PR #14693, opened 2018-11-30
Fix for accuracy problem for inplace operators when MKL-DNN mode is enabled

kbinias:
This fix eliminates accuracy problem for inplace operators (e.g. ReLu) when MKL-DNN mode is enabled

test=develop


luotao1 (approved):
LGTM！

PR #14697, opened 2018-12-01
Revert "Fix for accuracy problem for inplace operators when MKL-DNN mode is enabled"

luotao1:
Reverts PaddlePaddle/Paddle#14693
The reason is that object_detection fails on http://ce.paddlepaddle.org:8080/viewLog.html?buildId=5532&buildTypeId=PaddleCe_CEBuild&tab=buildChangesDiv


wopeizl (approved):

kbinias:
It looks the problem is with training on 4 GPUs (task object_detection/four_card_train_speed)
"details: [four_card_train_speed] failed, diff ratio: 0.025563791471089718 larger than 0.02"

Solution #14721

PR #14713, opened 2018-12-03
add more files to protected file list

panyx0718:
test=develop

### DIFF
--- a/paddle/scripts/paddle_build.sh
+++ b/paddle/scripts/paddle_build.sh
@@ -497,7 +497,9 @@ function assert_api_spec_approvals() {
                "paddle/fluid/framework/var_desc.h"
                "paddle/fluid/framework/scope.h"
                "paddle/fluid/framework/ir/node.h"
-               "paddle/fluid/framework/ir/graph.h")
+               "paddle/fluid/framework/ir/graph.h"
+               "paddle/fluid/framework/framework.proto"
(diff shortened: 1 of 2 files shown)


paddle/scripts/paddle_build.sh
```diff
@@ -487,7 +487,17 @@ function assert_api_spec_approvals() {
         BRANCH="develop"
     fi
 
-    API_FILES=("paddle/fluid/API.spec" "paddle/fluid/framework/operator.h")
+    API_FILES=("paddle/fluid/API.spec"
+               "paddle/fluid/framework/operator.h"
+               "paddle/fluid/framework/tensor.h"
+               "paddle/fluid/framework/lod_tensor.h"
+               "paddle/fluid/framework/selected_rows.h"
+               "paddle/fluid/framework/op_desc.h"
```
typhoonzero, line 495: how about adding `framework.proto` to replace `xx_desc.h`?

typhoonzero: also `send_recv.proto.in`

panyx0718: Added 2 more files. And keep xx_desc.h since they are also very important.

typhoonzero (approved):

qingqing01 (approved):
LGTM.

PR #14718, opened 2018-12-03
Fix the bug in gradient clip when the gradient is sparse

chengduoZH:


python/paddle/fluid/clip.py
```diff
@@ -271,7 +271,7 @@ def _process_context(self, context, param, grad):
                     "All parameters' 'clip_norm' of a same group should be the same"
                 )
 
-        square = grad * grad
+        square = layers.square(grad)
```
chengduoZH: In some models, the grad is sparse, but the `sparse * sparse` is meaningless, so I replace with `square`. here.

paddle/fluid/operators/activation_op.h
```diff
@@ -50,16 +56,38 @@ class ActivationKernel
...
+
+    if (CanBeUsedBySelectedRows.count(context.op().Type())) {
+      X = detail::Ref(
+          paddle::framework::GetLoDTensorOrSelectedRowsValueFromVar(*x_var),
+          "Cannot get input Tensor X, variable name = %s",
+          context.op().Input("X"));
+      Out = paddle::framework::GetMutableLoDTensorOrSelectedRowsValueFromVar(
+          out_var);
+    } else {
+      X = detail::Ref(context.Input<framework::Tensor>("X"),
```
panyx0718, line 78: Does this throw error if X is selected rows (Cannot be used by SelectedRows.

panyx0718 (commented):
API is fine. Some comments

python/paddle/fluid/clip.py
```diff
@@ -271,7 +271,12 @@ def _process_context(self, context, param, grad):
...
                 )
 
-        square = grad * grad
+        merge_grad = grad
+        if grad.type == core.VarDesc.VarType.SELECTED_ROWS:
+            merge_grad = layers.merge_selected_rows(grad)
+            merge_grad = layers.get_tensor_from_selected_rows(merge_grad)
+
+        square = layers.square(merge_grad)
         local_norm_var = layers.reduce_sum(input=square)
```
chengduoZH, line 280: `reduce_sum ` only accepts `LoDTensor`, so, if the grad is SELECTED_ROWS, there should use `get_tensor_from_selected_rows`.

paddle/fluid/operators/activation_op.h
chengduoZH: yup

chengduoZH: Yup

panyx0718 (approved):

paddle/fluid/operators/merge_selected_rows_op.cc
```diff
@@ -0,0 +1,72 @@
...
+             "The input type is SelectedRows, and the selected rows may be "
+             "duplicated.");
+    AddOutput("Out",
+              "The output type is SelectedRows, and the selected rows are not "
+              "duplicated.");
+    AddComment(
+        R"DOC(
+MergeSelectedRows Operator.
+
+MergeSelectedRows is used to merge the duplicated rows of the input.
```
qingqing01, line 46: Better to describe clearly how to merge. (add duplicated rows ?)

qingqing01 (commented):
Only reviewed the API.  LGTM.

paddle/fluid/operators/merge_selected_rows_op.cc
chengduoZH: Thanks, I will add more clear description in next PR.

typhoonzero (approved):
Sometimes people may think "merge" should have two inputs instead of one, it's actually doing dedup and sum. Since the internal implement is called "MergeAdd" I think it's fine to keep the name.

chengduoZH:
@typhoonzero Thanks, I will fix in next PR

PR #14727, opened 2018-12-04
Add run_startup_program() and save_model() for AsyncExecutor

wangguibao:


panyx0718 (changes requested):
I don't think this is necessary

PR #14728, opened 2018-12-04
Optimize hs op

JiabinYang:


paddle/fluid/operators/math/matrix_bit_code.cc
```diff
@@ -115,23 +116,34 @@ template <typename T>
...
+      GetBlas<platform::CPUDeviceContext, T>(platform::CPUDeviceContext());
   size_t num_samples = tmat.dims()[0];
   size_t input_width = input.dims()[1];
   size_t tmat_width = tmat.dims()[1];
   size_t weight_width = weight->dims()[1];
   auto tmat_value = tmat.data<T>();
   auto weight_value = weight->data<T>();
   auto input_value = input.data<T>();
+
+  std::unordered_map<int, std::vector<std::pair<T, const T*>>> ops;
```
jacquesqiao, line 129: `ops` is strange, what is the meaning?

jacquesqiao (approved):
LGTM!

paddle/fluid/operators/math/matrix_bit_code.cc
JiabinYang: this means operates in matrix calculate I will add more comments on that

PR #14732, opened 2018-12-04
[MKL-DNN] Concat Layer

Sand3r-:
Introduces the MKL-DNN implmentation of Concat Layer.
Provides **2.16x** speedup for whole topology on batch 128, 20 threads on GoogLeNet when tested against reference implementation.

### DIFF
--- a/paddle/fluid/operators/concat_mkldnn_op.cc
+++ b/paddle/fluid/operators/concat_mkldnn_op.cc
@@ -12,6 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */
 
+#include <memory>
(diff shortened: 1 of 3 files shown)


Sand3r-:
Waiting for #14775 to be resolved. Otherwise fixing the `test_concat_mkldnn_op` (which fails the TC build) can hardly be accomplished.

luotao1:
Thanks for this PR! 
Provides 2.7x speedup for concat op on batch 1, 4 threads on our detect model.

tensor-tang (commented):
@luotao1 Do we need follow this https://github.com/PaddlePaddle/Paddle/pull/14742 kernel register.

tensor-tang (approved):

luotao1:
@tensor-tang We can merge this PR at first, and if this PR need to follow kernel register, maybe could be down in next PR.

PR #14745, opened 2018-12-05
Fix eigen deallocate bug

sneaxiy:


paddle/fluid/platform/device_context.cc
```diff
@@ -127,9 +127,7 @@ class EigenCudaStreamDevice : public Eigen::StreamInterface {
     return retv;
   }
 
-  void deallocate(void* buffer) const override {
-    allocations_.erase(allocations_.find(buffer));
-  }
+  void deallocate(void* buffer) const override { allocations_.erase(buffer); }
```
chengduoZH: You should check whether `buffer` is nullptr here.

chengduoZH (dismissed):
Excellent

chengduoZH (approved):
Excellent

PR #14748, opened 2018-12-05
Refine merge_selected_rows Doc

chengduoZH:


paddle/fluid/framework/selected_rows.h
```diff
@@ -48,6 +48,11 @@ class SelectedRows {
  public:
   SelectedRows(const std::vector<int64_t>& rows, const int64_t& height)
       : rows_(rows), height_(height) {
+    for (auto row : rows_) {
```
panyx0718: This enforce can be expensive?

python/paddle/fluid/tests/unittests/test_get_tensor_from_selected_rows_op.py
```diff
@@ -29,7 +29,7 @@ def get_places(self):
 
     def check_with_place(self, place):
         scope = core.Scope()
-        x_rows = [0, 5, 5, 4, 20]
+        x_rows = [0, 5, 5, 4, 19]
```
panyx0718, line 32: why change this?

chengduoZH: If the height is 20, the value of x_rows should be 0~19.

paddle/fluid/framework/selected_rows.h
chengduoZH: No sure, but I think we should add this check, otherwise, the element value of rows may be out of the bound.

panyx0718: Leave this to the caller? feels expensive to me.

chengduoZH: Done

panyx0718 (approved):

PR #14762, opened 2018-12-05
fix bug of trt pool2d converter

NHZlX:


paddle/fluid/inference/tensorrt/convert/pool2d_op.cc
```diff
@@ -109,8 +109,8 @@ class Pool2dOpConverter : public OpConverter {
     }
 
     if (pool_type == "max") {
-      nvinfer1::DimsHW pre_pad(paddings[0], paddings[1]);
-      nvinfer1::DimsHW post_pad(paddings[0], paddings[1]);
+      nvinfer1::DimsHW pre_pad(0, 0);
```
Superjomn, line 116: Add more comments here to describe why the right way works here.

Superjomn (approved):
LGTM

Issue #14612, opened 2018-11-27
[Need to discuss]: Add more flexibility to pass for MKLDNN kernel


bingyanghuang:
For now, if we want to enable one or several ops' mkldnn kernel not all the ops who have mkldnn kernel support in C-API model, the way we use now is to add the specific pass to enable these kinds of ops. However, this way has poor scalability,  we need add different passes for different models. So, we'd like to propose whether
1. we could add some parameters when we add the new pass , i.e. we can create a general pass named enable_specific_mkldnn with parameters "graph" and a "list" containing the op names that the user want to enable their mkldnn kernel. When the user wants to enable one or two specific mkldnn ops, he can add this pass in his model with specific ops' names.
2. or let the pass function know the predictor configures informations  This is simpler way by following exist architecture. We can pass a list of op type names with predictor config, if pass mode is to enable specific mkldnn op kernel, the pass will read this list and change these specific ops' attribute "use_mkldnn" to "true".
-------------------------------------------------------------------------------------------------------------------
以防我英文没有说清楚，再用中文说一遍。
现在我们遇到了一个问题就是，无法在model的C-API端单独使能某几个op的mkldnn kernel， 因为一旦设置mkldnn的全局变量，就会使能所有支持mkldnn kernel op。基于现有框架可以实现的是，增加pass去使能我们所需要使能mkldnn的op，这就意味着对于每一个不同的model和不同的情况下我们都要创建不同的pass， 这样一来，可扩展性和可维护性就很差。我们初步讨论后，提出两种方案希望可以与你们在周三的例会上讨论一下。
1. 在创建pass的时候可以传入更多的参数。现有的pass实现下，传入给pass的参数为Graph，是否考虑再传入一个数组，这个数组里面包含了我希望更改的op的名字，这个数组的传入是可以由用户端定义和传入的。这样，在执行这个pass的时候就可以根据这个传入的数组进行特定的图修改；
2. 用现有的config这个属性传入一个数组，在pass的函数中可以读取predictor的config这个属性。现在我们在创建predictor的时候，用户可以通过config来传入一些他所需要的设置，我们是否可以也通过config传入一个我们需要更改的op名字的数组，然后在pass函数中可以读取到这个config的信息来单独设置某几个op的“use_mkldnn”为true。
-------------------------------------------------------------------------------------------------------------------
This is a general description for mkldnn enabling in C-API, for more details, we'd like to discuss with Baidu in our weekly meeting.

[...]

PR #14767, opened 2018-12-06
One possible solution to add flexibility for mkldnn placement pass

bingyanghuang:
#14612 This PR proposed one solution to selectively turn on "use_mkldnn" attribute in mkldnn_placement_pass. 
After investigation, we found that we can add attribute to each pass. So we decided to pass a string vector named op_list which contains all the op names we want to enable by this attribute of pass. 
You can add the following code to test this functionality in line 177.  
``` C++
 if (use_mkldnn) {    
      cfg.EnableMKLDNN();  
      std::vector<std::string> op_list ={"conv3d"};  
      cfg.SetMKLDNNOp(op_list);  
 }
```
https://github.com/PaddlePaddle/Paddle/blob/e90afec47b0498e03950369323e4f85335c406f5/paddle/fluid/inference/tests/api/analyzer_dam_tester.cc#L177

This test will enable just conv3d  mkldnn kernel in DAM model. Meanwhile, you can test it with resnet50 to make sure that this works well with resnet50 too.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn_placement_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn_placement_pass.cc
@@ -18,12 +18,20 @@ namespace paddle {
 namespace framework {
 namespace ir {
 
+
 std::unique_ptr<ir::Graph> MKLDNNPlacementPass::ApplyImpl(
     std::unique_ptr<ir::Graph> graph) const {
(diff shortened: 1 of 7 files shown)


bingyanghuang:
@luotao1 pls help remove this PR. I closed this PR because there is wrong git user name in the commit. I've created another new PR.

PR #14778, opened 2018-12-06
Async executor bugfix: Tensor changed to LoDTensor

wangguibao:


paddle/fluid/framework/data_feed.cc
```diff
@@ -348,36 +346,25 @@ void MultiSlotDataFeed::PutToFeedVec(
...
-      }
+      int64_t* tensor_ptr = feed_vec_[i]->mutable_data<int64_t>(
+          {total_instance, 1}, platform::CPUPlace());
+      memcpy(tensor_ptr, &feasign[0], total_instance * sizeof(int64_t));
+    }
+
+    LoD data_lod{offset};
+    feed_vec_[i]->set_lod(data_lod);
+    if (use_slots_is_dense_[i]) {
+      int dim = total_instance / batch_size_;
```
guru4elephant, line 366: seems buggy here. Is batch_size_ a fixed value or a runtime value?

guru4elephant: Oh，I see. batch_size_ is dynamic. default_batch_size_ is set by users.

guru4elephant (approved):
The code seems more clear to me than before, thanks for your pr!

PR #14795, opened 2018-12-07
Fix numpy bug on Ubuntu16 and Ubuntu18 which will cause segmentfault 

velconia:


python/paddle/dataset/image.py
```diff
@@ -32,11 +32,28 @@
...
-    import cv2
-except ImportError:
-    cv2 = None
+# FIXME(minqiyang): this is an ugly fix for the numpy bug reported here
+# https://github.com/numpy/numpy/issues/12497
+if six.PY3:
+    import subprocess
+    import sys
+    import_cv2_proc = subprocess.Popen(
+        [sys.executable, "-c", "import cv2"],
```
junjun315, line 43: `sys.executable` may not run under Windows

junjun315: forget this

junjun315 (approved):
LGTM

PR #14799, opened 2018-12-07
Fix/remove unnecessary mkldnn reorder

ghost:
When data flow from a MKLDNN OP kernel into a non-MKLDNN OP kernel, data layout transform will be done by framework automatically (see data_transform.cc & data_layout_transform.cc) via doing a MKLDNN reorder. 
However, when those two OP kernels share the same data layout (or "memory format" more accurately), this reorder should NOT be done.

@tensor-tang @tpatejko @kbinias Please help to review. Thanks!


luotao1 (approved):
LGTM

paddle/fluid/framework/data_layout_transform.cc
```diff
@@ -151,19 +151,22 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
...
 
-  platform::Reorder(in_memory, out_memory);
+    auto in_memory =
+        memory({{{in_tz}, in_type, in_format}, cpu_engine}, in_data);
+    auto out_memory =
+        memory({{{out_tz}, out_type, out_format}, cpu_engine}, out_data);
 
+    platform::Reorder(in_memory, out_memory);
+  } else {
+    out->ShareDataWith(in);
```
tensor-tang, line 168: Can we use like this?

http://paddlepaddle.org/documentation/docs/zh/1.2/advanced_usage/development/new_op/op_notes.html

tensor-tang: Synced with @chengduoZH this is fine.

tensor-tang (approved):

PR #14814, opened 2018-12-10
Add gperftools supports for PE

reyoung:


cmake/FindGperftools.cmake
```diff
@@ -0,0 +1,63 @@
+# Tries to find Gperftools.
```
reyoung, line 1: Shameless copy from https://github.com/vast-io/vast/blob/master/cmake/FindGperftools.cmake

Hope they do not mind.

cmake/generic.cmake
```diff
@@ -110,6 +110,14 @@ function(find_fluid_modules TARGET_NAME)
   endif()
 endfunction(find_fluid_modules)
 
+
+function(common_link TARGET_NAME)
+  if (WITH_PROFILER)
+    target_link_libraries(${TARGET_NAME} gperftools::profiler)
```
reyoung, line 116: Extract a function called `common_link`, because it might be more common libraries to link to all our target.

paddle/fluid/framework/parallel_executor.cc
```diff
@@ -270,6 +290,12 @@ void ParallelExecutor::BCastParamsToDevices(
 
 void ParallelExecutor::Run(const std::vector<std::string> &fetch_tensors,
                            const std::string &fetched_var_name) {
+#ifdef WITH_GPERFTOOLS
+  if (gProfileStarted) {
+    ProfilerFlush();
```
reyoung, line 298: Using `ProfilerFlush` instead of `ProfilerStop` because we always kill our program instead of waiting it done. So we can kill our program and generate profile result by using `ProfilerFlush`.

sneaxiy (dismissed):
Excellent

paddle/fluid/framework/parallel_executor.cc
```diff
@@ -30,13 +30,36 @@ limitations under the License. */
 #include "paddle/fluid/framework/details/threaded_ssa_graph_executor.h"
 #include "paddle/fluid/platform/profiler.h"
 
+#ifdef WITH_GPERFTOOLS
+#include "gperftools/profiler.h"
+#endif
+DEFINE_string(pe_profile_fname, "",
+              "Profiler filename for PE, which generated by gperftools."
+              "Only valid when compiled `WITH_PRIFILER=ON`. Empty if disable.");
```
velconia, line 38: should be ```WITH_PROFILER``` here

reyoung: No. Always define pe_profile_fname to keep API same.

sneaxiy (approved):
Excellent

PR #14826, opened 2018-12-10
add flag. test=develop

dzhwinter:
fix inference model bug.  if optimize flag is turn on, the program should not be saved.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -1452,6 +1452,10 @@ class Program(object):
         # maybe same, which conflict with save_inference_model.
         return self._is_optimized
 
+    @is_optimized.setter
+    def set_is_optimized(self, target):
+        self._is_optimized = target
+
     @property
     def op_role(self):
         """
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/io.py
```diff
@@ -643,6 +643,9 @@ def save_inference_model(dirname,
 
     if main_program is None:
         main_program = default_main_program()
+        if main_program.is_optimized:
+            raise RuntimeError(
+                "save_inference_model must put before you call memory_optimize.")
```
Superjomn: Better to add more description for users, such as

"save_inference_model must put before you call memory_optimize."
 "the memory_optimize will modify the original program, and that is not suitable for inference"

python/paddle/fluid/framework.py
```diff
@@ -1444,6 +1444,17 @@ def __init__(self):
         self._endpoints = []
         self._trainers_endpoints = []
         self._distributed_lookup_table = None
+        self._is_optimized = False
+
+    @property
+    def is_optimized(self):
```
panyx0718, line 1450: make these private? I assume they will be removed after pass migration?

PR #14829, opened 2018-12-10
Accelerate little models

velconia:
Take DDPG of RL as an example, before this PR:

| Tables        | ensemble_num=1           | ensemble_num=12  |
| ------------- |:-------------:| -----:|
| TF      | 0.48 | 4.25 |
| Paddle | 1.11 | 12.84 |

after this PR:

| Tables        | ensemble_num=1          | ensemble_num=12  |
| ------------- |:-------------:| -----:|
|  TF     | 0.48 | 4.25 |
| Paddle | 0.49 | 4.84 |


paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
```diff
@@ -64,21 +64,24 @@ FeedFetchList ScopeBufferedSSAGraphExecutor::Run(
   }
 
   platform::RecordEvent e("ScopeBufferedSSAGraphExecutorAfterRun", nullptr);
-  drop_scope_counter_ += 1;
+  ++drop_scope_counter_;
 
-  if (!fetch_tensors.empty() ||
-      drop_scope_counter_ == strategy_.num_iteration_per_drop_scope_) {
```
panyx0718, line 70: I guess this was added on purpose? Removing it might not be right.

paddle/fluid/framework/ir/graph.cc
```diff
@@ -44,27 +48,56 @@ void CheckProgram(const ProgramDesc &program) {
         break;
       case _INT(OpRole::kBackward):
       case _INT(OpRole::kBackward) | _INT(OpRole::kLoss):
-        PADDLE_ENFORCE(
-            visit.find(_INT(OpRole::kOptimize)) == visit.end(),
-            "Cannot add backward operator %s after optimize operator.",
-            op->Type());
+        if (!FLAGS_enforce_when_check_program) {
```
panyx0718: Update all of them to LOG ERROR and remove the flag?

paddle/fluid/framework/rw_lock.h
```diff
@@ -51,81 +53,44 @@ struct RWLock {
...
-  }
+  ~AutoWRLock() { UnLock(); }
+
+ private:
+  RWLock* lock_;
+};
+
+class AutoRDLock {
+ public:
+  explicit AutoRDLock(RWLock* rw_lock) : lock_(rw_lock) { Lock(); }
```
panyx0718, line 85: It might not be a good idea to mix lock_guard with normal lock. User can create AutoRDLock and call Lock. Can you split them?

paddle/fluid/framework/spin_lock.h
```diff
@@ -0,0 +1,71 @@
...
+#include <mutex>  // NOLINT
+#endif            // !_WIN32
+
+#include "paddle/fluid/platform/enforce.h"
+
+namespace paddle {
+namespace framework {
+
+#if !defined(_WIN32)
+struct SpinLock {
```
panyx0718: Is this used or tested? If not, don't add it?

paddle/fluid/framework/spin_lock.h
```diff
@@ -0,0 +1,71 @@
+/* Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
typhoonzero: this lock seems not used in this PR

paddle/fluid/framework/details/execution_strategy.h
```diff
@@ -25,7 +25,7 @@ struct ExecutionStrategy {
   size_t num_threads_{0};
   bool use_cuda_{true};
   bool allow_op_delay_{false};
-  size_t num_iteration_per_drop_scope_{100};
+  size_t num_iteration_per_drop_scope_{1};
```
panyx0718, line 28: why make this change?

paddle/fluid/framework/scope.h
```diff
@@ -124,7 +141,8 @@ class Scope {
   DISABLE_COPY_AND_ASSIGN(Scope);
 
  private:
-  mutable std::mutex mutex_;
+  mutable RWLock kids_lock_;
+  mutable RWLock vars_lock_;
```
typhoonzero, line 141: for discussion: We might need to have a new class `ThreadsafeScope` inherit from `Scope`. In cases when locking is not needed, they can use the parent class.

velconia: ```ON_INFER``` macro free all lock and lock is needed now... because a few ops use Scope::Var and Scope::Rename to rename it...

paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
```diff
@@ -64,21 +64,24 @@ FeedFetchList ScopeBufferedSSAGraphExecutor::Run(
...
-    drop_scope_counter_ = 0;
+  if (!fetch_tensors.empty()) {
     // Wait All computational streams
     for (auto p : places_) {
       platform::DeviceContextPool::Instance().Get(p)->Wait();
     }
+  }
+
+  if (drop_scope_counter_ == strategy_.num_iteration_per_drop_scope_) {
+    drop_scope_counter_ = 0;
```
chengduoZH: Before dropping the local scopes, you should call `Wait()`.

paddle/fluid/framework/rw_lock.h
velconia: ok, I will change AutoRDLock to RDLockGuard

paddle/fluid/framework/spin_lock.h
velconia: It's tested, but not used, I will remove it later.

paddle/fluid/framework/details/execution_strategy.h
velconia: because the original ```num_iteration_per_drop_scope_``` didn't take effect, and ```Scope``` was deleted every time ```executor.run``` end, so we should change to 1 here to make sure we get the same reaction

paddle/fluid/framework/spin_lock.h
velconia: It's tested, but not used, I will remove it later.

paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
```diff
@@ -64,21 +64,30 @@ FeedFetchList ScopeBufferedSSAGraphExecutor::Run(
...
     // Wait All computational streams
     for (auto p : places_) {
       platform::DeviceContextPool::Instance().Get(p)->Wait();
     }
+  }
+
+  if (drop_scope_counter_ == strategy_.num_iteration_per_drop_scope_) {
+    // Wait All computational streams
+    for (auto p : places_) {
+      platform::DeviceContextPool::Instance().Get(p)->Wait();
```
panyx0718: will be called twice (once above) since since num_iteration_per_drop_scope_=1?

paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
```diff
@@ -64,21 +64,30 @@ FeedFetchList ScopeBufferedSSAGraphExecutor::Run(
...
-      drop_scope_counter_ == strategy_.num_iteration_per_drop_scope_) {
-    drop_scope_counter_ = 0;
+  if (!fetch_tensors.empty()) {
     // Wait All computational streams
     for (auto p : places_) {
       platform::DeviceContextPool::Instance().Get(p)->Wait();
     }
+  }
+
+  if (drop_scope_counter_ == strategy_.num_iteration_per_drop_scope_) {
```
panyx0718, line 75: @chengduoZH @sneaxiy any comment about this change?

paddle/fluid/framework/scope.h
typhoonzero: So a base class can be used for all cases like inference with no locks, and child class can be used in threaded executors.

typhoonzero: Oh, can do it in a single pull request though

velconia: Good idea, I will fix it up in follow up PRs

chengduoZH (commented):
You should write the performance report in this PR, **for which model and the speed up data**.
It is important and useful, because reviewer can know performance improvement directly, and others can see how much performance improvement this PR brings.

panyx0718 (dismissed):

panyx0718 (dismissed):

panyx0718:
@chengduoZH @sneaxiy please comment or approve.

typhoonzero (dismissed):
LGTM, maybe @chengduoZH need to check changes for `scope_buffered_ssa_graph_executor.cc`

paddle/fluid/framework/ir/graph.cc
```diff
@@ -44,27 +48,56 @@ void CheckProgram(const ProgramDesc &program) {
...
+          PADDLE_ENFORCE(visit.find(_INT(OpRole::kBackward)) != visit.end(),
+                         "Optimize operators %s must follow backward operator.",
+                         op->Type());
+        } else {
+          if (visit.find(_INT(OpRole::kBackward)) == visit.end()) {
+            LOG(ERROR) << "Optimize operators %s must follow backward operator."
+                       << op->Type();
+          }
+        }
```
chengduoZH: You should use the flag to skip [`CheckProgram`](https://github.com/PaddlePaddle/Paddle/pull/14829/files#diff-2e5c77597541abca34ee662f78464558R118) directly.

paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
chengduoZH: It looks ok.

panyx0718 (dismissed):

panyx0718 (approved):

PR #14836, opened 2018-12-10
Featue/py_func op

sneaxiy:
To support self-defined Python Op, something like `tf.py_func`. But unlike `tf.py_func`, we support backward registration too.


paddle/fluid/operators/py_func_op.cc
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+
+class PyFuncOpShapeInference : public framework::InferShapeBase {
+ public:
+  void operator()(framework::InferShapeContext *ctx) const override {
+    PADDLE_ENFORCE(!ctx->IsRuntime(),
+                   "Infer shape cannot be called in runtime.");
+    PADDLE_ENFORCE(ctx->HasInputs("X") || ctx->HasOutputs("Out"),
+                   "Input(X) or Output(Out) must exist");
```
chengduoZH: Why the Input(X) or Output(Out) must exist but not both of them must exist.?

paddle/fluid/operators/py_func_op.cc
```diff
@@ -0,0 +1,285 @@
...
+    auto &in_arg_names = Inputs("X");
+    auto &out_arg_names = Outputs("Out");
+
+    std::vector<framework::LoDTensor> inputs(in_arg_names.size());
+    for (size_t i = 0; i < in_arg_names.size(); ++i) {
+      auto in_var = scope.FindVar(in_arg_names[i]);
+      if (in_var == nullptr) {
+        continue;
+      }
+      auto &in_tensor = in_var->Get<framework::LoDTensor>();
```
chengduoZH, line 279: Why the `in_var ` can be nullptr?
Maybe it is confusing in the code that you use `PyFuncOp` to run the forward and backward.

paddle/fluid/operators/py_func_op.cc
sneaxiy: Because it is more flexible to support Python func with no inputs or no outputs. For example, users want to print some variables using Python func with no outputs, or users want to init some variables using Python func with no inputs.

paddle/fluid/operators/py_func_op.cc
sneaxiy: `in_var` would be nullptr only when running backward. It is a little confused to mix the codes of forward and backward, but less codes are needed when mixing them. Moreover, when py_func_op is the backward op of py_func_op, it is easy to append second order derivatives if necessary in the future.

reyoung (dismissed):

python/paddle/fluid/layers/nn.py
```diff
@@ -9125,6 +9128,183 @@ def get_tensor_from_selected_rows(x, name=None):
...
+
+    Input orders of :code:`backward_func` would be: forward inputs
+    :code:`x`, forward outputs :code:`out` and backward input gradients of
+    :code:`out`. If some variables of :code:`out` have no gradient, the input
+    tensor would be None in Python side. If some variables of :code:`in` have
+    no gradient, users should return None.
+
+    This function can also be used to debug the running network. User can
+    add a :code:`py_func` operator without output, and print input 
+    :code:`x` inside :code:`func`.
```
panyx0718, line 9228: Can give 2~3 code examples here.

paddle/fluid/operators/py_func_op.cc
```diff
@@ -0,0 +1,285 @@
...
+#include <vector>
+#include "Python.h"
+#include "paddle/fluid/framework/op_registry.h"
+
+namespace paddle {
+namespace operators {
+
+namespace py = ::pybind11;
+
+static std::vector<py::object> g_py_callables;
```
panyx0718, line 27: Can we put this in Scope?

paddle/fluid/framework/shape_inference.h
```diff
@@ -25,7 +25,10 @@ limitations under the License. */
 namespace paddle {
 namespace framework {
 
+class OperatorBase;
+
 using InferShapeVarPtr = boost::variant<VarDesc *, Variable *>;
+using InferShapeOpPtr = boost::variant<const OpDesc *, const OperatorBase *>;
```
panyx0718: why make it variant? Seems only OpDesc is used?

paddle/fluid/operators/py_func_op.cc
```diff
@@ -0,0 +1,285 @@
...
+
+  std::vector<std::unique_ptr<framework::OpDesc>> operator()() const override {
+    auto &fwd_attrs = Attrs();
+    // no backward op when backward_id is less than 0
+    if (boost::get<int>(fwd_attrs.at(kBackwardPythonCallableId)) < 0) {
+      return {};
+    }
+
+    std::unique_ptr<framework::OpDesc> grad_op(new framework::OpDesc());
+    grad_op->SetType("py_func");
```
panyx0718, line 207: why not py_func_grad?

paddle/fluid/operators/py_func_op.cc
panyx0718: supporting higher order derivatives is an interesting question. Can we still do that with a seprate py_func_grad_op?

paddle/fluid/operators/py_func_op.cc
```diff
@@ -0,0 +1,285 @@
...
+    PADDLE_ENFORCE(ctx->HasInputs("X") || ctx->HasOutputs("Out"),
+                   "Input(X) or Output(Out) must exist");
+    PADDLE_ENFORCE_GE(ctx->Attrs().Get<int>(kForwardPythonCallableId), 0,
+                      "Function id cannot be less than 0");
+
+    // Transverse all outputs
+    // If name of any output ends with @GRAD,
+    // set its shape, dtype, lod_level, type to be the same as
+    // the correponding forward variable
+    auto *op = boost::get<const framework::OpDesc *>(ctx->GetOp());
```
panyx0718: why you need to fetch the op and block to set dim? It seems InferShapeContext has SetDim methods. Is it possible to simplify this InferShape?

paddle/fluid/operators/py_func_op.cc
sneaxiy: Maybe it is not a good idea to put it in Scope. Maintaining Python Object in C++ may be a risk. `tf.py_func` set Python object to be global objects too.

paddle/fluid/framework/shape_inference.h
sneaxiy: Yes. But `InferShapeContext` is an abstract class and `CompileTimeInferShapeContext` is a class hidden in source file. `Variant` makes it more general.

paddle/fluid/operators/py_func_op.cc
sneaxiy: As comment in https://github.com/PaddlePaddle/Paddle/pull/14836#discussion_r242010867, support high order derivatives.

paddle/fluid/operators/py_func_op.cc
sneaxiy: Because some forward inputs may not be needed in `backward_func`. In this case, `InferShapeContext` would not contain these unnecessary variables.

python/paddle/fluid/layers/nn.py
sneaxiy: OK. I would add some examples.

paddle/fluid/operators/py_func_op.cc
```diff
@@ -0,0 +1,310 @@
...
+    grad_op->SetInput("X", bwd_ins);
+    grad_op->SetOutput("Out", bwd_outs);
+
+    std::vector<std::unique_ptr<framework::OpDesc>> ret(1);
+    ret[0] = std::move(grad_op);
+    return ret;
+  }
+};
+
+class PyFuncOp : public framework::OperatorBase {
```
panyx0718, line 262: Should there be a PyFuncKernel that only supports CPU and this inherits OpWithKernel? The framework has built-in support to transfer data. adding again seems duplicated.

paddle/fluid/operators/py_func_op.cc
panyx0718: This optimization might not bit quite necessary at this moment.
With this GetOp, develop can use it to get any vars from block or program, which seems to open. I'd prefer to not exposing it now.

paddle/fluid/operators/py_func_op.cc
```diff
@@ -0,0 +1,310 @@
...
+      auto *out_var = scope.FindVar(out_arg_names[i]);
+      outputs[i] =
+          out_var ? out_var->GetMutable<framework::LoDTensor>() : nullptr;
+    }
+
+    auto callable_id = static_cast<size_t>(Attr<int>(kForwardPythonCallableId));
+    auto *py_callable = GetPythonCallableObject(callable_id);
+    VLOG(10) << "Call Python function with id " << callable_id << ": "
+             << PythonFuncDebugString(*py_callable);
+    CallPythonFunc(py_callable, inputs, &outputs);
```
panyx0718, line 302: will data be copied back to gpu? this should be a kernel and let framework handle it?

sneaxiy: No need to copy data back to GPU in py_func op. Framework would do that.

paddle/fluid/operators/py_func_op.cc
sneaxiy: OpWithKernel needs to know the data type of inputs. But it is possible that data types of input tensors is not the same.

panyx0718 (approved):

reyoung (approved):
LGAPI

PR #14842, opened 2018-12-11
Fix Eigen macro when using GPU

reyoung:
The macro should be defined by compiler rather than by source.

test=develop


sneaxiy (approved):
Cool

PR #14847, opened 2018-12-11
Fix the issue to run on AVX2 machine

yihuaxu:
The current status is that the AVX kernel can not be entered when run on AVX2 kernel, so add the workaround to ensure the AVX instructions' running.


luotao1:
please squash your commit log.

yihuaxu:
Recreate the new branch for this commit

PR #14851, opened 2018-12-11
Fix the issue to run on AVX2 machine

yihuaxu:
The current status is that the AVX kernel can not be entered when run on AVX2 kernel, so add the workaround to ensure the AVX instructions' running.


yihuaxu:
Start a review

luotao1:
Thanks for your PR. I test it on the small_dam model (batchsize=1 on E5-2620 v3 machine), the elapsed time is from 11.94ms to 11.087ms, about 7% speedup.

tensor-tang (approved):
Thanks @yihuaxu

PR #14858, opened 2018-12-12
Feature/tensor type

reyoung:


paddle/fluid/framework/data_layout_transform.cc
```diff
@@ -101,7 +101,7 @@ void* GetDataFromTensor(const Tensor& tensor, mkldnn::memory::data_type type) {
     case mkldnn::memory::data_type::f32:
       return platform::to_void_cast(tensor.data<float>());
     case mkldnn::memory::data_type::s8:
-      return platform::to_void_cast(tensor.data<char>());
```
reyoung, line 104: char is not always signed on all platform.

paddle/fluid/framework/data_type.h
```diff
@@ -22,46 +22,59 @@ limitations under the License. */
 namespace paddle {
 namespace framework {
 
+template <typename T>
+struct DataTypeTrait {};
+
+// Stub handle for void
```
reyoung, line 28: Add DataTypeTrait for void because we allow the user invoke `tensor.data<void>()` and `tensor.mutable_data<void>`, we will never use `sizeof(void)` because it will be a shortcut like `std::is_same<void, T> || SizeOf(DataTypeTrait<void>::DataType)`. However, we must write this template specification here.

paddle/fluid/framework/data_type.h
```diff
@@ -22,46 +22,59 @@ limitations under the License. */
...
+template <>
+struct DataTypeTrait<void> {
+  constexpr static auto DataType = proto::VarType::RAW;
+};
+
+#define _ForEachDataTypeHelper_(callback, cpp_type, proto_type) \
+  callback(cpp_type, ::paddle::framework::proto::VarType::proto_type);
+
+#define _ForEachDataType_(callback)                                     \
```
reyoung, line 37: A useful macro to `ForEach` data types which supported by tensors.

paddle/fluid/framework/details/fuse_vars_op_handle.h
```diff
@@ -33,7 +33,7 @@ struct FuseVarsOpHandle : public OpHandleBase {
   FuseVarsOpHandle(ir::Node *node, Scope *local_scope,
```
reyoung, line 33: I wonder where do we use this class?

paddle/fluid/framework/dlpack_tensor_test.cc
```diff
@@ -91,23 +91,11 @@ void TestMainLoop() {
     }
   }
 }
+TEST(dlpack, test_all) {
```
reyoung, line 94: It seems that we cannot use `test_##cpp_type` in the nested macro. Just move all tests into one function.

paddle/fluid/framework/data_type.h
dzhwinter: It's surprise that gcc define the sizeof(void). Although it is not defined in cpp standards.
```cpp
echo 'int main() { printf("%d", sizeof(void)); }' | g++ -xc -w - && ./a.out
```
it gives 1.

dzhwinter (dismissed):
Excellent Job. 
I'm curious that why both stl and boost implement use a hash based string in type_index.
```cpp
176 inline std::size_t stl_type_index::hash_code() const BOOST_NOEXCEPT {
177 #if (defined(_MSC_VER) && _MSC_VER > 1600) || (__GNUC__ == 4 && __GNUC_MINOR__ > 5 && defined(__GXX_EXPERIMENTAL_CXX0X__))
178     return data_->hash_code();
179 #else
180     return boost::hash_range(raw_name(), raw_name() + std::strlen(raw_name()));
181 #endif
182 }
183
184
185 /// @cond
186
187 // for this compiler at least, cross-shared-library type_info
188 // comparisons don't work, so we are using typeid(x).name() instead.
189 # if (defined(__GNUC__) && (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 5))) \
190     || defined(_AIX) \
191     || (defined(__sgi) && defined(__host_mips)) \
192     || (defined(__hpux) && defined(__HP_aCC)) \
193     || (defined(linux) && defined(__INTEL_COMPILER) && defined(__ICC))
194 #  define BOOST_CLASSINFO_COMPARE_BY_NAMES
195 # endif
196
197 /// @endcond
198
199 inline bool stl_type_index::equal(const stl_type_index& rhs) const BOOST_NOEXCEPT {
200 #ifdef BOOST_CLASSINFO_COMPARE_BY_NAMES
201     return raw_name() == rhs.raw_name() || !std::strcmp(raw_name(), rhs.raw_name());
202 #else
203     return !!(*data_ == *rhs.data_);
204 #endif
205 }
206
```

paddle/fluid/framework/tensor.h
```diff
@@ -14,6 +14,7 @@ limitations under the License. */
 
 #pragma once
 
+#include <paddle/fluid/framework/framework.pb.h>
```
panyx0718: use `""`?

panyx0718 (dismissed):

paddle/fluid/framework/tensor.h
reyoung: Done.

panyx0718 (approved):

qingqing01 (approved):

PR #14862, opened 2018-12-12
better sort. experiment

dzhwinter:


paddle/fluid/framework/details/multi_devices_graph_pass.cc
```diff
@@ -240,14 +241,68 @@ size_t MultiDevSSAGraphBuilder::GetAppropriateDeviceID(
   return dev_id;
 }
 
+std::vector<ir::Node *> BFSSortOpFromFeed(const ir::Graph &graph) {
+  std::vector<ir::Node *> ret_ops;
+  std::unordered_set<ir::Node *> ready_vars;
+  std::deque<ir::Node *> queue;
+  // find feed entrance. hard core for experimental.
+  for (auto *node : graph.Nodes()) {
+    if (node->Name() == "image") {
```
dzhwinter, line 250: For testing in resnet50. It's a experiment code, in fact, we should mark the feed entrance or use feed op in program then start BFS or DFS from it.

PR #14871, opened 2018-12-12
add NewTmpScope

jacquesqiao:
skip the lock in scope, test with async_executor, speed up 1 more times.


paddle/fluid/operators/distributed/parameter_prefetch.cc
```diff
@@ -173,7 +173,7 @@ void prefetch(const std::string& id_name, const std::string& out_name,
               const std::vector<std::string>& epmap,
               const std::vector<int>& height_sections,
               const framework::ExecutionContext& context) {
-  auto& local_scope = context.scope().NewScope();
+  framework::Scope* local_scope = context.scope().NewTmpScope();
```
typhoonzero: Can use unique_ptr here

PR #14872, opened 2018-12-12
[MKL-DNN ]Added transpose/transpose2 Op

jczaja:
Changes proposed here are introducing Transpose/Transpose2 op executed with MKL-DNN reorder primitive. This is **first** out of  three PR's related to transpose Op.

**Performance:**
_Model:_ Transformer (https://github.com/PaddlePaddle/models/tree/develop/fluid/PaddleNLP/neural_machine_translation/transformer)
_Platform:_  Intel(R) Xeon(R) Platinum 8180 CPU @ 2.50GHz
_Threads:_ 1
_Results[overall model speedup]:_ ~2 times(BS:1) , ~6 times(BS:50)

**Notes:**
- Forward Op only was implemented (works if is_test == True))
- No any layout support yet (will come in third installment)
- No reusing of reorder primitives (will come in second installment)
- Other models may benefit from this optimization, but for full performance on convolutional models (eg. mostly mkl-dnn based) any format support has to be added (will happen in third installment of this series)

@luotao1: 
1.This PR does not deliver corressponding Grad Op , so there is added is_test and assertion if this op is used during traning. Is this proper way of handling situation where we implement only Froward Op  ?
2. Could you also check if you see improvment on your target use case of Transformer?

### DIFF
--- a/paddle/fluid/operators/transpose_mkldnn_op.cc
+++ b/paddle/fluid/operators/transpose_mkldnn_op.cc
@@ -1,0 +1,122 @@
+/* Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License. */
+
+#include "paddle/fluid/framework/data_layout_transform.h"
+#include "paddle/fluid/framework/op_registry.h"
+#include "paddle/fluid/memory/malloc.h"
+#include "paddle/fluid/platform/mkldnn_reuse.h"
+
+namespace paddle {
+namespace operators {
+
+using Tensor = framework::Tensor;
+using framework::DataLayout;
+
+template <typename T>
+class TransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
+ public:
+  void Compute(const paddle::framework::ExecutionContext& ctx) const override {
+    PADDLE_ENFORCE(paddle::platform::is_cpu_place(ctx.GetPlace()),
+                   "It must use CPUPlace.");
+    const bool is_test = ctx.Attr<bool>("is_test");
+    PADDLE_ENFORCE(
+        is_test == true,
+        "ConvTransposeMKLDNN works only for inference!. Set is_test = True");
+    auto& dev_ctx =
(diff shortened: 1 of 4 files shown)


luotao1:
>  Is this proper way of handling situation where we implement only Froward Op?

Yes, it's OK.

> Could you also check if you see improvement on your target use case of Transformer

We are preparing the Transformer CAPI now.  After it's ready, we will check the improvement.

However, I test this PR with our detect model. There is accuracy diff between MKLDNN and MKL. Do you check the accuarcy on SSD model?

jczaja:
@luotao1 I checked only accuracy on Transformer apart from unit tests. I just realised that lack of MKL_DNN Any format (that I planned to add in third PR) may have introduced problems in colvolutional models (including SSD). I will update this PR with fix

luotao1:
I test with the latest commit:
```
I1213 21:20:48.606076 62237 analysis_predictor.cc:345] == optimize end ==
I1213 21:20:48.606884 62237 tester_helper.h:179] Warm up run...
unknown file: Failure
C++ exception with description "std::exception" thrown in the test body.
```

jczaja:
@luotao1 Thanks very much for immediate testing my changes. I have reproduction of problem so as soon as I diagnose it I will update review

jczaja:
@luotao1 I updated PR with changes that were tested both on Transformer as well as MobileNet SSD (local CAPI test). Accuracy looked good. Please confirm that it works functionally for your tests and test performance if possible.

**Note:** currently Transformer does not work on develop when MKL-DNN is to be used eg. FLAGS_use_mkldnn=True. To test it please disable mkldnn_concat (remove concat_mkldnn_op.cc and run cmake.. , make etc.).

luotao1:
It works correctly on our detect model. And the transpose2 op speedup 6x (BS=1, threads=4, E5-2650 v4).

luotao1:
please resolve conflicts.

paddle/fluid/operators/transpose_mkldnn_op.cc
```diff
@@ -0,0 +1,124 @@
...
+                       const T* data_i, std::vector<int>& nchw_dims,
+                       const mkldnn::engine& eng) const {
+    // Make Memory primitive descriptors
+    auto mpd_o = mkldnn::memory::primitive_desc(md_o, eng);
+    auto mpd_i = mkldnn::memory::primitive_desc(md_i, eng);
+
+    auto data_o = output->mutable_data<T>(
+        place, paddle::memory::Allocator::kDefault, mpd_o.get_size());
+
+    auto src = mkldnn::memory(mpd_i, const_cast<T*>(data_i));
```
luotao1: It's not recommended to use `const_cast` (see #14897). Maybe you can use `static_cast` instead.

jczaja: I worked around this issue. But problem is that MKLDNN memory primitive is execting non-const memory allocation to work with. So currently I do not know how to make paddle's const inputs to work with mkl-DNN memory.

luotao1 (approved):
LGTM！

PR #14875, opened 2018-12-12
[Feature] Add Temporary Allocator

chengduoZH:
Add Temporary Allocator.
Some operator need temporary memory during computation, for example, conv_gemm, which needs `col` to store the result of `im2col`. If we create a stack memory which is used by CUDA Kernel, before the `Computation(...)` returns, we should add `ctx->Wait()`, because the execution of CUDA is async, if there doesn't have `ctx->Wait()`, the temporary memory will be released.


paddle/fluid/platform/CMakeLists.txt
```diff
@@ -78,3 +78,11 @@ IF(WITH_GPU)
   nv_test(cuda_helper_test SRCS cuda_helper_test.cu)
 ENDIF()
 nv_library(cuda_device_guard SRCS cuda_device_guard.cc DEPS gpu_info)
+
+IF(WITH_GPU)
+    nv_library(temp_allocator SRCS temporal_allocator.cc DEPS  allocator_facade)
+    nv_test(temporal_allocator_test SRCS temporal_allocator_test.cc DEPS temp_allocator)
+ELSE()
```
reyoung: Why use `nv_library` when GPU?

chengduoZH: The `temporal_allocator` contains `cudaStream_t` and `cudaStreamSynchronize`. so it needs use nvcc compile.

chengduoZH: Have fixed.

paddle/fluid/framework/tensor.cc
```diff
@@ -101,5 +105,12 @@ const DDim& Tensor::dims() const { return dims_; }
 
 int64_t Tensor::numel() const { return product(dims_); }
 
+void Tensor::ReSetHolder(std::shared_ptr<memory::Allocation> holder) {
```
sneaxiy: `Reset` is a complete word.

paddle/fluid/platform/device_context.cc
```diff
@@ -271,8 +310,14 @@ CUDADeviceContext::~CUDADeviceContext() {
 Place CUDADeviceContext::GetPlace() const { return place_; }
 
 void CUDADeviceContext::Wait() const {
+  auto& allocator =
+      DeviceTemporaryAllocator::Instance().Get<CUDADeviceContext>(*this);
+  allocator.MoveToDeleteQueue();
```
sneaxiy: Do we really need 2 methods `MoveToDeleteQueue` and `Release`? It seems that they would not be invoked separately.

chengduoZH: `MoveToDeleteQueue`  is to move the temp_allocations that has not been held by any variable to delete_queue before call `cudaStreamSynchronize`,  `Release` is to delete the  allocation of delete_queue.
This is because, if we call the `Release` after `cudaStreamSynchronize`,  this will lead to some allocations are released before calling `cudaStreamSynchronize`.

qingqing01:
如果是在Wait的时候才释放，而不是计算完就释放，会不会遇到：比如conv需要较大的memory，conv的层数比较多，每次都申请一片，不同conv层用不同memory，导致总体的显存增加？

chengduoZH:
> 如果是在Wait的时候才释放，而不是计算完就释放，会不会遇到：比如conv需要较大的memory，conv的层数比较多，每次都申请一片，不同conv层用不同memory，导致总体的显存增加？

目前是会有这种情况的

paddle/fluid/platform/temporay_allocator.cc
```diff
@@ -0,0 +1,95 @@
...
+    std::unique_lock<std::mutex> lock(mtx_);
+    callback();
+    t_allocations = temp_mem_queue_;
+    temp_mem_queue_.reset(new std::deque<TemporayAllocation *>());
+    wait_delete_mem_ = 0;
+  }
+  for (auto tmp : *t_allocations) {
+    VLOG(10) << "Delete temporary allocation " << tmp->ptr()
+             << " size: " << tmp->size();
+    delete tmp;
```
typhoonzero: curious how this can delete actual memory on device, seems `Allocation` 's detor does nothing

typhoonzero: By the way, the typicall method of allocating caches in linux can use a "Slab" allocator.

chengduoZH: `tmp` is a `TemporaryAllocation` which has a `memory::allocation::AllocationPtr underlying_allocation_;`, and `underlying_allocation_` is a ` std::unique_ptr<Allocation, AllocationDeleter>`, the implementation of `AllocationDeleter` is 
```
void AllocationDeleter::operator()(Allocation* allocation) const {
  auto* allocator = allocation->allocator();
  allocator->Free(allocation);
}
```
When `tmp` is deleted, the `underlying_allocation_` will be released, so the `underlying_allocation_` will be freed.

typhoonzero (approved):
As discussed, maybe we need to put this in context rather than allocator.

qingqing01 (approved):

paddle/fluid/framework/tensor.cc
```diff
@@ -27,6 +28,9 @@ void Tensor::check_memory_size() const {
       "or maybe the required data-type mismatches the data already stored.");
 }
 
+Tensor::Tensor(std::type_index type)
```
panyx0718, line 31: why not support DataType directly?

panyx0718 (approved):

panyx0718:
Haven't look into the detail, the API change is fine to me

paddle/fluid/framework/tensor.cc
chengduoZH: Great, I will fix this in next PR.

paddle/fluid/operators/conv_op.h
```diff
@@ -286,13 +296,18 @@ class GemmConvGradKernel : public framework::OpKernel<T> {
     // to call the matrix multiplication interface.
     Tensor col_matrix;
     if (is_expand) {
-      col.mutable_data<T>(col_shape, context.GetPlace());
+      auto tmp_allocation_ptr =
+          platform::DeviceTemporaryAllocator::Instance().Get(dev_ctx).Allocate(
+              framework::product(col_shape) * sizeof(T));
+      Tensor tep_tensor =
```
panyx0718, line 302: why not just call this `col`?

paddle/fluid/operators/math/concat_and_split.cu
```diff
@@ -144,12 +143,9 @@ class ConcatFunctor<platform::CUDADeviceContext, T> {
       }
       out_col += t_cols;
       inputs_col[i + 1] = out_col;
-      inputs_ptr[i] = const_cast<T*>(input[i].data<T>());
+      inputs_data[i] = const_cast<T*>(input[i].data<T>());
```
panyx0718, line 146: this const_cast can be removed?

paddle/fluid/platform/create_tensor_with_allocationptr.h
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+#pragma once
+#include "paddle/fluid/framework/tensor.h"
+#include "paddle/fluid/platform/temporary_allocator.h"
+namespace paddle {
+namespace platform {
+
+template <typename T>
+paddle::framework::Tensor GetTensor(
```
panyx0718, line 22: this should probably live in tensor.h

paddle/fluid/platform/device_context.h
```diff
@@ -39,6 +41,50 @@ limitations under the License. */
 namespace paddle {
 namespace platform {
 
+/*! \brief device temporary allocator singleton */
```
panyx0718, line 44: document when should temp allocator be used? How is it different from other allocator?

chengduoZH: Done in this PR https://github.com/PaddlePaddle/Paddle/pull/15011.

paddle/fluid/platform/create_tensor_with_allocationptr.h
chengduoZH: Maybe it is better that move `GetTensor ` to `tensor_util.h`.

paddle/fluid/operators/math/concat_and_split.cu
chengduoZH: Done in https://github.com/PaddlePaddle/Paddle/pull/15011.

paddle/fluid/operators/conv_op.h
chengduoZH: Right, done in https://github.com/PaddlePaddle/Paddle/pull/15011.

PR #14878, opened 2018-12-13
MLP forward backward

panyx0718:
test=develop

### DIFF
--- a/paddle/fluid/imperative/tracer.h
+++ b/paddle/fluid/imperative/tracer.h
@@ -43,9 +43,12 @@ void CreateGradOp(const framework::OpDesc& op_desc,
 
 class Tracer {
  public:
-  explicit Tracer(framework::BlockDesc* root_block) : root_block_(root_block) {
+  explicit Tracer(framework::BlockDesc* root_block,
+                  framework::BlockDesc* startup_block)
+      : root_block_(root_block), startup_block_(startup_block) {
     root_scope_ = new framework::Scope();
     scopes_[root_block_] = root_scope_;
+    scopes_[startup_block_] = root_scope_;
   }
 
   virtual ~Tracer() { delete root_scope_; }
@@ -80,6 +83,8 @@ class Tracer {
       } else {
         op->pre_ops_->push_back(nullptr);
       }
+      VLOG(3) << "input vname " << vname << " "
+              << var->Get<framework::LoDTensor>().dims().size();
     }
 
     *op->output_vars_ = outputs;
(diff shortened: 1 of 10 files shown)


python/paddle/fluid/layers/nn.py
```diff
@@ -9426,3 +9427,47 @@ def huber_loss(input, label, delta):
...
+    def __init__(self,
+                 size,
+                 param_attr=None,
+                 num_flatten_dims=1,
+                 dtype=core.VarDesc.VarType.FP32):
+        super(FC, self).__init__()
+        self._size = size
+        self._num_flatten_dims = num_flatten_dims
+        self._dtype = dtype
+        self._helper = LayerHelper('FC', param_attr=param_attr)
```
velconia, line 9442: NO need to export layer helper to user, Better put to PyLayer

paddle/fluid/imperative/tracer.h
```diff
@@ -43,9 +43,12 @@ void CreateGradOp(const framework::OpDesc& op_desc,
 
 class Tracer {
  public:
-  explicit Tracer(framework::BlockDesc* root_block) : root_block_(root_block) {
+  explicit Tracer(framework::BlockDesc* root_block,
+                  framework::BlockDesc* startup_block)
+      : root_block_(root_block), startup_block_(startup_block) {
     root_scope_ = new framework::Scope();
     scopes_[root_block_] = root_scope_;
+    scopes_[startup_block_] = root_scope_;
```
velconia, line 51: Different opinion here, two different blocks ```root_block``` and ```startup_block``` will involve complexity and NO benefits, Parameter initiliazation is better to put into ```_build_once```

python/paddle/fluid/layers/nn.py
```diff
@@ -9426,3 +9427,47 @@ def huber_loss(input, label, delta):
                  'Residual': residual},
         attrs={'delta': delta})
     return out
+
+
+class FC(layers.PyLayer):
```
velconia, line 9432: Better to put this layer into ```imperative.nn``` or something

velconia (approved):
LGTM with a little comments

python/paddle/fluid/layers/nn.py
panyx0718: Currently, we are not exposing it via __all__. but eventually we want to make PyLayer support both static and dynamic

velconia (commented):
LGTM, polish the code later

PR #14886, opened 2018-12-13
add trt int8 for analysis predictor 

NHZlX:
add trt int8 for analysis predictor


NHZlX:
https://github.com/PaddlePaddle/Paddle/pull/15242#event-2099687522

PR #14888, opened 2018-12-13
Enable MKL-DNN INT8 Inference

xiaolil1:
We enabled MKLDNN INT8 inference.


luotao1 (commented):
please squash your commit log, which is too large

luotao1 (commented):
please update as less file as well. It is no need to updata `paddle/fluid/framework/ir/CMakeLists.txt`, `paddle/fluid/framework/ir/conv_elementwise_add_mkldnn_fuse_pass.cc`, `paddle/fluid/framework/ir/conv_elementwise_add_mkldnn_fuse_pass_tester.cc`, `python/setup.py.in` in this PR.

PR #14889, opened 2018-12-14
adam optimizer support lazy mode

jacquesqiao:
in my test case, in distribute training, when the parameter size is 800M, the adam optimizer will be 10 more time faster.


typhoonzero (dismissed):
LGTM

paddle/fluid/operators/optimizers/adam_op.h
```diff
@@ -351,11 +357,27 @@ class AdamOpKernel : public framework::OpKernel<T> {
...
-      for_range(functor);
+          grad_merge.rows().size(), lazy_mode);
+      VLOG(3) << "lazy_mode :" << lazy_mode;
+      if (lazy_mode) {
+        std::vector<int64_t> id_vector;
+        size_t row_count = grad_merge.rows().size();
+        std::vector<int64_t> cpu_rows(grad_merge.rows());
+        for (size_t row_index = 0; row_index < row_count; ++row_index) {
+          for (size_t offset = 0; offset < row_numel; ++offset) {
+            size_t i = cpu_rows[row_index] * row_numel + offset;
```
chengduoZH, line 371: What do you want to with `cpu_rows`? It is defined but not initialized.

paddle/fluid/operators/optimizers/adam_op.h
```diff
@@ -351,11 +357,27 @@ class AdamOpKernel : public framework::OpKernel<T> {
...
+            id_vector.push_back(i);
+          }
+        }
+        platform::ForRangeIn<DeviceContext> for_range_in(
+            static_cast<const DeviceContext&>(ctx.device_context()), id_vector);
```
chengduoZH: Note that the `for_range_in ` is in stack memory, once the function return, the memory will be released.
You define `range_` [here](https://github.com/PaddlePaddle/Paddle/pull/14889/files#diff-f3be7a1b3192291a4c06a311d5e643abR135), and it is used by CUDA Kernel, if the CUDA Kernel doesn't start to execute before the current function returns, the `range_` will be released, it is dangerous.

paddle/fluid/operators/optimizers/adam_op.h
```diff
@@ -351,11 +357,27 @@ class AdamOpKernel : public framework::OpKernel<T> {
...
+        for_range_in(functor);
```
chengduoZH: You can use `platform::ForRange` do above task.

paddle/fluid/operators/optimizers/adam_op.h
jacquesqiao: cpu_rows is inited https://github.com/PaddlePaddle/Paddle/pull/14889/files#diff-e616ec13ecb3784788f198a7812b4a0aR365
```cpp
std::vector<int64_t> cpu_rows(grad_merge.rows());
```

paddle/fluid/operators/optimizers/adam_op.h
jacquesqiao: No, ForRange can not to this task.

paddle/fluid/operators/optimizers/adam_op.h
jacquesqiao: This is a problem

python/paddle/fluid/optimizer.py
```diff
@@ -641,9 +641,14 @@ class AdamOptimizer(Optimizer):
...
-        regularization: A Regularizer, such as
-                        fluid.regularizer.L2DecayRegularizer.
+        regularization: A Regularizer, such as fluid.regularizer.L2DecayRegularizer.
         name: A optional name prefix.
+        lazy_mode(bool: false): The official Adam algorithm has two moving-average accumulators
+        the accumulators are updated at every step. Every element of the two moving-average is updated
+        in both dense mode and sparse mode. If the size of parameter is very large, then the update
+        may be very slow. The lazy mode only update the element that has gradient is the current
+        mini-batch, so it will be much more faster. But this mode has different semantics with the
+        original Adam algorithm and may lead to different result.
```
chengduoZH, line 651: Do you have the performance report in lazy_mode? If so, you can post the data in this PR.

jacquesqiao: Done

chengduoZH (approved):

python/paddle/fluid/optimizer.py
```diff
@@ -641,9 +641,14 @@ class AdamOptimizer(Optimizer):
...
         beta2 (float): The exponential decay rate for the 2nd moment estimates.
         epsilon (float): a small float value for numerical stability.
-        regularization: A Regularizer, such as
-                        fluid.regularizer.L2DecayRegularizer.
+        regularization: A Regularizer, such as fluid.regularizer.L2DecayRegularizer.
         name: A optional name prefix.
+        lazy_mode(bool: false): The official Adam algorithm has two moving-average accumulators
+        the accumulators are updated at every step. Every element of the two moving-average is updated
+        in both dense mode and sparse mode. If the size of parameter is very large, then the update
+        may be very slow. The lazy mode only update the element that has gradient is the current
```
qingqing01, line 649: The lazy mode only update -> The lazy mode only updates
is the current -> in current

panyx0718 (approved):
LG API

typhoonzero (approved):

PR #14904, opened 2018-12-16
refactor RunImpl

panyx0718:
There are a few reasons to make this change

1. RuntimeContext should be prepared before Run and scope lookup should be avoided in InferXXX and Compute
2. Imperative Mode should call Kernel directly and it should pass in Variable and get Variable without Scope.

test=develop

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -477,23 +477,22 @@ bool OpSupportGPU(const std::string& op_type) {
 
 class RuntimeInferShapeContext : public InferShapeContext {
  public:
-  RuntimeInferShapeContext(const OperatorBase& op, const Scope& scope)
-      : op_(op), scope_(scope) {}
+  RuntimeInferShapeContext(const OperatorBase& op, const Scope& scope,
+                           const RuntimeContext& ctx)
+      : op_(op), scope_(scope), ctx_(ctx) {}
 
   bool HasInput(const std::string& name) const override {
     // has only one input
-    const auto& ins = op_.Inputs();
+    const auto& ins = ctx_.inputs;
     auto it = ins.find(name);
     if (it == ins.end()) {
       return false;
     }
     const auto& in = it->second;
-    if (in.size() == 0 || in[0] == kEmptyVarName) {
-      return false;
-    }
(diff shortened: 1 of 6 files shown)


paddle/fluid/framework/operator.cc
```diff
@@ -412,11 +431,52 @@ bool ExecutionContext::HasOutput(const std::string& name) const {
...
+  auto it = ctx_.inputs.find(name);
+  if (it == ctx_.inputs.end()) return nullptr;
+
+  PADDLE_ENFORCE_LE(it->second.size(), 1UL,
+                    "Operator %s's input %s should contain only one variable.",
+                    op_.Type(), name);
+  return it->second.empty() ? nullptr : it->second[0];
+}
+
+Variable* ExecutionContext::OutputVar(const std::string& name) const {
```
hshen14: Shall we find the variable by name as well?

paddle/fluid/framework/operator.cc
```diff
@@ -797,13 +854,19 @@ void OperatorWithKernel::TransferInplaceVarsBack(
...
-Scope* OperatorWithKernel::TryTransferData(
+Scope* OperatorWithKernel::PrepareData(
     const Scope& scope, const OpKernelType& expected_kernel_key,
-    std::vector<std::string>* transfered_inplace_vars) const {
+    std::vector<std::string>* transfered_inplace_vars,
+    RuntimeContext* ctx) const {
   Scope* new_scope = nullptr;
   for (auto& var_name_item : Inputs()) {
-    for (auto& var_name : var_name_item.second) {
+    std::vector<Variable*>& input_vars = ctx->inputs[var_name_item.first];
```
typhoonzero, line 863: Maybe no need to construct the `input_vars` in every op run since all operator instance is long-lived, can do something like:
```c++
if (UNLIKELY(!prepared_)) {
  for ....
    do prepare
}
do datatransfers.
```

typhoonzero (approved):
Exellent change, one enhancement comment.

paddle/fluid/framework/operator.cc
panyx0718: Right, will consider this optimization in the follow up

PR #14914, opened 2018-12-17
Fix yolo loss

heavengate:
1. use stable Sigmoid Cross Entropy implement
2. use L1 Loss for w, h coordinates
3. remove loss weight attr
4. enhance loss performance


heavengate:
need fix

PR #14464, opened 2018-11-16
Enable MKL-DNN INT8 Inference

xiaolil1:
We enabled MKLDNN INT8 inference.
We got these inference result:

1. Accuracy

| Topology | FP32 Accuracy(Top-1/Top-5) | INT8 Accurary(Top-1/Top-5) |
| --- | :---: | :---: |
ResNet-50 (FB) | 76.63%/93.10%| 76.42%/93.07% 
MobileNet-V1 | 70.78%/89.69% | 70.10%/89.30%

2. Performance

| Topology | FP32 Throughput  | INT8 Throughput | FP32 latency(1x1)  | INT8 latency(1x1) |
| --- | :---: | :---: | :---: | :---: |
ResNet-50 (FB) | 260 | 527 | 62.4 |36.7 
MobileNet-V1 | 1207 |1934 | 9.3 | 8.4

Note: The above performance measured on SKX8180 1S (HT On, Turbo On) 
(Throughput imgs/sec; Latency: ms; 1x1: batch size 1 x thread 1)

PR #14929, opened 2018-12-17
Enable MKL-DNN INT8 inference

xiaolil1:
We enabled mkldnn INT8 inference with performance:

Paddle version | Topology | Performance measured on SKX8180 1S (HT On, Turbo On) (Throughput imgs/sec; Latency: ms;   1x1: batch size 1 x thread 1 |  
-- | -- | -- | --
 
FP32 Throughput | INT8 Throughput | FP32 Latency (1x1) | INT8 Latency (1x1) | FP32 Latency (1x4) | INT8 Latency (1x4) | FP32 Latency (1x28) | INT8 Latency (1x28) |  
Our master: 77ca0bce10f1c85de0cc04330958e2bf9db72fee | ResNet-50 (FB) | 338 | 537 | 54.0 | 32.5 | 16.3 | 10.0 | 7.2 | 4.1 |  
MobileNet-V1 | 1206 | 1958 | 8.8 | 6.5 | 3.1 | 2.6 | 2.1 | 1.8 |  
Baidu develop: b82a44ea8500edcdddbd3459ce89f9a44c1e2b41 | ResNet-50 (FB) | 335 |   | 54.7 |   | 17.8 |   | 9.0 |   |  
MobileNet-V1 | 1193 |   | 9.6 |   | 3.7 |   | 2.9 |   |  
  |   |   |   |   |   |   |   |   |   |  
Paddle version | Topology | Performance measured on SKX6148 1S (HT On, Turbo On) (Throughput imgs/sec; Latency: ms;   1x1: batch size 1 x thread 1 |  
 
FP32 Throughput | INT8 Throughput | FP32 Latency (1x1) | INT8 Latency (1x1) | FP32 Latency (1x4) | INT8 Latency (1x4) | FP32 Latency (1x20) | INT8 Latency (1x20) |  
Our master: 77ca0bce10f1c85de0cc04330958e2bf9db72fee | ResNet-50 (FB) | 240 | 384 | 54.3 | 32.8 | 17.1 | 10.5 | 8.0 | 5.4 |  
MobileNet-V1 | 1032 | 1598 | 9.0 | 6.7 | 3.3 | 2.7 | 2.3 | 1.9 |  
Baidu develop: b82a44ea8500edcdddbd3459ce89f9a44c1e2b41 | ResNet-50 (FB) | 239 |   | 55.7 |   | 18.6 |   | 10.2 |   |  
MobileNet-V1 | 1015 |   | 9.8 |   | 4.0 |   | 3.2 |   |

### DIFF
--- a/paddle/fluid/operators/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/conv_mkldnn_op.cc
@@ -12,6 +12,7 @@
    See the License for the specific language governing permissions and
    limitations under the License. */
 
+#include <unordered_map>
 #include "paddle/fluid/framework/data_layout_transform.h"
 #include "paddle/fluid/memory/malloc.h"
 #include "paddle/fluid/operators/conv_op.h"
@@ -68,13 +69,42 @@ inline mkldnn::memory::format GetWeightsFormat(mkldnn::memory::format format,
   }
 }
 
-template <typename T>
+template <typename T, typename K>
 class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
  public:
+  struct ConvInfo {
+    const paddle::framework::Tensor* input;
+    const paddle::framework::Tensor* bias;
+    const paddle::framework::Tensor* output;
+    const paddle::framework::Tensor* weight;
+    std::vector<int>* strides;
+    std::vector<int>* paddings;
+    std::vector<int>* dilations;
+    std::vector<int>* src_tz;
+    std::vector<int>* weights_tz;
+    std::vector<int>* dst_tz;
+    int g;
+  };
+  struct MkldnnInfo {
+    bool fuse_relu;
+    bool fuse_residual_conn;
+    bool force_fp32_output;
+    bool is_test;
+    bool is_conv3d;
+    const mkldnn::engine* mkldnn_engine;
+    std::vector<primitive>* pipeline;
+    const std::string* key_conv_pd;
+    std::string* key;
+    std::shared_ptr<platform::ConvMKLDNNHandler> handler;
+    std::shared_ptr<mkldnn::memory> src_memory_p;
+    std::shared_ptr<mkldnn::memory> user_src_memory_p;
+    std::shared_ptr<mkldnn::memory> dst_memory_p;
+    std::shared_ptr<mkldnn::convolution_forward> conv_p;
+    std::shared_ptr<mkldnn::convolution_forward::primitive_desc> conv_pd;
+  };
   void Compute(const paddle::framework::ExecutionContext& ctx) const override {
     PADDLE_ENFORCE(paddle::platform::is_cpu_place(ctx.GetPlace()),
                    "It must use CPUPlace.");
-
     const bool is_test = ctx.Attr<bool>("is_test");
 
     auto& dev_ctx =
@@ -107,9 +137,16 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     std::vector<int> strides = ctx.Attr<std::vector<int>>("strides");
     std::vector<int> paddings = ctx.Attr<std::vector<int>>("paddings");
     std::vector<int> dilations = ctx.Attr<std::vector<int>>("dilations");
+    int groups = ctx.Attr<int>("groups");
+
     bool fuse_relu = ctx.Attr<bool>("fuse_relu");
     bool fuse_residual_conn = ctx.Attr<bool>("fuse_residual_connection");
-    int groups = ctx.Attr<int>("groups");
+
+    bool force_fp32_output = ctx.Attr<bool>("force_fp32_output");
+    if (fuse_residual_conn) {
+      PADDLE_ENFORCE(force_fp32_output != true,
+                     "residual fusion does not support force output with fp32");
+    }
 
     bool is_conv3d = strides.size() == 3U;
     // TODO(tpatejko): add support for dilation
@@ -121,7 +158,6 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
         "dilation in convolution is not implemented yet");
 
     const T* input_data = input->data<T>();
-    const T* filter_data = filter->data<T>();
 
     std::vector<int> src_tz = paddle::framework::vectorize2int(input->dims());
     std::vector<int> weights_tz =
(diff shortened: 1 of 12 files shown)


hshen14:
@tpatejko @kbinias Please review the PR, thanks.

luotao1:
- First of all, @xiaolil1 why do you delete all the files in #14464, we could not see the corresponding codes on each comment.
image

- This PR is too large, to accelerate the PR merge, how about separate it into several PRs: (it seems first three small PRs could be merged ASAP)
   - `dequantize_mkldnn_op.cc, dequantize_op.cc, test_dequantize_mkldnn_op.py`
   - `quantize_mkldnn_op.cc, quantize_op.cc, test_quantize_mkldnn_op.py`
   - `pool_mkldnn_op.cc`, and **lack of its int8 test**.
   - `conv_mkldnn_op.cc`, and **lack of its int8 test**.

- For `conv_mkldnn_op.cc`, I agree the https://github.com/PaddlePaddle/Paddle/pull/14464#discussion_r235304592 and https://github.com/PaddlePaddle/Paddle/pull/14464#discussion_r236994935 from Poland team. There are so many if-else statements. It's very difficult to follow and read. How about pass the context instead and fetch necessary parameters from it?
```
void Compute(const paddle::framework::ExecutionContext& ctx) const override {
   xxxx;
   if (is_int8) {
       Compute_int8(ctx);
       return;
   }
   same codes for fp32;
}
```

paddle/fluid/operators/dequantize_op.cc
```diff
@@ -0,0 +1,45 @@
...
+
+  return framework::OpKernelType(ctx.Input<Tensor>("Input")->type(),
+                                 ctx.GetPlace(), layout_, library_);
+}
+
+void DeQuantOpMaker::Make() {
+  AddInput("Input", "input data");
+  AddOutput("Output", "output data");
+  AddAttr<float>("Scale", "scale data").SetDefault({1.0f});
+  AddComment(R"DOC(This op will quantize data from INT8 to FP32)DOC");
```
wojtuss, line 36: "quantize" -> "dequantize" ?

wojtuss (commented):
I agree with @luotao1's comments (https://github.com/PaddlePaddle/Paddle/pull/14929#issuecomment-448457447).

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -412,9 +924,12 @@ class ConvMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
...
     // This name will be used as key when saving info into device context
-    const std::string key = platform::ConvMKLDNNHandler::GetHash(
-        src_tz, weights_tz, strides, paddings, dilations, groups,
-        ctx.op().Input("Output"));
+    const int MaxKeyLength = 256;
+    std::string key;
+    key.reserve(MaxKeyLength);
+    platform::ConvMKLDNNHandler::AppendKey(key, src_tz, weights_tz, strides,
+                                           paddings, dilations, groups,
+                                           ctx.op().Input("Output"));
```
tpatejko, line 932: This piece of code is repeated in few places. Could you make a separate entity (function or class, maybe functor?), and reuse the code?

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -131,21 +167,198 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    if (conv_p == nullptr || !is_test) {
+      struct ConvInfo convinfo;
+      struct MkldnnInfo mkldnninfo;
+      convinfo.strides = &strides;
+      convinfo.paddings = &paddings;
+      convinfo.dilations = &dilations;
+      convinfo.src_tz = &src_tz;
+      convinfo.weights_tz = &weights_tz;
+      convinfo.dst_tz = &dst_tz;
+      convinfo.g = g;
```
tpatejko, line 217: I guess you could use aggregate initialization for `ConvInfo` and `MkldnnInfo` structures.

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -131,21 +167,198 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+      mkldnninfo.mkldnn_engine = &mkldnn_engine;
+      mkldnninfo.handler = handler;
+      mkldnninfo.pipeline = &pipeline;
+      mkldnninfo.key_conv_pd = &key_conv_pd;
+      mkldnninfo.key = &key;
+      mkldnninfo.src_memory_p = src_memory_p;
+      mkldnninfo.user_src_memory_p = user_src_memory_p;
+      mkldnninfo.dst_memory_p = dst_memory_p;
+      mkldnninfo.conv_p = conv_p;
+      mkldnninfo.conv_pd = conv_pd;
```
tpatejko, line 232: Check comment above.

paddle/fluid/operators/conv_op.cc
```diff
@@ -303,6 +320,9 @@ void Conv3DOpMaker::Make() {
       "Defaults to \"NHWC\". Specify the data format of the output data, "
       "the input will be transformed automatically. ")
       .SetDefault("AnyLayout");
+  AddAttr<bool>("force_fp32_output",
+                "(bool, default false) Only used in mkldnn INT8 kernel")
+      .SetDefault(false);
```
tpatejko, line 325: I think there are two attributes `force_fp32_output`. How is this attribute used? How is it set and who/what can set it?

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -131,21 +167,198 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+          if (fuse_relu) {
+            uint8_t* output_data = output->mutable_data<uint8_t>(
+                ctx.GetPlace(), ::paddle::memory::Allocator::kDefault,
+                handler->GetDstMemorySize());
+            dst_memory_p->set_data_handle(to_void_cast<uint8_t>(output_data));
+          } else {
+            int8_t* output_data = output->mutable_data<int8_t>(
+                ctx.GetPlace(), ::paddle::memory::Allocator::kDefault,
+                handler->GetDstMemorySize());
+            dst_memory_p->set_data_handle(to_void_cast<int8_t>(output_data));
```
tpatejko, line 317: In few places above you have a repeating patterns that allocates output for some type (float/int8/uint8) and sets a data handle for `dst_memory_p`. The only difference usually is a type, float/int8 or uint8.

Maybe it would be ok to move it to some template function that is parametrized with type of output data? That would make code above much more readable.

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -131,21 +167,198 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+
     const std::string key_conv_pd = key + "@conv_pd";
 
+    bool is_INT8 = false;
+    mkldnn::memory::data_type src_dt =
+        paddle::framework::ToMKLDNNDataType(input->type());
+    if (src_dt == mkldnn::memory::data_type::u8 ||
+        src_dt == mkldnn::memory::data_type::s8) {
+      is_INT8 = true;
+    }
```
tpatejko, line 185: If `input->type()` is consistent with template type `T`, the code between lines 179 and 185 could be reduced to, I think:
```
bool is_INT8 = std::is_same<T, int8_t>::value || std::is_same<T, uint8_t>::value;
```
This would make control paths between int8 and floating-points only depend on template parameters, which could lead to more refactoring.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -181,6 +194,42 @@ class MKLDNNHandler {
...
+                           const std::vector<int>& dims) {
+    for (unsigned int i = 0; i < dims.size(); i++) {
+      AppendKey(key, std::to_string(dims[i]));
+    }
+  }
+
+  static void AppendKey(std::string& key,  // NOLINT
+                        const std::string& s) {
+    key.append(s);
+  }
```
tpatejko, line 231: I think you can make AppendKey, AppendKeyDims, AppendKeyVec and AppendKey const and stateless. 

You could turn std::string& key into const std::string& prefix, and the rest of the parameters could be const &, and the methods could return std::string. Prefix could initialize local variable key. So:
```
static std::string AppendKey(const std::string& prefix, const &....) {
  std::string key = prefix;
  key += AppendKeyDims(input_dims);  //or key.append(AppendKeyDims(input_dims));
  key += AppendKeyDims(weights_dims);
  ...

  return key;
}
```

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -68,13 +69,42 @@ inline mkldnn::memory::format GetWeightsFormat(mkldnn::memory::format format,
...
 }
 
-template <typename T>
+template <typename T, typename K>
 class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
  public:
+  struct ConvInfo {
+    const paddle::framework::Tensor* input;
+    const paddle::framework::Tensor* bias;
+    const paddle::framework::Tensor* output;
```
Sand3r-, line 78: Tensor* output doesn't need to be const

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -68,13 +69,42 @@ inline mkldnn::memory::format GetWeightsFormat(mkldnn::memory::format format,
...
+    const paddle::framework::Tensor* weight;
+    std::vector<int>* strides;
+    std::vector<int>* paddings;
+    std::vector<int>* dilations;
+    std::vector<int>* src_tz;
+    std::vector<int>* weights_tz;
+    std::vector<int>* dst_tz;
+    int g;
```
Sand3r-, line 86: Consider renaming `g` to `groups`

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -107,9 +137,16 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    int groups = ctx.Attr<int>("groups");
+
     bool fuse_relu = ctx.Attr<bool>("fuse_relu");
     bool fuse_residual_conn = ctx.Attr<bool>("fuse_residual_connection");
-    int groups = ctx.Attr<int>("groups");
+
+    bool force_fp32_output = ctx.Attr<bool>("force_fp32_output");
+    if (fuse_residual_conn) {
+      PADDLE_ENFORCE(force_fp32_output != true,
+                     "residual fusion does not support force output with fp32");
```
Sand3r-, line 148: Isn't this check specific to int8? Perhaps it could be moved.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -181,6 +194,42 @@ class MKLDNNHandler {
     return dims2str(operand_dims) + suffix;
   }
 
+  static void AppendKey(std::string& key,                    // NOLINT
```
Sand3r-, line 197: Perhaps AppendKey could be renamed to CreateKey, and it would return an std::string, instead of filling a pre-determined one (it'd offload the need to create the string buffer and preallocating the size).

paddle/fluid/operators/conv_mkldnn_op.cc
Sand3r-: If they aren't consistent, these lines can be reduced to:
```C++
mkldnn::memory::data_type src_dt = paddle::framework::ToMKLDNNDataType(input->type());
bool is_INT8 = src_dt == mkldnn::memory::data_type::u8 || src_dt == mkldnn::memory::data_type::s8;
```

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -131,21 +167,198 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+        paddle::framework::ToMKLDNNDataType(input->type());
+    if (src_dt == mkldnn::memory::data_type::u8 ||
+        src_dt == mkldnn::memory::data_type::s8) {
+      is_INT8 = true;
+    }
+
+    bool need_s8_to_u8 = false;
+    if (fuse_residual_conn && is_INT8 && fuse_relu) {
+      need_s8_to_u8 = true;
+    }
```
Sand3r-, line 190: This can be reduced to:
```C++
bool need_s8_to_u8 = fuse_residual_conn && is_INT8 && fuse_relu;
```

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+  void CreateINT8Primitive(const paddle::framework::ExecutionContext& ctx,
+                           const paddle::platform::MKLDNNDeviceContext* dev_ctx,
+                           const paddle::framework::Tensor* input,
+                           const paddle::framework::Tensor* filter,
+                           const paddle::framework::Tensor* bias,
+                           paddle::framework::Tensor* output,
+                           ConvInfo* convinfo, MkldnnInfo* mkldnninfo) const {
+    const T* input_data = input->data<T>();
+    const K* filter_data = filter->data<K>();
+    bool is_INT8 = true;
```
Sand3r-, line 505: I think that the definition of this variable could be changed to `constexpr bool`, and also it could be moved to where it is actually used (619, 692).

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    auto scale_in_data = ctx.Attr<float>("Scale_in");
+    auto scale_in_eltwise_data = ctx.Attr<float>("Scale_in_eltwise");
+    auto scale_weights_data = ctx.Attr<std::vector<float>>("Scale_weights");
+    auto scale_out_data =
+        mkldnninfo->force_fp32_output ? 1.0f : ctx.Attr<float>("Scale_out");
+
+    bool is_multi_channel = scale_weights_data.size() > 1 ? true : false;
```
Sand3r-, line 512: Can be reduced to 
```C++
const bool is_multi_channel = scale_weights_data.size() > 1;
```

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+
+    std::vector<float> output_shift_scale;
+    float sum_scale = 1.0f;
+    int count =
+        is_multi_channel
+            ? (convinfo->g > 1
+                   ? (*convinfo->weights_tz)[1] * (*convinfo->weights_tz)[0]
+                   : (*convinfo->weights_tz)[0])
+            : 1;
```
Sand3r-, line 521: I find this part a bit unreadable. Fortunately since this code also appears in line 686, it can be extracted to the function and described with bool expressions and if conditions instead of ternary operators.

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    output_shift_scale.resize(count);
```
Sand3r-, line 522: The definition of `output_shift_scale` can be moved here to avoid unnecessary resize.
```C++
std::vector<float> output_shift_scale(count);
```

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+      if (scale_weights_data[i] == 0.0)
+        output_shift_scale[i] = scale_out_data;  // weights data will contain 0
+                                                 // in some models, then weights
+                                                 // scale couldn't be calculated
+      else
+        output_shift_scale[i] =
+            scale_out_data / (scale_in_data * scale_weights_data[i]);
+    }
+    if (mkldnninfo->fuse_residual_conn) {
+      sum_scale = scale_out_data / scale_in_eltwise_data;
```
Sand3r-, line 534: To avoid unnecessary ifs, this can be dragged into the following `if (mkldnninfo->fuse_residual_conn)` at line 571.

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+        platform::data_format_to_memory_format(data_format);
+
+    auto bias_tz = paddle::framework::vectorize2int(bias->dims());
+
+    auto src_md = platform::MKLDNNMemDesc(
+        *convinfo->src_tz, memory::data_type::u8, chosen_memory_format);
+    auto weights_md = platform::MKLDNNMemDesc(
+        *convinfo->weights_tz, memory::data_type::s8, chosen_memory_format);
+
+    auto dst_dt = mkldnninfo->fuse_relu
```
Sand3r-, line 560: The choice of `dst_dt` type could be turned into a `switch` statement which would consider `fuse_relu`, `force_fp32_output` and `fuse_residual_conn` which would make it a bit more readable.

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
-    output->set_layout(DataLayout::kMKLDNN);
-    output->set_format(GetMKLDNNFormat(*dst_memory_p));
+    // push primitive to stream and wait until it's executed
+    mkldnninfo->pipeline->push_back(*mkldnninfo->conv_p);
   }
 
- private:
-  mkldnn::primitive_attr CreatePostOps(bool fuse_relu,
-                                       bool fuse_residual_conn) const {
+  mkldnn::primitive_attr CreatePostOps(
```
Sand3r-, line 705: This function should reduced to:
```C++
auto conv_attr = CreatePostOps(fuse_relu, fuse_residual_conn);
int mask = output_shift_scale.size() > 1 ? 1 << 1 : 0;
conv_attr.set_output_scales(mask, output_shift_scale);
return conv_attr;
```
To reduce code duplication.

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -298,6 +753,34 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     return conv_attr;
   }
 
+  std::unique_ptr<mkldnn::convolution_forward::primitive_desc>
+  ConvFwdPrimitiveDesc(const memory::desc& src, const memory::desc& weights,
```
Sand3r-, line 757: This and the following ConvFwdPrimitiveDesc function (line 808) look very similar to it's fp32 counterparts. Couldn't we benefit here from the higher code re-use?

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
-                                          bias_memory_p, dst_memory_p);
+    // create reorder primitive if the input format is not the preferred one
+    mkldnninfo->src_memory_p =
+        mkldnninfo->handler->AcquireSrcMemoryFromPrimitive(
+            mkldnninfo->user_src_memory_p, *mkldnninfo->pipeline);
+
+    std::shared_ptr<mkldnn::memory> weights_memory_p;
+    int mask_reorder = is_multi_channel
+                           ? ((convinfo->g != 1) ? (1 << 1) + (1 << 0) : 1 << 0)
+                           : 0;
```
Sand3r-, line 616: Could we replace the bit shifts with the actual values? `3 : 1`

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+      PADDLE_ENFORCE_EQ(output->dims(), residual_param->dims(),
+                        "Output and elementwise parameter need to have the "
+                        "same dimension sizes");
+      auto residual_dt =
+          paddle::framework::ToMKLDNNDataType(residual_param->type());
+      PADDLE_ENFORCE_EQ(
+          residual_param->format(), mkldnninfo->handler->GetDstFormat(),
+          "Conv input dimension and filter dimension should be the same.");
+      output->ShareDataWith(*residual_param);
+      if (residual_dt == mkldnn::memory::data_type::u8) {
```
Sand3r-, line 632: This
```C++
output_data = (...)
dst_memory_p = (..., output_data)
```
Seem to be a recurring pattern until line 665. Perhaps we could create a template function that would perform this action supplied with different variable types (`uint8_t, int8_t, float`)?

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    std::vector<float> scale_bias_data;
+    auto scale_bias_key = *mkldnninfo->key + "@scale_bias";
+    if (bias) {
+      const float* bias_data = bias->data<float>();
+      auto user_bias_md = platform::MKLDNNMemDesc(
+          {bias_tz}, platform::MKLDNNGetDataType<float>(), memory::format::x);
+      auto user_bias_memory_p = mkldnninfo->handler->AcquireBiasMemory(
+          user_bias_md, to_void_cast<float>(bias_data));
+      std::shared_ptr<mkldnn::memory> bias_memory_p;
+      int mask_reorder = is_multi_channel ? 1 << 0 : 1;
```
Sand3r-, line 678: Isn't this the same as writing:
```C++
int mask_reorder = 1;
```
? `1 << 0 <=> 1`, so no matter if `is_multi_channel` is true or false, the returned value will be equal to `1`.

paddle/fluid/operators/conv_mkldnn_op.cc
```diff
@@ -216,72 +438,305 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+      int count =
+          is_multi_channel
+              ? (convinfo->g > 1
+                     ? (*convinfo->weights_tz)[1] * (*convinfo->weights_tz)[0]
+                     : (*convinfo->weights_tz)[0])
+              : 1;
+      scale_bias_data.resize(count);
```
Sand3r-, line 685: `scale_bias_data` Could be just created here in a form of:
```C++
std::vector<float> scale_bias_data(count);
```

Instead of being created before the `if (bias)`, especially since nothing apart from this branch uses it.

Sand3r- (commented):
It's great to see that you've refactored some of the code since the last PR. Good job on that. Please find my enclosed suggestions in the detailed review below. It mostly touches on increased readability and higher code reuse.

paddle/fluid/operators/conv_op.cc
xiaolil1: Calibration tool will set the value. By default it is false, which means INT8 convolution output is INT8. However, it is allowed to set output FP32.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: Right, it is an option to add such template function. However, it probably makes confusing since FP32 uses the template like you said. So, for INT8, we decide to specify the clear type to describe output memory type of residual block or force_fp32.

paddle/fluid/platform/mkldnn_reuse.h
xiaolil1: The new implementation is the optimization of key generation for reuse. The key is pre-allocated with buffer size 256 and then appends the dim string, etc (original implementation has a couple of malloc overhead). We verified the key reuse works well with the optimized version and the performance for both FP32 and INT8 was improved at latency. So, we probably cannot follow your comments.

paddle/fluid/operators/conv_op.cc
luotao1: > Calibration tool will set the value

If it only set in runtime, it could not be added here. You can use following codes to insert the attr at runtime.
https://github.com/PaddlePaddle/Paddle/blob/55af1168614edd1d9a0b9077304141f4e458c92b/paddle/fluid/framework/ir/is_test_pass.cc#L45-L46

paddle/fluid/operators/conv_mkldnn_op.cc
luotao1: > Maybe it would be ok to move it to some template function that is parametrized with type of output data? That would make code above much more readable.

I agree with @tpatejko. 

>  it probably makes confusing since FP32 uses the template like you said.

@xiaolil1 Why using template will make confuse?

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: yes, we will remove it.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: I will optimize it with template.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: yes, will optimize it.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: yes, will optimize it.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: I will remove it due to it will always be true for CreateINT8Primitive.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: yes , will optimize it.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: sure, will optimize it.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: It's according to MKL-DNN implementation for data format.

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: It's according to MKL-DNN implementation for data format.

paddle/fluid/operators/conv_op.cc
xiaolil1: Currently , our calibration tool is implemented on python, which will save such flags to model offline, we will do this runtime after we enable C PASS calibration.

paddle/fluid/operators/dequantize_op.cc
xiaolil1: yes, thanks for reminder

paddle/fluid/operators/conv_mkldnn_op.cc
xiaolil1: I will optimize it with template as your suggestion.

kbinias (approved):
The results are reproducible.
We accept these PR, but Intel team is going to start refactoring the source code in next pull requests.

PR #14944, opened 2018-12-18
[MKL-DNN] Fully Connected

Sand3r-:
Introduces Fully Connected operator. It has a comparable performance to the reference version, although it is necessary to be merged, so that the base for the int8 implementation and its performance gains is established.


Sand3r-:
This PR will be re-opened after the 3rd of January (holiday absence).

Sand3r-:
Because MKL-DNN doesn't support Tensors of dimension sizes different that 2 and 4 in FC operator, I've decided to make a fallback to native version in this sort of scenarios (currently it affects small_dam model only).

PR #14480, opened 2018-11-19
fea/infer memory optim

Superjomn:
Memory optimization pass for inference with pre-analysis of memory usage without GC.

It works with the following steps:

- Run the inference with several batches of fake data, important to try different sequence lengths if sequence data is used as inputs.
   - the predictor will collect the shapes of all the Tensors and persist to a cache file.
- Run the inference formally, the memory optimization pass will load the cache file, split the tensors into several clusters by the shape (currently just the batch size) change behaviors.
   - try to make a plan for reusing the memory of the existing tensors, attention that, a tensor can only reuse the ones in the same cluster.
      - the fixed shape tensor can reuse all the others with the fixed shape,
      - the tensor with LoD enabled can only reuse the others with the same sequence length,
      - the output of some `pool` op will change the first dimension to one(for each sequence), it can only share the memory with the other ones with the same behavior (TODO analysis all the shape dimensions).

It should work with both the LoD or Non-LoD models.

Currently, I've tested it with DAM, and that saves about 77.8% memory.

PR #14953, opened 2018-12-19
Fea/infer memory optim2

Superjomn:
Clone from https://github.com/PaddlePaddle/Paddle/pull/14480

Memory optimization pass for inference with pre-analysis of memory usage without GC.

It works with the following steps:

- Run the inference with several batches of fake data, important to try different sequence lengths if sequence data is used as inputs.
   - the predictor will collect the shapes of all the Tensors and persist to a cache file.
- Run the inference formally, the memory optimization pass will load the cache file, split the tensors into several clusters by the shape (currently just the batch size) change behaviors.
   - try to make a plan for reusing the memory of the existing tensors, attention that, a tensor can only reuse the ones in the same cluster.
      - the fixed shape tensor can reuse all the others with the fixed shape,
      - the tensor with LoD enabled can only reuse the others with the same sequence length,
      - the output of some `pool` op will change the first dimension to one(for each sequence), it can only share the memory with the other ones with the same behavior (TODO analysis all the shape dimensions).

It should work with both the LoD or Non-LoD models.

Currently, I've tested it with DAM, and that saves about 70% memory.

image


paddle/fluid/framework/ir/node.h
```diff
@@ -64,7 +64,7 @@ class Node {
 
   std::string Name() const { return name_; }
 
-  VarDesc* Var() {
+  VarDesc* Var() const {
```
panyx0718, line 67: LG

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -121,6 +124,108 @@ std::map<ir::Node *, std::unordered_set<ir::Node *>> BuildOperationAdjList(
...
+  }
+  return adj_list;
+}
+
+std::vector<ir::Node *> OpDFSSort(const Graph &graph) {
+  auto edge_table = BuildOperationOutAdjList(graph);
+  std::stack<Node *> stack;
+  // find the feed ops
+  for (auto &ele : edge_table) {
+    if (ele.first->Name() == "feed") {
```
panyx0718: seems not safe. there might not be feed operators in model. Do want to implement a function that search roots ops?

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -203,6 +308,28 @@ size_t GraphNum(const Graph &graph) {
...
+  }
+
+  for (auto *node : nodes2rm) {
+    graph->RemoveNode(node);
+  }
+}
+
+std::vector<Node *> TopologyVarientSort(const Graph &graph, int sort_kind) {
+  switch (sort_kind) {
+    case 0:
```
panyx0718: Need to be enum with a better name

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -121,6 +124,108 @@ std::map<ir::Node *, std::unordered_set<ir::Node *>> BuildOperationAdjList(
...
+    if (node->IsOp()) {
+      op_queue.push_back(node);
+    }
+  }
+
+  // traverse the graph
+  int num_ops = op_queue.size();
+  while (num_ops) {
+    for (auto it = op_queue.begin(); it != op_queue.end(); it++) {
+      auto *&cur_op = *it;
```
panyx0718, line 221: Why `*&`?

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -121,6 +124,108 @@ std::map<ir::Node *, std::unordered_set<ir::Node *>> BuildOperationAdjList(
...
+      if (!cur_op || in_degree[cur_op] > 0) continue;
```
panyx0718, line 222: why cur_op can be nullptr?

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -121,6 +124,108 @@ std::map<ir::Node *, std::unordered_set<ir::Node *>> BuildOperationAdjList(
...
+      // visit this node
+      // put all the output var of this op valid.
+      for (auto *out_var : cur_op->outputs) {
+        if (!out_var) continue;
```
panyx0718, line 226: why can be nullptr?

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -121,6 +124,108 @@ std::map<ir::Node *, std::unordered_set<ir::Node *>> BuildOperationAdjList(
...
+    for (auto *op : edge_table[cur]) {
+      if (!unique_set.count(op)) {
+        stack.push(op);
+      }
+    }
+  }
+  return res;
+}
+
+std::vector<ir::Node *> TopologyDfsSortOperations(const Graph &graph) {
```
panyx0718, line 190: @sneaxiy this algorithm seems to duplicate in a few places. Should unify in graph_helper

paddle/fluid/framework/ir/graph_helper.h
```diff
@@ -34,6 +34,15 @@ size_t GraphNum(const Graph &graph);
 // `graph` cannot contain circle.
 std::vector<ir::Node *> TopologySortOperations(const Graph &graph);
 
+// Topological sort, but try to DFS.
+std::vector<ir::Node *> TopologyDfsSortOperations(const Graph &graph);
```
panyx0718, line 38: Seems this one and the above one duplicate the one below? Remove one of them?

paddle/fluid/framework/ir/graph_helper.h
```diff
@@ -34,6 +34,15 @@ size_t GraphNum(const Graph &graph);
...
+
+// Several kinds of topological sort.
+std::vector<Node *> TopologyVarientSort(const Graph &graph, int sort_kind);
```
panyx0718: document the sort kind?

paddle/fluid/framework/naive_executor.cc
```diff
@@ -90,6 +92,7 @@ void NaiveExecutor::CreateVariables(const ProgramDesc &desc, int block_id,
       }
     }
   }
+  LOG(INFO) << "naive executor create " << num_vars << " vars";
```
panyx0718: VLOG?

paddle/fluid/inference/analysis/analyzer.cc
```diff
@@ -17,20 +17,23 @@
...
-void Analyzer::RunIrAnalysis(Argument *argument) {
-  std::vector<std::string> passes({"ir_analysis_compose_pass"});
-
-  for (auto &pass : passes) {
-    PassRegistry::Global().Retreive(pass)->Run(argument);
+void Analyzer::RunAnalysis(Argument *argument) {
+  PADDLE_ENFORCE(argument->analysis_passes_valid());
+  for (auto &pass : argument->analysis_passes()) {
+    string::PrettyLogH1("--- Running analysis [%s]", pass);
+    auto *ptr = PassRegistry::Global().Retreive(pass);
```
panyx0718, line 34: Why is this Retrieve? Seems this is ok:
`ir::PassRegistry::Instance().Get`

paddle/fluid/inference/api/analysis_predictor.cc
```diff
@@ -202,7 +211,7 @@ bool AnalysisPredictor::Run(const std::vector<PaddleTensor> &inputs,
 
 bool AnalysisPredictor::SetFeed(const std::vector<PaddleTensor> &inputs,
                                 framework::Scope *scope) {
-  VLOG(3) << "Predictor::set_feed";
+  VLOG(3) << "Pred5ictor::set_feed";
```
panyx0718: why?

paddle/fluid/inference/analysis/passes/memory_optimize_pass.cc
```diff
@@ -0,0 +1,642 @@
...
+  return res;
+}
+
+// Deserialize the batch var shapes from the cache file.
+std::vector<std::map<std::string, std::vector<int>>> DeseralizeBatchVarShapes(
+    const std::string& path) {
+  std::ifstream file(path);
+  PADDLE_ENFORCE(file.is_open(), "failed to open %s  to read cache", path);
+  std::string line;
+  std::vector<std::map<std::string, std::vector<int>>> batch_shapes;
```
panyx0718, line 429: doc the file format here?

paddle/fluid/inference/api/analysis_predictor.cc
```diff
@@ -331,19 +342,25 @@ void AnalysisPredictor::OptimizeInferenceProgram() {
...
   if (config_.use_mkldnn_) {
+    LOG(INFO) << "MKLDNN is enabled";
     argument_.SetMKLDNNEnabledOpTypes(config_.mkldnn_enabled_op_types_);
   }
 
   auto passes = config_.pass_builder()->AllPasses();
-  if (!config_.ir_optim()) passes.clear();
+  if (!config_.ir_optim()) {
+    passes.clear();
+    LOG(INFO) << "ir_optim is turned off, no IR pass will be executed";
```
panyx0718, line 360: there are many LOG, should change to VLOG if they are debug information.

paddle/fluid/framework/ir/graph_helper.cc
panyx0718: take a look at sequential_execution_pass.cc

paddle/fluid/framework/ir/graph_helper.cc
panyx0718: take a look as sequential_execution_pass.cc example

Superjomn: done

paddle/fluid/framework/ir/graph_helper.cc
Superjomn: done

paddle/fluid/framework/ir/graph_helper.cc
Superjomn: the cur_op in the queue will be set to null if visited.
The ops in the queue follow the DFS order, but we traverse the graph in topological order, so each iteration, just several ones in the queue will be visited, to simplify erasing records in the middle of the queue, just set them to null.

paddle/fluid/framework/ir/graph_helper.cc
Superjomn: Mark it as visited.

paddle/fluid/framework/ir/graph_helper.cc
Superjomn: Just in case, with a check, remove this line also works.

paddle/fluid/framework/ir/graph_helper.cc
Superjomn: done

paddle/fluid/framework/ir/graph_helper.h
Superjomn: the above one is the primary topology sort, the below one is tepology + DFS sort, they are different algorithm.

paddle/fluid/framework/ir/graph_helper.h
Superjomn: Add an enum for `sort_kind` and some documents above.

paddle/fluid/inference/analysis/analyzer.cc
Superjomn: These are `AnalysisPass`, which takes Argument as input, https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/inference/analysis/argument.h

The `AnalysisPass` are used to implement some complex inference specific pass, and can treat as the compose of `ir::Pass`es.

paddle/fluid/inference/api/analysis_predictor.cc
Superjomn: done

paddle/fluid/inference/api/analysis_predictor.cc
Superjomn: This information is important and should display to users, reminding them that this config is turned off.

paddle/fluid/inference/analysis/passes/memory_optimize_pass.cc
Superjomn: done

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -121,6 +124,119 @@ std::map<ir::Node *, std::unordered_set<ir::Node *>> BuildOperationAdjList(
...
+
+std::vector<ir::Node *> OpDFSSort(const Graph &graph) {
+  auto edge_table = BuildOperationOutAdjList(graph);
+  std::stack<Node *> stack;
+  for (auto &ele : edge_table) {
+    if (ele.first->inputs.empty()) {
+      // find the input ops (those without input vars)
+      stack.push(ele.first);
+    } else {
+      // find the ops with only persistable vars as inputs.
```
panyx0718, line 158: why only persistable vars?

paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -203,6 +319,29 @@ size_t GraphNum(const Graph &graph) {
   return graph_count;
 }
 
+void CleanIndividualNodes(Graph *graph) {
```
panyx0718, line 322: Call CleanUnconnectedNode? When can this happen?

paddle/fluid/inference/analysis/passes/memory_optimize_pass.cc
```diff
@@ -0,0 +1,647 @@
...
+
+  VLOG(3) << "Cluster by interval and get " << res.size() << " cluster";
+  return res;
+}
+
+std::string MemoryOptimizePass::repr() const { return "memory optimize pass"; }
+
+void MemoryOptimizePass::RunImpl(Argument* argument) {
+  // When force update, should not optimize memory.
+  if (!argument->enable_memory_optim() || argument->memory_optim_force_update())
```
panyx0718, line 543: this should normally live in pass manager

panyx0718 (approved):

panyx0718:
The strategies are fancy and heuristic. should follow up with more tests...

paddle/fluid/framework/ir/graph_helper.cc
Superjomn: The original program will create the VarNodes in the beginning. 
After memory reusing, some nodes of them are not needed and no edge links to them, those nodes need to be removed.

paddle/fluid/inference/analysis/passes/memory_optimize_pass.cc
Superjomn: Currently, the inference memory optimize performs in two stages:

1. Execute the inference program on the real data, store the shapes of variables into the disk. This stage is offline.
2. Online in the real deployment, load the shapes persisted on the disk, perform the memory reusing strategy, and run inference. This phase is where the memory optimize performs.

The `enable_memory_optim` config will make the predictor perform those two stages, where the first stage is not suitable to perform in the real online deployment, so the config needs to be carefully set in the real application, and cannot be switched on by default.

We will add another simple strategy that doesn't need the first stage, and then, the memory optimize can be turned on by default.

PR #14962, opened 2018-12-19
Rewrite variable type

sneaxiy:
Rewrite variable type. Users can only store registered types inside `Variable`.

Benefits may be from:
- Speed up `Variable`. Time complexity of all functions in `Variable` class is `O(1)` now. No virtual function calls (except destructor), and no comparing string in `IsType()`.
- We can know how many kinds of types that have been created inside `Variable`.

Moreover, registration of new variable types is easy. Users only need to add the type inside a varidiac template class `VarTypeRegistry`.


paddle/fluid/framework/operator.h
```diff
@@ -310,6 +310,18 @@ class ExecutionContext {
   const RuntimeContext& ctx_;
 };
 
+inline bool CanCUDNNBeUsed(const framework::ExecutionContext& ctx) {
```
Superjomn: CanUseCudnn or CanUseCUDNN?

sneaxiy: Maybe it is not important. Similar name appears such as `CanMKLDNNBeUsed`. I just move this function from `cudnn_helper.h` to `operator.h`, because it accepts `framework::ExecutionContext` as its parameter. It is better to put it inside `operator.h`.

paddle/fluid/framework/var_type.h
```diff
@@ -19,35 +19,33 @@ limitations under the License. */
...
-    return proto::VarType_Type_LOD_TENSOR_ARRAY;
-  } else if (IsType<SelectedRows>(type)) {
-    return proto::VarType_Type_SELECTED_ROWS;
-  } else if (IsType<ReaderHolder>(type)) {
-    return proto::VarType_Type_READER;
-  } else {
-    PADDLE_THROW("ToVarType:Unsupported type %s", type.name());
+inline proto::VarType::Type ToVarType(int type) {
+  switch (type) {
+    case proto::VarType::LOD_TENSOR:
```
dzhwinter, line 35: break?

sneaxiy: No break. Because `LOD_TENSOR`, `SELECTED_ROWS`, etc are all acceptable. We just need to cast `int` to proto id.

paddle/fluid/framework/var_type_traits.h
```diff
@@ -0,0 +1,184 @@
...
+#endif
+    int, float>;
+
+template <typename T>
+struct VarTypeTrait {
+  static_assert(VarTypeRegistry::IsRegistered<T>(), "Must be registered type");
+  using Type = T;
+  // Default id generation
+  static constexpr int kId = VarTypeRegistry::TypePos<T>() +
+                             static_cast<int>(proto::VarType::TUPLE) * 2;
```
chengduoZH, line 171: I am curious that why the `kId` is ` VarTypeRegistry::TypePos<T>() +     static_cast<int>(proto::VarType::TUPLE) * 2`?

sneaxiy: Because `proto::VarType::TUPLE` is the maximum protobuf id. To generate unique id, we should skip all protobuf ids defined in `framework.proto`.

chengduoZH (approved):
LGTM

PR #14333, opened 2018-11-08
Changed hardcoded format to any in convolution and bumped MKL-DNN version to 0.17rc

kbinias:
This changes give huge performance boost on MobileNet-v1

PR #14932, opened 2018-12-18
Fix the regression issue and add the group unitest for conv2d

yihuaxu:
Fix [ISSUE#14883](https://github.com/PaddlePaddle/Paddle/issues/14883) and add the unit testing for conv2d. 

**Command**: 
    FLAGS_use_mkldnn=true FLAGS_paddle_num_threads=20 KMP_AFFINITY=compact,granularity=fine ./paddle/fluid/inference/tests/api/test_analyzer_resnet50 --infer_model=third_party/inference_demo/mobilenet_depthwise_conv/model --repeat=100 --gtest_filter=Analyzer_resnet50.profile_mkldnn --batch_size=1 --num_threads=1 --profile --paddle_num_threads=28

PR #14970, opened 2018-12-19
fix weights memory format for grouped convolutions

wojtuss:
In the case of grouped convolution with MKL-DNN unnecessary reorders `nchw16c->nchw` and back occur. This was fixed previously with https://github.com/PaddlePaddle/Paddle/pull/14333 but came up again after some recent refactoring.

test=develop


sfraczek (approved):
LGTM

luotao1:
Is this duplicated with #14932?

yihuaxu:
@wojtuss It seems the issue be similar with [PR#14932](https://github.com/PaddlePaddle/Paddle/pull/14932). Could you help me review if it is the duplication for your PR?

hshen14:
14932 looks more general fixing with group convolution UT enabling.

wojtuss:
@luotao1 , @yihuaxu , @hshen14 , I have verified that #14932 also fixes this issue.

luotao1:
Thus, which PR should be merged?

PR #14973, opened 2018-12-19
Add Dequantize OP

xiaolil1:
Split from #14929 
test=develop

Quantize OP is in #14974

PR #14974, opened 2018-12-19
Add Quantize OP

xiaolil1:
Split from #14929 
test=develop

Dequantize OP is in #14973

### DIFF
--- a/paddle/fluid/operators/quantize_mkldnn_op.cc
+++ b/paddle/fluid/operators/quantize_mkldnn_op.cc
@@ -16,6 +16,7 @@ limitations under the License. */
 #include "paddle/fluid/framework/tensor.h"
 #include "paddle/fluid/operators/quantize_op.h"
 #include "paddle/fluid/platform/mkldnn_helper.h"
+#include "paddle/fluid/platform/mkldnn_reuse.h"
 
 namespace paddle {
 namespace operators {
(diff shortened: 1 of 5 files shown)


paddle/fluid/operators/quantize_mkldnn_op.cc
```diff
@@ -0,0 +1,100 @@
...
+      dst_memory.reset(
+          new mkldnn::memory(dst_pd, to_void_cast<int8_t>(output_data)));
+    } else {
+      uint8_t* output_data = output->mutable_data<uint8_t>(ctx.GetPlace());
+      auto dst_md = platform::MKLDNNMemDesc({dst_tz}, memory::data_type::u8,
+                                            memory::format::nhwc);
+      dst_pd = mkldnn::memory::primitive_desc(dst_md, engine);
+      dst_memory.reset(
+          new mkldnn::memory(dst_pd, to_void_cast<uint8_t>(output_data)));
+    }
```
tpatejko, line 71: The if part differs from else part with just a type of data. It's either `uint8_t` or `int8_t`.

You could move the code to a template function and parametrize it with a correct type.

paddle/fluid/operators/quantize_mkldnn_op.cc
```diff
@@ -0,0 +1,100 @@
...
+
+    auto src_md = platform::MKLDNNMemDesc({src_tz}, memory::data_type::f32,
+                                          input->format());
+    auto src_pd = mkldnn::memory::primitive_desc(src_md, engine);
+    auto src_memory =
+        std::make_shared<mkldnn::memory>(src_pd, to_void_cast<T>(input_data));
+    std::shared_ptr<primitive::at> src_memory_p =
+        std::shared_ptr<primitive::at>(new primitive::at(*src_memory));
+
+    bool is_negative = ctx.Attr<bool>("is_negative_input");
```
tpatejko, line 62: Where and how is this attribute used?

paddle/fluid/operators/quantize_mkldnn_op.cc
xiaolil1: yes, I will optimize it with template

paddle/fluid/operators/quantize_mkldnn_op.cc
xiaolil1: it is the flag of src data_type, is_negtive true for S8 and false for U8.

tpatejko: @xiaolil1 I understand it's set by the calibration tool?

paddle/fluid/operators/quantize_mkldnn_op.cc
```diff
@@ -0,0 +1,100 @@
...
+    int mask = 0;
+    attri.set_output_scales(mask, {scale_data});
+
+    auto src_md = platform::MKLDNNMemDesc({src_tz}, memory::data_type::f32,
+                                          input->format());
+    auto src_pd = mkldnn::memory::primitive_desc(src_md, engine);
+    auto src_memory =
+        std::make_shared<mkldnn::memory>(src_pd, to_void_cast<T>(input_data));
+    std::shared_ptr<primitive::at> src_memory_p =
+        std::shared_ptr<primitive::at>(new primitive::at(*src_memory));
```
Sand3r-, line 60: In the future it'd be worth to consider making a re-use so that all the memory allocations could be avoided.

paddle/fluid/operators/quantize_mkldnn_op.cc
```diff
@@ -0,0 +1,100 @@
...
+      dst_pd = mkldnn::memory::primitive_desc(dst_md, engine);
+      dst_memory.reset(
+          new mkldnn::memory(dst_pd, to_void_cast<int8_t>(output_data)));
+    } else {
+      uint8_t* output_data = output->mutable_data<uint8_t>(ctx.GetPlace());
+      auto dst_md = platform::MKLDNNMemDesc({dst_tz}, memory::data_type::u8,
+                                            memory::format::nhwc);
+      dst_pd = mkldnn::memory::primitive_desc(dst_md, engine);
+      dst_memory.reset(
+          new mkldnn::memory(dst_pd, to_void_cast<uint8_t>(output_data)));
```
Sand3r-: In the future it might be worth to extract both branches into a single templatized function that would accept int8/uint8 and memory::data_type as parameters.

Sand3r- (dismissed):
Generally it Looks Good To Me. Please consider introducing the proposed changes in the future.

paddle/fluid/operators/quantize_mkldnn_op.cc
luotao1: > You could move the code to a template function and parametrize it with a correct type

agree with it

paddle/fluid/operators/quantize_mkldnn_op.cc
luotao1: > extract both branches into a single templatized function

How about extract in this PR? if in the future, when will extract them?

paddle/fluid/operators/quantize_mkldnn_op.cc
luotao1: > it'd be worth to consider making a re-use so that all the memory allocations could be avoided.

How about make a re-use in this PR? if in the future, when will make a re-use?

xiaolil1: Sure, we will do it in next PR, thanks!

paddle/fluid/operators/quantize_mkldnn_op.cc
xiaolil1: yes, it's set by calibration tool.

paddle/fluid/operators/quantize_mkldnn_op.cc
xiaolil1: We will extract in this PR.

tensor-tang (approved):
LGTM

PR #14978, opened 2018-12-20
fix fuse_broadcast_op not exported in BuildStrategy

carlushuang:
in https://github.com/PaddlePaddle/Paddle/blob/develop/benchmark/fluid/fluid_benchmark.py#L182

```
  build_strategy.fuse_broadcast_op = args.fuse_broadcast_op
```

however `fuse_broadcast_op` of `BuildStrategy` is not exported by pybind, hence may result in crash when run fluid_benchmark.py

need to fix this by exporting fuse_broadcast_op in pybind.cc


carlushuang:
close due to nobody care anymore

PR #14992, opened 2018-12-21
Fp16 training

typhoonzero:


python/paddle/fluid/layers/learning_rate_scheduler.py
```diff
@@ -230,17 +231,17 @@ def polynomial_decay(learning_rate,
         if cycle:
             div_res = ops.ceil(global_step / decay_steps)
             zero_var = tensor.fill_constant(
-                shape=[1], dtype='float32', value=0.0)
+                shape=[1], dtype="float32", value=0.0)
             one_var = tensor.fill_constant(
-                shape=[1], dtype='float32', value=1.0)
+                shape=[1], dtype="float32", value=1.0)
```
chengduoZH: Why do you change this?

python/paddle/fluid/optimizer.py
```diff
@@ -74,6 +74,12 @@ def _create_global_learning_rate(self):
         lr = self._global_learning_rate()
 
         if isinstance(lr, framework.Variable):
+            # NOTE: cast to optimizer dtype if needed, this is for
+            # training with fp16 which learning rate should also be
+            # a 
+            if self._dtype:
+                if lr.dtype != self._dtype:
+                    layers.cast(lr, self._dtype)
```
chengduoZH: There doesn't have variable to accept the return of `layers.cast`.

python/paddle/fluid/layers/learning_rate_scheduler.py
typhoonzero: Done.

python/paddle/fluid/optimizer.py
typhoonzero: Removed and will add in next PR

chengduoZH (approved):

python/paddle/fluid/layers/nn.py
```diff
@@ -2797,6 +2797,10 @@ def batch_norm(input,
     helper = LayerHelper('batch_norm', **locals())
     dtype = helper.input_dtype()
 
+    # use fp32 for bn parameter
+    if dtype == core.VarDesc.VarType.FP16:
```
panyx0718, line 2801: I don't like this hardcode. For now, we should change the model and add cast op before bn.

python/paddle/fluid/layers/nn.py
```diff
@@ -2797,6 +2797,10 @@ def batch_norm(input,
...
+        dtype = core.VarDesc.VarType.FP32
```
panyx0718, line 2802: what's the output type, is it fp16 or fp32. What if the next op needs fp16?

python/paddle/fluid/layers/nn.py
typhoonzero: Seems currently all fp16 training should use batch norm with fp32 parameters, which includes moving mean, var and trainable scale, bias

python/paddle/fluid/layers/nn.py
typhoonzero: The output type is still fp16, but the parameters are all fp32. please ref: https://docs.nvidia.com/deeplearning/sdk/mixed-precision-training/index.html. and the current kernel implementation does this too.

typhoonzero: Well the only difficulty is that when we copy "master" parameters for mix precision, we must skip batch_norm layer's all parameters.

python/paddle/fluid/layers/nn.py
panyx0718: Then we should add arguments to bn layer to tell bn to cast parameter to fp16.

python/paddle/fluid/layers/nn.py
panyx0718: this is strange. why input is fp16, output is fp32? What if I want to infer type, output should by fp16 when input is fp16?

typhoonzero: The output is fp16.

explicit is better than implicit, should add an argument to control this. Will do this.

PR #15012, opened 2018-12-24
Add temp variable pool in Scope

sneaxiy:
This PR is mainly designed to remove `scope.newScope()` in `OperatorWithKernel::RunImpl()`.


chengduoZH (commented):
Great! 
The temporary variable can be managed by `TemporaryAllocator` and `TempVariablePool` now, but the time point of freeing temp var is different of the two strategies. Not sure which is better now.

PR #15020, opened 2018-12-24
Add the CUDA kernel for beam_search op

Xreki:
This PR do following things:
- Implement the CUDA kernel for `beam_search` op
- Refine the CPU kernel, ensure the selected items sorted in descending order by scores
- Add `is_accumulated` attribute. If `is_accumulated` is set to `false`, current `scores` will be transformed to the log field and accumulated `pre_scores`.
- Allow `ids` to be null. If it is null, `beam_search` op will calculate out `ids` according to `scores`.

In the develop version, we need 5 operators to configure `beam_search` algorithm.
```python
        current_scores = fluid.layers.softmax(noacted_current_score, use_cudnn=use_cudnn)

        topk_scores, topk_indices = fluid.layers.topk(current_scores, k=beam_size)

        # calculate accumulated scores after topk to reduce computation cost
        accu_scores = fluid.layers.elementwise_add(
            x=fluid.layers.log(topk_scores),
            y=fluid.layers.reshape(pre_score, shape=[-1], inplace=True),
            axis=0)

        selected_ids, selected_scores = fluid.layers.beam_search(
            pre_ids,
            pre_score,
            topk_indices,
            accu_scores,
            beam_size,
            1,  # end_id
            #level=0
        )
```

After this PR, only 1 operators is enough.
```python
        current_scores = fluid.layers.softmax(noacted_current_score, use_cudnn=use_cudnn)

        selected_ids, selected_scores = fluid.layers.beam_search(
            pre_ids=pre_ids,
            pre_scores=pre_score,
            ids=None, 
            scores=current_scores,
            beam_size=beam_size,
            end_id=1,
            is_accumulated=False
            #level=0
        )
```

Speedup for ocr's 1d-attention model: `16.0768 ms` -> `14.7455 ms`


paddle/fluid/operators/math/beam_search.cc
```diff
@@ -0,0 +1,279 @@
...
+            Insert(&top_beam, item, beam_size);
+          }
+        }
+      }
+
+      result.emplace_back(top_beam);
+    }
+
+    VLOG(3) << "SelectTopBeamSizeItems result size " << result.size();
+    for (auto &items : result) {
```
panyx0718: only run if `if VLOG_IS_3`?

paddle/fluid/framework/operator.cc
```diff
@@ -973,6 +973,10 @@ Scope* OperatorWithKernel::PrepareData(
     const Scope& scope, const OpKernelType& expected_kernel_key,
     std::vector<std::string>* transfered_inplace_vars,
     RuntimeContext* ctx) const {
+  if (type_ == "is_empty") {
```
panyx0718: why need this? Is there a better way? abstraction shouldn't depend on a specific operator implementation.

paddle/fluid/operators/beam_search_op.cc
```diff
@@ -242,6 +53,7 @@ class BeamSearchOpMaker : public framework::OpProtoAndCheckerMaker {
     AddAttr<int>("beam_size", "beam size for beam search");
     AddAttr<int>("end_id",
                  "the token id which indicates the end of a sequence");
+    AddAttr<bool>("is_accumulated", "").SetDefault(true);
```
panyx0718: doc?

paddle/fluid/operators/math/beam_search.cu
```diff
@@ -0,0 +1,375 @@
...
+  const int tid = threadIdx.x;
+  const int tid_of_seq = threadIdx.x % NumThreadsPerSeq;
+
+  int num_used_threads = (seq_width + beam_size - 1) / beam_size;
+  num_used_threads =
+      NumThreadsPerSeq < num_used_threads ? NumThreadsPerSeq : num_used_threads;
+
+  num_used_threads =
+      num_used_threads > 32
+          ? (num_used_threads >> 5) << 5
```
panyx0718: Can this code block be pre-computed in CPU? Seems slow on GPU.

paddle/fluid/operators/math/beam_search.cu
```diff
@@ -0,0 +1,375 @@
...
+    Triple* top_beam, const int64_t* pre_ids, const float* pre_scores,
+    const int64_t* ids, const float* scores, const int seq_offset_start,
+    const int seq_offset_end, const int seq_width, int beam_size, int end_id) {
+  // top_beam is shared memory
+  const int tid = threadIdx.x;
+  const int tid_of_seq = threadIdx.x % NumThreadsPerSeq;
+
+  int num_used_threads = (seq_width + beam_size - 1) / beam_size;
+  num_used_threads =
+      NumThreadsPerSeq < num_used_threads ? NumThreadsPerSeq : num_used_threads;
```
panyx0718: Can these be computed on CPU?

paddle/fluid/platform/cuda_device_function.h
```diff
@@ -30,6 +31,34 @@ namespace platform {
   mask = __ballot_sync(FULL_WARP_MASK, (predicate))
 #endif
 
+inline static int GetPowerOfTwo(int dim) {
```
panyx0718: RoundToPowerOfTwo?

python/paddle/fluid/layers/nn.py
```diff
@@ -3704,6 +3704,7 @@ def beam_search(pre_ids,
                 beam_size,
                 end_id,
                 level=0,
+                is_accumulated=True,
```
panyx0718, line 3878: comment for this arg?

paddle/fluid/framework/operator.cc
Xreki: Because the current `is_empty` only implements the CPU kernel. But there is no need to access the input tensor's elements, but just need to get the `dims`.

https://github.com/PaddlePaddle/Paddle/blob/ce093d2d5c9a3fd1dd546cc3c979f660588fdd1e/paddle/fluid/operators/is_empty_op.h#L31-L32

So if the input tensor is on GPU, in fact there is no need to transfer it to CPU. To avoid this and if we don't want to add this `if` statement here, we need to rewrite the `is_empty` op.

paddle/fluid/operators/beam_search_op.cc
Xreki: Sorry for forgetting to add doc. Done.

paddle/fluid/operators/math/beam_search.cc
Xreki: Done.

paddle/fluid/operators/math/beam_search.cu
Xreki: Done.

paddle/fluid/operators/math/beam_search.cu
Xreki: Done.

paddle/fluid/platform/cuda_device_function.h
Xreki: Done.

python/paddle/fluid/layers/nn.py
Xreki: Done.

panyx0718:
@guoshengCS Do you own this OP? Please review

paddle/fluid/operators/math/beam_search.h
```diff
@@ -0,0 +1,119 @@
+/* Copyright (c) 2016 PaddlePaddle Authors. All Rights Reserved.
+
```
Superjomn, line 2: beam_search 需要挪到 math 下面吗？ 理论上貌似不是数学计算

Xreki: `math`下面不全是数学计算函数吧，比如`im2col`、`seq2batch`这些？我理解`math`下面就是一个放`functor`的地方。

Superjomn: 那其实建议把 math rename 成 `impl` 或者 `funcs` ，math 不是 functor 的意思。

后续再看看，暂时没有问题

paddle/fluid/operators/math/beam_search.cu
```diff
@@ -0,0 +1,393 @@
...
+    Triple* top_beam_local, const int seq_offset_start,
+    const int seq_offset_end, const int selected_seq_start,
+    const int selected_seq_length) {
+  const int tid = threadIdx.x;  // use 1 thread only for each sequence
+  int global_index = selected_seq_start;
+  for (int global_offset = seq_offset_start; global_offset < seq_offset_end;
+       ++global_offset) {
+    for (int local_index = 0; local_index < selected_seq_length;
+         ++local_index) {
+      if (top_beam_local[local_index].offset == global_offset) {
```
guoshengCS, line 170: 确认下哈，这里还是和原先beam_search_op中一样，CPU和GPU kernel输出的selected_id是按照parent在pre_id中的index排序的是吧

paddle/fluid/operators/math/beam_search.h
Xreki: 好的，后面可以统一整理下。

paddle/fluid/operators/math/beam_search.cu
Xreki: 对的。但原来的`beam_search op`实现使用`std::nth_element`，当`index`一样时，不能保证所选取出的元素的顺序。这个PR进行了修改，当`index`一样是，按`score`降序存储。

paddle/fluid/framework/operator.cc
panyx0718: I think we should rewrite is_empty instead of adding if here.

paddle/fluid/platform/cuda_device_function.h
```diff
@@ -30,6 +31,34 @@ namespace platform {
   mask = __ballot_sync(FULL_WARP_MASK, (predicate))
 #endif
 
+inline static int RoundToPowerOfTwo(int dim) {
+  if (dim > 512) {
```
panyx0718, line 35: what if it's 2048?

Xreki: This is usually used to set the number of CUDA threads in a block, normally 1024 is the maximum. I just list some common values as here:

https://github.com/PaddlePaddle/Paddle/blob/d243e555eb5856d5642706e1aa0d0800b146cae5/paddle/fluid/operators/top_k_op.cu#L295-L305

paddle/fluid/framework/operator.cc
Xreki: Can I do this in next PR?

PR #15024, opened 2018-12-24
test=develop, fix ci by install requirement and add pip install valid…

JiabinYang:
…ation

### DIFF
--- a/paddle/scripts/paddle_build.sh
+++ b/paddle/scripts/paddle_build.sh
@@ -444,7 +444,7 @@ EOF
         if [ "$1" == "cp27-cp27m" ]; then
             set -e
             pip install --user ${INSTALL_PREFIX:-/paddle/build}/opt/paddle/share/wheels/*.whl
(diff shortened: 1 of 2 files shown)


paddle/scripts/paddle_build.sh
```diff
@@ -441,7 +442,9 @@ EOF
         # make install should also be test when unittest
         make install -j 8
         if [ "$1" == "cp27-cp27m" ]; then
+            set -e
             pip install --user ${INSTALL_PREFIX:-/paddle/build}/opt/paddle/share/wheels/*.whl
+            python ${PADDLE_ROOT}/paddle/scripts/installation_validate.py
```
velconia, line 450: Need some validation here?

JiabinYang: yes, this is used to validate `import paddle.fluid` here for now, and we can add more validate here later

wanghaoshuang (approved):
LGTM

PR #15048, opened 2018-12-25
Add conv+ affine channel fuse pass 

NHZlX:
1. The [faster rcnn model](https://github.com/PaddlePaddle/models/tree/develop/fluid/PaddleCV/faster_rcnn) contains the pattern of "conv + affine channel ", this patten can be detect and the two ops can be transformed to a fused op.

2. conv+ elementwise add bug.

3. Faster rcnn input is presistable, the paddle-trt will regard those variables as parameters, so,  I add a rule to avoid this. 

4. add min_subgraph_size attr for tensorrt config.


paddle/fluid/inference/api/paddle_pass_builder.h
```diff
@@ -118,7 +118,9 @@ class GpuPassStrategy : public PassStrategy {
  public:
   GpuPassStrategy() : PassStrategy({}) {
     passes_.assign({
-        "infer_clean_graph_pass",               //
+        "infer_clean_graph_pass",  //
+        "conv_affine_channel_fuse_pass",
```
Superjomn: ends with `//`

NHZlX: Done!

Superjomn (approved):
LGTM

PR #15062, opened 2018-12-26
[V1.3] Add the calibration tool code for int8 inference and focus test.

guomingz:
We developed the calibration tool for int8 inference.
Here is the document for this tool.

# Introduction of Accuracy Calibration Tool for 8 Bit Inference
The 8-bits (**INT8**) inference is also known as the Low Precision Inference which could speed up the inference with the lower accuracy loss. It has higher throughput and lower memory requirements compared to FP32. As the PaddlePaddle enables the INT8 inference supporting, we release a accuracy tool(Calibration.py) at the same time. This tool will generate the  quantization parameters and quantized model file finally.


guomingz:
Hi @luotao1 . Due to the paddle side [code](https://github.com/PaddlePaddle/Paddle/pull/14929) was not merged  yet, this PR may fail on pre-ci test. Anyway, would u please start the code review firstly, especially on int8_inference/utility.py ? @hshen14

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+            for i in output_index_name.keys():
+                if input_name in output_index_name[i]:
+                    search_end_index = i
+                    found_ancestor = True
+                    break
+
+            if not found_ancestor:  # Dangling var
+                return Calibrator.s8_max
+            elif search_end_index == first_conv_op_index:  # first conv
+                if program.current_block().ops[i].has_attr(
```
sfraczek: I think that if you `set search_end_index = i` in 103 it would make sense to use that variable now instead of `i` here and in line below.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+                        'fuse_relu') and program.current_block().ops[i].attr(
+                            'fuse_relu'):
+                    return Calibrator.u8_max
+                else:
+                    return Calibrator.s8_max
+            elif ops_type[search_end_index] != 'conv2d':
+                continue
+            else:
```
sfraczek, line 124: This case is for all convolutions except the first, right? It seems to be handled exactly the same as the one for the first convolution. Why is the first convolution separated then? Can you reduce it to two cases (any convolution vs other operators)?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+        Returns:
+            Two lists contains the quantize op and dequantize op index information.
+        """
+        quantize_op_index = []
+        dequantize_op_index = []
+
+        if len(self._conv_op_index) < 2:
+            return [], []
+
+        for index, value in enumerate(self._conv_op_index):
```
sfraczek, line 420: Is quantization inserted [some operators] ->[here] -> [one operator] -> conv_op_index[index + 1]. which is before the [one operator] proceeding the second convolution (`self._conv_op_index[index + 1] - 1`) instead of right before it?
Is dequantization inserted as 2nd operator after the last convolution if its output goes to int8 operator (`self._conv_op_index[index] + 2`)? Can you explain logic behind this?
And you force fp32 output if the last convolution is going to non int8 operator?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+                    program.current_block().ops[value]._set_attr(
+                        "force_fp32_output", True)
+
+            else:
+                if self._conv_op_index[index] + 1 < self._conv_op_index[index +
+                                                                        1]:
+                    for op_index in range(self._conv_op_index[index] + 1,
+                                          self._conv_op_index[index + 1]):
+                        op_type = program.current_block().ops[op_index].type
+                        if op_type in "conv2d" and self._conv_op_index[
```
sfraczek: 1. If conv_op_index has all the convolutions, why are you checking if there are any convolutions in between two consecutive convolutions? I see you would skip them but are there any?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+                                op_index] < self._conv_op_index[index + 1]:
+                            continue
+                        elif op_type in Calibrator.non_conv_int8_op_type:
+                            # dequantize_op_index.append(op_index) #disable dequantize if enable force_fp32_output
+                            program.current_block().ops[op_index - 1]._set_attr(
+                                "force_fp32_output", True)
+                            break
+                        else:
```
sfraczek: Why is there another `else` that does the same thing? This logic looks fine to me but I'm not sure about the logic of the case above in line 441?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+            else:
+                if self._conv_op_index[index] + 1 < self._conv_op_index[index +
+                                                                        1]:
+                    for op_index in range(self._conv_op_index[index] + 1,
+                                          self._conv_op_index[index + 1]):
+                        op_type = program.current_block().ops[op_index].type
+                        if op_type in "conv2d" and self._conv_op_index[
+                                op_index] < self._conv_op_index[index + 1]:
+                            continue
+                        elif op_type in Calibrator.non_conv_int8_op_type:
```
sfraczek: Do you have to force fp32 output on a layer if the next operator is int8 pooling?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+                                        op_index + 1].type == "conv2d":
+                                    continue
+                                elif program.current_block(
+                                ).ops[op_index +
+                                      1].type in Calibrator.non_conv_int8_op_type:
+                                    dequantize_op_index.append(op_index + 1)
+                                    break
+                                else:
+                                    program.current_block().ops[
+                                        op_index]._set_attr("force_fp32_output",
```
sfraczek: Why insert dequantize between int8 conv-> [here] -> pooling, but force_fp32_output of convolution which is after pooling? Do I get it wrong? What does force_fp32_output actually do?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+                                                            True)
+                                    continue
```
sfraczek: is this continue necessary?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+                                    ).ops[op_index].output(output_attr_name)[0]
+                                    if self.__get_op_index_by_output_var(
+                                            program, output_var_name,
+                                            op_index) != -1:
+                                        output_is_to_int8_op = True
+                                        break
+
+                                if share_input_flag or output_is_to_int8_op:
+                                    quantize_op_index.append(op_index - 1)
+                else:
```
sfraczek: is this and next line necessary?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,731 @@
...
+            The op index if the variable is the input of this op or -1 if the 
+            variable is not the input of any op. 
+        '''
+        for index, op in enumerate(program.current_block().ops[start_index:]):
+            if var_name in op.input_arg_names:
+                return index
+
+        return -1
+
+    def __get_quantize_dequantize_combination(self, program):
```
sfraczek, line 406: Could this function be made smaller? It's very complex, I'm having hard time analyzing this.

sfraczek (changes requested):
I couldn't finish the review yet because of little time. I can continue reviewing in the new year. Sorry.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Because we don't enable the int8 on first convolution, we separate the cases into two parts.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: No. The force_fp32_output flag is only enabled if the next op is a non-int8 op.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Oh, it's a legacy logic as we're intend to insert dequantize op if the next op is a non int8 op.
Later, we enabled the force_fp32_output attr on int8 conv op, so we abandoned the line442.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: This is a case we found on se-resnext. i'll paste the .dot file for this program. You may view the detail on http://www.webgraphviz.com/

guomingz: int8 conv-> [here] -> pooling is not correct as  the invalid indexing. For 'force_fp32_output of convolution which is after pooling', there's no this case. Force_fp32_output will remove extra dequantize op after conv to reduce the unnessarcy reorder

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Removable

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Removable

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: u're right, will update

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Two consecutive convolutions means no need to insert quantize/dequantize op anymore. The code here is for safety check.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: > Is quantization inserted [some operators] ->[here] -> [one operator] -> conv_op_index[index + 1]. which is before the [one operator] proceeding the second convolution (self._conv_op_index[index + 1] - 1) instead of right before it?

We just inserted the quantize op before  int8 conv2d in general.

> Is dequantization inserted as 2nd operator after the last convolution if its output goes to int8 operator (self._conv_op_index[index] + 2)? Can you explain logic behind this?
For example,  conv(int8)--->pooling(int8) where conv is the last conv op, if we inserted the dequantize after the conv, we can't levarage the pooling op which already enabled the int8 supporting. So we just insert the dequantize after the pool op.

> And you force fp32 output if the last convolution is going to non int8 operator?
yes.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Will have a try later

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+        ]
+
+        self._var_max_value_map = {}
+        self._var_max_range = {}
+        self._u8_output_var = []
+        self._s8_output_var = []
+        self._persistable_vars = []
+
+    def generate_sampling_program(self):
+        self.__init_analysis()
```
panyx0718, line 59: single underscore is enough.

python/paddle/fluid/contrib/tests/test_calibration.py
```diff
@@ -0,0 +1,221 @@
...
+            calibrator = ut.Calibrator(
+                program=infer_program,
+                pretrained_model=model_path,
+                iterations=100,
+                debug=False,
+                algo=algo)
+
+            sampling_data = {}
+
+            calibrator.generate_sampling_program()
```
panyx0718, line 175: Is this a transpiler? @wzzju Can this be implemented in pass?

wzzju: @panyx0718 Not yet. Because this pr is based on `Program`, however, our pass is based on `Graph`.

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: Hmm, but those two cases have the exact same instructions? Maybe they could be put together?
```python
 if program.current_block().ops[search_end_index].has_attr(
                        'fuse_relu') and program.current_block().ops[
                            search_end_index].attr('fuse_relu'):
                    return Calibrator.u8_max
```

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+        self.__display_debug()
+
+    def __display_debug(self):
+        if self.debug:
+            self.__dot(self._output_program)
+            print(self._output_program)
+
+    def __get_max_range_by_var_name(self, program, var_name):
+        """
+        Check the specified variable was generatd from Relu layer or not.
```
sfraczek: 1. 'generatd' -> 'generated' 
2. I think it would be super great to have a note that it returns Calibration.u8 if it was generated by relu and s8 otherrwise. I like the comment that starts at line 132.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+
+            for i in output_index_name.keys():
+                if input_name in output_index_name[i]:
+                    search_end_index = i
+                    found_ancestor = True
+                    break
+
+            if not found_ancestor:  # Dangling var
+                return Calibrator.s8_max
+            elif search_end_index == first_conv_op_index:  # first conv
```
sfraczek: Could you change this `elif` to `if` (and add a newline)? I think it should work the same and it would make the search for ancestor a separate small complete block of functionality easier to analyse.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+                if program.current_block().ops[search_end_index].has_attr(
+                        'fuse_relu') and program.current_block().ops[
+                            search_end_index].attr('fuse_relu'):
+                    return Calibrator.u8_max
+                else:
+                    return Calibrator.s8_max
+            elif ops_type[search_end_index] != 'conv2d':
+                continue
```
sfraczek, line 118: If this branch of code is executed, wouldn't it end up in infinite loop? I can't find anything changing after this `continue`. Is there something I missed?

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: Alternatively, you could put the below lines inside a `if found_ancestor:` and then rely on the last line of this function to return s8_max as default. Going further, you could also remove lines 114-115 because I think it would go back to that last return. By applying few changes you could make this function reduced to have only two return statements, one at the end and one in the merged from the two cases where u8_max is returned. If you would find any ambiguities after these changes they could be shortly explained in the function's comment but I think it shouldn't be necessary but actually easier to understand. What do you think?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+        Check the specified variable is the output of int8 op or not.
+        If true, return the original op index.
+        If false, return -1
+        '''
+        return self._int8_output_var_op_index_dict[
+            var_name] if var_name in self._int8_output_var_op_index_dict else -1
+
+    def __update_int8_output_var_op_index_dict(self, index, var_name=None):
+        '''
+        Update the int8_output_variable/op_index dictionary
```
sfraczek, line 157: What is this dictionary's purpose is not clear to me.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+        '''
+        for k, v in self._int8_output_var_op_index_dict.items():
+            if v >= index:
+                self._int8_output_var_op_index_dict[k] = v + 1
+        if var_name:
+            self._int8_output_var_op_index_dict[var_name] = index
+
+    def __update_program(self):
+        '''
+        Update the program with the quantize/dequantize op insertion.
```
sfraczek, line 167: This function is also long and I have trouble understanding it. There seems to be a lot going on, maybe you could add some comments or split it into smaller functions?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+                        self._s8_output_var.append(i.name)
+                else:
+                    max_range = self.__get_max_range_by_var_name(self.program,
+                                                                 i.name)
+
+                max_value = [[np.abs(np_data)]
+                             for np_data in sampling_data[i.name]]
+
+            self._var_max_range[i.name] = max_range
+
```
sfraczek: this newline seems unnecessary

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+            if i.name not in self._weights_var_name + self._conv_input_var_name + self._conv_output_var_name + self._residual_input_var_name:
+                continue
+
+            if i.name in self._weights_var_name:
+                max_value = []
+                data = sampling_data[i.name][0]
+                for j in range(data.shape[0]):
+                    if not self._is_close(float(np.max(np.abs(data[j]))), 0.0):
+                        max_value.append(Calibrator.s8_max /
+                                         float(np.max(np.abs(data[j]))))
```
sfraczek: For:
```python
                    if not self._is_close(float(np.max(np.abs(data[j]))), 0.0):
                        max_value.append(Calibrator.s8_max /
                                         float(np.max(np.abs(data[j]))))
```
How about storing `float(np.max(np.abs(data[j])))` in a variable so its calculated once?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+    def __sampling(self, sampling_data):
+        '''
+        Sampling the variables data range.
+        '''
+        for i in self.program.list_vars():
+            if i.name not in self._weights_var_name + self._conv_input_var_name + self._conv_output_var_name + self._residual_input_var_name:
+                continue
+
+            if i.name in self._weights_var_name:
+                max_value = []
```
sfraczek: Is sampling_data[i.name] in NHWC format? Because I heard this one is required for int8. Or is during calibration the format NCHW?
What is this max value per then, per height? 
Also, maybe a more descriptive name would be helpful for understanding like max_value_per_channel.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+                        max_range = Calibrator.u8_max
+                        self._u8_output_var.append(i.name)
+                    else:
+                        max_range = Calibrator.s8_max
+                        self._s8_output_var.append(i.name)
+                else:
+                    max_range = self.__get_max_range_by_var_name(self.program,
+                                                                 i.name)
+
+                max_value = [[np.abs(np_data)]
```
sfraczek, line 365: What is max_value here? Is it abs of all the sampling_data[i.name] in form of an array?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,711 @@
...
+                    cur_op = self.program.current_block().ops[
+                        self._conv_output_var_name.index(i.name) + 2]
+                    if cur_op.has_attr('fuse_relu') and cur_op.attr(
+                            'fuse_relu'):
+                        max_range = Calibrator.u8_max
+                        self._u8_output_var.append(i.name)
+                    else:
+                        max_range = Calibrator.s8_max
+                        self._s8_output_var.append(i.name)
+                else:
```
sfraczek, line 362: Is this else case not weight, not output, so for inputs and residual_inputs? Where are the biases?

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: For the first answer:  Correct me if I'm wrong, but isn't conv_op_index[index+1] where quantization should be inserted when index == 0, instead of conv_op_index[index+1]-1, to be right before whatever was at that index? 
Usually if we want to insert something before the first element of array (i=0) we would insert at i=0 and the previous element gets pushed to i=1, so that's what I don't get here. Is it working differently here?

For the second answer: Do you always expect pooling after last convolution?

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: I see. So how about adding a comment that says this is for safety check?

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: Ok.

sfraczek (changes requested):
I've added some more comments and questions. I will try to continue tomorrow.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Sure. I'll add the comments to resolve the potential confuse.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Will update the comments per your suggestion:)

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Yes, The array length is depend on(or equal to) the iterations. We sample max_value of all iterations firstly, then calculate the scale with max_range(s8/u8) later.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: As the Scale(Bias) = Scale(Bias)* Scale(Weights),  we don't need to acquire the bias scale separately.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Yes , it's outdated logic which should be abandoned. Will remove it and optimize the code per your previous suggestion.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: This dict record all int8-conv op output variable and their index as key-value mapping.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: which line?  Do you mean ` self._var_max_range[i.name] = max_range`?

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Agree

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: > Is sampling_data[i.name] in NHWC format? Because I heard this one is required for int8. Or is during calibration the format NCHW?

sampling_data[i.name] stands for FP32 data which format is NCHW. 

> What is this max value per then, per height?

per channel.

> Also, maybe a more descriptive name would be helpful for understanding like max_value_per_channel.

We're collect all data with var_name/max_value dict at the end of this function, so i don't want to use a different name to distinguish their source explicitly.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: > For the first answer: Correct me if I'm wrong, but isn't conv_op_index[index+1] where quantization should be inserted when index == 0, instead of conv_op_index[index+1]-1, to be right before whatever was at that index?
Usually if we want to insert something before the first element of array (i=0) we would insert at i=0 and the previous element gets pushed to i=1, so that's what I don't get here. Is it working differently here?

It's a trade-off `offset` here, we adjust the offset in the following __update_program to avoid much more `-1` indexing.

> For the second answer: Do you always expect pooling after last convolution?

Of coz no.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Will update the code accordingly.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+import paddle.fluid as fluid
+
+
+class Calibrator(object):
+    '''
+    The calibrator class transforms the program and updates the calculated scale into it.
+    '''
+    non_conv_int8_op_type = ("pool2d")
+    supported_int8_op_type = ("conv2d", "pool2d")
+    not_supported_int8_op_type = ("elementwise_add", "concat")
```
wojtuss: Please, remove the unused variables.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+            if i.name not in self._weights_var_name + self._conv_input_var_name + self._conv_output_var_name + self._residual_input_var_name:
+                continue
+
+            if i.name in self._weights_var_name:
+                max_value = []
+                data = sampling_data[i.name][0]
+                for j in range(data.shape[0]):
+                    var_value = float(np.max(np.abs(data[j])))
+                    if not self._is_close(var_value, 0.0):
+                        max_value.append(Calibrator.s8_max / var_value)
```
wojtuss: The `max_value` vector seems to contain scale factors which should be used to transform original weight values into the INT8 range. Am I right?
If yes, the variable's name is misleading.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+            self._conv_output_var_name.append(self.program.current_block().ops[
+                i].output('Output')[0])
+            self._int8_output_var_op_index_dict[self.program.current_block()
+                                                .ops[i].output('Output')[0]] = i
+            if self.program.current_block().ops[i].desc.input("ResidualData"):
+                self._residual_input_var_name.append(self.program.current_block(
+                ).ops[i].desc.input("ResidualData")[0])
+
+            if self.program.current_block().ops[i + 1].type == "pool2d":
+                self._conv_output_var_name.append(self.program.current_block()
```
wojtuss: the name of the `_conv_output_var_name` variable suggests it contains only convolution output's names, but here also the pool2d operator's output name is added

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+                    var_value = float(np.max(np.abs(data[j])))
+                    if not self._is_close(var_value, 0.0):
+                        max_value.append(Calibrator.s8_max / var_value)
+                    else:
+                        max_value.append(0.0)
+                max_range = Calibrator.s8_max
+            else:
+                if i.name in self._conv_output_var_name:
+                    cur_op = self.program.current_block().ops[
+                        self._conv_output_var_name.index(i.name) + 2]
```
wojtuss: Magic number: what does the "+2" stand for? What operator should be `cur_op`? It seems it can be any operator.

wojtuss: Does the index returned by `self._conv_output_var_name.index(i.name)` match the index of conv2d or pool2d operator in the current block? It does not look so.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+                    if cur_op.has_attr('fuse_relu') and cur_op.attr(
```
wojtuss, line 355: What is the purpose of this check?

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: ~Could you explain it in more detail please?~
Ok, I see you have changed the -1 and it looks good to me. Thank you.

python/paddle/fluid/contrib/int8_inference/utility.py
wojtuss: Could you elaborate more on this? @sfraczek and I do not understand the equation.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+
+        for i in self._conv_op_index[1:]:
+            weights_var_name = self.program.current_block().ops[i].input(
+                'Filter')[0]
+            input_var_name = self.program.current_block().ops[i].input('Input')[
+                0]
+            output_var_name = self.program.current_block().ops[i].output(
+                'Output')[0]
+            self._output_program.current_block().ops[i]._set_attr(
+                "Scale_weights", self._var_max_value_map[weights_var_name])
```
wojtuss: The `_var_max_value_map` contains the scale factors, not the max_values, doesn't it?
If yes, please rename it.
The same applies below.

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: Regarding this:
>For example, conv(int8)--->pooling(int8) where conv is the last conv op, if we inserted the dequantize after the conv, we can't levarage the pooling op which already enabled the int8 supporting. So we just insert the dequantize after the pool op.

and the fact that you said that you don't always expect pooling after the last convolution, then how can I know if this should always work? How do you know that you can safely insert the dequantize operator after the operator following the last convolution?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+
+                for original_input in self._output_program.current_block().ops[
+                        start_index].input_arg_names:
+                    index_res = self.__get_op_index_by_output_var(
+                        self._output_program, original_input)
+                    if index_res != -1:
+                        original_out_var = original_input
+                        break
+
+                if original_out_var:
```
wojtuss, line 230: Can the `original_out_var` be `None` here?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+                    op._set_attr("data_format", "MKLDNNLAYOUT")
+                    op._set_attr("use_mkldnn", 1)
+                    op._set_attr("Scale", self._var_max_range[original_out_var]
+                                 / calc_max_func(self._var_max_value_map[
+                                     original_out_var]))
+
+                    for op_index in range(
+                            start_index + 1,
+                            len(self._output_program.current_block().ops)):
+                        if self._output_program.current_block(
```
wojtuss, line 245: According to line 411 (`elif index == len(self._conv_op_index) - 1:`), there can't be any convolution after the dequantize index.

python/paddle/fluid/contrib/int8_inference/utility.py
sfraczek: I thought about this again and I really suggest removal of this `if`. Maybe I don't understand something so tell me if it is important. How I see it, it will never be executed based on the definition of conv_op_index because it has all convolutions already. Also for the `else` I see that only convolution has that parameter so it would also never be executed? If you insist on keeping this maybe there should be a raise exception if conv2d was found? What do you think?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,700 @@
...
+                    inserted_op_length += 1
+
+                    op._set_attr("data_format", "MKLDNNLAYOUT")
+                    op._set_attr("use_mkldnn", 1)
+
+    def __update_output_program_attr(self):
+        for i in self._output_program.list_vars():
+            if i.name in self._persistable_vars:
+                i.persistable = False
+                os.system("rm -rf {}/{}".format(self.pretrained_model, i.name))
```
wojtuss, line 271: Does it mean the files from the original pretrained model are being deleted here? We shouldn't modify them, should we?

sfraczek:
I'm going on vacation @wojtuss will take over reviewing.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Relu means the output value is the positive and its max range is 255(u8) accordingly.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: No. We don't delete them. We just delete some variables file because their persistable  attr were set to true temporarily for sampling.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: possible.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: i don't follow u clearly. Line 411 means there's one dequantize op after the last conv op if there's one pooling op follow the last conv op too.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Sorry that's my typo. It should be Scale(Bias) = Scale(Src)* Scale(Weights)

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Ok.

python/paddle/fluid/contrib/int8_inference/utility.py
hshen14: Guoming is right.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Yes, you understanding is right.  As the scaling factor for weights doesn't need to leverage the KL algorithm, so we put the scaling factor(which equal to 127/max_value) directly. For the non-weights variable, such as the output of conv op, we just saved the max_value of them. i agree that it may cause confuse and will save the scaling factor of weights in a unique map.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: It contains max_value, i created a new array to save weights scaling factor in following change

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Yes, it's bug because it should be indexed again as a parameter of self._conv_op_index.
Anyway, i abandoned this implementation as it may raise confuse. Please see my changes later.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Created one new array to save pool2d output variable.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: The purpose of this function is to insert the quantize/dequantize op into the original program.
The logic can be divided into two parts: 

1.  insert quantize. 
2.  insert dequantize.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: i read the snippet code several times and it looks ur comment is make sense. I'll delete those useless check and just set the force_fp32_output flag to true.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Maybe u misunderstanding  my previous comment and vice versa:).
what i said that i don't expect pooling after the last convolution means that the code doesn't aware there's pooling or other non-conv-int8 op following by last conv int8 op. So there're two cases here:

1.  there's int8-enabled-op following by last conv op, such as pooling, we'll add the dequantize op after the pool2d. 

2. there's no int8-enabled-op following by last conv op, we'll set the last conv op attribute `force_fp32_output` to True without dequantize op insertion.

guomingz:
hi @wojtuss would you please continue the review for this PR? I've update the code per the previous comments.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: oh, we consider this function as a private method.

guomingz:
Hi @luotao1 I've reverted the code to the 8952ecd as the initial version and checked-in the changes as the 0cc7dab for your side review.  Besides, the continuous-integration/travis-ci/pr  is passed while the rest of test may depend on the paddle side int8 op supporting.

Thanks.

wojtuss (approved):

wojtuss (dismissed):
LGTM

python/paddle/fluid/contrib/int8_inference/utility.py
wojtuss: the empty line, no. 362

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -423,107 +424,83 @@ def __get_quantize_dequantize_combination(self, program):
...
-                            program.current_block().ops[op_index - 1]._set_attr(
-                                "force_fp32_output", True)
-                            break
-                        else:
-                            program.current_block().ops[op_index - 1]._set_attr(
-                                "force_fp32_output", True)
+            elif self._conv_op_index[index] + 1 < self._conv_op_index[index +
+                                                                      1]:
+
+                program.current_block().ops[self._conv_op_index[
```
wojtuss, line 436: Sorry, I missed that in the earlier review. Why do we force fp32 output here?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -423,107 +424,83 @@ def __get_quantize_dequantize_combination(self, program):
...
-                                op_index].input_names:
-                            if program.current_block().ops[op_index].input(
-                                    input_attr_name):
+                    if op_has_int8_input:
+                        if op_type == "conv2d":
+                            if program.current_block().ops[op_index +
+                                                           1].type == "conv2d":
+                                continue
+                            elif program.current_block(
+                            ).ops[op_index +
```
wojtuss, line 460: In the following cases
`...->conv2d->pool2d->conv2d->...`
`...->conv2d->pool2d->pool2d->conv2d->...`
(from the Detect model) the dequantize op will be inserted after the first `pool2d` op, but it should not. Am I right?

python/paddle/fluid/contrib/int8_inference/utility.py
wojtuss: IMO these are very good names for helper functions to be used here.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: It means there's a pattern that conv(int8)-->non-int8(op)--->...--->conv(int8) op. We don't insert the dequantize op but set the first conv(int8) op attribute force_fp32_output to True to get the fp32 output as the input of next non-int8 op.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: yes, you're right. We're not considered such case. According to ur example, the ->conv2d->pool2d->pool2d->conv2d->.. will be ->conv2d(force_fp32_output)->pool2d->pool2d->quantize_op-->conv2d->... . It would low the int8 performance but w/o accuracy drop.
As we don't have plan to support detect model(BTW, would u please share me the model name of this detect model) for coming PaddlePaddle 1.3 release, do u think it's necessary to fix this issue for this PR?

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Will remove if there's another `change requestes`.

python/paddle/fluid/contrib/int8_inference/utility.py
hshen14: @wojtuss We did not support the feature in this PR because:
1. It is an optional feature to support INT8 for first convolution. Industries also have the similar thoughts, making the feature separate from other INT8 convolutions.
2. First convolution is critical for accuracy, in our practice, the recommended solution is to keep first convolution FP32.  

We will consider to add such support, when real requirement comes to achieve better performance. @guomingz

guomingz:
Hi @luotao1  Poland team has finished the review. Would u please notify the Baidu folks to review this PR?
thank you in advance.

python/paddle/fluid/contrib/int8_inference/utility.py
wojtuss: @hshen14 , @guomingz  I understand the concept of leaving the first conv op as fp32, but here that's not the case. The exact sequence of ops in one of the branches in the Detect model is:
```
...->conv2d->conv2d->pool2d->conv2d->conv2d->pool2d->pool2d->conv2d->...
```
and the calibration algorithm does not handle such cases properly. If the tool is designed for particular models, please add a list of models it covers in a README or elsewhere.

python/paddle/fluid/contrib/int8_inference/utility.py
wojtuss: But the sequence
```
...conv2d->...->conv2d*->pool2d->conv2d->...
```
also falls into this case and the conv2d* op should not force fp32 output.

python/paddle/fluid/contrib/int8_inference/utility.py
hshen14: @wojtuss Thansk for your reporting. INT8 v1 targets the pre-defined POR models, including ResNet-50, MobileNet, SSD-MobileNet, and Se-ResNeXT. INT8 v2 will have the new design and will enhance the graph transformation with the pattern you mentioned.

wojtuss: OK, please add some comments about the INT8 v1 targets. At least to the PR's description.

guomingz: Ok. Will add the comments on the class Calibrator doc.
Thank you.

guomingz: Updated.

python/paddle/fluid/contrib/tests/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 file(GLOB TEST_OPS RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "test_*.py")
 string(REPLACE ".py" "" TEST_OPS "${TEST_OPS}")
 
+if(APPLE OR WIN32)
+    list(REMOVE_ITEM TEST_OPS test_calibration)
+endif()
```
luotao1, line 6: remove line4-6, before line1
```
if (APPLE OR WIN32 OR NOT WITH_MKL)
   return()
endif()
```
to pass the Paddle_PrCiPython35 which runs on OPENBLAS

guomingz: oh， i saw there's another case which name is test_quantize_transpiler.py? Shall we ignore it too?
My original code just remove the test_calibration case only. Please advise.

guomingz: > remove line4-6, before line1
> 
> ```
> if (APPLE OR WIN32 OR NOT WITH_MKL)
>    return()
> endif()
> ```
> to pass the Paddle_PrCiPython35 which runs on OPENBLAS
oh， i saw there's another case which name is test_quantize_transpiler.py? Shall we ignore it too?
My original code just remove the test_calibration case only. Please advise.

luotao1: if `test_quantize_transpiler.py` could run on OPENBLAS, please remain it.

guomingz: > if `test_quantize_transpiler.py` could run on OPENBLAS, please remain it.

OK.

guomingz:
Hi @wojtuss @sfraczek  ,
I've made some changes since your last approved previous commit, the current change is focus on the one bug that invalid is_negative_input flag set.

Would you please review my change and approve if no further change needed?

Thank you in advance.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -72,9 +74,14 @@ def __display_debug(self):
     def __get_max_range_by_var_name(self, program, var_name):
         """
         Check the specified variable was generated from Relu layer or not.
+        If the variable was the output of one of the pool2d/reshape/concat
+        /transpose, we keep trace the ancesotr of this variable;
```
wojtuss: ancesotr -> ancestor

wojtuss (dismissed):
LGTM

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: > ancesotr -> ancestor

Updated in commit f201a48

python/paddle/fluid/contrib/tests/test_calibration.py
```diff
@@ -0,0 +1,229 @@
...
+
+    mapper = functools.partial(
+        process_image, mode=mode, color_jitter=color_jitter, rotate=rotate)
+
+    return paddle.reader.xmap_readers(mapper, reader, THREAD, BUF_SIZE)
+
+
+def val(data_dir=DATA_DIR):
+    file_list = os.path.join(data_dir, 'val_list.txt')
+    return _reader_creator(file_list, 'val', shuffle=False, data_dir=data_dir)
```
luotao1, line 116: @typhoonzero line45-114 is almost duplicated with https://github.com/PaddlePaddle/Paddle/blob/develop/benchmark/fluid/imagenet_reader.py, how do you think about moving `imagenet_reader.py` to `python/paddle/fluid/utils` or `python/paddle/fluid/reader`?

python/paddle/fluid/contrib/tests/test_calibration.py
```diff
@@ -0,0 +1,229 @@
...
+        with open(file_list) as flist:
+            full_lines = [line.strip() for line in flist]
+            if shuffle:
+                np.random.shuffle(full_lines)
+
+            lines = full_lines
+
+            for line in lines:
+                img_path, label = line.split()
+                #img_path = img_path.replace("JPEG", "jpeg")
```
luotao1: remove unused lines

python/paddle/fluid/contrib/tests/test_calibration.py
```diff
@@ -0,0 +1,229 @@
...
+        os.system(cmd)
+        # resnet50 fp32 data
+        resnet50_fp32_model_url = 'http://paddle-inference-dist.bj.bcebos.com/int8/resnet50_int8_model.tar.gz'
+        resnet50_zip_name = resnet50_fp32_model_url.split('/')[-1]
+        resnet50_unzip_folder_name = 'resnet50_fp32'
+        cmd = 'rm -rf {} {} && mkdir {} && wget {} && tar xvf {} -C {}'.format(
+            resnet50_unzip_folder_name, resnet50_zip_name,
+            resnet50_unzip_folder_name, resnet50_fp32_model_url,
+            resnet50_zip_name, resnet50_unzip_folder_name)
+        os.system(cmd)
```
luotao1, line 136: It seems this UT runs 5 minutes+, and it costs the second largest elapsed time.
```
[08:26:32]	[Step 1/1] 498/535 Test #535: test_calibration ................................   Passed  336.08 sec
```

Please put the download process in cmake, and put them in `"${THIRD_PARTY_PATH}/int8"` to cache it. Thus, CI system only needs download once, to decrease the download and unzip time.

http://ci.paddlepaddle.org/viewLog.html?buildId=51697&tab=buildLog&buildTypeId=Paddle_PrCi&logTab=tree&filter=all

You can refer to 
https://github.com/PaddlePaddle/Paddle/blob/f534c66d2d7d87aa580538513f4835439acd7bc0/paddle/fluid/inference/tests/test.cmake#L1-L16

python/paddle/fluid/contrib/tests/test_calibration.py
```diff
@@ -0,0 +1,229 @@
...
+
+        self.iterations = 100
```
luotao1, line 138: Could this `iterations` be much less?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,708 @@
...
+import os
+import paddle.fluid as fluid
+
+
+class Calibrator(object):
+    '''
+    The calibrator class transforms the program and updates the calculated scale into it.
+    This is INT8 v1 calibration tool, mainly for the support of ResNet-50 and MobileNet.
+    '''
+    # TODO: Below op list will be updated once more INT8 op kernels are supported.
```
luotao1: The correct format of `TODO` is `TODO(guomingz)`, or someboby others.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,708 @@
...
+        self.__display_debug()
+
+    def __display_debug(self):
+        if self.debug:
+            self.__dot(self._output_program)
+            print(self._output_program)
+
+    def __get_max_range_by_var_name(self, program, var_name):
+        """
+        Check the specified variable was generated from Relu layer or not.
```
luotao1, line 76: Check whether the

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,708 @@
...
+        If the variable was the output of one of the pool2d/reshape/concat
+        /transpose, we keep trace the ancestor of this variable;
+        If the variable was the output the conv op, we check it's has_relu
```
luotao1, line 79: we check whether it contains has_relu attr

python/paddle/fluid/contrib/int8_inference/utility.py
luotao1: please add `TODO` here.

python/paddle/fluid/contrib/int8_inference/utility.py
luotao1: @guomingz How do you think about https://github.com/PaddlePaddle/Paddle/pull/15062/files#r244694124?

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,708 @@
...
+                return Calibrator.s8_max
+
+        return Calibrator.s8_max
+
+    def __check_op_type_with_specified_var_as_input(self,
+                                                    program,
+                                                    var_name,
+                                                    start_index=0):
+        '''
+        Check all the type of ops that use the specified variable as the input.
```
luotao1: Check whether all the type of ops that use the specified variable as the input.

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,708 @@
...
+            if var_name in op.input_arg_names
+        ]
+        for i in op_type_list:
+            if not i in Calibrator.supported_int8_op_type:
+                return False
+        return True
+
+    def __check_var_source_dt(self, var_name):
+        '''
+        Check the specified variable is the output of int8 conv op or not.
```
luotao1: Check whether the

python/paddle/fluid/contrib/int8_inference/utility.py
```diff
@@ -0,0 +1,708 @@
...
+
+    def __check_force_fp32_attr_by_output_var(self, program, var_name):
+        for op in program.current_block().ops:
+            if op.type == "conv2d" and var_name in op.output_arg_names:
+                return op.attr("force_fp32_output")
+        return False
+
+    def __get_op_index_by_output_var(self, program, var_name, start_index=0):
+        '''
+        Check the specified input variable is the output of the conv/pool2d
```
luotao1: Check whether the

python/paddle/fluid/contrib/tests/test_calibration.py
luotao1: Discussed with @typhoonzero, it's better to unify these codes. This could be done in next PR. Please add a TODO here.

luotao1:
For ci elapsed times, you could decrease it in next PR. https://github.com/PaddlePaddle/Paddle/pull/15062#discussion_r249773021 Please add a TODO here as well.
Besides, could you speed up the elapsed time of this UT after caching the download and unzip process?

guomingz:
> For ci elapsed times, you could decrease it in next PR. [#15062 (comment)](https://github.com/PaddlePaddle/Paddle/pull/15062#discussion_r249773021) Please add a TODO here as well.
> Besides, could you speed up the elapsed time of this UT after caching the download and unzip process?

Thanks @luotao1  for your comments..
 I have got one weird pre-ci error for commit f201a48. This commit just modify the comment, but the PR_CI (Paddle)  said the `test_word2vec` case was failed.
i reran the test for three times. Would u please give me some hints on below error messages?
[log link](http://ci.paddlepaddle.org/viewLog.html?buildId=51697&buildTypeId=Paddle_PrCi&tab=buildLog)
> [07:54:23] :	 [Step 1/1] Test project /paddle/build
[07:54:23] :	 [Step 1/1]         Start 512: test_image_classification
[07:54:23] :	 [Step 1/1]         Start 513: test_recognize_digits
[07:54:23] :	 [Step 1/1]         Start 516: test_label_semantic_roles
[07:54:23] :	 [Step 1/1]         Start 517: test_word2vec
[07:54:42] :	 [Step 1/1]   1/535 Test #517: test_word2vec ...................................***Failed   18.92 sec
[07:54:42] :	 [Step 1/1] sss
[07:54:42] :	 [Step 1/1] W0122 07:54:33.379575 46937 device_context.cc:262] Please NOTE: device: 0, CUDA Capability: 61, Driver API Version: 9.0, Runtime API Version: 8.0
[07:54:42] :	 [Step 1/1] W0122 07:54:33.379622 46937 device_context.cc:270] device: 0, cuDNN Version: 7.0.
[07:54:42] :	 [Step 1/1] 
[07:54:42] :	 [Step 1/1] 
[07:54:42] :	 [Step 1/1] F
[07:54:42] :	 [Step 1/1] ======================================================================
[07:54:42] :	 [Step 1/1] FAIL: test_cuda_sparse_normal (__main__.W2VTest)
[07:54:42] :	 [Step 1/1] ----------------------------------------------------------------------
[07:54:42] :	 [Step 1/1] Traceback (most recent call last):
[07:54:42] :	 [Step 1/1]   File "test_word2vec.py", line 264, in __impl__
[07:54:42] :	 [Step 1/1]     is_parallel=is_parallel)
[07:54:42] :	 [Step 1/1]   File "test_word2vec.py", line 238, in main
[07:54:42] :	 [Step 1/1]     infer(use_cuda, save_dirname)
[07:54:42] :	 [Step 1/1]   File "test_word2vec.py", line 225, in infer
[07:54:42] :	 [Step 1/1]     assert g_a == g_b
[07:54:42] :	 [Step 1/1] AssertionError

luotao1:
`test_word2vec` is fixed by #15462, please rerun it again.
@guomingz @hshen14 I see there is only resnet50 UT, for mobilenet UT, do you add in next PR?

guomingz:
> `test_word2vec` is fixed by #15462, please rerun it again.
> @guomingz @hshen14 I see there is only resnet50 UT, for mobilenet UT, do you add in next PR?

Ofcoz, i'll add another TODO into the test_calibration.py.
And i think when u mentioned the next PR, i believe it should be a PR that only focus on one feature, right? So now, we need to add at least below three additional PRs, please confirm.

1. Remove the duplicated code on test_calibration.py line 45~114
2.  Decrease the ci elapsed time by enabling the download data via cmake.
3.  Enable the mobilenet case on the UT.

python/paddle/fluid/contrib/tests/test_calibration.py
guomingz: > Could this `iterations` be much less?

Oh, we just used the 100 pictures as the test set. i think it may be the least number for accuracy testing.

python/paddle/fluid/contrib/tests/test_calibration.py
guomingz: > @typhoonzero line45-114 is almost duplicated with https://github.com/PaddlePaddle/Paddle/blob/develop/benchmark/fluid/imagenet_reader.py, how do you think about moving `imagenet_reader.py` to `python/paddle/fluid/utils` or `python/paddle/fluid/reader`?

I think `python/paddle/fluid/reader` is a better place as we may have another kind of reader in the future.

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Done in commit 6157ae9

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: Done in commit 6157ae9

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: > Check whether all the type of ops that use the specified variable as the input.

Done in commit 6157ae9

python/paddle/fluid/contrib/int8_inference/utility.py
guomingz: > The correct format of `TODO` is `TODO(guomingz)`, or someboby others.

Done in commit 6157ae9

python/paddle/fluid/contrib/tests/test_calibration.py
guomingz: > remove unused lines

Done in commit 6157ae9

python/paddle/fluid/contrib/tests/test_calibration.py
guomingz: > It seems this UT runs 5 minutes+, and it costs the second largest elapsed time.
> 
> ```
> [08:26:32]	[Step 1/1] 498/535 Test #535: test_calibration ................................   Passed  336.08 sec
> ```
> Please put the download process in cmake, and put them in `"${THIRD_PARTY_PATH}/int8"` to cache it. Thus, CI system only needs download once, to decrease the download and unzip time.
> 
> http://ci.paddlepaddle.org/viewLog.html?buildId=51697&tab=buildLog&buildTypeId=Paddle_PrCi&logTab=tree&filter=all
> 
> You can refer to
> [Paddle/paddle/fluid/inference/tests/test.cmake](https://github.com/PaddlePaddle/Paddle/blob/f534c66d2d7d87aa580538513f4835439acd7bc0/paddle/fluid/inference/tests/test.cmake#L1-L16)
> 
> Lines 1 to 16 in [f534c66](/PaddlePaddle/Paddle/commit/f534c66d2d7d87aa580538513f4835439acd7bc0)
> 
>  set(INFERENCE_URL "http://paddle-inference-dist.cdn.bcebos.com" CACHE STRING "inference download url") 
>  set(INFERENCE_DEMO_INSTALL_DIR "${THIRD_PARTY_PATH}/inference_demo" CACHE STRING 
>      "A path setting inference demo download directories.") 
>  function (inference_download install_dir url filename) 
>      message(STATUS "Download inference test stuff from ${url}/${filename}") 
>      file(DOWNLOAD "${url}/${filename}" "${install_dir}/${filename}") 
>      message(STATUS "finish downloading ${filename}") 
>  endfunction() 
>   
>  function (inference_download_and_uncompress install_dir url filename) 
>      inference_download(${install_dir} ${url} ${filename}) 
>      execute_process( 
>              COMMAND ${CMAKE_COMMAND} -E tar xzf ${install_dir}/${filename} 
>              WORKING_DIRECTORY ${install_dir} 
>      ) 
>  endfunction()

Add the TODO label in the code and will update the code in another PR.

luotao1 (approved):
LGTM
> we need to add at least below three additional PRs

Yes, you could add three independent PRs.

Besides, 
- how much time of this UT except for downloading and unzipping?
- please add a readme to tell uses how to use the calibration tool.

luotao1:
@sfraczek Do you have any comments for this PR, Could you give a "Approve" for it?

guomingz:
> LGTM
> 
> > we need to add at least below three additional PRs
> 
> Yes, you could add three independent PRs.
> 
> Besides,
> 
> * how much time of this UT except for downloading and unzipping?
> * please add a readme to tell uses how to use the calibration tool.

Thanks,. i'll update the elapsed time data later as i need to measure it.
For readme. i'll add another markdown doc under int8_inference directory  per your previous [suggestion](https://github.com/PaddlePaddle/Paddle/issues/15063). And i think it  will be a new PR once 15602 merged.

guomingz:
> * how much time of this UT except for downloading and unzipping?
Hi @luotao1  i measured the UT time w/o downloading and unzipping  and the elapsed time are 46s/41s/42s( measured three times on our local master), it maybe not a precise value but a reference data.

guomingz:
> @sfraczek Do you have any comments for this PR, Could you give a "Approve" for it?

> I'm going on vacation @wojtuss will take over reviewing.

Hi @luotao1 ， @sfraczek already reviewed the first half of the code while @wojtuss reviewed the left part of the code due to @sfraczek took vacation.  @wojtuss has approved the code changes two day ago.

luotao1:
> i measured the UT time w/o downloading and unzipping and the elapsed time are 46s/41s/42s

It seems the elapsed time could speed up 3X times in our CI. We can observe the UT time after caching the downloading and unzipping at first.

wojtuss (approved):
LGTM, @luotao1 I have resolved @sfraczek 's conversations.

sfraczek (approved):

PR #15073, opened 2018-12-26
Add scope_pool

sneaxiy:


paddle/fluid/pybind/pybind.cc
```diff
@@ -474,17 +478,26 @@ All parameter, weight, gradient are variables in Paddle.
           param.set(param_array, place)
 
         )DOC")
+      .def("_remove_from_pool",
```
panyx0718, line 481: seems not used (and Remove()) currently

panyx0718 (approved):

PR #15075, opened 2018-12-26
Refactor distributed RPC

typhoonzero:
- [x] clean directory structure and cmake
- [ ] barrier get request when var is not ready
- [ ] clean rpc_server, rpc_client, request_handler bases
- [ ] keep only two rpc api type, send/recv and use request inner field "request type" to dispatch reqeusts


panyx0718 (approved):
LG proto change

gongweibao (approved):
LGTM

PR #15080, opened 2018-12-26
Optimize assign

jacquesqiao:


python/paddle/fluid/tests/unittests/test_layers.py
```diff
@@ -1015,6 +1015,18 @@ def test_batch_norm(self):
...
+        import numpy as np
+        startup = Program()
+        main = Program()
+        with program_guard(main, startup):
+            probs = layers.assign(
+                input=np.random.random([1, 2]).astype('float32'),
+                init_once=True)
+
+        print(str(main))
+        print(str(startup))
```
panyx0718: this is a bad test...

python/paddle/fluid/layers/nn.py
```diff
@@ -5010,10 +5010,12 @@ def nce(input,
             alias_probs_[little[0]] = 1.0
             alias_[little[0]] = -1
 
-        probs = assign(input=np.array(custom_dist).astype('float32'))
-        custom_alias = assign(input=np.array(alias_).astype('int32'))
+        probs = assign(
+            input=np.array(custom_dist).astype('float32'), init_once=True)
```
panyx0718: why not create var in startup program here?

typhoonzero: agree to create var here.

python/paddle/fluid/tests/unittests/test_layers.py
jacquesqiao: real test is here https://github.com/PaddlePaddle/Paddle/pull/15080/files#diff-68cdbd1cba20d9e8a5ddfcdd1ffa88f6L35. It's disabled now, i enabled it and use it for the test.

python/paddle/fluid/layers/nn.py
jacquesqiao: persistable var must be created both in startup program and main program for type and shape inference.

typhoonzero: `assign` should be doing what the name is. or you can add a layer named `layers.static_tensor` but not an attirbute on `assign`

jacquesqiao: fix done

python/paddle/fluid/initializer.py
```diff
@@ -683,6 +684,64 @@ def __call__(self, var, block):
...
+        if dtype == VarDesc.VarType.FP32:
+            value_name = "fp32_values"
+            values = [float(v) for v in self._value.flat]
+        elif dtype == VarDesc.VarType.INT32:
+            value_name = "int32_values"
+            values = [int(v) for v in self._value.flat]
+        else:
+            raise ValueError("Unsupported dtype %s", self._value.dtype)
+        if self._value.size > 1024 * 1024 * 5:
+            raise ValueError("The size of input is too big. Please consider "
```
typhoonzero, line 730: warn the max size

typhoonzero (approved):

panyx0718 (approved):
LG API

PR #15089, opened 2018-12-27
try unify Executor and ParallelExecutor

panyx0718:
test=develop

### DIFF
--- a/python/paddle/fluid/executor.py
+++ b/python/paddle/fluid/executor.py
@@ -14,11 +14,15 @@
 
 from __future__ import print_function
 
+import os
+import multiprocessing
 import numpy as np
 import contextlib
 import six
 from .framework import Program, default_main_program, Variable
 from . import core
+from . import compiler
+from .. import compat as cpt
 
 __all__ = ['Executor', 'global_scope', 'scope_guard']
 
@@ -275,11 +279,8 @@ class Executor(object):
 
     def __init__(self, place):
         self.place = place
-        p = core.Place()
-        p.set_place(place)
-        self.executor = core.Executor(p)
-
         self.program_caches = dict()
+        self.executor = None
         self._closed = False
 
     def _get_program_cache(self, program_cache_key):
@@ -361,6 +362,7 @@ class Executor(object):
         You can no long use this executor after calling this method.
         For the distributed training, this method would free the resource on PServers related to
         the current Trainer.
+        TODO(panyx0718): Why ParallelExecutor doesn't have close?
 
         Example:
             >>> cpu = core.CPUPlace()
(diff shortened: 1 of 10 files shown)


python/paddle/fluid/executor.py
```diff
@@ -253,6 +258,95 @@ def _as_lodtensor(data, place):
...
+                self._exec_strategy.num_threads = len(self._places) * 4
+            else:
+                cpu_num = int(
+                    os.environ.get('CPU_NUM', multiprocessing.cpu_count()))
+                self._exec_strategy.num_threads = cpu_num * 2
+
+        trainers_endpoints = self._program._trainers_endpoints
+        if self._build_strategy.num_trainers > 1 and trainers_endpoints:
+            assert self._build_strategy.num_trainers == len(
+                trainers_endpoints), "num_trainers == len(end_points)"
```
hshen14: The code here indicates the num_trainers should be equal to trainers_endpoints; otherwise assertion will fail. So, to make the condition consistent, should change to "self._build_strategy.num_trainers > 0" since it should align the semantics on whether a list is non empty.

panyx0718: @typhoonzero WDYT?

typhoonzero: Well, when `self._build_strategy.num_trainers` is equal to 1, which means local training, we should not check `trainers_endpoints` though. by the way, `trainers_endpoints` is for sparse GPU gather only.

hshen14: Got it. Probably adding some comments to explain the assumptions.

python/paddle/fluid/executor.py
```diff
@@ -372,6 +467,76 @@ def close(self):
...
+        assert isinstance(compiled_program, _ProgramCompiler)
+
+        if compiled_program._is_data_parallel:
+            return self._run_parallel(compiled_program._executor,
+                                      compiled_program._scope, feed, fetch_list,
+                                      return_numpy)
+        else:
+            return self.run(compiled_program._program,
+                            feed,
+                            fetch_list,
```
chengduoZH: Note that: the fetch_list's element type of the `Executor` and `ParallelExecutor` are different,  `Executor`'s fetch_list is a list of variables or variables' name, but `ParallelExecutor`'s fetch_list only can be a list of variables' name.

python/paddle/fluid/tests/unittests/parallel_executor_test_base.py
```diff
@@ -87,29 +80,28 @@ def run_executor(exe, feed, fetch_list, program=None):
...
-                    use_cuda,
+                binary = exe._start_compile(main)._with_data_parallel(
                     loss_name=loss.name,
-                    exec_strategy=exec_strategy,
-                    build_strategy=build_strategy)
+                    build_strategy=build_strategy,
+                    exec_strategy=exec_strategy)._compile()
             else:
-                exe = fluid.Executor(place=place)
+                binary = exe._start_compile(main)._compile()
```
chengduoZH: Not sure whether it is a good idea, because, for user, before running, they must compile the program explicitly, and it may add to the user's burden.

python/paddle/fluid/executor.py
panyx0718: Can ParallelExecutor support variables? it not friendly to users to be different

python/paddle/fluid/tests/unittests/parallel_executor_test_base.py
panyx0718: They don't "have to" compile. Compile is optional

python/paddle/fluid/executor.py
```diff
@@ -428,6 +502,50 @@ def run(self,
...
+                return_numpy=return_numpy,
+                use_program_cache=use_program_cache)
+
+        program._compile(scope, self.place)
+        self.executor = program._executor
+        if program._is_data_parallel:
+            return self._run_parallel(
+                scope=scope,
+                feed=feed,
+                fetch_list=fetch_list,
```
chengduoZH, line 530: If the `fetch_list ` is a list of variable, we should convert it to a list of variable's name here.

python/paddle/fluid/executor.py
chengduoZH: ParallelExecutor can support variables, if the fetch_list is a list of variable, just convert it to a list of variable's name.

chengduoZH: ParallelExecutor can support variables, if the fetch_list is a list of variable, just convert it to a list of variable's name.

python/paddle/fluid/parallel_executor.py
```diff
@@ -167,9 +167,8 @@ def place_obj(place):
         # step7: init ParallelExecutor
         self.executor = core.ParallelExecutor(
             places, persistable_vars, main.desc,
-            cpt.to_text(loss_name)
-            if loss_name else six.u(''), scope, local_scopes, exec_strategy,
-            build_strategy, num_trainers, trainer_id)
+            cpt.to_text(loss_name) if loss_name else six.u(''), scope,
+            local_scopes, exec_strategy, build_strategy)
```
typhoonzero, line 171: Maybe add deprecation warnings when user passed `trainer_id` and `num_trainers` to parallelexecutor API

python/paddle/fluid/executor.py
panyx0718: Done

python/paddle/fluid/parallel_executor.py
panyx0718: ParallelExecutor API will be deprecated after a few PR if things go well, so no need to add warning for these 2 fields now.

chengduoZH: You should add warning in `ParallelExecutor` to notify user that `ParallelExecutor` will be deprecated, and give the user an simple example that how to use the new API.

panyx0718: It will happen after a few PRs

python/paddle/fluid/parallel_executor.py
```diff
@@ -279,7 +279,7 @@ def run(self, fetch_list, feed=None, feed_dict=None, return_numpy=True):
                 res.append(res_dict)
             self.executor.feed_tensors_into_local_scopes(res)
 
-        fetch_var_name = '@FETCHED_VAR_NAME@'
+        fetch_var_name = 'fetch'
```
chengduoZH, line 282: I think `fetch_var_name` should be special flag and it doesn't conflict with the var's name of the user's program.
So we should check whether the program have a var that the name is exactly `fetch`.

panyx0718: we make them consistent with Executor first. Then we can consider special name

chengduoZH (approved):
LGTM

shanyi15:
Please note that **any fix on documentation** must be checked by PM before merge, thanks!!!!!!!!!

PR #15096, opened 2018-12-28
Simplify multi_box_head API in detection.py

qingqing01:
Enable to use fusion_transpose_flatten_concat_op in inference.

### DIFF
--- a/python/paddle/fluid/layers/detection.py
+++ b/python/paddle/fluid/layers/detection.py
@@ -1358,10 +1358,6 @@ def multi_box_head(inputs,
             clip=True)
     """
 
-    def _reshape_with_axis_(input, axis=1):
-        out = nn.flatten(x=input, axis=axis)
-        return out
-
     def _is_list_or_tuple_(data):
         return (isinstance(data, list) or isinstance(data, tuple))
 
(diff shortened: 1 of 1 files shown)


wanghaoshuang (dismissed):

python/paddle/fluid/layers/detection.py
```diff
@@ -1487,14 +1468,23 @@ def _is_list_or_tuple_and_equal(data, length, err_info):
...
-            reshaped_boxes.append(_reshape_with_axis_(box_results[i], axis=3))
-            reshaped_vars.append(_reshape_with_axis_(var_results[i], axis=3))
+            reshaped_boxes.append(_flatten_(box_results[i]))
+            reshaped_vars.append(_flatten_(var_results[i]))
 
         box = tensor.concat(reshaped_boxes)
         var = tensor.concat(reshaped_vars)
         mbox_locs_concat = tensor.concat(mbox_locs, axis=1)
+        mbox_locs_concat = nn.reshape(
+            mbox_locs_concat, shape=[0L, -1L, 4L], inplace=True)
```
reyoung: It seems that `0L` is not valid syntax for Python 3+. Just use `[0, -1, 4]` is ok.

qingqing01: Done. Thanks!

NHZlX (approved):
LGTM

PR #15097, opened 2018-12-28
adaptive_pool support pool_size as int. test=develop

heavengate:
support adaptive_pool2d/3d API, pool_size can be int type, such as pool_size=3 in adaptive_pool2d, pool_size will be convert to [3, 3] automatically


qingqing01 (approved):
LGTM.

PR #15087, opened 2018-12-27
test=develop

tianshuo78520a:
Previous submissions seem to be problematic, forcing me to submit revisions again.

Add API document audit by Shanyi15

PR #15010, opened 2018-12-24
Tools

tianshuo78520a:
test=develop
New features have been added
If you want to modify the API documentation, you need to find Shanyi15 audit

PR #15104, opened 2018-12-28
update CI rules for checking change of python reference

tianshuo78520a:
Add API document audit by Shanyi15
previous version: #15010 #15087


reyoung (approved):
LG except for the hard code files in python. Maybe we can find a better way later.

PR #15111, opened 2018-12-28
Fix adam tmp var on cpu

jacquesqiao:


panyx0718 (commented):
seems other ops have this program as well. GPU also need to be fixed

jacquesqiao:
@panyx0718 will have another pr for all other op and GPU kernel. This pr is need by a model used now.

panyx0718 (approved):
Others should be fixed.

sneaxiy (approved):
LGTM

PR #15139, opened 2019-01-02
Remove op handle lock

sneaxiy:


chengduoZH (approved):
LGTM

paddle/fluid/platform/cuda_helper.h
```diff
@@ -0,0 +1,58 @@
...
+#include "paddle/fluid/platform/macros.h"
+
+#if CUDA_VERSION < 9000
+enum cublasMath_t { CUBLAS_DEFAULT_MATH = 0 };
+#endif
+
+namespace paddle {
+namespace platform {
+
+class CublasHandleHolder {
```
panyx0718, line 29: do we need similar thing for cudnn?

paddle/fluid/platform/device_context_test.cu
```diff
@@ -43,9 +43,6 @@ TEST(Device, CUDADeviceContext) {
     ASSERT_NE(nullptr, gpu_device);
     cudnnHandle_t cudnn_handle = device_context->cudnn_handle();
     ASSERT_NE(nullptr, cudnn_handle);
-    cublasHandle_t cublas_handle = device_context->cublas_handle();
```
panyx0718, line 46: still need some cublas test?

paddle/fluid/platform/device_context.cc
```diff
@@ -245,8 +245,15 @@ CUDADeviceContext::CUDADeviceContext(CUDAPlace place)
   eigen_stream_.reset(new EigenCudaStreamDevice());
   eigen_stream_->Reinitialize(&stream_, place);
   eigen_device_.reset(new Eigen::GpuDevice(eigen_stream_.get()));
-  PADDLE_ENFORCE(dynload::cublasCreate(&cublas_handle_));
-  PADDLE_ENFORCE(dynload::cublasSetStream(cublas_handle_, stream_));
+  cublas_handle_.reset(new CublasHandleHolder(stream_, CUBLAS_DEFAULT_MATH));
+
+  if (TensorCoreAvailable()) {
+#if CUDA_VERSION >= 9000
```
panyx0718, line 251: this seems unnecessary since TensorCoreAvailable already check it.

paddle/fluid/platform/cuda_helper.h
sneaxiy: Not quite sure whether cudnnHandle_t raise the same hang. No the same phenomenon occurs yet. Maybe we should add guard to cudnnHandle_t when errors are found.

paddle/fluid/platform/device_context_test.cu
sneaxiy: cublasHandle_t would be tested in all cublas unittests. I just remove `cublas_handle()` to prevent users from calling cublas functions without lock guard.

paddle/fluid/platform/device_context.cc
sneaxiy: Cannot remove this macro guard, since `CUBLAS_TENSOR_OP_MATH` is not defined in CUDA < 9000.

panyx0718: I know, but TensorCoreAvailable() already checks it.

panyx0718 (approved):

paddle/fluid/platform/device_context.cc
sneaxiy: No. `CUBLAS_TENSOR_OP_MATH` is not defined in CUDA < 9000. Removing the macro guard would cause compilation error in CUDA < 9000.

PR #15156, opened 2019-01-03
fix gpu buils issue on windows test=develop

wopeizl:


cmake/generic.cmake
```diff
@@ -448,6 +448,9 @@ function(nv_test TARGET_NAME)
     cmake_parse_arguments(nv_test "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
     cuda_add_executable(${TARGET_NAME} ${nv_test_SRCS})
     target_link_libraries(${TARGET_NAME} ${nv_test_DEPS} paddle_gtest_main lod_tensor memory gtest gflags glog)
+    if(WIN32)
+      target_link_libraries(${TARGET_NAME} shlwapi)
```
dzhwinter: Please move to glog、gtest library cmake file.

dzhwinter (dismissed):

paddle/fluid/platform/float16.h
```diff
@@ -59,7 +59,7 @@ limitations under the License. */
 #if !defined(_WIN32)
 #define PADDLE_ALIGN(x) __attribute__((aligned(x)))
 #else
-#define PADDLE_ALIGN(x) /*do nothing*/
+#define PADDLE_ALIGN(x) __declspec(align(x))
```
dzhwinter, line 62: nice catch

dzhwinter (approved):
LGTM

dzhwinter:
By the way, I thought that we can just archive the shawpi.lib into the gtest.lib and glog.lib, then other place will work fine.

dzhwinter:
By the way, I thought that we can just archive the shawpi.lib into the gtest.lib and glog.lib, then other place will work fine.

PR #15164, opened 2019-01-04
polish cudnn related code and fix bug.

dzhwinter:
mainly change
- increase drop scope counter in parallel executor
- async drop kids scope in executor(used in while and condition_block)
- add cudnn activation, for speed up fc activation. 
resnet50 per-minibatch **0.24s -> 0.14s.** (mainly reason is 1)


dzhwinter:
It seems that this can take effect on models heavily. Here is CE benchmark result.
http://yq01-paddle-online.epc.baidu.com:8098/commit/details?commit=1c15f00925727f7655e1b857ff3dd7b203f57b1c

resnet50 speed up 1.08x on single card, transformer speed up 0.23x on single card, almost every model have a 10% speed up. Further more, it seems also save some memory, less repeated allocate recycle happens, less memory fragmentation occurs. 

I am shocked that the resnet and vgg model speed up with a 100% percentage.

paddle/fluid/API.spec
```diff
@@ -64,7 +64,7 @@ paddle.fluid.initializer.BilinearInitializer.__init__ ArgSpec(args=['self'], var
 paddle.fluid.initializer.MSRAInitializer.__init__ ArgSpec(args=['self', 'uniform', 'fan_in', 'seed'], varargs=None, keywords=None, defaults=(True, None, 0))
 paddle.fluid.initializer.force_init_on_cpu ArgSpec(args=[], varargs=None, keywords=None, defaults=None)
 paddle.fluid.initializer.init_on_cpu ArgSpec(args=[], varargs='args', keywords='kwds', defaults=None)
-paddle.fluid.layers.fc ArgSpec(args=['input', 'size', 'num_flatten_dims', 'param_attr', 'bias_attr', 'act', 'is_test', 'name'], varargs=None, keywords=None, defaults=(1, None, None, None, False, None))
+paddle.fluid.layers.fc ArgSpec(args=['input', 'size', 'num_flatten_dims', 'param_attr', 'bias_attr', 'act', 'is_test', 'name', 'use_cudnn'], varargs=None, keywords=None, defaults=(1, None, None, None, False, None, False))
```
panyx0718: Don't expose this option? Use the fastest one by default?

paddle/fluid/framework/details/execution_strategy.h
```diff
@@ -25,7 +25,7 @@ struct ExecutionStrategy {
   size_t num_threads_{0};
   bool use_cuda_{true};
   bool allow_op_delay_{false};
-  size_t num_iteration_per_drop_scope_{1};
+  size_t num_iteration_per_drop_scope_{100};
```
panyx0718: what's the difference after changing this?

paddle/fluid/framework/executor.cc
```diff
@@ -466,7 +467,7 @@ void Executor::RunPreparedContext(ExecutorPrepareContext* ctx, Scope* scope,
       // the sub scopes it created should not be dropped immediately, because
       // while_grad_op will use some variables created during while_op run, so
       // we need to keep the kids and wait for the outer executor to drop them.
-      scope->DropKids();
+      Async([scope] { scope->DropKids(); });
```
panyx0718: what if the thread is not started before the next Run starts?

paddle/fluid/framework/details/execution_strategy.h
dzhwinter: It will not delete the scope after each iteration.

paddle/fluid/API.spec
dzhwinter: our other API expose this attribution to user, may be we should unify the API consistency to user.  Both way is fine to me

paddle/fluid/framework/executor.cc
dzhwinter: I think it's fine. Because each time the executor will create a brand new kid scope. 
https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/executor.cc#L420

paddle/fluid/API.spec
panyx0718: we should hide it from other API as well

panyx0718: remove this?

panyx0718 (commented):
Is this PR still being worked on?

paddle/fluid/API.spec
dzhwinter: done.

JiabinYang (approved):
LGTM

PR #15116, opened 2018-12-29
Upgrade MKL-DNN commit 

hshen14:
The commit includes the optimizations for:
* INT8 depthwise convolution on CLX
* INT8 1x1 convolution on CLX
* FP32/INT8 gemm convolution (group convolution)

test=develop

PR #15175, opened 2019-01-04
Update ngraph to resolve issue with mkldnn upgrade

baojun-nervana:
This is resolve the Relu primitive issue raised from the mkldnn upgrade https://github.com/PaddlePaddle/Paddle/pull/15116


luotao1 (approved):
LGTM

PR #15184, opened 2019-01-07
Add temp variable pool to Scope and Remove NewScope in OperatorWithKernel::RunImpl

sneaxiy:


paddle/fluid/framework/operator.cc
```diff
@@ -1018,15 +1007,29 @@ Scope* OperatorWithKernel::PrepareData(
...
+        if (!new_scope) {
+          new_scope = &scope.NewScope();
+        }
+
+        trans_var = new_scope->Var(var_name);
+      } else {
+        trans_var = scope.TempVar();
+      }
+
+      if (!has_get_out_vars) {
```
panyx0718, line 1023: no need for this bool? Just test if out_var_name_set is empty?

paddle/fluid/framework/scope.cc
```diff
@@ -87,13 +87,26 @@ Variable* Scope::Var(const std::string& name) {
 }
 
 Variable* Scope::Var(std::string* name) {
-  SCOPE_VARS_WRITER_LOCK
-  auto new_name = std::to_string(reinterpret_cast<uintptr_t>(this)) + "." +
-                  std::to_string(vars_.size());
-  if (name != nullptr) {
-    *name = new_name;
+  if (name == nullptr) {
+    return TempVar();
```
panyx0718, line 91: This seems quite implicit. User should call TempVar if they need TempVar?

paddle/fluid/framework/scope.cc
```diff
@@ -87,13 +87,26 @@ Variable* Scope::Var(const std::string& name) {
...
+  } else {
+    *name = std::to_string(reinterpret_cast<uintptr_t>(this)) + "." +
+            std::to_string(counter_.fetch_add(1));
```
panyx0718, line 94: My feeling is that this change is not worth it unless there is some experiment to prove it...creating a string in CPU might just be faster than cross-core fetch_add. who knows.

paddle/fluid/framework/scope.cc
```diff
@@ -168,9 +176,12 @@ void Scope::Rename(const std::string& origin_name,
 }
 
 std::string Scope::Rename(const std::string& origin_name) const {
-  SCOPE_VARS_WRITER_LOCK
-  auto new_name = string::Sprintf("%p.%d", this, vars_.size());
-  RenameInternal(origin_name, new_name);
+  auto new_name = std::to_string(reinterpret_cast<uintptr_t>(this)) + "." +
+                  std::to_string(counter_.fetch_add(1));
```
panyx0718, line 180: same as above

paddle/fluid/framework/temp_variable_pool.h
```diff
@@ -0,0 +1,41 @@
...
+#include <deque>
+#include <memory>
+#include <mutex>  // NOLINT
+#include "paddle/fluid/framework/variable.h"
+#include "paddle/fluid/platform/macros.h"
+
+namespace paddle {
+namespace framework {
+
+class TempVariablePool {
```
panyx0718, line 26: this class is so simple that there seems no need for it. Just implement it in Scope?

paddle/fluid/framework/scope.cc
panyx0718: Also, you can update users Var call to TempVar if they really need TempVar, not Var

panyx0718 (commented):
There seems not much difference in terms of performance, except that you are creating vars in existing Scope instead of creating new Scope. What's your plan for deleting temp vars? It seems there is no way to delete it unless the Scope is deleted? In the past, I can delete these vars by deleting sub-scope, now I need to delete the parent scope? Should temp vars live in GC and cleared in Wait (Need to be considered with GC and TempAllocator and be consistent?)

paddle/fluid/framework/operator.cc
sneaxiy: Sometimes output is empty. Add this bool variable to skip this corner case.

paddle/fluid/framework/scope.cc
sneaxiy: This is a bug fix, not for performance.

Consider the following code:
```
framework::Scope scope;
scope.Var("var1");
auto new_var1 = scope.Var();  // name of new_var1 would be address+"."+1
scope.EraseVars(std::vector<std::string>("var1"));

// name of new_var2 would be address+"."+1
// which would be the same as new_var1
auto new_var2 = scope.Var();
```

The codes above would be a bug.

paddle/fluid/framework/scope.cc
sneaxiy: Bug fix, same as above.

paddle/fluid/framework/scope.cc
sneaxiy: Yes. I would change this PR to use temporary allocator. For example, collect the newly created variables into a vector, and release them when op ends. Therefore, the newly allocated gpu memory would come back to temporary allocator.

The temporary variable would be deleted when scope is deleted in this PR. But when using executor and parallel_executor, the scope passed to `OperatorBase::Run` is not the root scope but a sub-scope, so the temp vars would be deleted when `Executor::Run` ends. But I think you are right. Using temporary allocator would be better.

paddle/fluid/framework/temp_variable_pool.h
sneaxiy: Of course.

paddle/fluid/framework/scope.cc
panyx0718: I see

PR #15185, opened 2019-01-07
enhance cudnn symbol check

Superjomn:
test=develop

### DIFF
--- a/paddle/fluid/platform/dynload/cudnn.cc
+++ b/paddle/fluid/platform/dynload/cudnn.cc
@@ -53,15 +53,15 @@ bool HasCUDNN() {
   return cudnn_dso_handle != nullptr;
 }
 
-void EnforceCUDNNLoaded(const char* fn_name) {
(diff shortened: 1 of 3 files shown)


paddle/fluid/platform/dynload/cudnn.cc
```diff
@@ -57,6 +57,11 @@ void EnforceCUDNNLoaded(const char* fn_name) {
   PADDLE_ENFORCE(cudnn_dso_handle != nullptr,
                  "Cannot load cudnn shared library. Cannot invoke method %s",
                  fn_name);
+  // if the function fn_name can't be found in the so, dlsym returns NULL.
+  PADDLE_ENFORCE(dlsym(cudnn_dso_handle, fn_name),
```
qingqing01: This will be caused call dlsym twice.  The second is at https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/platform/dynload/cudnn.h#L43

PR #15186, opened 2019-01-07
Remove workspace_handle in conv_cudnn

chengduoZH:
This PR includes two works:
 - Replace `cudnn_workspace_handle` with `TemporaryAllocator` in conv_cudnn_op.
 - Make `TemporaryAllocator` support reusing the temporary allocation which is waiting to be deleted. If one temporary allocation's size of `tmp_mem_map` equals to the size of the requirement, the temporary allocation will be removed from `tmp_mem_map`, and it can be used by another operation. It is useful for conv_cudnn_op, because if the cudnn_workspace of `conv_cudnn_op` is 4G, and the computation graph has many conv_cudnn_op which are in the same pass, we only need allocate one 4G for those conv_cudnn_op.


paddle/fluid/platform/device_context.cc
```diff
@@ -92,25 +92,28 @@ platform::TemporaryAllocator& DeviceTemporaryAllocator::Get(
...
-    std::unique_lock<std::mutex> lock(mtx_);
-    if (!device_allocator_.count(place_stream)) {
-      device_allocator_[place_stream].reset(new TemporaryAllocator(place));
-      device_allocator_[place_stream]->SetCallback([stream]() {
-        PADDLE_ENFORCE(cudaStreamSynchronize(stream));
-        PADDLE_ENFORCE(cudaGetLastError());
-      });
-    }
+  std::unique_lock<std::mutex> lock(mtx_);
+  if (!device_allocator_.count(place_stream)) {
```
sneaxiy: Why not just use
```C++
std::unique_lock<std::mutex> lock(mtx_);
auto it = device_allocator_.find(place_stream);
if (it != device_allocator_.end()) {
   return it->second;
} else {
   // Create new temp allocator as what you want.
}
```

paddle/fluid/platform/temporary_allocator.cc
```diff
@@ -84,9 +92,27 @@ void TemporaryAllocator::SetCallback(const std::function<void()> &callback) {
 
 alloc::Allocation *TemporaryAllocator::AllocateImpl(
     size_t size, alloc::Allocator::Attr attr) {
+  {
+    // Find available allocation in temp_mem_map.
+    std::unique_lock<std::mutex> lock(mtx_);
+    if (temp_mem_map_->size()) {
+      auto it = temp_mem_map_->equal_range(size);
```
sneaxiy: Why use `equal_range` instead of `lower_bound`?

paddle/fluid/operators/conv_cudnn_op.cu.cc
```diff
@@ -158,6 +157,16 @@ class CUDNNConvOpKernel : public framework::OpKernel<T> {
       VLOG(5) << "NOT use cudnn_tensor_op_math";
     }
 #endif
+    Tensor cudnn_workspace;
+    void* cudnn_workspace_ptr = nullptr;
+    if (exhaustive_search && (!half_float)) {
```
sneaxiy: How about move this code to Line 179? Maybe codes would be more readable if moved.

paddle/fluid/platform/temporary_allocator.h
```diff
@@ -65,8 +66,8 @@ class TemporaryAllocator : public memory::allocation::Allocator {
 
   // When the allocation is not held by any variable, it should be placed
   // to temp_mem_queue immediately.
-  std::shared_ptr<std::deque<TemporaryAllocation *>> temp_mem_queue_{nullptr};
-
+  std::shared_ptr<std::multimap<size_t, TemporaryAllocation *>> temp_mem_map_{
```
sneaxiy: Seems that `temp_mem_map_` can be a unique_ptr. Anyway, it is not important.

paddle/fluid/operators/conv_cudnn_op.cu.cc
chengduoZH: Done

paddle/fluid/platform/device_context.cc
chengduoZH: Done

paddle/fluid/platform/device_context.cc
```diff
@@ -92,25 +92,30 @@ platform::TemporaryAllocator& DeviceTemporaryAllocator::Get(
...
-  return *device_allocator_.at(place_stream);
 }
 
 template <>
 platform::TemporaryAllocator& DeviceTemporaryAllocator::Get(
     const platform::CUDADeviceContext& dev_ctx) {
-  auto place_stream = std::make_pair(dev_ctx.GetPlace(), dev_ctx.stream());
-  if (device_allocator_.count(place_stream)) {
-    return *device_allocator_.at(place_stream);
+  {
```
sneaxiy: Why not just call function in Line 91? This function may be inefficient if `dev_ctx` does not exist inside `device_allocator_`.

paddle/fluid/platform/temporary_allocator.cc
```diff
@@ -29,53 +31,61 @@ TemporaryAllocation::TemporaryAllocation(
...
+    temp_mem_map_.reset(new std::multimap<size_t, TemporaryAllocation *>());
     wait_delete_mem_ = 0;
   }
+
   for (auto tmp : *t_allocations) {
-    VLOG(10) << "Delete temporary allocation " << tmp->ptr()
-             << " size: " << tmp->size();
-    delete tmp;
+    delete tmp.second;
+    VLOG(10) << "Delete temporary allocation " << tmp.second->ptr()
```
sneaxiy, line 55: Why reorder these two lines? If `tmp.second` has been deleted, the following logging would cause segfault.

chengduoZH: Done

paddle/fluid/platform/temporary_allocator.cc
```diff
@@ -15,8 +15,10 @@
 #include "paddle/fluid/platform/temporary_allocator.h"
 #include "paddle/fluid/memory/allocation/allocator_facade.h"
 
-DEFINE_double(limit_of_temporary_allocation, -1,
-              "The up limit of temporary_allocation size.");
+DEFINE_int64(limit_of_tmp_allocation, -1,
+             "The up limit of temporary_allocation size.");
+DEFINE_double(reuse_tmp_allocation_excess_fraction, 2,
```
panyx0718: why fraction is larger than 1.0, why double is 2?

paddle/fluid/platform/temporary_allocator.cc
```diff
@@ -15,8 +15,10 @@
...
+              "The excess fraction of the reused tmp_allocation.");
```
panyx0718: Not sure what does this mean

paddle/fluid/platform/temporary_allocator.cc
chengduoZH: First, I am not sure the best value of excess fraction.

`reuse_tmp_allocation_excess_fraction` indicates the max memory size that `TemporaryAllocator` can return. 

This is to say, if user want to get 4KB from `TemporaryAllocator`, but the available memory of `TemporaryAllocator` only have 3KB, 4GB, we should not return 4GB generally, because it is too waste memory. 
If the available memory is 3KB, 6KB, 9KB, we should return 6KB; if user want to get the exacted equal size from `TemporaryAllocator`, he should set `reuse_tmp_allocation_excess_fraction` below 1.0  **this is why the fraction should be larger than 1.0.**

paddle/fluid/platform/temporary_allocator.cc
```diff
@@ -29,53 +31,60 @@ TemporaryAllocation::TemporaryAllocation(
...
+    if (FLAGS_limit_of_tmp_allocation > 0 &&
+        wait_delete_mem > static_cast<size_t>(FLAGS_limit_of_tmp_allocation)) {
+      PADDLE_ENFORCE(callback_ != nullptr, "The callback is non-initialized.");
       Release(callback_);
     }
     return;
   }
+  VLOG(10) << "Delete temporary allocation " << temp_allocation->ptr()
+           << " size: " << temp_allocation->size();
   delete temp_allocation;
```
panyx0718, line 87: Can this be missed if returned early?

chengduoZH: Why？If the memory is in CPU, it can be deleted directly.

panyx0718 (approved):

PR #15189, opened 2019-01-07
Enforce dim check in data feeder

sneaxiy:
This PR is for: 
- Enforce dim check in data feeder 
- Make data feeder support fixed dim inputs

### DIFF
--- a/python/paddle/fluid/data_feeder.py
+++ b/python/paddle/fluid/data_feeder.py
@@ -71,10 +71,20 @@ class DataToLoDTensorConverter(object):
             for each_data in data:
                 self._feed_impl_(each_data, lod[1:], lod_level - 1)
 
+    def _check_shape_(self, shape):
+        for s1, s2 in zip(self.shape, shape):
+            if s1 != s2 and s1 >= 0 and s2 >= 0:
+                raise ValueError(
+                    "Shape not match. What is defined in data layer is {}, but receive {}".
+                    format(self.shape, shape))
+
     def done(self):
         arr = numpy.array(self.data, dtype=self.dtype)
-        if self.shape and len(arr.shape) != len(self.shape):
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/data_feeder.py
```diff
@@ -71,10 +71,20 @@ def _feed_impl_(self, data, lod, lod_level):
             for each_data in data:
                 self._feed_impl_(each_data, lod[1:], lod_level - 1)
 
+    def _check_shape_(self, shape):
```
panyx0718: no need for `_` suffix.

python/paddle/fluid/data_feeder.py
```diff
@@ -71,10 +71,20 @@ def _feed_impl_(self, data, lod, lod_level):
...
+                    "Shape not match. What is defined in data layer is {}, but receive {}".
+                    format(self.shape, shape))
+
     def done(self):
         arr = numpy.array(self.data, dtype=self.dtype)
-        if self.shape and len(arr.shape) != len(self.shape):
-            arr = arr.reshape(self.shape)
+        if self.shape:
+            if len(arr.shape) != len(self.shape):
+                arr = arr.reshape(self.shape)
```
panyx0718: reshape can throw exception. Can catch this and throw a more user-friendly error message.

panyx0718 (dismissed):

python/paddle/fluid/data_feeder.py
sneaxiy: Done.

python/paddle/fluid/data_feeder.py
sneaxiy: Done.

panyx0718 (approved):

Issue #15032, opened 2018-12-25
test_analyzer_* random crashs at `compare_mkldnn`


luotao1:
- test_analyzer_small_dam random crashes at `compare_mkldnn` in three nightly CI at a same machine (5117).
```
test_analyzer_small_dam (NUMERICAL)
```
http://ci.paddlepaddle.org/viewLog.html?buildId=40622&tab=buildLog&buildTypeId=Paddle_PrCiNight&logTab=tree&filter=all&_focus=21635
http://ci.paddlepaddle.org/viewLog.html?tab=buildLog&buildTypeId=Paddle_PrCiNight&buildId=40460&_focus=22113
http://ci.paddlepaddle.org/viewLog.html?buildId=40324&tab=buildLog&buildTypeId=Paddle_PrCiNight&logTab=tree&filter=all&_focus=21764
-  test_analyzer_dam random crashes at `compare_mkldnn` in nightly CI
http://ci.paddlepaddle.org/viewLog.html?buildId=44600&tab=buildLog&buildTypeId=Paddle_PrCiNight&logTab=tree&filter=all&_focus=22935
- test_analyzer_mobilenet_depthwise_conv crashes at `compare_mkldnn` in nightly CI
http://ci.paddlepaddle.org/viewLog.html?buildId=44596&tab=buildLog&buildTypeId=Paddle_PrCiNight&logTab=tree&filter=all&_focus=23004

luotao1:
Is the same problem like #14174?

ghost:
@jczaja Could you help to have a check at this? Thanks!

luotao1:
It seems that mkldnn hangs from the log, is a openmp problem?

jczaja:
@luotao1 We will look to that and get back to You

jczaja:
@luotao1 I did not get to reproduce problem. 
1. Could you make a changes so that : compare_mkldnn works without mkl_dnn eg. compare(false) and then run experiments. So we know if MKL-DNN is actual problem ?
2. How did you observe that mkldnn hangs on logging? Did you attach debugger and observe where execution is looping over?
3. Do you use parallel mode of execution of tests of ctest? What is ctest commandline to run tests?

luotao1:
> Could you make a changes so that : compare_mkldnn works without mkl_dnn eg. compare(false) and then run experiments. So we know if MKL-DNN is actual problem ?

[...]

PR #15196, opened 2019-01-07
run analyzer_tester serial in multi-thread

luotao1:
try to solve #15032 
run analyzer_tester with multi-thread and mkldnn in serial mode.


jczaja (approved):
Looks good to me.

PR #15204, opened 2019-01-08
add the python callstack for debug support test=develop

wopeizl:


paddle/fluid/framework/operator.cc
```diff
@@ -19,17 +19,25 @@ limitations under the License. */
...
 #include "paddle/fluid/framework/var_type.h"
+#include "paddle/fluid/platform/debug_support.h"
 #include "paddle/fluid/platform/profiler.h"
 
 DECLARE_bool(benchmark);
 DEFINE_bool(check_nan_inf, false,
             "Checking whether operator produce NAN/INF or not. It will be "
             "extremely slow so please use this flag wisely.");
 
+DEFINE_bool(
```
panyx0718: Can this be a config in executor run API?

panyx0718: we want to support different debug levels. For level=1 print shape, for level=2 print debug NaN/Inf, etc

paddle/fluid/framework/operator.cc
```diff
@@ -155,7 +163,20 @@ RuntimeContext::RuntimeContext(const VariableNameMap& innames,
   }
 }
 
+void OperatorBase::PreHook(const Scope& scope, const platform::Place& place) {
+  auto attrName = OpProtoAndCheckerMaker::OpCreationCallstackAttrName();
+  if (HasAttr(attrName)) {
+    auto& callstack = Attr<std::vector<std::string>>(attrName);
+    platform::PythonDebugSupport::GetInstance()->SetInformation(callstack);
```
panyx0718: This python stack should be thrown all the time when there is an exception

paddle/fluid/framework/operator.h
```diff
@@ -160,6 +160,10 @@ class OperatorBase {
                                  const platform::Place& place,
                                  const RuntimeContext& ctx) const {}
 
+  // Add the hooks
+  virtual void PreHook(const Scope& scope, const platform::Place& place);
```
panyx0718: No need for hook. Just print the debug information at the right code place.

wopeizl:
image

paddle/fluid/framework/operator.h
wopeizl: remove them

paddle/fluid/framework/operator.cc
wopeizl: ok

python/paddle/fluid/framework.py
```diff
@@ -625,6 +626,11 @@ def __init__(self,
         if type is None:
             raise ValueError(
                 "`type` to initilized an Operator can not be None.")
+        else:
```
panyx0718, line 629: seems no need for this else

panyx0718 (approved):

PR #15248, opened 2019-01-09
add python inference api

fc500110:
Python inference API, support `AnalysisPredictor` and `NativePredictor`


python/paddle/fluid/executor.py
```diff
@@ -452,7 +454,9 @@ def run(self,
             fetch_var_name='fetch',
             scope=None,
             return_numpy=True,
-            use_program_cache=False):
+            use_program_cache=False,
+            infer_inputs=None,
```
panyx0718: reuse `feed` and behave the same?

python/paddle/fluid/executor.py
```diff
@@ -452,7 +454,9 @@ def run(self,
...
+            infer_attrs=None):
```
panyx0718: what is this for? If this is config, it should be put in with_inference_optimize.

python/paddle/fluid/executor.py
```diff
@@ -522,6 +528,15 @@ def run(self,
                 scope=scope,
                 return_numpy=return_numpy,
                 use_program_cache=use_program_cache)
+        else:
```
panyx0718: take a look at L543, and do something similar.

python/paddle/fluid/executor.py
```diff
@@ -27,6 +27,8 @@
 __all__ = ['Executor', 'global_scope', 'scope_guard']
 
 g_scope = core.Scope()
+InferNativeConfig = core.NativeConfig
+InferAnalysisConfig = core.AnalysisConfig
```
panyx0718, line 31: What do you need 2 Config? Can it be done in one config?

python/paddle/fluid/compiler.py
```diff
@@ -107,11 +109,18 @@ def with_data_parallel(self,
...
+            isinstance(config, InferAnalysisConfig)
+        ])
+        self._infer_config = config
+
     def _with_distributed(self):
         raise NotImplementedError()
 
     def _with_inference_optimize(self):
-        raise NotImplementedError()
+        self._infer_predictor = core.create_paddle_predictor(self._infer_config)
```
panyx0718: Predictor is created in _compile

panyx0718 (commented):
Please follow the example of data_parallel and be as consistent as possible.

python/paddle/fluid/executor.py
fc500110: config is used to create paddle predictor, `infer_attrs` is inference attribute, etc. `batch_size`, `wormup`.

python/paddle/fluid/executor.py
Superjomn: The `AnalysisConfig` and `NativeConfig` are exposed from C++ inference predictor directly, we hope the python API works as similar to the c++ inference as possible, so the existing inference users will be more clear facing the new python APIs.

The `AnalysisConfig` will create AnalysisPredictor while the NativeConfig will create NativePredictor, that is a quite common concept in the C++ inference.

fc500110: The two `Config` class correspond to `NativePredictor` and `AnalysisPredictor`. There is no class inheritance relations.

python/paddle/fluid/compiler.py
fc500110: fixed.

python/paddle/fluid/executor.py
fc500110: Could you tell me more details？

python/paddle/fluid/executor.py
fc500110: fixed.

python/paddle/fluid/executor.py
fc500110: The code has been refactored.

python/paddle/fluid/executor.py
panyx0718: warmup is config? batch size can be found in `feed`?

python/paddle/fluid/executor.py
panyx0718: Then they should be wrapped in one config?

python/paddle/fluid/executor.py
panyx0718: Please read the executor.py can try harder to be consistent. I believe you can do it.

```
        if program._is_data_parallel:
            return self._run_parallel(...)
```

python/paddle/fluid/executor.py
Superjomn: warmup is not an option of the existing AnalysisConfig.

warmup is just a configuration for profiling, not for inference.

We port the inference to python just for convenient of profiling and testing, so warm up is important.

python/paddle/fluid/executor.py
Superjomn: I don't think so.

For inference, it is important is to make C++ APIs concepts consistent in python, make the python inference program consistent with the C++ version.
To wrap the existing widely used C++ program and pretend to be something else is not the first goal.

If these APIs are not suitable for the executor, we can just expose the Pybind version, make a prefix with each API like `__` to warn users not to use it.

The current `PyNativeConfig` will be something like `__NativeConfig`, and we can write some testing program for convenient.

Superjomn (dismissed):
LGTM

python/paddle/fluid/compiler.py
```diff
@@ -109,10 +112,17 @@ def with_data_parallel(self,
             self._build_strategy = BuildStrategy()
         return self
 
-    def _with_distributed(self):
-        raise NotImplementedError()
+    def with_inference_optimize(self, config, infer_attrs):
```
panyx0718: need doc.

panyx0718 (dismissed):
LG API

panyx0718 (dismissed):
LG API

panyx0718 (approved):

panyx0718:
LG API

PR #15296, opened 2019-01-12
Async double buffered py reader

cjld:
This PR changes sync copy to async copy in double buffered py reader. It allowed the memcpy process can overlap with kernels. This PR improves single GPU performance about 1%, and 8 GPUs performance about 1~3%

before(memcpy can not overlap with kernels)
image

after (memcpy now overlapped with kernels)
image

### DIFF
--- a/paddle/fluid/API.spec
+++ b/paddle/fluid/API.spec
@@ -218,7 +218,7 @@ paddle.fluid.layers.shuffle ArgSpec(args=['reader', 'buffer_size'], varargs=None
 paddle.fluid.layers.batch ArgSpec(args=['reader', 'batch_size'], varargs=None, keywords=None, defaults=None)
 paddle.fluid.layers.double_buffer ArgSpec(args=['reader', 'place', 'name'], varargs=None, keywords=None, defaults=(None, None))
 paddle.fluid.layers.random_data_generator ArgSpec(args=['low', 'high', 'shapes', 'lod_levels', 'for_parallel'], varargs=None, keywords=None, defaults=(True,))
-paddle.fluid.layers.py_reader ArgSpec(args=['capacity', 'shapes', 'dtypes', 'lod_levels', 'name', 'use_double_buffer'], varargs=None, keywords=None, defaults=(None, None, True))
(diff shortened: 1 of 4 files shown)


paddle/fluid/API.spec
```diff
@@ -218,7 +218,7 @@ paddle.fluid.layers.shuffle ArgSpec(args=['reader', 'buffer_size'], varargs=None
 paddle.fluid.layers.batch ArgSpec(args=['reader', 'batch_size'], varargs=None, keywords=None, defaults=None)
 paddle.fluid.layers.double_buffer ArgSpec(args=['reader', 'place', 'name'], varargs=None, keywords=None, defaults=(None, None))
 paddle.fluid.layers.random_data_generator ArgSpec(args=['low', 'high', 'shapes', 'lod_levels', 'for_parallel'], varargs=None, keywords=None, defaults=(True,))
-paddle.fluid.layers.py_reader ArgSpec(args=['capacity', 'shapes', 'dtypes', 'lod_levels', 'name', 'use_double_buffer'], varargs=None, keywords=None, defaults=(None, None, True))
+paddle.fluid.layers.py_reader ArgSpec(args=['capacity', 'shapes', 'dtypes', 'lod_levels', 'name', 'use_double_buffer', 'use_cuda_pinned_place'], varargs=None, keywords=None, defaults=(None, None, True, None))
```
panyx0718: can we just default on this feature? without this argument?

panyx0718 (commented):
The idea sounds reasonable to me. @sneaxiy please comment, approve or reject

paddle/fluid/API.spec
cjld: Ok

cjld: Now both `py_reader` and `create_py_reader_by_data` will be affected by pinned place by default, and API will not changed. @sneaxiy

sneaxiy (approved):
LTGM.

One concern is the performance when we use Paddle compiled with GPU but only run CPU model. Not quite sure whether using CUDAPinnedPlace is better than CPUPlace. But it is difficult to know the place of double buffer inside py_reader. Let's add more tests to check the benchmark.

PR #15306, opened 2019-01-14
fix github issue 15267 test=develop

liupluswei:
fix issue: https://github.com/PaddlePaddle/Paddle/issues/15267
from the documentation, the return type should be int32: http://www.paddlepaddle.org/documentation/docs/zh/1.2/api_cn/layers_cn.html#permalink-149-shape


dzhwinter (dismissed):
LGTM

dzhwinter (approved):
Cheers

PR #15347, opened 2019-01-16
Fixdropscoe

gongweibao:


paddle/fluid/operators/distributed/variable_response.cc
```diff
@@ -117,8 +117,9 @@ bool VariableResponse::CopyLodTensorData(
       tensor->mutable_data(ctx.GetPlace(), ToVarType(meta_.data_type()));
 
   VLOG(6) << "Tensor.memory_size = " << tensor->memory_size()
-          << ", Buffer Size = " << length;
-  PADDLE_ENFORCE_EQ(tensor->memory_size(), static_cast<unsigned int>(length));
+          << ", Buffer Size = " << length << ", dims:" << dims
+          << ", numel:" << tensor->numel();
+  PADDLE_ENFORCE_GE(tensor->memory_size(), static_cast<unsigned int>(length));
```
jacquesqiao, line 122: can we use ?
``` cpp
PADDLE_ENFORCE_EQ(tensor->numel(), static_cast<unsigned int>(length));
```

gongweibao: 已经resize过了，numel()*sizeof(type)是resize过的值，没有必要在这里在ENFORCE一遍了。

jacquesqiao (approved):
LGTM,  please @velconia review 
```
num_iteration_per_drop_scope_
```

paddle/fluid/operators/distributed/variable_response.cc
gongweibao: 另外,memory_size() != data_size

paddle/fluid/framework/details/execution_strategy.h
```diff
@@ -25,7 +25,10 @@ struct ExecutionStrategy {
   size_t num_threads_{0};
   bool use_cuda_{true};
   bool allow_op_delay_{false};
-  size_t num_iteration_per_drop_scope_{1};
+  // If we set this to 1, we will delete all variables when finish a batch. and
+  // this will loss 15%+ performance.
+  // Please be aware about this parameters.
+  size_t num_iteration_per_drop_scope_{100};
```
typhoonzero, line 31: @chengduoZH What do you think about this change?

chengduoZH: In some models, this change will make training faster. Because if the local scopes are not dropped, the next iteration will reuse the memory, it can avoid alloc operation again.
But, in [`TransformData`](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/operator.cc#L1054-L1068), the transformed data will be stored in the local scope, and it will be not deleted until the drop_scope is called, for example, Transformer(big model).
So this option doesn't have a best default value.

PR #15363, opened 2019-01-16
make clone thread safe

Superjomn:
Users are used to create a multi-thread service using Inference API like

```c++
auto main_predictor = CreatePaddlePredictor(some_config);

for (int i = 0; i < num_threads; i++) {
   auto predictor = main_predictor->Clone();
   predictor.xxx
}
```

Our `Clone` API doesn't support this kind of usage due to thread unsafe issue.

This PR add a mutex and make it thread-safe.


luotao1:
Does this PR solve the multi-thread hang in `test_analyzer_rnn1`?
http://ci.paddlepaddle.org/viewLog.html?tab=buildLog&buildTypeId=Paddle_PrCiNight&buildId=49613

luotao1 (approved):
LGTM

PR #15365, opened 2019-01-16
fix save_inferece_model bug

Superjomn:
test=develop
image

`save_inference_model` will insert scale(1) after each target
image

the `identity_scale_op_clean_pass` will delete that `scale(1)` when inference executes

image


fc500110 (dismissed):

python/paddle/fluid/io.py
```diff
@@ -644,6 +645,17 @@ def save_inference_model(dirname,
     if main_program is None:
         main_program = default_main_program()
 
+    # fix the bug that the activation op's output as target will be pruned.
+    # will affect the inference performance.
```
luotao1, line 936: > will affect the inference performance

Could you give some benchmark?

luotao1 (approved):
LGTM. Could you add a unit test for identity_scale_op_clean_pass later?

Superjomn:
Will add one latter. @luotao1

PR #15376, opened 2019-01-16
Remove workspace_handle

chengduoZH:


paddle/fluid/operators/conv_fusion_op.cu.cc
```diff
@@ -104,7 +104,9 @@ class CUDNNConvFusionOpKernel : public framework::OpKernel<T> {
     // ------------------- cudnn conv algorithm ---------------------
     cudnnConvolutionFwdAlgo_t algo;
     auto handle = dev_ctx.cudnn_handle();
-    auto workspace_handle = dev_ctx.cudnn_workspace_handle();
+
+    Tensor cudnn_workspace;
```
typhoonzero, line 108: Maybe we don't need a tensor object for workspace?

chengduoZH: This is because the workspace of conv_cudnn may be `workspace_size_limit`(when `exhaustive_search` is true) or `workspace_size_in_bytes`(when `exhaustive_search` is false), and `workspace_size_in_bytes` is littler than `workspace_size_limit`. So if `exhaustive_search` is true, we only need alloc one `workspace_size_limit` space, and the following workspace can reuse this workspace.
But the return of`platform::DeviceTemporaryAllocator::Instance().Get(dev_ctx).Allocate(...)` is a `unique_ptr`, so we should use a `Tensor` to hold the `ptr`, because the under holder of `Tensor` is a `shared_ptr`.

typhoonzero: I see, DeviceTemporaryAllocator allocated memory will be put into a queue when "delete", and that memory can be reused by later on operators. Thanks~

typhoonzero (approved):

PR #15381, opened 2019-01-17
add multi process start script

typhoonzero:
Add tool to start multi-process mode to speed up multi gpu training.

Usage: `python tools/run_mp.py --gups 8 --cmd "python train.py --model ResNet101 --class_dim 1000"`


Yancey0623 (approved):
lgtm

PR #15417, opened 2019-01-18
Fix brpc compilation.

gongweibao:


typhoonzero (approved):
You may need to squash and merge to reduce the num of commits.

Issue #14988, opened 2018-12-20
Transformer Fails when MKLDNN is used


kbinias:
This commit made Transformer (used with MKL-DNN) crashed:
a872eb90c2384596ee46d29ad9f39bfe3f6ebc1b

@panyx0718 Could you please take a look and advice on this issue?

if there is a problem with reproduction of issue or more information is needed, feel free to ask.

Example log & reference command line for reproduction:

```
 FLAGS_use_mkldnn=True OMP_NUM_THREADS=1    python -u infer.py --src_vocab_fpath /nfs/fm/home/jczaja/data/wmt16_ende_data_bpe_clean/vocab_all.bpe.32000  --trg_vocab_fpath /nfs/fm/home/jczaja/data/wmt16_ende_data_bpe_clean/vocab_all.bpe.32000   --special_token '<s>' '<e>' '<unk>'   --test_file_pattern /nfs/fm/home/jczaja/data/wmt16_ende_data_bpe_clean/newstest2016.tok.bpe.32000.en-de   --token_delimiter ' '    model_path /nfs/fm/home/jczaja/models/iter_100000.infer.model   beam_size 4 max_out_len 255
-----------  Configuration Arguments -----------
batch_size: 1
device: CPU
display_output: False
num_profiling_passes: 100
opts: ['model_path', '/nfs/fm/home/jczaja/models/iter_100000.infer.model', 'beam_size', '4', 'max_out_len', '255']
pool_size: 10000
profile: True
skip_pass_num: 0
special_token: ['<s>', '<e>', '<unk>']
src_vocab_fpath: /nfs/fm/home/jczaja/data/wmt16_ende_data_bpe_clean/vocab_all.bpe.32000
test_file_pattern: /nfs/fm/home/jczaja/data/wmt16_ende_data_bpe_clean/newstest2016.tok.bpe.32000.en-de
token_delimiter:
trg_vocab_fpath: /nfs/fm/home/jczaja/data/wmt16_ende_data_bpe_clean/vocab_all.bpe.32000
------------------------------------------------
Traceback (most recent call last):
  File "infer-profile.py", line 308, in <module>
    infer(args)
  File "infer-profile.py", line 296, in infer
    inferencer(test_data, trg_idx2word)
  File "infer-profile.py", line 208, in fast_infer
    return_numpy=False)
  File "/home/jczaja/paddle/virtualpython/local/lib/python2.7/site-packages/paddle/fluid/executor.py", line 472, in run
[...]

PR #14900, opened 2018-12-14
[WIP] add analyzer_transformer_test

luotao1:
This PR can run `analyzer_transformer_test`, but could not be merged since it depends on #14632.

PR #11040, opened 2018-05-30
Mkldnn layout

mozga-intel:
This PR contains changes required to support MKLDNN memory layouts improving performance of MKLDNN computations.

This implementation differs from the proposed one in #10291. It is based mainly on Brian Liu implementation, which is simplified version of what I proposed previously. 

 The latest version of the pull request is splits into a few part of code ([information](https://github.com/PaddlePaddle/Paddle/pull/10915)).

This pull request contains:
- Implementation of the layout as a supports for mkldnn's operators.

Pull-request is related to:
- MKLDNN layout: Support for sum operator #11102
- MKLDNN layout: Support for pool operator #11101
- MKLDNN layout: Support for convolution operator #11099
- MKLDNN layout: Support for batch norm operator #11098 
- MKLDNN layout: Support for activation operator #11124

PR #15429, opened 2019-01-18
[MKL-DNN] MKL-DNN specific Tensor modification

jczaja:
This is first of series of PRs on tensor modifications for MKL-DNN operations that are replacing mkl-dnn format storing with MKL-DNN primitive descriptor storing.

Idea is that currently MKL-DNN format is stored  in Tensor. format is not good enough for Transpose MKL-DNN as format is ambigious in terms of idenditifcations of actual mkldnn-layout. More over it does require recreation of primitive descriptions for given tensor. 

Adventages:
-  Design should be clearer
-  Redundant recreation of user primitive desc should be removed

In this PR:
- Tensor was modified to store MKLDNN primitive desc
-  Transpose op was adjusted to work with new functions
-  ConvGrad MKL-DNN was barely adjusted as well as Gaussian Random mkl-dnn.
-  All other mkl-dnn ops were not modified . They will be adjusted in next PRs (convolution will come next)
- **No changes to non MKL-DNN PaddlePaddle (core PaddlePaddle concepts are not modified)**

### DIFF
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -27,6 +27,10 @@ limitations under the License. */
 #include "paddle/fluid/platform/enforce.h"
 #include "paddle/fluid/platform/place.h"
 
+#ifdef PADDLE_WITH_MKLDNN
+#include "paddle/fluid/platform/mkldnn_utils.h"
+#endif
+
 namespace paddle {
 
 namespace framework {
@@ -37,10 +41,38 @@ class Tensor {
 #ifdef PADDLE_WITH_MKLDNN
 
  public:
-  inline mkldnn::memory::format format() const { return format_; }
+  // TODO(jczaja): This is depracted and will be removed
+  inline mkldnn::memory::format format() const {
+    if (layout_ == DataLayout::kMKLDNN) {
+      return static_cast<mkldnn::memory::format>(mem_pd_.desc().data.format);
+    } else {
+      return mkldnn::memory::format::format_undef;
+    }
+  }
 
-  inline void set_format(const mkldnn::memory::format format) {
-    format_ = format;
+  // TODO(jczaja): This is depracted and will be removed
+  inline void set_format(
+      const mkldnn::memory::format fmt,
+      mkldnn::memory::data_type data_type = mkldnn::memory::f32) {
+    mem_pd_ = paddle::platform::create_prim_desc_from_format(
+        paddle::framework::vectorize2int(dims()), fmt, data_type);
+  }
+
+  inline mkldnn::memory::primitive_desc get_prim_desc() const {
+    return mem_pd_;
+  }
+
+  inline void set_prim_desc(mkldnn::memory::primitive_desc& mem_pd) {
+    // Internally MKL-DNN is just copying (increasing reference counter)
+    // to shared_ptr. So asignment should be quite cheap
+    mem_pd_ = mem_pd;
+  }
+
+  inline void create_prim_desc_from_dims(
+      mkldnn::memory::format fmt,
+      mkldnn::memory::data_type data_type = mkldnn::memory::data_type::f32) {
+    mem_pd_ = paddle::platform::create_prim_desc_from_dims(
+        paddle::framework::vectorize2int(dims()), fmt);
   }
 
  protected:
@@ -48,12 +80,9 @@ class Tensor {
    * @brief the detail format of memory block which have layout as kMKLDNN
    *
    * @note MKLDNN lib support various memory format like nchw, nhwc, nChw8C,
-   *       nChw16c, etc. For a MKLDNN memory block, layout will be set as
-   *       DataLayout::kMKLDNN meanwhile detail memory format will be kept in
-   *       this field.
+   *       nChw16c, etc. For a MKLDNN memory block, we store memory descriptor
    */
-
-  mkldnn::memory::format format_ = mkldnn::memory::format::format_undef;
+  mutable mkldnn::memory::primitive_desc mem_pd_;
 #endif
 
  public:
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -59,22 +59,14 @@ class Tensor {
     layout_ = DataLayout::kMKLDNN;
   }
 
-  inline mkldnn::memory::primitive_desc get_prim_desc() const {
+  inline mkldnn::memory::primitive_desc get_mkldnn_prim_desc() const {
     return mem_pd_;
   }
 
-  inline void set_prim_desc(mkldnn::memory::primitive_desc& mem_pd) {
+  inline void set_mkldnn_prim_desc(mkldnn::memory::primitive_desc& mem_pd) {
     // Internally MKL-DNN is just copying (increasing reference counter)
     // to shared_ptr. So asignment should be quite cheap
     mem_pd_ = mem_pd;
-    layout_ = DataLayout::kMKLDNN;
-  }
-
-  inline void create_prim_desc_from_dims(
-      mkldnn::memory::format fmt,
-      mkldnn::memory::data_type data_type = mkldnn::memory::data_type::f32) {
-    mem_pd_ = paddle::platform::create_prim_desc_from_dims(
-        paddle::framework::vectorize2int(dims()), fmt);
     layout_ = DataLayout::kMKLDNN;
   }
 
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -63,7 +63,8 @@ class Tensor {
     return mem_pd_;
   }
 
-  inline void set_mkldnn_prim_desc(mkldnn::memory::primitive_desc& mem_pd) {
+  inline void set_mkldnn_prim_desc(
+      const mkldnn::memory::primitive_desc& mem_pd) {
     // Internally MKL-DNN is just copying (increasing reference counter)
     // to shared_ptr. So asignment should be quite cheap
     mem_pd_ = mem_pd;
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -56,6 +56,7 @@ class Tensor {
       mkldnn::memory::data_type data_type = mkldnn::memory::f32) {
     mem_pd_ = paddle::platform::create_prim_desc_from_format(
         paddle::framework::vectorize2int(dims()), fmt, data_type);
+    layout_ = DataLayout::kMKLDNN;
   }
 
   inline mkldnn::memory::primitive_desc get_prim_desc() const {
@@ -66,6 +67,7 @@ class Tensor {
     // Internally MKL-DNN is just copying (increasing reference counter)
     // to shared_ptr. So asignment should be quite cheap
     mem_pd_ = mem_pd;
+    layout_ = DataLayout::kMKLDNN;
   }
 
   inline void create_prim_desc_from_dims(
@@ -73,6 +75,7 @@ class Tensor {
       mkldnn::memory::data_type data_type = mkldnn::memory::data_type::f32) {
     mem_pd_ = paddle::platform::create_prim_desc_from_dims(
         paddle::framework::vectorize2int(dims()), fmt);
+    layout_ = DataLayout::kMKLDNN;
   }
 
  protected:
--- a/paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
@@ -52,7 +52,7 @@ class TransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
                                              mkldnn_engine, key);
 
     auto transpose_src_memory_p = handler.AcquireSrcMemory(
-        input->format(), platform::to_void_cast<T>(input_data));
+        input->get_prim_desc(), platform::to_void_cast<T>(input_data));
     auto transpose_dst_memory_p =
         handler.AcquireDstMemory(output, ctx.GetPlace());
     auto transpose_p = handler.AcquireTranspose(transpose_dst_memory_p,
@@ -61,6 +61,15 @@ class TransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     std::vector<mkldnn::primitive> pipeline;
     pipeline.push_back(*transpose_p);
     mkldnn::stream(mkldnn::stream::kind::eager).submit(pipeline).wait();
+
+    // Transpose did change logical dimensions of Tensor, but reorder does not.
+    // Reorder does change only physical layout eg. format , strides
+    // so we need to create new primitive descriptor with changed logical layout
+    // so
+    // it match output shape
+
+    output->set_layout(DataLayout::kMKLDNN);
+    output->create_prim_desc_from_dims(mkldnn::memory::format::blocked);
   }
 };
 
@@ -103,7 +112,7 @@ class TransposeMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
                                              mkldnn_engine, key);
 
     auto transpose_src_memory_p = handler.AcquireSrcMemory(
-        out_grad->format(), platform::to_void_cast<T>(out_grad_data));
+        out_grad->get_prim_desc(), platform::to_void_cast<T>(out_grad_data));
     auto transpose_dst_memory_p =
         handler.AcquireDstMemory(x_grad, ctx.GetPlace());
     auto transpose_p = handler.AcquireTranspose(transpose_dst_memory_p,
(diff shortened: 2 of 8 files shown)


jczaja:
@panyx0718 To have CI passed I need your approve as PR is modifying tensor.h file. Coule you please take a look and comment/approve ?

luotao1:
> Transpose op was adjusted to work with new functions

Do you have some benchmark of transpose op before and after this PR?

jczaja:
@luotao1 This changes do not improve performance. After all PRs are merged, performance should be a tiny bit better. But it is more redesign MKL-DNN tensor to make transpose MKL-DNN working and have cleaner design. I know you were working on c-API transfomer (diffrent flavour of transformer than python transformer: https://github.com/PaddlePaddle/Paddle/pull/14900) . How can I test this new C-API transformer (python transformer currently does not work due to: https://github.com/PaddlePaddle/Paddle/issues/14988) ?

jczaja:
@luotao1 This PR does not improve performance. It is on making Transpose MKL-DNN working
in every scenario & make foundation for cleaner design. Performance results were made using resnet50 to make sure performance did not go down. Actually with this PR it is a bit slower (x < 0.5%), but will be improved with next PR. I would like to make extensive performance tests (when I have other ops modified)

jczaja:
@luotao1 I updated PR with some changes to convolution , so performance on resnet-50 (SKX 8180, B1 & B128) is restored to original level.

jczaja:
@luotao1 @panyx0718 Do you need any more information to have this PR reviewed?

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +37,84 @@ class Tensor {
...
    * @note MKLDNN lib support various memory format like nchw, nhwc, nChw8C,
-   *       nChw16c, etc. For a MKLDNN memory block, layout will be set as
-   *       DataLayout::kMKLDNN meanwhile detail memory format will be kept in
-   *       this field.
+   *       nChw16c, etc. For a MKLDNN memory block, we store memory descriptor
    */
-
-  mkldnn::memory::format format_ = mkldnn::memory::format::format_undef;
+  mutable mkldnn::memory::primitive_desc mem_pd_;
 #endif
```
tensor-tang, line 82: I think there is too much MKLDNN things in base `Tensor`.

MKLDNNTensor is better to read for adding library.

Need @panyx0718 approve this I think.

luotao1:
@tensor-tang How do think about https://github.com/PaddlePaddle/Paddle/pull/11040#discussion_r192349103

paddle/fluid/framework/tensor.h
luotao1: @tensor-tang How do think about https://github.com/PaddlePaddle/Paddle/pull/11040#discussion_r192349103

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +37,84 @@ class Tensor {
...
+
+  inline void setMKLDNNMemPrimDesc(mkldnn::memory::primitive_desc& mem_pd) {
+    // Internally MKL-DNN is just copying (increasing reference counter)
+    // to shared_ptr. So asignment should be quite cheap
+    mem_pd_ = mem_pd;
+  }
 
-  inline void set_format(const mkldnn::memory::format format) {
-    format_ = format;
+  void CreateMemoryPrimDescFromDims(
```
tensor-tang: 1. please unify the name style. like `setMKLDNNMemPrimDesc`, `CreateMemoryPrimDescFromDims`
2. MemoryPrimDesc specify the  data_type, but tensor do not aware this only if get data.

paddle/fluid/framework/tensor.h
tensor-tang: That 's just how this current version comes from, but not the reason we cloud continue adding more.

`primitive_desc` contains more info like dims, format and data_type which is duplicated with many conceptions in tensor itself.

panyx0718:
@sneaxiy Can you help take a look and see if there is a cleaner way?

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +37,84 @@ class Tensor {
...
+  // TODO(jczaja): This is depracted and will be removed
+  inline mkldnn::memory::format format() const {
+    if (layout_ == DataLayout::kMKLDNN) {
+      return static_cast<mkldnn::memory::format>(mem_pd_.desc().data.format);
+    } else {
+      return mkldnn::memory::format::format_undef;
+    }
+  }
+
+  // TODO(jczaja): This is depracted and will be removed
```
panyx0718, line 53: it will be removed? why add it?

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +37,84 @@ class Tensor {
 #ifdef PADDLE_WITH_MKLDNN
 
  public:
-  inline mkldnn::memory::format format() const { return format_; }
+  // TODO(jczaja): This is depracted and will be removed
+  inline mkldnn::memory::format format() const {
```
panyx0718, line 45: what if these format are put in Paddle's Layout enum? will it be cleaner without these ifdef PADDLE_WITH_MKLDNN

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +37,84 @@ class Tensor {
...
    *
    * @note MKLDNN lib support various memory format like nchw, nhwc, nChw8C,
-   *       nChw16c, etc. For a MKLDNN memory block, layout will be set as
-   *       DataLayout::kMKLDNN meanwhile detail memory format will be kept in
-   *       this field.
+   *       nChw16c, etc. For a MKLDNN memory block, we store memory descriptor
    */
-
-  mkldnn::memory::format format_ = mkldnn::memory::format::format_undef;
+  mutable mkldnn::memory::primitive_desc mem_pd_;
```
panyx0718, line 81: what does this contain? Can it be mapped to current tensor apis? such as numel,layer?

jczaja: @panyx0718 mem_pd_ is a primitive descriptor of MKL-DNN eg. This a structure that describes the implementation details of memory primitive eg. object does represent memory. So in that case mem_pd_ describes the dims of memory allocations , sizes, layout(way how memory is arranged, strides), and location of memory eg. cpu (MKL-DNN currently only supports CPU).

paddle/fluid/framework/tensor.h
jczaja: @panyx0718 format() and set_format() where modified . And both they will be removed. Why they are present is that all(apart from transpose mkldnn op) MKL-DNN ops are using set_format() , format() functions. but mkl-dnn format is not used internally in data_layout_transform . So As We want to step by step replace all mkl-dnn ops ' set_format() with setMKLDNNMemPrimDesc(). After all mkl-dnn ops have set_format() removed and replaced, then We can remove set_format()/ format() implementation.

paddle/fluid/framework/tensor.h
jczaja: @panyx0718 Yes, but as we are to move away from format based conversions of Tensors, then there is no point of doing so. Moreover MKL-DNN format does "map" some MKL-DNN layout configuration eg. dims and strides etc. There are no enough format names so they map to all possible configurations of layouts. Transpose MKL-DNN op is using layouts that do not have format assigned to them , other than generic name: mkldnn_blocked. So it is not that format <=> mkl-dnn layout. format is not enough , to represent mkl-dnn layouts , so it has to go

paddle/fluid/framework/tensor.h
jczaja: 1. It will be done
2. Yes. This is correct. data() / mutable_data() , returns pointer to allocation. Op has to ask separately for layout and/or Memory Primitive descriptor. Perhaps it would be good to  extend function data()/ mutable_data() so they can be aware of layouts and return converted data. Definitely having data() function that can return pointer to data in Paddle Layout , would help with implementing unit test so they can operate on block layout , and then we get result in paddle layout (NCHW or other) for comparison with reference. As paddlepaddle supports many devices (CPU,GPU,MKL-DNN...) I would like someone more familiar with other devices to trigger change in interface.

jczaja:
@tensor-tang ,
1. names of functions are unified.
2. I tried to create MKLDNNTensor as suggessted. Eg. Inheritance chain of  LODTensor  -> Tensor, now becomes LODTensor -> MKLDNNtensor -> Tensor .  This approach require quite alot of changes
In data_transform.cc , data_layout_layout_transform.cc  and in operator.cc (possibly more..) As Tensor is quite important entity tigtly integrated with ops and data loading. So it does require alot of changes eg. ifdef PADDLE_WITH_MKLDNN code.
    Other approach I tried is to Rename Tensor to TensorBase  and then for non-MKLDNN , alias TensorBase as Tensor , and for mkl-dnn builds MKLDNNTensor inherits from TensorBase and MKLDNNTensor is renamed to Tensor. This approach is also quite complex , aldo debugging it later may be complex.

So what are your suggestions on how to create MKLDNNTensor so it match current design ?

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +37,84 @@ class Tensor {
...
+
+  inline void set_prim_desc(mkldnn::memory::primitive_desc& mem_pd) {
+    // Internally MKL-DNN is just copying (increasing reference counter)
+    // to shared_ptr. So asignment should be quite cheap
+    mem_pd_ = mem_pd;
+  }
 
-  inline void set_format(const mkldnn::memory::format format) {
-    format_ = format;
+  void create_prim_desc_from_dims(
```
panyx0718: this doesn't need to be a method of tensor? Can this be independent function and create prim_desc elsewhere and then call set_prim_desc_?

tensor-tang:
> LODTensor -> MKLDNNtensor -> Tensor 

How about MKLDNNtensor -> LODTensor -> Tensor

jczaja:
@tensor-tang I haven't tried to implement MKLDNNTensor -> LoDTensor -> Tensor, as by looking into ops that are clearly checking if Tensor is LoDTensor etc. I suspected lots of changes . So I updated PR with original concept (no MKLDNNTensor) , but I moved most of MKLDNN functionality away to mkldnn_utils.h . Eg. Tensor contains thin wrapper functions to MKLDNN specific functions.

paddle/fluid/framework/tensor.h
jczaja: @panyx0718 I moved body of create_prim_desc_from_dims into utils file eg. mkldnn_utils.h  and left thin wrapper that only calls moved out function. Is this good enough or you suggest to remove entirely create_prim_desc_from_dims() ?

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +41,51 @@ class Tensor {
...
+  }
+
+  inline void set_prim_desc(mkldnn::memory::primitive_desc& mem_pd) {
+    // Internally MKL-DNN is just copying (increasing reference counter)
+    // to shared_ptr. So asignment should be quite cheap
+    mem_pd_ = mem_pd;
+    layout_ = DataLayout::kMKLDNN;
+  }
+
+  inline void create_prim_desc_from_dims(
```
panyx0718: I think the whole function can be moved to util? Then call set_mkldnn_prim_desc?

paddle/fluid/framework/tensor.h
```diff
@@ -37,23 +41,51 @@ class Tensor {
...
+      const mkldnn::memory::format fmt,
+      mkldnn::memory::data_type data_type = mkldnn::memory::f32) {
+    mem_pd_ = paddle::platform::create_prim_desc_from_format(
+        paddle::framework::vectorize2int(dims()), fmt, data_type);
+    layout_ = DataLayout::kMKLDNN;
+  }
 
-  inline void set_format(const mkldnn::memory::format format) {
-    format_ = format;
+  inline mkldnn::memory::primitive_desc get_prim_desc() const {
```
panyx0718: this method and the ones above should be called `mkldnn_prim_desc` to make its usage more obvious.

paddle/fluid/framework/tensor.h
jczaja: Ok, I moved whole functions to util and called set_mkldnn_prim_desc

paddle/fluid/framework/tensor.h
jczaja: @panyx0718 Ok, methods renamed as suggested.

jczaja:
@panyx0718 , @tensor-tang , @luotao1  I updated review following your suggestions. Please review

paddle/fluid/platform/mkldnn_utils.h
```diff
@@ -0,0 +1,72 @@
...
+  return mkldnn::memory::primitive_desc(mem_fmt, cpu_engine);
+}
+
+inline mkldnn::memory::primitive_desc create_prim_desc_from_format(
+    const std::vector<int>& ltz, const mkldnn::memory::format format,
+    const mkldnn::memory::data_type data_type) {
+  auto md = mkldnn::memory::desc({ltz}, data_type, format);
+  auto& pool = platform::DeviceContextPool::Instance();
+  auto place = paddle::platform::CPUPlace();
+  auto* dev_ctx = dynamic_cast<platform::MKLDNNDeviceContext*>(pool.Get(place));
```
panyx0718: check nullptr?

panyx0718 (commented):
@tensor-tang @luotao1 any comments? seems ok to me.

paddle/fluid/platform/mkldnn_utils.h
```diff
@@ -0,0 +1,72 @@
+/* Copyright (c) 2019 PaddlePaddle Authors. All Rights Reserved.
```
luotao1, line 1: There is `paddle/fluid/platform/mkldnn_reuse.h`, why not combine `mkldnn_utils.h` into `mkldnn_reuse.h`?

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
```diff
@@ -61,6 +61,17 @@ class TransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     std::vector<mkldnn::primitive> pipeline;
     pipeline.push_back(*transpose_p);
     mkldnn::stream(mkldnn::stream::kind::eager).submit(pipeline).wait();
+
+    // Transpose did change logical dimensions of Tensor, but reorder does not.
+    // Reorder does change only physical layout eg. format , strides
+    // so we need to create new primitive descriptor with changed logical layout
+    // so
+    // it match output shape
+
```
tensor-tang: redundant line.

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
```diff
@@ -112,6 +124,17 @@ class TransposeMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
...
```
tensor-tang: redundant line.

paddle/fluid/platform/mkldnn_utils.h
tensor-tang: Agreed.

paddle/fluid/platform/mkldnn_utils.h
```diff
@@ -0,0 +1,72 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */
+
+#pragma once
+#include <mkldnn.h>
+#include <string>
+
+#ifdef PADDLE_WITH_MKLDNN
```
tensor-tang: Why `mkldnn_helper.h` and `mkldnn_reuse.h` do not need this?

paddle/fluid/platform/mkldnn_utils.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace paddle {
+
```
tensor-tang: redundant line.

paddle/fluid/platform/mkldnn_utils.h
jczaja: @tensor-tang Good catch. mkldnn_utils, should not have mentioned line either. This line was removed.

paddle/fluid/platform/mkldnn_utils.h
jczaja: @luotao1 , @tensor-tang  mkldnn_utils.h contains routines that are used within tensor.h. We cannot move mkldnn_utils.h into mkldnn_reuse.h or mkldnn_helper.h , as both of those headers are including tensor.h (via operator.h). So we would have tensor.h including mkldnn_reuse.h , mkldnn_reuse.h including tensor.h  ... which is cyclic dependency. That is why it is separate file

paddle/fluid/platform/mkldnn_utils.h
jczaja: @panyx0718 good suggestion

jczaja:
@panyx0718 , @luotao1 , @tensor-tang I adjusted changes to your suggestions. Please take a look. CI cannot pass without @panyx0718 approval.

panyx0718 (approved):

PR #15434, opened 2019-01-19
 [enhancement] Make lod reset op support for appending lod level

wanghaoshuang:


python/paddle/fluid/layers/nn.py
```diff
@@ -6128,6 +6128,7 @@ def lod_reset(x, y=None, target_lod=None):
                            from :attr:`y`.
         target_lod (list|tuple|None): One level LoD which should be considered
                                       as target LoD when :attr:`y` not provided.
+        append (bool): A flag indicating whether to append lod level instead of reseting.
```
chengduoZH, line 6131: It is confusing that adding `append` for `lod_reset`. In my mind that `lod_reset` is just used to reset the lod info with `target_lod`, but not have other function.

panyx0718: agree

PR #15449, opened 2019-01-21
Check collective server's data.

gongweibao:


paddle/fluid/operators/distributed/collective_server_test.cc
```diff
@@ -80,6 +81,20 @@ void Gather(const std::vector<distributed::RemoteVar>& vars,
...
+  ASSERT_EQ(dst[0]->height(), 20000);
+  ASSERT_EQ(dst[0]->rows().size(), static_cast<size_t>(3));
+  for (int i = 0; i < 3; i++) {
+    ASSERT_EQ(dst[0]->rows()[i], i);
+  }
+
+  std::vector<float> vec;
+  TensorToVector(dst[0]->value(), *dev_ctx, &vec);
+  for (size_t i = 0; i < 20000 * 1024; i++) {
+    ASSERT_TRUE(vec[i] - 32.7 < 0.000001);
```
typhoonzero: ASSERT_FLOAT_EQ

paddle/fluid/operators/distributed/collective_server_test.cc
```diff
@@ -80,6 +81,20 @@ void Gather(const std::vector<distributed::RemoteVar>& vars,
...
+  }
```
typhoonzero, line 97: Why need to call `Gather(vars, &ctx);` twice?

paddle/fluid/operators/distributed/collective_server_test.cc
```diff
@@ -80,6 +81,20 @@ void Gather(const std::vector<distributed::RemoteVar>& vars,
...
+  ASSERT_EQ(dst[0]->value().dims(), framework::make_ddim({20000, 1024}));
+  ASSERT_EQ(dst[0]->height(), 20000);
+  ASSERT_EQ(dst[0]->rows().size(), static_cast<size_t>(3));
+  for (int i = 0; i < 3; i++) {
+    ASSERT_EQ(dst[0]->rows()[i], i);
+  }
+
+  std::vector<float> vec;
+  TensorToVector(dst[0]->value(), *dev_ctx, &vec);
+  for (size_t i = 0; i < 20000 * 1024; i++) {
```
typhoonzero: should only check sparse rows.

typhoonzero: should initialize the selected rows with only several rows but not all.

paddle/fluid/operators/distributed/collective_server_test.cc
gongweibao: It starts a ps server and starts 2 trainers, so it can test if there's a deadlock.

paddle/fluid/operators/distributed/collective_server_test.cc
gongweibao: Done. Thanks!

paddle/fluid/operators/distributed/collective_server_test.cc
gongweibao: Done. Thanks!

PR #15450, opened 2019-01-21
Disable cache for scope creation to workaround CI failure

ghost:
Current code in TryCreateTransferScope (use thread_local as cache to avoid creating transfer scope) may have random failure which cause this CI failure (hangs as timeout, segfault as crash). By simply comment out it (i.e. don't use cache; always create "new_scope"), we won't run into any error anymore. We trial with several times to enable/disable this cache and proved no issue occur when cache is disabled.

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1054,8 +1054,8 @@ Scope* OperatorWithKernel::PrepareData(
       // called by a NaiveExecutor, the NaiveExecutor will cache the scopes and
       // variables, that behavior a lot different.
       if (!run_by_executor_) {
-        new_scope = TryCreateTransferScope(kernel_type_for_var,
-                                           expected_kernel_key, &scope);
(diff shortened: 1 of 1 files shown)


paddle/fluid/framework/operator.cc
```diff
@@ -1054,8 +1054,8 @@ Scope* OperatorWithKernel::PrepareData(
       // called by a NaiveExecutor, the NaiveExecutor will cache the scopes and
       // variables, that behavior a lot different.
       if (!run_by_executor_) {
-        new_scope = TryCreateTransferScope(kernel_type_for_var,
-                                           expected_kernel_key, &scope);
+        // new_scope = TryCreateTransferScope(kernel_type_for_var,
```
Superjomn, line 1057: This will result in memory leak in inference when `ON_INFER` turned on.

bingyanghuang:
@jianhang-liu  Do we need to keep this PR and target to merge?

PR #15454, opened 2019-01-21
tweak the executor implementation to better match origin behavior

panyx0718:
test=develop

### DIFF
--- a/python/paddle/fluid/executor.py
+++ b/python/paddle/fluid/executor.py
@@ -305,7 +305,9 @@ class Executor(object):
     def __init__(self, place):
         self.place = place
         self.program_caches = dict()
-        self.executor = None
+        p = core.Place()
+        p.set_place(self.place)
+        self._default_executor = core.Executor(p)
         self._closed = False
 
     def _get_program_cache(self, program_cache_key):
@@ -397,12 +399,13 @@ class Executor(object):
             >>> ...
             >>> exe.close()
         """
-        if not self._closed and self.executor:
-            self.executor.close()
+        if not self._closed:
+            self._default_executor.close()
             self._closed = True
 
     def _run_parallel(self, program, scope, feed, fetch_list, fetch_var_name,
                       return_numpy):
+        exe = program._executor
         if isinstance(feed, dict):
             feed_tensor_dict = dict()
             for feed_name in feed:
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/executor.py
```diff
@@ -589,11 +588,11 @@ def _run(self, program, feed, fetch_list, feed_var_name, fetch_var_name,
...
+        exe.run(program.desc, scope, 0, True, True)
         outs = self._fetch_data(fetch_list, fetch_var_name, scope)
         if return_numpy:
             outs = as_numpy(outs)
         return outs
 
-    def _run_inference(self, program, feed):
-        return self.executor.run(feed)
+    def _run_inference(self, exe, feed):
+        return exe.run(feed)
```
chengduoZH, line 598: Doesn't `Inference` need fetch data?

python/paddle/fluid/executor.py
```diff
@@ -397,12 +399,13 @@ def close(self):
...
         """
-        if not self._closed and self.executor:
-            self.executor.close()
+        if not self._closed:
+            self._default_executor.close()
             self._closed = True
 
     def _run_parallel(self, program, scope, feed, fetch_list, fetch_var_name,
                       return_numpy):
+        exe = program._executor
```
chengduoZH, line 408: `program` is a bit confusing with `fluid.Program`, it should be `compiled_program`.
And the `exe` should be `parallel_exe`.

python/paddle/fluid/executor.py
```diff
@@ -414,8 +417,7 @@ def _run_parallel(self, program, scope, feed, fetch_list, fetch_var_name,
                     feed_tensor.set(feed[feed_name], core.CPUPlace())
                 feed_tensor_dict[feed_name] = feed_tensor
 
-            self.executor.feed_and_split_tensor_into_local_scopes(
-                feed_tensor_dict)
+            exe.feed_and_split_tensor_into_local_scopes(feed_tensor_dict)
         elif isinstance(feed, list) or isinstance(feed, tuple):
             if len(feed) != len(program._places):
```
chengduoZH, line 422: Same above, program is a bit confusing with fluid.Program

panyx0718: not a big problem now, can improve in the future.

python/paddle/fluid/executor.py
panyx0718: not a big problem now, can improve in the future.

python/paddle/fluid/executor.py
panyx0718: This is inference's problem, should improve in the future.

chengduoZH (approved):

PR #15457, opened 2019-01-22
add dynamic memory optim

Superjomn:
Inference memory optimization without the need of offline pre-execution.

It will co-operate with the existing static memory optimization, and is easier to use.


NHZlX (approved):
LGTM， I will test this module with other models later.

PR #15461, opened 2019-01-22
fix trt stream bug.

NHZlX:
BUG: After continuing to input different data, the output cannot be aligned


paddle/fluid/operators/tensorrt/tensorrt_engine_op_test.cc
```diff
@@ -99,7 +99,7 @@ TEST(TensorRTEngineOp, manual) {
   SetAttr<std::string>(engine_op_desc.Proto(), "subgraph",
                        block_->SerializeAsString());
   SetAttr<int>(engine_op_desc.Proto(), "max_batch_size", 2);
-  SetAttr<int>(engine_op_desc.Proto(), "workspace_size", 2 << 10);
+  SetAttr<int>(engine_op_desc.Proto(), "workspace_size", 2 << 20);
```
Superjomn: Why change to `2<<20` ?

Superjomn (commented):
Add a test to check that different inputs and get right outputs.

paddle/fluid/operators/tensorrt/tensorrt_engine_op_test.cc
NHZlX: We should change it to 1 << 20 (the Official default). 1 << 10 is too small for trt5.

Superjomn (dismissed):
LGTM

Superjomn (approved):
LGTM

PR #15472, opened 2019-01-22
[RFC] add INT8 placement pass

wojtuss:
This is a request for comments.

There is a need for applying INT8 quantization strategy in PaddlePaddle models. This patch adds a mechanism for applying INT8 quantization strategy similar to MKL-DNN approach.

A `use_int8` op attribute and `int8_placement_pass` in the simplest form (similar to the `use_mkldnn` attribute and `mkldnn_placement_pass`) are added. The pass will allow a user to choose which operators should be quantized and use INT8 kernels.

We envisage more passes for INT8 optimization that would utilize the `use_int8` attribute.

test=develop


sfraczek (commented):
LGTM, but push the pass code also. :D

hshen14 (commented):
As discussed earlier, I would suggest to maintain such INT8 kernel op list into calibration tool since it is better not to expose such information in non-INT8-MKLDNN-device model (e.g., FP32, GPU). Calibration tool should be the better place to handle/maintain such list.

wojtuss:
> As discussed earlier, I would suggest to maintain such INT8 kernel op list into calibration tool since it is better not to expose such information in non-INT8-MKLDNN-device model (e.g., FP32, GPU). Calibration tool should be the better place to handle/maintain such list.

@hshen14 , the method `SetINT8Op()` (https://github.com/PaddlePaddle/Paddle/blob/e0fb2578a53de189fd9c675850c00cc80b657cd4/paddle/fluid/inference/api/paddle_analysis_config.h#L193) can be used in the Calibration tool to choose operators for quantization (cf. the `SetMKLDNNOp()` usage in https://github.com/PaddlePaddle/Paddle/blob/5a8bd82c0cef63bd7313171e8049953aa2db43f6/paddle/fluid/inference/tests/api/analyzer_dam_tester.cc#L198)
Furthermore, the `int8_placement_pass` can be more sophisticated. Here it applies only the simplest strategy.

paddle/fluid/framework/ir/int8_placement_pass.cc
```diff
@@ -0,0 +1,48 @@
+/* Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
sfraczek: 2019 :)

ghost:
@wojtuss I understood this is the 1st version so now it only implement the simplest "placement" strategy, i.e. set "use_int8 = true" for a OP if it's in the passed in "int8 op list". Similar as what we had done for mkldnn placement. However, it's not enough even as the 1st version. We have to be able to specify which instances of OP (e.g. 1st conv) should be included (or excluded) for quantization. Remember that the 1st conv Op can't be quanzied due to we (Xeon) only support s8u8 now and there is no RELU before the 1st conv OP.

hshen14:
> @wojtuss I understood this is the 1st version so now it only implement the simplest "placement" strategy, i.e. set "use_int8 = true" for a OP if it's in the passed in "int8 op list". Similar as what we had done for mkldnn placement. However, it's not enough even as the 1st version. We have to be able to specify which instances of OP (e.g. 1st conv) should be included (or excluded) for quantization. Remember that the 1st conv Op can't be quanzied due to we (Xeon) only support s8u8 now and there is no RELU before the 1st conv OP.

No, MKL-DNN provides all the capability of conv Op INT8, including first conv Op.

hshen14:
> This is a request for comments.
> 
> There is a need for applying INT8 quantization strategy in PaddlePaddle models. This patch adds a mechanism for applying INT8 quantization strategy similar to MKL-DNN approach.
> 
> A `use_int8` op attribute and `int8_placement_pass` in the simplest form (similar to the `use_mkldnn` attribute and `mkldnn_placement_pass`) are added. The pass will allow a user to choose which operators should be quantized and use INT8 kernels.
> 
> We envisage more passes for INT8 optimization that would utilize the `use_int8` attribute.
> 
> test=develop

Please also consider to make the design more general, e.g., to support FP16, BF16, or other mixed precision. Is it scalable to add other attributes and maintain different separate pass? Probably we can refer to FP16 training in GPU already done in Paddle. Any example to specify such information? @luotao1 @panyx0718  Thanks.

paddle/fluid/framework/ir/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ cc_library(fuse_pass_base SRCS fuse_pass_base.cc DEPS pass)
 pass_library(graph_to_program_pass base)
 pass_library(graph_viz_pass base)
 pass_library(lock_free_optimize_pass base)
+pass_library(int8_placement_pass base)
```
Superjomn, line 35: to mkldnn_int8_placement_pass ?

paddle/fluid/inference/analysis/argument.h
```diff
@@ -123,6 +123,10 @@ struct Argument {
   DECL_ARGUMENT_FIELD(mkldnn_enabled_op_types, MKLDNNEnabledOpTypes,
                       std::unordered_set<std::string>);
 
+  // Pass a set of op types to enable its int8 kernel
+  DECL_ARGUMENT_FIELD(int8_enabled_op_types, INT8EnabledOpTypes,
```
Superjomn, line 127: mkldnn_int8 ? 

There might be GPU int8 latter.

paddle/fluid/inference/api/paddle_analysis_config.h
```diff
@@ -151,6 +151,13 @@ struct AnalysisConfig {
    */
   bool mkldnn_enabled() const { return use_mkldnn_; }
 
+  /** Turn on INT8.
+   */
+  void EnableINT8();
```
Superjomn, line 156: EnableMkldnnInt8 ?

paddle/fluid/inference/api/paddle_analysis_config.h
```diff
@@ -180,6 +187,13 @@ struct AnalysisConfig {
     mkldnn_enabled_op_types_ = op_list;
   }
 
+  /** Specify the operator type list to use INT8 kernel.
+   * @param op_list the operator type list.
+   */
+  void SetINT8Op(std::unordered_set<std::string> op_list) {
```
Superjomn, line 193: MkldnnInt8 ?

paddle/fluid/framework/ir/CMakeLists.txt
ghost: I personally don't suggest to name it as "mkldnn_int8" considering we may use various mechanism (int8 mkl, int8 mkldnn, etc.) to implement int8 kernel.

paddle/fluid/framework/ir/int8_placement_pass.cc
```diff
@@ -0,0 +1,48 @@
...
+namespace ir {
+
+std::unique_ptr<ir::Graph> INT8PlacementPass::ApplyImpl(
+    std::unique_ptr<ir::Graph> graph) const {
+  VLOG(3) << "Aplies INT8 placement strategy.";
+  const auto& op_types_list =
+      Get<std::unordered_set<std::string>>("int8_enabled_op_types");
+  for (const Node* n : graph->Nodes()) {
+    if (n->IsOp()) {
+      auto* op = n->Op();
```
Superjomn, line 29: Add an enforce here.

`PADDLE_ENFORCE_NOT_NULL(n->Op())`

wojtuss:
> @wojtuss I understood this is the 1st version so now it only implement the simplest "placement" strategy, i.e. set "use_int8 = true" for a OP if it's in the passed in "int8 op list". Similar as what we had done for mkldnn placement. However, it's not enough even as the 1st version. We have to be able to specify which instances of OP (e.g. 1st conv) should be included (or excluded) for quantization. Remember that the 1st conv Op can't be quanzied due to we (Xeon) only support s8u8 now and there is no RELU before the 1st conv OP.

@jianhang-liu 
As @hshen14 wrote, quantization of all the convolutions (including the first one) can be supported.
With a pass like the `int8_placement_pass` and using `GraphPatternDetector` you can very precisely pick operators for quantization according to a given pattern (eg. look for a sequence of more than one conv ops and start quantization from the second one). This is how it works for any other fuse pass.

paddle/fluid/framework/ir/CMakeLists.txt
wojtuss: @Superjomn , INT8 optimization is independent of MKL-DNN, we would like to keep them separately. There are operators which we are going to quantize without MKL-DNN.

luotao1: @wojtuss Do you prefer `mkldnn_int8_placement_pass` as well?

luotao1: @wojtuss @Superjomn @jianhang-liu How about `cpu_int8_placement_pass`?

hshen14:
> > @wojtuss I understood this is the 1st version so now it only implement the simplest "placement" strategy, i.e. set "use_int8 = true" for a OP if it's in the passed in "int8 op list". Similar as what we had done for mkldnn placement. However, it's not enough even as the 1st version. We have to be able to specify which instances of OP (e.g. 1st conv) should be included (or excluded) for quantization. Remember that the 1st conv Op can't be quanzied due to we (Xeon) only support s8u8 now and there is no RELU before the 1st conv OP.
> 
> @jianhang-liu
> As @hshen14 wrote, quantization of all the convolutions (including the first one) can be supported.
> With a pass like the `int8_placement_pass` and using `GraphPatternDetector` you can very precisely pick operators for quantization according to a given pattern (eg. look for a sequence of more than one conv ops and start quantization from the second one). This is how it works for any other fuse pass.

Brian mentioned the op instance in the meeting and Baidu suggested to use sorted topology in a vector, which can probably meet the requirement.

paddle/fluid/framework/ir/int8_placement_pass.cc
sfraczek: When can op be null?

wojtuss:
> Brian mentioned the op instance in the meeting and Baidu suggested to use sorted topology in a vector, which can probably meet the requirement.

@hshen14 what do you mean by a sorted topology?

panyx0718 (approved):
LG

wojtuss:
This PR is updated and included in the quantization core: https://github.com/PaddlePaddle/Paddle/pull/15987

wojtuss:
Closed, as became a part of the INT8v2 core: https://github.com/PaddlePaddle/Paddle/pull/15987

PR #15498, opened 2019-01-23
fix check_doc_approval.py ;test=develop

tianshuo78520a:
There was a problem with the previous method of checking documents. I fix this problem and use the same judgment method as API.
Generates a new text build fluid_document.spec .If you want to modify the comment under all of fluid, you need to modify the file first. The file is followed by the MD5 value of the function doc. Check_doc_approval.py is used to get the file and find the MD5 value of the corresponding function to modify.


paddle/fluid/fluid_document.spec
```diff
@@ -0,0 +1,376 @@
...
+ ('paddle.fluid.regularizer.L1Decay', '5fa1107d292765dc06fe81b7612d6ea2')
+ ('paddle.fluid.regularizer.L2Decay', 'f83c7a9862940772424eb0ae79ca21d7')
+ ('paddle.fluid.regularizer.L1DecayRegularizer', '5fa1107d292765dc06fe81b7612d6ea2')
+ ('paddle.fluid.regularizer.L2DecayRegularizer', 'f83c7a9862940772424eb0ae79ca21d7')
+ ('paddle.fluid.transpiler.DistributeTranspiler', '8653d0a1568d1524ab2e664cc443f8b1')
+ ('paddle.fluid.transpiler.memory_optimize', 'eda17d0f1639bc6ca215cecf87f588a4')
+ ('paddle.fluid.transpiler.release_memory', 'ac4114d3df16264f1946deb3a8434a6f')
+ ('paddle.fluid.transpiler.HashName', 'a519f5335c70bd17e467d5d9af559bc1')
+ ('paddle.fluid.transpiler.RoundRobin', '7b22b6f83d9f3f782f58df18933dc7cb')
+ ('paddle.fluid.transpiler.DistributeTranspilerConfig', 'cb46bd12b047faf5eeb405ff60f01b0a')])
```
typhoonzero, line 376: I'm thinking a better way to not to add `.spec` files in the reponsitory, e.g. we can generate the develop spec file before applying current PR changes, then generate again the changed spec after the PR branch is applied. So that developers won't care about to update this file if anything was going to be changed.

The reason for the API.spec is that it can be generated only when the branch was built. `.spec` file can reduce build time on CI environment. But for this change seems we can apply the method said above.

typhoonzero: @panyx0718 please take a look at this? seems this file still need to build and generate `whl` package first before checking.

paddle/scripts/paddle_build.sh
```diff
@@ -531,16 +531,17 @@ function assert_api_spec_approvals() {
     fi
 
     pip install ${PADDLE_ROOT}/build/opt/paddle/share/wheels/*.whl
-    CHECK_DOCK_MD5=`python ${PADDLE_ROOT}/tools/check_doc_approval.py`
-    if [ "True" != ${CHECK_DOCK_MD5} ]; then
+    python ${PADDLE_ROOT}/tools/check_doc_approval.py > new_md5.spec
+    sed -i 's#),#)\n#g' new_md5.spec
+    check_fluid_document=`python ${PADDLE_ROOT}/tools/diff_api.py ${PADDLE_ROOT}/paddle/fluid/fluid_document.spec new_md5.spec`
```
panyx0718: do we need to update fluid_document.spec in the PR like API.spec? How to do that?

tianshuo78520a: I only thought of this method, and I tried to make it the same as api. spec.

tools/print_signatures.py
```diff
@@ -68,7 +73,8 @@ def visit_all_module(mod):
             visit_member(mod.__name__, instance)
 
 
-modules = sys.argv[1].split(",")
+#modules = sys.argv[1].split(",")
```
luotao1: remove unused line76

luotao1 (commented):
Besides, could API.spec check and const_cast check are moved to travis_ci, in order to reduce the rerun time of CI?

tools/print_signatures.py
tianshuo78520a: ok

paddle/fluid/fluid_document.spec
```diff
@@ -0,0 +1,376 @@
+OrderedDict([('paddle.fluid.Program', '54ea7f106c3e8337f48653a56e9793e0')
```
panyx0718, line 1: Can this file be removed since the information is in API.spec?

PR #15503, opened 2019-01-24
fix bug in merge_ids

seiriosPlus:
merge_ids在ID超过int32范围后出错，表现为参数中包含nan或计算auc的时候出nan:

the exception is:
```
paddle.fluid.core.EnforceNotMet: Tensor embedding_28.tmp_0 contains NAN at [/paddle/paddle/paddle/fluid/framework/operator.cc:851]
PaddlePaddle Call Stacks: 
0       0x7fc577330ad5p void paddle::platform::EnforceNotMet::Init<char const*>(char const*, char const*, int) + 357
1       0x7fc577330e59p 
...
```

### DIFF
--- a/paddle/fluid/operators/distributed_ops/merge_ids_op.h
+++ b/paddle/fluid/operators/distributed_ops/merge_ids_op.h
@@ -89,7 +89,7 @@ class MergeIdsOpKernel : public framework::OpKernel<T> {
           framework::make_ddim({nums, embedding_size}), place);
       for (int j = 0; j < nums; ++j) {
         auto id = out_ids->data<int64_t>()[j];
-        auto row_tuple = selected_rows_idx_map[id];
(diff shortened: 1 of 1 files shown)


jacquesqiao:
```cpp
    int embedding_size = 0;

    for (size_t i = 0; i < x_tensors.size(); ++i) {
      const auto *x_tensor = x_tensors[i];
      const auto *row_id = row_ids[i];

      if (embedding_size == 0) {
        embedding_size = x_tensor->dims()[1];
      }
      PADDLE_ENFORCE_EQ(embedding_size, x_tensor->dims()[1],
                        "embedding size of all input should be the same");
      row_size += x_tensor->dims()[0];
      row_ids_size += row_id->dims()[0];
    }
```
the type of embedding_size also need to set to int64_t

jacquesqiao (approved):
LGTM!

PR #15509, opened 2019-01-24
Rewrite is_empty op to avoid unnecessary data transform.

Xreki:
`is_empty op` is used to determine whether the input has elements. So that only need to access the `dims` of input. If the input has elements, the op will set `output[0] = true`. Now `is_empty` op is always run on CPU. If the input tensor is on GPU, there is no need to transform the input data from GPU to CPU, because the transformed CPU data will never be accessed. 

Test the op's performance in `1d-attention` model:

- Before
```
Event                              Calls       Total       CPU Time (Ratio)        GPU Time (Ratio)        Min.        Max.        Ave.        Ratio.
thread0::is_empty                  1300        71.7622     69.393406 (0.966991)    2.368766 (0.033009)     0.019874    0.238775    0.0552017   0.0158786
```

- After: inherit `OperatorBase` and rewrite `RunImpl`
```
Event                              Calls       Total       CPU Time (Ratio)        GPU Time (Ratio)        Min.        Max.        Ave.        Ratio.
thread0::is_empty                  1300        3.96455     3.961327 (0.999187)     0.003224 (0.000813)     0.00165     0.03931     0.00304965  0.000836897
```

- Register GPU kernel but still do the kernel on CPU
```
Event                              Calls       Total       CPU Time (Ratio)        GPU Time (Ratio)        Min.        Max.        Ave.        Ratio.
thread0::is_empty                  1300        5.52518     5.521585 (0.999349)     0.003598 (0.000651)     0.002836    0.03293     0.00425014  0.00130384
```


panyx0718 (commented):
can you give more detail why you need to avoid data transform?

Xreki:
@panyx0718 I think the reason is intuitive. Because data transform is time-consuming, and there is no need to access input data, but just need to get the total number of input elements. So the data transform is not necessary at all. I do this change to reduce the time overhead.

panyx0718 (commented):
If you register some CUDAKernel, then it doesn't need to transform data?

Xreki:
@panyx0718 Yes. What `is_empty` op do is to determine whether the input tensor has elements. Thus, there is no need to do any data transform no matter the input is on CPU or GPU. In the current implementation in develop branch, `is_empty` is always done on CPU, that means the output tensor is always on CPU. When other operators are executed by GPU, the input tensor of `is_empty` will be on CPU, data transform will happen but the resulted transformed CPU tensor will never be accessed.

panyx0718 (approved):

PR #15514, opened 2019-01-24
Revert conv transpose cudnn

chengduoZH:


dzhwinter:
what's the reason?

dzhwinter (approved):
LGTM

chengduoZH:
TmpAllocator will release the tmpAllocations to Allocation system when `context.Wait()` is called, but for inference, It is time-consuming.
According to @qingqing01 said, This PR induces some inference models's time consume increase from 2.5ms to 181ms.

Issue #15447, opened 2019-01-21
gemm acc issue on AVX512 and AVX


tensor-tang:
code base:
https://github.com/PaddlePaddle/Paddle/pull/15448

# how to reproduce
1. `cmake .. -DCMAKE_BUILD_TYPE=Release -DWITH_GPU=OFF -DWITH_MKL=ON -DWITH_TESTING=ON -DWITH_FLUID_ONLY=ON -DWITH_DOC=OFF -DWITH_MKLDNN=OFF -DWITH_CONTRIB=OFF`
2. `make jit_kernel_test -j`
3. `make test ARGS="-R jit_kernel_test -V"`

This would pass.

But when delete line [here](https://github.com/PaddlePaddle/Paddle/pull/15448/files#diff-278ba0d236f1106f6db54b838ecaab52R495), and do step2 and step3, then it will fail like:

```
...
73: /home/tangjian/paddle-tj-docker/paddle/fluid/operators/jit/test.cc:42: Failure
73: The difference between target[i] and refer[i] is 5.340576171875e-05, which exceeds FLAGS_acc, where
73: target[i] evaluates to -55.43963623046875,
73: refer[i] evaluates to -55.439689636230469, and
73: FLAGS_acc evaluates to 1.0000000000000001e-05.
73: /home/tangjian/paddle-tj-docker/paddle/fluid/operators/jit/test.cc:42: Failure
73: The difference between target[i] and refer[i] is 8.392333984375e-05, which exceeds FLAGS_acc, where
73: target[i] evaluates to 93.255752563476562,
73: refer[i] evaluates to 93.255836486816406, and
73: FLAGS_acc evaluates to 1.0000000000000001e-05.
73: /home/tangjian/paddle-tj-docker/paddle/fluid/operators/jit/test.cc:42: Failure
73: The difference between target[i] and refer[i] is 6.103515625e-05, which exceeds FLAGS_acc, where
73: target[i] evaluates to -67.680465698242188,
73: refer[i] evaluates to -67.680526733398438, and
73: FLAGS_acc evaluates to 1.0000000000000001e-05.
...

```

Failed on both 2620v2 and 5117.

tensor-tang:
This is a very urgent issue. Please help to fix this in high priority @jianhang-liu .
Thanks.

luotao1:
[...]

PR #15515, opened 2019-01-24
jit benchmark use tensor

tensor-tang:
Since tensor has alignment.

test=develop


luotao1 (approved):
> Since tensor has alignment.

Do you mean the memory alignment? Is the part solution of #15447? 
Please add more descriptions in the first comment or create another issue to describe the problem. 
And add comments in the codes, like `To xxx, we use Tensor instead of std::vector<T>, please see issue:`.

tensor-tang:
This is not the solution of the issue, but just we need make the benchmark more credibility since we leant from the #15447 that some MKL env need memory allocated with alignment.

PR #15545, opened 2019-01-28
Fix nccl unittest error in debug mode

sneaxiy:
All nccl unittests fail when compiled with `CMAKE_BUILD_TYPE=Debug`. It is because `PADDLE_ENFORCE` cannot recognize `ncclResult_t` rightly even though the result is equal to `ncclSuccess`.

Also clean up `enforce.h` in this PR.


chengduoZH (dismissed):
LGTM

chengduoZH (approved):

PR #15555, opened 2019-01-28
transpiler.py code clean

seiriosPlus:
1. move ```var structed``` to vars_distributed.py

### DIFF
--- a/python/paddle/fluid/transpiler/details/__init__.py
+++ b/python/paddle/fluid/transpiler/details/__init__.py
@@ -17,3 +17,4 @@ from __future__ import print_function
 from .program_utils import *
 from .ufind import *
 from .checkport import *
+from .vars_distributed import *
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/transpiler/details/__init__.py
```diff
@@ -17,3 +17,4 @@
 from .program_utils import *
 from .ufind import *
 from .checkport import *
+from .vars_distributed import *
```
typhoonzero, line 20: util functions in `io.py` also need same operations?

seiriosPlus: just wait for a second, I am testing it locally.

python/paddle/fluid/transpiler/distribute_transpiler.py
```diff
@@ -39,7 +39,7 @@
 from .. import core, framework, unique_name
 from ..framework import Program, default_main_program, \
     default_startup_program, Block, \
-    Parameter, Variable, grad_var_name
+    Parameter, grad_var_name
 from .details import *
```
typhoonzero: try to avoid import * to reduce namespace conflict

typhoonzero (approved):

PR #15571, opened 2019-01-29
fix ir debug config

Superjomn:
The bug:

- skip the `graph_viz_pass`
- when the `AnalysisConfig::SwitchIrDebug` is called, the config is not updated


fc500110 (approved):
LGTM

PR #15582, opened 2019-01-29
Add the `FLAGS_enable_callstack` to enable the callstack attribute's …

gtvforever:
During the  personal debugging of the PaddlePaddle, i found the new-added `callstack` attribute generate a lot of the call stack info as i don't need it.

So i checked the git log and i found there's a option to disable it but removed in the latest commit.

I made this PR to add the `FLAGS_enable_callstack` option to enable  or disable the callstack attribute output.

If you wannan to get the callstack information, you may need to call os.environ['FLAGS_enable_callstack']='True' firstly.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -642,7 +642,7 @@ class Operator(object):
         if type is None:
             raise ValueError(
                 "`type` to initilized an Operator can not be None.")
-        else:
+        if os.getenv("FLAGS_enable_callstack"):
             callstack_var_name = op_maker.kOpCreationCallstackAttrName()
             op_attrs[callstack_var_name] = list(
                 reversed(traceback.format_stack()))[1:]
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/framework.py
```diff
@@ -642,7 +642,8 @@ def __init__(self,
         if type is None:
             raise ValueError(
                 "`type` to initilized an Operator can not be None.")
-        else:
+
+        if os.getenv("FLAGS_enable_callstack"):
```
wopeizl, line 646: Please don't disable this since http://newicafe.baidu.com/issue/DLTP-124/show?from=page

gtvforever: > Please don't disable this since http://newicafe.baidu.com/issue/DLTP-124/show?from=page

Sorry, i can't open this link. Is it a internal link?
Besides, is it possible to provide an option to disable the callstack detail? 
Those debug-level information will make it's hard to read the program.

PR #15584, opened 2019-01-29
Support imperative learning rate scheduler

velconia:
1. move imperative mnist ut to test_imperative_mnist
2. add optimizer lr scheduler ut into test_imperative_optimizer.py
3. implement lr scheduler in imperative mode


python/paddle/fluid/imperative/learning_rate_scheduler.py
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+
+from __future__ import print_function
+
+from .. import unique_name
+
+__all__ = ['PiecewiseDecay']
+
+
+class LearningRateDecay(object):
```
panyx0718: does this support static graph? why not use existing lr_scheduler?

velconia: static lr and dynamic lr is used via interface in ```learning_rate_scheduler.py``` under the fluid dir

python/paddle/fluid/optimizer.py
```diff
@@ -76,24 +87,43 @@ def get_opti_var_name_list(self):
...
     def _create_global_learning_rate(self):
-        lr = self._global_learning_rate()
-
-        if isinstance(lr, framework.Variable):
-            return
+        if imperative_base.enabled():
+            # create learning rate Variable
+            if isinstance(self._learning_rate, float):
+                self._learning_rate_map[framework.default_main_program(
+                )] = layers.create_global_var(
```
panyx0718: can this be created many times?

python/paddle/fluid/optimizer.py
```diff
@@ -76,24 +87,43 @@ def get_opti_var_name_list(self):
...
+                    name=unique_name.generate("learning_rate"),
+                    shape=[1],
+                    value=float(self._learning_rate),
+                    dtype='float32' if self._dtype is None else self._dtype,
+                    persistable=True)
+            # get learning rate Variable from LearningRateDecay
+            elif isinstance(self._learning_rate, LearningRateDecay):
+                self._learning_rate_map[framework.default_main_program(
```
panyx0718, line 111: why use main program?

python/paddle/fluid/optimizer.py
velconia: no, this should only be called once Optimizer.__init__

python/paddle/fluid/optimizer.py
velconia: to keep the same with static mode code

python/paddle/fluid/optimizer.py
velconia: Done

panyx0718 (approved):

PR #15585, opened 2019-01-29
Some improvements to support bert mixed precision training

kuke:


python/paddle/fluid/layers/nn.py
```diff
@@ -2966,6 +2966,18 @@ def layer_norm(input,
     helper = LayerHelper('layer_norm', **locals())
     dtype = helper.input_dtype()
 
+    # cast to fp32 if using fp16
+    if dtype == core.VarDesc.VarType.FP16:
```
typhoonzero: Should put this in model configurations? Explicit should be better?

kuke: Agreed and made the changes

typhoonzero (approved):
LGTM

PR #15586, opened 2019-01-29
refine bert

tensor-tang:
- add cache to find kernel
- refine vadd kernel choice

----
before:
`native 180.302ms , analysis 224.744ms`

--- 
after
`analysis 159.413ms`

repeat 10 times

analysis提升41%， 优于native 13%


luotao1 (approved):
LGTM

PR #15596, opened 2019-01-30
add device info to tensor

xuezhong:
in order to get place id of tensor with python api


paddle/fluid/pybind/pybind.cc
```diff
@@ -665,6 +667,8 @@ All parameter, weight, gradient are variables in Paddle.
 
   py::class_<platform::Place>(m, "Place")
       .def(py::init<>())
+      .def("is_gpu_place", [](platform::Place &self) { return platform::is_gpu_place(self); })
+      .def("device_id", [](platform::Place &self) { return boost::get<platform::CUDAPlace>(self).device; })
```
sneaxiy: change it to `gpu_device_id`?

paddle/fluid/pybind/pybind.cc
```diff
@@ -287,7 +287,9 @@ PYBIND11_MODULE(core, m) {
       .def("_get_float_element", TensorGetElement<float>)
       .def("_set_double_element", TensorSetElement<double>)
       .def("_get_double_element", TensorGetElement<double>)
-      .def("_dtype", [](Tensor &self) { return self.type(); });
+      .def("_place", [](Tensor &self) { return self.place(); })
+      .def("_dtype", [](Tensor &self) { return ToDataType(self.type()); });
```
sneaxiy: It may be wrong. Returning `self.type()` is ok. `ToDataType` accepts `std::type_info`.

paddle/fluid/pybind/pybind.cc
xuezhong: ok

paddle/fluid/pybind/pybind.cc
xuezhong: ok

PR #15602, opened 2019-01-30
bundle allocator. test=develop

dzhwinter:


paddle/fluid/memory/allocation/buffered_allocator.cc
```diff
@@ -50,29 +57,32 @@ void BufferedAllocator::FreeCache(size_t size) {
 bool BufferedAllocator::IsAllocThreadSafe() const {
   return this->underlying_allocator_->IsAllocThreadSafe();
 }
+
 void BufferedAllocator::Free(Allocation *allocation) {
   platform::LockGuardPtr<std::mutex> guard(mtx_);
   allocations_.emplace(allocation->size(), AllocationPtr(allocation));
+  cache_size_.fetch_add(allocation->size());
```
liupluswei, line 64: why not just cache_size_ += allocation->size()?  since we do not need the previous held value

PR #15605, opened 2019-01-30
Fix bug for lstmp

xuezhong:
1 add cell_clip
2 add proj_clip
3 fix bug of h0
4 change default proj_activation to identity
test=develop


kuke:
Please add `test=develop` in commit message to trig ci

python/paddle/fluid/layers/nn.py
```diff
@@ -770,6 +780,11 @@ def dynamic_lstmp(input,
                               the bias is initialized zero. Default: None.
         use_peepholes(bool): Whether to enable diagonal/peephole connections,
                              default `True`.
+        cell_clip(float): If provided the cell state is clipped
```
kuke: How about another name like `cell_clip_threshold`, and adding an assert `assert(cell_clip_threshold>=0)` ?

paddle/fluid/operators/lstmp_op.h
```diff
@@ -60,16 +94,37 @@ class LSTMPKernel : public framework::OpKernel<T> {
       PADDLE_THROW("unsupported activation type");
   }
 
+  void Print(const Tensor& t, std::string name) const {
```
kuke: Can remove this function？

xuezhong: yes just a function for debug

python/paddle/fluid/layers/nn.py
xuezhong: this parameter uses the same name in API of tensorflow rnn cell, so the name may be more convenient to user.

xuezhong: can add assert(cell_clip >=0)

xuezhong:
> Please add `test=develop` in commit message to trig ci

ok

kuke (dismissed):
LGTM

panyx0718 (approved):

kuke (approved):

PR #15611, opened 2019-01-30
To make CUDA_LAUNCH_KERNEL_HELPER support large size.

guoshengCS:
To make CUDA_LAUNCH_KERNEL_HELPER support large size. Otherwise, [beam_search_op](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/beam_search.cu#L361) with `beam_size * num_seqs * 32 > 256` might failed.


kuke (approved):
LGTM

PR #15619, opened 2019-01-31
Fix unittest failure in debug mode

sneaxiy:


python/paddle/fluid/tests/unittests/test_eager_deletion_transformer.py
```diff
@@ -15,13 +15,12 @@
 import os
 import unittest
 os.environ['FLAGS_eager_delete_tensor_gb'] = "0.0"
+# os.environ['FLAGS_fast_eager_deletion_mode'] = "1"
```
chengduoZH: If line18 is unnecessary, it should be removed.

sneaxiy: Done.

chengduoZH (approved):
LGTM

PR #15624, opened 2019-01-31
add skip send.recv in memory optimize

dzhwinter:
add skip send.recv


paddle/fluid/framework/details/analysis_var_pass.cc
```diff
@@ -247,20 +246,21 @@ void AnalysisVarPass::SubGraphOptimize(OpDesc* op_desc) const {
   }
 }
 
-std::unordered_set<std::string> AnalysisVarPass::GetSubBlockVars(
+void AnalysisVarPass::CollectSkipSet(
```
jacquesqiao, line 249: 建议保留 GetSubBlockVars，加入 GetCommunicateVars
```cpp
std::unordered_set<std::string> CollectSkipSet() {
  GetSubBlockVars();
  GetCommunicateVars();
}
```

Issue #15622, opened 2019-01-31
paddle::operators::jit::gen::SeqPoolJitCode needs an aligned allocator


GBuella:
The standard new operator is used with an overaligned type:
```
paddle/fluid/operators/jit/gen/seqpool.cc:74:60:   required from here
paddle/fluid/operators/jit/registry.h:33:29: error: ‘new’ of type ‘paddle::operators::jit::gen::SeqPoolJitCode’ with extended alignment 32 [-Werror=aligned-new=]
   return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
Obviously this can result in runtime errors when using certain AVX instructions, if the allocated SeqPoolJitCode instance is misaligned.

tensor-tang:
Thanks for reporting issues. 

But could you give some more reproduce infos, it seems some compile issue?

tensor-tang:
maybe your gcc version? we only tried below gcc5.4.

> Obviously this can result in runtime errors when using certain AVX instructions

BTW, this jitcode instance is only an instance that generate code, the code buffer would be allocated by xbyak with alignment, so I did not get your point about the runtime error if I understand right.

GBuella:
This issue is independent of compilers. The warning comes from GCC 7.3 BTW.
The offending allocation is at:
https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/jit/gen/seqpool.h#L196

The 32 char aligned data member was added in this commit:
https://github.com/PaddlePaddle/Paddle/commit/123b98f417d064e780412f316f4ca43988f4d0d2

As I look at it now, it is only a single float variable that is "aligned", so that seems rather pointless.
I think it would make sense to just remove those ALIGN32* macros from `seqpool.h`.

tensor-tang:
OK, thanks, now I understand your issue.

Firstly, this new operator in this `SeqPoolJitCode` instance itself would not result in runtime error using AVX. 

Secondly, since I used new operator, then this ALIGEN code would cause high level GCC e.g. GCC 7.3 (which warning check more strictly with `-Werror=aligned-new`) failed.
https://github.com/PaddlePaddle/Paddle/blob/546eefae6fdf6b9848b91c8246b17a0fa72cc982/paddle/fluid/operators/jit/gen/seqpool.h#L196


Maybe you could replace here 

[...]

PR #15625, opened 2019-01-31
Remove superfluous alignment requirement from seqpool.h

GBuella:
See #15622

### DIFF
--- a/paddle/fluid/operators/jit/gen/seqpool.h
+++ b/paddle/fluid/operators/jit/gen/seqpool.h
@@ -193,7 +193,7 @@ class SeqPoolJitCode : public JitCode {
   }
 
  private:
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/jit/gen/seqpool.h
```diff
@@ -193,7 +193,7 @@ class SeqPoolJitCode : public JitCode {
   }
 
  private:
-  float ALIGN32_BEG fp_h_[1] ALIGN32_END;
+  float fp_h_[1];
```
tensor-tang, line 196: no need remove this since #15626

GBuella: I still have the question, what is this alignment for? What does it achieve here?

PR #15626, opened 2019-01-31
fix aligned-new error in jitkernel

### LINKED ISSUE #15622: paddle::operators::jit::gen::SeqPoolJitCode needs an aligned allocator

tensor-tang:
fix #15622 

related #15625

### DIFF
--- a/paddle/fluid/operators/jit/gen_base.cc
+++ b/paddle/fluid/operators/jit/gen_base.cc
@@ -17,7 +17,13 @@
 #include <iostream>
 #include <sstream>
 #include <vector>
+#include "paddle/fluid/memory/allocation/cpu_allocator.h"  // for posix_memalign
 #include "paddle/fluid/platform/cpu_info.h"
+#include "paddle/fluid/platform/enforce.h"
+
+#ifndef _WIN32
+#define posix_memalign_free free
+#endif
 
 DEFINE_bool(dump_jitcode, false, "Whether to dump the jitcode to file");
 
@@ -39,6 +45,17 @@ void GenBase::dumpCode(const unsigned char* code) const {
     }
   }
 }
+
+void* GenBase::operator new(size_t size) {
+  void* ptr;
+  constexpr size_t alignment = 32ul;
+  PADDLE_ENFORCE_EQ(posix_memalign(&ptr, alignment, size), 0,
+                    "GenBase Alloc %ld error!", size);
+  PADDLE_ENFORCE(ptr, "Fail to allocate GenBase CPU memory: size = %d .", size);
+  return ptr;
+}
+
+void GenBase::operator delete(void* ptr) { posix_memalign_free(ptr); }
 
 std::vector<int> packed_groups(int n, int k, int* block_out, int* rest_out) {
   int block;
--- a/paddle/fluid/operators/jit/registry.h
+++ b/paddle/fluid/operators/jit/registry.h
@@ -17,6 +17,7 @@
 #include <memory>
 #include <tuple>
 #include <type_traits>
+#include "paddle/fluid/memory/allocation/cpu_allocator.h"  // for posix_memalign
 #include "paddle/fluid/operators/jit/kernel_base.h"
 #include "paddle/fluid/operators/jit/kernel_pool.h"
 #include "paddle/fluid/platform/place.h"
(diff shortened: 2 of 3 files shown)


tensor-tang:
@GBuella please help to review

GBuella:
Right now I see two problems with this:

1.
This function is named `make_unique`, it is obviously a substitue for `std::make_unique`(which is missing from C++11). As long as it has this name, I think it should follow the expected `make_unique` behavior.
If/when this project is updated to use C++14, one natural step of such update is removing the substitute `make_unique` implementation, and using `std::make_unique` instead, but then suddenly the standard implementation would seem broken, as that one doesn't have this "fix"...

https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

2.
It only fixes this particular instance. Using this approach would require taking care of all allocations, e.g.`std::vector` might just use an anilgned allocator.
```
struct some_type {
  alignas(32) float[16];
};

std::vector<some_type> v(7);
perform_avx_computations(v); // <- this could fail, as the data in v might be misaligned
```

GBuella:
By the way, I already see an aligned_allocator under:
https://github.com/PaddlePaddle/Paddle/tree/develop/paddle/fluid/memory/allocation

Anyways, the best way to do it I believe, is to add the appropriate operators to such classes:

```
struct some_type {
  alignas(32) float[16];

  static void* operator new(size_t);
  {
     return ... // return an aligned pointer
  }

  static void* operator new[](size_t);
  {
     return ... // return an aligned pointer
  }

  static void operator delete(void* pointer)
  {
      //delete these...
  }

    static void operator delete[](void* pointer)
  {
      //delete these...
  }
};

std::vector<some_type> v(7);
perform_avx_computations(v); // <- the data in v should be properly aligned
```

paddle/fluid/operators/jit/registry.h
```diff
@@ -30,7 +31,12 @@ namespace jit {
 template <typename T, typename... Args>
 inline std::unique_ptr<T> make_unique(Args&&... args) {
   static_assert(!std::is_array<T>::value, "T must not be array");
-  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+  void* ptr;
+  constexpr size_t alignment = 64;
+  PADDLE_ENFORCE_EQ(posix_memalign(&ptr, alignment, sizeof(T)), 0,
```
GBuella: `PADDLE_ENFORCE_EQ(posix_memalign(&ptr, alignof(T), sizeof(T)), 0,`

tensor-tang:
1. is a good point, that name should be inappropriate, we could find more appropriate one.

2. maybe this is not a big problem, this kernel do not accept been used like your way, and the alignment could only be something like 32, 64 or 4096. we can just chose 4096 like mkldnn did. but here i use 64 since it's not really needed of 4096 now.

GBuella:
"this kernel do not accept been used like your way"
Excuse me, what do you mean by "this kernel"?

tensor-tang:
I mean this class `SeqPoolJitcode` which is `some_type` in your example
```
std::vector<some_type> v(7); perform_avx_computations(v);
```
we will not call avx like this example, but,
as you said, if some_type is any type with inside alignment then the better way should be override new operator.

Anyway, l will make this class more general.

paddle/fluid/operators/jit/gen_base.cc
```diff
@@ -40,6 +46,17 @@ void GenBase::dumpCode(const unsigned char* code) const {
   }
 }
 
+void* GenBase::operator new(size_t size) {
```
GBuella, line 49: Yes, but once again, are these needed?
As you said, no such AVX instructions are used with these classes, thus it would make sense to just remove the alignment requirement.

tensor-tang: It would be possible that I may create some protected members which needed alignment, such as change this https://github.com/PaddlePaddle/Paddle/blob/c356bd01e9dffa4ff2af8285b9175363d87d6083/paddle/fluid/operators/jit/gen/act.cc#L24
 to class member in later plan.

This code is writing in hurry so just used as  global variables. We may polish them later, then this patch is definitely needed as your raised the issue #15622  which is very useful.

> As you said, no such AVX instructions are used with these classes

BTW, I am not saying no AVX used in these classes, I am saying it would not be used like the way, such as `some_avx_instrction(these classes)`.

GBuella:
No need for this since #15625
Unless that alignment specification is somehow really useful in `seqpool.h`.

tensor-tang:
As I said here https://github.com/PaddlePaddle/Paddle/pull/15626#discussion_r253000863, #15622 is not the only reason of this pr.

luotao1 (approved):
LGTM

PR #15639, opened 2019-02-01
Revert "Async double buffered py reader"

qingqing01:
Reverts PaddlePaddle/Paddle#15296

Since failed CE http://ce.paddlepaddle.org:8080/viewLog.html?buildId=21819&buildTypeId=PaddleModesl_BuildModels&tab=buildLog


cjld:
Does the tolerance of the transformer is too strict?
```
test_cost_card4_kpi = CostKpi('test_cost_card4', 0.005, 0, actived=True)
```
current error is 0.01

cjld (approved):
No idea. Don't know why this affects the transformer. I need further digging.

PR #15640, opened 2019-02-01
Remove ALIGN32_BEG, ALIGN32_END macros

GBuella:
The `alignas` keyword is part of ISO C++11
I'm not sure how old VC++ versions support it though.


wojtuss:
@GBuella , clang-format complaints about the code and two unit tests fail (test_analyzer and test_analysis_predictor).

PR #15642, opened 2019-02-01
Add ngraph sum, sigmoid, relu_grad and tanh_grad op

baojun-nervana:


python/paddle/fluid/tests/unittests/ngraph/test_activation_ngraph_op.py
```diff
@@ -18,17 +18,7 @@
...
-
-
-class TestNGRAPHReluDim2(TestRelu):
-    def setUp(self):
-        super(TestNGRAPHReluDim2, self).setUp()
-
-
-class TestNGRAPHTanhDim2(TestTanh):
-    def setUp(self):
-        super(TestNGRAPHTanhDim2, self).setUp()
```
tensor-tang, line 31: where is test sigmoid?

python/paddle/fluid/tests/unittests/ngraph/test_sum_ngraph_op.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import print_function
+import unittest
+from paddle.fluid.tests.unittests.test_sum_op import TestSumOp, TestSelectedRowsSumOp, TestLoDTensorAndSelectedRowsOp
+
```
tensor-tang, line 17: need add ngraph tests?

baojun-nervana: The testcase is imported, it will be picked up by unittest. Is that correct? The difference of this test from the original one is it will run by ngraph (trigger set in cmake)

python/paddle/fluid/tests/unittests/ngraph/test_activation_ngraph_op.py
baojun-nervana: The TestSigmoid is imported

tensor-tang: seems we have `TestNGRAPHReluDim4` and `TestNGRAPHTanhDim4`, but no sigmoid one? I thought you may forget one. You can considerate adding one in next PR.

tensor-tang (approved):
LGTM

python/paddle/fluid/tests/unittests/ngraph/test_activation_ngraph_op.py
baojun-nervana: The TestSigmoid test case was imported. It will run as part of the unittest through ngraph. It wasn't derived to another testcase as no change is needed from the base except it will run through ngraph.
https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/ngraph/test_activation_ngraph_op.py#L21

PR #15649, opened 2019-02-02
fix ctr reader read svm data

jacquesqiao:


paddle/fluid/operators/reader/ctr_reader.cc
```diff
@@ -213,7 +213,7 @@ void ReadSvmData(const DataDesc& data_desc, std::shared_ptr<Reader> reader,
       framework::LoD lod{lod_data};
       lod_tensor.set_lod(lod);
       int64_t* tensor_data = lod_tensor.mutable_data<int64_t>(
```
seiriosPlus, line 215: ```auto``` is better?

seiriosPlus (approved):
LGTM

PR #15659, opened 2019-02-05
Tests - add some missing to_string calls

GBuella:
E.g.:

```
/home/tej/code/gbuella_paddle/paddle/fluid/framework/ir/seqpool_concat_fuse_pass_tester.cc:167:40: error: adding 'int' to a string does not append to the string [-Werror,-Wstring-plus-int]
    std::string prefix = "seqpool_op_" + i;
                         ~~~~~~~~~~~~~~^~~
/home/tej/code/gbuella_paddle/paddle/fluid/framework/ir/seqpool_concat_fuse_pass_tester.cc:167:40: note: use array indexing to silence this warning
    std::string prefix = "seqpool_op_" + i;
                                       ^
                         &             [  ]
1 error generated.
```


luotao1:
Could you tell us how to reproduce this error? Is it caused by GCC 7.3?

tensor-tang (approved):
LGTM

GBuella:
> Could you tell us how to reproduce this error? Is it caused by GCC 7.3?

I'm not sure if GCC has this particular warning. With clang, it is the result of the `-Wstring-plus-int` flag.

PR #15661, opened 2019-02-07
cpu reduce mode did not need to broadcast params test=develop

jacquesqiao:


python/paddle/fluid/compiler.py
```diff
@@ -110,6 +120,8 @@ def with_data_parallel(self,
             self._exec_strategy = ExecutionStrategy()
         if self._build_strategy is None:
             self._build_strategy = BuildStrategy()
+        self._build_strategy.is_distribution = _is_pserver_mode(
```
panyx0718: what is `is_distribution` strategy?

panyx0718: @chengduoZH

jacquesqiao: means it use parameter server, will have send/recv in program

paddle/fluid/framework/details/build_strategy.cc
```diff
@@ -133,12 +133,15 @@ class ParallelExecutorPassBuilder : public ir::PassBuilder {
...
     if (strategy_.is_distribution_) {
+      VLOG(3) << "dist train mode";
       multi_devices_pass = AppendPass("dist_multi_devices_pass").get();
     } else {
       if (strategy.reduce_ == BuildStrategy::ReduceStrategy::kAllReduce) {
+        VLOG(3) << "allreduce mode";
         multi_devices_pass =
             AppendPass("allreduce_mode_multi_devices_pass").get();
       } else if (strategy.reduce_ == BuildStrategy::ReduceStrategy::kReduce) {
+        VLOG(3) << "reduce mode";
```
chengduoZH: This log is too simple, others may not know what it means if just look at the output logs.

paddle/fluid/framework/details/multi_devices_graph_pass.cc
```diff
@@ -925,9 +925,7 @@ void DistSSAGraphBuilder::InsertCollectiveOp(ir::Graph *result,
 }
 
 void DistSSAGraphBuilder::InsertPostprocessOps(ir::Graph *result) const {
-  if (need_broadcast_var_ ||
-      (UseGPU() &&
-       strategy_.reduce_ == BuildStrategy::ReduceStrategy::kReduce)) {
+  if (UseGPU() && strategy_.reduce_ == BuildStrategy::ReduceStrategy::kReduce) {
```
chengduoZH, line 928: If `need_broadcast_var_` is not useful, it should be removed exactly.

python/paddle/fluid/compiler.py
chengduoZH: It should be `is_distributed`.

chengduoZH: It should be `is_distributed`.

typhoonzero: maybe `or self._build_strategy.num_trainers > 1` should not be added here. `is_distribution` is only for pserver mode.

paddle/fluid/framework/details/multi_devices_graph_pass.cc
jacquesqiao: done

python/paddle/fluid/compiler.py
jacquesqiao: removed

typhoonzero (approved):
LGTM

python/paddle/fluid/compiler.py
```diff
@@ -34,6 +35,15 @@ def _place_obj(place):
     return p
 
 
+def _is_pserver_mode(main_program):
+    main = main_program if main_program \
+        else framework.default_main_program()
+    for op in main.global_block().ops:
+        if op.type in ["send", "recv"]:
```
panyx0718, line 42: this is very fragile. Can you use a better solution? tag program in transpiler?

python/paddle/fluid/compiler.py
panyx0718: `is_distribution` is not a build strategy, maybe should get rid of it.

python/paddle/fluid/compiler.py
```diff
@@ -110,6 +120,7 @@ def with_data_parallel(self,
             self._exec_strategy = ExecutionStrategy()
         if self._build_strategy is None:
             self._build_strategy = BuildStrategy()
+        self._build_strategy.is_distribution = _is_pserver_mode(self._program)
```
panyx0718, line 123: this should be reverted. pserver mode is not equal to is_distribution

panyx0718:
Can we revert this PR and redo review?

PR #15662, opened 2019-02-08
fea/pile allocator

Superjomn:
# Pile Allocator
This is a buddy-allocator based high-level memory allocator for both CPU and GPU memory.
Beside a normal memory allocator, it offers the ability to save memory consumption in some special scenarios:

- In inference, several models are executed in sequential, it can help to reuse the memory for temporary variables,
- In training, it can also squeeze the memory used by sequential models execution.

The example scenarios:
 
We have `N` models, each has an average memory size `W` for weight and `T` for temporary variables, and all these models are used in sequential. By default, these models will take `N*(W+T)` size of memory.

By using PileAllocator, we can share the temporary variable memory space for all the models, and in total these model will only take `N*W+T` size of memory. That effect is remarkable when `T` is large.

TODOs:

- TODO add performance benchmark
- TODO re-consider the overall allocator interface, it seems that the design of `Allocation` as the return type add the overhead, and make it less flexible.

### DIFF
--- a/paddle/fluid/memory/allocation/pile_allocator.h
+++ b/paddle/fluid/memory/allocation/pile_allocator.h
@@ -116,21 +116,38 @@ struct BuddyManager {
   // to store this integer to keep memory aligned by 8.
   const uint32_t kHeaderSize{sizeof(size_t)};
 
-  BuddyManager(size_t min_mem_log_size, size_t max_mem_log_size)
+  BuddyManager(size_t min_mem_log_size, size_t max_mem_log_size,
+               size_t realloc_mem_log_size, bool allow_realloc = true)
       : min_mem_log_size_(min_mem_log_size),
         max_mem_log_size_(max_mem_log_size),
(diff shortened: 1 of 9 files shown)


paddle/fluid/memory/allocation/pile_allocator.h
```diff
@@ -0,0 +1,568 @@
...
+ * The same underlying memory pool can be reused that, each buddy system can
+ * allcate memory based on the same memory pool, regardless of others, so the
+ * same memory block can be allcated by more than one buddy systems, on
+ * condition that the buddy systems are used sequentially and individually.
+ */
+struct BuddyResource {
+  using byte_t = uint8_t;
+  using bucket_t = StackSet<byte_t*>;
+
+  BuddyResource(int max_mem_log_size, int min_mem_log_size,
```
NHZlX, line 126: need comments for `max_mem_log_size, min_mem_log_size, realloc_mem_log_size`

PR #15667, opened 2019-02-11
Fix decorator signature error

sneaxiy:
Python decorator would change the signature of the wrapped functions, so that the signature of the wrapped functions would have `*args` and `**kwargs`. Use the method described [here](https://stackoverflow.com/questions/48746567/python-decorator-to-keep-signature-and-user-defined-attribute) to fix.


python/paddle/fluid/wrapped_decorator.py
```diff
@@ -0,0 +1,30 @@
...
+def wrap_decorator(decorator_func):
+    @decorator.decorator
+    def __impl__(func, *args, **kwargs):
+        wrapped_func = decorator_func(func)
+        return wrapped_func(*args, **kwargs)
+
+    return __impl__
+
+
+contextmanager = wrap_decorator(contextlib.contextmanager)
```
panyx0718: call this `signature_safe_contextmanager` to avoid confusion with `contextlib.contextmanager`?

sneaxiy: Done.

panyx0718 (approved):

PR #15678, opened 2019-02-12
change default option related to softmax, test=develop

jerrywgz:


panyx0718 (commented):
why make this change?

jerrywgz:
> why make this change?

softmax_with_cross_entropy may cause NaN in models and it can be solved by setting use_cudnn to False

python/paddle/fluid/layers/nn.py
```diff
@@ -1734,7 +1734,7 @@ def sequence_softmax(input, use_cudnn=False, name=None):
     return softmax_out
 
 
-def softmax(input, use_cudnn=True, name=None):
+def softmax(input, use_cudnn=False, name=None):
```
panyx0718, line 1737: add some comments to explain the difference between true or false?

jerrywgz: Done

panyx0718 (approved):

PR #15687, opened 2019-02-13
fix enforce

sneaxiy:
Previously, `__PADDLE_BINARY_COMPARE` would calculate expressions of `__VAL0` and `__VAL1` twice when exception raises, which may cause undefined behavior sometimes (for example, if `__VAL0` is some expressions of `pthread_unlock`).


chengduoZH (dismissed):
LGTM

chengduoZH (approved):
LGTM

PR #15691, opened 2019-02-13
fix generate doc error in activation ops

luotao1:
- fix generate doc error in several activation ops
- fix the wrong formula in sigmoid, soft sign, ceil, floor
- before:
image
http://www.paddlepaddle.org/documentation/docs/en/1.2/api/layers.html#permalink-184-sigmoid
- after:
image
image
image
image


shanyi15 (approved):
LGTM

PR #15692, opened 2019-02-13
Feature/add memory profile

FrostML:
profiler, memory monitor


paddle/fluid/memory/allocation/legacy_allocator.cc
```diff
@@ -328,19 +329,23 @@ size_t Usage::operator()(const platform::CUDAPinnedPlace &cuda_pinned) const {
...
-
 LegacyMemMonitor GPUMemMonitor;
 
 Allocation *LegacyAllocator::AllocateImpl(size_t size, Allocator::Attr attr) {
   void *ptr = boost::apply_visitor(legacy::AllocVisitor(size), place_);
-  return new Allocation(ptr, size, place_);
+  Allocation *tmp_alloc = new Allocation(ptr, size, place_);
+  platform::RecordMemEvent tmp_record;
+  record_mem.insert({tmp_alloc, tmp_record});
+  record_mem[tmp_alloc].InitRecordMem(size, place_);
```
cjld: Is it thread safe?
Would it be better if the `RecordMemEvent` is a member of `Allocation`, so you do not need to manage the life cycle of `RecordMemEvent`, and `unordered_set` is slow, This profiler may disturb the original process.

paddle/fluid/memory/allocation/legacy_allocator.cc
```diff
@@ -328,19 +329,26 @@ size_t Usage::operator()(const platform::CUDAPinnedPlace &cuda_pinned) const {
 }  // namespace legacy
 
 namespace allocation {
-
 LegacyMemMonitor GPUMemMonitor;
+std::mutex mem_profiler;
 
 Allocation *LegacyAllocator::AllocateImpl(size_t size, Allocator::Attr attr) {
+  std::lock_guard<std::mutex> l(mem_profiler);
```
chengduoZH: This lock maybe makes `AllocateImpl` slower. 
You should test the performance influence in CI and CE.

paddle/fluid/memory/allocation/legacy_allocator.cc
```diff
@@ -328,19 +329,26 @@ size_t Usage::operator()(const platform::CUDAPinnedPlace &cuda_pinned) const {
 }  // namespace legacy
 
 namespace allocation {
-
 LegacyMemMonitor GPUMemMonitor;
+std::mutex mem_profiler;
```
chengduoZH: Using global variables here is a bad design.

paddle/fluid/memory/allocation/legacy_allocator.cc
```diff
@@ -328,19 +329,26 @@ size_t Usage::operator()(const platform::CUDAPinnedPlace &cuda_pinned) const {
...
-  return new Allocation(ptr, size, place_);
+  Allocation *tmp_alloc = new Allocation(ptr, size, place_);
+  platform::RecordMemEvent tmp_record;
+  record_mem.insert({tmp_alloc, tmp_record});
+  record_mem[tmp_alloc].InitRecordMem(size, place_);
+  return tmp_alloc;
 }
 
 void LegacyAllocator::Free(Allocation *allocation) {
+  std::lock_guard<std::mutex> l(mem_profiler);
```
chengduoZH: This lock maybe makes `Free` slower.

paddle/fluid/memory/allocation/legacy_allocator.cc
chengduoZH: `RecordMemEvent` should not be a member of `Allocation`, because they are different modules.

cjld: How about this:
```
void LegacyAllocator::AllocateImpl(...) {
  Allocation *tmp_alloc = new Allocation(ptr, size, place_);
  RecordMemAlloc(tmp_alloc->size, place_);
  return tmp_alloc;
}

void LegacyAllocator::Free(...) {
  RecordMemFree(allocation->size, place_);
  delete allocation;
}

void RecordMemAlloc(size, place) {
  // lock is not needed
  start_ns_ = PosixInNsec();
  GetMemEventList().Record(MemEventType::Alloc, start_ns_, bytes, place, g_mem_thread_id);
  tracer->AddMemAllocRecord(start_ns_, bytes_, place_, g_mem_thread_id);
}


void RecordMemFree(size, place) {
  // lock is not needed
  end_ns_ = PosixInNsec();
  PushMemEvent(end_ns_, bytes_, place_);
  GetMemEventList().Record(MemEventType::Free, end_ns, bytes, place, g_mem_thread_id);
  tracer->AddMemFreeRecord(end_ns_, bytes_, place_, g_mem_thread_id);
}
```
It is thread safe and faster. unordered_set is too heavy.

PR #15696, opened 2019-02-13
cherry picked modifies.

dzhwinter:
1. fix the default config in parallelexecutor and compiler.
2. fix the reuse condition, skip the small loss variables, avoid reuse losses. 
3. polish code for reading.


JiabinYang (approved):
LGTM

PR #15710, opened 2019-02-13
Add adam op for ngraph engine

baojun-nervana:


baojun-nervana:
@tensor-tang can you check this as well?

tensor-tang (approved):
LGTM

PR #15712, opened 2019-02-14
disable DAM test in MAC

Superjomn:
test=develop


luotao1:
How about only disable `compare_with_dynamic_memory_optim` and `compare_with_static_memory_optim` of `analyzer_dam_tester.cc` in MAC?

Issue #15708, opened 2019-02-13
Add ngraph related unit-tests


luotao1:
To improve the code (line+function) coverage, please **add ngraph related unit-tests**:
http://ce.paddlepaddle.org:8080/viewLog.html?buildId=25255&buildTypeId=PaddlePaddleFramework_TestCoverage_TestCoverage&tab=report_project15_C___Coverage
- operators/ngraph/ops
![image](https://user-images.githubusercontent.com/6836917/52716182-324fa580-2fd9-11e9-9a42-4ca09643aad3.png)
![image](https://user-images.githubusercontent.com/6836917/52716213-40052b00-2fd9-11e9-89b9-9f03e68ae6e6.png)
![image](https://user-images.githubusercontent.com/6836917/52716235-4d221a00-2fd9-11e9-917b-3462289968a5.png)

mozga-intel:
@luotao1 I will have a look at this, thank you.

PR #15721, opened 2019-02-15
Add ngraph op coverage

baojun-nervana:
https://github.com/PaddlePaddle/Paddle/issues/15708

CC @mozga-intel


python/paddle/fluid/tests/unittests/ngraph/test_pool2d_ngraph_op.py
```diff
@@ -13,62 +13,247 @@
...
+from paddle.fluid.tests.unittests.op_test import OpTest
 
-class TestNGRAPHPool2D_Op(TestPool2D_Op):
-    def setUp(self):
-        super(TestNGRAPHPool2D_Op, self).setUp()
-        self._cpu_only = True
 
-    def init_test_case(self):
-        super(TestNGRAPHPool2D_Op, self).init_test_case()
+def adaptive_start_index(index, input_size, output_size):
```
luotao1: 1. Why not use `super(TestNGRAPHxxx_Op, self).setUp()`?
2. Many codes in `test_pool2d_ngraph_op.py` are duplicated in `test_pool2d_op.py`. For simplify, you can refer to `test_pool2d_int8_mkldnn_op.py`. Same for the other three test files.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -374,6 +375,9 @@ def _get_places(self):
                 return []
         places = [fluid.CPUPlace()]
         cpu_only = self._cpu_only if hasattr(self, '_cpu_only') else False
+        use_ngraph = bool(os.getenv("FLAGS_use_ngraph", False))
+        if use_ngraph:
+            cpu_only = True
```
luotao1, line 380: If `cpu_only=True` here, do you remove all the `self._cpu_only = True` in `test_xxx_ngraph_op.py`? Such as https://github.com/PaddlePaddle/Paddle/blob/4da291c6a38f532302b4996b1a078dfa11f9bc7e/python/paddle/fluid/tests/unittests/ngraph/test_mean_ngraph_op.py#L23

python/paddle/fluid/tests/unittests/ngraph/test_conv2d_ngraph_op.py
```diff
@@ -17,60 +17,5 @@
...
-        self._cpu_only = True
-
-    def init_kernel_type(self):
-        super(TestNGRAPHWith1x1, self).init_kernel_type()
-
-
-class TestNGRAPHWithInput1x1Filter1x1(TestWithInput1x1Filter1x1):
-    def setUp(self):
-        super(TestNGRAPHWithInput1x1Filter1x1, self).setUp()
-        self._cpu_only = True
```
luotao1, line 69: Why do you remove all the unit-tests? remove `self._cpu_only = True` is enough?

baojun-nervana: Here is my understanding. The testcases are imported to the file and they will be picked up by unittest. https://github.com/PaddlePaddle/Paddle/blob/b010354b12126dcbf3e0878380d7824970ae84f5/python/paddle/fluid/tests/unittests/ngraph/test_conv2d_ngraph_op.py#L18

Those testcases of TestConv2dOp, TestWithPad, TestWithStride, TestWithGroup, TestWith1x1, and TestWithInput1x1Filter1x1 will run and they will run through ngraph. 

The derived testcase has no change except it is set as cpu_only. Since this cpu_only setting is no longer needed, the test case is not needed.

luotao1 (approved):

PR #15725, opened 2019-02-15
Correct the doc in Python API

### LINKED ISSUE #15208: layers  op文档问题
在对1,2英文文档走查后，发现layers op类别的文档出现了下面的问题：

bug

以下是具有上图相同问题的api：

|同类问题api列表|
|-----|
abs 
ceil
cos
exp
floor
logsigmoid
reciprocal
round
sigmoid
sin
softplus
softsign
sqrt
square
tanh
tanh_shrink

Xreki:
Fix #15208 

- DynamicRNN
  - Before:
image
image

  - After:
image
image

- equal
  - Before:
image

  - After:
image

- abs etc.
  - After:
image


Xreki:
Review comment from @panyx0718 , `use_mkldnn` and `is_test` should not be configured by users.

Change to:
image

python/paddle/fluid/framework.py
```diff
@@ -555,7 +555,10 @@ def generated_op_attr_names():
         return {
             core.op_proto_and_checker_maker.kOpRoleAttrName(),
             core.op_proto_and_checker_maker.kOpRoleVarAttrName(),
-            core.op_proto_and_checker_maker.kOpNameScopeAttrName()
+            core.op_proto_and_checker_maker.kOpNameScopeAttrName(),
+            core.op_proto_and_checker_maker.kOpCreationCallstackAttrName(),
+            core.op_proto_and_checker_maker.kOpUseMKLDNNAttrName(),
```
panyx0718: why put this here? not every op has mkldnn attr?

paddle/fluid/framework/op_proto_maker.h
```diff
@@ -48,6 +48,8 @@ class OpProtoAndCheckerMaker {
   static const char *OpRoleVarAttrName() { return "op_role_var"; }
   static const char *OpNamescopeAttrName() { return "op_namescope"; }
   static const char *OpCreationCallstackAttrName() { return "op_callstack"; }
+  static const char *OpUseMKLDNNAttrName() { return "use_mkldnn"; }
```
panyx0718: remove? why put it here? not every op has mkldnn kernel?

python/paddle/fluid/framework.py
Xreki: It is used to skip some attrs when generating doc string for op, and this function `generated_op_attr_names` is only used here.

https://github.com/PaddlePaddle/Paddle/blob/fc87ef741b7e6dc38a3100d4c8d6b9170863efac/python/paddle/fluid/layers/layer_function_generator.py#L91

panyx0718 (approved):

PR #15736, opened 2019-02-15
fix test_transpiler random fail

typhoonzero:
fix random fails on mac, like http://ci.paddlepaddle.org/viewLog.html?buildId=58799&buildTypeId=PaddleMac_MacPrCiPython35&tab=buildLog


seiriosPlus (approved):
LGTM!

PR #15752, opened 2019-02-17
Added resnet script for ngraph benchmark

baojun-nervana:
Will follow up on README


tensor-tang:
@baojun-nervana when it's ready to review, please let us know.

baojun-nervana:
@tensor-tang We need PR https://github.com/PaddlePaddle/Paddle/pull/15925 to get better performance. Can you help to review and get the PR merged?

@mozga-intel will provide a readme on the instruction and benchmark data.

tensor-tang:
Since it's the standard Resnet50 model, maybe you could just refer to https://github.com/PaddlePaddle/models/blob/develop/fluid/PaddleCV/image_classification/models/resnet.py and add flag `use_graph=True` to trigger NGraph.

luotao1:
@tensor-tang How about add README in models repo?

tensor-tang:
> How about add README in models repo?

Agreed. But which folder you suggest.

luotao1:
@tensor-tang How about in https://github.com/PaddlePaddle/models/tree/develop/fluid/PaddleCV/image_classification? And add some ngraph info in READEME.md?

tensor-tang:
@baojun-nervana

Discussed with @luotao1 , we think you should be add README_ngraph.md 
 in https://github.com/PaddlePaddle/models/blob/develop/fluid/PaddleCV/image_classification.


As for the scripts, we thought it should be only one flag needed be added in the https://github.com/PaddlePaddle/models/blob/develop/fluid/PaddleCV/image_classification/run.sh or  add few changes in the python scripts.

luotao1:
@tensor-tang @baojun-nervana 
1. I think README_ngraph.md and README_ngraph_cn.md is better.
2. I think that https://github.com/PaddlePaddle/models/blob/develop/fluid/PaddleCV/image_classification/run.sh and python scripts could be the same as before, you could write in README_ngraph.md to tell how to use ngraph, such as `FLAGS_ngraph=true ./run.sh`

mozga-intel:
@luotao1 and @tensor-tang 
I don't want to dwell on this subject. I suppose that it what is so-called the image_classification script is a bit different than the script stored in the model repository. It has a few additional parameters checking the performance after each iteration, the script gives us opportunity to run it with dummy_data and so on.  But if you want we can undertake the job of put these script to the repository dedicated for it.

tensor-tang:
Thanks @mozga-intel 

That's two things:
1. Enabling ngraph and running with resnet50 just in the models repo is an independent job. And maybe you can compare with `use_mkldnn=True` for performance comparison. Then the comparison is fair.
2. If you think the script is valuable like "gives us opportunity to run it with dummy_data", of course, we are appreciate and it cloud be added to models repo. But, it's another story like you said. For example, we may need guys from model's team to review if that's necessary and we do not think we have enough time to do this before release1.4.

Hope you cloud understand.
Any suggestions are welcome.

baojun-nervana:
@tensor-tang and @luotao1 For the resnet model which I refer to the following file ([link](https://github.com/PaddlePaddle/Paddle/blob/b25252ae37ff09ffda69ec2e508a059c707ff01a/benchmark/fluid/ngraph/script/models/resnet.py.)), we can use what it has in the models repo. We may add a readme and a link to the model). This way I assume we can avoid to process of model review from another team.

As fort the train script, it is preferred to have a separate one as we need some modification as aforementioned. Also the script in the model repo uses parallel executor and we don't support yet. For a fair comparison, we will compare against mkldnn with the same script which is what we are working on with our benchmarking team.

baojun-nervana:
@tensor-tang and @luotao1 For the resnet model [(link),](https://github.com/PaddlePaddle/Paddle/blob/b25252ae37ff09ffda69ec2e508a059c707ff01a/benchmark/fluid/ngraph/script/models/resnet.py) we can use what it has in the models repo. We may add a readme and a link to the model. This way I assume we can avoid the process of model review from another team.

As for the train script, it is preferred to have a separate one as we need some modification as aforementioned. The script in the model repo uses parallel executor and we don't support yet. For a fair comparison, we will compare against mkldnn with the same script which is what we are working on with our benchmarking team.

luotao1:
> we need some modification as aforementioned

What's modification will you do? If ngraph don't support parallel-executor, you can use `if-else` in the original script. By the way, parallel-executor are unified to the executor, and this `if-else` should be encapsulated.

> For a fair comparison, we will compare against mkldnn with the same script which is what we are working on with our benchmarking team.

A separate script will make user very confused. Easy-using is very important as well, and `FLAGS_ngraph` is enough.

baojun-nervana:
It seems the script in model repo is written for gpu. We will need more if else besides the executor. eg. 
https://github.com/PaddlePaddle/models/blob/develop/fluid/PaddleCV/image_classification/train.py#L267

luotao1:
It can be used for CPU by `use_gpu` argument.
https://github.com/PaddlePaddle/models/blob/develop/fluid/PaddleCV/image_classification/train.py#L252

luotao1:
close due to https://github.com/PaddlePaddle/models/pull/1904

Issue #15615, opened 2019-01-31
Add mkldnn related unit-tests


luotao1:
To improve the code (line+function) coverage, please **add mkldnn related unit-tests**:
http://ce.paddlepaddle.org:8080/viewLog.html?buildId=21110&buildTypeId=PaddlePaddleFramework_TestCoverage_TestCoverage&tab=report_project15_C___Coverage
- framework/ir/mkldnn
![image](https://user-images.githubusercontent.com/6836917/52030394-a22e4c80-2552-11e9-9328-29466a662d06.png)
- operators/elementwise/mkldnn
![image](https://user-images.githubusercontent.com/6836917/52032590-8fb91080-255c-11e9-91d4-a92950f18ced.png)
![image](https://user-images.githubusercontent.com/6836917/52032629-c727bd00-255c-11e9-9510-48872ec95503.png)
![image](https://user-images.githubusercontent.com/6836917/52031027-6e085b00-2555-11e9-9644-97fa91897c62.png)
![image](https://user-images.githubusercontent.com/6836917/52031039-7c567700-2555-11e9-8447-343850f02912.png)
![image](https://user-images.githubusercontent.com/6836917/52031049-8b3d2980-2555-11e9-92ae-bfeb2231c8b4.png)
![image](https://user-images.githubusercontent.com/6836917/52032662-dc045080-255c-11e9-96b0-a00438226f47.png)

kbinias:
We reproduced your results with coverage testing tool.
- cmake .. -DWITH_GPU=OFF -DWITH_PROFILER=ON -DWITH_STYLE_CHECK=OFF -DWITH_MKLDNN=ON -DWITH_TESTING=ON -DWITH_COVERAGE=ON
- lcov --capture --directory . --output-file coverage.info
- genhtml coverage.info --output-directory out

We will add missing tests.

![image](https://user-images.githubusercontent.com/19502037/52485988-64689e00-2bba-11e9-8c1a-67703ffd723f.png)
![image](https://user-images.githubusercontent.com/19502037/52486012-6fbbc980-2bba-11e9-9ed4-dbdc59f3550e.png)
![image](https://user-images.githubusercontent.com/19502037/52486031-73e7e700-2bba-11e9-813b-66bf8de106e6.png)

kbinias:
[...]

PR #15824, opened 2019-02-20
MKL-DNN: Add test for conv bias fuse pass

Sand3r-:
Implements C-API Unit Test for correct fusing of convolution and bias when paddle is built with MKL-DNN.

image
Related to: #15615

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/conv_bias_mkldnn_fuse_pass_tester.cc
+++ b/paddle/fluid/framework/ir/mkldnn/conv_bias_mkldnn_fuse_pass_tester.cc
@@ -1,0 +1,122 @@
+// Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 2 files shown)


paddle/fluid/framework/ir/mkldnn/conv_bias_mkldnn_fuse_pass_tester.cc
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
sfraczek: 2019 :)

sfraczek (changes requested):

paddle/fluid/framework/ir/mkldnn/conv_bias_mkldnn_fuse_pass_tester.cc
```diff
@@ -0,0 +1,122 @@
...
+TEST(ConvBiasFusePass, basic) {
+  auto prog = BuildProgramDesc();
+  std::unique_ptr<ir::Graph> graph(new ir::Graph(prog));
+
+  NaiveExecutor exe{paddle::platform::CPUPlace()};
+  Scope scope;
+  // Init scope, as it is used in pass
+  exe.CreateVariables(prog, 0, true, &scope);
+  graph->Set(kParamScopeAttr, new framework::Scope*(
+                    const_cast<framework::Scope*>(&scope)));
```
luotao1: It is not recommended to use `const_cast`.

paddle/fluid/framework/ir/mkldnn/conv_bias_mkldnn_fuse_pass_tester.cc
Sand3r-: This breaking change has been introduced as of the most recent commit.

paddle/fluid/framework/ir/mkldnn/conv_bias_mkldnn_fuse_pass_tester.cc
Sand3r-: I've removed the cast since it wasn't even necessary.

sfraczek (approved):
LGTM

PR #15844, opened 2019-02-21
add per kernel config and remove const_cast.

panyx0718:
kernel config is per_kernel instance and not thread-safe. allowed Type is defined in operator.h and reviewed by me. So, people cannot use it to hold any states.

op writer should still favor input/output. Since kernel is selected at runtime, kernel-specific input/output is not ideal

test=develop

### DIFF
--- a/paddle/fluid/framework/operator.h
+++ b/paddle/fluid/framework/operator.h
@@ -28,6 +28,7 @@ limitations under the License. */
 #include "paddle/fluid/framework/lod_tensor.h"
 #include "paddle/fluid/framework/op_info.h"
 #include "paddle/fluid/framework/op_kernel_type.h"
+#include "paddle/fluid/framework/operator_kernel_configs.h"
 #include "paddle/fluid/framework/scope.h"
 #include "paddle/fluid/framework/selected_rows.h"
 #include "paddle/fluid/framework/tensor.h"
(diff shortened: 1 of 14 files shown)


paddle/fluid/framework/operator.h
```diff
@@ -184,12 +184,122 @@ class OperatorBase {
                        const platform::Place& place) const = 0;
 };
 
+template <typename TAlgorithm>
+class AlgorithmsCache {
```
chengduoZH: It is not appropriate that placing `AlgorithmsCache ` in `operator.h` because it is only used by conv_op.

panyx0718: moved to a header file.

paddle/fluid/framework/operator.h
```diff
@@ -184,12 +185,30 @@ class OperatorBase {
...
 };
 
+#ifdef PADDLE_WITH_CUDA
+using KernelConfig = boost::variant<
+    std::shared_ptr<AlgorithmsCache<cudnnConvolutionFwdAlgo_t>>,
+    std::shared_ptr<AlgorithmsCache<cudnnConvolutionBwdDataAlgo_t>>,
+    std::shared_ptr<AlgorithmsCache<cudnnConvolutionBwdFilterAlgo_t>>>;
+#else
+using KernelConfig = boost::variant<boost::blank>;
+#endif
```
chengduoZH, line 195: line 188-195 should be moved to `operator_kernel_configs.h`.

chengduoZH:
This PR seems ok, please @qingqing01 has a look.

chengduoZH (approved):

PR #15856, opened 2019-02-21
MKL-DNN: Add UT to check whether primitives already exist in backward

kbinias:
Implements C-API Unit Test to check if primitives already exist in backward when Paddle is built with MKL-DNN and remove unusable code for attributes.

Releated to #15615


luotao1 (approved):
LGTM

PR #15862, opened 2019-02-21
add override to ApplyImpl and precommit fixes

sfraczek:
added override to ApplyImpl
also required pre-commit fixes
* and #pragma once in edited headers
* add #include<string> in edited headers


luotao1 (approved):
LGTM！Thanks very much!

Issue #15853, opened 2019-02-21
the output tensor's 'dtype' of 'uniform_random' is not set


houj04:
**Brief**
According to the API doc of 'uniform_random' (see http://paddlepaddle.org/documentation/docs/zh/1.2/api/layers.html#uniform-random), the default 'dtype' is:
> dtype (INT) – Output tensor data type. [default 5(FP32)].

But in fact, the dtype of returned tensor is not set to 5.
**System information**
-PaddlePaddle version: 1.2.1
-OS Platform and Distribution: Ubuntu 18.04
-Python version: Python 2.7.15rc1
**To Reproduce**
Just use the code provided in API doc:
```python
import paddle.fluid as fluid
result = fluid.layers.uniform_random(shape=[32, 784])
print result
```
**Describe your current behavior**
```
Traceback (most recent call last):
  File "uniform_random.py", line 3, in <module>
    print result
  File "/usr/local/lib/python2.7/dist-packages/paddle/fluid/framework.py", line 350, in __str__
    return self.to_string(True)
  File "/usr/local/lib/python2.7/dist-packages/paddle/fluid/framework.py", line 370, in to_string
    res_str = _debug_string_(proto, throw_on_error)
  File "/usr/local/lib/python2.7/dist-packages/paddle/fluid/framework.py", line 201, in _debug_string_
    format(error_fields, proto))
ValueError: ['type.lod_tensor.tensor.data_type'] are not initialized.
```

qingqing01:
I use your code to test, the output is:

```
name: "uniform_random_0.tmp_0"
type {
  type: LOD_TENSOR
  lod_tensor {
    tensor {
      data_type: FP64
      dims: 32
      dims: 784
    }
  }
}
persistable: false
```

The def of `uniform_random`:

https://github.com/PaddlePaddle/Paddle/blob/afc885dace69d97531ebe1066af2636d1adbfeea/python/paddle/fluid/layers/ops.py#L63-L66

->

https://github.com/PaddlePaddle/Paddle/blob/afc885dace69d97531ebe1066af2636d1adbfeea/python/paddle/fluid/framework.py#L131-L146

Found that python 2.7 converts None to float64:

```python
Python 2.7.3 (default, Sep  7 2018, 13:59:46)
[GCC 4.4.6 20120305 (Red Hat 4.4.6-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
[...]

PR #15873, opened 2019-02-22
Refine doc of uniform_random and fix dtype

qingqing01:
Fix https://github.com/PaddlePaddle/Paddle/issues/15853

Refine doc since the type of dtype is not right in origin doc.


python/paddle/fluid/layers/ops.py
```diff
@@ -61,7 +61,29 @@
...
 
 def uniform_random(shape, dtype=None, min=None, max=None, seed=None):
+    """
+    This operator initializes a variable with random values sampled from a
+    uniform distribution. The random result is in set [min, max].
+
+    Args:
+        shape (list): The shape of output variable.
+        dtype(np.dtype|core.VarDesc.VarType|str): The type of data, such as
+            float32, float64 etc. Default: float32.
```
chengduoZH, line 71: It may be not unified because the doc says the default value of `dtype` is `float32 `, but `dtype` is None in the parameter list.
http://paddlepaddle.org/documentation/docs/zh/1.2/api_cn/layers_cn.html#uniform-random

qingqing01: @chengduoZH 

In following code,

```python
   if dtype is None:
        dtype = 'float32'
```

Here, I don't change the None to 'float32'.  Maybe should change.

sneaxiy (dismissed):
LGTM

python/paddle/fluid/layers/ops.py
qingqing01: @chengduoZH I update the doc.

chengduoZH (approved):
LGTM

qingqing01:
```
[10:56:47][Step 1/1] + APPROVALS=FALSE
[10:56:47][Step 1/1] + echo 'current pr 15873 got approvals: FALSE'
[10:56:47][Step 1/1] + '[' FALSE == FALSE ']'
[10:56:47][Step 1/1] + echo 'You must have panyx0718 approval for the api change! paddle/fluid/API.spec'
[10:56:47][Step 1/1] + exit 1
[10:56:47][Step 1/1] current pr 15873 got approvals: FALSE
[10:56:47][Step 1/1] You must have panyx0718 approval for the api change! paddle/fluid/API.spec
```

@panyx0718 please help to review. Thanks!

panyx0718 (approved):

PR #15883, opened 2019-02-22
INT8 Pool kernel Key Creation Optimization.

xiaolil1:
Optimize key creation of INT8 pool kernel to improve the peformance of ResNet-50 and MobileNet, especially for latency.

test=develop

### DIFF
--- a/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
@@ -15,6 +15,7 @@ limitations under the License. */
 #include "paddle/fluid/framework/data_layout_transform.h"
 #include "paddle/fluid/operators/pool_op.h"
 #include "paddle/fluid/platform/mkldnn_helper.h"
+#include "paddle/fluid/platform/mkldnn_reuse.h"
 
 namespace paddle {
 namespace operators {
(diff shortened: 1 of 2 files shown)


xiaolil1:
@kbinias @sfraczek @wojtuss  Please help to review this PR, thanks!

paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
```diff
@@ -29,23 +30,23 @@ using mkldnn::stream;
...
-  auto dims2str = [](const memory::dims& operand_dims) {
-    std::string dstr = "";
-    for (size_t i = 0; i < operand_dims.size(); ++i) {
-      dstr += std::to_string(operand_dims[i]) + "-";
-    }
-    return dstr;
-  };
-  return dims2str(input_dims) + dims2str(ksize) + dims2str(strides) +
-         dims2str(paddings) + std::to_string(dt) + pooling_type + suffix;
+std::string CreateKey(const paddle::framework::ExecutionContext& ctx,
```
sfraczek, line 33: I wonder if there was any particular reason why you have removed the static keyword here?

sfraczek (approved):
LGTM. I just have one comment (see above).

paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
hshen14: Thanks @sfraczek. We thought the current design on hash key function depends on OP implementation with MKL-DNN kernel and such key is only visible to the corresponding op, so we removed the static qualifier to make it simpler.

PR #15889, opened 2019-02-22
Enable the operator for the nGraph bridge

mozga-intel:
Enable the sum operator for the nGraph bridge.
test=develop


tensor-tang (approved):
LGTM

PR #15890, opened 2019-02-22
add pipeline io method for industrial usage

guru4elephant:
This pull request is about running a linux pipe inside a data reader. The input files are specified by users, and the preprocessing should be linux pipe command that can generate c++ reader recognized file format.


wangguibao (dismissed):
LGTM

PR #15913, opened 2019-02-25
Enable function coverage for U8/S8 ConvMKLDNNOpKernel

liangan1:
This PR aims to enable the coverage for U8/S8 ConvMKLDNNOpKernel and push to the full coverage on expected functions. According to the mechanism of gcov coverage tool, if there is no information about U8/S8 ConvMKLDNNOpKernel in operators.cmake, the gcno files (produced in build process) cannot have the related information which will be used in execution process. So, gcov tool cannot generate the related statistics information about U8/S8 ConvMKLDNNOpKernel.
In this PR, the coverage of  `"TouchOpKernelRegistrar_conv2d_MKLDNN_S8()"` and
`"TouchOpKernelRegistrar_conv2d_MKLDNN_U8()"` has been enabled. The gcov collects the static type info and the dynamic type info cannot be handled well (usually treat false positives in gcov). For example, `"TouchOpKernelRegistrar_conv2d_MKLDNN_S8()" `is registered for S8 type, which means `"paddle::operators::ConvMKLDNNOpKernel<signed char, float>::ComputeINT8(paddle::framework::ExecutionContext const&) const"` function can be hit , while` "paddle::operators::ConvMKLDNNOpKernel<signed char, float>::ComputeFP32(paddle::framework::ExecutionContext const&) const "` cannot be hit.
@sfraczek and @kbinias please help review. Thanks.


sfraczek (approved):
LGTM.

PR #15914, opened 2019-02-25
Refactor MKL-DNN Sum to use reference version on fallback

Sand3r-:
Reimplement Sum's fallback to use reference version directly instead of using outdated copy-paste code.

### DIFF
--- a/paddle/fluid/operators/mkldnn/sum_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/sum_mkldnn_op.cc
@@ -79,15 +79,6 @@ class SumMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
 
       memory::format input_format = input0.format();
 
-      if (src_tz.size() == 1 && (input_format == memory::format::nchw ||
-                                 input_format == memory::format::nhwc)) {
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/mkldnn/sum_mkldnn_op.cc
```diff
@@ -147,105 +138,10 @@ class SumMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
-              result.device(*ctx.template device_context<MKLDNNDeviceContext>()
-                                 .eigen_device()) = result + in;
-            }
-          }
-        }
-      }
     } else {
-      PADDLE_THROW("Unexpected branch, output variable type is %s",
-                   framework::ToTypeName(out_var->Type()));
+      // TODO(@mozga-intel) Add MKLDNN SelectedRows & LoDTensorArray support
```
mozga-intel: This comment is not true.

mozga-intel:
@Sand3r- Why the comment was removed? - it's really necessary. Could you please restore it and only change up a part of it.

mozga-intel (approved):
LGTM!

PR #15922, opened 2019-02-25
MKL-DNN: Add Activations and Softmax UTs to check if primitives already exist in backward

kbinias:
Implements Activations and Softmax Unit Tests to check if primitives already exist in backward when PaddlePaddle is built with MKL-DNN, move all MKL-DNN Softmax tests to one file.

Releated to #15615


luotao1 (approved):
LGTM

PR #15925, opened 2019-02-26
remove const_cast and refactor ngraph engine code 

baojun-nervana:
1. close a TODO to remove the const_cast. This is to avoid altering the progdesc. Instead of substituting ngraph_engine op in progdesc, it is changed back to scan the op list after creation and replace the op sublist with ngraph_engine (the original way).

2. Refactor the engine code to reduce the overhead. The idea is to minimize the repetitives iteration to iteration. If the graph has no change, the NgraphEngine::Run will grab the data and run the computation.
    - an "engine_key" was added to verify if there is a graph change.
    - the repetitive info for NgraphEngine::Run will be cached

### DIFF
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -34,11 +34,11 @@ limitations under the License. */
 
 #ifdef PADDLE_WITH_NGRAPH
 #include "paddle/fluid/operators/ngraph/ngraph_engine.h"
+DEFINE_bool(use_ngraph, false, "Use NGRAPH to run");
 #endif
 
 DECLARE_bool(benchmark);
 DEFINE_bool(use_mkldnn, false, "Use MKLDNN to run");
-DEFINE_bool(use_ngraph, false, "Use NGRAPH to run");
 
 namespace paddle {
 namespace framework {
@@ -372,6 +372,12 @@ std::unique_ptr<ExecutorPrepareContext> Executor::Prepare(
   for (auto& op_desc : block.AllOps()) {
     ctx->ops_.push_back(OpRegistry::CreateOp(*op_desc));
   }
+#ifdef PADDLE_WITH_NGRAPH
+  if (FLAGS_use_ngraph) {
+    paddle::operators::NgraphEngine::FuseNgraphOps(
+        ctx->prog_.Block(ctx->block_id_), &ctx->ops_);
+  }
+#endif
   return ctx;
 }
 
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -194,9 +194,6 @@ void Executor::Run(const ProgramDesc& pdesc, Scope* scope, int block_id,
                    bool force_disable_gc) {
   platform::RecordBlock b(block_id);
   if (FLAGS_use_mkldnn) EnableMKLDNN(pdesc);
-#ifdef PADDLE_WITH_NGRAPH
-  if (FLAGS_use_ngraph) operators::NgraphEngine::EnableNgraph(pdesc);
-#endif
   auto ctx = Prepare(pdesc, block_id, skip_ref_cnt_vars, force_disable_gc);
   RunPreparedContext(ctx.get(), scope, create_local_scope, create_vars);
 }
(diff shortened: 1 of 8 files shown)


baojun-nervana:
[23:45:10]	+ '[' ']'
[23:45:10]	++ git diff -U0 upstream/develop
[23:45:10]	++ grep -o -m 1 const_cast
[23:45:10]	+ HAS_CONST_CAST=const_cast
[23:45:10]	+ '[' const_cast ']'
[23:45:10]	+ '[' 15925 '!=' '' ']'
[23:45:10]	++ curl -H 'Authorization: token 95a329e820eeae12c4c5c8a0b6e3e39654dc01a4' 'https://api.github.com/repos/PaddlePaddle/Paddle/pulls/15925/reviews?per_page=10000'
[23:45:10]	++ python /paddle/tools/check_pr_approval.py 1 2887803
[23:45:10]	  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[23:45:10]	                                 Dload  Upload   Total   Spent    Left  Speed
[23:45:10]	
[23:45:11]	  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[23:45:12]	  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[23:45:12]	100     5  100     5    0     0      3      0  0:00:01  0:00:01 --:--:--     3
[23:45:12]	+ APPROVALS=FALSE
[23:45:12]	+ echo 'current pr 15925 got approvals: FALSE'
[23:45:12]	+ '[' FALSE == FALSE ']'
[23:45:12]	+ echo 'You must have panyx0718 approval for the const_cast'
[23:45:12]	+ exit 1
[23:45:19]	Process exited with code 1

paddle/fluid/framework/executor.cc
```diff
@@ -14,6 +14,10 @@ limitations under the License. */
 
 #include "paddle/fluid/framework/executor.h"
 #include <deque>
+#include <memory>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
```
tensor-tang: is that necessary include these?

tensor-tang (commented):
I haven't digger deep into this code, but since we are reviewing the design doc inside, maybe you could add this conception as well, especially about the cache method you used.

At the most possibility, you should use PASS to modify the program.

paddle/fluid/framework/executor.cc
baojun-nervana: @tensor-tang It is the cpplint requirement from pre-commit to add those. It seems cpp-lint is on for CI. I have to add those to pass CI. 

I can add the cache to the design doc. 

We will support pass for ngraph enabling, but I think we will still need the current ngraph triggering in executor to support unittest.

tensor-tang: > I have to add those to pass CI.

OK, got it. we have updated the cpp-lint in CI.

tensor-tang:
Please fix the errors in CI. 
BTW, the cpplint version is 1.4.4 in CI/PR, you can check style locally.

paddle/fluid/operators/ngraph/ngraph_engine.h
```diff
@@ -12,12 +12,18 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */
 
+#ifndef PADDLE_FLUID_OPERATORS_NGRAPH_NGRAPH_ENGINE_H_
```
tensor-tang, line 15: use `pragma once` instead.

tensor-tang (approved):
LGTM

paddle/fluid/operators/ngraph/ngraph_engine.h
baojun-nervana: will do. Maybe later PR (not sure if it will dismiss your approval if I do it in this PR)

panyx0718 (approved):
LG const_cast removal

paddle/fluid/operators/ngraph/ngraph_engine.h
tensor-tang: You can address this in #15752 .

PR #15941, opened 2019-02-26
The flag of mkldnn engine is enabled iff it is necessary

mozga-intel:
The flag of mkldnn (#define) is always enabled even as the global flag is disabled.


luotao1 (approved):
LGTM

PR #15943, opened 2019-02-26
MKL-DNN: Add placement pass tester

kbinias:
Implements C-API Unit Test for add placement pass when paddle is built with MKL-DNN.

@sfraczek created the base code.

Releated to #15615


luotao1 (approved):
LGTM！

PR #15953, opened 2019-02-27
Add Event for TensorCopy

chengduoZH:


panyx0718:
what's the performance effect?

chengduoZH:
After using the copy_stream to replace default stream, so unit tests failed. So this PR only adds Event for `Memory Copy` firstly.

panyx0718 (approved):

PR #15955, opened 2019-02-27
set stop_gradient sign to all loss layers

velconia:
test=develop


python/paddle/fluid/layers/nn.py
```diff
@@ -1472,6 +1472,7 @@ def bpr_loss(input, label, name=None):
 
     helper = LayerHelper('bpr_loss', **locals())
     out = helper.create_variable_for_type_inference(dtype=input.dtype)
+    label.stop_gradient = True
```
chengduoZH, line 1475: **We shouldn't secretly change the user's configuration.**

velconia: This is a hot fix

chengduoZH: **This is not hot fix. It's introducing new bugs.**

velconia: So... what's the best way to solve the problem?

PR #15977, opened 2019-02-28
add continuous value model op

colourful-tree:


python/paddle/fluid/layers/nn.py
```diff
@@ -10561,3 +10562,50 @@ def tree_conv(nodes_vector,
...
         pre_activation = out
     return helper.append_activation(pre_activation)
+
+
+def cvm(input, cvm, use_cvm=True):
+    """
+    **cvm layers**
+
+    This layer accepts a tensor named input which is ID after embedded and lod level is 1 ,
+         cvm is a show_click info and returns the cvm output.
```
panyx0718: not sure what does "cvm is a show_click info and returns the cvm output" mean. what is `cvm`?

python/paddle/fluid/layers/nn.py
```diff
@@ -10561,3 +10562,50 @@ def tree_conv(nodes_vector,
...
+
+    Args:
+        input (Variable):  a 2-D LodTensor with shape [N x D], where N is the
+                                batch size, D is the embedding dim. 
+                                lod level = 1.
+        cvm (Variable):    a 2-D Tensor with shape [N x 2], where N is the batch size, 2 is show and click.
+        use_cvm  (bool):  use cvm or not. if use cvm, the output dim is the same as input
```
panyx0718, line 10925: do you think it's confusing? how can cvm layer not using cvm?

python/paddle/fluid/layers/nn.py
colourful-tree: cvm is continuous value moded. now, it only consider show and click value in ctr project.

python/paddle/fluid/layers/nn.py
colourful-tree: yeah，it's a little confusing. cvm op is a  customized op, which input is a sequence had embedd_with_cvm default, so we need a op named cvm to decided whever use it or not.

panyx0718 (dismissed):

shanyi15:
please preview the comment first, thanks

PR #15986, opened 2019-02-28
add op type in check nan/inf

seiriosPlus:


paddle/fluid/framework/operator.cc
```diff
@@ -882,7 +882,8 @@ class RuntimeInferShapeContext : public InferShapeContext {
   const RuntimeContext& ctx_;
 };
 
-static void CheckTensorNANOrInf(const std::string& name,
+static void CheckTensorNANOrInf(const std::string& type,
```
typhoonzero: name it `op_type` should be easy to read

typhoonzero (dismissed):

paddle/fluid/framework/operator.cc
seiriosPlus: done.

paddle/fluid/framework/operator.cc
```diff
@@ -988,9 +989,10 @@ void OperatorWithKernel::RunImpl(const Scope& scope,
       auto* var = exec_scope.FindVar(vname);
       if (var == nullptr) continue;
       if (var->IsType<framework::LoDTensor>()) {
-        CheckTensorNANOrInf(vname, var->Get<framework::LoDTensor>());
+        CheckTensorNANOrInf(type_, vname, var->Get<framework::LoDTensor>());
       } else if (var->IsType<framework::SelectedRows>()) {
-        CheckTensorNANOrInf(vname, var->Get<framework::SelectedRows>().value());
+        CheckTensorNANOrInf(type_, vname,
+                            var->Get<framework::SelectedRows>().value());
```
chengduoZH, line 995: In fact, if we set the `FLAGS_check_nan_inf=1` and `GLOG_vmodule=operator=4`, we can know the `nan/inf` is happend in which op.

seiriosPlus: yes, I'm agreed. But it may print much more logs I will not expect.

chengduoZH (approved):

PR #16013, opened 2019-03-01
MKL-DNN: Add a unit test for pooling when ceil mode is enabled

kbinias:
Adds a unit test for pooling when ceil mode is enabled (instead of floor) to compute the output shape.

Code coverage (lines/functions):
- Previous: 85.9% / 76.7%
- Current: 94.6% / 83.3%

Releated to #15615


luotao1 (approved):
LGTM！

PR #16045, opened 2019-03-04
Imperative remove all descs

velconia:
1. Move all implementations of vars and op's constructor from python into CXX.
2. Remove all descs in imperative mode


paddle/fluid/imperative/layer.cc
```diff
@@ -316,19 +328,14 @@ void PyLayer::RegisterFunc(int func_id, const py::object& py_func) {
 
 int PyLayer::NumFuncs() { return py_funcs_.size(); }
 
-std::vector<VarBase*> PyLayer::Apply(int func_id,
-                                     const std::vector<VarBase*>& inputs) {
+std::vector<Variable*> PyLayer::Apply(int func_id,
```
panyx0718, line 331: why change this? pylayer can be followed by other layers?

paddle/fluid/imperative/tracer.cc
```diff
@@ -85,6 +89,62 @@ platform::Place GetExpectedPlace(platform::Place place, VarBasePtrMap inputs) {
   return result;
 }
 
+framework::VariableNameMap CreateInputVarNameMap(
```
panyx0718, line 92: Is this code copied from somewhere?

paddle/fluid/imperative/tracer.cc
```diff
@@ -152,25 +199,38 @@ std::set<std::string> Tracer::Trace(OpBase* op, const VarBasePtrMap& inputs,
     auto& outvars = outvars_map[it.first];
     const std::vector<VarBase*>& outputs = it.second;
     outvars.reserve(outputs.size());
-    for (size_t i = 0; i < outputs.size(); ++i) {
+    for (size_t i = 0U; i < outputs.size(); ++i) {
```
panyx0718, line 202: interesting. Is this required?

paddle/fluid/imperative/tracer.cc
```diff
@@ -152,25 +199,38 @@ std::set<std::string> Tracer::Trace(OpBase* op, const VarBasePtrMap& inputs,
...
+  if (info.Checker() != nullptr) {
+    info.Checker()->Check(&attrs_map);
+  }
+
+  std::unique_ptr<framework::OperatorBase> op_base =
+      framework::OpRegistry::CreateOp(op->Type(), invars_name_map,
+                                      outvars_name_map, attrs_map);
+
+  // TODO(minqiyang): Support infer var type in imperative mode
+  // Run forward op
```
panyx0718, line 232: how? do you need to change all ops?

python/paddle/fluid/tests/unittests/test_imperative_resnet.py
```diff
@@ -277,7 +277,7 @@ def test_resnet_float32(self):
 
                 dy_grad_value = {}
                 for param in resnet.parameters():
-                    if not param.stop_gradient:
+                    if param.trainable:
```
panyx0718, line 280: what's the relation between trainable and stop_gradient.

paddle/fluid/pybind/pybind_boost_headers.h
```diff
@@ -0,0 +1,115 @@
+/* Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
JiabinYang, line 1: why we need this

paddle/fluid/imperative/tracer.cc
velconia: this will tell compiler that this is an unsigned integer explicitly in compile time.

paddle/fluid/imperative/tracer.cc
velconia: two ways: 1. add a runtime new infer var type in all ops with infer var type 2. create block desc when need infer var type.

python/paddle/fluid/tests/unittests/test_imperative_resnet.py
velconia: stop_gradient controls the gradient flow and trainable controls whether the parameter could be trained in optimizers.

paddle/fluid/pybind/pybind_boost_headers.h
velconia: Because Pybind11 needs custom boost::variant type caster in C++ 11, this could be removed in C++14

paddle/fluid/imperative/tracer.cc
velconia: No...

paddle/fluid/imperative/layer.cc
velconia: Because every VarBase should have a name now.. so VarBase's construnction was moved into PyTrace

paddle/fluid/imperative/layer.h
```diff
@@ -112,31 +112,53 @@ class OpBase;
...
+          const framework::DDim& shape, const platform::Place& place,
+          bool stop_gradient, bool persistable)
+      : VarBase(name, dtype, shape, place, nullptr, nullptr, stop_gradient,
+                persistable) {}
 
  private:
-  VarBase(framework::Variable* var, VarBase* grad, bool stop_gradient)
-      : name_(),
-        var_desc_(nullptr),
+  VarBase(const std::string& name, framework::proto::VarType::Type dtype,
```
JiabinYang, line 138: what usage for this VarBase constructor, Add some Comments?

JiabinYang (approved):
LGTM

paddle/fluid/imperative/tracer.cc
panyx0718: both ways seems strange

PR #16063, opened 2019-03-05
Enhance gc

sneaxiy:
This PR works for:

- To support partial GC, that is to say, not all variables need to be deleted.
- To support GC in while_op, and even recursive nested while_op.
- To disable GC in recurrent_op temporarily.


paddle/fluid/framework/details/eager_deletion_pass.cc
```diff
@@ -12,20 +12,170 @@
...
 #include <vector>
 
 #include "paddle/fluid/framework/details/computation_op_handle.h"
 #include "paddle/fluid/framework/details/eager_deletion_op_handle.h"
-#include "paddle/fluid/framework/details/eager_deletion_pass.h"
 #include "paddle/fluid/framework/details/multi_devices_helper.h"
 #include "paddle/fluid/framework/ir/graph_helper.h"
 
+DEFINE_double(memory_fraction_of_eager_deletion, 1.0,
+              "Fraction of eager deletion");
```
panyx0718: more detail?

paddle/fluid/framework/details/eager_deletion_pass.cc
```diff
@@ -12,20 +12,170 @@
...
+static int64_t GetMemorySize(
+    const std::unordered_map<std::string, std::vector<VarHandle *>> &vars,
+    const std::string &var_name) {
+  auto *var_desc = TryGetLatestVarDesc(vars.at(var_name));
+  PADDLE_ENFORCE_NOT_NULL(var_desc);
+  PADDLE_ENFORCE(IsLoDTensor(var_desc));
+  auto dims = var_desc->GetShape();
+  return SizeOfType(var_desc->GetDataType()) *
+         std::accumulate(dims.begin(), dims.end(), static_cast<int64_t>(1),
+                         std::multiplies<int64_t>());
```
panyx0718, line 56: what if there is -1?

paddle/fluid/framework/executor.cc
```diff
@@ -428,6 +428,10 @@ void Executor::RunPreparedContext(ExecutorPrepareContext* ctx, Scope* scope,
 #ifdef PADDLE_WITH_CUDA
     }
 #endif
+    if (gc) {
+      operators::PrepareSafeEagerDeletionOnWhileOpAndWhileGradOp(ctx->block_id_,
+                                                                 ctx->ops_);
```
panyx0718, line 445: can this be slow if put in Run?

paddle/fluid/operators/controlflow/while_op_helper.cc
```diff
@@ -0,0 +1,289 @@
...
+static void SetSkipVars(const OpVariant &op, std::vector<std::string> attr) {
+  auto &attrs = const_cast<framework::AttributeMap &>(op.Attrs());
+  VLOG(2) << "Prepare to skip " << attr.size()
+          << " var(s): " << GetDebugString(attr);
+  attrs[kSkipEagerDeletionVars] = std::move(attr);
+}
+
+// Check whether the forward while_op and while_grad_op match
+// The program may have many while_ops.
+static bool IsMatchedWhileOpAndWhileGradOp(const OpVariant &fwd_op,
```
panyx0718, line 133: maybe you can do something in python to mark matched while and while_grad op? It maybe easier?

paddle/fluid/framework/details/eager_deletion_pass.cc
sneaxiy: -1 would be treated as positive 1. `GCVarInfo` only cares the absolute value of memory size.

paddle/fluid/framework/executor.cc
sneaxiy: Add `prog_.Size() > 1` to speed up when there is no while op in the program.

paddle/fluid/operators/controlflow/while_op_helper.cc
sneaxiy: Maybe it is the same to do it in Python and C++. Python `append_backward` considers block not while_op.

paddle/fluid/framework/details/eager_deletion_pass.cc
sneaxiy: Done.

panyx0718 (approved):

PR #16069, opened 2019-03-05
Fix reshape bug

chengduoZH:


paddle/fluid/operators/reshape_op.cc
```diff
@@ -56,6 +56,9 @@ class ReshapeOp : public framework::OperatorWithKernel {
   static framework::DDim ValidateShape(const std::vector<int> shape,
                                        const framework::DDim &in_dims) {
     const int64_t in_size = framework::product(in_dims);
+    auto in_dims_vec = framework::vectorize(in_dims);
+    bool all_positive = std::all_of(in_dims_vec.cbegin(), in_dims_vec.cend(),
+                                    [](int64_t i) { return i > 0; });
```
chengduoZH, line 61: If the `in_dims` has two negative value, the result of `framework::product(in_dims)` will be positive, it is not consistent with line94.

python/paddle/fluid/layers/nn.py
```diff
@@ -4740,11 +4740,6 @@ def matmul(x, y, transpose_x=False, transpose_y=False, alpha=1.0, name=None):
     """
 
     def __check_input(x, y):
-        if len(y.shape) > len(x.shape):
-            raise ValueError(
-                "Invalid inputs for matmul. "
-                "x's rank should be always greater than or equal to y'rank.")
-
```
chengduoZH, line 4747: The matmul doesn't require `len(y.shape) > len(x.shape)`.

python/paddle/fluid/layers/nn.py
```diff
@@ -4760,10 +4755,11 @@ def __check_input(x, y):
         if x_shape[-1] != y_shape[-2]:
             raise ValueError("Invalid inputs for matmul.")
 
-        if len(y_shape) > 2:
+        if len(y_shape) > 2 and len(x_shape) > 2:
```
chengduoZH, line 4758: If the len(y_shape)  and `len(x_shape) > 2` are equal, the value of `y_shape[0:-2]` and `x_shape[0:-2]` should be equal.

paddle/fluid/operators/reshape_op.cc
panyx0718: do we have a test for this?

panyx0718 (approved):

python/paddle/fluid/layers/nn.py
panyx0718: why

paddle/fluid/operators/reshape_op.cc
chengduoZH: Currently, we don't have a test for this.
There is an example: when the input's dim is (8, -1, -1,) and the `shape` is (8, -1), the output's dim should be (8, -1), but the origin implementation will infer the outout'dim to (8, 1).

PR #16080, opened 2019-03-06
fix a bug which cause crash when empty variable is used, test=develop

liupluswei:
put the code of checking empty variable ahead to avoid dereference null pointer


JiabinYang (approved):
LGTM

PR #16086, opened 2019-03-06
Refactor Allocator

sneaxiy:
This PR works for:

- Fix Allocator bug in `Allocator::Free()`. Beforehand, some allocators required that the allocated allocation could be `delete`d directly, but it does not make sense, which may cause memory leak. Use `decorated_allocators_` to fix this bug. `InlinedVector` is introduced to record `decorated_allocators_ ` to avoid frequent heap allocation.

- Add `AutoGrowthBestFitAllocator` to cache memory allocated from `cudaMalloc`. Unlike `BuddyAllocator`, memory block whose size is larger than `max_chunk_size_` would not be returned to GPU by calling `cudaFree` when it is freed. Tests on Transformer model show that the speed would not change when using `AutoGrowthBestFitAllocator`, but defragmentation would exist when `FLAGS_fraction_of_gpu_memory_to_use` is too small.

- Enable `MultiBinBufferedAllocator` in `naive_best_fit` strategy. `AutoGrowthBestFitAllocator` maintains buffers internally, so that it does not need to be decorated by `MultiBinBufferedAllocator`.


panyx0718:
It should be spit into at least 3 PRs

1. InlinedVector
2. Bug fix
3. New Allocation strategy (and need improvements)

paddle/fluid/memory/allocation/allocator.h
```diff
@@ -72,17 +77,31 @@ class Allocation {
...
+  inline Allocator* TopDecoratedAllocator() {
+    return decorated_allocators_.back();
+  }
 
  private:
-  Allocator* allocator_;
   void* ptr_;
   size_t size_;
   platform::Place place_;
+  framework::InlinedVector<Allocator*, 8> decorated_allocators_;
```
chengduoZH, line 101: Why not use `std::vector` directly? Does there have some different efficient?

sneaxiy: `std::vector<T>`  would do heap allocation, and increases its capacity in a mode like `0->1->2->4->...`, which may be inefficient because `decorated_allocators_` is a very small vector. `InlinedVector` would be more efficient.

PR #16092, opened 2019-03-06
add layer norm to Layers, add transformer test in imperative mode

JiabinYang:
1. add transformer test in imperative mode
2. refactor layer norm in Layers and add test in test_layers
4. build parameter name in standard way in Layers' `create_parameter` method
5. fix minimize error about param being overwritten 
6. fix `one_hot` layer by set it stop gradient as true
7. change source from `cdn` to `bj` to pass ci


python/paddle/fluid/imperative/nn.py
```diff
@@ -432,13 +434,14 @@ def __init__(self,
...
                  param_attr=None,
+                 stop_gradient=False,
                  dtype='float32'):
 
         super(Embedding, self).__init__(name_scope)
         self._size = size
         self._is_sparse = is_sparse
         self._is_distributed = is_distributed
-
+        self._stop_gradient = stop_gradient
```
velconia: why?

python/paddle/fluid/optimizer.py
```diff
@@ -388,13 +388,14 @@ def minimize(self,
...
                 if not param.trainable:
                     continue
-                # create gradient variable
-                grad_var = Variable(
-                    block=loss.block,
-                    name=param._ivar._grad_name(),
-                    stop_gradient=True,
-                    ivar=param._ivar._grad_ivar())
-                params_grads.append((param, grad_var))
+                if param._ivar._grad_ivar() is not None:
```
velconia, line 391: Add a comment here to indicate the reason why grad ivar could be None

python/paddle/fluid/imperative/layers.py
```diff
@@ -189,11 +196,19 @@ def __getattr__(self, name):
...
+            else:
+                if name in dict:
+                    del dict[name]
+
         if isinstance(value, framework.Parameter):
             params = self.__dict__.get('_parameters', None)
             if params is None:
                 raise ValueError(
                     "super(YourLayer, self).__init__() should be called first")
+            remove_from_layers(self.__dict__.get('_parameters', None))
```
panyx0718: why? the parameter might be referenced somewhere else?

python/paddle/fluid/imperative/nn.py
```diff
@@ -432,13 +434,14 @@ def __init__(self,
                  is_distributed=False,
                  padding_idx=None,
                  param_attr=None,
+                 stop_gradient=False,
```
panyx0718: why? how does pytorch do?

python/paddle/fluid/optimizer.py
panyx0718, line 391: why?

JiabinYang: Here we will get all params to minimize, however some param shouldn't be minimized cuz it has been overwritten by other params. So we should avoid creating wrong grads for those params

python/paddle/fluid/imperative/layers.py
JiabinYang: in that case, parameter should be in the parameter list of the layer which truly hold it, here I just want to make sure each layer hold their own parameters after weight sharing

paddle/fluid/imperative/tracer.cc
```diff
@@ -277,6 +277,7 @@ std::set<std::string> Tracer::Trace(OpBase* op, const VarBasePtrMap& inputs,
           VarBase* var = current_vars_map[var_it->second];
           InitGrad(var, prepared_op.GetDeviceContext());
           grad_out_vars.push_back(var->grads_);
+          VLOG(3) << "grads output var name: " << var->name_;
```
velconia, line 280: Better use Name()

velconia (approved):
LGTM with little comments and need to align cost with static mode in more batchs

python/paddle/fluid/imperative/nn.py
JiabinYang: one_hot should not have gradient and if we don't set this it will cause segment fault in imperative mode, better option is set all op to stop gradient first and then change to false case by case

python/paddle/fluid/optimizer.py
JiabinYang: ok, I will fix this in next PR

paddle/fluid/imperative/tracer.cc
JiabinYang: K

PR #16106, opened 2019-03-07
Make parent_idx a dispensable output for beam_search op to support models saved by older paddle version.

Xreki:


guoshengCS (approved):
LGTM

PR #16107, opened 2019-03-07
Add clone function for IrGraph.

wzzju:


paddle/fluid/framework/ir/graph.h
```diff
@@ -71,6 +72,7 @@ namespace ir {
 class Graph {
  public:
   explicit Graph(const ProgramDesc &program);
+  Graph(const Graph &o);
```
panyx0718: can we use a `Clone` method and ban this method.
In the `Clone` comment, say that it only clones graph structure, not the attributes.

wzzju: OK. Thx.

python/paddle/fluid/framework.py
```diff
@@ -1950,6 +1950,19 @@ def __init__(self, graph, for_test=False):
...
+        Create a new and duplicated IrGraph.
+
+        Warns:
+            The method only clones the graph structure, not its attributes.
+
+        Returns:
+            IrGraph: A new and duplicated graph.
+        """
+        g = self.graph.clone()
+        return IrGraph(g, self._for_test)
```
wanghaoshuang, line 2016: `for_test` should be an argument of clone function.

wzzju: Hi, I think you should change the `for_test` value when you perform the clone operation. It is different from `Program`.

panyx0718 (approved):

paddle/fluid/framework/ir/graph.cc
```diff
@@ -152,6 +152,39 @@ void Graph::ResolveHazard(
   }
 }
 
+std::shared_ptr<Graph> Graph::Clone() {
```
panyx0718, line 155: no test?

panyx0718 (changes requested):

paddle/fluid/framework/ir/graph.cc
wzzju: Have added the executor test for the graph clone API.

panyx0718 (approved):

PR #16127, opened 2019-03-08
Add cpu_quantize_pass for C-API quantization

wojtuss:
We add `cpu_quantize_pass`, which quantizes `conv2d` and `pool2d` operators in a model. After the pass, a `quantize` op and a `dequantize` op come before and after a quantized operator.

This pass works with the C-API quantization core (https://github.com/PaddlePaddle/Paddle/pull/15987)
and is the first pass in a series of quantization passes.

After the pass we get the following accuracy
MobileNet-v1:
• FP32: 70.448%
• INT8v2: 70.172%
ResNet50:
• FP32: 76.054%
• INT8v2: 75.716%

There are still a few improvements in progress for this pass.


paddle/fluid/operators/conv_op.cc
```diff
@@ -194,6 +195,10 @@ void Conv2DOpMaker::Make() {
   AddAttr<bool>("use_mkldnn",
                 "(bool, default false) Only used in mkldnn kernel")
       .SetDefault(false);
+  AddAttr<bool>("use_quantizer", "(bool, default false) Only used on CPU")
+      .SetDefault(false);
+  AddAttr<bool>("quantized", "(bool, default false) Only used on CPU")
+      .SetDefault(false);
```
luotao1, line 203: Why do you add `use_quantizer` and `quantized` attribute? Could you not add them?
In #16002, we use `RuntimeContextCachePass` to add `kEnableRuntimeContext` attribute, but we don't need add attr in each op.

wojtuss: The `use_quantizer` attribute is set to `true` during `cpu_quantize_placement_pass` for operators that are to be quantized, according to settings in `QuantizerConfig`. It works similarly to `use_mkldnn` attribute and is quite essential to this approach.

The `quantized` attribute we use to mark already quantized operators. In the first place we match and quantize all the `conv2d` ops with residual data, and then all other `conv2d` ops. For the latter, we use graph pattern for all `conv2d` and to exclude already quantized ops we use the attribute `quantized`. We could go without the `quantized` attribute by adding more graph patterns, if you prefer.

luotao1: @panyx0718 @Superjomn @wzzju @qingqing01 How do you think about adding `use_quantizer` and `quantized` attribute for ops who support INT8 kernels?

sfraczek: We will try to eliminate `quantized` attribute.

paddle/fluid/operators/conv_op.cc
```diff
@@ -194,6 +195,10 @@ void Conv2DOpMaker::Make() {
   AddAttr<bool>("use_mkldnn",
                 "(bool, default false) Only used in mkldnn kernel")
       .SetDefault(false);
+  AddAttr<bool>("use_quantizer", "(bool, default false) Only used on CPU")
```
panyx0718: more detail?

paddle/fluid/operators/conv_op.cc
```diff
@@ -194,6 +195,10 @@ void Conv2DOpMaker::Make() {
...
+      .SetDefault(false);
+  AddAttr<bool>("quantized", "(bool, default false) Only used on CPU")
```
panyx0718: more detail?

paddle/fluid/operators/conv_op.cc
luotao1: @wojtuss @sfraczek 
Discussed with @panyx0718, you can hold `use_quantizer ` but delete `quantized`. Besides, could you add more explanation for `use_quantizer` attribute, since `(bool, default false) Only used on CPU` is too simple.

hshen14: Right. I also suggested in design meeting that quantized attribute seems duplicated here, since int8 kernel will be scheduled by the input data type. On use_quantizer attribute, I would also suggest whether cpu_quantize_pass can accept the INT8 op list, instead of another pass with INT8 op list and produce such attribute that is used in another pass. The attribute seems intermediate during pass, better not expose in op.

luotao1: > I would also suggest whether cpu_quantize_pass can accept the INT8 op list

I think that is OK as well.

wojtuss: The sequence is as follows:
1. mark operators for quantization (cpu_quantize_placement_pass with the list of ops),
2. run warmup iteration and calculate scales for inputs and outputs of marked operators only,
3. run quantization passes.

If we passed the list of operators for quantization directly to the quantization passes in step 3., we would have to calculate scales for all the variables in the graph in step 2.

hshen14: Yes, I understood the sequence. My concern is the attribute does not contribute for op computation, but to provide the hint for sampling (offline stage). Pass itself is okay for me.

wojtuss: @hshen14 , got your point. I would think about getting rid of this attribute in the future.

paddle/fluid/operators/conv_op.cc
wojtuss: The `quantized` attribute is removed.

paddle/fluid/operators/conv_op.cc
wojtuss: Added.

luotao1 (approved):
LGTM

PR #16144, opened 2019-03-11
PaddingRNN model memory optimize

sneaxiy:
Refine cross_entropy and expand ops to save memory. About 12G GPU memory would be saved.


paddle/fluid/operators/cross_entropy2_op.cc
```diff
@@ -0,0 +1,218 @@
...
+    if ((!ctx->IsRuntime()) && (framework::product(x_dims) <= 0 ||
+                                framework::product(label_dims) <= 0)) {
+      check = false;
+    }
+    if (check) {
+      PADDLE_ENFORCE_EQ(framework::slice_ddim(x_dims, 0, rank - 1),
+                        framework::slice_ddim(label_dims, 0, rank - 1),
+                        "Input(X) and Input(Label) shall have the same shape "
+                        "except the last dimension.");
+    }
```
chengduoZH: Why make a different check here?

paddle/fluid/operators/cross_entropy2_op.cc
```diff
@@ -0,0 +1,218 @@
...
+#include <unordered_map>
+
+namespace paddle {
+namespace operators {
+
+class CrossEntropyOp2 : public framework::OperatorWithKernel {
+ public:
+  using framework::OperatorWithKernel::OperatorWithKernel;
+
+  void InferShape(framework::InferShapeContext* ctx) const override {
```
chengduoZH: Most of the logic of `InferShape` is the same as `CrossEntropyOp::InferShape`, you can make an inheritance class.

paddle/fluid/operators/cross_entropy2_op.h
```diff
@@ -0,0 +1,188 @@
...
+#ifdef __NVCC__
+  return exp(x);
+#else
+  return std::exp(x);
+#endif
+}
+
+template <typename T>
+struct CrossEntropyForwardFunctor {
+  CrossEntropyForwardFunctor(const T *x, T *y, const int64_t *label,
```
chengduoZH: Why don't you reuse [`CrossEntropyKernel`](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/cross_entropy.cu#L35)?
If `platform::ForRange` is faster, we should replace `CrossEntropyKernel` with`platform::ForRange`.

paddle/fluid/operators/cross_entropy2_op.h
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  T *dx_;
+  const T *y_;
+  const T *dy_;
+  const int64_t *label_;
+  int64_t ignore_index_;
+  int64_t feature_size_;
+};
```
chengduoZH: The above code should be placed to `cross_entropy.h`.

paddle/fluid/operators/expand_op.cc
```diff
@@ -138,12 +138,28 @@ class ExpandGradOp : public framework::OperatorWithKernel {
...
+  }
+};
+
 }  // namespace operators
 }  // namespace paddle
 
 namespace ops = paddle::operators;
 REGISTER_OPERATOR(expand, ops::ExpandOp, ops::ExpandOpMaker,
-                  paddle::framework::DefaultGradOpDescMaker<true>);
+                  ops::ExpandGradOpDescMaker);
```
chengduoZH, line 163: We should take care of compatibility. The saved training model may be unavailable if we replace `DefaultGradOpDescMaker<true>` with `ops::ExpandGradOpDescMaker` directly.

paddle/fluid/operators/cross_entropy2_op.cc
sneaxiy: Just copy from `CrossEntropyOp`.

paddle/fluid/operators/cross_entropy2_op.h
sneaxiy: It is not about speed. I just make a different op here to avoid confusing with original cross entropy op.

paddle/fluid/operators/expand_op.cc
sneaxiy: Compatibility is OK. The saved training model may contain extra variable names but it does not matter.

paddle/fluid/operators/cross_entropy2_op.cc
sneaxiy: Done.

paddle/fluid/operators/cross_entropy2_op.h
sneaxiy: Done. Move to cross_entropy_op_base.h

paddle/fluid/operators/cross_entropy2_op.cc
```diff
@@ -0,0 +1,137 @@
+/* Copyright (c) 2016 PaddlePaddle Authors. All Rights Reserved.
```
chengduoZH: 2016 -> 2019

chengduoZH (approved):
LGTM

PR #16149, opened 2019-03-11
Fix bugs for static quantization strategy (mutil-cards train).

wzzju:
Fix bugs for the static quantization strategy. Now the static quantization strategy can use mutil-cards to train.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -1829,6 +1829,17 @@ class IrOpNode(IrNode):
         super(IrOpNode, self).__init__(node)
         self.node = node
 
+    def block(self):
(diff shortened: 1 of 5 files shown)


python/paddle/fluid/framework.py
```diff
@@ -1829,6 +1829,17 @@ def __init__(self, node):
         super(IrOpNode, self).__init__(node)
         self.node = node
 
+    def block(self):
```
qingqing01, line 1832: 在IRGraph里还有block的概念？

wzzju: 这个需要讨论一下，谢谢青青姐。

PR #16153, opened 2019-03-11
fix default value of epsilon in l2_normalize func

shippingwang:
fix 2nd of [PaddlePaddle/FluidDoc#672](https://github.com/PaddlePaddle/FluidDoc/issues/672)


python/paddle/fluid/layers/nn.py
```diff
@@ -4731,7 +4731,7 @@ def l2_normalize(x, axis, epsilon=1e-12, name=None):
             the dimension to normalization is rank(X) + axis. -1 is the
             last dimension.
         epsilon(float): The epsilon value is used to avoid division by zero, \
-            the defalut value is 1e-10.
+            the defalut value is 1e-12.
```
shanyi15: default

shippingwang: ：）copy that

shanyi15 (dismissed):
LG comments

shippingwang:
already done by #16805

PR #16156, opened 2019-03-11
Enable MKL-DNN INT8 Concat Kernel.

xiaolil1:
Enable INT8 Concat Kernel to improve the performance of MobileNet-SSD, theoretical improvement is 4X for memory bandwidth saving.

test=develop


xiaolil1:
[06:02:47]	[Step 1/1] 99% tests passed, 1 tests failed out of 577
[06:02:47]	[Step 1/1] 
[06:02:47]	[Step 1/1] Total Test time (real) = 3055.28 sec
[06:02:47]	[Step 1/1] 
[06:02:47]	[Step 1/1] The following tests FAILED:
[06:02:47]	[Step 1/1] 	492 - test_layers (Failed)
[06:02:47]	[Step 1/1] Errors while running CTest
[06:02:57]	[Step 1/1] Process exited with code 8
[06:02:57]	[Step 1/1] Process exited with code 8 (Step: Build and test (Command Line))
[06:02:57]	[Step 1/1] Step Build and test (Command Line) failed

This UT always fail in PR_CI, but it seems not related to my commitment. Is it a common issue?
@luotao1 Could you please help to check it?

luotao1:
fixed by #16158. Please re-run again.

luotao1:
fixed by #16158. Please re-run again.

xiaolil1:
> fixed by #16158. Please re-run again.

Got it, thanks for your quick reply!

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
```diff
@@ -120,6 +158,16 @@ template <typename T>
 class ConcatMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
  public:
   void Compute(const paddle::framework::ExecutionContext& ctx) const override {
+    bool is_INT8 =
+        std::is_same<T, int8_t>::value || std::is_same<T, uint8_t>::value;
```
Sand3r-: Wouldn't it be better to use template specialisation for `int8_t` and `uint8_t` for Compute method? That way, the method to call would be known at the compile time and we'd avoid both RTTI and a branch which are a bit performance costy.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
```diff
@@ -134,7 +182,65 @@ class ConcatMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     auto concat = prim_creator.CreateConcatPrimitive(concat_pd, output, place);
     stream(stream::kind::eager).submit({concat}).wait();
 
-    output->set_mkldnn_prim_desc(concat_pd.dst_primitive_desc());
+    output->set_layout(DataLayout::kMKLDNN);
+    output->set_format(GetDstMemFormat(concat_pd));
```
Sand3r-: It is advised to stick to the new `set_mkldnn_prim_desc` API created by @jczaja, since the old `set_format` and `set_layout` are going to be deprecated.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
```diff
@@ -134,7 +182,65 @@ class ConcatMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    int64_t concat_axis = static_cast<int64_t>(ctx.Attr<int>("axis"));
+    auto& dev_ctx =
+        ctx.template device_context<paddle::platform::MKLDNNDeviceContext>();
+    auto place = GetCpuPlace(ctx);
+
+    memory::data_type dt =
+        paddle::framework::ToMKLDNNDataType(multi_input[0]->type());
+
+    ConcatPrimitiveFactory<T> prim_creator;
+    std::string key = CreateKey(ctx, multi_input, concat_axis, dt);
```
Sand3r-, line 173: Regarding the re-use mechanism in general - it'd be much easier to just store the ConcatPrimitiveFactory in the context instead of creating all the keys and storing primitives separately, while logically they create a single whole. It'd also make the code much shorter.

Sand3r- (commented):
Thank you @xiaolil1 for this PR! Please consider introducing changes I've posted underneath.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
hshen14: @Sand3r- No, compile time does not know the data type. You may refer to INT8 kernel at conv_mkldnn_op.cc as well.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
hshen14: Thanks for your proposal. The reason why we put the key generation out of ConcatPrimitiveFactory is to follow the similar pattern of design and implementation in other MKL-DNN FP32/INT8 kernels, e.g., Conv, Pooling, Quantize, and Dequantize.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
xiaolil1: Thanks for your suggestion, i have refine it.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
Sand3r-: It actually does, because it happens during the registration of an op. Then during the runtime an appropriate instance of such template method would be called.

hshen14: Thanks @Sand3r- . How about adding a JIRA internally to experiment your suggestion, with other INT8 kernels, like Conv? With that, we can have a unified solution.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
luotao1: > since the old set_format and set_layout are going to be deprecated

@Sand3r- @jczaja Do you need to use `set_mkldnn_prim_desc` to replace all the `set_format` and `set_layout` in the codes?

Sand3r-: That's the plan, yes.

Sand3r-:
I have thought about this recently, and wondered whether this branch and division into fp32 and int8 functions makes sense in case of concat. Typically, there are scales to be used in other operators, and I guess an approach of dividing the code into int8 and fp32 counterparts is acceptable, but concat doesn't have any scales, or typical for int8 parameters. Perhaps it'd be enough to simplify the code by modifying the old code with adding `memory::data_type dt = paddle::framework::ToMKLDNNDataType(multi_input[0]->type());` to the old Compute function and just creating the ConcatPrimitiveFactory using the constructor which uses dt. And correct me if I'm wrong, but the rest should pretty much stay the same, but the code would just look simpler.

hshen14:
Thanks @Sand3r- for your comments.

Hi @luotao1, what do you think? This PR aims to newly introduce INT8 kernel for concat and keep FP32 kernel same as before. As suggested, it is better to separate the kernel. The major reason is to avoid any potential issues for the deployed FP32 model with MKL-DNN kernel.

luotao1:
> Perhaps it'd be enough to simplify the code by modifying the old code with adding memory::data_type dt = paddle::framework::ToMKLDNNDataType(multi_input[0]->type()); to the old Compute function and just creating the ConcatPrimitiveFactory using the constructor which uses dt.

If this way is simple, makes codes less, not includes a lot of `if-else` like `conv_mkldnn_op` and easy to debug FP32 codes as well, I prefer this way.

hshen14:
@Sand3r- @tensor-tang Please help review the PR, thanks.

paddle/fluid/operators/mkldnn/concat_mkldnn_op.cc
```diff
@@ -120,21 +158,59 @@ template <typename T>
...
+    const std::string key_concat_pd = key + "@concat_pd";
+    const std::string key_srcs = key + "@concat_srcs";
+    const std::string key_dst = key + "@concat_dst";
+
+    std::shared_ptr<concat::primitive_desc> concat_pd;
+    std::shared_ptr<std::vector<memory>> srcs;
+    std::shared_ptr<memory> dst_mem;
+    auto concat_p = std::static_pointer_cast<concat>(dev_ctx.GetBlob(key_prim));
+
+    if (concat_p == nullptr) {
```
Sand3r-, line 184: What is the benefit of using this re-use mechanism instead of just caching the PrimitiveFactory?

hshen14: Re-use will help improve the latency with batch size 1 in general. The single op benefit is not so big, but topology level is considerable.

Sand3r- (approved):
It's been agreed with @hshen14 that the primitive-reuse will be limited to just reusing PrimitiveFactory in another refactoring PR.

luotao1:
> in another refactoring PR.

@Sand3r- @hshen14 When and where is the refracting PR?

hshen14:
Hi @luotao1 , I talked with @Sand3r- . Since the complete primitive factory idea was in another PR (FC) https://github.com/PaddlePaddle/Paddle/pull/15226, we would consider whether @Sand3r- will adapt the primitive factory for FP32 in Concat first after https://github.com/PaddlePaddle/Paddle/pull/15226 is merged and we can refine it for INT8 kernel. With that, we can have a better unified primitive reuse solution. @xiaolil1

Sand3r-:
Well, the concept of factory is different for each op unfortunately, due to high variety of needed memory primitves. Hence these can be hardly unified, but it'd certainly be worth to consider rewriting other mkldnn operators to use this scheme.

luotao1:
> it'd certainly be worth to consider rewriting other mkldnn operators to use this scheme.

Do you have a schedule for rewriting? @Sand3r- @kbinias @jianhang-liu

PR #16157, opened 2019-03-12
refine anakin subgraph.

NHZlX:
support change input size
refine simple anakin detection patten pass.
align facebox model


Superjomn (approved):
LGTM

PR #16159, opened 2019-03-12
Enable INT8 transpose kernel for MobileNet-SSD improvement.

xiaolil1:
Enable INT8 Transpose Kernel to improve the performance of MobileNet-SSD, theoretical improvement is 4X for memory bandwidth saving.

test=develop


xiaolil1:
@kbinias @sfraczek @wojtuss Please help to review this PR, thanks!

python/paddle/fluid/tests/unittests/mkldnn/test_transpose_int8_mkldnn_op.py
```diff
@@ -0,0 +1,78 @@
+# Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
sfraczek: Shouldn't there be 2019?

hshen14: @luotao1 Shall we use 2019 or keep 2018? You probably have a tool to unify all the copyright year?

luotao1: > You probably have a tool to unify all the copyright year?

We don't unify all the copyright year, since it records the first created time. Thus, if a file is first created in 2019, it should be 2019.

hshen14: > > You probably have a tool to unify all the copyright year?
> 
> We don't unify all the copyright year, since it records the first created time. Thus, if a file is first created in 2019, it should be 2019.

Understood. Thanks.

xiaolil1: Thanks for the reminder, I have changed the license year in latest version.

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
```diff
@@ -70,6 +70,21 @@ class TransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
         paddle::framework::vectorize2int(output->dims()),
         mkldnn::memory::format::blocked);
     output->set_mkldnn_prim_desc(output_mem_pd);
+    output->set_layout(DataLayout::kMKLDNN);
```
jczaja: This line is likely not needed. set_mkldnn_prim_desc is internally setting layout to DataLayout::kMKLDNN.

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
```diff
@@ -70,6 +70,21 @@ class TransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+
+template <typename T>
+class TransposeINT8MKLDNNOpKernel : public paddle::framework::OpKernel<T> {
+ public:
+  void Compute(const paddle::framework::ExecutionContext& ctx) const override {
+    VLOG(4) << "The format of MKL-DNN INT8 will always be 'nhwc', do not need "
+               "to do transpose for it.";
+    auto* input = ctx.Input<Tensor>("X");
+    auto* output = ctx.Output<Tensor>("Out");
+    output->ShareDataWith(*input);
```
jczaja, line 93: Could you explain a reason why there is no any data rearrangement here? Transpose Op's intention is to swap axes even if finally paddle format is intact. I'm guessing that for some INt8 models that could be a shortcut, but for some others still rearranging should take place ?

hshen14: @jczaja Right, as for now, INT8 MKL-DNN kernel supports the only format NHWC, that's why we can safely use the shortcut in Transpose INT8 kernel.

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
xiaolil1: Yes, i missed this new modification when merging code, will update code.

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
sfraczek: I think @jczaja made a good point. If Baidu team is fine with this proposition then I'm cool with it. However thinking about a long term solution, I would expect that the **axis** attribute would have a meaning in a network.
I think it would be safer to check during execution that axis is [1,2,3,4] and then do this pass-through instead of always performing it regardless of the axis attribute because if someone used different axis attribute then it could be invalid (no longer nhwc). 
For example, if we know that the network was designed with nchw in mind and we quantize it, then maybe we could add a pass (or extend the quantization pass) to modify the axis attribute of this layer accordingly. We should this way end up with [1,2,3,4] for example. After a check that there is no reorder needed then it would pass through as it does in this PR. It could throw an exception at least when the axis has unexpected order.

hshen14: @sfraczek Yes, axis should be checked. Thanks for your suggestion.

xiaolil1: done.

sfraczek (approved):
Okay

PR #16162, opened 2019-03-12
Fix NaN bugs for static quantization strategy (mutil-cards train). 

wzzju:
Fix bugs for the static quantization strategy. Now the static quantization strategy can use mutil-cards to train.


qingqing01 (dismissed):
LGTM

panyx0718 (dismissed):

panyx0718 (approved):

PR #16177, opened 2019-03-13
remove graph visualizer tool, which can be replaced by python IrGraph draw API

fc500110:


wzzju (approved):
The visualizer tool should be deleted, because this function has been included in `IrGraph`. What's more, the visualizer tool is very huge, and is about 1.4GB.  What a mess!

luotao1 (approved):

PR #16205, opened 2019-03-14
temoprarily disable the code of use kCUDNN, test=develop

liupluswei:


jerrywgz (commented):
better to add TODO to deal with cudnn act

jerrywgz (approved):

PR #16206, opened 2019-03-14
Impl fp16 compute kernel for slice_op

kuke:
For the eigen function `pad` doesn't seem to support fp16


typhoonzero (approved):
LGTM for the code.

PR #16154, opened 2019-03-11
add all_kernels_must_compute_runtime_shape example for speedup infershape

luotao1:
We plan to implement computing runtime shape for all ops, in order to skip runtime infershape for speedup. 

This PR add `all_kernels_must_compute_runtime_shape` attributes and make an example of three ops: `fused_embedding_seq_pool_op`, `hash_op` and `sequence_enumerate_op`. 

Related #15959

PR #16213, opened 2019-03-15
Skip compile infer shape in box_coder_op

qingqing01:
To support face detection depolyment.


luotao1:
Could you follow #16154 to skip compile infershape?

qingqing01:
@luotao1 Each op needs to set the flag?  And here skip the compile infer-shape, not runtime infer-shape.

luotao1:
> Each op needs to set the flag? 

Yes, each op needs to set this flag in the future. 

> And here skip the compile infer-shape, not runtime infer-shape

Got it. If you use this flag, it will skip `infershape()` function in runtime, and you can put line65-70 into `Compute()` fucntion. You could try it.

NHZlX (approved):

PR #16225, opened 2019-03-15
Fix mkldnn related irpasses being added triple times

### LINKED ISSUE #16174: mkldnn related irpasses are added triple times when mkldnn is enabled in AnaylsisPredictor

luotao1:
- fix #16174 
- Inspired by `pass_builder()->EnableMKLDNN()` is called again in `Update()` after pass_builder reset (see https://github.com/PaddlePaddle/Paddle/pull/16175#issuecomment-472345856).

### DIFF
--- a/paddle/fluid/inference/api/analysis_config.cc
+++ b/paddle/fluid/inference/api/analysis_config.cc
@@ -204,20 +204,6 @@ void AnalysisConfig::Update() {
     }
   }
 
-  if (use_mkldnn_) {
-    if (!enable_ir_optim_) {
-      LOG(ERROR)
-          << "EnableMKLDNN() only works when IR optimization is enabled.";
-    }
-#ifdef PADDLE_WITH_MKLDNN
-    pass_builder()->EnableMKLDNN();
-    use_mkldnn_ = true;
-#else
-    LOG(ERROR) << "Please compile with MKLDNN first to use MKLDNN";
-    use_mkldnn_ = false;
-#endif
-  }
-
   if (enable_memory_optim_) {
     pass_builder()->AppendAnalysisPass("memory_optimize_pass");
   }


sfraczek:
Roughly looking, this could work fine unless a user calls AnalysisConfig::EnableMKLDNN() twice. Now the flag use_mkldnn_ has no use in both:
1. AnalysisConfig - because it is not checked.
2. PassStrategy - because now I think we assume CpuPassStrategy::EnableMKLDNN() is called only once.

luotao1:
>  Now the flag use_mkldnn_ has no use in both:

flag use_mkldnn_ would be used in 
https://github.com/PaddlePaddle/Paddle/blob/c49e604906551e1802c1da213450631b660d9daa/paddle/fluid/inference/api/analysis_predictor.cc#L376-L379

sfraczek:
ok

LeoZhao-BJ:
with this change, the issue looks like fixed, but it seems to cause status mismatch between Passes_ and use_mkldnn_ in CpuPassStrategy.  use_mkldnn_ in old CpuPassStrategy instance was not copied into new one, while Passes_ copied.

Better to add use_mkldnn_  in CpuPassStrategy copy constructor to avoid this mismatch.

luotao1:
>  use_mkldnn_ in old CpuPassStrategy instance was not copied into new one, while Passes_ copied

`use_mkldnn_` will be copied here.
https://github.com/PaddlePaddle/Paddle/blob/7458114b5b0919b26da56396dd30c1f2443bbd9a/paddle/fluid/inference/api/analysis_config.cc#L108

tensor-tang (commented):
LGTM.
Need @Superjomn 's double check.

LeoZhao-BJ:
> > use_mkldnn_ in old CpuPassStrategy instance was not copied into new one, while Passes_ copied
> 
> `use_mkldnn_` will be copied here.
> [Paddle/paddle/fluid/inference/api/analysis_config.cc](https://github.com/PaddlePaddle/Paddle/blob/7458114b5b0919b26da56396dd30c1f2443bbd9a/paddle/fluid/inference/api/analysis_config.cc#L108)
> 
> Line 108 in [7458114](/PaddlePaddle/Paddle/commit/7458114b5b0919b26da56396dd30c1f2443bbd9a)
> 
>  CP_MEMBER(use_mkldnn_);

I mean use_mkldnn_ in CpuPassStrategy/PassStrategy, instead of AnalysisConfig, 

```
#ifdef PADDLE_WITH_MKLDNN
  use_mkldnn_ = true;
#else
  LOG(ERROR) << "Please compile with MKLDNN first to use MKLDNN";
  use_mkldnn_ = false;
#endif

  Update();
}
```
There are 2 members named use_mkldnn_ , one is AnaysisConfig, the other in PassStrategy.
In Update(), pass_builder_ will be reset with passes_ copied, but in new pass_builder_ instance, use_mkldnn_  value is still false while passes_ has included mkldnn irpasses already, which is what I mean "mismatch".  That's why irpasses were added triple times. 

with your change, passes_ will not be added again, but use_mkldnn_ in PassStrategy will be false.

sfraczek:
I agree with @LeoZhao-Intel that the use_mkldnn_ left in the PassStrategy now has no clear purpose and something should be done about it.

PR #16233, opened 2019-03-15
[MKL-DNN] Fix to crash of Transformer when mkldnn is to be used

jczaja:
This PR is fixing transformer analyzer test when MKL-DNN is to be used. Root cause of crash was that  TensorCopy was not setting MKLDNN primitive descriptor when layout was to be kMKLDNN

### DIFF
--- a/paddle/fluid/inference/tests/api/analyzer_transformer_tester.cc
+++ b/paddle/fluid/inference/tests/api/analyzer_transformer_tester.cc
@@ -183,16 +183,25 @@ void SetInput(std::vector<std::vector<PaddleTensor>> *inputs) {
 }
 
 // Easy for profiling independently.
-TEST(Analyzer_Transformer, profile) {
-  AnalysisConfig cfg;
-  SetConfig(&cfg);
-  std::vector<PaddleTensor> outputs;
-
-  std::vector<std::vector<PaddleTensor>> input_slots_all;
-  SetInput(&input_slots_all);
-  TestPrediction(reinterpret_cast<const PaddlePredictor::Config *>(&cfg),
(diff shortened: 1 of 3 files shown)


luotao1:
Please add `TEST(Analyzer_transformer, compare_mkldnn)` and `TEST(Analyzer_transformer, profile_mkldnn)` like `analyer_resnet50_tester.cc` to verity this PR. And make `test_analyzer_transformer` runs in `SERIAL` in `CMAKELISTS.txt`.

jczaja:
@luotao1 I have updated changes with your suggestions.

paddle/fluid/inference/tests/api/analyzer_transformer_tester.cc
```diff
@@ -183,17 +183,27 @@ void SetInput(std::vector<std::vector<PaddleTensor>> *inputs) {
...
 // Easy for profiling independently.
-TEST(Analyzer_Transformer, profile) {
+void profile(bool use_mkldnn = false) {
   AnalysisConfig cfg;
   SetConfig(&cfg);
   std::vector<PaddleTensor> outputs;
+  if (use_mkldnn) {
+    cfg.EnableMKLDNN();
+    std::unordered_set<std::string> op_list = {"transpose2"};
+    cfg.SetMKLDNNOp(op_list);
```
luotao1: Why should we specify op_list for transformer ut? Could we remove line192-193?

jczaja: @luotao1 Good point. Manual selection of mkl-dnn ops for transformer was removed.

luotao1 (approved):
LGTM! Thanks very much!

PR #16238, opened 2019-03-17
Train with pipe reader

guru4elephant:
a merged version of https://github.com/PaddlePaddle/Paddle/pull/16072 with current develop branch
support multiple queue data reader
support remote file system reading through streaming
support automatically unzip files
support dataset_generator which allows user to define data preprocessing easily


python/paddle/fluid/async_executor.py
```diff
@@ -77,6 +78,12 @@ class AsyncExecutor(object):
     """
 
     def __init__(self, place=None, run_mode=""):
+        """
```
shanyi15, line 81: any comments, please preview first, thanks!
refer to [here](http://wiki.baidu.com/pages/viewpage.action?pageId=705892848)

paddle/fluid/framework/data_feed.cc
```diff
@@ -102,13 +110,14 @@ template <typename T>
 void PrivateQueueDataFeed<T>::ReadThread() {
   std::string filename;
   while (PickOneFile(&filename)) {
-    file_.open(filename.c_str());  // is_text_feed
-    PADDLE_ENFORCE(file_.good(), "Open file<%s> fail.", filename.c_str());
+    int err_no = 0;
+    fp_ = fs_open_read(filename, &err_no, pipe_command_);
+    __fsetlocking(&*fp_, FSETLOCKING_BYCALLER);
+    thread_local string::LineFileReader reader;
```
gavin1332: What's the reader declaration used for?

paddle/fluid/framework/data_feed.cc
```diff
@@ -102,13 +110,14 @@ template <typename T>
...
     T instance;
-    while (ParseOneInstance(&instance)) {
+    while (ParseOneInstanceFromPipe(&instance)) {
```
gavin1332, line 118: I think it is not a good way to decorate  method name "ParseOneInstance" with "FromPipe", for it narrows the data sources. E.g. if I need to create a data feeder parsing data from files or file mappings, they seem not fit for this method. As the result, I need to override ReadThread and append methods like ParseOneInstanceFromMmap in "PrivateQueueDataFeed", such a base class level.
If the method name "ParseOneInstance" could be kept, I can simply override this method in a concrete sub-class  without modifying PrivateQueueDataFeed

guru4elephant: I think you can just write a c++ pipe binary program to use mmap for training. and print out the instance with the same format DataFeed can recognize. We will provide a C++ framework similar to data_generator.py in next version so that users can load binary data into pipe program.

gavin1332: Good comrade, I will try

paddle/fluid/framework/data_feed.cc
guru4elephant: fixed, thx.

paddle/fluid/framework/data_feed.cc
guru4elephant: @@:)

guru4elephant:
design doc：http://agroup.baidu.com/paddlepaddle/md/article/1649944

guru4elephant:
examples：http://wiki.baidu.com/pages/viewpage.action?pageId=717062461

PR #16265, opened 2019-03-18
Add cpu_quantize_placement_pass for C-API quantization

wojtuss:
The `cpu_quantize_placement_pass` is a pass that is run after all graph optimization passes and before quantization passes to apply quantization strategy by setting the `use_quantizer` attribute to `true` for operators which are to be quantized.

This pass is a part of INT8v2 approach with the core implementation submitted in PR https://github.com/PaddlePaddle/Paddle/pull/15987.

test=develop


paddle/fluid/framework/ir/cpu_quantize_placement_pass.cc
```diff
@@ -0,0 +1,56 @@
...
+}
+
+}  // namespace ir
+}  // namespace framework
+}  // namespace paddle
+
+REGISTER_PASS(cpu_quantize_placement_pass,
+              paddle::framework::ir::CPUQuantizePlacementPass)
+    .RequirePassAttr("quantize_enabled_op_types")
+    .RequirePassAttr("quantize_excluded_op_ids");
```
luotao1, line 58: Could you add some description for `quantize_enabled_op_types` and `quantize_excluded_op_ids`?

wojtuss: Done.

luotao1 (approved):
LGTM

PR #16270, opened 2019-03-18
Minor ngraph fix

baojun-nervana:
* update the ngraph engine to catch some edge cases (ops with or w/o feed or fetch)
* use pragma instead of header guard


tensor-tang (approved):
LGTM

PR #16274, opened 2019-03-19
Remove unused variables in op grad maker

sneaxiy:
These ops are found by PR #16211.

### DIFF
--- a/paddle/fluid/framework/scope.cc
+++ b/paddle/fluid/framework/scope.cc
@@ -34,7 +34,7 @@ DEFINE_double(
     "Memory size threshold (GB) when the garbage collector clear tensors."
     "Disabled when this value is less than 0");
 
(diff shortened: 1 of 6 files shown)


paddle/fluid/framework/scope.cc
```diff
@@ -34,7 +34,7 @@ DEFINE_double(
     "Memory size threshold (GB) when the garbage collector clear tensors."
     "Disabled when this value is less than 0");
 
-DEFINE_bool(fast_eager_deletion_mode, false,
+DEFINE_bool(fast_eager_deletion_mode, true,
```
sneaxiy, line 37: Make it true as default. It would be faster.

chengduoZH (dismissed):

chengduoZH (approved):

PR #16275, opened 2019-03-19
inference_shared_library support profile

luotao1:
In demo_ci
- when use static library, we can use command like `/vis_demo --modeldir=../data/mobilenet/model --data=../data/mobilenet/data.txt --refer=../data/mobilenet/result.txt --use_gpu=0 --profile` to print profile result. 
- But when use shared library, the result of same command is `ERROR: unknown command line flag 'profile'`
This PR try to fix this problem.


fc500110 (approved):
LGTM

bingyanghuang:
I built inference engine with Paddle commit ee2f296ef8653488d453f6adf4dede9d4e6a6678 to run lite RNN model, firstly, I cannot get any profile output with the `--profile`. After I merged this PR and rebuilt the Paddle inference lib, I got the right profile output.

PR #16294, opened 2019-03-19
Add support for init_memory and reallocate_memory

zhhsplendid:
Main changes include:

1. Define two flags in paddle/fluid/platform/gpu_info.cc
    FLAGS_gpu_init_memory_in_mb
    FLAGS_gpu_reallocate_memory_in_mb
    then, changed function of max chunk size, max alloc size etc.

2. Changed the RefillPool() in paddle/fluid/memory/detail/buddy_allocator.cc

3. Added unit test which used the two new flags in
    paddle/fluid/memory/allocation/allocator_facade_test.cc

4. Changed corresponding logs.


paddle/fluid/memory/allocation/allocator_facade_test.cc
```diff
@@ -19,20 +19,17 @@
 #ifdef PADDLE_WITH_CUDA
 DECLARE_double(fraction_of_gpu_memory_to_use);
 DECLARE_double(fraction_of_cuda_pinned_memory_to_use);
+DECLARE_uint64(gpu_init_memory_in_mb);
```
sneaxiy, line 22: Maybe change it to be `initial_gpu_memory_in_mb`?  Because we have another flag named `initial_cpu_memory_in_mb`. BTW, define it as double type.

PR #16295, opened 2019-03-19
Add support for init_memory and re-allocate_memory

zhhsplendid:
Main changes include:

Define two flags in paddle/fluid/platform/gpu_info.cc
FLAGS_gpu_init_memory_in_mb
FLAGS_gpu_reallocate_memory_in_mb
then, changed function of max chunk size, max alloc size etc.

Changed the RefillPool() in paddle/fluid/memory/detail/buddy_allocator.cc

Added unit test which used the two new flags in
paddle/fluid/memory/allocation/allocator_facade_test.cc

Changed corresponding logs.


paddle/fluid/platform/gpu_info.cc
```diff
@@ -31,13 +30,32 @@ constexpr static float fraction_of_gpu_memory_to_use = 0.92f;
...
+constexpr static float fraction_reserve_gpu_memory = 0.05f;
+
 DEFINE_double(fraction_of_gpu_memory_to_use, fraction_of_gpu_memory_to_use,
               "Allocate a trunk of gpu memory that is this fraction of the "
               "total gpu memory size. Future memory usage will be allocated "
               "from the trunk. If the trunk doesn't have enough gpu memory, "
               "additional trunks of the same size will be requested from gpu "
               "until the gpu has no memory left for another trunk.");
 
+DEFINE_uint64(gpu_init_memory_in_mb, 0ul,
```
sneaxiy: Maybe we can change it to be `initial_gpu_memory_in_mb` because we have already have a flag named `initial_cpu_memory_in_mb`. BTW, change the type to be `double`.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -31,13 +30,32 @@ constexpr static float fraction_of_gpu_memory_to_use = 0.92f;
...
+              "truck. If the trunk doesn't have enough gpu memory, additional "
+              "trunks of the gpu memory will be requested from gpu with size "
+              "speified by FLAGS_gpu_reallocate_memory_in_mb until the gpu has "
+              "no memory left for the additional trunk. Note: if you set this "
+              "flag, the memory size set by "
+              "FLAGS_fraction_of_gpu_memory_to_use will be overrided by this "
+              "flag. If you don't set this flag, PaddlePaddle will use "
+              "FLAGS_fraction_of_gpu_memory_to_use to allocate gpu memory");
+
+DEFINE_uint64(gpu_reallocate_memory_in_mb, 0ul,
```
sneaxiy: Similar comments as the above flag `gpu_init_memory_in_mb`.

zhhsplendid:
After offline discussed with Jinle, I'm still making some changes on this PR. Will add comment after I finish the change.

zhhsplendid:
Hi Reviewers,

This PR is ready for review again. Please take another look.

**Note**:

Now I implemented **FLAGS_initial_gpu_memory_in_mb** and **FLAGS_reallocate_gpu_memory_in_mb** as override of old fractional initial and re-allocate methods. For example, if user set **FLAGS_reallocate_gpu_memory_in_mb** but doesn't set **FLAGS_initial_gpu_memory_in_mb**, we will allocate the initial memory using old fraction method **FLAGS_fraction_of_gpu_memory_to_use** and re-allocate using **FLAGS_reallocate_gpu_memory_in_mb**.

An alternative implementation is that we only allow user to allocate initial and re-allocate size in same way. If user allocates in **MB** way, user must set both **FLAGS_initial_gpu_memory_in_mb** and **FLAGS_reallocate_gpu_memory_in_mb**. We could add check somewhere for the two MB flags in this PR

I personally prefer the current implementation because users are more free to config initial and re-allocate size, it also matches the meaning of flag names. But I am open to discuss.

paddle/fluid/memory/detail/buddy_allocator.cc
```diff
@@ -71,7 +75,7 @@ void* BuddyAllocator::Alloc(size_t unaligned_size) {
 
   // refill the pool if failure
   if (it == pool_.end()) {
-    it = RefillPool();
+    it = RefillPool(size);
```
sneaxiy, line 79: Maybe better to change destructor as well. Change [https://github.com/PaddlePaddle/Paddle/pull/16295/files#diff-6b33aef9f13c1a2c3753d5d3fabf36adR45](url) to be `system_allocator_->Free(block, block->size(cache_), block->index(cache_));`.

paddle/fluid/memory/allocation/allocator_facade_test.cc
```diff
@@ -82,6 +79,32 @@ TEST(allocator, allocator) {
...
+#endif
+
+  AllocateTestCases();
+}
+
+TEST(Allocator, SpecifyGpuMemory) {
+#ifdef PADDLE_WITH_CUDA
+  // Set to 0.0 to test FLAGS_initial_gpu_memory_in_mb and
+  // FLAGS_reallocate_gpu_memory_in_mb
+  FLAGS_fraction_of_gpu_memory_to_use = 0.0;
```
sneaxiy: These environment variables setting may not take effect, because `AllocatorFacade::Instance()` is a process-wide singleton, which would only be initialized once.

paddle/fluid/memory/detail/buddy_allocator.cc
zhhsplendid: Not only constructor, I also found some similar places when call system_allcator->Free in this file. Please see my new change.

paddle/fluid/memory/allocation/allocator_facade_test.cc
zhhsplendid: Spited into two unit test files.

sneaxiy (approved):
Excellent.

PR #16303, opened 2019-03-19
for Checkpoint save and load

junjun315:
1. change save、load、save_combine、load_combine to opkernal
2.checkpoint save and load


velconia (approved):
LGTM

PR #16322, opened 2019-03-20
fix pattern maching conv2d with(out) ResidualData

wojtuss:
This patch fixes quantization of `conv2d` ops without `ResidualData` input.

test=develop


luotao1 (approved):
LGTM

Issue #16248, opened 2019-03-18
Performance Regression Caused by Tensor Modification.


xiaolil1:
We found performance regression measured by single batch size and 28 cores in our regular test. We root-caused the regression was caused by the code change of "set_format" introduced by Tensor modification, which commit ID is dec9cf53c89e0acc605a053b436ba24be68f62c7.
There are 5% regression for FP32, and over 10% regression for INT8 on CLX (below table):

INT8 ( BS = 1, Cores = 28)| ResNet-50 | MobileNet | SSD-MobileNet
-- | -- | -- | --
Current (dec9c)| 236 | 494 | 169 
Last (08c96) | 265 | 536 | 185
Curr / Last | 0.89 | 0.92 | 0.92

FP32 ( BS = 1, Cores = 28)| ResNet-50 | MobileNet | SSD-MobileNet
-- | -- | -- | --
Current (dec9c)| 97 | 298 | 116 
Last (08c96) | 100 | 309 | 122
Curr / Last | 0.97 | 0.97 | 0.95
 
The code implementation always creates the primitive descriptor in "set_format", which seems unnecessary and brings potential performance overhead. Moreover, “set_format” accepts the default data type FP32, however, the API should consider both FP32 and INT8. (e.g., conv_mkldnn_op.cc:632)

xiaolil1:
@jacek @kbinas @tensor-tang @luotao
Please help to solve this issue, thanks!

xiaolil1:
@kbinias

tensor-tang:
This should be inappropriate adding this hard code in paddle @jczaja :
https://github.com/PaddlePaddle/Paddle/blob/2579ade45fb0d2d698ee4ee0a8f6f4162f9ddb5f/paddle/fluid/platform/mkldnn_utils.h#L27-L47

If MKL-DNN memory structure changed, such as update version, then this should crash at somewhere unknown inside  MKLDNN without any error information in paddle side.

And this should have some bugs, for example, if the dims is larger than 5 which is very normal for users, it should fail and throw `std::exception` at creating MKLDNN pd `mkldnn::memory::primitive_desc(mem_fmt, cpu_engine)` since is not support by MKLDNN.

jczaja:
@tensor-tang This general way of creating non-blocked format was suggested by MKL-DNN team:
https://github.com/intel/mkl-dnn/issues/156#issuecomment-345772775
[...]

PR #16325, opened 2019-03-20
[MKL-DNN] Fix to performance regression of #16248

jczaja:
This PR introduces changes to  performance issue described in #16248 . Changes here Are adapting  mkl-dnn ops to tensor modifications so overhead by setting legacy function: set_format() is removed. 
Also set_format() is removed .

@xiaolil1 If you could run same performance experiement (just Int8 part) as You did in #16248
It would be helpful confirmation that modifications are going into good direction.

### DIFF
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -18,6 +18,7 @@
 #include "paddle/fluid/operators/math/math_function.h"
 #ifdef PADDLE_WITH_MKLDNN
 #include "paddle/fluid/platform/mkldnn_helper.h"
+#include "paddle/fluid/platform/mkldnn_utils.h"
 #endif
 
 namespace paddle {
@@ -145,18 +146,30 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
   // output tensor has the same dims as input. Reorder don't change dims
   out->Resize(in.dims());
 
-  // tempory mem pd fr out , to make reorder
-  auto out_mem_pd = paddle::platform::create_prim_desc_from_dims(
-      paddle::framework::vectorize2int(out->dims()),
-      mkldnn::memory::format::blocked, out_type);
-  if (in.get_mkldnn_prim_desc() != out_mem_pd) {
-    void* in_data = GetDataFromTensor(in, in_type);
-    auto out_data = out->mutable_data(expected_kernel_type.place_, in.type());
+  auto in_format = in.format();
+  auto out_format = platform::mkldnn_fmt(in_tz.size());
 
-    auto in_memory = memory(in.get_mkldnn_prim_desc(), in_data);
-    auto out_memory = memory(out_mem_pd, out_data);
+  // Comparing formats is faster than comparing
+  // MKLDNN Memory Primitive Descriptors (PD)
+  // So for non-blocked we can use format comparison
+  if (out_format == mkldnn::memory::format::blocked ||
+      out_format != in_format) {
+    // just in case one of tensors' formats is blocked
+    // create temporary PD and make comparison of PDs
+    auto out_mem_pd = paddle::platform::create_prim_desc_from_dims(
+        paddle::framework::vectorize2int(out->dims()),
+        mkldnn::memory::format::blocked, out_type);
+    if (in.get_mkldnn_prim_desc() != *out_mem_pd) {
+      void* in_data = GetDataFromTensor(in, in_type);
+      auto out_data = out->mutable_data(expected_kernel_type.place_, in.type());
 
-    platform::Reorder(in_memory, out_memory);
+      auto in_memory = memory(in.get_mkldnn_prim_desc(), in_data);
+      auto out_memory = memory(*out_mem_pd, out_data);
+
+      platform::Reorder(in_memory, out_memory);
+    } else {
+      out->ShareDataWith(in);
+    }
   } else {
     out->ShareDataWith(in);
   }
--- a/paddle/fluid/framework/data_layout_transform.h
+++ b/paddle/fluid/framework/data_layout_transform.h
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <map>
+#include <unordered_map>
 #include <vector>
 #include "paddle/fluid/framework/op_kernel_type.h"
 #include "paddle/fluid/framework/tensor.h"
@@ -36,17 +37,6 @@ inline MKLDNNFormat ToMKLDNNFormat(const DataLayout& layout) {
     default:
       PADDLE_THROW("Fail to convert layout %s to MKLDNN format",
                    DataLayoutToString(layout));
-  }
-}
-
-inline DataLayout ToPaddleLayout(const MKLDNNFormat& format) {
-  switch (format) {
-    case MKLDNNFormat::nhwc:
-      return DataLayout::kNHWC;
-    case MKLDNNFormat::nchw:
-      return DataLayout::kNCHW;
-    default:
-      PADDLE_THROW("Fail to convert MKLDNN format to paddle layout");
   }
 }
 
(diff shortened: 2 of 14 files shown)


jczaja:
Added modifcations of other int8 ops: transpose int8, quantize, dequantize and requantize

tensor-tang:
@jczaja Thanks for your quick solution.

But please fix this issue in one PR and make sure **all the problems** would be fixed in #16248 .

Then the PR cloud be merged.

And make sure we have backup plan for reverting the changes as we discussed before @jianhang-liu @jczaja

jczaja:
Updated PR with adapting elementwise_mul_mkldnn (last missing op) and removing set_format() from Tensor.

ghost:
@tensor-tang CI fail due to "need approval since tensor.h is modified". Please help to check. All changed needed are all included in this PR. set_format() was removed. Thanks!

xiaolil1:
@jczaja @tensor-tang @Sand3r-
hi @jczaja , I found you removed the "set_format" API in this PR, this change may block some features like "dequantize op" and "reshape op" MKL-DNN support. For example, for Dequantize OP, we don't need to set MKLDNN layout for the output, because it will bring unnecessary data transform from "fp32 nchw mkldnn" to "fp32 nchw cpu", and for "Reshape OP", we will change the output format as output dims size without "dst_pd" creation. So i think current  "set_mkldnn_prim_desc" couldn't involve all these necessary functions. How about we keep the "set_format" API and remove the pd creation from it for performance consideration? Or do you have plan to support all these functions in "set_mkldnn_prim_desc"?

ghost:
@tensor-tang All exiting OPs were all updated. As a result, set_format() was fully removed. For future OPs, it should follow the new way (i.e. set_mkldnn_prim_desc). PD was cached in the 1st creation so no performance drop will exist.

ghost:
@jczaja @tensor-tang @kbinias To be brief, with all modification in this PR, performance drop "caused by Tensor Modification" should be **gone**. This is a "complete" change. 
Meanwhile, even with this PR, there are STILL some "minor" performance drop in FP32. It's quite possiblly caused by MKLDNN upgrade (to v0.18). Krysztof will measure "latest code + change in this PR + revert back to MKLDNN v0.17" to further prove. He should post the result today.
Suggest to merge this PR since it already done it's job. Meanwhile, keep checking whether there is other performance drop caused by other commit.

xiaolil1:
> @tensor-tang All exiting OPs were all updated. As a result, set_format() was fully removed. For future OPs, it should follow the new way (i.e. set_mkldnn_prim_desc). PD was cached in the 1st creation so no performance drop will exist.
For "Dequantize OP", replace "set_format" with "set_mkldnn_prim_desc" will bring unnecessary reorder, which will decrease INT8 latency performance, and for "MKLDNN Reshape OP", format need to be changed as output dims, which also needs the "set_format" API.

xiaolil1:
@tensor-tang  For Dequantize OP, if use "set_format" the op chain is "dequantize -> fp32 op", but if use "set_mkldnn_prim_desc", the op chain will be "dequantize->data transform (from mkldnn to cpu) -> fp32 op", one unnecessary reorder will decrease the latency.

tensor-tang:
@jianhang-liu @hshen14 @jczaja 

>  For Dequantize OP, if use "set_format" the op chain is "dequantize -> fp32 op", but if use "set_mkldnn_prim_desc", the op chain will be "dequantize->data transform (from mkldnn to cpu) -> fp32 op", one unnecessary reorder will decrease the latency.

What do you think about this problem?

hshen14:
> @jianhang-liu @hshen14 @jczaja
> 
> > For Dequantize OP, if use "set_format" the op chain is "dequantize -> fp32 op", but if use "set_mkldnn_prim_desc", the op chain will be "dequantize->data transform (from mkldnn to cpu) -> fp32 op", one unnecessary reorder will decrease the latency.
> 
> What do you think about this problem?

Xiaoli's finding is correct. Can we keep set_format and make it co-exist with set_mkldnn_prim_desc? From API perspective, set_mkldnn_prim_desc and set_format provides their own responsibilities which are required in different use cases. DequantizeOP is the case that needs set_format directly to avoid the extra reorder.

jczaja:
@xiaolil1 , @hshen14 
Following the code from this PR. Three will be no additional reorder. When NCHW MKLDNN is to be transformed to NCHW cpu. Tensor::ShareData will be called , which does not perform reorder:
https://github.com/jczaja/Paddle/blob/fd76c5961690695802228526cf1a1032840c9272/paddle/fluid/framework/data_layout_transform.cc#L174

tensor-tang:
Thanks for your quick actions of this PR @jczaja @kbinias  @jianhang-liu.
Thanks for your deep investigations and experiments @xiaolil1 @hshen14 .

From our perspective, the purpose of adding `set_mkldnn_prim_desc` is to remove `set_format`. And it turns out does not work well for all of MKLDNN cases.
As of the performance issue of #16248 , it seems still remain about 9% fp32 regression and need go deeper.

So, as a hotfix of #16248 this PR helps, but for long term it does not work fine.
We should discuss more about this.

jczaja:
@tensor-tang , @xiaolil1 After profiling it was revealed that mkldnn prim desc creation time increased and I cannot explain reasons for that. But we had it on our todo list to hide primitive desc creation eg. reuse it . Integer conv does reuse once created prim desc , so that is why performance problem does not manifest there.

I can provide performance fix  but It will be ready Thursday EOD this week.

xiaolil1:
> > @jianhang-liu @hshen14 @jczaja
> > > For Dequantize OP, if use "set_format" the op chain is "dequantize -> fp32 op", but if use "set_mkldnn_prim_desc", the op chain will be "dequantize->data transform (from mkldnn to cpu) -> fp32 op", one unnecessary reorder will decrease the latency.
> > 
> > 
> > What do you think about this problem?
> 
> Xiaoli's finding is correct. Can we keep set_format and make it co-exist with set_mkldnn_prim_desc? From API perspective, set_mkldnn_prim_desc and set_format provides their own responsibilities which are required in different use cases. DequantizeOP is the case that needs set_format directly to avoid the extra reorder.

We found the mechanism of data transform was changed from "layout" judgement to "pd" judgement, so the reorder issue we met before has gone, and we have verified it.

jczaja:
@tensor-tang I want to explain the reasons for Tensor modifications. There were four reasons that led to making up such a decision:
1. After implementing Transpose MKL-DNN op , using Reorder , led to appearance of Memory formats: mkldnn::memory::format::blocked , which crashed DataTransformer. But this was hacked, by forcing Transpose MKL-DNN op to produce non MKL-DNN tensors eg. layout=Paddle, format: undef.
2. Will to tidy up all "hacky" guessing of what format to use:
```
  auto src_format =
      src_tz.size() == 2 ? mkldnn::memory::format::nc : x->format();
```
or this:
https://github.com/PaddlePaddle/Paddle/blob/4f2278f03264f96c5ed165dc17b4755b53cd1cdc/paddle/fluid/platform/mkldnn_helper.h#L110

3. By using input tensor PD in ops, instead of recreating one, We expected code to be a bit more tidy and performance should be tiny better (no recreation PD from format). This turned out to be only important for ops that do not  use reusing or not cache primitive descriptors


4. Prepare paddlepaddle for mkl-dnn 1.0
In section 5 Memory descriptor rework of MKL-DNN 1.0 RFC:
(https://github.com/intel/mkl-dnn/tree/rfc-api-changes-v1.0/doc/rfc/api-v1.0)

> Users will still be able to initialize memory descriptors with format tags like nchw, but the memory_desc_t.format_kind will be set to a canonicalized kind like blocked, and **the format name will not be recorded in the memory descriptor structure**. Initialization with strides will always result in blocked format. The API will also use different types for memory format tags and kinds to aid correctness.

My reading of this is that we no longer will be able to read format name from Primitive Descriptors . For example when creating primitive descriptor of convolution we pass "any" as we want primitive to get
best implementation for our situation. Currently we get NCHW16C  , but in new API we may not be able to get that "NCHW16C" , so we cannot do comparison based on it. Hence Using primitive descriptors to comparison was step not to have such a problem. I'm currently trying to get more details on that from MKL-DNN.

PR #16326, opened 2019-03-20
Add enabling quantization

wojtuss:
This patch adds option to enable `cpu_quantize_placement_pass` applying quantization strategy.

test=develop

### DIFF
--- a/paddle/fluid/inference/api/paddle_pass_builder.h
+++ b/paddle/fluid/inference/api/paddle_pass_builder.h
@@ -30,6 +30,10 @@ class PaddlePassBuilder {
   explicit PaddlePassBuilder(const std::vector<std::string> &passes)
       : passes_(passes) {}
 
+  void SetPasses(std::initializer_list<std::string> passes) {
+    passes_ = passes;
+  }
+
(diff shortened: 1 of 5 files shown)


paddle/fluid/inference/api/paddle_pass_builder.h
```diff
@@ -30,6 +30,10 @@ class PaddlePassBuilder {
   explicit PaddlePassBuilder(const std::vector<std::string> &passes)
       : passes_(passes) {}
 
+  void SetPasses(std::initializer_list<std::string> passes) {
+    passes_ = passes;
+  }
```
luotao1: why use `SetPasses`? I don't find it used in #15987 as well.

wojtuss: The `SetPasses` is used in `quantizer.cc` (https://github.com/PaddlePaddle/Paddle/pull/15987/files?file-filters%5B%5D=.cc#diff-2c882d108544e86a1f517bd247ac68ceR352). I have removed its definition from this PR.

luotao1 (approved):
LGTM

PR #16004, opened 2019-03-01
Cache the chosen kernel of operators'.

Xreki:

PR #16339, opened 2019-03-21
Cache the chosen kernel of operators

luotao1:
related #16004
In a same name scope and same place, since the expected kerenl of this Op does not change in the execution, it could be recorded only at the first iteration of this Op's execution to save the elapsed time.

In the inference of PyramidDNN (a small model):

|---|---|---|
|before|  0.212784 | 0.179449 |
|after| 0.208145  | 0.170931 |
|speedup| 2.1%  | 4.7% |


luotao1:
The speedup of this PR is not better than https://github.com/PaddlePaddle/Paddle/pull/16004#issuecomment-468685738, maybe the reason is the cost of `HasAttr`. The `find` in `std::unorder_map` is slow: https://stackoverflow.com/questions/42588264/why-is-stdunordered-map-slow-and-can-i-use-it-more-effectively-to-alleviate-t.
If I remove `!HasAttr(kEnableCacheExpectedKernel)` in https://github.com/PaddlePaddle/Paddle/pull/16339/files#diff-bb4dff0e76921ba253151a33c5b6d1f4R898, the result is

|---|---|---|
|before|  0.212784 | 0.179449 |
|after| 0.201955  | 0.162867 |
|speedup| 5%  | 9.2% |

The speedup is 2X than use `HasAttr`.

luotao1:
http://ce.paddlepaddle.org:8080/viewLog.html?buildId=33382&buildTypeId=PaddleModesl_ModelDebug_Model11Build&tab=buildResultsDiv
CE已通过。

luotao1:
```
[05:37:22][Step 1/1] current pr 16339 got approvals: FALSE
[05:37:22][Step 1/1] You must have one RD (velconia,panyx0718,XiaoguangHu01,chengduoZH,Xreki,luotao1,sneaxiy,tensor-tang,jacquesqiao,typhoonzero) approval for the api change! paddle/fluid/framework/operator.h
[05:37:32][Step 1/1] Process exited with code 1
```

luotao1:
It seems that the develop branch is slower than before. The result of this PR is:

| | 2620 v3 (ms per sample) |
|---|---|
|before| 0.186852 |
|after| 0.179725  | 
|speedup| 3.8%  |

Xreki (approved):
LGTM

PR #16342, opened 2019-03-21
Remove reshape and transpose operators from attention module

yihuaxu:
According to the performance status of Bert/Transformer model, fused matmul/reshape/transpose operators to reduce memory's copy.

**Platform:** Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz
**Model Path:** third_party/inference_demo/bert_emb128/model
**Batch Size:** 1
**Command:** ./paddle/fluid/inference/tests/api/test_analyzer_bert --infer_model=third_party/inference_demo/bert_emb128/model/ --infer_data=third_party/inference_demo/bert_emb128/data.txt --gtest_filter=Analyzer_bert.profile --paddle_num_threads=1 --repeat=10 --batch_size=1 --test_all_data
**Data Source:** third_party/inference_demo/bert_emb128/data.txt.

The following is the comparison with the different scenarios.

image

**Model Comparison:**
(a).Before Optimization:
image

(b).After Optimization:
image

**Reference:**
[Can we avoid head split_merge in Transformer.pdf](https://github.com/PaddlePaddle/Paddle/files/3088454/Can.we.avoid.head.split_merge.in.Transformer.pdf)
image


yihuaxu:
start a review

ghost:
@tensor-tang Please help to review this PR. This is one critical patch for BERT (and apply to Transformer) also. Thanks!

yihuaxu:
@tensor-tang Please help us review this PR and give some suggestion. Thanks a lot!

luotao1:
> The following is the comparison with the different scenarios.

Do you have the model level comparison before and after this PR?

yihuaxu:
> > The following is the comparison with the different scenarios.
> 
> Do you have the model level comparison before and after this PR?

Just updated the description included the comparison of this model

paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -137,7 +137,8 @@ CpuPassStrategy::CpuPassStrategy() : PassStrategy({}) {
                   // following two passes should be located in the last, since
                   // they will work on all fused ops.
                   "expected_kernel_cache_pass",  //
-                  "runtime_context_cache_pass"});
+                  "runtime_context_cache_pass",  //
+                  "fuse_reshape_transpose_scale_matmul_pass"});
```
luotao1: see line137, put `fuse_reshape_transpose_scale_matmul_pass` before `expected_kernel_cache_pass`.

paddle/fluid/operators/math/blas_impl.h
```diff
@@ -563,6 +563,35 @@ void Blas<platform::CPUDeviceContext>::BatchedGEMM(
...
+#endif
+  }
+  for (int k = 0; k < batchCount; ++k) {
+    auto *Ak = (*a_array)[k];
+    auto *Bk = (*b_array)[k];
+    auto *Ck = (*c_array)[k];
+    this->template GEMM<T>(transA == CblasTrans, transB == CblasTrans, M, N, K,
+                           alpha, Ak, lda, Bk, ldb, beta, Ck, ldc);
+  }
+}
```
luotao1, line 593: - What's the difference between old `BatchedGEMM` and new `BatchedGEMM` in your PR?
- I see the difference is the input format from `const T` to `std::vector<const T *> *a_array`.
- Could you reuse the old one or unify them? Or why do you create a new one?

Same for the `Matmul`.
There is already 
```
void Blas<DeviceContext>::MatMul(const framework::Tensor &mat_a,
                                 const MatDescriptor &dim_a,
                                 const framework::Tensor &mat_b,
                                 const MatDescriptor &dim_b, T alpha,
                                 framework::Tensor *mat_out, T beta) 
```

paddle/fluid/operators/matmul_op.cc
```diff
@@ -339,6 +609,14 @@ class MatMulOpMaker : public framework::OpProtoAndCheckerMaker {
         )DOC")
         .SetDefault(false);
     AddAttr<float>("alpha", "The scale of Out").SetDefault(1.0f);
+    AddAttr<std::vector<int>>("last_dim",
+                              "(vector<int>) "
+                              "the shape of inputs and output.")
+        .SetDefault(std::vector<int>{-1, -1, -1});
+    AddAttr<bool>("is_test",
+                  "(bool, default false) Set to true for inference only, false "
+                  "for training. Some layers may run faster when this is true.")
```
luotao1, line 620: Why matmul need `is_test` attribute?
Why add `last_dim` attribute?

paddle/fluid/operators/math/blas_impl.h
yihuaxu: > * What's the difference between old `BatchedGEMM` and new `BatchedGEMM` in your PR?
**Transfer the arrays of input and output  into BatchedGEMM directly.**
> * I see the difference is the input format from `const T` to `std::vector<const T *> *a_array`.
**According to the transpose dimensions's difference and the stride's requirement,  the array calculation of MKL BatchedGEMM need be get though the special calculation. So it tends to move the calculation into the internal of  matmul operator.**
> * Could you reuse the old one or unify them? Or why do you create a new one?
**The initial idea is that it can avoid the completion into the common blas's implementation.  If we need implement this or others  array's calculation into blas_impl.h, it can not keep the code's clean.**
> 
> Same for the `Matmul`.
> There is already
> 
> ```
> void Blas<DeviceContext>::MatMul(const framework::Tensor &mat_a,
>                                  const MatDescriptor &dim_a,
>                                  const framework::Tensor &mat_b,
>                                  const MatDescriptor &dim_b, T alpha,
>                                  framework::Tensor *mat_out, T beta) 
> ```

paddle/fluid/operators/matmul_op.cc
yihuaxu: > Why matmul need `is_test` attribute?
**Add the "is_test" attribute for inference mode and don't influence other requirement.** 
> Why add `last_dim` attribute?
**To decrease the count of matmul operator's attributes, but it will result in that it is only for the special dimensions of reshape and transpose.**

paddle/fluid/inference/api/paddle_pass_builder.cc
yihuaxu: done

paddle/fluid/operators/matmul_op.cc
luotao1: > Add the "is_test" attribute for inference mode and don't influence other requirement.

Matmul is a common and base op, and it should not have the difference between train and inference.

ghost: @luotao1 Score (inference) and Forward part of Training sometimes have difference. "is_test" attribute is used to distinguish between them and it's widely used in many OPs. For example:
* Batch Norm: Use fixed mean/variance instead of computing on batch
* Softmax: skip epson for performance improvement
* sequence_pool: don't create index buffer for performance improvement

The optimization here for attention (i.e. remove transpose/reshape via enhanced MatMul) only need apply to Inference only at this time. So we add "is_test" to contain our code.

In case this optimization need be applied to training also, we can add backward part and remove this "is_test" in fwd.

bingyanghuang:
@jianhang-liu This PR should be moved to Release 1.6

paddle/fluid/operators/math/blas.h
```diff
@@ -176,11 +177,24 @@ class Blas {
...
+                   int batchCount) const;
+
   template <typename T>
   void MatMul(const framework::Tensor& mat_a, const MatDescriptor& dim_a,
               const framework::Tensor& mat_b, const MatDescriptor& dim_b,
               T alpha, framework::Tensor* mat_out, T beta) const;
 
+  template <typename T>
+  void MatMul(std::vector<const T*>* a_array, const MatDescriptor& dim_a,
+              const int ld_a, std::vector<const T*>* b_array,
```
tensor-tang, line 194: why matmul input should be a vector?

paddle/fluid/operators/math/blas.h
```diff
@@ -176,11 +177,24 @@ class Blas {
                    int K, T alpha, const T* A, const T* B, T beta, T* C,
                    int batchCount, int64_t strideA, int64_t strideB) const;
 
+  template <typename T>
+  void BatchedGEMM(CBLAS_TRANSPOSE transA, CBLAS_TRANSPOSE transB, int M, int N,
+                   int K, T alpha, std::vector<const T*>* a_array,
```
tensor-tang, line 182: `const std:vector<const T*>* ` ?

paddle/fluid/operators/matmul_op.cc
```diff
@@ -54,13 +228,69 @@ class MatMulKernel : public framework::OpKernel<T> {
     auto *out = context.Output<framework::Tensor>("Out");
     out->mutable_data<T>(context.GetPlace());
 
+    auto transpose_x = context.Attr<bool>("transpose_X");
+    auto transpose_y = context.Attr<bool>("transpose_Y");
+    auto dim_x = x.dims();
+    auto dim_y = y.dims();
+
+    int ld_x = 0, ld_y = 0, ld_out = 0;
+    std::vector<const T *> a_array, b_array;
```
tensor-tang, line 237: maybe could use `const T (*p)[n]` to avoid using std:vector? so can reuse some matmul function in blas.h?

tensor-tang:
Such a good point and thanks to the foresight  `GEMM`!

paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -136,7 +136,8 @@ CpuPassStrategy::CpuPassStrategy() : PassStrategy({}) {
                   "is_test_pass",                  //
                   // following two passes should be located in the last, since
                   // they will work on all fused ops.
-                  "expected_kernel_cache_pass",  //
+                  "expected_kernel_cache_pass",                //
+                  "fuse_reshape_transpose_scale_matmul_pass",  //
```
tensor-tang, line 140: why add at last?

this is a very big fuse, maybe should be earlier.

GaoWei8:
This PR is tested on Ernie in CPU and the num of threads is 20. The original time is 72.5432ms and the time decrease to 69.9925ms after optimization.

GaoWei8:
This PR is tested on Ernie in CPU and the num of threads is set as 20. 
The original time after fuse is 72.5432ms (without this PR) and the time decreases to 69.9925ms with merged this PR.
Is this situation correct？

paddle/fluid/inference/api/paddle_pass_builder.cc
yihuaxu: Now Bingyang are ready to re-implement the pass in future. This PR will be aborted.

paddle/fluid/operators/math/blas.h
yihuaxu: Now Bingyang are ready to re-implement the pass in future. This PR will be aborted.

paddle/fluid/operators/math/blas.h
yihuaxu: Now Bingyang are ready to re-implement the pass in future. This PR will be aborted.

paddle/fluid/operators/matmul_op.cc
yihuaxu: Now Bingyang are ready to re-implement the pass in future. This PR will be aborted.

PR #16409, opened 2019-03-23
Enhance gc to support deleting tensor buffer in advance

sneaxiy:
For example, gc can collect data buffer of input `Y` in `elementwise_add_grad` op before this op runs. In the meanwhile, shape and lod of `Y` can be kept when `elementwise_add_grad` op runs.

Op developers can use `DECLARE_NO_NEED_BUFFER_VARS_INFERENCE` to declare a class that indicates the unused-buffer tensors.

For example, inputs of concat op `x0`, `x1`, `x2` can be deleted before concat_grad op runs:
image


paddle/fluid/framework/details/op_registry.h
```diff
@@ -156,6 +216,18 @@ struct OpInfoFiller<T, kInplaceOpInference> {
   }
 };
 
+template <typename T>
+struct OpInfoFiller<T, kNoNeedBufferVarsInference> {
+  void operator()(const char* op_type, OpInfo* info) const {
+    info->infer_no_need_buffer_vars_ = [](const VariableNameMap& inputs,
```
liupluswei, line 222: curious about when will these three parameters be used to get the NoNeedBufferVars, seems now we just return the parameters specified in the macro as an unordered_set?

sneaxiy: I reserve these parameters for future use. Some ops may not need some forward inputs or outputs when some attribute is true/false. For example, batch_norm_grad_op does not need Bias when use_mkldnn is false.

paddle/fluid/framework/details/op_registry.h
```diff
@@ -36,27 +37,86 @@ enum OpInfoFillType {
...
 };
 
+namespace internal {
+template <typename T, OpInfoFillType kType>
+struct TypePair {
+  using Type = T;
+  static constexpr OpInfoFillType kFillType = kType;
+};
+
+using OpRegistryClasses = std::tuple<                                // NOLINT
```
sneaxiy, line 52: Ugly but scalable codes here. I rewrite `OpInfoFillTypeID::ID()` method because the character number limit is set to be 80 in a line.

paddle/fluid/operators/add_position_encoding_op.cc
```diff
@@ -39,13 +40,8 @@ class AddPositionEncodingOpGrad : public framework::OperatorWithKernel {
...
-    PADDLE_ENFORCE(ctx->HasInput("X"), "X(Input) must not be null.");
-    PADDLE_ENFORCE(ctx->HasInput("Out"), "Out must not be null.");
-    PADDLE_ENFORCE(ctx->HasInput(framework::GradVarName("Out")),
-                   "Out@GRAD must not be null.");
-
-    auto out_dims = ctx->GetInputDim("Out");
     if (ctx->HasOutput(framework::GradVarName("X"))) {
+      auto out_dims = ctx->GetInputDim(framework::GradVarName("Out"));
       ctx->SetOutputDim(framework::GradVarName("X"), out_dims);
     }
```
chengduoZH, line 46: Maybe the above code is confusing, if the `ctx->HasOutput(framework::GradVarName("X")` is False, there not need call `AddPositionEncodingOpGrad `.

chengduoZH (dismissed):

chengduoZH (approved):

paddle/fluid/operators/add_position_encoding_op.cc
sneaxiy: I am confused with these codes too. I just follow the original logic.

paddle/fluid/framework/details/reference_count_pass.cc
```diff
@@ -229,17 +302,43 @@ std::unique_ptr<ir::Graph> ReferenceCountPass::ApplyImpl(
...
-      auto result = ExtractComputationOpFromLastLivedVar(
-          name_var_pair.second.back(), i, shrink_func, &ok);
+      auto &var_name = name_var_pair.first;
+      auto &var_handles = name_var_pair.second;
+
+      for (auto iter = var_handles.rbegin(); iter != var_handles.rend();
+           ++iter) {
+        bool ok;
+        auto result =
+            ExtractComputationOpFromLastLivedVar(*iter, i, shrink_func, &ok);
```
liupluswei, line 312: actually I am a little bit confused about ExtractComputationOpFromLastLivedVar function. I know it want to get the last op which will use this variable, but in reference count part, do we need to care about which op generate this variable?  What's the impact if we only care about the ops use this variable (the variable is used in the input part)

paddle/fluid/framework/details/reference_count_pass.cc
```diff
@@ -193,6 +193,79 @@ ExtractComputationOpFromLastLivedVar(VarHandle *var, size_t scope_idx,
...
+        occurred_in_other_vars = true;
+        break;
+      }
+    }
+
+    if (occurred_in_other_vars) {
+      continue;
+    }
+
+    for (auto &out_pair : op_base->Outputs()) {
```
liupluswei, line 243: Do we need to care about outputs here?

sneaxiy: Yes, we should care about outputs for in-place operation.

paddle/fluid/framework/details/reference_count_pass.cc
sneaxiy: `ExtractComputationOpFromLastLivedVar` returns (1) the last ops which read this variable if read ops exist, or (2) the last one op which writes this variables if read op does not exist. See [HERE](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/details/reference_count_pass.cc#L154) in details.

PR #16411, opened 2019-03-24
link the linux-x86_64 libwbaes.so into paddle

chenwhql:


luotao1:
please use `test=develop` in commit to trigger the CI

PR #16424, opened 2019-03-25
Fix allocator bug

sneaxiy:
Use `decorated_allocators_` to fix allocator bug described in #16086.


paddle/fluid/memory/allocation/allocator.h
```diff
@@ -47,10 +49,14 @@ class Allocator;
 class Allocation {
  public:
   Allocation(void* ptr, size_t size, platform::Place place)
-      : allocator_(nullptr), ptr_(ptr), size_(size), place_(place) {}
+      : ptr_(ptr), size_(size), place_(place) {
+    decorated_allocators_.reserve(8);
```
chengduoZH, line 92: Why do you reserve `decorated_allocators_ ` as 8?

paddle/fluid/memory/allocation/allocator.h
```diff
@@ -72,17 +78,31 @@ class Allocation {
...
 
-  void set_allocator(Allocator* allocator) { allocator_ = allocator; }
+  inline void RegisterDecoratedAllocator(Allocator* allocator) {
+    decorated_allocators_.push_back(allocator);
+  }
 
-  virtual ~Allocation();
+  inline void PopDecoratedAllocator() { decorated_allocators_.pop_back(); }
+
+  inline Allocator* TopDecoratedAllocator() {
```
chengduoZH, line 133: The name of `TopDecoratedAllocator` is confusing, it seems that the function of this functor returns the current Allocator.

paddle/fluid/memory/allocation/allocator.h
```diff
@@ -132,9 +152,12 @@ class Allocator {
   // True if the `Allocate` is thread safe.
   virtual bool IsAllocThreadSafe() const;
 
+  // This function should not be called outside
```
panyx0718, line 194: outside what? why public

paddle/fluid/memory/allocation/buffered_allocator.h
```diff
@@ -31,7 +31,7 @@ namespace allocation {
 // underlying_allocator_
 class BufferedAllocator : public Allocator {
  public:
-  explicit BufferedAllocator(std::unique_ptr<Allocator> &&allocator);
+  explicit BufferedAllocator(std::shared_ptr<Allocator> allocator);
```
panyx0718, line 34: why shared?

paddle/fluid/memory/allocation/cpu_allocator.h
```diff
@@ -31,19 +31,13 @@ namespace allocation {
...
-class CPUAllocator;
-class CPUAllocation : public Allocation {
- public:
-  CPUAllocation(void* ptr, size_t size);
-};
-
 class CPUAllocator : public Allocator {
  public:
-  constexpr static size_t kAlignment = 64u;
+  constexpr static size_t kAlignment = 4096UL;
```
panyx0718, line 36: why change?

paddle/fluid/memory/allocation/allocator.cc
```diff
@@ -27,16 +27,24 @@ bool Allocator::IsAllocThreadSafe() const { return false; }
...
 }
 
-void Allocator::Free(Allocation* allocation) { delete allocation; }
+void Allocator::FreeImpl(Allocation* allocation) {
+  Allocator* allocator = allocation->TopDecoratedAllocator();
+  allocator->Free(allocation);
+}
+
+void Allocator::Free(Allocation* allocation) {
+  allocation->PopDecoratedAllocator();
```
panyx0718, line 40: is this poping the last allocator? what not pop in FreeImpl

sneaxiy: `FreeImpl` is a virtual function that must be overwritten by derived class. If we pop allocator in `FreeImpl`, each derived class must write this code in their implemented function. I just want to hide the implementation of `decorated_allocators_` in base class.

paddle/fluid/memory/allocation/allocator.h
sneaxiy: Because some decorated allocator would call something like `underlying_allocator_->Free(...)`. Comment above is saying that we should not call this method outside the implementation of allocator.

paddle/fluid/memory/allocation/buffered_allocator.h
sneaxiy: Because some allocator accepts `std::shared_ptr<Allocator>` when constructing. We can cast `std::unique_ptr<Allocator>` to `std::shared_ptr<Allocator>` easily, but it is impossible to do the reverse casting.

paddle/fluid/memory/allocation/cpu_allocator.h
sneaxiy: Because legacy CPU allocator aligns to 4096.

paddle/fluid/memory/allocation/allocator.h
sneaxiy: Add comment to explain. It is designed to prevent frequently heap allocation.

paddle/fluid/memory/allocation/allocator.h
sneaxiy: `decorated_alllocators_` is a stack, so the method is named `Top`, `Pop`, etc, following the naming rules of `std::stack`.

paddle/fluid/memory/allocation/allocator.h
```diff
@@ -44,13 +46,56 @@ class Allocator;
...
+ *
+ * Suppose we have an allocator which is decorated by several allocators:
+ *
+ *   A(1) <- A(2) <- A(3) <- ... <- A(n)
+ *
+ * , and the public allocator is A(1).
+ *
+ * The allocation process would be:
+ *
+ *   A(n).Allocate() -> ... -> A(2).Allocate() -> A(1).Allocate()
```
panyx0718, line 68: should the public Allocator be called first? so public Allocator can do some caching and retry before calling the underlying allocators

panyx0718 (approved):

paddle/fluid/memory/allocation/allocator.h
sneaxiy: Yes. I am sorry that I have not declared clearly. The actual allocating process would be:

A(1).Allocate() called -> A(2).Allocate() called -> .... A(n).Allocate() called 
-> 
A(n).Allocate() returned -> ... -> A(2).Allocate() returned -> A(1).Allocate() returned


The public allocator should call first. But the allocating chain can break if A(i).AllocateImpl() does not call A(i+1).Allocate(). For example, BufferedAllocator would break the allocating chain and return the cached memory.

PR #16438, opened 2019-03-25
Move cpu_quantize_* passes into mkldnn subfolder

wojtuss:
As currently the `cpu_quantize_*` passes runs only with MKL-DNN, this patch moves them to the `mkldnn` subfolder.

test=develop


wojtuss:
@luotao1 , there are problems on CI not related to the PRs

luotao1:
Yes, we are repairing CI now.

luotao1 (approved):
LGTM

PR #16443, opened 2019-03-25
Fix allreducedep bug

gongweibao:
Error:
image

Now:
image

性能对比：

|    bert    | step/s |
| ---------- | --- |
| before |  0.86*32 |
| after       |  1.47*32 |

### DIFF
--- a/paddle/fluid/framework/details/all_reduce_deps_pass.cc
+++ b/paddle/fluid/framework/details/all_reduce_deps_pass.cc
@@ -52,13 +52,30 @@ std::unique_ptr<ir::Graph> AllReduceDepsPass::ApplyImpl(
   //               Note that must assert topology sort is stable
   auto& ops = graph->Get<const std::vector<OpDesc*>>(kStaleProgramOpDescs);
   for (auto* op_desc : ops) {
-    auto outputs = op_desc->Outputs();
-    for (auto& o_it : outputs) {
-      for (auto& v : o_it.second) {  // values
-        vars[v] = order;
+    try {
+      bool is_bk_op =
+          static_cast<bool>(boost::get<int>(op_desc->GetAttr(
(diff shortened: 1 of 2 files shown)


paddle/fluid/framework/details/all_reduce_deps_pass.cc
```diff
@@ -52,13 +53,28 @@ std::unique_ptr<ir::Graph> AllReduceDepsPass::ApplyImpl(
...
+
+      auto backward_vars =
+          boost::get<std::vector<std::string>>(op_desc->GetNullableAttr(
+              OpProtoAndCheckerMaker::OpRoleVarAttrName()));
+      PADDLE_ENFORCE_EQ(backward_vars.size() % 2, 0);
+
+      auto outputs = op_desc->Outputs();
+      for (auto& o_it : outputs) {
+        for (auto& v : o_it.second) {  // values
+          vars[v] = order;
```
chengduoZH, line 71: Why do you record the gradient but not parameter here? because the gradient's name may be changed after mem_opt.

paddle/fluid/framework/parallel_executor.cc
```diff
@@ -192,6 +192,7 @@ ParallelExecutor::ParallelExecutor(const std::vector<platform::Place> &places,
                                    const BuildStrategy &build_strategy,
                                    ir::Graph *graph)
     : member_(new ParallelExecutorPrivate(places)) {
+  VLOG(10) << "enter ParallelExecutor";
```
typhoonzero: should remove this

gongweibao: Thanks!

paddle/fluid/framework/details/all_reduce_deps_pass.cc
gongweibao: Use the op_role_var attr

Issue #15599, opened 2019-01-30
ngraph compile fail


tensor-tang:
## how to reproduce

```
cmake .. -DCMAKE_BUILD_TYPE=Release -DWITH_GPU=OFF -DWITH_MKL=ON -DWITH_TESTING=ON -DWITH_FLUID_ONLY=ON -DWITH_DOC=OFF -DWITH_CONTRIB=OFF   -DWITH_INFERENCE_API_TEST=ON -DON_INFER=ON -DWITH_MKLDNN=ON -DWITH_NGRAPH=ON
make -j
```

## error
```
[ 99%] Linking CXX static library libpaddle_fluid.a
../../operators/ngraph/libngraph_engine.a(ngraph_engine.cc.o): In function `paddle::operators::NgraphEngine::BuildNgFunction()':
ngraph_engine.cc:(.text+0x43d4): undefined reference to `ngraph::Function::Function(ngraph::NodeVector const&, ngraph::ParameterVector const&, std::string const&)'
../../operators/ngraph/libngraph_engine.a(ngraph_engine.cc.o): In function `_GLOBAL__sub_I_ngraph_engine.cc':
ngraph_engine.cc:(.text.startup+0x268): undefined reference to `ngraph::runtime::Backend::create(std::string const&)'
../../operators/ngraph/libngraph_bridge.a(ngraph_bridge.cc.o): In function `ngraph::op::Constant::Constant<float>(ngraph::element::Type const&, ngraph::Shape, std::vector<float, std::allocator<float> > const&)':
ngraph_bridge.cc:(.text._ZN6ngraph2op8ConstantC2IfEERKNS_7element4TypeENS_5ShapeERKSt6vectorIT_SaIS9_EE[_ZN6ngraph2op8ConstantC5IfEERKNS_7element4TypeENS_5ShapeERKSt6vectorIT_SaIS9_EE]+0x99): undefined reference to `ngraph::Node::Node(std::string const&, ngraph::NodeVector const&, unsigned long)'
ngraph_bridge.cc:(.text._ZN6ngraph2op8ConstantC2IfEERKNS_7element4TypeENS_5ShapeERKSt6vectorIT_SaIS9_EE[_ZN6ngraph2op8ConstantC5IfEERKNS_7element4TypeENS_5ShapeERKSt6vectorIT_SaIS9_EE]+0x674): undefined reference to `ngraph::node_validation_assertion_string(ngraph::Node const*)'
[...]

PR #16447, opened 2019-03-25
fix ngraph compile issue 

baojun-nervana:
Fix compile issue with gcc 4.8 https://github.com/PaddlePaddle/Paddle/issues/15599


tensor-tang:
Could you give a little bit explanation why fail before?

baojun-nervana:
the gcc 4.8 was given by an env variables which was received by paddle but not by nGraph. NGraph used the system gcc which is 5.4 and caused a mismatch.

tensor-tang:
Got it, Thanks.

tensor-tang (approved):
fixed the problem

PR #16399, opened 2019-03-22
create test for quantized resnet50

sfraczek:

PR #16448, opened 2019-03-25
Fix order of mkldnn passes

wojtuss:
This patch sets the mkldnn passes in correct order.
With the previous order the relu op was not fused in ResNet50 model.

test=develop


luotao1:
Does this PR solve https://github.com/PaddlePaddle/Paddle/pull/16399#issuecomment-476326800?

bingyanghuang:
> Does this PR solve [#16399 (comment)](https://github.com/PaddlePaddle/Paddle/pull/16399#issuecomment-476326800)?

@sfraczek What do you think about this fix?

wojtuss:
Sorry, it seems it is not working well in all cases.

sfraczek:
Probably one problem is that "mkldnn depthwise conv pass" runs after "conv bn fuse pass" so the depthwise convolutions are not replaced by regular convolution and cannot be fused with batch norm.

PR #16450, opened 2019-03-26
Add SpectralNormGradOpDescMaker

zhhsplendid:
Use SpectralNormGradOpDescMaker instead of DefaultGradOpDescMaker to
avoid registering useless variables to improve GPU usage.

test=develop


zhhsplendid:
Add comment for reviewers:

The "Out" variable is useless when calculate the backward gradient for SpectralNormGradOp. So in the  SpectralNormGradOpDescMaker, we set input for "Out_grad" instead of default both "Out" and "Out_grad"

sneaxiy (dismissed):
LGTM.

sneaxiy (approved):
LGTM.

PR #16454, opened 2019-03-26
polish deepCF model to support real dataset

panyx0718:


python/paddle/fluid/initializer.py
```diff
@@ -756,7 +756,7 @@ def __call__(self, var, block):
             values = [int(v) for v in self._value.flat]
         else:
             raise ValueError("Unsupported dtype %s", self._value.dtype)
-        if self._value.size > 1024 * 1024 * 5:
+        if self._value.size > 1024 * 1024 * 1024:
```
JiabinYang, line 759: why change this?

JiabinYang (commented):
one question

python/paddle/fluid/initializer.py
panyx0718: NumpyArrayInitializer limits numpy size. I think a few GBs is OK

JiabinYang (approved):
LGTM

PR #16461, opened 2019-03-26
[Paddle/GPU] Add SyncBatchNormGradOpDescMaker

zhhsplendid:
Use SyncBatchNormGradOpDescMaker instead of DefaultGradOpDescMaker
to avoid registering useless variables to improve GPU usage.

Bias are useless variable.

test=develop


qingqing01:
> Use SyncBatchNormGradOpDescMaker instead of DefaultGradOpDescMaker

SyncBatchNormGradOpDescMaker use atchNormGradOpDescMaker, not DefaultGradOpDescMaker.

zhhsplendid:
> > Use SyncBatchNormGradOpDescMaker instead of DefaultGradOpDescMaker
> 
> SyncBatchNormGradOpDescMaker use atchNormGradOpDescMaker, not DefaultGradOpDescMaker.

Thank you a lot for taking a look. I noticed it in the past and the BatchNormGradOpDescMaker doesn't fit in this case:

1. We found and tested that "Bias" is redundant input, but BatchNormGradOpDescMaker still set it at this line: https://github.com/PaddlePaddle/Paddle/blob/8d22bc17a42f97b96d560e513fd7deaea33ae0b8/paddle/fluid/operators/batch_norm_op.cc#L570

2. In addition, sync_batch_norm doesn't support use_global_stats True, SyncBatchNormGradOpDescMaker doesn't have to run some lines in BatchNormGradOpDescMaker.

That was why I made this PR instead of using existing BatchNormGradOpDescMaker.

PR #16490, opened 2019-03-27
Refine default MKL-DNN Pass order

Sand3r-:
As mentioned in the https://github.com/PaddlePaddle/Paddle/pull/16399#issuecomment-476702198
> I have refined the default pass order, by adding conv batch norm passes to the list of mkl-dnn passes. That way:
> 
> * The correct order of pass execution is preserved and the fusings behave as intended
> * There is no need for user to manually set passes for his/her application

### DIFF
--- a/paddle/fluid/inference/api/paddle_pass_builder.h
+++ b/paddle/fluid/inference/api/paddle_pass_builder.h
@@ -117,6 +117,8 @@ class CpuPassStrategy : public PassStrategy {
 
       for (auto &pass : std::vector<std::string>(
                {"depthwise_conv_mkldnn_pass",    //
+                "conv_bn_fuse_pass",             //
+                "conv_eltwiseadd_bn_fuse_pass",  //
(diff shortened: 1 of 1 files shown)


paddle/fluid/inference/api/paddle_pass_builder.h
```diff
@@ -117,6 +117,8 @@ class CpuPassStrategy : public PassStrategy {
 
       for (auto &pass : std::vector<std::string>(
                {"depthwise_conv_mkldnn_pass",    //
+                "conv_bn_fuse_pass",             //
```
luotao1: please add some explanation here, why need `conv_bn_fuse_pass` again in mkldnn_passes.

luotao1 (approved):

luotao1:
If this PR passes all the CI in night, please squash and merge it yourself.

wojtuss (approved):

PR #16514, opened 2019-03-28
Fix the bug of the fast threaded executor

chengduoZH:
Fix the bug of the fast threaded executor. In some case, the fetch_op are ready at the beginning, so we should deal with the ready_fetch_ops.


Yancey0623 (approved):
LGTM!

paddle/fluid/framework/details/fast_threaded_ssa_graph_executor.cc
```diff
@@ -56,6 +56,7 @@ FeedFetchList FastThreadedSSAGraphExecutor::Run(
   fetches.resize(fetch_tensors.size());
   std::unordered_map<std::string, std::vector<VarHandleBase *>> fetched_vars;
   std::vector<FetchOpHandle *> fetch_ops;
+  std::vector<OpHandleBase *> ready_fetch_ops;
```
chengduoZH, line 59: In some case, the fetch_op are ready at the beginning.

PR #16526, opened 2019-03-28
refine subgraph trt and anakin

NHZlX:
add anakin ssd support
refine trt first run
add quant dequant fuse pass
omit simplify_anakin_priorbox_detection template
omit transpose_flatten_concat_fuse template


Superjomn (approved):
LGTM

PR #16529, opened 2019-03-28
preprocess with PIL the full val dataset and save binary

lidanqing-intel:
test=develop
preprocess the full val of ILSVRC2012 with python PIL and save to binary file so as to align with [test_calibration](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/contrib/tests/test_calibration.py) INT8v1 
provide full val data for analyzer_int8_resnet50_test in PR #16399

### DIFF
--- a/paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
+++ b/paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
@@ -21,6 +21,7 @@ import functools
 import contextlib
 from PIL import Image, ImageEnhance
 import math
+from paddle.dataset.common import download
 
 random.seed(0)
 np.random.seed(0)
@@ -29,8 +30,6 @@ DATA_DIM = 224
 
 SIZE_FLOAT32 = 4
 SIZE_INT64 = 8
-
-DATA_DIR = './data/ILSVRC2012/data.bin'
 
 img_mean = np.array([0.485, 0.456, 0.406]).reshape((3, 1, 1))
 img_std = np.array([0.229, 0.224, 0.225]).reshape((3, 1, 1))
@@ -71,15 +70,60 @@ def process_image(img_path, mode, color_jitter, rotate):
     return img
 
 
+def download_unzip():
+
+    tmp_folder = 'int8/download'
+
+    cache_folder = os.path.expanduser('~/.cache/' + tmp_folder)
+
+    data_urls = []
+    data_md5s = []
+
+    data_urls.append(
+        'https://paddle-inference-dist.bj.bcebos.com/int8/ILSVRC2012_img_val.tar.gz.partaa'
+    )
+    data_md5s.append('60f6525b0e1d127f345641d75d41f0a8')
+    data_urls.append(
+        'https://paddle-inference-dist.bj.bcebos.com/int8/ILSVRC2012_img_val.tar.gz.partab'
+    )
+    data_md5s.append('1e9f15f64e015e58d6f9ec3210ed18b5')
+
+    file_names = []
+    for i in range(0, len(data_urls)):
+        download(data_urls[i], tmp_folder, data_md5s[i])
+        file_names.append(data_urls[i].split('/')[-1])
+
+    zip_path = os.path.join(cache_folder, 'full_imagenet_val.tar.gz')
+
+    if not os.path.exists(zip_path):
+        cat_command = 'cat'
+        for file_name in file_names:
+            cat_command += ' ' + os.path.join(cache_folder, file_name)
+        cat_command += ' > ' + zip_path
+        os.system(cat_command)
+
+    if not os.path.exists(cache_folder):
+        cmd = 'mkdir {0} && tar xf {1} -C {0}'.format(cache_folder, zip_path)
+
+    cmd = 'rm -rf {3} && ln -s {1} {0}'.format("data", cache_folder, zip_path)
+
+    os.system(cmd)
+
+    data_dir = os.path.expanduser(cache_folder + 'data')
+
+    return data_dir
+
+
 def reader():
-    data_dir = DATA_DIR
+    data_dir = download_unzip()
     file_list = os.path.join(data_dir, 'val_list.txt')
-    bin_file = os.path.join(data_dir, 'data.bin')
+    output_file = os.path.join(data_dir, 'int8_full_val.bin')
     with open(file_list) as flist:
         lines = [line.strip() for line in flist]
         num_images = len(lines)
 
-        with open(bin_file, "w+b") as of:
+        with open(output_file, "w+b") as of:
+            #save num_images(int64_t) to file
             of.seek(0)
             num = np.array(int(num_images)).astype('int64')
             of.write(num.tobytes())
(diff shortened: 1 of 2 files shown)


lidanqing-vv:
@luotao1 Please check if the DATA_DIR is ok?

luotao1:
Could you add the command from `wget and unzip`? Otherwise, users don't know how to get the original dataset. You can refer to `test_calibration.py`.
Then, I use `preprocess.py` to implement `wget, unzip and preprocess`. Maybe the name of `preprocess.py` changes to `full_ ILSVRC2012_val.py`?

paddle/fluid/inference/tests/api/preprocess.py
```diff
@@ -0,0 +1,109 @@
...
+
+random.seed(0)
+np.random.seed(0)
+
+DATA_DIM = 224
+
+SIZE_FLOAT32 = 4
+SIZE_INT64 = 8
+
+DATA_DIR = '/data/ILSVRC2012'
```
luotao1: If I run `cd build; python ../paddle/fluid/inference/tests/api/preprocess.py`, where will the output data?

lidanqing-vv: Currently it outputs to /data/ILSVRC2012/data.bin.  I also wanted to ask this. Where should I put the output data.bin

lidanqing-intel:
> Could you add the command from `wget and unzip`? Otherwise, users don't know how to get the original dataset. You can refer to `test_calibration.py`.
Ok I will do this.
> Then, I use `preprocess.py` to implement `wget, unzip and preprocess`. Maybe the name of `preprocess.py` changes to `full_ ILSVRC2012_val.py`?
Yes I think 'full_ ILSVRC2012_val.py' is good

lidanqing-vv:
> Could you add the command from `wget and unzip`? Otherwise, users don't know how to get the original dataset. You can refer to `test_calibration.py`.
> Then, I use `preprocess.py` to implement `wget, unzip and preprocess`. Maybe the name of `preprocess.py` changes to `full_ ILSVRC2012_val.py`?

Ok I will add wget and unzip part. I may change the name to full_ILSVRC2012_val.py, but it is about preprocess. Maybe 'full_ILSVRC2012_val_preprocess.py'? What do you think?

paddle/fluid/inference/tests/api/preprocess.py
luotao1: `/data/ILSVRC2012/data.bin` is not better, since we don't have `/data` authority. `./data/ILSVRC2012/data.bin`?

luotao1:
> Maybe 'full_ILSVRC2012_val_preprocess.py'? What do you think?

I think it's OK.

paddle/fluid/inference/tests/api/preprocess.py
bingyanghuang: @lidanqing-intel you can put in the .cache like V1 did :https://github.com/PaddlePaddle/Paddle/blob/2632327429ed823cf1a0c2593cfa411fb2f111b9/python/paddle/fluid/contrib/tests/test_calibration.py#L120

lidanqing-vv:
> Could you add the command from `wget and unzip`? Otherwise, users don't know how to get the original dataset. You can refer to `test_calibration.py`.
> Then, I use `preprocess.py` to implement `wget, unzip and preprocess`. Maybe the name of `preprocess.py` changes to `full_ ILSVRC2012_val.py`?

Do I need to give the option to download 100 val images? Or only downloading full val is good.

luotao1:
only downloading full val is enough

paddle/fluid/inference/tests/api/preprocess.py
luotao1: I think `.cache` is OK.

lidanqing-vv:
> only downloading full val is enough

ok

paddle/fluid/inference/tests/api/preprocess.py
lidanqing-vv: ok will use .cache

lidanqing-vv: @luotao1  the data.bin path will be `~/.cache/int8_full_val_bin/data.bin` Is it ok ? I am worried if I put directly in  `~/.cache/` may cause some misunderstanding

luotao1: ~/.cache/int8_full_val_bin/data.bin is OK

luotao1: Since V1 put into `~/.cache/paddle/dataset/int8/download`, how about put into `~/.cache/paddle/dataset/int8/download/int8_full_val.bin`

lidanqing-vv: > Since V1 put into `~/.cache/paddle/dataset/int8/download`, how about put into `~/.cache/paddle/dataset/int8/download/int8_full_val.bin`

Yes! Agree.

lidanqing-vv:
@bingyanghuang The generated file is `~/.cache/paddle/dataset/int8/download/data/ILSVRC2012/int8_full_val.bin`

lidanqing-vv:
> only downloading full val is enough

Done

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def reader():
+    data_dir = download_unzip()
+    file_list = os.path.join(data_dir, 'val_list.txt')
+    output_file = os.path.join(data_dir, 'int8_full_val.bin')
+    with open(file_list) as flist:
+        lines = [line.strip() for line in flist]
+        num_images = len(lines)
+        if not os.path.exists(output_file):
```
bingyanghuang, line 128: we cannot only judge the existence of the output file. Because the process of generating the the "data.bin" is too long , it is possible that the process is not finished but user stop this running or some error happens like "no space left". These kinds of interruption  will leave the uncompleted file in the folder. And next time when you run again this python script, it will not generate the new output binary file.

paddle/fluid/inference/tests/api/preprocess.py
luotao1: I see the unzip dataset is not in this location.

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
```diff
@@ -0,0 +1,162 @@
...
+            print(
+                'Preprocessing to binary file...<num_images><all images><all labels>...\n'
```
luotao1, line 130: This `Print` is hard to understand.

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+            with open(output_file, "w+b") as of:
```
luotao1, line 132: please add some print every 1000 images.

PR #16543, opened 2019-03-29
fix fluid.metrics.Precision bug

Genie-Liu:
1 metric update should be in the for loop. 
2 sample_num is the shape's first element 
3 preds cast should use round, astype("int32") method will cast all the elements(which is less than 1) to 0.


shanyi15:
please add "test=develop" in your commit message to trigger CI, thanks @Genie-Liu

Genie-Liu:
has added
> please add "test=develop" in your commit message to trigger CI, thanks @Genie-Liu

qingqing01 (approved):

shanyi15 (approved):
LG comments

Issue #16552, opened 2019-03-29
test_analyzer_int8_mobilenet timeout


luotao1:
http://ci.paddlepaddle.org/viewLog.html?buildId=77536&tab=buildLog&buildTypeId=Paddle_PrCi&logTab=tree&filter=all&_focus=19228
```
[08:35:28]The following tests FAILED:
[08:35:28]	165 - test_analyzer_int8_mobilenet (Timeout)
```
Each unit-test have only 1200s limits.

PR #16558, opened 2019-03-29
remove profiling from int8 analyzer test

wojtuss:
This should fix CI timeout: https://github.com/PaddlePaddle/Paddle/issues/16552

test=develop


luotao1 (approved):
LGTM

PR #16584, opened 2019-04-01
enhance analyzer_tests download

luotao1:
When I download analyzer_tests' model and dataset on local machine:
image
This PR adds `--no-check-certificate` for `wget` to fix the problem.


Xreki (approved):
LGTM

PR #16617, opened 2019-04-02
Reove dgc dependency from inference.

gongweibao:


CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ if(WITH_GPU)
     include(anakin_subgraph)
 endif()
 
-if(WITH_GPU AND NOT WIN32)
+if(WITH_GPU AND NOT WIN32 NOT ON_INFER)
```
luotao1, line 199: AND NOT ON_INFER

paddle/fluid/operators/CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ endif()
 
 set(COMMON_OP_DEPS ${OP_HEADER_DEPS})
 
-if (WITH_GPU AND NOT WIN32)
+if (WITH_GPU AND NOT WIN32 NOT ON_INFER)
```
luotao1, line 75: AND NOT ON_INFER

PR #16640, opened 2019-04-02
Enable the runtime_context_cache pass in train phase

Xreki:
Add an option in `build_strategy` to enable the `runtime_context_cache` when training. But, there is not explicitly speedup when training PaddingRNN small model.


luotao1 (approved):
LGTM! 可以补充下PR的description（目前是空），能否贴一下该pass在train阶段的效果呢？

PR #16652, opened 2019-04-03
fix dataset bug

xjqbest:
fix timeout error when global shuffle
fix dataset testcase error

### DIFF
--- a/paddle/fluid/framework/data_feed.cc
+++ b/paddle/fluid/framework/data_feed.cc
@@ -238,11 +238,6 @@ void InMemoryDataFeed<T>::SetThreadNum(int thread_num) {
 }
 
 template <typename T>
-void InMemoryDataFeed<T>::SetTrainerId(int trainer_id) {
-  trainer_id_ = trainer_id;
-}
-
-template <typename T>
 void InMemoryDataFeed<T>::SetTrainerNum(int trainer_num) {
   trainer_num_ = trainer_num;
 }
@@ -372,12 +367,10 @@ void InMemoryDataFeed<T>::GlobalShuffle() {
   auto fleet_ptr = FleetWrapper::GetInstance();
   std::vector<std::vector<T*>> send_vec(trainer_num_);
   std::vector<int> send_index(trainer_num_);
-  std::vector<T> local_send_vec;
   uint64_t reserve_len = fleet_send_batch_size_ / trainer_num_;
   for (auto& vec : send_vec) {
     vec.reserve(reserve_len);
   }
-  local_send_vec.reserve(reserve_len);
   for (int i = 0; i < trainer_num_; ++i) {
     send_index[i] = i;
   }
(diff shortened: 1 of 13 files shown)


paddle/fluid/framework/data_feed.h
```diff
@@ -91,9 +91,13 @@ class DataFeed {
   // This function will do nothing at default
   virtual void SetThreadId(int thread_id) {}
   // This function will do nothing at default
+  virtual void SetTrainerId(int trainer_id) {}
```
guru4elephant: why do you have to add trainer_id to dataset? is there anyway we can decouple dataset and trainer?

paddle/fluid/pybind/data_set_py.cc
```diff
@@ -49,12 +49,18 @@ void BindDataset(py::module* m) {
       }))
       .def("set_filelist", &framework::Dataset::SetFileList)
       .def("set_thread_num", &framework::Dataset::SetThreadNum)
+      .def("set_trainer_id", &framework::Dataset::SetTrainerId)
       .def("set_trainer_num", &framework::Dataset::SetTrainerNum)
+      .def("set_fleet_send_batch_size",
```
guru4elephant, line 53: do you have default value for this?

paddle/fluid/framework/data_feed.cc
```diff
@@ -388,8 +419,14 @@ void InMemoryDataFeed<T>::GlobalShuffle() {
       }
     }
   }
-  for (int j = 0; j < send_vec.size(); ++j) {
-    if (send_vec[j].size() != 0) {
+  // shuffle the sequence of sending to avoid network timeout error
```
guru4elephant, line 404: so the previous timeout problem is fixed?

paddle/fluid/framework/data_feed.h
xjqbest: because when global shuffle，if reader knows some instance are send to local，then  no need to use client to client communication

paddle/fluid/pybind/data_set_py.cc
xjqbest: yes, default value is 80000 in c++ InMemoryDataFeed constructor，and in python api (not user api) can also set this value

paddle/fluid/framework/data_feed.cc
xjqbest: yes,  this can solve the timeout problem～

paddle/fluid/framework/data_feed.h
guru4elephant: but I think the timeout problem has been solved. Not sure whether this is necessary

xjqbest: ok , I will try removing the optimization of "instance sent to local"  , and check timeout problem.

xjqbest: after testing,  there is no timeout problem when removing trainer id.

python/paddle/fluid/incubate/fleet/parameter_server/__init__.py
```diff
@@ -123,18 +123,25 @@ def init_pserver(self):
             print("You should run DistributedOptimizer.minimize() first")
             sys.exit(-1)
 
-    def init_worker(self, programs):
+    def init_worker(self, programs, scopes=None):
```
guru4elephant, line 126: It seems that a user has to specify a scope for each program. Maybe you can add the scope into python program class in the future.

guru4elephant (approved):
LGTM

PR #16667, opened 2019-04-03
[MKL-DNN] Added reusing of primitive descriptors (fp32)

jczaja:
Changes discussed here are are introducing Reusing concept for MKL-DNN primitive descriptors (computational ones) for fp32 mkl-dnn ops (conv, conv transpose, softmax and batch norm). 

Motivation:
- To be independent from creation time of primitive descriptors, that may vary from version of MKL-DNN.
- Performance improvement

Performance improvements (SKX8180, ResNet50):
- Significant boost for small workload when lots of computing  resources ( BS=1, num threads=28) : ~10% 
- Tiny improvement for more often use scenarios (BS=1, num_threads=1) : ~0.5%

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -144,7 +144,6 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     const std::string key = platform::ConvMKLDNNHandler::GetHash(
         src_tz, weights_tz, strides, paddings, dilations, groups,
         ctx.op().Input("Input") + ctx.op().Input("Filter"));
-    const std::string key_conv_pd = key + "@conv_pd";
 
     std::vector<primitive> pipeline;
 
@@ -183,6 +182,8 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     auto dst_md = platform::MKLDNNMemDesc(
         dst_tz, platform::MKLDNNGetDataType<T>(), chosen_memory_format);
 
+    platform::ConvMKLDNNHandler handler(dev_ctx, mkldnn_engine, key);
+
     // create a conv primitive descriptor and save it for usage in backward
     std::shared_ptr<mkldnn::convolution_forward::primitive_desc> conv_pd;
     auto fwd_prop_kind = is_test ? mkldnn::prop_kind::forward_inference
@@ -191,18 +192,14 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
       bias_tz = paddle::framework::vectorize2int(bias->dims());
       auto bias_md = platform::MKLDNNMemDesc(
           bias_tz, platform::MKLDNNGetDataType<T>(), memory::format::x);
-      conv_pd = ConvFwdPrimitiveDesc(
+      conv_pd = handler.AcquireConvolutionPrimitiveDescriptor(
           src_md, weights_md, bias_md, dst_md, strides, paddings, mkldnn_engine,
           fuse_relu, fuse_residual_conn, fwd_prop_kind);
     } else {
-      conv_pd = ConvFwdPrimitiveDesc(src_md, weights_md, dst_md, strides,
-                                     paddings, mkldnn_engine, fuse_relu,
-                                     fuse_residual_conn, fwd_prop_kind);
+      conv_pd = handler.AcquireConvolutionPrimitiveDescriptor(
+          src_md, weights_md, boost::none, dst_md, strides, paddings,
+          mkldnn_engine, fuse_relu, fuse_residual_conn, fwd_prop_kind);
     }
-    // Save conv_pd/src_memory/weights_memory for backward pass
-    if (!is_test) dev_ctx.SetBlob(key_conv_pd, conv_pd);
-
-    platform::ConvMKLDNNHandler handler(conv_pd, dev_ctx, mkldnn_engine, key);
 
     // create mkldnn memory from input tensors (data/weights)
     auto user_src_memory_p =
@@ -633,31 +630,6 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
   }
 
  private:
-  mkldnn::primitive_attr CreatePostOps(bool fuse_relu,
-                                       bool fuse_residual_conn) const {
-    mkldnn::primitive_attr conv_attr;
-    mkldnn::post_ops post_operations;
-    // Fusion with Elementwise layer relies on adding a sum post-operation with
-    // the scale parameter. It is assumed that when fuse_residual_connection is
-    // true, the output tensor contains the data coming from residual
-    // connection. The result of this post_op is:
-    // Output = scale * Output + Conv_Out.
-    if (fuse_residual_conn) {
-      post_operations.append_sum(1.0f);
-    }
-    // Fusion with ReLU layer is executed through the PostOps feature. Create a
-    // PostOps object and configure it to execute an eltwise relu operation.
-    if (fuse_relu) {
-      constexpr float scale = 1.0f;
-      constexpr float negative_slope = 0.0f;
-      constexpr float placeholder = 0.0f;
-      post_operations.append_eltwise(scale, mkldnn::algorithm::eltwise_relu,
-                                     negative_slope, placeholder);
-    }
-    conv_attr.set_post_ops(post_operations);
-    return conv_attr;
-  }
-
   mkldnn::primitive_attr CreatePostOps(
       bool fuse_relu, bool fuse_residual_conn,
       const std::vector<float> output_shift_scale, float sum_scale) const {
@@ -685,30 +657,6 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
                        const std::vector<int>& paddings,
                        const mkldnn::engine& engine, const bool fuse_relu,
                        const bool fuse_residual_conn,
-                       mkldnn::prop_kind fwd_prop_kind) const {
-    memory::dims stride_dims = strides;
-    memory::dims padding_dims = paddings;
-
-    auto conv_desc = mkldnn::convolution_forward::desc(
-        fwd_prop_kind, mkldnn::convolution_direct, src, weights, dst,
-        stride_dims, padding_dims, padding_dims, mkldnn::padding_kind::zero);
-
-    mkldnn::primitive_attr conv_attr =
-        CreatePostOps(fuse_relu, fuse_residual_conn);
-
-    auto p_conv_pd = new mkldnn::convolution_forward::primitive_desc(
-        conv_desc, conv_attr, engine);
-
-    return std::unique_ptr<mkldnn::convolution_forward::primitive_desc>(
-        p_conv_pd);
-  }
-
-  std::unique_ptr<mkldnn::convolution_forward::primitive_desc>
-  ConvFwdPrimitiveDesc(const memory::desc& src, const memory::desc& weights,
-                       const memory::desc& dst, const std::vector<int>& strides,
-                       const std::vector<int>& paddings,
-                       const mkldnn::engine& engine, const bool fuse_relu,
-                       const bool fuse_residual_conn,
                        const std::vector<float> output_shift_scale,
                        const float sum_scale, bool is_test) const {
     memory::dims stride_dims = {strides[0], strides[1]};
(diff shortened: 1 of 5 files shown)


jczaja:
@luotao1 , @tensor-tang , @kbinias Could you please review this PR?

kbinias (dismissed):
LGTM

jczaja:
@luotao1 , @tensor-tang Do you need any more information on that PR ?

tensor-tang:
Thanks @jczaja 

What it the background of this PR?

Is this the final solution of replacing tensor modifications discussed in #16248 ?

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -12,6 +12,7 @@
    See the License for the specific language governing permissions and
    limitations under the License. */
 
+#include <boost/optional.hpp>
```
tensor-tang: prefer use `""` since boost is in third-party.

jczaja:
@tensor-tang This is not directly related to tensor modifications. It is just optimization. As for tensor modifications, We are looking to have it updated/reimplemented to match mkl-dnn 1.0 new API.

MKL-DNN Primitive descriptor creation time takes quite some time and it varies from MKL-DNN version to another. This change makes PD to be stored in context and reuse in next iterations.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -395,9 +396,28 @@ class TransposeMKLDNNHandler : public MKLDNNHandler {
   std::vector<int> logical_axis_;
 };
 
+template <typename T>
+struct mode;
```
tensor-tang: what is mode? seems like `mkldnn::algorithm`

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -542,6 +562,73 @@ class ConvMKLDNNTemplateHandler : public MKLDNNHandler {
...
+
+    auto conv_pd = std::static_pointer_cast<typename forward_t::primitive_desc>(
+        dev_ctx_.GetBlob(key_conv_pd));
+
+    if (conv_pd == nullptr) {
+      mkldnn::memory::dims stride_dims = strides;
+      mkldnn::memory::dims padding_dims = paddings;
+
+      auto conv_desc =
+          bias ? typename forward_t::desc(fwd_prop_kind, mode<forward_t>::T,
```
tensor-tang: And it is just for conv now, just use what is needed seems clearer.

tensor-tang:
> This is not directly related to tensor modifications. It is just optimization. As for tensor modifications, We are looking to have it updated/reimplemented to match mkl-dnn 1.0 new API.

Got it~

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -633,31 +630,6 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
   }
 
  private:
-  mkldnn::primitive_attr CreatePostOps(bool fuse_relu,
```
tensor-tang, line 636: Seems just move this to `mkldnn_reuse.h`?

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -542,6 +562,73 @@ class ConvMKLDNNTemplateHandler : public MKLDNNHandler {
...
+      constexpr float placeholder = 0.0f;
+      post_operations.append_eltwise(scale, mkldnn::algorithm::eltwise_relu,
+                                     negative_slope, placeholder);
+    }
+    conv_attr.set_post_ops(post_operations);
+    return conv_attr;
+  }
+
+  std::shared_ptr<typename forward_t::primitive_desc>
+  AcquireConvolutionPrimitiveDescriptor(
```
tensor-tang, line 591: Just find that why only `AcquireConvolutionPrimitiveDescriptor` should be place in `mkldnn_reuse.h` while others do not.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: @tensor-tang Yes. Apart from moving CreatePostOps and ConvFwdPrimitiveDesc and ConvTransposeFwdPrimitiveDesc into mkldnn_reuse.h, they were also unified eg.
Two implementations of CreatePostOps were reimplemented into one and two instances of convFwdPrimitiveDesc and two instances ConvTransposeFwdPrimitiveDesc where also replaced with one function:  AcquireConvolutionPrimitiveDescriptor

tensor-tang: OK, I see.

So what's the difference with line 661 https://github.com/PaddlePaddle/Paddle/pull/16667/files#diff-ff3a710419e42895ac7587353a7af5bcL661.

Can this be reused there as well ?

paddle/fluid/platform/mkldnn_reuse.h
jczaja: @tensor-tang AcquireConvolutionPrimitiveDescriptor does use CreatePostOps. CreatePostOps where used in conv_mkldnn as well as conv_transpose_mkldnn . I merged two implementations of CreatePostOps into one so it is shared by conv and conv_transpose  and also because it is used to AcquireConvolutionPrimitiveDescriptor , I decided to put it in mkldnn_reuse.h . Do you suggest other location?

tensor-tang: That's fine keeping here, if both conv and transpose conv using these functions.

paddle/fluid/platform/mkldnn_reuse.h
jczaja: @tensor-tang Yes, mode was introduced so that When AcquireConvolutionPrimitiveDescriptor
is working for conv_mkldnn eg. template parameter is "convolution_forward"  then mkldnn::algorithm should be "convolution_direct" . When AcquireConvolutionPrimitiveDescriptor
is working for conv_transpose_mkldnn eg. template parameter is "deconvolution_forward" then mkldnn::algorithm should be "desconvolution_direct" . So mode was introduced to have this mapping done

paddle/fluid/platform/mkldnn_reuse.h
jczaja: @tensor-tang AcquireConvolutionPrimitiveDescriptor is used for both conv and conv_transpose.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
jczaja: ok, will update that

paddle/fluid/platform/mkldnn_reuse.h
tensor-tang: All reference just: https://github.com/PaddlePaddle/Paddle/pull/16667/files#diff-a60a81d3c97ec5a7cf7cf3d7e1aa9adbR608 and https://github.com/PaddlePaddle/Paddle/pull/16667/files#diff-a60a81d3c97ec5a7cf7cf3d7e1aa9adbR612 right?

seems no need add `mode` for these.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: @tensor-tang This is Int8 CreatePostOps it has additional params. In this PR I haven't touched int8 conv. Later on yes we could do similar reusing of functions for conv int8.

tensor-tang: Cool.

paddle/fluid/platform/mkldnn_reuse.h
jczaja: @tensor-tang I do not quite understand. AcquireConvolutionPrimitiveDescriptor is used twice: for conv and conv_transpose. So for conv , template paramter is "convolution_forward" and later on "mode" does set algorithm based on that to "convolution_direct" . For conv_transpose template parameter is "deconvolution_forward" and later on "mode" maps this deconvolution_forward into "deconvolution_direct". So in my opinion it is useful at that scenario.

jczaja:
@tensor-tang  I have updated code to Include boost as third part as suggested.

paddle/fluid/platform/mkldnn_reuse.h
tensor-tang: OK.
Then the name should be adapted, `mode` is not a good name for what is mean. 
`algorithm` should be better, or other names you suggest.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -144,7 +144,6 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     const std::string key = platform::ConvMKLDNNHandler::GetHash(
```
xiaolil1, line 144: GetHash can be replaced by AppendKey, which is more efficient by saving multiple string creation.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -191,18 +192,14 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
       bias_tz = paddle::framework::vectorize2int(bias->dims());
       auto bias_md = platform::MKLDNNMemDesc(
           bias_tz, platform::MKLDNNGetDataType<T>(), memory::format::x);
-      conv_pd = ConvFwdPrimitiveDesc(
+      conv_pd = handler.AcquireConvolutionPrimitiveDescriptor(
```
xiaolil1, line 195: It seems these pds (primitive descriptors) still be created in every iterations, right? I have no idea how this PR do "primitive descriptors reusing", could you please explain it? Thanks.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -542,6 +562,73 @@ class ConvMKLDNNTemplateHandler : public MKLDNNHandler {
                                scale_data, mask);
   }
 
+  mkldnn::primitive_attr CreatePostOps(bool fuse_relu,
```
xiaolil1, line 565: @jczaja I have no idea about why these code moving can improve the performance, It seems that the creation of primitive descriptors was not decreased as I commented on "conv_mkldnn_op.cc", could you please explain it?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: @xiaolil1 Yes, but someone else from our team is reworking Hash generation so We plan to migrate when he is done

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: @xiaolil1  AcquireConvolutionPrimitiveDescriptor internally does create PD or get it from Context eg. reuse it

paddle/fluid/platform/mkldnn_reuse.h
jczaja: @xiaolil1 Just moving this code does not improve performacne. PD reusing improves performance. CreatePostOps was present in conv as well as in conv_transpose, and now instead of two instances we have only one in mkldnn_reuse. So that particular change is refactoring to avoid redundant code. See some comments above

paddle/fluid/platform/mkldnn_reuse.h
jczaja: I introduced change : "mode" -> "convolutional_algorithm"  . Hopefully it is more descriptive that way.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
xiaolil1: @jczaja Thanks for your answer. 
I understand pds was reused in  AcquireConvolutionPrimitiveDescriptor, but these "md" like bias_md and "memory" like weights_memory_p, "format", "data_type" ... still were created, which is unnecessary right?

jczaja: @xiaolil1  Memory descriptors eg. md  objects are still being created and yes it would be good to remove then later on.

xiaolil1: @jczaja OK, you can refer to conv INT8 primitive reuse, which have reused all these "descriptors".

jczaja:
@tensor-tang , @xiaolil1 I have rebased changes to have CI passing. Please continue with review

kbinias (approved):
LGTM

xiaolil1 (approved):
LGTM

tensor-tang:
Thanks @jczaja

PR #16677, opened 2019-04-03
[Paddle/GPU] Fix op registry

zhhsplendid:
list of fixed ops:
lookup_table_op
space_to_depth_op
squared_l2_distance_op
squared_l2_norm_op
teacher_student_sigmoid_loss_op
tree_conv_op
warpctc_op

test=develop


paddle/fluid/operators/warpctc_op.cc
```diff
@@ -118,6 +120,27 @@ Recurrent Neural Networks](
...
+  using framework::SingleGradOpDescMaker::SingleGradOpDescMaker;
+
+ protected:
+  std::unique_ptr<framework::OpDesc> Apply() const override {
+    std::unique_ptr<framework::OpDesc> op(new framework::OpDesc());
+
+    op->SetType("warpctc_grad");
+
+    op->SetInput("WarpCTCGrad", Output("WarpCTCGrad"));
+    op->SetInput("Logits", Input("Logits"));
```
sneaxiy, line 134: Seems that buffer of `Logits` is not needed. You can register `NoNeedBufferVarsInference` here. All other codes are good except here. Also add a test to `test_eager_deletion_no_need_buffer_vars_inference.py`.

sneaxiy (dismissed):
LGTM

paddle/fluid/operators/lookup_table_op.cc
```diff
@@ -131,7 +155,7 @@ class LookupTableOpGrad : public framework::OperatorWithKernel {
  protected:
   framework::OpKernelType GetExpectedKernelType(
       const framework::ExecutionContext& ctx) const override {
-    auto data_type = framework::GetDataTypeOfVar(ctx.InputVar("Out"));
+    auto data_type = framework::GetDataTypeOfVar(ctx.InputVar("W"));
```
sneaxiy: Please use `framework::GradVarName("Out")` here, since `W` is not always initialized especially in distributed mode.

zhhsplendid: Done.

zhhsplendid:
Based on comment, I will still change this PR a little:

1. lookup_table_op.cc, try to use framework::GradVarName("Out"), since W is not always initialized especially in distributed mode. Will test under distributed mode

2. When I was developing, using "Logits" as no need inference broke unit tests. Try to debug or not set "Logits" as no need inference.

zhhsplendid:
The CL is ready for review again. Changes after last review:

1. Use framework::GradVarName("Out") in lookup_table_op, since W is not always initialized especially in distributed mode.

2. Did not add "Logits" to the no need inference, because unit test will fail on eager deletion mode at CUDNN place.

sneaxiy (approved):
LGTM.

PR #16683, opened 2019-04-04
fix avx option

tensor-tang:
fix  http://ci.paddlepaddle.org/viewLog.html?buildId=80079&buildTypeId=Paddle_PrCiReleasePython3514&tab=buildLog


junjun315 (approved):
LGTM

PR #16697, opened 2019-04-07
Fix DGC bug.

gongweibao:


python/paddle/fluid/parallel_executor.py
```diff
@@ -123,6 +124,10 @@ def __init__(self,
             exec_strategy=exec_strategy,
             share_vars_from=share_vars_from._compiled_program
             if share_vars_from else None)
+
+        if num_trainers > 1:
+            self._compiled_program._build_strategy.is_distribution = True
```
typhoonzero, line 130: May not be true, nccl2 mode should set this to false.

gongweibao: Use it for DGC now. And I will fix it in next pr.

PR #16698, opened 2019-04-07
Fix training validation convergence - v1.4 bug fix

baojun-nervana:
This is a bug fix for v1.4 release ("overfitting" issue).
In inference, the intermediate mkldnn layout was saved for performance boost.  In training, the feature needs to be disabled as the weights will be updated each iteration. 

This fix is to disable the intermediate layout save, so the validation will be done correctly. Thus this will resolve the "overfitting" issue. The updated convergence curve will be sent for review as I had collected more data points.

Another change in this PR is the "read" op was included to check inputs as py_reader uses read op instead of feed op.

CC. @mozga-intel @jianhang-liu

### DIFF
--- a/paddle/fluid/operators/ngraph/ngraph_engine.cc
+++ b/paddle/fluid/operators/ngraph/ngraph_engine.cc
@@ -75,6 +75,7 @@ std::vector<std::string> NgraphEngine::feed_vars = {};
 std::vector<std::string> NgraphEngine::fetch_vars = {};
 framework::Variable* NgraphEngine::pre_var_ptr = nullptr;
 const framework::BlockDesc* NgraphEngine::p_bdesc = nullptr;
+bool NgraphEngine::is_training = false;
 
 std::unordered_map<std::string, EngineCache> NgraphEngine::engine_cache = {};
 std::unordered_map<std::string,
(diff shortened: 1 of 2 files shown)


tensor-tang:
When would you send the latest convergence curve?

The bugfix needed be cherry-pick to branch release/1.4.

paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -590,7 +594,7 @@ void NgraphEngine::Run(const framework::Scope& scope,
       }
       bool is_persistable =
           (p_persistables->find(vi) != p_persistables->end()) ? true : false;
-      if (is_test && is_persistable) {
+      if (!is_training && is_test && is_persistable) {
```
luotao1, line 597: `is_training` and `is_test` are duplicated?

paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -93,11 +94,13 @@ static std::vector<std::vector<int>> NgraphOpIntervals(
...
   int left = 0;
   while (left < size && ops->at(left)->Type() != framework::kFeedOpType &&
+         ops->at(left)->Type() != "read" &&
          ops->at(left)->Type() != framework::kFetchOpType) {
     ++left;
   }
 
-  while (left < size && ops->at(left)->Type() == framework::kFeedOpType) {
+  while (left < size && (ops->at(left)->Type() == framework::kFeedOpType ||
+                         ops->at(left)->Type() == "read")) {
```
luotao1, line 103: Do you add `ops->at(left)->Type() == "read"` enough?

baojun-nervana, line 103: We can expand this when we see other cases, but we will need to know and understand the case first. So far it can handle the user cases we know.

paddle/fluid/operators/ngraph/ngraph_engine.cc
baojun-nervana: This can be simplified, but it will need changes otherwhere as well. I am trying to minimize change at this late release stage. That may involve more time for review and  validation.  Will follow up to remove duplicates after the release.

tensor-tang (approved):
LGTM

Can address the comments in develop branch

PR #16701, opened 2019-04-08
Make test_parallel_executor_seresnet.py Faster

chengduoZH:
Reduce the time consuming from 432.11 sec to  287.07 sec.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_parallel_executor_seresnext.py
+++ b/python/paddle/fluid/tests/unittests/test_parallel_executor_seresnext.py
@@ -215,7 +215,7 @@ class TestResnet(TestParallelExecutorBase):
             return
 
         global remove_bn
-        remove_bn = True if use_cuda else False
+        remove_bn = True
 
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/tests/unittests/test_parallel_executor_seresnext.py
```diff
@@ -181,43 +180,82 @@ def optimizer(learning_rate=0.01):
...
+        use_reduce=False,
+        optimizer=optimizer)
+
+    return first_loss, last_loss
+
+
+origin_cpu_first_loss, origin_cpu_last_loss = _get_result_of_origin_model(False)
+if core.is_compiled_with_cuda():
+    origin_gpu_first_loss, origin_gpu_last_loss = _get_result_of_origin_model(
+        True)
```
chengduoZH, line 228: Pre-compute the results of the original model and the following test cast only run the model that with the new strategy.

sneaxiy (approved):
LGTM.

PR #16711, opened 2019-04-08
reduce hasAttr elapsed time in RunImpl

luotao1:
`HasAttr` costs a lot of time in some small model. For example, it costs 14% in pyramid_dnn inference.
image
The reason is that `std::unordered_map` is slow in `find`: https://stackoverflow.com/questions/35238001/why-does-the-unordered-map-perform-about-100-times-slower-than-a-2d-vector-for-l

This pr use two `bool` variable to record the result of `HasAttr`. Results:

| | 2620 v3 (ms per sample) |
|---|---|
|before| 0.186852 |
|after| 0.168889 | 
|speedup| 9.6%  |

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -884,6 +884,8 @@ void OperatorWithKernel::RunImpl(const Scope& scope,
   // result of HasAttr.
   if (!enable_cache_runtime_context && HasAttr(kEnableCacheRuntimeContext))
     enable_cache_runtime_context = true;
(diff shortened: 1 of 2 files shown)


luotao1:
After merge #16339, the result for pyramid_dnn is: 

|---|---|---|
|before(develop)|  0.220637 | 0.179725 |
|after(develop+this PR)|0.196298   |0.156329  |
|speedup| 11%  | 13% |

Xreki (approved):
LGTM

PR #16713, opened 2019-04-08
Fix unittests which takes too long time

sneaxiy:
test=develop

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_gradient_clip.py
+++ b/python/paddle/fluid/tests/unittests/test_gradient_clip.py
@@ -20,24 +20,7 @@ import paddle
 import paddle.fluid.core as core
 import paddle.fluid as fluid
 import six
-
-
-def fake_reader_creator(word_dict_size,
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/tests/unittests/test_gradient_clip.py
```diff
@@ -19,6 +19,25 @@
 import paddle
 import paddle.fluid.core as core
 import paddle.fluid as fluid
+import six
+
+
+def fake_reader_creator(word_dict_size,
```
chengduoZH: Now that `test_gradient_clip.py` and `test_eager_deletion_dynamic_rnn_base.py` use `fake_reader_creator` as the reader, `fake_reader_creator ` should be place to a separate file.

sneaxiy: Done.

chengduoZH (approved):
Excellent

PR #16724, opened 2019-04-09
Add an option to enable the cache of expected kernel in train phase.

Xreki:


paddle/fluid/framework/details/build_strategy.h
```diff
@@ -108,6 +108,7 @@ struct BuildStrategy {
   bool remove_unnecessary_lock_{true};
 
   bool cache_runtime_context_{false};
+  bool cache_expected_kernel_{false};
```
luotao1: Could `cache_expected_kernel_ ` default be true?

PR #16729, opened 2019-04-09
modify the script to adapt to new ci

wopeizl:
test=develop


wopeizl:
remove this since it relate to invalid account

PR #16732, opened 2019-04-09
modify the build script for new ci test=develop

wopeizl:
1. give a parallel number for make process on linux platform during runtime since the new CI would support distcc and the make number can be modified via CI configuration.
2. separate the compile and unit-test process.


luotao1 (approved):
LGTM

PR #16744, opened 2019-04-09
Fix auto growth bug of optimizer in dygraph mode

velconia:
cherry pick from https://github.com/PaddlePaddle/Paddle/pull/16744


junjun315 (approved):
LGTM

PR #16746, opened 2019-04-09
move split filelist from trainer.py to fleet & fix error

xjqbest:
(1) move split filelist from trainer.py to fleet 
(2) fix error: rm mpi finalize because MPI_Finalize() is registered  for being automatically called when Python processes exit, so there is no finalize api. refer to [this](https://mpi4py.readthedocs.io/en/stable/overview.html)
(3) fix release_memory not found & fix doc string example error


python/paddle/fluid/incubate/fleet/parameter_server/__init__.py
```diff
@@ -241,6 +241,34 @@ def save_pserver_model(self, save_path):
...
+    def split_filelist(self, filelist):
+        """
+        split filelist before distributed training,
+        for example, filelist is [a, b, c ,d, e]  and trainer_num = 2,
+        then trainer 0 gets [a, b, c] and trainer 1 gets [d, e]
+
+        Args:
+            filelist(list): list of filename, can be local or hdfs/afs.
+
+        Returns:
```
guru4elephant, line 259: You should add a python Example here.

python/paddle/fluid/incubate/fleet/base/role_maker.py
```diff
@@ -128,7 +128,7 @@ def _finalize(self):
         """
         finalize the current MPI instance.
         """
-        self.comm_.finalize()
+        pass
```
guru4elephant, line 131: why do you remove this?

python/paddle/fluid/incubate/fleet/parameter_server/__init__.py
xjqbest: added.

guru4elephant (approved):

PR #16747, opened 2019-04-09
Reduce test_calibration test time by using large batchsize

chuanqi129:
To reduce test_calibration test time. Please @luotao1 have a review on it.

### DIFF
--- a/python/paddle/fluid/contrib/tests/test_calibration.py
+++ b/python/paddle/fluid/contrib/tests/test_calibration.py
@@ -147,10 +147,11 @@ class TestCalibrationForResnet50(unittest.TestCase):
                                                    self.data_cache_folder)
         os.system(cmd)
 
-        self.batch_size = 1
-        self.sample_iterations = 50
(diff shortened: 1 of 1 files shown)


luotao1:
please pass the travis-ci.

luotao1:
```
[01:52:41]582/624 Test #616: test_calibration ....................................   Passed  134.42 sec
```
It seems `using large batchsize` don't have many benefits.

chuanqi129:
> ```
> [01:52:41]582/624 Test #616: test_calibration ....................................   Passed  134.42 sec
> ```
> It seems `using large batchsize` don't have many benefits.

Is it necessary to download data and models for each test? If so, the download phase will take up most of the time. I have verified it locally, this test takes about 50 seconds when the download phase is not included.

luotao1:
> Is it necessary to download data and models for each test? If so, the download phase will take up most of the time. 

Download data and models are already cached.

luotao1 (approved):
LGTM. But we still need speedup this unit-test.

PR #16760, opened 2019-04-10
disable memory_optimize and inpalce strategy by default, test=develop

liupluswei:


sneaxiy (approved):
LGTM. Please add warnings when enabling memory_optimize and in-place strategy.

PR #16762, opened 2019-04-10
Add async sparse param update recorder

jacquesqiao:


seiriosPlus (approved):
LGTM, But I think We need another compatible way to communicate with  gRPC.

PR #16763, opened 2019-04-10
Fix cosine_decay, l2_normalize, pixel_shuffle API comments

shippingwang:
fix some comments
- [x] pixel_shuffle, fix output shape comments
- [x] l2_normalize, fix 2nd of [PaddlePaddle/FluidDoc#672](https://github.com/PaddlePaddle/FluidDoc/issues/672)
- [x] cosine_decay, fix format


shippingwang:
预览
image

image

python/paddle/fluid/layers/learning_rate_scheduler.py
```diff
@@ -360,13 +362,13 @@ def cosine_decay(learning_rate, step_each_epoch, epochs):
         Variable: The decayed learning rate.
 
      Examples:
```
shanyi15: can't find this part in preview pic

shippingwang: > can't find this part in preview pic

Do we have a better self-check process? for now, the developer will repeat "compile -> preview -> rewrite ->compile", most caused by careless though,  I suggest supporting a tool to preview badly, like just copy these comments to the pasteboard, then preview the comments online without compiling the whole program

shippingwang:
image

XiaoguangHu01 (approved):
LGTM

python/paddle/fluid/layers/learning_rate_scheduler.py
shanyi15: > > can't find this part in preview pic
> 
> Do we have a better self-check process? for now, the developer will repeat "compile -> preview -> rewrite ->compile", most caused by careless though, I suggest supporting a tool to preview badly, like just copy these comments to the pasteboard, then preview the comments online without compiling the whole program

we will consider it, thanks for your suggestion

shanyi15 (approved):
LG comments

PR #16769, opened 2019-04-10
test=develop, fix hsigmoid dereference nullptr

JiabinYang:


python/paddle/fluid/layers/nn.py
```diff
@@ -5617,6 +5617,10 @@ def hsigmoid(input,
         raise ValueError(
             "num_classes must not be less than 2 with default tree")
 
+    if (not is_custom) and (is_sparse):
```
liupluswei, line 5620: per my understanding, the core issue here is we need to make sure path_table is not None when is_sparse==True
the case: "is_custom = false, is_spase=true, path_table is not none" is acceptable, right?

JiabinYang: I don't think so, in that case is_sparse should be False

JiabinYang: I think add `if (not is_costum) and ((path_table is not None) or (path_code is not None))` case is better choice, I'll fix that

liupluswei (dismissed):

liupluswei (approved):
LGTM

PR #16792, opened 2019-04-11
re-enable activation inplace in python side, test=release/1.4

liupluswei:


wanghaoshuang (dismissed):

liupluswei:
the model compress memory usage issue can be solved by using GC, so this is not needed

PR #16794, opened 2019-04-11
fix overflow by int32 mul test=develop

Aurelius84:
修复int32相乘overflow的问题

### DIFF
--- a/paddle/fluid/operators/sequence_ops/sequence_conv_op.h
+++ b/paddle/fluid/operators/sequence_ops/sequence_conv_op.h
@@ -41,8 +41,8 @@ class SequenceConvKernel : public framework::OpKernel<T> {
 
     PADDLE_ENFORCE_EQ(in->lod().size(), 1UL,
                       "Only support one level sequence now.");
-    Tensor tensor;
-    const Tensor* padding_data = &tensor;
+
+    const Tensor* padding_data = nullptr;
     if (padding_trainable) {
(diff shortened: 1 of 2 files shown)


luotao1:
Each commit should add `test=develop` to trigger the CI.

paddle/fluid/operators/sequence_ops/sequence_conv_op.h
```diff
@@ -41,15 +41,15 @@ class SequenceConvKernel : public framework::OpKernel<T> {
 
     PADDLE_ENFORCE_EQ(in->lod().size(), 1UL,
                       "Only support one level sequence now.");
-
-    const Tensor* padding_data = nullptr;
```
liupluswei: I think modify the [ContextProjectFunctor](https://github.com/PaddlePaddle/Paddle/blob/6864370a9e7ba991b0dba2553150901bcf63234f/paddle/fluid/operators/math/context_project.h#L90) API will be cleaner, update the  **const Tensor& padding_data** parameter to  **const Tensor* padding_data**, and inside the API, if **padding_trainable** is true, PADDLE_ENFORCE_NOT_NULL(padding_data)

Aurelius84: Thanks，API has been modified from reference to point.

liupluswei (approved):
LGTM

PR #16822, opened 2019-04-12
Optimize merge add

jacquesqiao:
if we find that all the input selected_rows of merge_add has no duplicated id and the row does not need to sort, we can just concat all the rows and tensors to output.


seiriosPlus (approved):
LGTM!

PR #16824, opened 2019-04-12
disable test_elementwise_mul_mkldnn_op case

### LINKED ISSUE #16764: test_elementwise_mul_mkldnn_op fails
In a new CI machine `Intel(R) Xeon(R) CPU E5-2660 v4 @ 2.00GHz`, `test_elementwise_mul_mkldnn_op` fails
```
[08:21:40]550/624 Test #583: test_elementwise_mul_mkldnn_op ......................***Failed    5.15 sec
[08:21:40]W0410 08:21:38.063987 53917 device_context.cc:261] Please NOTE: device: 0, CUDA Capability: 61, Driver API Version: 9.1, Runtime API Version: 8.0
[08:21:40]W0410 08:21:38.066761 53917 device_context.cc:269] device: 0, cuDNN Version: 7.0.
[08:21:40]test_elementwise_mul_mkldnn_op failed
[08:21:40] .......Fssss....................................................
[08:21:40]======================================================================
[08:21:40]FAIL: test_check_output (test_elementwise_mul_mkldnn_op.TestElementwiseMulMKLDNNOp_BroadcastNCHW16c)
[08:21:40]----------------------------------------------------------------------
[08:21:40]Traceback (most recent call last):
[08:21:40]  File "/paddle/build/python/paddle/fluid/tests/unittests/test_elementwise_mul_op.py", line 44, in test_check_output
[08:21:40]    self.check_output()
[08:21:40]  File "/paddle/build/python/paddle/fluid/tests/unittests/op_test.py", line 503, in check_output
[08:21:40]    check_dygraph)
[08:21:40]  File "/paddle/build/python/paddle/fluid/tests/unittests/op_test.py", line 451, in check_output_with_place
[08:21:40]    str(actual_t) + " in class " + self.__class__.__name__)
[08:21:40]AssertionError: Output (Out) has diff at CPUPlace
[08:21:40]Expect [[[[0.00125048 0.08492047]
[08:21:40]   [0.29046938 0.386396  ]]
[08:21:40]
[08:21:40]  [[0.36969334 0.11319863]
[08:21:40]   [0.21415007 0.1986888 ]]
[08:21:40]
[08:21:40]  [[0.00177243 0.25010154]
[08:21:40]   [0.40215877 0.04279045]]
[08:21:40]
[08:21:40]  [[0.41721553 0.3375377 ]
[08:21:40]   [0.46301916 0.00820544]]
[08:21:40]
[08:21:40]  [[0.10934631 0.40422767]
[08:21:40]   [0.2904416  0.22166422]]
[08:21:40]
[08:21:40]  [[0.49167052 0.0152072 ]
[08:21:40]   [0.12244622 0.41933575]]
[08:21:40]
[08:21:40]  [[0.10907405 0.1227215 ]
[08:21:40]   [0.00730317 0.6629001 ]]
[08:21:40]
[08:21:40]  [[0.5631185  0.8816506 ]
[08:21:40]   [0.4704359  0.36099708]]
[08:21:40]
[08:21:40]  [[0.2228665  0.8427488 ]
[08:21:40]   [0.07571415 0.1031361 ]]
[08:21:40]
[08:21:40]  [[0.01434026 0.22389263]
[08:21:40]   [0.06033051 0.18588263]]
[08:21:40]
[08:21:40]  [[0.15922903 0.2645933 ]
[08:21:40]   [0.05752328 0.9053555 ]]
[08:21:40]
[08:21:40]  [[0.25251266 0.14107876]
[08:21:40]   [0.0689055  0.0038731 ]]
[08:21:40]
[08:21:40]  [[0.08460373 0.29914436]
[08:21:40]   [0.16119799 0.5224081 ]]
[08:21:40]
[08:21:40]  [[0.15163004 0.05331361]
[08:21:40]   [0.13301256 0.23944747]]
[08:21:40]
[08:21:40]  [[0.04832048 0.0109666 ]
[08:21:40]   [0.3521237  0.81271416]]
[08:21:40]
[08:21:40]  [[0.51281095 0.8913848 ]
[08:21:40]   [0.33166745 0.04245876]]]]
[08:21:40]But Got[[[[0.00125048 0.10934631]
[08:21:40]   [0.2228665  0.08460373]]
[08:21:40]
[08:21:40]  [[0.08492047 0.40422767]
[08:21:40]   [0.8427488  0.29914436]]
[08:21:40]
[08:21:40]  [[0.29046938 0.2904416 ]
[08:21:40]   [0.07571415 0.16119799]]
[08:21:40]
[08:21:40]  [[0.386396   0.22166422]
[08:21:40]   [0.1031361  0.5224081 ]]
[08:21:40]
[08:21:40]  [[0.36969334 0.49167052]
[08:21:40]   [0.01434026 0.15163004]]
[08:21:40]
[08:21:40]  [[0.11319863 0.0152072 ]
[08:21:40]   [0.22389263 0.05331361]]
[08:21:40]
[08:21:40]  [[0.21415007 0.12244622]
[08:21:40]   [0.06033051 0.13301256]]
[08:21:40]
[08:21:40]  [[0.1986888  0.41933575]
[08:21:40]   [0.18588263 0.23944747]]
[08:21:40]
[08:21:40]  [[0.00177243 0.10907405]
[08:21:40]   [0.15922903 0.04832048]]
[08:21:40]
[08:21:40]  [[0.25010154 0.1227215 ]
[08:21:40]   [0.2645933  0.0109666 ]]
[08:21:40]
[08:21:40]  [[0.40215877 0.00730317]
[08:21:40]   [0.05752328 0.3521237 ]]
[08:21:40]
[08:21:40]  [[0.04279045 0.6629001 ]
[08:21:40]   [0.9053555  0.81271416]]
[08:21:40]
[08:21:40]  [[0.41721553 0.5631185 ]
[08:21:40]   [0.25251266 0.51281095]]
[08:21:40]
[08:21:40]  [[0.3375377  0.8816506 ]
[08:21:40]   [0.14107876 0.8913848 ]]
[08:21:40]
[08:21:40]  [[0.46301916 0.4704359 ]
[08:21:40]   [0.0689055  0.33166745]]
[08:21:40]
[08:21:40]  [[0.00820544 0.36099708]
[08:21:40]   [0.0038731  0.04245876]]]] in class TestElementwiseMulMKLDNNOp_BroadcastNCHW16c
[08:21:40]
[08:21:40]----------------------------------------------------------------------
[08:21:40]Ran 64 tests in 3.246s
[08:21:40]
[08:21:40]FAILED (failures=1, skipped=4)
```
http://ci.paddlepaddle.org/viewLog.html?buildId=82557&tab=buildLog&buildTypeId=Paddle_PrCi&logTab=tree&filter=all&_focus=37940

LeoZhao-BJ:
test = develop
resolve #16764

### DIFF
--- a/paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
+++ b/paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
@@ -130,6 +130,13 @@ class ElementwiseMulMKLDNNKernel : public framework::OpKernel<T> {
 
       z->set_layout(DataLayout::kMKLDNN);
       z->set_format(x->format());
+
+      // convert to nchw format to align with native version
+      using platform::MKLDNNDeviceContext;
+      auto& dev_ctx = ctx.template device_context<MKLDNNDeviceContext>();
+      const auto& mkldnn_engine = dev_ctx.GetEngine();
+      ReorderInput<T>(const_cast<Tensor*>(z), ctx.GetPlace(), mkldnn_engine,
+                      z->dims().size() == 4);
     } else {
       // Fallback to naive version:
       const bool are_inputs_in_same_format = x->format() == y->format();
--- a/paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
+++ b/paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
@@ -130,13 +130,6 @@ class ElementwiseMulMKLDNNKernel : public framework::OpKernel<T> {
 
       z->set_layout(DataLayout::kMKLDNN);
       z->set_format(x->format());
-
-      // convert to nchw format to align with native version
-      using platform::MKLDNNDeviceContext;
-      auto& dev_ctx = ctx.template device_context<MKLDNNDeviceContext>();
-      const auto& mkldnn_engine = dev_ctx.GetEngine();
-      ReorderInput<T>(const_cast<Tensor*>(z), ctx.GetPlace(), mkldnn_engine,
-                      z->dims().size() == 4);
     } else {
       // Fallback to naive version:
       const bool are_inputs_in_same_format = x->format() == y->format();
--- a/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
+++ b/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
@@ -31,7 +31,6 @@ class TestElementwiseMulMKLDNNOp_BroadcastNCHW16c(ElementwiseMulOp):
         self.y = np.random.rand(1, 16).astype(self.dtype)
 
         self.out = x * self.y.reshape(1, 16, 1, 1)
-
         self.out = self.out.transpose(0, 2, 3, 1).reshape(1, 16, 2, 2)
 
     def setUp(self):
--- a/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
+++ b/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
@@ -21,6 +21,9 @@ from paddle.fluid.op import Operator
 from paddle.fluid.tests.unittests.test_elementwise_mul_op import *
 
 
+# TODO(LeoZhao-Intel): re-enable this case
+# https://github.com/PaddlePaddle/Paddle/issues/16764
+@unittest.skip("Not supported well on avx2.")
 class TestElementwiseMulMKLDNNOp_BroadcastNCHW16c(ElementwiseMulOp):
     def init_input_output(self):
         x = np.random.rand(1, 16, 2, 2).astype(self.dtype)
@@ -29,7 +32,7 @@ class TestElementwiseMulMKLDNNOp_BroadcastNCHW16c(ElementwiseMulOp):
 
         self.out = x * self.y.reshape(1, 16, 1, 1)
 
-#        self.out = self.out.transpose(0, 2, 3, 1).reshape(1, 16, 2, 2)
+        self.out = self.out.transpose(0, 2, 3, 1).reshape(1, 16, 2, 2)
 
     def setUp(self):
         super(TestElementwiseMulMKLDNNOp_BroadcastNCHW16c, self).setUp()
--- a/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
+++ b/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
@@ -28,7 +28,8 @@ class TestElementwiseMulMKLDNNOp_BroadcastNCHW16c(ElementwiseMulOp):
         self.y = np.random.rand(1, 16).astype(self.dtype)
 
         self.out = x * self.y.reshape(1, 16, 1, 1)
-        self.out = self.out.transpose(0, 2, 3, 1).reshape(1, 16, 2, 2)
+
+#        self.out = self.out.transpose(0, 2, 3, 1).reshape(1, 16, 2, 2)
 
     def setUp(self):
         super(TestElementwiseMulMKLDNNOp_BroadcastNCHW16c, self).setUp()


LeoZhao-BJ:
@luotao1 @kbinias

paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
```diff
@@ -130,6 +130,13 @@ class ElementwiseMulMKLDNNKernel : public framework::OpKernel<T> {
 
       z->set_layout(DataLayout::kMKLDNN);
       z->set_format(x->format());
+
+      // convert to nchw format to align with native version
+      using platform::MKLDNNDeviceContext;
+      auto& dev_ctx = ctx.template device_context<MKLDNNDeviceContext>();
+      const auto& mkldnn_engine = dev_ctx.GetEngine();
+      ReorderInput<T>(const_cast<Tensor*>(z), ctx.GetPlace(), mkldnn_engine,
+                      z->dims().size() == 4);
```
hshen14: I understand element-wise should not be aware of format, which means input format is equal to output format in op semantics. If adding the above code to fix UT failure but break the op semantics, we may need to rethink carefully.

LeoZhao-BJ: This is to align with native version, avx512 branch code is only case for.
```
    if (is_x_format_correct && is_y_format_correct && are_dims_divisable &&
        is_avx512_enabled)
```
which output format is different with other naive version, so this patch is used to alignment for both output format and test case.

LeoZhao-BJ: currently, for naive version, whatever x, y format is, the output is nchw or nc, while avx512 new added branch, the output is nchw16c, which made format difference on SKUs with different avx support

hshen14: My point is to follow the op semantics, namely element-wise op should not change the format (output format = input format), no matter how kernel implements and how UT works. Otherwise, it may introduce unnecessary reorders at topology level. E.g., conv1 (nchw16c) -> element-wise mul1; conv2 (nchw16c) -> element-wise mul1; element-wise mul1 -> conv3; for this pattern, it is expected to execute with nchw16c in the entire flow (no reorder from nchw to nchw16c). Add @tensor-tang for comments.

hshen14: > currently, for native version, whatever x, y format is, the output is nchw or nc, while avx512 new added branch, the output is nchw16c, which made format difference on SKUs with different avx support

From your description, it looks more like a test case issue, not kernel issue. Better to fix in the test code, instead of kernel.

LeoZhao-BJ: I agree, we shall avoid unnecessary reordering, e.g. output format follows x format.
I think that's the way for further code and test case optimization.

LeoZhao-BJ: > > currently, for native version, whatever x, y format is, the output is nchw or nc, while avx512 new added branch, the output is nchw16c, which made format difference on SKUs with different avx support
> 
> From your description, it looks more like a test case issue, not kernel issue. Better to fix in the test code, instead of kernel.

My opinion is test case should be transparent for low level SKU difference, and output format should be unified, that more makes sense.

kbinias: We are planning to prepare assembly version of this operator which supports AVX2 and AVX-512 VNNI. Maybe until then we could consider detecting CPU features in this UT with Python API (e.g. cpufeature module)

LeoZhao-BJ: Want to understand the rule of input/output format mapping, e.g. when x_data_format is nchw/nchw16c/xx, y is nchw/nchw16c/xx, what's the expected output format ? or it may be dynamic due to CPU SKU ? 

The purpose of this patch is just a quick solution to align avx512 version with naive version to unblock CI, may not for a long-term solution of giving consideration to both performance and interface definition.

LeoZhao-BJ: https://pypi.org/project/py-cpuinfo/
python for cpu info, if needed for detecting cpu features

Sand3r-: Yes, output format shall be equivalent to the x_data_format.

It is because the y_data is broadcasted to x_data during multiplications, and x_data defines the main path of the data flow. Usually elementwise_mul is used for scaling of the output parameters of some certain layer in a given topology. Since it is quite often the case that topologies such as ResNeXt-50 consist of repeating blocks of layers, it is most often the case that the input format to such block will be equivalent to the output format it yields (excluding the last block which is linked to the topology's output, e.g. the `fetch` op). Hence, keeping the same data format across almost the whole data path allows avoiding the reorders.

LeoZhao-BJ: Then, for naive version (working on non-avx512 machine), need follow same rule? or output the most performance-optimized format for specific CPU?

I was trying to convert format in naive version code, but found it is hard to do since it will break some existing test cases in mkldnn mul op.  Maybe need further adjustment on test cases?

Sand3r-: If the input format is non standard (other than nchw, ncw, nc or x) **and** is different from the case when x_data_format = nchw16c & y_data_format = nc, then it needs to be converted to the standard format so that the naive version can handle it.

And yes, it would generally suffice to check for avx512 support before applying the transpose and reshape operators to the reference output in the test code.

(pseudocode)
```
if avx512_supported:
    self.out = self.out.transpose(0, 2, 3, 1).reshape(1, 16, 2, 2)
```

LeoZhao-BJ: I actually totally agree with you on the initiative. I think currently the only problem we need to solve is just to avoid the mismatch between different CPU SKU.

My confusion was, given the x input format is nchw16c on non-avx512 machine, what's the expect output format? your answer by my understanding is nchw, right? 

While I am thinking if we can unify and simply the rule between SKUs, e.g. output follows x input, it means naive version also outputs nchw16c by doing convert internally, since on non-avx512 platform, nchw16c should not happen, and if it happens, it is reasonable for performance cost.

Sand3r-: nchw16c format shall be chosen by mkl-dnn engine only on avx512 compliant machines (so that they can use avx512 instructions to process binary operations on pairs of 16 floats at once), so at least in theory it should never occur that we encounter nchw16c on non-avx512 machine.

If magic happens, then you are right, the nchw16c would need to be converted to nchw so that the naive kernel can correctly process it. However, instead I would propose adding a PADDLE_ENFORCE that would verify that there is no nchw16c format on non-avx512 compliant machine, because if this occurs, something is clearly wrong.

LeoZhao-BJ: Good, this approach makes more sense. Then we may also need to update test case since current case design doesn't consider CPU flag difference, and better to disable this case on non-avx512 machine.

Sand3r-: Good idea, I totally agree. For now, in case of non-avx512 machine only the fallback test case should be executed (at least for until we introduce an avx2 kernel).

luotao1:
> and better to disable this case on non-avx512 machine.

Discussed with @tensor-tang, it's better not to disable this case on non-avx512 machine. For the sake of users, running successfully is more important than performance. Thus, we prefer to modify the codes.

Sand3r-:
> > and better to disable this case on non-avx512 machine.
> 
> Discussed with @tensor-tang, it's better not to disable this case on non-avx512 machine. For the sake of users, running successfully is more important than performance. Thus, we prefer to modify the codes.

Why wouldn't they run it succesfully? If users are using non-avx512 machine, the elementwise_mul would always just execute the naive version when running any topology, be it with or without MKL-DNN.

luotao1:
please add `test=develop` in each commit to trigger the CI.

LeoZhao-BJ:
@luotao1 can you help to check PR_CI failure? from details, seems it failed on test_analyzer_mobilenet_depthwise_conv which is not related with my patch.

image

luotao1:
It is another old issue #15032, I have rerun it again at first.

luotao1 (approved):
LGTM

PR #16838, opened 2019-04-13
fix the hang bugs of memory copying. test=develop

wzzju:
Recently, when doing more quantization labs on different models,  we find the GPU memory copy operation used in `fake_quantize_range_abs_max` and `fake_quantize_moving_average_abs_max` ops will cause hang bugs. This pr tries to fix these bugs.


sneaxiy (approved):
LGTM. Please remove CUDA kernel by overwriting `GetExpectedKernelType` in the future.

wzzju:
> LGTM. Please remove CUDA kernel by overwriting `GetExpectedKernelType` in the future.

I find that `FindMovingAverageAbsMaxFunctor` is only part of `FakeQuantizeMovingAverageAbsMaxKernel`.

qingqing01 (approved):

PR #16849, opened 2019-04-14
fix var fetching issue, test=develop

liupluswei:


liupluswei:
CE [link](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=37977&buildTypeId=PaddleModesl_ModelDebug_Model11Build&tab=buildResultsDiv)

PR #16857, opened 2019-04-15
Fix sum infershape issue

zhaoyuchen2018:
if dim is -1, compile time check fails.

test=develop


phlrain (approved):
LGTM

PR #16860, opened 2019-04-15
Fix bug: long vgg-utest testing time 

junjun315:
Use only part of the test set data validation program, Solve the problem of long unit testing time.Reduce test time from 300 seconds to less than 30 seconds（with dataset cached）


luotao1 (approved):
LGTM

PR #16875, opened 2019-04-15
Improve preprocessing script and read from tar

lidanqing-intel:
Improve the script. Read from tar file directly without extracting to folders.

Resnet50 and mobilenet accuracy on generated full dataset:

**Resnet50:**
Avg top1 INT8 accuracy: 0.7652
Avg top1 FP32 accuracy: 0.7663

**Mobilenet:**
Avg top1 INT8 accuracy: 0.7035
Avg top1 FP32 accuracy: 0.7078

The checksum changed because the order of the data now is the same with the contents order in tar file, not the order of image filenames.


lidanqing-vv:
@kbinias

PR #16880, opened 2019-04-15
[MKL-DNN] Made Tensor::Resize() invalidate mkl-dnn related content

jczaja:
Changes discussed here are introducing invalidation of MKL-DNN format and layout stored in Tensor, when Tensor::Resize is called. Reason is that when Resize is called  Tensor dims are to change so
mkldnn format is no longer relevant. For example we have Tensor of 3D shape (mkldnn format: ncw) and then Resize is called to change Tensor dim's shape to 2D . Leaving format set to mkldnn::format::ncw may result in a functional problems later on.

Problem was detected when Transformer using MKL-DNN was executed, but using only
one mkldnn op: "elementwise_add". Result was a crash. With this fix crash is not appearing.

### DIFF
--- a/paddle/fluid/framework/tensor.cc
+++ b/paddle/fluid/framework/tensor.cc
@@ -96,7 +96,15 @@ Tensor Tensor::Slice(int64_t begin_idx, int64_t end_idx) const {
 }
 
 Tensor& Tensor::Resize(const DDim& dims) {
-  dims_ = dims;
+  if (dims_ != dims) {
+// When Tensor is to have different dimensions eg. 4D instead of 3D
+// then for MKL-DNN layout we invalidate format as
+// having eg. mkldnn::ncw for 4D data does not make sense
+#ifdef PADDLE_WITH_MKLDNN
+    mkldnn_reset();
+#endif
+    dims_ = dims;
+  }
   return *this;
 }
 
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -44,6 +44,13 @@ class Tensor {
     format_ = format;
   }
 
+  inline void mkldnn_reset() {
+    if (layout_ == DataLayout::kMKLDNN) {
+      layout_ = DataLayout::kNCHW;
+      format_ = mkldnn::memory::format::format_undef;
+    }
+  }
+
  protected:
   /**
    * @brief the detail format of memory block which have layout as kMKLDNN


jczaja:
@luotao1 PR_Windows_CI fails, I cannot get build logs for some reason. Could you please advice on how to proceed?

jczaja:
@luotao1 , @tensor-tang , @reyoung  Please start review

jczaja:
@luotao1 , @tensor-tang , @reyoung Do you need any additional explanation on that PR?

kbinias:
@jczaja Is it possible to move `#ifdef PADDLE_WITH_MKLDNN` block of code to other place ? IMO the reset of the MKL-DNN layout in the Tensor::resize() function can be misleading in the future.

luotao1:
>  Is it possible to move `#ifdef PADDLE_WITH_MKLDNN` block of code to other place?

Since `Tensor& Tensor::Resize(const DDim& dims)` is a very base interface, do you have any other solution to solve this problem?

jczaja:
@luotao1 , @kbinias Changes introduced by this PR are conceptually following the way Tensor class was modified for MKL-DNN . Alternatively We could introduce MKLDNNTensor, but few people tried(including myself) and this would require reimplementing of significant part of paddlepaddle and it is far from easy task. Crash to be fixed by this PR  is a general problem of MKL-DNN integration so similar crash may appear in other models than transformer. I currently do not see other solution to this problem.

jczaja:
@luotao1 I made a tiny refactoring of changes  as suggested (offline) by @kbinias and currently to get CI passing I need an approve from your team.

kbinias:
`I currently do not see other solution to this problem`
Agree, it would be nice to have a new kind of tensor with MKL-DNN properties.

LGTM

luotao1:
> Problem was detected when Transformer using MKL-DNN was executed, but using only
one mkldnn op: "elementwise_add". Result was a crash.

We have `analyzer_transformer_tester.cc` which runs on MKLDNN successfully. Could you tell us how to produce this issue?

jczaja:
@luotao1 To reproduce a crash in transformer it is needed to modify analyzer_transformer_tester.cc eg. to limit MKL-DNN to use only "elementwise_add".  

To be more precise add following code:
```
  std::unordered_set<std::string> op_list = {"elementwise_add"};
  cfg.SetMKLDNNOp(op_list);
```

after https://github.com/PaddlePaddle/Paddle/blob/dab71e8d97702f40c23772356711d162e8c6caa1/paddle/fluid/inference/tests/api/analyzer_transformer_tester.cc#L191

As I mentioned it is generic problem it just found it when working with transformer. This kind of problem may appear in any other model.

jczaja:
@luotao1 Were you able to reproduce a problem that this PR is fixing?

luotao1:
> Were you able to reproduce a problem that this PR is fixing?

I can reproduce this problem, and this PR fixes it.

jczaja:
@luotao1 Great! To have CI passed I need and approval , so It would be beneficial If you could carry on with review. If you have any other questions/comments I will be glad to hear them

kbinias (approved):
LGTM

jczaja:
Same functionality will be implemented when MKL-dNN 1.0 is to be introduced

PR #16899, opened 2019-04-16
Support cross compilation for ARM

hjchen2:
Enable setting thread number in Anakin Config

Fix code style

Reduce cmake file changes

test=release/1.4


Superjomn (approved):
LGTM

PR #16904, opened 2019-04-16
fix save and load bug, test=release/1.4

junjun315:
fix save bug for digraph

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_imperative_checkpoint.py
+++ b/python/paddle/fluid/tests/unittests/test_imperative_checkpoint.py
@@ -99,7 +99,7 @@ class MNIST(fluid.Layer):
 
 
 class TestDygraphCheckpoint(unittest.TestCase):
-    def save_load_persistables(self):
+    def test_save_load_persistables(self):
         seed = 90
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/tests/unittests/test_imperative_checkpoint.py
```diff
@@ -135,23 +135,26 @@ def save_load_persistables(self):
 
                     avg_loss.backward()
                     sgd.minimize(avg_loss)
-                    fluid.dygraph.save_persistables(mnist, "save_dir")
+                    fluid.dygraph.save_persistables(mnist.state_dict(),
```
velconia, line 138: why ```state_dict``` ? should we use mnist directly?

velconia (dismissed):

python/paddle/fluid/tests/unittests/test_imperative_checkpoint.py
junjun315: Thanks! I will modify it later.

JiabinYang (approved):
LGTM

PR #16914, opened 2019-04-16
Enable In-Place for MKLDNN Op. (solution1)

xiaolil1:
test=develop

We met in-place issue when enable MKLDNN Reshape OP. 
For the case: no-mkldnn op ---> mkldnn op ( Reshape or other in-place supported Ops).

We investigated this issue and found the "data transform" will generate new tensor and change the input tensor of Reshape Op, while the output not, which will break the in-place between input and output.

We submit this PR solve this problem.
We also filed issue #16916 to track it.

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1106,10 +1106,10 @@ Scope* OperatorWithKernel::PrepareData(
         for (auto& out_var_name_item : Outputs()) {
           std::vector<Variable*>& output_vars =
               ctx->outputs[out_var_name_item.first];
-          for (size_t i = 0; i < out_var_name_item.second.size(); ++i) {
-            auto& out_var_name = out_var_name_item.second[i];
+          for (size_t id = 0; id < out_var_name_item.second.size(); ++id) {
+            auto& out_var_name = out_var_name_item.second[id];
             if (out_var_name == var_name) {
-              output_vars[i] = trans_var;
+              output_vars[id] = trans_var;
             }
           }
         }
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1068,8 +1068,11 @@ Scope* OperatorWithKernel::PrepareData(
       }
 
       auto out_var_names = OutputVars(true);
+      bool is_inplace = false;  // Output variable of in-place op should be
+                                // transfered as well.
       if (std::find(out_var_names.begin(), out_var_names.end(), var_name) !=
           out_var_names.end()) {
+        is_inplace = true;
         transfered_inplace_vars->emplace_back(var_name);
       }
 
@@ -1098,6 +1101,19 @@ Scope* OperatorWithKernel::PrepareData(
 
       auto* trans_var = new_scope->Var(var_name);
       input_vars[i] = trans_var;
+
+      if (is_inplace) {
+        for (auto& out_var_name_item : Outputs()) {
+          std::vector<Variable*>& output_vars =
+              ctx->outputs[out_var_name_item.first];
+          for (size_t i = 0; i < out_var_name_item.second.size(); ++i) {
+            auto& out_var_name = out_var_name_item.second[i];
+            if (out_var_name == var_name) {
+              output_vars[i] = input_vars[i];
+            }
+          }
+        }
+      }
 
       Tensor out;
       TransformData(expected_kernel_key, kernel_type_for_var, *tensor_in, &out);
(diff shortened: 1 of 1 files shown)


luotao1:
please label 'intel' for each PR.

paddle/fluid/framework/operator.cc
```diff
@@ -1099,6 +1102,19 @@ Scope* OperatorWithKernel::PrepareData(
...
+            if (out_var_name == var_name) {
+              output_vars[i] = input_vars[i];
+            }
+          }
+        }
+      }
+
       Tensor out;
       TransformData(expected_kernel_key, kernel_type_for_var, *tensor_in, &out);
       SetTensorToVariable(*var, out, trans_var);
```
jczaja, line 1120: @xiaolil1 Could you please explain why scenario you are describing : non-mkldnn tensor --> mkldnn tensor  when op is doing inplace computation does not work without this PR? My understanding is that
if this is inplace then var_name for input and output is the same. Both output and input var_names should point to the same variable , so by calling SetTensorToVariable we should have both input and output updated?

xiaolil1: Hi @jczaja , thanks for your quickly review. As you said, var_name for input and output is the same and point to the same tensor at the beginning, but "data transform" between "non-mkldnn tensor --> mkldnn" tensor will create new tensor and replace the original input tensor, you can find it in line:1100 ( i have marked it), while output doesn't. So after the replacement, input and output will not point to same tensor, we need to add the same mechanism as input to keep the "in-place". What's more, i think not only the "non-mkldnn tensor --> mkldnn", but also other cases need "data_transform" will meet this issue. @hshen14

paddle/fluid/framework/operator.cc
```diff
@@ -1099,6 +1102,19 @@ Scope* OperatorWithKernel::PrepareData(
       auto* trans_var = new_scope->Var(var_name);
       input_vars[i] = trans_var;
```
xiaolil1, line 1103: Here, trans_var will replace input_vars[i], as well as its' tensor.

jczaja: ok

jczaja:
@xiaolil1 changes looks good to me , but due to fact that this is change in a core of Tensor transformations, I would feel more confident if someone more familiar with that area also took a look.
@luotao1 Please review those changes as well or assign someone familiar with that code.

paddle/fluid/framework/operator.cc
```diff
@@ -1099,6 +1102,19 @@ Scope* OperatorWithKernel::PrepareData(
...
 
+      if (is_inplace) {
+        for (auto& out_var_name_item : Outputs()) {
+          std::vector<Variable*>& output_vars =
+              ctx->outputs[out_var_name_item.first];
+          for (size_t i = 0; i < out_var_name_item.second.size(); ++i) {
+            auto& out_var_name = out_var_name_item.second[i];
+            if (out_var_name == var_name) {
+              output_vars[i] = input_vars[i];
```
ghost: here output_vars[i] = input_vars[i] is incorrect. "i" in the index for output_vars[] and should NOT apply to input_vars[] which is totally independant with output_vars[]. Actually this local var "i" in used in both two nested "for" which is not a GOOD practice.

xiaolil1: @jianhang-liu Thanks for you reminder, i will use "output_vars[i] = trans_var;" here, due to line 1103 "input_vars[i] = trans_var", which will be safer and make sense.

ghost: @xiaolil1 output_vars[i] = input_vars[i] is NOT correct, not only "unsafe". Unfortunately there are two "i" defined in nested for loop. [The first "i"](https://github.com/PaddlePaddle/Paddle/blob/8d0d6185ce3b3c7571b34316152ece5dd361f859/paddle/fluid/framework/operator.cc#L1049) is index for input_vars[]  and [the second "i"](https://github.com/PaddlePaddle/Paddle/blob/8d0d6185ce3b3c7571b34316152ece5dd361f859/paddle/fluid/framework/operator.cc#L1109) (in inner loop) is index for output_vars[].
So the code "[output_vars[i] = input_vars[i]](https://github.com/PaddlePaddle/Paddle/blob/8d0d6185ce3b3c7571b34316152ece5dd361f859/paddle/fluid/framework/operator.cc#L1112)" doesn't make sense. It's totally wrong. It happen to work because you run with output_vars[0] = input_vars[0].
As mentioned in my last comment, duplicated "i" in nested loop is NOT a good practice.

ghost:
Close without merge. Risk in solution.

PR #16915, opened 2019-04-16
Enable In-Place for MKLDNN Op. (solution2)

xiaolil1:
test=develop

We met in-place issue when enable MKLDNN Reshape OP.
For the case: no-mkldnn op ---> mkldnn op ( Reshape or other in-place supported Ops).

We investigated this issue and found the "data transform" will generate new tensor and change the input tensor of Reshape Op, while the output not, which will break the in-place between input and output.

We developed two solutions to solve this problem, current PR is the second solution and the first solutions is #16914.
We also file issue #16916 to track it.


xiaolil1:
We prefer #16914 , the first solution, please do not review this PR without our requirement, thanks!

ghost:
Close without merge. Risk in solution.

PR #16973, opened 2019-04-18
Refine gpu mem of models

sneaxiy:
This PR works for:
1. Speed up GC by highest priority. Thread switching would do harm to train speed.
2. Refine affine_channel op to save memory when using gc.
3. Add relu memory optimize pass to replace relu with relu2, relu_grad with relu2_grad, with outputing an extra mask in forward part to save memory. The reason why I do not replace relu with relu2 in Python API is that too many passes handles relu (many op fusion passes). We can use a flag to disable relu_memory_optimize_pass to support the former op fusion passes.

This PR improves 15% speed of Transformer model in V100 8-GPUs machine, and makes the maximum batch size of Mask RCNN model increase from 3 to 5.


paddle/fluid/framework/details/build_strategy.cc
```diff
@@ -53,6 +53,9 @@ class ParallelExecutorPassBuilder : public ir::PassBuilder {
       viz_pass->Set<std::string>("graph_viz_path", new std::string(graph_path));
     }
 
+    // Note(zcd): record_skip_memory_opt_vars_pass should be the first pass.
```
zhhsplendid, line 56: Recommend to change to Github ID "chengduoZH" ...

paddle/fluid/framework/details/fast_threaded_ssa_graph_executor.cc
```diff
@@ -131,32 +132,53 @@ FeedFetchList FastThreadedSSAGraphExecutor::Run(
...
+
       auto &outputs = op_to_run->Outputs();
       op_to_run = nullptr;
       for (auto &output : outputs) {
         for (auto &pending_op : output->PendingOps()) {
           std::atomic<int> &deps = op_deps->at(pending_op);
-          if (deps.fetch_sub(1) == 1) {  // pending_op ready
+          if (deps.fetch_sub(1) != 1) continue;
+
+          // NOTE(zjl): op with highest priority should run
```
zhhsplendid, line 177: Suggest to change to Github ID to prevent outside developer from confusion. Same at other places

python/paddle/fluid/dygraph/layer_object_helper.py
```diff
@@ -200,10 +200,20 @@ def append_activation(self,
         act_type = act.pop('type')
 
         tmp = self.create_variable_for_type_inference(dtype=input_var.dtype)
+
+        if act_type == 'relu':
+            mask = self.create_variable_for_type_inference(
+                dtype=core.VarDesc.VarType.UINT8)
+            mask.stop_gradient = True
+            act_type = 'relu2'
```
zhhsplendid: You created relu2 op and if act_type == 'relu', replace it with the new 'relu2'. I guess it is for backward compatibility. Do we have a future plan to remove "2" and replace 'relu' completely?

python/paddle/fluid/layers/nn.py
```diff
@@ -7860,8 +7860,18 @@ def relu(x, name=None):
     helper = LayerHelper('relu', **locals())
     dtype = helper.input_dtype(input_param_name='x')
     out = helper.create_variable_for_type_inference(dtype)
-    helper.append_op(
-        type="relu", inputs={"X": helper.input('x')}, outputs={"Out": out})
+    if x.type != core.VarDesc.VarType.LOD_TENSOR:
```
zhhsplendid: I have two questions.

1. Why replace relu2 when x is LoDTensor, but in layer_helper, the activation function is always replaces. Replacing with different logic makes more prior knowledge requirement for users/software engineers .

2. TODO in the future.
Is relu2 a good name as a new op? Maybe relu_with_mask (Or in the future, mask is an optional field for old relu)

python/paddle/fluid/dygraph/layer_object_helper.py
sneaxiy: Replace with an automatic pass.

python/paddle/fluid/layers/nn.py
sneaxiy: Replace with an automatic pass.

sneaxiy:
Close this PR, since relu2_op is too complicated and is not necessary for models after testing.

Move to #16993.

Issue #16841, opened 2019-04-13
[MKL-DNN] Advice on runtime context related crash in Transformer


jczaja:
When test_analyzer_transformer.profile_mkldnn is run with "--test_all_data" option there is a crash (matmul inputs has mismatch of dims). Problem is that when MKL-DNN is used (any op)
then starting from third iteration (when Input data shapes will differ from Warmup and first iteration) then wrong Tensor is chosen for op to work. 

Problem is that runtime context is having a Variables pointing to tensors "scale_0.tmp_0" that 
comes from previous iterations when shape of those Tensors was diffrent. At the same time
scope does contain "scale_0_tmp_0" of good proper dims.
As a test/workround I forced paddle to recreate runtime context each time. With this hack crash goes away (it all pass). Code I'm refering to (that was disabled to force recreate of runtime context):
https://github.com/PaddlePaddle/Paddle/blob/4267a81afcab6ccc4d84eab8ffad0dff24fd8d65/paddle/fluid/framework/operator.cc#L897

@luotao1 Could you please advice/suggest on how to fix this problem?

luotao1:
When use the latest develop branch. I can run it normally without `MKLDNN`
```
./paddle/fluid/inference/tests/api/test_analyzer_transformer --gtest_filter=Analyzer_Transformer.profile --infer_model=third_party/inference_demo/transformer/model --infer_data=third_party/inference_demo/transformer/data.txt --paddle_num_threads=8 --repeat=1 --batch_size=8 --test_all_data
Note: Google Test filter = Analyzer_Transformer.profile
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from Analyzer_Transformer
[ RUN      ] Analyzer_Transformer.profile
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0414 19:27:37.383009 26581 analyzer_transformer_tester.cc:176] The number of samples to be test: 200
I0414 19:27:37.523564 26581 tester_helper.h:85] AnalysisConfig {
  NativeConfig {
    PaddlePredictor::Config {
      model_dir:
    }
    use_gpu: 0
    device: 0
    fraction_of_gpu_memory: 0
    specify_input_name: 1
    cpu_num_threads: 1
  }
  prog_file: third_party/inference_demo/transformer/model/model
  param_file: third_party/inference_demo/transformer/model/params
  enable_ir_optim: 1
  enable_ir_optim: 1
[...]

PR #16981, opened 2019-04-19
disable runtime_context_cache pass by default

luotao1:
Hotfix #16841 

For gpu inference, if a gpu model has an op which could only run on CPU, each result of different input will be the same with the first one.
analysis predictor gpu模式下，打开pass优化的时候，如果一个模型中有op只有cpu kernel， 会出现，所有的运行的结果和第一次运行的一样。

The reason is that if a gpu tensor is the input of a cpu kernel, we will create a new cpu tensor in new scope. if enable cache runtime context, we get the cpu tensor each time, not the gpu tensor.
一个gpu 的tensor 做为cpu kernel 输入的话，会在一个新的scope中创建一个新的cpu tensor， 如果进行cache runtime context, 下次直接拿到的是cpu tensor， 而不是gpu tensor
https://github.com/PaddlePaddle/Paddle/blob/4098ba29ed180e5035fc90573b8eeef8224f373e/paddle/fluid/framework/operator.cc#L1099-L1100

### DIFF
--- a/paddle/fluid/inference/api/paddle_pass_builder.cc
+++ b/paddle/fluid/inference/api/paddle_pass_builder.cc
@@ -86,8 +86,7 @@ const std::vector<std::string> kAnakinSubgraphPasses({
 
 GpuPassStrategy::GpuPassStrategy() : PassStrategy({}) {
   passes_.assign({
-    "infer_clean_graph_pass",          //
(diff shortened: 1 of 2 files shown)


NHZlX (approved):
LGTM

PR #16993, opened 2019-04-20
Refine model gpu memory

sneaxiy:
This PR works for:

- Speed up GC by highest priority. Thread switching would do harm to train speed.
- Refine affine_channel op to save memory when using gc.
- Combining PR #16968 to fix inplace bug.

This PR improves 15% speed of Transformer model in V100 8-GPUs machine, and makes the maximum batch size of Mask RCNN model increase from 3 to 5.


paddle/fluid/framework/details/op_handle_base.h
```diff
@@ -31,6 +33,17 @@ constexpr char kLocalExecScopeName[] = "@LOCAL_SCOPE@";
 // It's responsible for populating necessary fields of ir::Node.
 class OpHandleBase {
  public:
+  /**
+   * NOTE(zjl): Some op should have higher priority than others.
+   * The higher priority op would run first without switching
+   * threads in FastThreadedSSAGraphExecutor.
```
chengduoZH: The `OpHandle` should not be bound with a specified Executor. So the note should be `The higher priority op should be run first without switching threads in Executor.`

paddle/fluid/framework/details/op_handle_base.h
```diff
@@ -31,6 +33,17 @@ constexpr char kLocalExecScopeName[] = "@LOCAL_SCOPE@";
...
+   *
+   * Currently, EagerDeletionOpHandle has the highest priority.
+   * This priority settings speed up gc 15% in Transformer
+   * V100 8-GPU model.
+   */
```
chengduoZH, line 40: line40~44 should be placed to [here](https://github.com/PaddlePaddle/Paddle/pull/16993/files#diff-d9f7d8760a10c9ad6e5e0f8049971939R43).

paddle/fluid/framework/details/eager_deletion_op_handle.h
```diff
@@ -37,14 +40,16 @@ class EagerDeletionOpHandle : public OpHandleBase {
...
+
  protected:
   void RunImpl() override;
 
  private:
   void ClearGarbages(std::deque<std::shared_ptr<memory::Allocation>> *garbages);
 
   const Scope *scope_;
-  std::unordered_set<std::string> var_names_;
+  std::vector<std::string> var_names_;
```
chengduoZH, line 57: Why do you replace `std::unordered_set` with `std::vector` here.

paddle/fluid/framework/details/inplace_op_pass.cc
```diff
@@ -78,6 +78,13 @@ const std::string kInplacedOpWhiteList[] = { // NOLINT
     "elementwise_add",
     "elementwise_add_grad",
 };
+
+// FIXME(zjl): Shapes of in-out of some ops are exactly the same,
+// but the static size during compiling time would be wrong.
+// Use a flag to indicate such ops. Please fix me when found a better way.
+static const std::unordered_set<std::string> kSameShapeOpWhiteSet{ // NOLINT
+    "reshape2"
+};
```
chengduoZH, line 87: Why? not understand what's the function of `kSameShapeOpWhiteSet`.

paddle/fluid/framework/details/op_handle_base.h
sneaxiy: Done.

paddle/fluid/framework/details/op_handle_base.h
sneaxiy: Done.

paddle/fluid/framework/details/eager_deletion_op_handle.h
sneaxiy: It does not matter. Just for faster transversing of `var_names_`.

paddle/fluid/framework/details/inplace_op_pass.cc
sneaxiy: Currently, inplace pass would only make in-out inplace when their compile-time shapes are the same. But in some nlp tasks, the compile-time shapes of in-out of reshape op are not the same. For example, shape of X is [-1, 3, 30] and shape of Out would be [-1, 30]. But in-out of some ops (such as reshape op) are exactly the same, regardless of the mismatching compile-time shapes. See [here](https://github.com/PaddlePaddle/Paddle/pull/16993/files#diff-076f3065cd0e7534a4f82bf22457869cR348).

This changing decreases the GPU memory usage of Padding RNN model a lot.

chengduoZH: I see, thanks.

chengduoZH (approved):
LGTM

PR #16999, opened 2019-04-21
Move gc test to each test of op

sneaxiy:


python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -371,6 +371,20 @@ def check_output_with_place(self,
                                 no_check_set=None,
                                 equal_nan=False,
                                 check_dygraph=False):
+        self._check_output_with_place(place, atol, no_check_set, equal_nan,
+                                      check_dygraph)
+        if self._is_gc_test_enabled():
+            self._enable_gc(True)
+            self._check_output_with_place(place, atol, no_check_set, equal_nan,
+                                          check_dygraph)
+            self._enable_gc(False)
```
chengduoZH: If ` self._is_gc_test_enabled()` is true, the `self._check_output_with_place` will be called two times, is it necessary?

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -503,6 +530,13 @@ def check_output(self,
                                          check_dygraph)
 
     def check_output_customized(self, checker):
+        self._check_output_customized(checker)
+        if self._is_gc_test_enabled():
+            self._enable_gc(True)
+            self._check_output_customized(checker)
+            self._enable_gc(False)
```
chengduoZH: Same above.

python/paddle/fluid/tests/unittests/test_fill_constant_batch_size_like_op.py
```diff
@@ -21,6 +21,7 @@
 
 class TestFillConstantBatchSizeLikeWhenFirstDimIsBatchSize(OpTest):
     def setUp(self):
+        self.test_gc = True
```
chengduoZH: Which unit tests should use `self.test_gc = True`?

python/paddle/fluid/tests/unittests/op_test.py
sneaxiy: For safety check, though I have not figured out any case that the calculating result would be different if using gc or not.

python/paddle/fluid/tests/unittests/op_test.py
sneaxiy: For safety check, though I have not figured out any case that the calculating result would be different if using gc or not.

python/paddle/fluid/tests/unittests/test_fill_constant_batch_size_like_op.py
sneaxiy: Currently, only op that registers `NoNeedBufferVarsInference` should add this flag.

chengduoZH (approved):
LGTM

PR #17000, opened 2019-04-21
Enable conv+simoid fusion (FP32)

liangan1:
Conv + Sigmoid fusion is a key optimization for Squeeze-and-Excitation Networks (https://arxiv.org/pdf/1709.01507.pdf). The representative model is SE-ResNext and its variant. With this fusion, we can reduce the memory footprint for Conv + Sigmoid and therefore improve the performance. As  MKL-DNN supports the primitive post op in FP32, we submit the PR to support FP32 first and will add INT8 once MKL-DNN provides the support.


liangan1:
@sfraczek  @wojtuss  please help to review this PR. Thanks.

liangan1:
unexpected code conflicts after merging the public code. to simplify the PR process, I created another one.

PR #17005, opened 2019-04-22
fix nccl wrapper on windows

wopeizl:
test=develop


guru4elephant (approved):
Thx for your fix, I will care more about compiling on windows later.

PR #17015, opened 2019-04-22
Enhance concat op to support empty input.

jerrywgz:
Enhance concat op to allow empty variable in inputs.
For example, In Faster-RCNN-FPN model, RoIs are distributed onto different levels and get roi features by roi_align. In such case, the number of rois in one level might be 0, then causes empty input and output in roi_align. As result, we should support empty variable when use concat op to concatenate roi features.

### DIFF
--- a/paddle/fluid/operators/concat_op.h
+++ b/paddle/fluid/operators/concat_op.h
@@ -37,7 +37,7 @@ class ConcatKernel : public framework::OpKernel<T> {
     if (axis == 0 && ins.size() < 10) {
       size_t output_offset = 0;
       for (auto* in : ins) {
-        if (in->numel() == 0UL) {
+        if (!in || in->numel() == 0UL) {
           continue;
         }
         auto in_stride = framework::stride_numel(in->dims());
@@ -50,10 +50,10 @@ class ConcatKernel : public framework::OpKernel<T> {
     } else {
       std::vector<framework::Tensor> inputs;
       for (size_t j = 0; j < ins.size(); ++j) {
-        if (ins[j]->numel() == 0UL) {
-          continue;
-        } else {
+        if (ins[j] || ins[j]->numel() > 0) {
           inputs.push_back(*ins[j]);
+        } else {
+          continue;
         }
       }
       auto& dev_ctx = ctx.template device_context<DeviceContext>();
--- a/paddle/fluid/operators/concat_op.h
+++ b/paddle/fluid/operators/concat_op.h
@@ -37,6 +37,9 @@ class ConcatKernel : public framework::OpKernel<T> {
     if (axis == 0 && ins.size() < 10) {
       size_t output_offset = 0;
       for (auto* in : ins) {
+        if (in->numel() == 0UL) {
+          continue;
+        }
         auto in_stride = framework::stride_numel(in->dims());
         auto out_stride = framework::stride_numel(out->dims());
         StridedNumelCopyWithAxis<T>(ctx.device_context(), axis,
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/concat_op.h
```diff
@@ -37,6 +37,9 @@ class ConcatKernel : public framework::OpKernel<T> {
     if (axis == 0 && ins.size() < 10) {
       size_t output_offset = 0;
       for (auto* in : ins) {
+        if (in->numel() == 0UL) {
```
qingqing01: Maybe better to check:

```c++
if (!in || in->numel() == 0UL) {
}
```

paddle/fluid/operators/concat_op.h
```diff
@@ -45,9 +48,13 @@ class ConcatKernel : public framework::OpKernel<T> {
...
     } else {
-      std::vector<framework::Tensor> inputs(ins.size());
+      std::vector<framework::Tensor> inputs;
       for (size_t j = 0; j < ins.size(); ++j) {
-        inputs[j] = *ins[j];
+        if (ins[j]->numel() == 0UL) {
+          continue;
+        } else {
+          inputs.push_back(*ins[j]);
+        }
```
qingqing01, line 57:
```c++
if (ins[j] || in->numel() > 0) {
  inputs.push_back(*ins[j]);
} else {
  continue;
}
```

qingqing01 (commented):
@chengduoZH Please also help to review. Thanks!

chengduoZH:
@jerrywgz You should add unit test for this change.

paddle/fluid/operators/concat_op.h
jerrywgz: Done

paddle/fluid/operators/concat_op.h
jerrywgz: Done, but here should be || or &&?

jerrywgz:
> @jerrywgz You should add unit test for this change.

Done

paddle/fluid/operators/concat_op.h
qingqing01: Yeah, should be &&

python/paddle/fluid/tests/unittests/test_concat_op.py
```diff
@@ -64,5 +64,16 @@ def test_check_grad(self):
...
 
+class TestConcatOp4(TestConcatOp):
+    def init_test_data(self):
+        self.x0 = np.random.random((2, 3, 4, 5)).astype('float32')
+        self.x1 = np.random.random((2, 3, 4, 5)).astype('float32')
+        self.x2 = np.random.random((0, 3, 4, 5)).astype('float32')
+        self.axis = 0
+
+    def test_check_grad(self):
+        pass
```
qingqing01, line 75: Why pass check_grad ?

jerrywgz: Because it will meet fetch error when there is a null grad

paddle/fluid/operators/concat_op.h
qingqing01: @jerrywgz please to fix code

chengduoZH:
@jerrywgz Why the inputs have some empty argument?

jerrywgz:
> @jerrywgz Why the inputs have some empty argument?

In FPN detection model, we need to divide rois into four parts and it may produce empty rois in one part. Concat op in caffe2 can support empty argument. 
Or maybe we could add a warning when there's empty argument?

paddle/fluid/operators/concat_op.h
jerrywgz: Done

qingqing01 (approved):

PR #17029, opened 2019-04-22
add checkpoint functions for graph. test=develop

wzzju:
* Add the `fluid.io._save_persistable_nodes`  function for saving persistable nodes of a graph.
* Add the `fluid.io._load_persistable_nodes`  function for loading persistable nodes of a graph.

`fluid.io._save_persistable_nodes` and `fluid.io._load_persistable_nodes` can be used to implement the checkpoint function for graph training. But these two graph helper functions should not be visible to the user.


wanghaoshuang (approved):
LGTM.

PR #17036, opened 2019-04-22
Make conv cudnn workspace size configurable

sneaxiy:
Currently, CUDNN workspace size cannot be controlled by environment variable `FLAGS_conv_workspace_size_limit`. This PR enables this control.


paddle/fluid/operators/conv_cudnn_op.cu.cc
```diff
@@ -127,7 +129,7 @@ class CUDNNConvOpKernel : public framework::OpKernel<T> {
     int group_offset_filter = filter->numel() / groups;
     // ------------------- cudnn conv workspace ---------------------
     size_t workspace_size_in_bytes;  // final workspace to allocate.
-    size_t workspace_size_limit = kCONV_CUDNN_WORKSPACE_LIMIT_BYTES;
+    size_t workspace_size_limit = 0;
     if (FLAGS_conv_workspace_size_limit > 0 || user_workspace_size > 0) {
       int64_t max_user_size =
           std::max(static_cast<int64_t>(FLAGS_conv_workspace_size_limit),
```
chengduoZH: I think that we should use `std::min` here.

sneaxiy: Done.

chengduoZH (approved):
Excellent

PR #17048, opened 2019-04-23
fix runtime_context_cache bug when gpu model has an op runs only on cpu

### LINKED ISSUE #16841: [MKL-DNN] Advice on runtime context related crash in Transformer

luotao1:
related #16981
fix #16841

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1095,6 +1095,17 @@ Scope* OperatorWithKernel::PrepareData(
       if (!new_scope) {
         new_scope = &scope.NewScope();
       }
+      // For inference, if a gpu model has an op which could only run on CPU,
+      // each result of different input will be the same with the first one.
+      // The reason is that if a gpu tensor is the input of a cpu kernel,
+      // we will create a new cpu tensor in new scope.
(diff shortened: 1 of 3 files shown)


NHZlX (approved):
Have Tested. LGTM.

PR #17054, opened 2019-04-23
remove all SERIAL since the parallel test would not use that anymore …

wopeizl:
…test=develop


luotao1 (commented):
去掉后需要加`RUN_TYPE=EXCLUSIVE`么?
或者先并行跑看是否有问题？

luotao1:
close due to #17500

PR #17071, opened 2019-04-24
add test_conv2d_op to exclusively test case list

wopeizl:
test=develop


luotao1:
Why add `test_conv2d_op` to exclusively test case list, is there some error?

wopeizl:
> Why add `test_conv2d_op` to exclusively test case list, is there some error?

Seems this case run out of memory when GPU memory is under 4GB, and we have one machine with 4GB only.

luotao1:
> Seems this case run out of memory when GPU memory is under 4GB, and we have one machine with 4GB only.

How about adding this explanation in the codes?

wopeizl:
close this since the test case may randomly fail on other part

PR #17082, opened 2019-04-24
Use CudnnWorkspaceHandle in exhaustive search

zhhsplendid:
1. Use CudnnWorkspaceHandle in exhaustive search of conv_cudnn.
2. For ops using CudnnWorkspaceHandle in exhaustive search, release their GPU memory after exhaustive search.

test=develop

### DIFF
--- a/paddle/fluid/platform/device_context.h
+++ b/paddle/fluid/platform/device_context.h
@@ -163,6 +163,13 @@ class CudnnHolder {
     cudnn_func(WorkspacePtr());
   }
 
+  /*! \brief Reset workspace thus release the memory */
+  inline void ResetWorkspace() {
+    if (workspace_) {
+      workspace_ = nullptr;
+    }
+  }
+
   inline void* WorkspacePtr() {
     if (workspace_) {
       return workspace_->ptr();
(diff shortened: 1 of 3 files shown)


paddle/fluid/platform/device_context.h
```diff
@@ -207,6 +214,20 @@ class CudnnWorkspaceHandle {
...
+   *  functions */
+  template <typename Callback>
+  inline void RunFuncReleaseMem(Callback&& cudnn_func,
+                                size_t required_workspace_len) {
+    if (!guard_) {
+      guard_.reset(new std::lock_guard<std::mutex>(holder_->Mutex()));
+    }
+    holder_->RunFuncImpl(std::forward<Callback>(cudnn_func),
+                         required_workspace_len);
+    holder_->ResetWorkspace();
```
chengduoZH, line 232: The execution of CUDA code is async, so you should call `cudaStreamSync` before reset workspace.

sneaxiy: Maybe we should call this method `RunFuncSync`? Because this function may be only used when cudnn exhaustive searching, and `cudnnFindConvolutionForwardAlgorithmEx` is a synchronized method as described in [here](https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnFindConvolutionForwardAlgorithmEx).

@zhhsplendid Please also add some comments on this method, something like: you have to guarantee that the function is host blocking.

zhhsplendid: @chengduoZH Done. I added inside ResetWorkspace() to enforce users to call cudaStreamSynchronize.

@sneaxiy Done

sneaxiy (approved):
Excellent.

PR #17086, opened 2019-04-25
fix runtimeerror in uniform_random

fc-liu:
fix runtimeerror : dictionary changed size during iteration when calling uniform_random in python3+

### DIFF
--- a/python/paddle/fluid/layers/ops.py
+++ b/python/paddle/fluid/layers/ops.py
@@ -86,12 +86,11 @@ def uniform_random(shape, dtype='float32', min=-1.0, max=1.0, seed=0):
             result = fluid.layers.uniform_random(shape=[32, 784])
     """
 
-    locals_var = locals().keys()
+    locals_var = locals()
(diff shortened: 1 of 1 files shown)


luotao1:
Please add `test=develop` in commit info to trigger the CI.
http://paddlepaddle.org/documentation/docs/zh/1.4/advanced_usage/development/contribute_to_paddle/local_dev_guide.html#permalink-8--commit-
image

fc-liu:
> Please add `test=develop` in commit info to trigger the CI.
> http://paddlepaddle.org/documentation/docs/zh/1.4/advanced_usage/development/contribute_to_paddle/local_dev_guide.html#permalink-8--commit-
> image

I modified the commit infomation to "test=develop", but happend to change the user name to Liu(not exist in github, but on my PC)，the signature check is pending here...

luotao1:
> but happend to change the user name to Liu(not exist in github, but on my PC)，the signature check is pending here

This PR LGTM, but we need CLA pass. How about create a new PR with Lfc1993 not Liu again? Thanks very much for your contribution!

PR #17097, opened 2019-04-25
Enable MKLDNN INT8 FC kernel.

xiaolil1:
This PR is only for reference, no need to run pre-ci and merge.

To make INT8 kernel more readable, i make a temp file named "fc_mkldnn_int8_op.cc", it will be merged and align to MKLDNN FP32 kernel "fc_mkldnn_op.cc" coding style as well as adding UT if there is PR merge requirement in the future.

### DIFF
--- a/paddle/fluid/operators/fc_op.cc
+++ b/paddle/fluid/operators/fc_op.cc
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 #include "paddle/fluid/operators/fc_op.h"
+#include <string>
 #include <vector>
 #include "paddle/fluid/operators/math/blas.h"
 #include "paddle/fluid/operators/math/fc_compute.h"
@@ -66,12 +67,20 @@ framework::OpKernelType FCOp::GetExpectedKernelType(
     const framework::ExecutionContext& ctx) const {
   framework::LibraryType library = framework::LibraryType::kPlain;
   framework::DataLayout layout = framework::DataLayout::kAnyLayout;
+  int customized_type_value =
+      framework::OpKernelType::kDefaultCustomizedTypeValue;
+  auto input_data_type = ctx.Input<Tensor>("Input")->type();
   if (ctx.Attr<bool>("use_mkldnn")) {
     library = framework::LibraryType::kMKLDNN;
     layout = framework::DataLayout::kMKLDNN;
+    customized_type_value =
+        (input_data_type == framework::DataTypeTrait<int8_t>::DataType ||
+         input_data_type == framework::DataTypeTrait<uint8_t>::DataType)
+            ? kFCMKLDNNINT8
+            : kFCMKLDNNFP32;
   }
-  return framework::OpKernelType(ctx.Input<Tensor>("Input")->type(),
-                                 ctx.GetPlace(), layout, library);
+  return framework::OpKernelType(input_data_type, ctx.GetPlace(), layout,
+                                 library, customized_type_value);
 }
 
 void FCOpGrad::InferShape(framework::InferShapeContext* ctx) const {
@@ -110,6 +119,25 @@ void FCOpMaker::Make() {
   AddInput("W", "(Tensor), The weight fc op with shape (I, O).");
   AddInput("Bias", "(Tensor, optional) Bias vector with shape (1 x O")
       .AsDispensable();
+
+  AddAttr<float>("Scale_in",
+                 "(float, default 1.0f), The quantize scale of input data")
+      .SetDefault(1.0f);
+  AddAttr<std::vector<float>>("Scale_weights",
+                              "(std::vector<float>, default {1.0f}), The "
+                              "quantize scale of weights data")
+      .SetDefault({1.0f});
+  AddAttr<float>("Scale_out",
+                 "(float, default 1.0f), The quantize scale of output data")
+      .SetDefault(1.0f);
+  AddAttr<std::string>(
+      "data_format",
+      "(string, default NCHW) Only used in "
+      "An optional string from: \"NHWC\", \"NCHW\". "
+      "Defaults to \"NHWC\". Specify the data format of the output data, "
+      "the input will be transformed automatically. ")
+      .SetDefault("AnyLayout");
+
   AddAttr<int>("in_num_col_dims",
                "(int, default 1), The fc op can take tensors with more than "
                "two dimensions as its inputs.")
@@ -122,6 +150,12 @@ void FCOpMaker::Make() {
   AddAttr<bool>(framework::kAllKernelsMustComputeRuntimeShape,
                 "Skip calling InferShape() function in the runtime.")
       .SetDefault(true);
+  AddAttr<bool>("fuse_relu", "(bool, default false) Only used in mkldnn kernel")
+      .SetDefault(false);
+  AddAttr<bool>("force_fp32_output",
+                "(bool, default false) Force INT8 kernel output FP32, only "
+                "used in MKL-DNN INT8")
+      .SetDefault(false);
   AddComment(R"DOC(
   Fully Connected Operator.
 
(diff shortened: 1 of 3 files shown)


xiaolil1:
@jianhang-liu @luotao1

luotao1:
Does `fc_mkldnn_int8_op.cc` need weights transposed before run like `fc_mkldnn_op.cc` in #15226?

xiaolil1:
> Does `fc_mkldnn_int8_op.cc` need weights transposed before run like `fc_mkldnn_op.cc` in #15226?

@luotao1 Yes, we base on the #15226, weights transpose.

luotao1:
If weights transpose:
- Is the `cpu_int8_save_model` different with other `int8_save_model`? And you should do another weights transpose before `save_model`?
- How do you deal with `fc` + `fc`?

paddle/fluid/operators/fc_op.cc
```diff
@@ -110,6 +119,25 @@ void FCOpMaker::Make() {
...
+  AddAttr<float>("Scale_out",
+                 "(float, default 1.0f), The quantize scale of output data")
+      .SetDefault(1.0f);
+  AddAttr<std::string>(
+      "data_format",
+      "(string, default NCHW) Only used in "
+      "An optional string from: \"NHWC\", \"NCHW\". "
+      "Defaults to \"NHWC\". Specify the data format of the output data, "
+      "the input will be transformed automatically. ")
+      .SetDefault("AnyLayout");
```
luotao1, line 139: Although this is a temp file, please take it much seriously for reviewers. fc doesn't need data_format.

wzzju, line 139: Does FC have `data_format` NCHW or NHWC? What exactly is its default value?

paddle/fluid/operators/fc_op.cc
```diff
@@ -110,6 +119,25 @@ void FCOpMaker::Make() {
   AddInput("W", "(Tensor), The weight fc op with shape (I, O).");
   AddInput("Bias", "(Tensor, optional) Bias vector with shape (1 x O")
       .AsDispensable();
+
+  AddAttr<float>("Scale_in",
+                 "(float, default 1.0f), The quantize scale of input data")
+      .SetDefault(1.0f);
+  AddAttr<std::vector<float>>("Scale_weights",
```
wzzju, line 126: Channel-wise quantization should only be applied on the convolution op. There is no channel concept for fc/mul op. So, I think the weight of fc/mul  op should do the whole tensor quantization, namely one scale value. TensoRT also takes this strategy.

paddle/fluid/operators/fc_op.cc
xiaolil1: @luotao1 "data_format" was needed by FC INT8 kernel, please see [line150](https://github.com/PaddlePaddle/Paddle/pull/17097/files/edd7bd3ecce6bc9d619f2e415273b977abac9165#diff-cd3543641d6c976ea46111fc3dfab3f6R150) for it.

paddle/fluid/operators/fc_op.cc
xiaolil1: We supporting both tensor wise and channel wise weights scale in kernel, channel is dim-1 exactly. Accuracy with channel wise scale will be better.

paddle/fluid/operators/fc_op.cc
hshen14: The attribute was firstly introduced by @dzhwinter in conv op for cuda device and followed by other ops. Normally for CPU device, the default value is described in attribute comment, namely AnyLayout. with AnyLayout, the op kernel will ask MKL-DNN to do the reorder to the performant format, e,g., NHWC or NCHW16C etc.

hshen14:
> If weights transpose:
> 
> * Is the `cpu_int8_save_model` different with other `int8_save_model`? And you should do another weights transpose before `save_model`?
> * How do you deal with `fc` + `fc`?

1. It depends on how the API defines. E.g., if int8_save_model aims to provide device-agnostic int8 model (for both GPU and CPU), you can either save plain weight or transposed weight (but need to let device know with clear API definition).
2. It does not matter. We verified in AlexNet, GoogleNet, and VGGxx.

paddle/fluid/operators/fc_op.cc
luotao1: Fc op is a general op which don't need attribute `data_format`. `data_format` NCHW or NHWC is used for images, and Fc op could be used in any other models, such as NLP, DNN. These models don't have channel concepts.

hshen14: Thanks, we understood your point. We will align FC FP32 finally: if FC FP32 does not require the attribute, we will refine the code for sure.

paddle/fluid/operators/fc_op.cc
wzzju: In Paddle, the data layout of the fc/mul weight is input_node_num * output_node_num.  You said "the channel is dim-1 exactly",  what does it mean `output_node_num` ?

wzzju:
> > If weights transpose:
> > 
> > * Is the `cpu_int8_save_model` different with other `int8_save_model`? And you should do another weights transpose before `save_model`?
> > * How do you deal with `fc` + `fc`?
> 
> 1. It depends on how the API defines. E.g., if int8_save_model aims to provide device-agnostic int8 model (for both GPU and CPU), you can either save plain weight or transposed weight (but need to let device know with clear API definition).
> 2. It does not matter. We verified in AlexNet, GoogleNet, and VGGxx.

1. I think you should save the int8 weight as origin data layout, not the transposed data layout. When you load saved weights, do the transpose operation in the pass at first.
2. In Paddle, data layout described as follow: A = batch_size * input_node_num, W = input_node_num * output_node_num,
O = A * W = batch_size * output_node_num.
If weights transpose:
W should be transposed as WT = output_node_num *  input_node_num, A should be transposed as 
AT = input_node_num *  batch_size. Then:
OT = WT * AT = output_node_num * batch_size.
 In conclusion, is OT your output's data layout? If so, I think you should transpose OT to get O(data layout).

paddle/fluid/operators/fc_op.cc
hshen14: > In Paddle, the data layout of the fc/mul weight is input_node_num * output_node_num. You said "the channel is dim-1 exactly", what does it mean `output_node_num` ?

Yes, it means output.

> > > If weights transpose:
> > > 
> > > * Is the `cpu_int8_save_model` different with other `int8_save_model`? And you should do another weights transpose before `save_model`?
> > > * How do you deal with `fc` + `fc`?
> > 
> > 
> > 
> > 1. It depends on how the API defines. E.g., if int8_save_model aims to provide device-agnostic int8 model (for both GPU and CPU), you can either save plain weight or transposed weight (but need to let device know with clear API definition).
> > 2. It does not matter. We verified in AlexNet, GoogleNet, and VGGxx.
> 
> 1. I think you should save the int8 weight as origin data layout, not the transposed data layout. 
When you load saved weights, do the transpose operation in the pass at first.

Agree, that's a better choice.

> 2. In Paddle, data layout described as follow: A = batch_size * input_node_num, W = input_node_num * output_node_num,
>    O = A * W = batch_size * output_node_num.
>    If weights transpose:
>    W should be transposed as WT = output_node_num *  input_node_num, A should be transposed as
>    AT = input_node_num *  batch_size. Then:
>    OT = WT * AT = output_node_num * batch_size.
>    In conclusion, is OT your output's data layout? If so, I think you should transpose OT to get O(data layout).

Thanks, I understood your point here. My point is that MKL-DNN supports FC INT8 with channel-wise quantization scales per output channel. In order to enable channel-wise for better INT8 accuracy, frameworks or ops can decide whether to transpose based on the API definition of saving weight format.

luotao1: > In order to enable channel-wise for better INT8 accuracy

This better INT8 accuracy is on CV models or all other models? Do you have detail benchmark before and after enable channel-wise?

ghost:
Close without merge. Duplicated to PR #17641

PR #17106, opened 2019-04-25
Change buffered_reader stream to a non-default one

sneaxiy:


chengduoZH (dismissed):
Excellent

paddle/fluid/memory/allocation/retry_allocator_test.cc
```diff
@@ -41,7 +41,7 @@ TEST(RetryAllocator, RetryAllocator) {
 
   size_t thread_num = 32;
   size_t sleep_time = 40;
-  size_t extra_time = 10;
+  size_t extra_time = 15;
```
chengduoZH: Why do you change `extra_time` here?

sneaxiy: I do not know why `retry_allocator_test` fails randomly yet since CI is changed to run in parallel recently.

chengduoZH (approved):
Excellent

PR #17125, opened 2019-04-27
test=develop, add gradient sort backward strategy

JiabinYang:
This PR add a new backward strategy under dygraph mode, this strategy can sort the gradient with reverse order of forward and then add them together , it may avoid some diff while comparing static models with their dygraph version.

it can be used like this: 
    
    backward_strategy = fluid.dygraph.BackwardStrategy()
    backward_strategy.sort_sum_gradient = True
    loss.backward(backward_strategy)
      
and the default backward strategy is `backward_strategy.sort_sum_gradient = False`


paddle/fluid/pybind/pybind.cc
```diff
@@ -182,6 +182,18 @@ PYBIND11_MODULE(core, m) {
   m.def("print_mem_usage",
         []() { return memory::allocation::GPUMemMonitor.PrintMemUsage(); });
 
+  py::class_<imperative::detail::BackwardStrategy> backward_strategy(
+      m, "BackwardStrategy", R"DOC()DOC");
```
junjun315, line 186: empty doc?

junjun315 (approved):
GJ

PR #17135, opened 2019-04-28
Reformat fleet API

seiriosPlus:
like title

### DIFF
--- a/python/paddle/fluid/incubate/fleet/base/fleet_base.py
+++ b/python/paddle/fluid/incubate/fleet/base/fleet_base.py
@@ -15,18 +15,20 @@
 from __future__ import print_function
 
 import abc
-import sys
 
 from enum import Enum
 
 from paddle.fluid.optimizer import SGD
 
-from role_maker import RoleMakerBase, Role
+from role_maker import RoleMakerBase
 from role_maker import MPISymetricRoleMaker
 from role_maker import UserDefinedRoleMaker
 
 
 class Mode(Enum):
+    """
+    There are various mode for fleet, each of them is designed for different model.
+    """
     TRANSPILER = 1,
     PSLIB = 2,
     COLLECTIVE = 3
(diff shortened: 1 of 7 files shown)


python/paddle/fluid/incubate/fleet/base/fleet_base.py
```diff
@@ -15,18 +15,21 @@
 from __future__ import print_function
 
 import abc
-import sys
 
 from enum import Enum
 
 from paddle.fluid.optimizer import SGD
+from paddle.fluid.executor import Executor
```
guru4elephant, line 22: why do you import Executor here?

python/paddle/fluid/incubate/fleet/parameter_server/pslib/__init__.py
```diff
@@ -28,63 +28,56 @@ class PSLib(Fleet):
...
         init_worker(): will be called by user. When a user knows current process is_server(), he/she
                     should call init_worker() to initialize global information about worker and connect
                     worker with pserver. You should run startup program before init_worker.
 
         Args:
-            programs(Program|list): a Program or a list of Programs
-            scopes(Scope|list): a Scope or  a list of Scopes, default None.
+            executor(Executor): The executor to run for init server.
+            programs(Program|None): The program that need to run.
         """
```
xjqbest, line 49: what if a user uses  more than one scope , each for a specific main_program.  
in this case,  init_worker no longer supports this.

xjqbest: oh，these scopes are passed to fleet in minimize，ignore the above.

python/paddle/fluid/incubate/fleet/base/fleet_base.py
seiriosPlus: for value check, `isinstance of`

guru4elephant (approved):
LGTM. You need to update official document with replicable examples and API docs.

fuyinno4 (approved):

PR #17158, opened 2019-04-29
Remove unnecessary set_devices

chengduoZH:
Don't need to invoke `platform::SetDeviceId(gpu_id);` every time when calling `GetGPUBuddyAllocator `.

### DIFF
--- a/paddle/fluid/memory/allocation/legacy_allocator.cc
+++ b/paddle/fluid/memory/allocation/legacy_allocator.cc
@@ -170,7 +170,6 @@ BuddyAllocator *GetGPUBuddyAllocator(int gpu_id) {
     }
   });
 
-  platform::SetDeviceId(gpu_id);
   auto pos = std::distance(devices.begin(),
                            std::find(devices.begin(), devices.end(), gpu_id));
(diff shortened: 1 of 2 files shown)


paddle/fluid/memory/allocation/legacy_allocator.cc
```diff
@@ -168,9 +170,9 @@ BuddyAllocator *GetGPUBuddyAllocator(int gpu_id) {
                << ". Current 'FLAGS_reallocate_gpu_memory_in_mb' value is "
                << FLAGS_reallocate_gpu_memory_in_mb << "\n\n";
     }
+    platform::SetDeviceId(gpu_id);
   });
 
-  platform::SetDeviceId(gpu_id);
```
chengduoZH, line 173: Don't need to invoke `platform::SetDeviceId(gpu_id);` every time when calling `GetGPUBuddyAllocator `.

sneaxiy:
Maybe we can make cuda device id be a thread local variable?

sneaxiy (approved):
LGTM.

PR #17170, opened 2019-04-29
fix python3 run_time_error inlayer. ops. test=develop

fc-liu:
fix python3 run_time_error in layers.ops caused by locals(). The same as PR#17152


junjun315 (approved):
LGTM

luotao1:
Thanks very much for your contribution!

PR #17174, opened 2019-04-29
remove async executor python api to fix document

guru4elephant:
test=develop


wangguibao (approved):
OK

shanyi15 (approved):

chengduoZH:
@guru4elephant Can you explain in detail why do you remove those API?

guru4elephant:
> @guru4elephant Can you explain in detail why do you remove those API?

AsyncExecutor and Executor are merge in release 1.4, users can use Executor.train_from_dataset and Executor.infer_from_dataset for AsyncExecutor execution. Previously, the document of these APIs are not public in our official document, so users are not affected except the examples of text classification in Paddle/models. I will refine the text classification example of AsyncExecutor when this pr got merged.

chengduoZH:
@guru4elephant In this case, you need to specify in the release note that these APIs have been deleted and what alternative APIs are when the next version is released.

chengduoZH (approved):
**Considering the compatibility of versions, it is better not to add and delete APIs at will, which is very unfriendly to users.**

PR #17182, opened 2019-04-30
Fix mem leak when converting Tensor to numpy array

### LINKED ISSUE #17176: fetch_var 从GPU拷贝parameter到CPU的np.array，内存泄漏
fetch_var 从GPU拷贝parameter到CPU的np.array，内存泄漏。
示例代码如下：

    import paddle
    import paddle.fluid as fluid
    from paddle.fluid import layers
    import numpy as np

    def fetch_var(program, scope):
        for param in program.global_block().all_parameters():
            var = scope.find_var(param.name)
            t = var.get_tensor()
            array = np.array(t)             # TODO: This line will cause memory leak

    def net():
        x = layers.data(name='x', shape=[-1, 1000], dtype='float32')
        h = layers.fc(x, size=1000)
        y = layers.reduce_mean(h)
        return y

    def train():
        place = fluid.CUDAPlace(0)

        scope = fluid.Scope()
        program = fluid.Program()
        start_up = fluid.Program()
        with fluid.scope_guard(scope):
            with fluid.program_guard(program, start_up):
                with fluid.unique_name.guard():
                    out = net()
            exe = fluid.Executor(place)
            exe.run(start_up)

        print('start showing memory leak')
        while True:
            fetch_var(program, scope)
                                    
    if __name__ == "__main__":
        train()

How to run: 

    CUDA_VISIBLE_DEVICES=0 python fetch_var.py
系统信息：
- centos
- fluid 1.3
- CUDA9 CUDNN7

sneaxiy:
Fix #17176.


paddle/fluid/pybind/tensor_py.h
```diff
@@ -531,5 +428,90 @@ inline void PyCUDAPinnedTensorSetFromArray(
...
+  return false;
+}
+
+#define DECLARE_VALID_DTYPE_TO_PY_ARRAY(type)    \
+  template <>                                    \
+  constexpr bool IsValidDTypeToPyArray<type>() { \
+    return true;                                 \
+  }
+
+DECLARE_VALID_DTYPE_TO_PY_ARRAY(float);
```
zhhsplendid, line 445: 1. If we are supporting Numpy for this change. Had better to support basic Numpy types.
https://www.numpy.org/devdocs/user/basics.types.html
See the section of non-platform-dependent list: np.int8, np.int16, np.int32 ...

2. Suggest to sort these so you can see some types are missed. For example, uint16_t, uint32_t etc.

chengduoZH (dismissed):
LGTM

zhhsplendid (commented):
Great

paddle/fluid/pybind/tensor_py.h
sneaxiy: 1. I just follow the previous logic. Not all data types were supported previously. See [here](https://github.com/PaddlePaddle/Paddle/pull/17182/files#diff-e63f2a3aed0a7dc4f649dabb8272344dL134).

2. Done.

zhhsplendid: Sure, LGTM, also

zhhsplendid (dismissed):

chengduoZH (approved):

PR #17187, opened 2019-05-02
use two GPUs to run the exclusive test test=develop

wopeizl:
Originally some test cases would be run exclusively, which mean there can be only one test case run one time, now we force to use two GPUs to run the exclusively test cases, so that to speed up the process.


junjun315 (approved):
good job, LGTM

PR #17195, opened 2019-05-05
Fix tensor_py.h

sneaxiy:
Previously, `PADDLE_ENFORCE(IsValidDTypeToPyArray<T>, ...)` is wrong because it is checking whether the function pointer is nullptr, but not checking whether the returned value of the function is true. This PR fixes it.


chengduoZH (approved):
LGTM

PR #17204, opened 2019-05-05
Fix ParallelExecutor bug

chengduoZH:
The `trainer_id` and `num_trainers` of `build_strategy` should not be overwritten by the passed `trainer_id` and `num_trainers` of `ParallelExecutor`.

### DIFF
--- a/python/paddle/fluid/parallel_executor.py
+++ b/python/paddle/fluid/parallel_executor.py
@@ -96,19 +96,31 @@ class ParallelExecutor(object):
 
         if build_strategy is None:
             build_strategy = BuildStrategy()
-        build_strategy.num_trainers = num_trainers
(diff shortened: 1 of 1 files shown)


Xreki (approved):
LGTM

python/paddle/fluid/parallel_executor.py
```diff
@@ -96,19 +96,31 @@ def __init__(self,
 
         if build_strategy is None:
             build_strategy = BuildStrategy()
-        build_strategy.num_trainers = num_trainers
-        build_strategy.trainer_id = trainer_id
```
chengduoZH, line 100: The `trainer_id` and `num_trainers` of `build_strategy` should not be overwritten by the passed `trainer_id` and `num_trainers` of `ParallelExecutor`.

PR #17205, opened 2019-05-05
Add use_cuda to inplace pass

sneaxiy:
This PR adds `bool use_cuda` to `InplaceOpInference` for in-placing `Softmax` and `Logits` in softmax_with_cross_entropy op when using CUDA. 

I have tested that `cudnnSoftmaxForward` can be performed in-place, and added a unittest to guarantee it.

Also enhance inplace pass to avoid wrong in place when there are duplicate inputs and outputs with same var names.

### DIFF
--- a/paddle/fluid/framework/details/inplace_op_pass.cc
+++ b/paddle/fluid/framework/details/inplace_op_pass.cc
@@ -111,9 +111,9 @@ class InplacePass : public ir::Pass {
   // Check whether all `ops` is the preceding ops of `op`
   bool CheckOpDeps(ir::Node *op, const std::vector<ir::Node *> &ops) const;
 
-  // Find node whose name is equal to the given name
-  static ir::Node *FindNodeByName(const std::string &name,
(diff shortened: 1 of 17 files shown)


chengduoZH (approved):

paddle/fluid/framework/details/inplace_op_pass.cc
```diff
@@ -111,9 +111,9 @@ class InplacePass : public ir::Pass {
   // Check whether all `ops` is the preceding ops of `op`
   bool CheckOpDeps(ir::Node *op, const std::vector<ir::Node *> &ops) const;
 
-  // Find node whose name is equal to the given name
-  static ir::Node *FindNodeByName(const std::string &name,
-                                  const std::vector<ir::Node *> &nodes);
+  // Find nodes whose name are equal to the given name
```
zhhsplendid, line 114: Just grammar: "whose name are equal" is incorrect
You can say "whose names are equal to" or "whose name is equal to"
or simply "whose name equals the given name"

PR #17215, opened 2019-05-05
Quant output scale

wzzju:
`MovingAverageAbsMaxScale` operator is only used for calculating the quantization scale.
And it will not quantize the input tensor. It is necessary for  TensorRT.


NHZlX (dismissed):
LGTM

NHZlX (approved):
LGTM

PR #17225, opened 2019-05-06
Enhance inplace/mem-opt pass and enhance softmax_with_cross_entropy op inplace

sneaxiy:
This PR works for:

- Add more skip variables to mem opt/inplace pass.
- Enhance inplace pass to check whether in_arg/out_arg occur in inputs/outputs of `OpDesc` multiple times. If so, inplace cannot be performed.
- Enhance softmax_with_cross_entropy to support in-place in both CPU/GPU kernels, and in both soft_label=False and soft_label=True.

### DIFF
--- a/paddle/fluid/framework/details/inplace_op_pass.cc
+++ b/paddle/fluid/framework/details/inplace_op_pass.cc
@@ -111,9 +111,9 @@ class InplacePass : public ir::Pass {
   // Check whether all `ops` is the preceding ops of `op`
   bool CheckOpDeps(ir::Node *op, const std::vector<ir::Node *> &ops) const;
 
-  // Find node whose name is equal to the given name
-  static ir::Node *FindNodeByName(const std::string &name,
-                                  const std::vector<ir::Node *> &nodes);
+  // Find nodes whose name are equal to the given name
+  static std::unordered_set<ir::Node *> FindNodesByName(
+      const std::string &name, const std::vector<ir::Node *> &nodes);
 
   // Get all versions vars named var_name
   std::vector<ir::Node *> *AllVersionVars(const std::string &var_name) const;
@@ -290,17 +290,15 @@ void InplacePass::RenameInOut(ir::Node *op, ir::Node *in_var,
   op->Op()->Flush();
 }
 
-ir::Node *InplacePass::FindNodeByName(const std::string &name,
-                                      const std::vector<ir::Node *> &nodes) {
-  ir::Node *found_node = nullptr;
+std::unordered_set<ir::Node *> InplacePass::FindNodesByName(
+    const std::string &name, const std::vector<ir::Node *> &nodes) {
+  std::unordered_set<ir::Node *> ret;
   for (auto *node : nodes) {
     if (node->Name() == name) {
-      PADDLE_ENFORCE(found_node == nullptr, "Find duplicate input nodes %s",
-                     name);
-      found_node = node;
+      ret.insert(node);
     }
   }
-  return found_node;
+  return ret;
 }
 
 void InplacePass::ApplyImpl(ir::Graph *graph) const {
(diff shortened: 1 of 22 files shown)


zhhsplendid (dismissed):

paddle/fluid/framework/details/record_skip_memory_opt_vars_pass.cc
```diff
@@ -30,26 +33,102 @@ class RecordSkipMemoryOptVarsPass : public ir::Pass {
...
+      // issue yet. Currently, we should skip these variables when
+      // memory optimization is enabled.
+      auto op_type = op_desc->Type();
+      if (op_type == "while_grad") {
+        UpdateSkipVarSet(skip_vars, {ToGradVarName(op_desc->Input("X"))});
+      } else if (op_type == "conditional_block_grad") {
+        UpdateSkipVarSet(skip_vars, {ToGradVarName(op_desc->Input("Input")),
+                                     ToGradVarName(op_desc->Input("Cond"))});
+      } else if (op_type == "recurrent" || op_type == "recurrent_grad") {
+        auto& ex_states =
```
liupluswei, line 143: better add some comments here to explain the skip reason

sneaxiy: Done.

paddle/fluid/framework/details/inplace_op_pass.cc
```diff
@@ -301,6 +275,21 @@ std::unordered_set<ir::Node *> InplacePass::FindNodesByName(
...
+  in_args->clear();
+  out_args->clear();
+
+  for (auto &in_name : op_desc->InputArgumentNames()) {
+    in_args->insert(in_name);
+  }
+
+  for (auto &out_name : op_desc->OutputArgumentNames()) {
+    out_args->insert(out_name);
+  }
```
chengduoZH: Why the outputs of some op may have the same name?

paddle/fluid/framework/details/inplace_op_pass.cc
```diff
@@ -387,6 +381,20 @@ void InplacePass::ApplyImpl(ir::Graph *graph) const {
...
+                << " occurs " << in_arg_occur_times << " times in input of op "
+                << op_type;
+        continue;
+      }
+
+      size_t out_arg_occur_times = all_out_args.count(out_arg);
+      PADDLE_ENFORCE_EQ(
+          out_arg_occur_times, 1,
+          "Wrong graph: Output(%s)=%s occurs in other outputs of op %s",
+          out_param, out_arg, op_type);
```
chengduoZH: I am thinking that maybe it is not appropriate that adding the check here, it should be placed in program->graph.

paddle/fluid/framework/details/record_skip_memory_opt_vars_pass.cc
```diff
@@ -30,26 +33,129 @@ class RecordSkipMemoryOptVarsPass : public ir::Pass {
...
+    ret.reserve(names.size());
+    for (auto& name : names) {
+      if (name != framework::kEmptyVarName) {
+        ret.emplace_back(framework::GradVarName(name));
+      }
+    }
+    return ret;
+  }
+
+  static void InsertSkipMemOptOpInOutToSkipVarSet(
```
chengduoZH, line 92: Why do you use `static` here?

sneaxiy: Because it is really a static method.

paddle/fluid/framework/details/inplace_op_pass.cc
sneaxiy: Just to check wrong case. I think you are right. We should add these check when converting program to graph.

paddle/fluid/framework/details/inplace_op_pass.cc
sneaxiy: I agree.

paddle/fluid/framework/details/inplace_op_pass.cc
sneaxiy: Done.

paddle/fluid/framework/details/inplace_op_pass.cc
sneaxiy: Done.

chengduoZH (approved):

PR #17227, opened 2019-05-06
disable the test_distillation_strategy temporarily

wopeizl:
1. it always failed on a machine with 4 gpus only, need to figure out the root cause and add it back later
http://ci.paddlepaddle.org/viewLog.html?buildId=92473&buildTypeId=Paddle_PrCiPython35&tab=buildLog
http://ci.paddlepaddle.org/viewLog.html?buildId=92622&buildTypeId=Paddle_PrCiPython35&tab=buildLog


seiriosPlus (approved):
LGTM

PR #17228, opened 2019-05-06
Code Clean: Move all pass to paddle::framework::ir

chengduoZH:
Move all the pass to paddle/framework/ir. It will make the code clear.


sneaxiy (approved):
Excellent.

PR #17259, opened 2019-05-07
Add scale pass for calculating the output scales

wzzju:
Add scale pass for calculating the output scales. These output scales may be used by tensorRT or some other inference engines.


NHZlX (approved):
LGTM

PR #17261, opened 2019-05-07
Async exe support communicator

jacquesqiao:


paddle/fluid/operators/distributed/communicator.cc
```diff
@@ -15,6 +15,7 @@ limitations under the License. */
 #include "paddle/fluid/operators/distributed/communicator.h"
```
guru4elephant, line 15: Can you move communicator into paddle/fluid/framework/fleet/? Since communicator is a singleton instance that can be called from anywhere, it is not reasonable to put communicator under paddle/fluid/operators.

paddle/fluid/operators/distributed/communicator.cc
```diff
@@ -83,6 +83,47 @@ Communicator::Communicator(const RpcCtxMap &send_varname_to_ctx,
   recv_threadpool_.reset(new ::ThreadPool(FLAGS_communicator_thread_pool_size));
 }
 
+Communicator::Communicator(const paddle::framework::ProgramDesc &program,
+                           Scope *param_scope) {
```
guru4elephant: why do you have to use param_scope in constructor?

paddle/fluid/operators/distributed/communicator.cc
```diff
@@ -83,6 +83,47 @@ Communicator::Communicator(const RpcCtxMap &send_varname_to_ctx,
...
+  using RpcCtxMap = operators::distributed::RpcCtxMap;
+  VLOG(3) << "ProcessGraph";
+  RpcCtxMap send_varname_to_ctx;
+  RpcCtxMap recv_varname_to_ctx;
+  for (auto *op : program.Block(0).AllOps()) {
+    VLOG(3) << "node name " << op->Type();
+    if (op->Type() == "send") {
```
guru4elephant: Can I use communicator beyond send_op and recv_op?

jacquesqiao:
move to https://github.com/PaddlePaddle/Paddle/pull/17386

paddle/fluid/operators/distributed/communicator.cc
jacquesqiao: ok can do this in next pr

paddle/fluid/operators/distributed/communicator.cc
jacquesqiao: currently it can only be used inside send and recv op.

PR #17266, opened 2019-05-07
improved unit test output

wojtuss:
added printing data type to differentiate int8 and fp32 latency results

test=develop


luotao1 (approved):
LGTM！ Thanks very much!

PR #17273, opened 2019-05-08
Fix softmax_with_cross_entropy op

sneaxiy:
When feature_size = 1, softmax_with_cross_entropy op does not add `PADDLE_ENFORCE` to CUDA system calls. This PR fixes it.

### DIFF
--- a/paddle/fluid/operators/softmax_with_cross_entropy_op.cu
+++ b/paddle/fluid/operators/softmax_with_cross_entropy_op.cu
@@ -10,6 +10,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 #include <cub/cub.cuh>
 #include "paddle/fluid/operators/math/cross_entropy.h"
+#include "paddle/fluid/operators/math/math_function.h"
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/softmax_with_cross_entropy_op.cu
```diff
@@ -359,7 +359,8 @@ static void HardLabelSoftmaxWithCrossEntropy(
                                                 kMaxBlockDim,
                                             kMaxBlockDim, 0, stream>>>(
           softmax_data, grid_dim);
-      cudaMemsetAsync(loss_data, 0, grid_dim * sizeof(T), stream);
+      PADDLE_ENFORCE(
+          cudaMemsetAsync(loss_data, 0, grid_dim * sizeof(T), stream));
```
chengduoZH: Why not use SetConstant here?

sneaxiy: Done.

chengduoZH (approved):

PR #17277, opened 2019-05-08
delete mul_lstm_fuse_pass

fc500110:
Sometimes `mul_lstm_fuse_pass` result in the precision difference between CPU and GPU computation.

Disable it temporarily.  

TODO @fc500110 @lijianshe02 repair it.


Superjomn (dismissed):
LGTM

PR #17279, opened 2019-05-08
Fix two graphics memory problem When doing inference.

NHZlX:
1. infernce multi card occupy
We specific gpu id using `config->EnableUseGpu(0 /*gpu_id*/, 10);`, but multiple cards will be occupied when during inference.

2. the facebox model  occupy too much graphic memory when during inference.

The reason is that the default value of conv attr "workspace_size_MB" is 4G, we modify this attr's value to 64M, which is applicable for most inference tasks..


Superjomn (approved):
LGTM

PR #17291, opened 2019-05-08
Use sync copy in fetch_op_handle

chengduoZH:
It seems that, in the origin implementation, the copy action is not synced.


Xreki (approved):
LGTM

PR #17295, opened 2019-05-08
Add fill_zeros_like op to ngraph engine

baojun-nervana:
Added fill_zeros_like op to ngraph engine. vgg model uses it


tensor-tang (approved):
LGTM

PR #17296, opened 2019-05-08
Enable gather operator for a NGraph

mozga-intel:
Enable the gather operator for the nGraph bridge.  BERT model uses it.


tensor-tang (approved):
LGTM

PR #17297, opened 2019-05-08
Add DropLocalExeScopes in ParallelExecutor

chengduoZH:
Add DropLocalExeScopes in ParallelExecutor.

During the execution of the Program, the generate intermediate results are placed in local execution scope, in some model the creation and deletion of those intermediate results are time-consuming. To resolve that problem, ParallelExecutor provides an option in ExecutionStrategy, i.g. num_iteration_per_drop_scope, this option indicates how many iterations to run before dropping the local execution scope. But in some situation, each iteration generates different intermediate results, it will lead to the result that the memory which is needed by local execution scope gradually increase. And if you want to run another program at this time, there may be insufficient storage, At this point, you should drop the local execution scope of other Programs.

FireShot Capture 017 - PaddlePaddle - Docs - 10 255 125 37


paddle/fluid/framework/details/op_handle_base.h
```diff
@@ -27,7 +27,7 @@ namespace paddle {
 namespace framework {
 namespace details {
 
-constexpr char kLocalExecScopeName[] = "@LOCAL_SCOPE@";
+constexpr char kLocalExecScopeName[] = "@LOCAL_EXE_SCOPE@";
```
sneaxiy, line 30: Why change this?

paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.h
```diff
@@ -47,14 +47,9 @@ class ScopeBufferedSSAGraphExecutor : public SSAGraphExecutor {
...
- private:
-  inline void WaitComputationalStreams() {
-    // Wait All computational streams
-    for (auto p : places_) {
-      platform::DeviceContextPool::Instance().Get(p)->Wait();
-    }
-  }
+  void DropLocalExeScopes();
 
+ private:
```
sneaxiy: Remove `private`?

paddle/fluid/framework/details/op_handle_base.h
chengduoZH: To make the name more explicitly. The local scope and local execution scope are different.

sneaxiy (approved):
LGTM.

XiaoguangHu01 (approved):

shanyi15 (approved):
LG comments

PR #17320, opened 2019-05-09
Adding dropout and dropout_grad to ngraph engine

baojun-nervana:
Added dropout and dropout_grad op to ngraph engine. VGG and BERT models use them. Mask uses uint8, so the data type was added to ngraph engine.


paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -61,6 +61,7 @@ static std::map<framework::proto::VarType::Type, ngraph::element::Type>
         {framework::proto::VarType::FP64, ngraph::element::f64},
         {framework::proto::VarType::INT32, ngraph::element::i32},
         {framework::proto::VarType::INT64, ngraph::element::i64},
+        {framework::proto::VarType::UINT8, ngraph::element::u8},
```
tensor-tang, line 64: why add u8 in this dropout implementation?

tensor-tang (approved):
LGTM

PR #17394, opened 2019-05-14
Fix/Fix memory leak in dygraph

JiabinYang:
Fix memory leak problem in dygraph mode and using `unique_ptr` to hold `Variable` to make sure it can be freed automatically  when `Varbase` is released by python gc


paddle/fluid/imperative/layer.cc
```diff
@@ -132,8 +132,8 @@ class Autograd {
       return;
     }
     VLOG(3) << "start autograd";
-    bck_map = new BackwardSumMap();
-    grad_ref = new GradientRef();
+    bck_map = BackwardSumMap();
+    grad_ref = GradientRef();
```
sneaxiy: These two lines are unnecessary. BTW, it seems that it is not necessary that`bck_map` and `grad_ref` are data member of `Autograd`. Just make them temporary variable inside the function.

paddle/fluid/imperative/layer.cc
```diff
@@ -185,12 +185,12 @@ class Autograd {
         for (const auto& map : candidate->grad_output_vars_) {
           for (const auto& it : map) {
             for (const auto& vb : it.second) {
-              if (grad_ref->find(vb) == grad_ref->end()) {
-                grad_ref->insert(std::make_pair(vb, 1));
+              if (grad_ref.find(vb) == grad_ref.end()) {
+                grad_ref.insert(std::make_pair(vb, 1));
```
sneaxiy: C++ guarantees zero-initialization of primitive types. See [here](https://en.cppreference.com/w/cpp/language/zero_initialization). You can just write `++grad_ref[vb];`

paddle/fluid/imperative/layer.cc
```diff
@@ -213,8 +213,8 @@ class Autograd {
     return ret;
   }
 
-  BackwardSumMap* bck_map;
-  GradientRef* grad_ref;
+  BackwardSumMap bck_map;
+  GradientRef grad_ref;
```
sneaxiy: No need to be data members.

sneaxiy (approved):
Excellent.

PR #17396, opened 2019-05-14
Fix dygraph memory leak

sneaxiy:
In dygraph train mode, vars created in forward op without backward would not be released after each batch ends, which causes memory leak. Also, without calling `dev_ctx->Wait()` in GPU, memory allocated by `TemporaryAllocator` would not be released. This PR fixes these two bugs.

### DIFF
--- a/paddle/fluid/imperative/layer.cc
+++ b/paddle/fluid/imperative/layer.cc
@@ -118,7 +118,7 @@ void AddGradBySort(BackwardSumMap* bck_map, VarBase* target) {
     VLOG(2) << "added grad: " << var_pair.second->Name()
             << " trace id is: " << var_pair.first;
     AddTo(grad_to_add, origin_grad, current.first);
-    delete grad_to_add;
+    delete var_pair.second;
     var_pair.second = nullptr;
   }
 }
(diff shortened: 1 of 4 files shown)


chengduoZH (approved):
LGTM

paddle/fluid/imperative/layer.cc
```diff
@@ -230,16 +230,14 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
-                              tensor);
-
-    dev_ctx->Wait();
-  } else {
-    framework::TensorCopy(var_->Get<framework::LoDTensor>(), dst_place, tensor);
+    platform::DeviceContextPool::Instance().Get(dst_place)->Wait();
+    auto src_place = src_tensor.place();
+    if (!(src_place == dst_place)) {
+      platform::DeviceContextPool::Instance().Get(src_place)->Wait();
+    }
```
chengduoZH, line 240: Just need to wait for GPU here.

sneaxiy: Just make it more common.

PR #17405, opened 2019-05-14
[Speed] Refine the Executor when the num_thread=1

chengduoZH:
Refine the Executor when the num_thread=1.
For the small model, this PR maybe make the speed faster when num_thread is 1.

---

If the num_thread is 1, the executor will run the operators of computation graph by analysis the operators' dependency during the first iteration and record the operators' execution order in `traced_ops`, and then in subsequence iterations, run the operators of `traced_ops` one by one. This strategy is faster than before, and in the small model, the speed increase is more obvious.

I attempt to use run the topo sorted operators directly, but it is a bit slower than the above strategy.
This is to say, the operators' execution order will affect the program speed in some case. If we have time, we should analyze and build the operators' execution order to make the program run faster.

### DIFF
--- a/paddle/fluid/framework/details/fast_threaded_ssa_graph_executor.cc
+++ b/paddle/fluid/framework/details/fast_threaded_ssa_graph_executor.cc
@@ -107,6 +107,7 @@ FeedFetchList FastThreadedSSAGraphExecutor::Run(
   for (auto op : ready_fetch_ops) {
     RunOpAsync(op_deps.get(), op, complete_q);
   }
+
   while (num_complete != op_deps->size()) {
     size_t num_comp = complete_q->Pop();
     if (num_comp == -1UL) {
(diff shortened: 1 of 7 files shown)


Xreki:
I test this PR in PaddingRNN models which set `exec_strategy.num_threads=1`:
- PaddingRNN small model (static): 41.8 s -> 35.8 s
- PaddingRNN large model (static): 75.6 s -> 70.3 s

paddle/fluid/framework/details/fast_threaded_ssa_graph_executor.cc
```diff
@@ -94,55 +150,22 @@ FeedFetchList FastThreadedSSAGraphExecutor::Run(
...
-      op->Run(strategy_.use_cuda_);
+      RecordOps(op);
     }
     ++(*complete);
     return true;
-  } catch (...) {
-    exception_.Catch(std::current_exception());
+  } else {
     --remaining_;
     complete_q->Push(-1UL);
```
chengduoZH, line 173: Maybe `complete_q->Push` would throw exception?

sneaxiy (approved):
LGTM.

PR #17406, opened 2019-05-14
Enable stack operator for a NG

mozga-intel:
Enable the stack operator for the nGraph bridge. BERT model uses it.


tensor-tang (approved):
LGTM

PR #17436, opened 2019-05-15
Enable abs operator for a ngraph

mozga-intel:
Enable the abs operator for the nGraph bridge. BERT model uses it.


luotao1 (approved):
LGTM!

PR #17437, opened 2019-05-15
Enable assign operator for a ngraph, test=develop

mozga-intel:
Enable the assign operator for the nGraph bridge. BERT model uses it.

Additionally, the corss_entropy test was updated, due to that CI's test fails.

### DIFF
--- a/paddle/fluid/operators/ngraph/ops/assign_op.h
+++ b/paddle/fluid/operators/ngraph/ops/assign_op.h
@@ -1,0 +1,43 @@
+/*Copyright (c) 2019 PaddlePaddle Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


tensor-tang (commented):
> Additionally, the corss_entropy test was updated, due to that CI's test fails.

Could u give some more details about this ?

mozga-intel:
@tensor-tang The assign operator needs to support different shapes so any changes of xe operator are needed. The previous version of xe operator did not include these changes.

tensor-tang:
> xe operator are needed

did not understand this

> due to that CI's test fails.

Actually I mean what is the test fails you said?

mozga-intel:
@tensor-tang 
1) xe operator are needed - it means that I must have updated the cross_entoropy operator
2) Actually, I mean what is the test fails you said? - cross_entropy_ngraph

tensor-tang (approved):
LGTM

PR #17443, opened 2019-05-16
support sparse table get shard_num from TableParameter

xjqbest:
support sparse table get shard_num from TableParameter when using fleet，
TableParameter is in fleet_desc.prototxt


seiriosPlus (approved):
LGTM

PR #17448, opened 2019-05-16
add Concat quantization

sfraczek:
in files `graph_pattern_detector.cc, graph_pattern_detector.h, cpu_quantize_pass.cc, cpu_quantize_pass.h, mkldnn_quantizer.cc, mkldnn_quantizer.cc`:

- added Concat quantization code 

in file `mkldnn_quantizer.cc`:
-  handling of multiple tensors wired to single input
- extended list of ops that do not modify the sign of values in tensor,
- added setting type to unsigned after regular ReLU op

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.h"
+#include <limits>
 #include <utility>
 #include <vector>
 #include "paddle/fluid/framework/eigen.h"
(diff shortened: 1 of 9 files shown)


paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -50,34 +50,39 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
...
+
+            // force unsigned type if already know it
+            bool is_unsigned = false;
+            if (is_output && op->Type() == "conv2d") {
+              // output of conv2d with relu must be unsigned
+              is_unsigned = op->HasAttr("fuse_relu") &&
+                            boost::get<bool>(op->GetAttr("fuse_relu"));
+            } else if (is_output && op->Type() == "relu") {
+              is_unsigned = true;
+            } else if (is_output &&
```
wojtuss, line 71: We need to check whether op's input has `is_unsigned == true`

sfraczek: Lines 75-80 are supposed to check that. For all values linked to `op->Input("X")` we check all `scales_[input_var_name].first`

wojtuss: right

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -216,6 +264,46 @@ void CPUQuantizePass::QuantizePool(Graph* graph) const {
...
+    if (!concat_op_desc->HasAttr("use_quantizer") ||
+        !boost::get<bool>(concat_op_desc->GetAttr("use_quantizer")))
+      return;
+
+    GET_IR_NODE_FROM_SUBGRAPH(concat_out, concat_out, concat_pattern);
+
+    // get scales calculated after warmup, they scale variables to MAX=1.0
+    auto scales = Get<VarQuantScale>("quant_var_scales");
+
+    bool are_all_inputs_unsigned = scales[concat_out->Name()].first;
```
wojtuss, line 292: It would be good to have a comment here, explaining that checking signs of all inputs was done during the scale calculation phase and based on that check the `is_unsigned` value for the concat's output was set.

sfraczek:
I have to add use_quantizer flag to concat operator which because its missing in this PR

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
sfraczek: We can do that later

wojtuss (dismissed):

wojtuss (approved):

luotao1:
Which model uses this quantization, do you have some benchmark before and after this quantization?

sfraczek:
Googlenet and mobilenet-ssd benefit from this (but we don't have test for mobilenet-ssd yet).
GoogleNet is about 1.67x faster with concat quantized vs before (on my i9 for development with batch size 50 and 1000 iterations).

PR #17453, opened 2019-05-16
Update ngraph engine to parse serialized graph

baojun-nervana:
This PR is to update ngraph engine to be able to parse a serialized graph. We are adding a pass to enable ngraph, and the subgraph will be carried to ngraph engine through a serialized graph.


tensor-tang (approved):
LGTM

PR #17475, opened 2019-05-18
fix the random compilation failure on windows test=develop

wopeizl:
build may fail on windows since the dependency miss, add the dependency back


gongweibao (approved):
LGTM

PR #17489, opened 2019-05-20
Add LAMB Optimizer support

kuke:
Add [LAMB Optimizer](https://arxiv.org/abs/1904.00962) for large-batch training of BERT. 

We can see that Adam is chosen to be the base Optimizer, and they share the same way in moments' updating :

image


paddle/fluid/operators/optimizers/lamb_op.cc
```diff
@@ -0,0 +1,101 @@
...
+where $m$ is the 1st moment, and $v$ the 2nd moment, $\eta$ the 
+learning rate, $\lambda$ the weight decay rate.
+)DOC");
+  }
+};
+
+class LambOpVarTypeInference : public framework::VarTypeInference {
+ public:
+  void operator()(framework::InferVarTypeContext* ctx) const override {}
+};
```
chengduoZH: Why the `LambOpVarTypeInference` is empty?

paddle/fluid/operators/optimizers/lamb_op.cc
```diff
@@ -0,0 +1,101 @@
...
+}  // namespace operators
+}  // namespace paddle
+
+namespace ops = paddle::operators;
+REGISTER_OPERATOR(lamb, ops::AdamOp, ops::LambOpMaker,
+                  paddle::framework::EmptyGradOpMaker,
+                  ops::LambOpVarTypeInference);
```
chengduoZH: You could use `REGISTER_OP_WITHOUT_GRADIENT` here, please refer to adam_op.cc.

paddle/fluid/operators/optimizers/lamb_op.h
```diff
@@ -0,0 +1,316 @@
...
+    auto& param_out =
+        Ref(ctx.Output<LoDTensor>("ParamOut"), "Must set ParamOut.");
+    auto& mom1_out =
+        Ref(ctx.Output<LoDTensor>("Moment1Out"), "Must set Moment1Out.");
+    auto& mom2_out =
+        Ref(ctx.Output<LoDTensor>("Moment2Out"), "Must set Moment1Out.");
+
+    platform::ForRange<DeviceContext> for_range(
+        static_cast<const DeviceContext&>(ctx.device_context()), param.numel());
+    framework::Tensor trust_ratio_div;
```
chengduoZH: There is a potential bug because the GPU execution is async, this is to say that when the calling of `LambOpKernel::Compuate` returns, the computation of the GPU side may be not finished. But `trust_ratio_div` is placed in stack memory, and the memory is to be freed when the calling of `LambOpKernel::Compuate` returns.

chengduoZH: I found that there are many files use `Tensor tmp`, but the program doesn't have abnormal results, this is because we use only use one CUDA stream, and all the tasks are placed in this stream.

paddle/fluid/operators/optimizers/lamb_op.cc
kuke: Removed

paddle/fluid/operators/optimizers/lamb_op.cc
kuke: Done

paddle/fluid/operators/optimizers/lamb_op.h
kuke: Changed these declarations.

chengduoZH (dismissed):
LGTM

kuke:
image

chengduoZH (approved):

shanyi15 (approved):

PR #17491, opened 2019-05-20
remove two memory optimization related flags

liupluswei:
remove FLAG_enable_subgraph_optimize and FLAG_memory_optimize_debug
these two flags are added by the original author for debug purpose, but they are rarely used. Delete them to reduce the total number of exposed flags


shanyi15 (approved):

sneaxiy (approved):
LGTM.

PR #17500, opened 2019-05-20
remove unused SERIAL option

luotao1:
update #17054
remove all SERIAL since the parallel test would not use this option.


wopeizl (approved):
LGTM

PR #17522, opened 2019-05-21
refine softmax fwd on CPU

tensor-tang:
improve 6X on v2620 v3

Before
image

After
image


luotao1:
> If reason is that input is to small to vsEXP then  how about on moving vsEXP out of batch loop. vsEXP takes most of time  perhaps we could first in a loop (for all batches) compute max and sub and then vsEXP  for all data followed by loop executing sums and scals?

@jczaja Do you mean update this PR?

paddle/fluid/operators/math/cpu_vec.h
```diff
@@ -128,6 +135,47 @@ inline void vec_scal<float, platform::avx512f>(const int n, const float a,
   vec_scal<float, platform::avx2>(n, a, x, y);
 }
 
+template <typename T, platform::cpu_isa_t isa = platform::isa_any>
+inline void vec_sum(const size_t n, const T* x, T* s) {
+  s[0] = x[0];
```
jczaja, line 140: Just small suggestion. It could be worth to add OpenMP SIMD vectorization (the way it is described in our article). Reasoning is that for sum reduction it gives a boost and works for SSE4 so code would benefit from it on older platform. Details in our article (https://arxiv.org/pdf/1904.12380.pdf)  Figure 6 (lines 27 - 31) in section 2.4.2 . That would work starting from GCC 4.9.2

paddle/fluid/operators/math/softmax_impl.h
```diff
@@ -38,16 +40,37 @@ template <typename DeviceContext, typename T, bool is_test, typename Enable>
...
-
+  auto in_dims = X->dims();
   const int kBatchDim = 0;
   const int kClassDim = 1;
 
-  const int batch_size = logits.dimension(kBatchDim);
-  const int num_classes = logits.dimension(kClassDim);
+  const int num_classes = in_dims[kClassDim];
+  const int batch_size = in_dims[kBatchDim];
   const int num_remain = num_classes / axis_dim;
```
jczaja: axis_dim can be (to my understanding): [0, rank-1] . So it may happen that  there is exception here when dividing by zero . If my understanding is wrong then please explain?

paddle/fluid/operators/math/softmax_impl.h
```diff
@@ -38,16 +40,37 @@ template <typename DeviceContext, typename T, bool is_test, typename Enable>
...
 
+  if (num_remain == 1 && platform::MayIUse(platform::avx)) {
```
jczaja: Could you please why num_remain has to be equal 1 ? I understand that efficient softmax code can be used when axis_dim == rank -1  eg. we noramlize after final dimension.  Why not has condition: axis_dim + 1 = in.dims.size()

paddle/fluid/operators/math/softmax_impl.h
```diff
@@ -38,16 +40,37 @@ template <typename DeviceContext, typename T, bool is_test, typename Enable>
...
+
+      T sum = 0;
+      vec_sum<T, platform::avx>(num_classes, out_data, &sum);
+      sum = static_cast<T>(1) / sum;
+      vec_scal<T, platform::avx>(num_classes, sum, out_data, out_data);
+
+      in_data += num_classes;
+      out_data += num_classes;
+    }
+    return;
```
jczaja: What will happen when we have situation that we exectue softmax normalization after first axe or no AVX support. It seems there will no be any computation? Is this correct

jczaja (commented):
@luotao1 That was just suggestion on how to implement softmax op to deal with small inputs (while batch are bigger). It is not directly related to this PR

paddle/fluid/operators/math/cpu_vec.h
tensor-tang: goog point by `# pragma omp simd redution  (+ : sum)`

But will this use multi-threads? I am afraid we could only use single thread.

paddle/fluid/operators/math/softmax_impl.h
tensor-tang: https://github.com/PaddlePaddle/Paddle/blob/dfdcd91869595c7f068767e15f05f325d8a1de9c/paddle/fluid/operators/softmax_op.h#L56

It's from here, so could no be zero.

paddle/fluid/operators/math/softmax_impl.h
tensor-tang: It's refer to [here](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/jit/refer/refer.h#L444)

paddle/fluid/operators/math/softmax_impl.h
tensor-tang: This only improve on avx platform and only axis=1, if not, use original implementations.

paddle/fluid/operators/math/cpu_vec.h
jczaja: No. this is single threaded eg. omp simd does perform vectorization

paddle/fluid/operators/math/softmax_impl.h
jczaja: Ok, makes sense

paddle/fluid/operators/math/softmax_impl.h
jczaja: ok

jczaja (dismissed):
LGTM

paddle/fluid/operators/math/softmax_impl.h
jczaja: ok

paddle/fluid/operators/math/cpu_vec.h
tensor-tang: OK thanks.

jczaja:
@tensor-tang Just one more question: This code is to be run for inference only eg. it is enable when PaddlePaddle for inference is build and no ValueClip is there that is needed in softmax fwd training. But performance results are shown from training which should not contain this optimization ?

tensor-tang:
This is used for training actually.

tensor-tang:
http://ci.paddlepaddle.org/viewLog.html?buildId=101697&buildTypeId=Paddle_PrCi&tab=buildLog&branch_Paddle=pull%2F16684 CI failed on test_parallel_executor_seresnext

luotao1 (approved):
LGTM

PR #17524, opened 2019-05-21
TRT: Support set dynamic range in int8 mode.

NHZlX:
Align fluid int8 training and trt int8 inference .

We first do this on resnet50 model and alignment is made on accuracy and speed.

**Env: p4 card**  
**Dataset: imagenet**

|resnet50 |speed|accuracy|
|:--:|:--:|:--:|
|fp32|5.28776ms|76.35%|
|int8 |2.22264ms|76.26%|

image


Superjomn (approved):
LGTM

PR #17529, opened 2019-05-21
Adding AddQuantDequantPass for TensorRT int8

wzzju:
Add quant_dequant before some ops, such as the `elementwise_add` op. This is required by TensorRT.


NHZlX (approved):
LGTM

PR #17536, opened 2019-05-21
polish_executor_and_add_ctx_cache

guru4elephant:
test=develop

### DIFF
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -63,11 +63,7 @@ ExecutorPrepareContext::~ExecutorPrepareContext() {
   VLOG(5) << "destroy ExecutorPrepareContext";
 }
 
-Executor::Executor(const platform::Place& place) : place_(place) {
-  cur_step_ = 0;
-  open_ctx_cache_ = false;
-  ctx_is_cached_ = false;
-}
+Executor::Executor(const platform::Place& place) : place_(place) {}
 
 void Executor::Close() {
 #ifdef PADDLE_WITH_DISTRIBUTE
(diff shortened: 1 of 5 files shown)


guru4elephant:
This PR mainly aims to speedup ExecutorPrepareContext creation. In particular, when a user set use_ctx_cache() in python API, ExecutorPrepareContext can be cached in Executor. This is useful for declarative programing that a graph is static and we do not need to modify the graph during training. 

I've tested the efficiency of this change on bert. The timeline before and after optimization are as follows. Note that this change can be generalized to other scenario such as resnet50, transformer and so on.

Before Optimization
image

After Optimization
image

In the future, ExecutorPrepareContext will be exposed to python API, and we can cache program and ExecutorPrepareContext together to make the cache strategy more clearly.

paddle/fluid/framework/executor.cc
```diff
@@ -244,77 +252,21 @@ static bool has_fetch_operators(
...
-      op->SetOutput("Out", {var_name});
-      op->SetAttr("col", {static_cast<int>(i)});
-      op->CheckAttrs();
-
-      i++;
-    }
+void Executor::PrepareCtxCache(
+    const ProgramDesc& program, int block_id,
+    const std::vector<std::string>& skip_ref_cnt_vars, bool force_disable_gc) {
+  ctx_.reset(new ExecutorPrepareContext(program, block_id));
```
jacquesqiao: add an mutex?

paddle/fluid/framework/executor.cc
```diff
@@ -68,7 +63,11 @@ ExecutorPrepareContext::~ExecutorPrepareContext() {
   VLOG(5) << "destroy ExecutorPrepareContext";
 }
 
-Executor::Executor(const platform::Place& place) : place_(place) {}
+Executor::Executor(const platform::Place& place) : place_(place) {
+  cur_step_ = 0;
+  open_ctx_cache_ = false;
+  ctx_is_cached_ = false;
```
gavin1332: could we do a nullptr checking of ctx_ to judge whether ctx_is_cached is false? And we should pay attention to caching multiple ctx

paddle/fluid/framework/executor.cc
```diff
@@ -68,7 +63,11 @@ ExecutorPrepareContext::~ExecutorPrepareContext() {
   VLOG(5) << "destroy ExecutorPrepareContext";
 }
 
-Executor::Executor(const platform::Place& place) : place_(place) {}
+Executor::Executor(const platform::Place& place) : place_(place) {
+  cur_step_ = 0;
```
gavin1332: what if i use this exe to run other programs, like startup program. Do we need a reseting method for context pamameters?

paddle/fluid/framework/executor.cc
gavin1332: why?

gavin1332: > add an mutex?

why?

gavin1332 (dismissed):
LGTM

jacquesqiao (approved):
LGTM!

jacquesqiao:
Can you add a unit test to ensue that the cache works well.

PR #17543, opened 2019-05-21
Enable logical operators for the nGraph Bridge

kbinias:
Enable logical operators for the nGraph Bridge. BERT model uses logical_not operator.


luotao1 (approved):
LGTM

PR #17544, opened 2019-05-21
support NCCLContextPool singleton for collective op strategy

gavin1332:
some pointer and reference signs is formatted automatically by clang-format


gongweibao (commented):
Where and who will use `NCCLCommGroup`?
This pr is not completed.

gongweibao (changes requested):
Where and who will use NCCLCommGroup?
This pr is not completed.

gavin1332:
> Where and who will use NCCLCommGroup?
> This pr is not completed.

some comments have been added.
NCCLCommGroup is specially useful for hierarchical ring communication, in which, communicators are initialized with different ncclUniqueId, even though with the save world ranks.

guru4elephant (dismissed):
LGTM, you need verify that hierarchical all reduce is valid through NCCLCommGroup

PR #17549, opened 2019-05-21
Enable SQRT operator for the nGraph Bridge

kbinias:
Enable SQRT operator for the nGraph Bridge. BERT model uses this operator.


luotao1 (approved):
LGTM

PR #17565, opened 2019-05-22
fix trt ci error

NHZlX:
Tests were conducted locally and no errors were reproduced.
So, i ban all trt op converter ut temporary.


luotao1 (approved):
LGTM

PR #17567, opened 2019-05-22
decrease the train loop number to avoid run too long to fail the ci p…

wopeizl:
…rocess test=develop

some imperative unit test may run too long to fail the CI, decrease the run loop times to fix, and would not affect the final result.

image


gongweibao (approved):
LGTM

PR #17571, opened 2019-05-22
Fix distributed launch.py

chengduoZH:
Fix bug of launch.py.
The `CUDA_VISIBLE_DEVICES` may not begin with 0.


gongweibao (approved):
LGTM

PR #17037, opened 2019-04-22
Capi: Ngraph engine

mozga-intel:
Enable the capi pass for the nGraph bridge. BERT for an inference.

PR #17578, opened 2019-05-22
Batch norm operator, easiest way to align NG precision with PP precision

mozga-intel:
The BN update is to pass the test due to the BN fast math issue. That is the easiest way to make it align with paddle precision.  Capi issue,  #17037


baojun-nervana:
close the PR as the compensation is not preferred

PR #17588, opened 2019-05-23
add __str__ method for tensor and lodtensor to support print test=dev…

wopeizl:
…elop

extend the python __str__ method to support print operation on python side directly.


phlrain (approved):
LGTM

PR #17592, opened 2019-05-23
Fix dygraph unique name bug

sneaxiy:
`unique_name.ids` increases as batch increases, which make memory leak in dygraph mode since dygraph mode creates new variables across batches. This bug makes ptb model run slower as batch increases. This PR fixes this bug.

### DIFF
--- a/python/paddle/fluid/unique_name.py
+++ b/python/paddle/fluid/unique_name.py
@@ -58,6 +58,14 @@ def generate(key):
     return generator(key)
 
 
+def generate_tmp(key):
+    from .framework import in_dygraph_mode
+    if in_dygraph_mode():
+        key = "tmp"
+
+    return generator(key)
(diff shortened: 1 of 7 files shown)


python/paddle/fluid/unique_name.py
```diff
@@ -58,6 +58,14 @@ def generate(key):
     return generator(key)
 
 
+def generate_tmp(key):
```
JiabinYang: Add Comment to explain why we need generate_tmp

chengduoZH (dismissed):
LGTM

JiabinYang (approved):
LGTM

python/paddle/fluid/unique_name.py
sneaxiy: Done.

PR #17594, opened 2019-05-23
add support for cuda9 on windows test=develop

wopeizl:
add a patch for eigen since eigen has a bug for cuda9 on windows
add cuda binary adaption


qingqing01 (commented):
LGTM.

qingqing01 (commented):
LGTM.

cmake/external/eigen.cmake
```diff
@@ -12,6 +12,13 @@ if(NOT WITH_FAST_MATH)
   add_definitions(-DEIGEN_FAST_MATH=0)
 endif()
 
+if(WIN32)
+    set(PATCH_COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            ${PADDLE_SOURCE_DIR}/patches/eigen/Half.h.txt
+            ${EIGEN_INCLUDE_DIR}/Eigen/src/Core/arch/CUDA/Half.h)
```
chengduoZH: **I don't think we should do that for the users who need to compile Paddle in Windows, we should let users do that themselves to modify this file.**

wopeizl: remove the patch

qingqing01 (approved):
LGTM.

PR #17602, opened 2019-05-23
Code clean of Allocator

sneaxiy:
Clean code of Allocator.

1. Remove `AllocationWithUnderlying`
2. Unify allocation request when size = 0.


paddle/fluid/memory/allocation/allocator_facade.cc
```diff
@@ -252,12 +258,40 @@ class AllocatorFacadePrivate {
 #endif
   }
 
-  void WrapZeroSizeAllocator() {
-    for (auto& pair : allocators_) {
-      pair.second =
-          std::make_shared<ZeroSizeAllocator>(pair.second, pair.first);
+  class ZeroSizeAllocator : public Allocator {
```
chengduoZH, line 261: Why do you move `ZeroSizeAllocator` here?

sneaxiy: Because I think that it may be too complex that each allocator must be decorated by `ZeroSizeAllocator`. `ZeroSizeAllocator` should only handle allocation request size = 0 alone.

chengduoZH (approved):
LGTM

PR #17612, opened 2019-05-24
remove useless input from softmax_with_cross_entropy op

hutuxian:
Remove the useless input 'Softmax@GRAD' from backward phase of softmax_with_cross_entropy op.


phlrain (approved):
LGTM

PR #17616, opened 2019-05-24
add Run Prepared Ctx

guru4elephant:
test=develop


guru4elephant:
this PR fix previous merged PR https://github.com/PaddlePaddle/Paddle/pull/17536
one more RunPreparedCtx interface is added. This is more fast than previous one although in run-time both are ok in efficiency.

gavin1332 (approved):

PR #17630, opened 2019-05-24
Add auto pruning strategy.

wanghaoshuang:
1. Fix compressor.
2. Enhence graph executor.
3. Add SAController
4. Add auto pruning strategy.
5. Add unitest for auto pruning strategy.
test=develop


python/paddle/fluid/contrib/slim/core/config.py
```diff
@@ -20,11 +20,14 @@
 from ..quantization import *
 from .strategy import *
 from ..distillation import *
+from ..searcher import *
 
 __all__ = ['ConfigFactory']
 """This factory is used to create instances by loading and parsing configure file with yaml format.
 """
 
+PLUGINS = ['pruners', 'quantizers', 'quantizers', 'strategies', 'controllers']
```
kk12333: Distillers is missing.

wanghaoshuang: Thx. Fixed.

PR #16865, opened 2019-04-15
[Not for merge] INT8 MKL-DNN v2 integrate to slim first runnable version

bingyanghuang:
This is the first version to run the INT8 v2 solution in current slim.
User perspective:
1. User needs to indicate the FP32 model path
2. User needs to create a config.yaml file to indicate the save model path and different int8 quantization strategies 
3. User needs to give the ImageNet validation dataset location(hard code) 
4. After running, user can get a converted int8 model

Supported functionalities:
1. Run the int8v2 solution from slim. Command line:
`python python/paddle/fluid/contrib/slim/tests/test_infer_quant_strategy.py --infer_model_path=/path/to/Paddle/build/third_party/inference_demo/int8v2/resnet50/model --device=CPU`
2. Save the params after running the test

There exists some hard codes in this version,  will be removed when the functions are more stable(e.g Quantizer Config will be independent). This is just for sharing a feasibility plan.
@luotao1 @wzzju @wojtuss

Issue #656, opened 2016-11-29
docker环境下paddledev/paddle:gpu-demo-v0.9.0a0 ，执行quickstart，Cuda Error: CUDA driver version is insufficient for CUDA runtime version


beckett1124:
概述：
        docker环境下paddledev/paddle:gpu-demo-v0.9.0a0 ，执行quickstart，Cuda Error: CUDA driver version is insufficient for CUDA runtime version。

I1129 11:18:51.543231   335 Util.cpp:155] commandline: /usr/local/bin/../opt/paddle/bin/paddle_trainer --config=trainer_config.lr.py --save_dir=./output --trainer_count=4 --log_period=100 --num_passes=15 --use_gpu=True 
F1129 11:18:51.543578   335 hl_cuda_device.cc:526] Check failed: cudaSuccess == cudaStat (0 vs. 35) Cuda Error: CUDA driver version is insufficient for CUDA runtime version
*** Check failure stack trace: ***
    @     0x7f6d59f61daa  (unknown)
    @     0x7f6d59f61ce4  (unknown)
    @     0x7f6d59f616e6  (unknown)
    @     0x7f6d59f64687  (unknown)
    @           0x895bf0  hl_specify_devices_start()
    @           0x895dfd  hl_start()
    @           0x813362  paddle::initMain()
    @           0x52ac5b  main
    @     0x7f6d5916df45  (unknown)
    @           0x540c05  (unknown)
    @              (nil)  (unknown)
/usr/local/bin/paddle: line 109:   335 Aborted                 ${DEBUGGER} $MYDIR/../opt/paddle/bin/paddle_trainer ${@:2}
I1129 11:18:52.197794   356 Util.cpp:155] commandline: /usr/local/bin/../opt/paddle/bin/paddle_trainer --config=trainer_config.lr.py --use_gpu=false --job=test --init_model_path=output/pass-00003 --config_args=is_predict=1 --predict_output_dir=. 
I1129 11:18:52.197998   356 Util.cpp:130] Calling runInitFunctions
I1129 11:18:52.198359   356 Util.cpp:143] Call runInitFunctions done.
[WARNING 2016-11-29 11:18:52,343 networks.py:1438] `outputs` routine try to calculate network's inputs and outputs order. It might not work well.Please see follow log carefully.
[INFO 2016-11-29 11:18:52,343 networks.py:1466] The input order is [word]
[...]

Issue #654, opened 2016-11-29
data_layer的作用是什么呢？


keain:
Thank you for contributing to PaddlePaddle. Submitting an issue is a great help for us.
Both Chinese and English issues are welcome.

It's hard to solve a problem when important details are missing.
Before submitting the issue, look over the following criteria before handing your request in.

- [ ] Was there a similar issue submitted or resolved before ? You could search issue in the github.
- [ ] Did you retrieve your issue from widespread search engines ?
- [ ] Is my description of the issue clear enough to reproduce this problem?
   * If some errors occurred, we need details about `how do you run your code?`, `what system do you use?`, `Are you using GPU or not?`, etc.
   * If you use an recording [asciinema](https://asciinema.org/) to show what you are doing to make it happen, that's awesome! We could help you solve the problem more quickly.
- [ ] Is my description of the issue use the github markdown correctly?
   * Please use the proper markdown syntaxes for styling all forms of writing, e.g, source code, error information, etc.
   * Check out [this page](https://guides.github.com/features/mastering-markdown/) to find out much more about markdown.

reyoung:
请完善这个issue描述

PR #17634, opened 2019-05-24
INT8 MKL-DNN v2 integrate to slim

zhaify:
Refactor the code from PR [16865](https://github.com/PaddlePaddle/Paddle/pull/16865)

please help to review
@luotao1 @wzzju @wojtuss @bingyanghuang


zhaify:
@luotao1  MAC_PR_CI failed and the error message is bellow, is this the build server issue?

[02:38:51]W:	 [Step 1/1] + export PATH=/usr/local/opt/opencv@2/bin:/usr/local/bin:/Library/Frameworks/Python.framework/Versions/2.7/bin:./usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Java/JavaVirtualMachines/jdk-10.0.2.jdk/Contents/Home/bin
[02:38:51]W:	 [Step 1/1] + PATH=/usr/local/opt/opencv@2/bin:/usr/local/bin:/Library/Frameworks/Python.framework/Versions/2.7/bin:./usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Java/JavaVirtualMachines/jdk-10.0.2.jdk/Contents/Home/bin
[02:38:51]W:	 [Step 1/1] + export PATH=/usr/local/opt/opencv@2/bin:/usr/local/bin:/Library/Frameworks/Python.framework/Versions/2.7/bin:./usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Java/JavaVirtualMachines/jdk-10.0.2.jdk/Contents/Home/bin:/Users/paddle/Library/Python/2.7/bin
[02:38:51]W:	 [Step 1/1] + PATH=/usr/local/opt/opencv@2/bin:/usr/local/bin:/Library/Frameworks/Python.framework/Versions/2.7/bin:./usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Java/JavaVirtualMachines/jdk-10.0.2.jdk/Contents/Home/bin:/Users/paddle/Library/Python/2.7/bin
[02:38:51]W:	 [Step 1/1] + echo /usr/local/opt/opencv@2/bin:/usr/local/bin:/Library/Frameworks/Python.framework/Versions/2.7/bin:./usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Java/JavaVirtualMachines/jdk-10.0.2.jdk/Contents/Home/bin:/Users/paddle/Library/Python/2.7/bin
[02:38:51]W:	 [Step 1/1] + git config --global user.name 'Paddle CI_MAC'
[02:38:51]W:	 [Step 1/1] + git config --global user.email paddle_ci@example.com
[02:38:51]W:	 [Step 1/1] + set +e
[02:38:51]W:	 [Step 1/1] + git remote
[02:38:51]W:	 [Step 1/1] + grep upstream
[02:38:51]W:	 [Step 1/1] + '[' 1 == 1 ']'
[02:38:51]W:	 [Step 1/1] + git remote add upstream https://github.com/PaddlePaddle/Paddle.git
[02:38:51]W:	 [Step 1/1] + set -e
[02:38:51]W:	 [Step 1/1] + git fetch upstream
[02:38:51]W:	 [Step 1/1] fatal: unable to access 'https://github.com/PaddlePaddle/Paddle.git/': Failed to connect to 172.19.57.45 port 3128: Connection refused
[02:38:51]W:	 [Step 1/1] Process exited with code 128
[02:34:27]E:	 [Step 1/1] Process exited with code 128 (Step: Build and test (Command Line))
[02:38:51]E:	 [Step 1/1] Step Build and test (Command Line) failed
[02:38:51]i: Stopping performance monitoring process
[02:38:51]i: Performance monitoring process stopped
[02:38:51]i: Publishing performance monitoring build stages data
[02:38:51] : Publishing artifacts
[02:38:51] :	 [Publishing artifacts] Collecting files to publish: [/home/teamcity/system/perfmon/temp/104459/perfmon.csv=>.teamcity/perfmon/, /home/teamcity/temp/agentTmp/build_stages.txt=>.teamcity/perfmon/]
[02:38:51] :	 [Publishing artifacts] Publishing 2 files using [ArtifactsCachePublisher]: /home/teamcity/system/perfmon/temp/104459/perfmon.csv, /home/teamcity/temp/agentTmp/build_stages.txt => .teamcity/perfmon
[02:38:51] :	 [Publishing artifacts] Publishing 2 files using [WebPublisher]: /home/teamcity/system/perfmon/temp/104459/perfmon.csv, /home/teamcity/temp/agentTmp/build_stages.txt => .teamcity/perfmon
[02:38:51] : Publishing internal artifacts
[02:38:51] :	 [Publishing internal artifacts] Publishing 1 file using [ArtifactsCachePublisher]
[02:38:51] :	 [Publishing internal artifacts] Publishing 1 file using [WebPublisher]
[02:38:51]W: Publishing artifacts
[02:38:51] :	 [Publishing artifacts] Collecting files to publish: [output => output]
[02:38:51]W:	 [Publishing artifacts] Artifacts path 'output' not found
[02:34:27] : Build finished

luotao1:
I help you rerun at first. You can rerun yourself since sometimes ci is not stable.

zhaify:
@luotao1 
I use the following python code to download test_data and models (url: http://paddle-inference-dist.bj.bcebos.com/int8/imagenet_val_100_tail.tar.gz) before testing TestInferQuantizeStrategy, it works fine on Linux and Windows but always failed on MAC_PR_CI with http timeout. 
Do I need to add some additional configuration to the unit test on MAC_PR_CI?
``` python
from paddle.dataset.common import download
def setUp(self):
    self.cache_folder = os.path.expanduser(
        '~/.cache/paddle/dataset/int8/download')
    if not os.path.exists(self.cache_folder):
        os.makedirs(self.cache_folder)

    self.mobilenetv1 = 'mobilenetv1_int8_model.tar.gz'
    self.small_data = 'imagenet_val_100_tail.tar.gz'
    self.url_base = 'http://paddle-inference-dist.bj.bcebos.com/int8/'
    urls = [
        self.url_base + 'imagenet_val_100_tail.tar.gz',
        self.url_base + 'mobilenetv1_int8_model.tar.gz'
    ]
    md5s = [
        'b6e05365252f12f75e7daf3fcbc62e96',
        '13892b0716d26443a8cdea15b3c6438b'
    ]
    for url, md5 in zip(urls, md5s):
        download(url, self.cache_folder, md5)
        file_name = url.split('/')[-1]
        tar_file = tarfile.open(os.path.join(self.cache_folder, file_name))
        target_dir = file_name.split('.')[0]
        target_dir = os.path.join(self.cache_folder, target_dir)
        tar_file.extractall(target_dir)
``` 
CI errors
``` txt
[03:14:52] :	 [Step 1/1]     r = requests.get(url, stream=True)
[03:14:52] :	 [Step 1/1]   File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/requests/api.py", line 71, in get
[03:14:52] :	 [Step 1/1]     return request('get', url, params=params, **kwargs)
[03:14:52] :	 [Step 1/1]   File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/requests/api.py", line 57, in request
[03:14:52] :	 [Step 1/1]     return session.request(method=method, url=url, **kwargs)
[03:14:52] :	 [Step 1/1]   File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/requests/sessions.py", line 475, in request
[03:14:52] :	 [Step 1/1]     resp = self.send(prep, **send_kwargs)
[03:14:52] :	 [Step 1/1]   File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/requests/sessions.py", line 585, in send
[03:14:52] :	 [Step 1/1]     r = adapter.send(request, **kwargs)
[03:14:52] :	 [Step 1/1]   File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/requests/adapters.py", line 442, in send
[03:14:52] :	 [Step 1/1]     raise ConnectionError(e, request=request)
[03:14:52] :	 [Step 1/1] requests.exceptions.ConnectionError: HTTPConnectionPool(host='paddle-inference-dist.bj.bcebos.com', port=80): Max retries exceeded with url: /int8/imagenet_val_100_tail.tar.gz (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0x14258f438>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known',))
[03:14:52] :	 [Step 1/1] 
```

luotao1:
We will look at the network problem of mac ci machines.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,127 @@
...
+import struct
+import numpy as np
+import paddle.fluid as fluid
+from mobilenet import MobileNet
+from paddle.fluid.contrib.slim.core import Compressor
+from paddle.fluid.contrib.slim.graph import GraphWrapper
+from paddle.dataset.common import download
+
+
+class TestInferQuantizeStrategy(unittest.TestCase):
```
luotao1: Why this UT enabled on MAC/windows? It should be on LInux only. 
Please see CMakelists.txt.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    Test API of Post Training quantization strategy for int8 with MKLDNN.
+    """
+
+    def setUp(self):
```
luotao1: duplicated codes in `test_calibration_resnet50.py`, please refine it.

luotao1:
```
[03:38:23]	Cache file /Users/paddle/.cache/paddle/dataset/int8/download/imagenet_val_100_tail.tar.gz not found, downloading http://paddle-inference-dist.bj.bcebos.com/int8/imagenet_val_100_tail.tar.gzE
```
Besides, the download name is not right: `imagenet_val_100_tail.tar.gzE`.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: understand much thanks!

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: done

zhaify:
@luotao1 CI passed, should we start to review this PR ?

paddle/fluid/inference/api/analysis_predictor_tester.cc
```diff
@@ -260,7 +260,8 @@ class MkldnnQuantizerTest : public testing::Test {
     predictor.reset(new AnalysisPredictor(config));
     auto* predictor_p = static_cast<AnalysisPredictor*>(predictor.get());
 
-    auto qconfig = std::make_shared<MkldnnQuantizerConfig>();
+    // The MkldnnQuantizerConfig will be deleted by MkldnnQuantizer's destructor
```
sfraczek: Hey, this comment is no longer necessary. :)

zhaify: done

zhaify:
@wojtuss  @sfraczek Can you also help to review this PR?

paddle/fluid/inference/api/analysis_config.cc
```diff
@@ -168,11 +168,10 @@ void AnalysisConfig::EnableMkldnnQuantizer() {
   Update();
```
luotao1: why do you change `analysis_config.cc`, `mkldnn_quantizer.h` and `paddle_analysis_config.h`?

paddle/fluid/pybind/inference_api.cc
```diff
@@ -249,13 +257,41 @@ void BindAnalysisConfig(py::module *m) {
       .def("cpu_math_library_num_threads",
            &AnalysisConfig::cpu_math_library_num_threads)
       .def("to_native_config", &AnalysisConfig::ToNativeConfig)
+#ifdef PADDLE_WITH_MKLDNN
+      .def("enable_quantizer", &AnalysisConfig::EnableMkldnnQuantizer)
+      .def("quantizer_enabled", &AnalysisConfig::mkldnn_quantizer_enabled)
```
luotao1: Do you need polish `quantizer_enabled`?
- it is similar to `enable_quantizer`
- I don't see its usage in this PR.

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+import sys
+
+__all__ = ['InferQuantStrategy']
+
+FORMAT = '%(asctime)s-%(levelname)s: %(message)s'
+logging.basicConfig(level=logging.INFO, format=FORMAT, stream=sys.stdout)
+logger = logging.getLogger(__name__)
+
+
+class InferQuantStrategy(Strategy):
```
luotao1: The name `InferQuantStrategy` and `infer_quant_strategy.py` are not better, since it means Post Training quantization strategy.

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    def __init__(self,
+                 start_epoch=0,
+                 end_epoch=0,
+                 int8_model_save_path=None,
+                 fp32_model_path=None):
+        """
+        Args:
+            start_epoch(int): The 'on_epoch_begin' function will be called in start_epoch. default: 0
```
luotao1: where is `on_epoch_begin` function?

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+            end_epoch(int): The 'on_epoch_end' function will be called in end_epoch. default: 0
```
luotao1: where is `on_epoch_end` funcition.

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+            int8_model_save_path(str): The path to save model with int8_t weight.
+                            None means it doesn't save int8 model. defalut: None.
```
luotao1: > The path to save model with int8_t weight.
None (?) means it doesn't save int8 model

- it's confusing. do you mean fp32 programdesc + int8 weight?
- int8_t ->int8

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+                data = warmup_reader.__next__()
+
+            num_images = len(data)
+            dshape = [3, 224, 224]
+            shape = [num_images, 3, 224, 224]
+            image = core.PaddleTensor()
+            image.name = "x"
+            image.shape = shape
+            image.dtype = core.PaddleDType.FLOAT32
+            image.data.resize(num_images * 3 * 224 * 224 * sys.getsizeof(float))
```
luotao1: `[3, 224, 224]` should not be in a common function.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,112 @@
...
+
+        data_path = os.path.join(self.cache_folder,
+                                 'imagenet_val_100_tail/data.bin')
+        #warmup dataset, only use the first batch data
+        test_reader = paddle.batch(
+            self._reader_creator(data_path, False), batch_size=100)
+        com_pass = Compressor(
+            None,
+            None,
+            None,
```
luotao1: please write line93-95 as line96, we don't know the parameter name of None

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def test_infer_quant(self):
+        if not fluid.core.is_compiled_with_mkldnn():
+            return
+        self.download()
+        config_path = './quantization/config_mkldnn_int8.yaml'
+        with open(config_path, 'r+') as fp:
+            data = fp.read()
+        model_path = os.path.join(self.cache_folder,
+                                  'mobilenetv1_int8_model/model')
```
luotao1: please add resnet50 test as well.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,112 @@
...
+                                                          img_h), img)
+                    img = np.array(img)
+                    img.shape = (img_ch, img_w, img_h)
+                    fp.seek(labels_offset + label_size * step)
+                    label = fp.read(label_size)
+                    label = struct.unpack('q', label)[0]
+                    yield img, int(label)
+                    step += 1
+                    if cycle and step == num:
+                        step = 0
```
luotao1: could you reuse the reader in `test_calibraion_resnet50.py`?

zhaify: if reuse, the result (top1, top5) will not align with CAPI test (analyzer_int8_image_classification_tester.cc) since CAPI test use "data.bin" file for it's test.

luotao1: Do you mean only the dataset different?
https://github.com/PaddlePaddle/Paddle/blob/e4a52e082fe84a4713d05370f077eeda413a870c/python/paddle/fluid/contrib/tests/test_calibration_resnet50.py#L85-L90
I see the parameter has `file_list` and `data_dir` to pass the dataset.

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
zhaify: > * it's confusing. do you mean fp32 programdesc + int8 weight?

it will save int8 programdesc + fp32 weight, because mkldnn conv int8 kernel only support fp32 weight now.

paddle/fluid/inference/api/analysis_config.cc
zhaify: these changes can avoid share_ptr double free.
@wojtuss @bingyanghuang

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
luotao1: Got it. But the `int8_model_save_path` and its notes are too confusing. @bingyanghuang

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: > Do you mean only the dataset different? **Right**

> I see the parameter has `file_list` and `data_dir` to pass the dataset.

`file_list` parameter need a file contains all the files name like
``` txt
val/ILSVRC2012_val_00000093.JPEG 366
val/ILSVRC2012_val_00000094.JPEG 780
...
```
`data_dir` is the root path to all the pictures list in `file_list`,
the reader in test_calibration_resnet50.py will read each picture by path (data_dir + val/ILSVRC2012_val_*.JPEG), so i think it cannot pass "data.bin"

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
zhaify: how about 
`MKLDNNPostTrainingQuantStrategy`
`mkldnn_post_training_strategy.py`

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
zhaify: I will refine the notes

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
luotao1: I think it is OK. @bingyanghuang how do you think about?

luotao1:
please use `start a review` to reply comments
http://paddlepaddle.org/documentation/docs/zh/1.4/advanced_usage/development/contribute_to_paddle/submit_pr_guide.html#permalink-8--
image

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
bingyanghuang: > I think it is OK. @bingyanghuang how do you think about?

Looks fine to me.

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
zhaify: done

paddle/fluid/pybind/inference_api.cc
zhaify: quantizer_enabled removed

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
zhaify: removed

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/infer_quant_strategy.py
zhaify: done

paddle/fluid/pybind/inference_api.cc
```diff
@@ -249,13 +257,40 @@ void BindAnalysisConfig(py::module *m) {
       .def("cpu_math_library_num_threads",
            &AnalysisConfig::cpu_math_library_num_threads)
       .def("to_native_config", &AnalysisConfig::ToNativeConfig)
+#ifdef PADDLE_WITH_MKLDNN
+      .def("enable_quantizer", &AnalysisConfig::EnableMkldnnQuantizer)
```
wojtuss, line 260: The `EnableMkldnnQuantizer()` method is defined also without MKL-DNN, so I think we do not have to wrap it with the `#ifdef`.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
wojtuss: `2019`

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,108 @@
...
+
+        super(MKLDNNPostTrainingQuantStrategy, self).__init__(0, 0)
+        self.int8_model_save_path = int8_model_save_path
+        if fp32_model_path is None:
+            raise Exception("fp32_model_path is None")
+        self.fp32_model_path = fp32_model_path
+
+    def on_compression_begin(self, context):
+        """
+	Prepare the data and quantify the model
```
wojtuss: Please add indentation to the comment

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        super(MKLDNNPostTrainingQuantStrategy,
+              self).on_compression_begin(context)
+        logger.info('InferQuantStrategy::on_compression_begin')
```
wojtuss: `InferQuantStrategy` name is not used anymore. `MKLDNNPostTrainingQuantStrategy` should be used instead?

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,108 @@
...
+
+        warmup_data = [images, labels]
+
+        #Enable the int8 quantization
+        infer_config.enable_quantizer()
+        infer_config.quantizer_config().set_quant_data(warmup_data)
+        infer_config.quantizer_config().set_quant_batch_size(num_images)
+        #Run INT8 MKL-DNN Quantization
+        predictor = core.create_paddle_predictor(infer_config)
+        if self.int8_model_save_path and self.int8_model_save_path != '':
```
wojtuss: empty string gives `false`, the first check should be enough

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,108 @@
...
+            if not os.path.exists(self.int8_model_save_path):
+                os.makedirs(self.int8_model_save_path)
+            predictor.SaveOptimModel(self.int8_model_save_path)
+
+        logger.info('Finish InferQuantStrategy::on_compresseion_begin')
```
wojtuss: `InferQuantStrategy` name is not used anymore

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,134 @@
...
+            eval_program=None,
+            eval_reader=test_reader,
+            eval_feed_list=[],
+            eval_fetch_list=[],
+            teacher_programs=[],
+            checkpoint_path='',
+            train_optimizer=None,
+            distiller_optimizer=None)
+        com_pass.config(config_path)
+        eval_graph = com_pass.run()
```
wojtuss: Please add the accuracy result after the quantization. Otherwise, we cannot verify the quantization was done properly. I know the INT8v2 quantization is verified by the c-api unit tests, but the check should be added here as well, in my opinion.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def test_compression(self):
+        if not fluid.core.is_compiled_with_mkldnn():
+            return
+
+        base_url = 'http://paddle-inference-dist.bj.bcebos.com/int8/'
```
wojtuss: If it is possible, it would be good to reuse models and datasets downloaded for the c-api unit tests. They seem to be the same.
You can take a look at `paddle/fluid/inference/tests/api/CMakeLists.txt` and `paddle/fluid/inference/tests/test.cmake`.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

paddle/fluid/pybind/inference_api.cc
zhaify: done

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
```diff
@@ -1,15 +1,57 @@
...
 
-if(APPLE OR WIN32 OR NOT WITH_MKL)
-    list(REMOVE_ITEM TEST_OPS test_mkldnn_int8_quantization_strategy)
+# int8 image classification python api test
+if(LINUX AND WITH_MKLDNN)
+  set(INT8_DATA_DIR "${INFERENCE_DEMO_INSTALL_DIR}/int8v2")
+
+  #resnet50 int8
+  set(INT8_RESNET50_MODEL_DIR "${INT8_DATA_DIR}/resnet50")
+  inference_analysis_python_api_int8_test(test_analyzer_int8_resnet50 ${INT8_RESNET50_MODEL_DIR} ${INT8_DATA_DIR} test_mkldnn_int8_quantization_strategy.py)
```
wojtuss: With this configuration running
```
ctest -R test_analyzer_int8_resnet50
```
executes both: c-api unit test and the python slim test for ResNet50.
Please, make those tests separate ones, i.e. rename the tests here `test_analyzer_int8_resnet50`->`test_slim_int8_resnet50`, and so on.

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
```diff
@@ -1,15 +1,57 @@
...
+  #vgg19 int8
+  set(INT8_VGG19_MODEL_DIR "${INT8_DATA_DIR}/vgg19")
+  inference_analysis_python_api_int8_test(test_analyzer_int8_vgg19 ${INT8_VGG19_MODEL_DIR} ${INT8_DATA_DIR} test_mkldnn_int8_quantization_strategy.py)
+
+  #googlenet int8
+  set(INT8_GOOGLENET_MODEL_DIR "${INT8_DATA_DIR}/googlenet")
+  inference_analysis_python_api_int8_test(test_analyzer_int8_googlenet ${INT8_GOOGLENET_MODEL_DIR} ${INT8_DATA_DIR} test_mkldnn_int8_quantization_strategy.py)
 endif()
 
+list(REMOVE_ITEM TEST_OPS test_mkldnn_int8_quantization_strategy)
```
wojtuss, line 59: Please, add a comment on why the test is disabled.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -93,42 +153,59 @@ def _test_int8_quant(self, data_path, config_path):
...
+        _logger.info('--- FP32 prediction start ---')
+        val_reader = paddle.batch(
+            self._reader_creator(data_path, False), batch_size=batch_size)
+        fp32_model_result = self._predict(val_reader, fp32_model_path)
+        _logger.info('--- INT8 prediction start ---')
+        val_reader = paddle.batch(
+            self._reader_creator(data_path, False), batch_size=batch_size)
+        int8_model_result = self._predict(val_reader, int8_model_path)
+        _logger.info('--- comparing outputs ---')
+        _logger.info('Avg top1 INT8 accuracy: {0:.4}'.format(int8_model_result[
```
wojtuss: Please, make it printing accuracy with at least 4 digits precision after period, padding with zeros if necessary.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -12,22 +12,52 @@
...
+        type=str,
+        default='./output',
+        help='path to save int8 model with fp32 weights')
+    parser.add_argument(
+        '--warmup_batch_size',
+        type=int,
+        default=100,
+        help='batch size for quantization warmup')
+    parser.add_argument(
+        '--quantized_accuracy',
```
wojtuss: Please make the option name more meaningful, e.g. `--accuracy_diff_threshold`.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,104 @@
...
+
+        labels = core.PaddleTensor()
+        labels.name = "y"
+        labels.shape = [num_images, 1]
+        labels.dtype = core.PaddleDType.INT64
+        label_data = [label for (_, label) in data]
+        labels.data = core.PaddleBuf(label_data)
+
+        warmup_data = [images, labels]
+
```
wojtuss: Please, make it running on multiple cores by adding
```
infer_config.set_cpu_math_library_num_threads(???)
```
with a proper argument from `FLAGS_OMP_NUM_THREADS` env variable. Otherwise, the test does not utilize multiple cores.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,104 @@
...
+        super(MKLDNNPostTrainingQuantStrategy, self).__init__(0, 0)
+        self.int8_model_save_path = int8_model_save_path
+        if fp32_model_path is None:
+            raise Exception("fp32_model_path is None")
+        self.fp32_model_path = fp32_model_path
+
+    def on_compression_begin(self, context):
+        """
+	    Prepare the data and quantify the model
```
wojtuss: Please, make the comment aligned with indentation of `"""`.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
wojtuss: It is still not aligned with `"""`.

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
zhaify: done

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,211 @@
...
+            self._reader_creator(data_path, False), batch_size=batch_size)
+        int8_model_result = self._predict(val_reader, int8_model_path)
+        _logger.info('--- comparing outputs ---')
+        _logger.info('Avg top1 INT8 accuracy: {0:.4}'.format(int8_model_result[
+            0]))
+        _logger.info('Avg top1 FP32 accuracy: {0:.4}'.format(fp32_model_result[
+            0]))
+        _logger.info('Accepted accuracy drop threshold: {0}'.format(
+            quantized_accuracy))
+        assert abs(int8_model_result[0] - fp32_model_result[
```
wojtuss: Please, modify the assert condition so it pass only when
```
fp32_model_result[0] - int8_model_result[0] <= quantized_accuracy
```
It may happen that int8 accuracy is higher than fp32 accuracy, and that is OK.

zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

wojtuss (dismissed):

wojtuss (dismissed):

python/paddle/fluid/contrib/slim/tests/quantization/config_mkldnn_int8.yaml
```diff
@@ -0,0 +1,11 @@
+version: 1.0
+strategies:
+    mkldnn_post_training_strategy:
+        class: 'MKLDNNPostTrainingQuantStrategy'
+        int8_model_save_path: 'OUTPUT_PATH'
+        fp32_model_path: 'MODEL_PATH'
```
luotao1, line 22: Do you need `cpu_math_library_num_threads` here, i,e do not directly use `FLAGS_omp_num_thread` to change it? @bingyanghuang @wojtuss 
Maybe you can update in next PR.

luotao1 (commented):
@wzzju Do you have any other suggestions?

luotao1:
```
Test #651: test_slim_int8_mobilenet .....................   Passed   62.47 sec
Test #653: test_slim_int8_resnet101 .....................   Passed  162.41 sec
Test #656: test_slim_int8_googlenet ......................   Passed   63.16 sec
Test #652: test_slim_int8_mobilenetv2 ....................   Passed   78.97 sec
Test #650: test_slim_int8_resnet50 .........................   Passed   78.49 sec
Test #655: test_slim_int8_vgg19 .........................   Passed  234.20 sec
Test #654: test_slim_int8_vgg16 ............................   Passed  153.20 sec
```
These tests cost a lot of time on CI, do you have any solutions to speedup CI time?

bingyanghuang:
> ```
> Test #651: test_slim_int8_mobilenet .....................   Passed   62.47 sec
> Test #653: test_slim_int8_resnet101 .....................   Passed  162.41 sec
> Test #656: test_slim_int8_googlenet ......................   Passed   63.16 sec
> Test #652: test_slim_int8_mobilenetv2 ....................   Passed   78.97 sec
> Test #650: test_slim_int8_resnet50 .........................   Passed   78.49 sec
> Test #655: test_slim_int8_vgg19 .........................   Passed  234.20 sec
> Test #654: test_slim_int8_vgg16 ............................   Passed  153.20 sec
> ```
> 
> These tests cost a lot of time on CI, do you have any solutions to speedup CI time?

I have some suggestions: 1. use more omp num threads; 2. Decrease the accuracy test batch size, need test the influence for the accuracy comparison @zhaify 3. @wojtuss  could we decrease the warmup batch size < 100?

luotao1:
> use more omp num threads

We could not use more than 4 threads. Since the least cpu cores on ci machine are 16, and ci parallel level is 4.

luotao1:
@bingyanghuang @wojtuss Could we reduce the unit-tests and only choose the least one or two time-cost unit-tests such as mobilenet and googlenet? The reason is:
- The core computation of these tests are the same in the `analyzer_test_xxx`
- This PR is only for python wrapper, don't focus on the int8 computation.

python/paddle/fluid/contrib/slim/tests/quantization/config_mkldnn_int8.yaml
wojtuss: @luotao1 , I think you are right, the number of threads could be set here instead of using the `FLAGS_OMP_NUM_THREADS` environment variable.

wojtuss:
> I have some suggestions: 1. use more omp num threads; 2. Decrease the accuracy test batch size, need test the influence for the accuracy comparison @zhaify 3. @wojtuss could we decrease the warmup batch size < 100?
@bingyanghuang , @luotao1 , usually warmup batch size 50 or even less is enough for the test to pass. It has to be verified per model.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,109 @@
...
+        image_data = np.array(image_data).astype("float32")
+        image_data = image_data.ravel()
+        images.data = core.PaddleBuf(image_data.tolist())
+
+        labels = core.PaddleTensor()
+        labels.name = "y"
+        labels.shape = [num_images, 1]
+        labels.dtype = core.PaddleDType.INT64
+        label_data = [label for (_, label) in data]
+        labels.data = core.PaddleBuf(label_data)
```
qingqing01, line 103: 1. Here, the image and label will limit  the `MKLDNNPostTrainingQuantStrategy` only can be used for image classification.
2. Maybe need add `TODO` to remove this limits?

bingyanghuang:
> @bingyanghuang @wojtuss Could we reduce the unit-tests and only choose the least one or two time-cost unit-tests such as mobilenet and googlenet? The reason is:
> 
> * The core computation of these tests are the same in the `analyzer_test_xxx`
> * This PR is only for python wrapper, don't focus on the int8 computation.

The reason why we add these unit test in the python is to make sure that there is no diff for the results between C-API and Python API. @luotao1  If you think that one or two models tests can be enough for this results alignment test, we can reduce this test to 1 or 2 the least time-consuming test at once.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
bingyanghuang: Thanks qingqing for your review. We will add `TODO` to remove this limits.  @zhaify Please follows qiangqiang' suggestions.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,109 @@
...
+
+class MKLDNNPostTrainingQuantStrategy(Strategy):
+    """
+    The strategy for Post Training quantization strategy.
+    """
+
+    def __init__(self, int8_model_save_path=None, fp32_model_path=None):
+        """
+        Args:
+            int8_model_save_path(str): The path to save int8 model with fp32 weights.
```
wzzju: I don't understand what is the `int8 model with fp32 weights`?Please give more detail explanation.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,109 @@
...
+                            None means it doesn't save int8 model. defalut: None.
+            fp32_model_path(str): The path to model with fp32 weight. defalut: None.
```
wzzju: The same above.

luotao1:
@bingyanghuang I think you can reduce the ci time by
- first try https://github.com/PaddlePaddle/Paddle/pull/17634#issuecomment-498071688?
- if above try fails, then reduce this test to 1 or 2 the least time-consuming test at once. If you need 
 to reduce the tests, you could use some flags to make our QA tests them in local machine.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
luotao1: The same to me. I comment in https://github.com/PaddlePaddle/Paddle/pull/17634#discussion_r287941929, but it seems still confusing here.

zhaify: It means `MKLDNNPostTrainingQuantStrategy` will save **int8 programdesc + fp32 weights** to the path `int8_model_save_path`

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: For post training quantization, MKLDNNPostTrainingQuantStrategy support to convert a **fp32 programdesc with fp32 weights** to a **int8 programdesc with fp32 weights**, so fp32_model_path means the path to original fp32 model need to be quantized.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
bingyanghuang: > It means `MKLDNNPostTrainingQuantStrategy` will save **int8 programdesc + fp32 weights** to the path `int8_model_save_path`

@zhaify we should make the instructions/explanation in code more clearly.

bingyanghuang:
> @bingyanghuang I think you can reduce the ci time by
> 
> * first try [#17634 (comment)](https://github.com/PaddlePaddle/Paddle/pull/17634#issuecomment-498071688)?
> * if above try fails, then reduce this test to 1 or 2 the least time-consuming test at once. If you need
>   to reduce the tests, you could use some flags to make o.ur QA tests them in local machine.

Got it. Thanks

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
bingyanghuang: > The same above.

Will refactor the args instructions for them.

python/paddle/fluid/contrib/slim/tests/quantization/mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,212 @@
+#   copyright (c) 2019 paddlepaddle authors. all rights reserved.
```
wzzju: Is this file a unit test? If so, it should be placed in the `python/paddle/fluid/contrib/slim/tests/` path. What's more, its name should be started with `test_`. Otherwise, it will not be run in the `CI`.

python/paddle/fluid/contrib/slim/tests/quantization/mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,212 @@
...
+                    label = struct.unpack('q', label)[0]
+                    yield img, int(label)
+                    step += 1
+                    if cycle and step == num:
+                        step = 0
+
+        return reader
+
+    def _update_config_file(self, fp32_model_path, output_path):
+        config_path = './quantization/config_mkldnn_int8.yaml'
```
wzzju: Do you run this unit test file locally? If this file in `quantization` directory,  I don't think this UT can be run successfully because of the `config_path` value. When you raise a pull request, please be sure that its UT can be run successfully in your local machine.

python/paddle/fluid/contrib/slim/tests/quantization/mkldnn_int8_quantization_strategy.py
wzzju: `python/paddle/fluid/contrib/slim/tests/quantization` should only place some yaml config files, not python files.

zhaify: This is the suggestion comes form @wojtuss, he think that adding `test_mkldnn_int8_quantization_strategy.py` in `python/paddle/fluid/contrib/slim/tests/` and removed if not running on Linux platform compiled with MKLDNN is a bad idea.

python/paddle/fluid/contrib/slim/tests/quantization/mkldnn_int8_quantization_strategy.py
zhaify: It runs ok with cmd `ctest -R test_slim_int8_*` on my local machine.

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
zhaify, line 59: @wzzju how about this version of cmake

python/paddle/fluid/contrib/slim/tests/quantization/config_mkldnn_int8.yaml
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
zhaify: done

python/paddle/fluid/contrib/slim/tests/quantization/config_mkldnn_int8.yaml
```diff
@@ -4,6 +4,7 @@ strategies:
         class: 'MKLDNNPostTrainingQuantStrategy'
         int8_model_save_path: 'OUTPUT_PATH'
         fp32_model_path: 'MODEL_PATH'
+        omp_num_threads: 4
```
luotao1: How about `cpu_math_library_num_threads`?

zhaify: sounds good, I will change the name.

zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
luotao1: Do you refactor the args instructions? I see it is the same as before. The comments https://github.com/PaddlePaddle/Paddle/pull/17634#discussion_r289680391 and https://github.com/PaddlePaddle/Paddle/pull/17634#discussion_r289679585 are more clear than the args notes.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,116 @@
...
+    The strategy for Post Training quantization strategy.
+    """
+
+    def __init__(self,
+                 int8_model_save_path=None,
+                 fp32_model_path=None,
+                 cpu_math_library_num_threads=1):
+        """
+        Args:
+            int8_model_save_path(str): The path to save model for MKL-DNN int8 inference.
```
luotao1: I see there is no refactor.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,116 @@
...
+                            None means it doesn't save model. defalut: None.
+            fp32_model_path(str): The path of fp32 model to be converted to int8 model
+                            None means it doesn't have fp32 model. defalut: None.
+            cpu_math_library_num_threads(int): The num of omp threads user want
```
luotao1: The number of cpu math library threads.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,116 @@
...
+                 fp32_model_path=None,
+                 cpu_math_library_num_threads=1):
+        """
+        Args:
+            int8_model_save_path(str): The path to save model for MKL-DNN int8 inference.
+                            None means it doesn't save model. defalut: None.
+            fp32_model_path(str): The path of fp32 model to be converted to int8 model
+                            None means it doesn't have fp32 model. defalut: None.
+            cpu_math_library_num_threads(int): The num of omp threads user want 
+                            MKLDNNPostTrainingQuantStrategy to use. defalut: 1.
```
luotao1: defalut->default. @zhaify @bingyanghuang Please check all the notes again, thanks very much!

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,116 @@
...
+                            1 means only use one omp thread.
```
luotao1: 1 means only use one omp thread.
There are two verbs （means and use）in one sentence.

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
```diff
@@ -1,11 +1,59 @@
 file(GLOB TEST_OPS RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "test_*.py")
 string(REPLACE ".py" "" TEST_OPS "${TEST_OPS}")
 
+function(inference_analysis_python_api_int8_test target model_dir data_dir filename)
+    py_test(${target} SRCS ${filename}
+        ENVS FLAGS_OMP_NUM_THREADS=${CPU_NUM_THREADS_ON_CI}
+        ARGS --infer_model ${model_dir}/model
+             --infer_data ${data_dir}/data.bin
+             --int8_model_save_path int8_models/${target}
+             --warmup_batch_size 100
```
luotao1, line 10: I see you change it to 10 in https://github.com/PaddlePaddle/Paddle/pull/17634/commits/085642a9fdba219180e2084a8ef5bf2d117b207f#diff-1d840481eb99c8cb8ae2116875d02d57R9. Why do you change back to 100 again?

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
luotao1: I still don't understand line53: 
- why remove this test here? 
- Do you need to move this test after line 22?
- current CI not support for testing, which machine could test it?

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
zhaify: I have tried  `warmup_batch_size 10, 20,30,...100`, and `warmup_batch_size=10` is a very special  case that can pass CI (all the 7 models) which can reduce time cost from 30s+ to several seconds (googlenet) on our local machine, @bingyanghuang think it is not a good idea, so I change it back and reduce the number of images used in accuracy check from 100 to 10 that also can reduce some time cost by CI. 
And if you prefer the previous solution, I can rollback it.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done with your [suggestion](https://github.com/PaddlePaddle/Paddle/pull/17634#discussion_r290305425)

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
luotao1: Got it. But are 10 images too less to ensure the accuracy?

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
luotao1: how about https://github.com/PaddlePaddle/Paddle/pull/17634#discussion_r290310558?

python/paddle/fluid/contrib/slim/tests/quantization/config_mkldnn_int8.yaml
```diff
@@ -0,0 +1,12 @@
+version: 1.0
+strategies:
+    mkldnn_post_training_strategy:
+        class: 'MKLDNNPostTrainingQuantStrategy'
+        int8_model_save_path: 'OUTPUT_PATH'
+        fp32_model_path: 'MODEL_PATH'
+        cpu_math_library_num_threads: 4
```
luotao1, line 23: please comment why do you use 4 here.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,120 @@
...
+    The strategy for MKL-DNN Post Training quantization strategy.
+    """
+
+    def __init__(self,
+                 int8_model_save_path=None,
+                 fp32_model_path=None,
+                 cpu_math_library_num_threads=1):
+        """
+        Args:
+            int8_model_save_path(str): The path to save an int8 programdesc with fp32 weights for
```
luotao1: 1. programdesc-> ProgramDesc , please correct all.
2. `The path to save xxx`, is grammar right？

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,120 @@
...
+                        MKL-DNN int8 inference. For post training quantization,
+                        MKLDNNPostTrainingQuantStrategy only support to convert a fp32 programdesc
+                        with fp32 weights to a int8 programdesc with fp32 weights now. The saved
```
luotao1: a int8 -> an int8

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
```diff
@@ -0,0 +1,120 @@
...
+                        int8 programdesc with fp32 weights only can be executed with MKL-DNN enabled.
+                        None means it doesn't save model. default: None.
+            fp32_model_path(str): The path to the original fp32 programdesc with fp32 weights.
+                        None means it doesn't have a fp32 programdesc with fp32 weights. default: None.
+            cpu_math_library_num_threads(int): The number of cpu math library threads are used on
```
luotao1: which are used on

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
zhaify: > I still don't understand line53:

I have write a comment in CMakeList.txt before and refactor again. 
> * why remove this test here?


> * Do you need to move this test after line 22?

test_mkldnn_int8_quantization_strategy only support for testing on Linux platform with MKL-DNN enabled, otherwise it should be disabled (here is removed). If on Linux and enable MKL-DNN test_mkldnn_int8_quantization_strategy has been tested from line 20 -line 51.

> * current CI not support for testing, which machine could test it?

current CI means: on MAC OS is MAC_PR_CI, On windows OS is PR_Windows_CI.

python/paddle/fluid/contrib/slim/tests/quantization/config_mkldnn_int8.yaml
zhaify: done

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
zhaify: done with remaining googlenet & mobilenetv1 UTs and remove others,
and add cmake flags `WITH_SLIM_MKLDNN_FULL_TEST` for QA testing all models locally!

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
```diff
@@ -0,0 +1,220 @@
...
+
+class TestMKLDNNPostTrainingQuantStrategy(unittest.TestCase):
+    """
+    Test API of Post Training quantization strategy for int8 with MKL-DNN.
+    """
+
+    def _reader_creator(self,
+                        data_file='data.bin',
+                        cycle=False,
+                        predict_num=100):
```
wojtuss: Please, add an option to set the number of images to process. Currently, the value of 10 is hardcoded in lines 199 and 203, and cannot be modified.

wojtuss (changes requested):

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
zhaify: done with removed, since reduce predict number of images in accuracy check is not a good idea.

python/paddle/fluid/contrib/slim/quantization/mkldnn_post_training_strategy.py
zhaify: done

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
```diff
@@ -1,11 +1,63 @@
 file(GLOB TEST_OPS RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "test_*.py")
 string(REPLACE ".py" "" TEST_OPS "${TEST_OPS}")
 
+function(inference_analysis_python_api_int8_test target model_dir data_dir filename)
+    py_test(${target} SRCS ${filename}
+        ENVS FLAGS_OMP_NUM_THREADS=${CPU_NUM_THREADS_ON_CI}
```
sfraczek, line 6: Unfortunately it always runs on 1 thread regardless of  this parameter :(

sfraczek (changes requested):
We can't provide performance measurement yet because it always runs on 1 thread.

sfraczek (approved):

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
sfraczek: ok this can be added later

python/paddle/fluid/contrib/slim/tests/test_mkldnn_int8_quantization_strategy.py
wojtuss: OK. This can be done later.

wojtuss (approved):
There are still a few things to improve, but they can be refactored later.
LGTM.

luotao1:
@zhaify @bingyanghuang Please do refactor in next PR.
PR1: 
1. https://github.com/PaddlePaddle/Paddle/pull/17634#pullrequestreview-246201587
2. Investigate the CI unit test running time problem

PR2: 

3. Finish the test doc with performance and accuracy data

PR #17641, opened 2019-05-24
INT8 Fully-connected

### LINKED ISSUE #17335: INT8 FC MKLDNN kernel need be created
Create INT8 FC MKLDNN kernel based on new designed FP32 FC MKLDNN kernel.

Sand3r-:
Introduces int8 operator, integration with int8v2.
Resolves #17335


luotao1:
Since #15226 is merged, could you clean the commit log in this PR?

Sand3r-:
> Since #15226 is merged, could you clean the commit log in this PR?

Sure, I'll look into it :)

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -68,8 +69,8 @@ class FCPrimitiveFactory {
  private:
   void UpdateDataPointers(const ExecutionContext& ctx, Tensor* out,
                           const Tensor* in) {
-    input_->set_data_handle(const_cast<T*>(in->data<T>()));
-    output_->set_data_handle(out->mutable_data<T>(ctx.GetPlace()));
+    input_->set_data_handle(to_void_cast(in->data<T_in>()));
```
Sand3r-, line 81: I'm removing a const cast here, hence the approval requirement. (It's necessary so that I can replace former `T` type for `T_in` which indicates input type.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -221,31 +325,57 @@ class FCPrimitiveFactory {
...
-    const MKLDNNDeviceContext& dev_ctx, const ExecutionContext& ctx,
-    const Tensor* input, const Tensor* weights,
-    const mkldnn::engine& mkldnn_engine) {
+template <typename T_in, typename T_w, typename T_out>
+static std::shared_ptr<FCPrimitiveFactory<T_in, T_w, T_out>>
+GetPrimitiveFactory(const MKLDNNDeviceContext& dev_ctx,
+                    const ExecutionContext& ctx, const Tensor* input,
+                    const Tensor* weights,
+                    const mkldnn::engine& mkldnn_engine) {
   const std::string key = platform::CreateKey(
```
jczaja, line 400: Please update your key with per thread in all instances where key is created of this Op

Sand3r-: Done.

jczaja (dismissed):
LGTM

jczaja (dismissed):

bingyanghuang:
@luotao1 @wzzju Please start a review. Thanks

paddle/fluid/inference/tests/api/CMakeLists.txt
```diff
@@ -210,37 +212,37 @@ if(WITH_MKLDNN)
...
   set(INT8_MOBILENETV2_MODEL_DIR "${INT8_DATA_DIR}/mobilenetv2")
   download_int8_data(${INT8_MOBILENETV2_MODEL_DIR} "mobilenet_v2_int8_model.tar.gz" )
-  inference_analysis_api_int8_test_run(test_analyzer_int8_mobilenetv2 ${INT8_IMG_CLASS_TEST_APP} ${INT8_MOBILENETV2_MODEL_DIR} ${IMAGENET_DATA_PATH})
+  inference_analysis_api_int8_test_run(test_analyzer_int8_mobilenetv2 ${INT8_IMG_CLASS_TEST_APP} ${INT8_MOBILENETV2_MODEL_DIR} ${IMAGENET_DATA_PATH} ${ALLOWED_DIFF})
  
   # resnet101 int8
   set(INT8_RESNET101_MODEL_DIR "${INT8_DATA_DIR}/resnet101")
   download_int8_data(${INT8_RESNET101_MODEL_DIR} "Res101_int8_model.tar.gz" )
-  inference_analysis_api_int8_test_run(test_analyzer_int8_resnet101 ${INT8_IMG_CLASS_TEST_APP} ${INT8_RESNET101_MODEL_DIR} ${IMAGENET_DATA_PATH})
+  inference_analysis_api_int8_test_run(test_analyzer_int8_resnet101 ${INT8_IMG_CLASS_TEST_APP} ${INT8_RESNET101_MODEL_DIR} ${IMAGENET_DATA_PATH} 0.031)
```
luotao1: Why each test has different accuray_diff?

Sand3r-: Unfortunately the int8 tests aren't meant to run on Broadwell-like architecture (for both performance and accuracy results), and we had to lower the tolerance so that they pass on CI.
I can however set a common `accuracy_diff` for all tests if desired.

luotao1: If use common `accuracy_diff`, which value would be set?

Sand3r-: I'd propose 0.035 as to satisfy resnet101 broadwell avx2 requirements

lidanqing-intel:
Test machine: 6248
Command:
``
Colons can be used to align columns.

| Performance        | sha [FPS]           | sha + pr17641 [FPS]  |
| ------------- |:-------------:| -----:|
| googlenet      |       |      |
| resnet50 |         |     |
| resnet101|         |     |
| mobilenetv1|         |     |
| mobilenetv2 |         |     |
| vgg16 |         |     |
| vgg19 |         |     |
| mobilenet-ssd |         |     |

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -228,6 +238,66 @@ void CPUQuantizePass::QuantizeConv(Graph* graph,
...
+    // skip if should not be quantized
+    if (fc_op_desc->GetAttrIfExists<bool>("use_quantizer") != true ||
+        fc_op_desc->GetAttrIfExists<bool>("use_mkldnn") != true)
+      return;
+
+    GET_IR_NODE_FROM_SUBGRAPH(weights, weights, fc_pattern);
+    GET_IR_NODE_FROM_SUBGRAPH(input, input, fc_pattern);
+    GET_IR_NODE_FROM_SUBGRAPH(output, output, fc_pattern);
+
+    // get scales calculated after warmup, they scale variables to MAX=1.0
```
wzzju, line 272: Could you please give more comments about this ? I can't understand this sentence `they scale variables to MAX=1.0`.

paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -138,7 +143,10 @@ void AnalysisPredictor::MkldnnQuantizer::CalculateSingleScale(
       scales_[var_name] = GetMaxScalingFactor(var_tensor, is_unsigned);
       break;
     case ScaleAlgo::MAX_CH:
```
wzzju, line 145: What is the difference between `ScaleAlgo::MAX_CH` and `ScaleAlgo::MAX_CH_T`? Could you please add some comments here?

paddle/fluid/inference/tests/api/CMakeLists.txt
wzzju: I don't think use the highest common `accuracy_diff`, such as `0.035`,  is a good idea. Because  some tests' diff can be lower. What is your opinion about this? @luotao1  What's more, why `fc` has the most highest diff?  Is this a bug? @Sand3r-

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -101,18 +104,38 @@ class FCPrimitiveFactory {
     return dst_mem;
   }
 
+  mkldnn::memory Reorder(const memory& src_mem,
```
wzzju, line 123: What's the function of the `Reorder`? Please add some comments here.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -126,9 +149,85 @@ class FCPrimitiveFactory {
...
+      if (scale_weights_data[i] == 0.0)
+        bias_scales[i] = 1.0f;
+      else
+        bias_scales[i] = scale_in_data * scale_weights_data[i];
+    }
+
+    return bias_scales;
+  }
+
+  std::vector<float> ComputeOutputShiftScale(const ExecutionContext& ctx) {
```
wzzju, line 204: What is the shift scale?

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -37,7 +37,7 @@ using mkldnn::primitive;
 using mkldnn::stream;
```
wzzju, line 37: This `fc` op is too complex to understand. Please give more comments.

wzzju (changes requested):

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
Sand3r-: What is essentially meant here is that the scales are normalised. Later on through the value of `is_input_unsigned ` it is determined whether the scale should be multiplied by max_int8 (127) or max_uint8 (255) inside `QuantizeInput` method. The similar happens for output as well.

paddle/fluid/inference/api/mkldnn_quantizer.cc
Sand3r-: These enums (`ScaleAlgo::MAX_CH` for conv and `ScaleAlgo::MAX_CH_T` for fc) denote scale-finding algorithms which perform similar task, essentially __they find the per-output-channel scale of weights tensor__ (be it fc or conv).

They do that by finding the max absolute value of all weight values that correspond to a given output channel.

What is the difference between them?
The difference lays in the data format that `conv` and `fc` are using. 
For their weights, convolution uses `oihw` (output_channels, input_channels, height, width) data format, while fc uses `io` (input_channels, output_channels) data format.
To make it easier to reason about, let's flatten the dimensions of convolution's weights to 2D. This way we find that it uses an `oI` format where `I=i*h*w`. Because row-major order is assumed, the fastest changing dimension is `I`  which consecutive values can be find in consecutive columns and the slowest is `o` which is found in consecutive rows.

Now, since we want to find **per-output-channel** scales, we search for max value along each row (each `o`) of that 2D matrix. Hence the `eigen_tensor_mat.cast<double>().abs().rowwise().maxCoeff();` code in the `GetMaxChScalingFactor` function.

Untitled Diagram


Since FC is using an `io` format, the situation changes so that the fastest changing dimension is `o`, and the slowest is `i`. Hence to find per-channel scales, we can no longer use the same `.rowwise()` code. Therefore, we treat the fc weights as if they were transposed, and perform search through find absolute values of each column using `.colwise().maxCoeff()`. 

io

The `_T` in the `ScaleAlgo::MAX_CH_T` indicates that we should treat the weights as if they were transposed from `oi` format.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
Sand3r-: Reorder transforms mkl-dnn memory from one data format to another. More data can be found [here](https://intel.github.io/mkl-dnn/understanding_memory_formats.html).

In the particular overload you have mentioned, it also scales the resulting data. The particular use case of this function is to quantize weights.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
Sand3r-: Please take a look at the following equations where `y` denotes output, `x` input, `w` weight, `b` bias and `q` a scale, and where the subscripts of `q` denote the scale it corresponds to, while the subscripts of the tensors (f or q) imply whether **f**loat or **q**uantized values are meant:
image
(source: https://nervanasystems.github.io/distiller/algo_quantization.html#range-based-linear-quantization Section Symmetric mode)

The output scale provided to fc is relative to the absolute fp32 values.
When we want to set the desired output scale, we first have to take into consideration the influence of input and weight scaling factors (`q_x` and `q_w`). To compensate for the scaling of these, we first have to divide our output by the product of input and weight scales (`q_x * q_w`) so that the output scale is appropriately applied.

Since there are as many weight scales as there are output channels, we have to compute such a "shifted scale" for each  output channel. Later on, these scales are used to rescale fc's output.

paddle/fluid/inference/tests/api/CMakeLists.txt
Sand3r-: It's the nature of int8 computations and a dose of unluckiness that the fc provides higher diff than before. It so just happens that on the small dataset we have selected for CI tests, the resulting accuracy when FC is applied is lower, but when the tests are carried out for full imagenet validation dataset the accuracy is not much lower than before (at least on my SKX CPU). @lidanqing-intel is going to post accuracy results here carried out on full dataset, as well as benchmark results.

lidanqing-vv:
**Test machine:** 
CLX 6248

**Performance Command:**
`/home/danqing/Paddle/build/paddle/fluid/inference/tests/api/test_analyzer_int8_image\
_classification "ARGS" "--infer_model=/home/danqing/Paddle/build/third_party/inferen\
ce_demo/int8v2/googlenet/model" "--infer_data=/home/danqing/.cache/paddle/dataset/in\
t8/download/int8_full_val.bin" "--warmup_batch_size=100" "--batch_size=1" "--paddle_\
num_threads=1" "--iterations=1000"  >develop_googlenet_bs1_iter1000.txt 2>&1`

**Performance Benchmark:**

| INT8 Performance        | develop [FPS]    | PR17641 [FPS]  | Improvement|
| ------------- |-------------| ----- |--------
| googlenet      |    95.3486   |  101.7831    |  1.067 |
| resnet50 |    56.8499     |  69.3127   | 1.219  |
| resnet101|   31.5507      |  35.9413   | 1.139 |
| mobilenetv1|    270.8641    |  303.0129   |  1.118 |
| mobilenetv2 |  234.9245      |  281.5531   | 1.198  |
| vgg16 |    10.3644   | 12.6128   | 1.217 |
| vgg19 |     9.2676   |   11.1056  | 1.198 |
| mobilenet-ssd |   63.5919     | 68.8618  | 1.082 |

**Accuracy Command:**

`/home/danqing/Paddle/build/paddle/fluid/infer                                        ence/tests/api/test_analyzer_int8_image_classification "ARGS" "--infer_model=/home/danqing/Paddle/build/third_part                                        y/inference_demo/int8v2/vgg19/model" "--infer_data=/home/danqing/.cache/paddle/dataset/int8/download/int8_full_val                                        .bin" "--warmup_batch_size=100" "--batch_size=50" "--paddle_num_threads=10" "--iterations=1000" > vgg19_bs50_iter1000.txt 2>&1`

**Accuracy Benchmark:**

| PR17641 Accuracy | FP32 | INT8 | Diff |
| ------------- |---------|-----| -----
| googlenet      |    0.7050   |  0.6992     |0.58%  |
| resnet50 |     0.7663  |   0.7645   |  0.18%  |
| resnet101|     0.7750   |  0.7743    | 0.07% |
| mobilenetv1|      0.7078      |  0.7033   |  0.45%|
| mobilenetv2 |     0.7190     |   0.7131   | 0.59% |
| vgg16 |   0.7208     |    0.7160  | 0.48% |
| vgg19 |     0.7257     |   0.7075 | 1.82% |
| mobilenet-ssd |   0.7386  | 0.7320   |0.66% |

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
Sand3r-: This PR introduces changes on 3 levels:
1. Pass that enables FC (`fc_mkldnn_pass`)
Pass
The requirement in the last step is required so that int8 kernel can work properly on one of the FC layers found in VGG16 and VGG19 network.

2. MKL-DNN Quantizer, so that FC-related tensors can have their scales computed.
Quantizer

3. `fc_mkldnn_op` - so that the int8 part can be incorporated.
The control flow when instruction pointer enters MKL-DNN's FC op looks as follows:
FC Kernel

bingyanghuang:
**Compare** Accuracy | INT8-develop | INT8-PR | Diff
-- | -- | -- | --
googlenet | 0.7003 | 0.6992 | -0.11%
resnet50 | 0.7657 | 0.7645 | -0.12%
resnet101 | 0.7743 | 0.7743 | 0.00%
mobilenetv1 | 0.7041 | 0.7033 | -0.08%
mobilenetv2 | 0.7134 | 0.7131 | -0.03%
vgg16 | 0.7205 | 0.7160 | -0.45%
vgg19 | 0.7257 | 0.7075 | -1.82%

Tested on CLX6248 with latest branch.

lidanqing-vv:
Change `--warmup_batch_size=100` to `--warmup_batch_size=50`, Keep other parameters same as before.

Full dataset accuracy tested on CLX6248

| Compare Accuracy | Int8-develop | Int8-PR | Diff |
| ------------- |---------|-----| -----
| googlenet      |  0.7009     |    0.6991    |  -0.18% |
| resnet50 |   0.7650     |   0.7643   |  -0.07%   |
| resnet101|    0.7741     |   0.7743    | 0.02%  |
| mobilenetv1|       0.7044      |  0.7032    |  -0.12% |
| mobilenetv2 |    0.7153      |  0.7143     |  -0.10% |
| vgg16 |    0.7204     |   0.7184    | -0.20%  |
| vgg19 |     0.7255      |  0.7232  | -0.23%  |

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -228,6 +238,66 @@ void CPUQuantizePass::QuantizeConv(Graph* graph,
...
+  int quantize_fc_count = 0;
+  auto handler = [&](const GraphPatternDetector::subgraph_t& subgraph,
+                     Graph* g) {
+    VLOG(4) << "Quantize fc op";
+    GET_IR_NODE_FROM_SUBGRAPH(fc, fc, fc_pattern);
+    auto* fc_op_desc = fc->Op();
+
+    // skip if should not be quantized
+    if (fc_op_desc->GetAttrIfExists<bool>("use_quantizer") != true ||
+        fc_op_desc->GetAttrIfExists<bool>("use_mkldnn") != true)
```
wojtuss, line 265: I think checking for 'use_mkldnn' attribute value is superfluous here. It was checked when setting 'use_quantizer` attribute.

paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -331,18 +339,23 @@ AnalysisPredictor::MkldnnQuantizer::GetMaxChScalingFactor(
...
 
-  int channels = var_tensor.dims()[0];
-  LoDTensor scale_tensor = CreateScaleTensor(channels);
-  auto* scale_ptr = scale_tensor.mutable_data<double>(CPUPlace());
-
-  for (int i = 0; i < channels; ++i) {
-    const auto tensor = var_tensor.Slice(i, i + 1);
+  auto dims = var_tensor.dims();
+  bool is_transposed = oc_axis == (dims.size() - 1);
+  int num_col_dims = is_transposed ? oc_axis : oc_axis + 1;
```
wojtuss: Maybe the `int oc_axis` argument could be turned into `bool is_transposed` with the following simiplification:
(line 343 removed)
```
int num_col_dims = is_transposed ? dims.size() - 1 : dims.size();
```

paddle/fluid/inference/api/paddle_mkldnn_quantizer_config.h
```diff
@@ -26,10 +26,11 @@ namespace paddle {
...
 // Algorithms for finding scale of quantized Tensors.
 enum class ScaleAlgo {
-  NONE,    // Do not compute scale
-  MAX,     // Find scale based on the maximum absolute value
-  MAX_CH,  // Find scale based on the maximum absolute value per channel
-  KL,      // Find scale based on KL Divergence
+  NONE,      // Do not compute scale
+  MAX,       // Find scale based on the max absolute value
+  MAX_CH,    // Find scale based on the max absolute value per output channel
+  MAX_CH_T,  // Find scale based on the max absolute value per input channel
```
wojtuss: imo `... per output channel of a transposed tensor` would convey the meaning a little bit closer

paddle/fluid/inference/tests/api/analyzer_int8_image_classification_tester.cc
```diff
@@ -28,6 +28,7 @@ void SetConfig(AnalysisConfig *cfg) {
   cfg->SwitchSpecifyInputNames();
   cfg->SetCpuMathLibraryNumThreads(FLAGS_paddle_num_threads);
   cfg->EnableMKLDNN();
+  cfg->pass_builder()->AppendPass("fc_mkldnn_pass");
```
wojtuss: Why not add the pass to the suite of MKL-DNN passes?

paddle/fluid/inference/api/mkldnn_quantizer.cc
Sand3r-: In that case, we would need to derive `oc_axis` again anyway so that it can be used here: https://github.com/PaddlePaddle/Paddle/pull/17641/files/904cd05467add542e690b5bd81f8e5ec78d2a7b4#diff-16a31361a76290e4220c11ea16e4e8f5R355

It would probably then something like 
```C++
int oc_axis = is_transposed;
```

But even doing this would make it more clear, thank you for the suggestion.

paddle/fluid/inference/api/paddle_mkldnn_quantizer_config.h
Sand3r-: I remember I spent quite some time figuring out how to not exceed 80 characters per line limit still maintaining the message that would suffice 😄 But yes, your description is definitely much more accurate. I'll modify it.

paddle/fluid/inference/tests/api/analyzer_int8_image_classification_tester.cc
Sand3r-: Please see https://github.com/PaddlePaddle/Paddle/pull/15226#pullrequestreview-238328905
Since MKL-DNN's FC was only bringing performance improvement on some topologies, it has been decided to have it enabled optionally.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
Sand3r-: Could you please provide my the link to the code that does that? I can't find it in `cpu_quantize_placement_pass.cc`

Sand3r-: As far as I recall, I have added it specifically here, since without this check when the topology was run without `fc_mkldnn_pass` it was actually quantizing the fp32 fc and then it was failing on finding a kernel that supports uint8 or int8 input.

Sand3r-: Yup, it's this commit a417bdb0de244febc72fd72c1526180ccb048fed that introduced that. Please see its title.

wojtuss: True, thanks.

paddle/fluid/inference/tests/api/analyzer_int8_image_classification_tester.cc
wojtuss: OK

wojtuss (dismissed):
LGTM
In my opinion, the code is simplified as much as possible and can't be simplified much further. Using MKL-DNN imposes some code complexity, but it becomes clear after getting used to the pattern.

luotao1:
Please solve the conflict

Sand3r-:
> Please solve the conflict

It's done, thanks :)

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
```diff
@@ -71,13 +75,13 @@ void SetOp(ProgramDesc* prog, const std::string& type, const std::string& name,
 
 namespace {
 static const std::initializer_list<std::string> variable_names{
-    "a", "w1", "c",  "d", "w2", "e",  "f", "g",
-    "h", "w3", "b1", "i", "j",  "w4", "b2"};
+    "a",  "w1", "c", "d", "w2", "e",  "f",  "g", "h",
```
wzzju, line 78: Why there are two blanks before some chars, such as "w1"/"g"? Please check more.

paddle/fluid/inference/api/mkldnn_quantizer.cc
wzzju: OK. Thanks.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
wzzju: OK. Thanks.

wzzju (dismissed):
Mostly LGTM.

luotao1:
It fails ON CI_COVERAGE

luotao1:
```
[11:24:17]	[Step 1/1] 1 . You must have one RD (XiaoguangHu01,Xreki,luotao1,sneaxiy) approval for the usage (either add or delete) of const_cast.
[11:24:17]	[Step 1/1] 2 . The error message you wrote in PADDLE_ENFORCE{_**} or PADDLE_THROW does not meet our error message writing specification. Possible errors include 1. the error message is empty / 2. the error message is too short / 3. the error type is not specified. Please read the specification [ https://github.com/PaddlePaddle/Paddle/wiki/Paddle-Error-Message-Writing-Specification ], then refine the error message. If it is a mismatch, please specify chenwhql (Recommend), luotao1 or lanxianghit review and approve.
[11:24:17]	[Step 1/1] The PADDDLE_ENFORCE or PADDLE_THROW entries that do not meet the specification are as follows:
[11:24:17]	[Step 1/1] PADDLE_ENFORCE_EQ(name_found, true, "%s isn't the input of the %s operator",
[11:24:17]	[Step 1/1] +                    input_name, op->Op()->Type());
[11:24:17]	[Step 1/1] PADDLE_ENFORCE_EQ(name_found, true, "%s isn't the output of the %s operator",
[11:24:17]	[Step 1/1] +                    output_name, op->Op()->Type());
[11:24:17]	[Step 1/1] PADDLE_ENFORCE_NOT_NULL(graph);
[11:24:17]	[Step 1/1] PADDLE_ENFORCE_NOT_NULL(scope);
[11:24:17]	[Step 1/1] PADDLE_ENFORCE_EQ(platform::is_cpu_place(ctx.GetPlace()), true,
[11:24:17]	[Step 1/1] +                      "It must use CPUPlace.");
[11:24:17]	[Step 1/1] 
[11:24:17]	[Step 1/1] There are 2 approved errors.
[11:24:17]	[Step 1/1] ****************
```
@Sand3r- http://ci.paddlepaddle.org/viewLog.html?buildId=228829&tab=buildLog&buildTypeId=YYTest_PrCiCpuPy2&logTab=tail

luotao1 (approved):
LGTM

Issue #17535, opened 2019-05-21
conv_transpose_mkldnn has accuracy diff


luotao1:
`conv_transpose_mkldnn_op` has accuracy diff on our internal face model.
commit: 3fe2def is OK.
commit: 1eb1517 is wrong.
If I remove `conv_transpose` in op_lists, the accuracy is OK.
```
if (use_mkldnn) {
     cfg.EnableMKLDNN();
     std::unordered_set<std::string> op_list={"conv", "batch_norm", "relu", "leaky_relu", "elementwise_add", "tanh"};
     cfg.SetMKLDNNOp(op_list);
   }
```

jczaja:
@luotao1 I understand that exact commit that introduce inaccuracy is  unknown. Since we do not have access to your face workload, is it possible your validation narrow down and trace commit that introduce a problem? Or perhaps some public test can be run to show inaccuracy?

jczaja:
I can see that conv2d transpose of MKL-DNN was merged 29th of November 2018. So this suggest that inaccuracy
was there from the moment of introduction of conv2d transpose mkl-dnn.

luotao1:
I test accuracy when your PR merged https://github.com/PaddlePaddle/Paddle/pull/14608#issuecomment-442442212, and it runs OK on accuracy diff.
I will see how to give you more clues about this issue.
By the way, do you have any suggestion?

luotao1:
I replace the `conv_transpose_mkldnn_op` with https://github.com/jczaja/Paddle/blob/1540df51cf5b88a08afec4e94d5fc86d4ddfe5c5/paddle/fluid/operators/conv_transpose_mkldnn_op.cc from #14608, but it still has accuracy diff.

jczaja:
My understanding:
1. So when you roll back to conv transpose 2d mkldnn as it was introduced all is fine. 
2. When you put conv2d transpose mkldnn from PR as it was merged into current develop there is a problem.

If this is what is seems then:
a) Meantime MKL-DNN was updated and then have implemented efficient JIT version for conv2d transpose , perhaps there is mistake there. Moving back mkl-dnn is not an option as INT8 kerenls were added. I can prepare (tomorow when I'm in a office)
branch with disabled efficent deconvolution (disabling inside mkl-dnn) and we can see if accuracy diff is fine.

Meantime:
1) please provide MKLDNN_VERBOSE=1 log from develop
So I can see which implementation is called on develop where problems are reported.

[...]

PR #17644, opened 2019-05-24
[MKL-DNN] conv_transpose mkldnn bias pass

jczaja:
This is PR is a fix to issue #17535. Changes here are introducing pass fusing conv_transpose and elementwise_add mkl-dnn ops which is used in Face Model.


luotao1 (approved):
LGTM

PR #17645, opened 2019-05-24
Adding matmul op and its grad to ngraph engine 

baojun-nervana:
Added matmul and matmul_grad op (need for bert)
- added a layout set in ngraph engine to ensure the right layout (backend may set to different layout due to optimization)


mozga-intel (approved):
LGTM!

tensor-tang (approved):
LGTM

PR #17647, opened 2019-05-24
Add lookup_table op and its grad to ngraph engine

baojun-nervana:
Added lookup_table and lookup_grad op. The ngraph is updated to support those ops. The ops are needed for BERT model


mozga-intel (approved):
LGTM!

cmake/external/ngraph.cmake
```diff
@@ -37,7 +37,7 @@ INCLUDE(GNUInstallDirs)
 INCLUDE(ExternalProject)
 
 SET(NGRAPH_PROJECT         "extern_ngraph")
-SET(NGRAPH_GIT_TAG         "096ad6ef0c04d57db1522940dbdf9a0652768065")
+SET(NGRAPH_GIT_TAG         "4ec94acc11084a5d53418f565529310fa584899a")
```
tensor-tang, line 40: why change this？

baojun-nervana: The ngraph::op::ScatterAdd (used in this lookup_table_op) is supported with new ngraph version.

tensor-tang (approved):
LGTM

PR #17653, opened 2019-05-26
Feature/init grad in time

JiabinYang:
This PR using lazy `Init Grad` to save GPU memory in dygraph mode.

Make test on  [`Resnet`](https://github.com/PaddlePaddle/models/tree/develop/dygraph/resnet), and with `Tesla K40m` (GPU memory of 11439MiB) :
With lazy mode: max batch size is 56
With previous mode: max batch size is 29


JiabinYang:
> No description provided.

PR #17662, opened 2019-05-27
Gather dup index grad fix

wawltor:
test=develop


shanyi15 (approved):
LG doc

chengduoZH (approved):
LG api.spec

paddle/fluid/operators/scatter.cu.h
```diff
@@ -28,13 +28,17 @@ using Tensor = framework::Tensor;
...
   CUDA_1D_KERNEL_LOOP(i, index_size * slice_size) {
     int indices_i = i / slice_size;
     int slice_i = i - indices_i * slice_size;  // offset inside the slice
     IndexT scatter_i = indices[indices_i];
     IndexT out_i = scatter_i * slice_size + slice_i;
-    *(output + out_i) = *(params + i);
+    if (overwrite) {
+      *(output + out_i) = *(params + i);
+    } else {
+      *(output + out_i) += *(params + i);
```
kuke, line 40: https://github.com/PaddlePaddle/Paddle/blob/ad61e1b22c1625db7e096d207c4240fdfea9b2b8/paddle/fluid/operators/lookup_table_op.cu#L67 Should use atomic add here

python/paddle/fluid/tests/unittests/test_gather_op.py
```diff
@@ -79,5 +79,23 @@ def config(self):
         self.index_type = "int64"
 
 
+class TestCase4(TestGatherOp):
```
kuke, line 82: Please add a unit case for duplicate indices in overwrite mode.

paddle/fluid/operators/scatter.cu.h
```diff
@@ -28,13 +28,17 @@ using Tensor = framework::Tensor;
...
-                                  size_t slice_size) {
+                                  size_t slice_size, bool overwrite) {
   CUDA_1D_KERNEL_LOOP(i, index_size * slice_size) {
     int indices_i = i / slice_size;
     int slice_i = i - indices_i * slice_size;  // offset inside the slice
     IndexT scatter_i = indices[indices_i];
     IndexT out_i = scatter_i * slice_size + slice_i;
-    *(output + out_i) = *(params + i);
+    if (overwrite) {
+      *(output + out_i) = *(params + i);
```
kuke, line 38: This element-wisely overwriting may produce a wrong result. Assume that we assign two vectors A and B to the same position in this way,  the elements of the vector in this position may partially come from A and partial comes from B. Please consider the vector-wise assignment.

PR #17667, opened 2019-05-27
Add Event in ScopeBuffer Executor

chengduoZH:
Add Event in ScopeBuffer Executor And FastThreadedSSAGraphExecutor.


paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
```diff
@@ -31,31 +31,18 @@ ScopeBufferedSSAGraphExecutor::ScopeBufferedSSAGraphExecutor(
       underlying_executor_(std::move(underlying_executor)),
       local_scopes_(std::move(local_scopes)),
       var_infos_(std::move(var_infos)),
-      places_(std::move(places)) {}
+      places_(std::move(places)),
+      threads_pool_(1) {
+  PrepareLocalExeScopes();
```
chengduoZH: `PrepareLocalExeScopes` is time-consuming, because the operations of scope contain locks, especially for the big model with data parallelism mode, because in this case, there are so many variables that need be initialized.

Xreki (approved):
LGTM

PR #17690, opened 2019-05-28
Gather dupliate index fix

wawltor:
test=develop


wawltor:
close the same pr

PR #17700, opened 2019-05-28
Modify keys in the dictionary of layer.stat_dict

junjun315:
Modify keys in the dictionary of layer.stat_dict in dynamic graph mode.

use the var.name instead


phlrain (approved):
LGTM

PR #17705, opened 2019-05-28
Fix the gather op grad calc bug when has same index

wawltor:
test=develop


luotao1:
why not create a new commit in #17662, #17690 and #17696 and #17705 are the same pr with #17662.

wawltor:
> why not create a new commit in #17662, #17690 and #17696 and #17705 are the same pr with #17662.

Sorry, I didn't realize  it was wrong, I will close the duplicate pr.

PR #17707, opened 2019-05-29
Split test_dist_mnist into multiple unit tests to avoid timeout

sandyhouse:
Split the unit test _test_dist_mist_ into multiple unit tests (test_dist_mnist, test_dist_mnist_nccl and test_dist_mnist_lars) to avoid timeout.
We do that because _test_dist_mnist_ contains multiple unit tests which is very easy to timeout.


gongweibao (approved):
LGTM

PR #17709, opened 2019-05-29
enable multi-threading for elementwise ops

LeoZhao-BJ:
enable omp for elementwise ops, which can speed up BERT training on CPU by using multi-core and threads.

test=develop


paddle/fluid/operators/elementwise/elementwise_op_function.h
```diff
@@ -258,14 +258,26 @@ class TransformFunctor {
 
   inline void RunRowWise(int n, int pre) const {
     platform::Transform<DeviceContext> trans;
-    trans(ctx_, x_, x_ + nx_, RowwiseTransformIterator<T, DeviceContext>(y_, n),
-          z_, func_);
+#ifdef PADDLE_WITH_MKLML
+#pragma omp parallel for
+#endif
+    for (int ni = 0; ni < pre; ni++) {
```
luotao1, line 264: Is this `for` slowers on `threads=1`?

LeoZhao-BJ: I suppose the cost is tiny, given in internal implementation, it uses std::transform which is a kind of loop actually. But for some cases, e.g. pre, n, post is not big enough, multi-theading may be slower than single due to thread spawn/converge cost.
Anyway, it is better to measure performance delta on single thread with this patch, @luotao1, do you have suggestion which workload can be used for measurement?

luotao1: I don't have inspired workload for measurement. Could you give some benchmark on it?
Or could you distinguish multi-thread and one-thread in codes?

luotao1:
> which can speed up BERT training on CPU by using multi-core and threads.

Do you use executor+ngraph multi-thread? We may use parallel_executor+ngraph one-thread in deployment.

paddle/fluid/operators/elementwise/elementwise_op_function.h
LeoZhao-BJ: sure, I will give some benchmark for our typical workloads.

LeoZhao-BJ:
> > which can speed up BERT training on CPU by using multi-core and threads.
> 
> Do you use executor+ngraph multi-thread? We may use parallel_executor+ngraph one-thread in deployment.

This patch is actually for ParallelExecutor DO for training case, without ngraph. By using multi-thread, on both ParallelExecutor and Executor, I see performance improved. 
For one-thread in deployment, I understand you mean for inference, right?  yes, for inference, we still use single-thread mode.

luotao1:
> By using multi-thread, on both ParallelExecutor and Executor, I see performance improved.

Do you mean ParallelExecutor + Op multi-thread? I mean the multi-thread called by ParallelExecutor.

LeoZhao-BJ:
> > By using multi-thread, on both ParallelExecutor and Executor, I see performance improved.
> 
> Do you mean ParallelExecutor + Op multi-thread? I mean the multi-thread called by ParallelExecutor.

two level multi-thread, one is ParallelExecutor thread pool for multi-node, the other is op level multi-thread. I tried both, e.g. use CPU_NUM to control ParallelExecutor level data parallelism, use OMP_NUM_THREADS to control op level. so final computing thread num is CPU_NUM * OMP_NUM_THREADS.

for CPU_NUM > 1 && OMP_NUM_THREADS = 1 case, the performance is restrained, not better than CPU_NUM > 1 && OMP_NUM_THREADS > 1 combination.

LeoZhao-BJ:
I did test on 2650 machine for resnet, small_dam, bert, transformer, and didn't see performance drop
image

PR #17711, opened 2019-05-29
add ops and kernels that mul, scale, fc, relu, softmax, dropout, elem…

lijianshe02:
add ops and kernels that mul, scale, fc, relu, softmax, dropout, elementwise_add, activation for X86


Superjomn (approved):
LGTM

PR #17715, opened 2019-05-29
Refactor dygraph using smart pointers

sneaxiy:
1. Refactor dygraph using `std::shared_ptr` and `std::weak_ptr`, since it is hard to maintain object using both in C++ and Python without smart pointer.

2. Move most of the codes from Python to C++, like op tracing, lifecycle management of `OpBase`, etc.


paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
-      }
+OpBase::OpBase(Tracer* tracer, size_t id, const framework::OpDesc& op_desc,
+               const NameVarBaseMap& ins, const NameVarBaseMap& outs,
+               const platform::Place& place, bool trace_backward)
+    : tracer_(tracer), id_(id), type_(op_desc.Type()), place_(place) {
+  platform::RecordEvent event(type_);
+  info_ = &(framework::OpInfoMap::Instance().Get(type_));
+  auto op = framework::OpRegistry::CreateOp(op_desc);
+
+  RunOp(op, ins, outs);
```
JiabinYang: `RunOp` in `OpBase`'s  constructor is a little weird, why don't u do this in `Trace`

paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
-    if (info.infer_var_type_) {
-      RuntimeInferVarTypeContext infer_var_type_ctx(
-          &grad_input_vars_[k], &tmp_grad_outputs[k], &attrs_);
-      info.infer_var_type_(&infer_var_type_ctx);
+  prepared_op.Run();
+
+  VLOG(3) << "Running Op " << op->Type() << " ends";
+}
+
+void OpBase::TraceBackward(const framework::OpDesc& fwd_op,
```
JiabinYang, line 389: looks like this func is also better to belong to trace...

paddle/fluid/imperative/layer.h
```diff
@@ -13,403 +13,285 @@
...
+      const std::vector<std::shared_ptr<VarBase>>& inputs) {
+    return {};
+  }
 };
 
-/* The wrapper for OpDesc which holds a OpDesc and a OpDesc of its
- * gradient. This object should be managed totally by Python intepreter.
- */
-class PYBIND11_HIDDEN OpBase {
+using NameVarBaseMap =
```
JiabinYang: Move this to `type_def.h` ?

paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
+    : tracer_(tracer), id_(id), type_(type), place_(place) {
+  platform::RecordEvent event(type_);
+  info_ = &(framework::OpInfoMap::Instance().Get(type_));
+
+  // Step 1: Run forward
+  if (info_->Checker() != nullptr) {
+    info_->Checker()->Check(&attrs);
+  }
+
+  auto input_name_map = CreateVarNameMap<true>(info_, type_, ins);
```
JiabinYang: how about build `name_to_var` here instead of scan `ins` twice in `TrackBackward`

paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
+  auto output_name_map = CreateVarNameMap<false>(info_, type_, outs);
```
JiabinYang: same as before

paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
+
+      for (auto& grad_out_var_name : grad_outs.second) {
+        auto iter = grad_to_var.find(grad_out_var_name);
+        PADDLE_ENFORCE(iter != grad_to_var.end(),
+                       "Cannot find output of input grad %s in op %s",
+                       grad_out_var_name, type_);
+        auto fwd_var_iter = name_to_var.find(iter->second);
+        PADDLE_ENFORCE(fwd_var_iter != name_to_var.end(),
+                       "Cannot find forward variable named %s", iter->second);
+        bwd_out.emplace_back((*(fwd_var_iter->second))->GradVarBase());
```
JiabinYang, line 492: Init grad `VarBase` with `shape` and  `dtype` here

paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
-  // call backward hooks
-  for (py::object& callable : backward_hooks_) {
-    callable(this);
+void AutoGradImpl::CheckBackwardOutputs(OpBase* op) {
+  for (auto& out : op->BackwardOutputs()) {
+    for (auto& pair : out) {
+      for (auto& var : pair.second) {
+        if (!var) continue;
+
+        auto pair = accumulators_.emplace(
```
JiabinYang: No need to give each var a different accumulator?

paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
+      NameVarBaseMap tmp_outs;
+      // A var may be coresponding to several grad var in one op
+      std::unordered_map<VarBase*, std::vector<std::shared_ptr<VarBase>>>
+          var_map;
+      size_t counter = 0;
+      for (auto& bwd_out : bwd_outs[i]) {
+        auto& tmp_var_list = tmp_outs[bwd_out.first];
+        tmp_var_list.reserve(bwd_out.second.size());
+        for (auto& var : bwd_out.second) {
+          auto tmp_var = std::make_shared<VarBase>(false);  // Do not need grad
```
JiabinYang: why share_ptr here?

paddle/fluid/imperative/layer.cc
```diff
@@ -240,189 +350,452 @@ std::unique_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
...
+  void operator()();
+
+ private:
+  void PrepareDeps();
+
+  bool CheckBackwardInputs(OpBase* op);
+
+  void CheckBackwardOutputs(OpBase* op);
+
+  void SumGradient(OpBase* op, std::shared_ptr<VarBase> src, VarBase* dst) {
```
JiabinYang: why `share_ptr` here?

python/paddle/fluid/framework.py
```diff
@@ -981,8 +981,9 @@ def __init__(self,
             if type is None:
                 raise ValueError(
                     "`type` to initialized an Operator can not be None.")
-            self.iop = core.OpBase(type)
```
JiabinYang, line 984: no need for clear op then

paddle/fluid/imperative/layer.cc
sneaxiy: `RunOp` has been moved to another method of `OpBase`,  and it is called in `Tracer`.

paddle/fluid/imperative/layer.cc
sneaxiy: `TraceBackward` has moved to anther method of `OpBase`, and it is called in `Tracer`.

paddle/fluid/imperative/layer.h
sneaxiy: Done.

paddle/fluid/imperative/layer.cc
sneaxiy: Not suitable to do it, because not all ops need `name_to_var`.

paddle/fluid/imperative/layer.cc
sneaxiy: Not suitable to do it, because not all ops need name_to_var.

paddle/fluid/imperative/layer.cc
sneaxiy: Not needed. Maybe it is confused to set `shape` in `VarBase` because `VarBase` is not always a `LoDTensor`. It may be a `SelectedRows` in the future, and `SelectedRows` does not support `shape`.

paddle/fluid/imperative/layer.cc
sneaxiy: Each var should have its own accumulator. Accumulator is designed to unify the codes when `sorted_sum_gradient_=false` and `sorted_sum_gradient=true`.

paddle/fluid/imperative/layer.cc
sneaxiy: Because `Tracer` needs the input type to be `std::shared_ptr`.

paddle/fluid/imperative/layer.cc
sneaxiy: Because `Tracer` needs the input type to be `std::shared_ptr`.

paddle/fluid/imperative/layer.cc
sneaxiy: Because `Tracer` needs the input type to be `std::shared_ptr`.

python/paddle/fluid/framework.py
sneaxiy: Yes. But I keep clear ops now, because some function may still create backward ops, like gradient clipping, optimize.minimize.

PR #17720, opened 2019-05-29
digraph-load: Use tensor.set instead of `=` to restore parameters

junjun315:
Fix potential bugs: use numpy assignment to restore parameters at build-once


phlrain (approved):
LGTM

PR #17736, opened 2019-05-29
Correct Role Enum for role_maker, test=develop.

lzha106:
For Enum, it should not contain comma in its member defination.


seiriosPlus:
please add a new commit with the message "Correct Role Enum for role_maker, test=develop".

PR #17737, opened 2019-05-29
Add input format in Transpose GetHash

lidanqing-vv:
Add input format in Transpose GetHash, resolving transpose quantization bug. With this fix, mobilenet-ssd INT8 tester will work.

### DIFF
--- a/paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
@@ -124,8 +124,7 @@ class TransposeMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
         paddle::framework::vectorize2int(out_grad->dims());
 
     const std::string key = platform::TransposeMKLDNNHandler::GetHash(
-        nchw_tz, axis, ctx.op().Output(framework::GradVarName("X")) +
-                           std::to_string(out_grad->format()));
+        nchw_tz, axis, ctx.op().Output(framework::GradVarName("X")));
 
     platform::TransposeMKLDNNHandler handler(nchw_tz, reversed_axis, dev_ctx,
                                              mkldnn_engine, key);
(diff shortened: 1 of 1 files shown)


wojtuss (approved):

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
```diff
@@ -124,8 +124,7 @@ class TransposeMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
         paddle::framework::vectorize2int(out_grad->dims());
 
     const std::string key = platform::TransposeMKLDNNHandler::GetHash(
-        nchw_tz, axis, ctx.op().Output(framework::GradVarName("X")) +
-                           std::to_string(out_grad->format()));
+        nchw_tz, axis, ctx.op().Output(framework::GradVarName("X")));
```
luotao1: Why remove the `out_grad->format()` in `TransposeMKLDNNGradOpKernel`?

lidanqing-vv:
Without adding input->format() into hash, INT8 `warmup stage` and `inference stage` may have same hashkey for `transpose` operator.

For example, when batch_size=warm_up_size=50, 
warm_up stage: `Transpose:    in:f32_nChw16c out:f32_blocked,num:1,50x12x19x19`
inference stage: `Transpose:     in:s8_nhwc out:f32_nchw,num:1,50x12x19x19`

paddle/fluid/operators/mkldnn/transpose_mkldnn_op.cc
lidanqing-vv: @luotao1  Hi! Because in INT8 inference, we don't use `TransposeMKLDNNGradOpKernel`(which is for backpropogation and training). @wojtuss suggests that we should not influence others' design. About why we add `input->format()`, please check following comment.

luotao1: Got it!

PR #17739, opened 2019-05-29
some ngraph updates to enable bert

baojun-nervana:
This PR includes a number of changes to enable BERT through ngraph:
1. set ngraph to support block 0 only, the subblock from condition_block op will run through paddle
2. Infershape was delayed before getting output dimension. 
    - reason for this change is some of the variables are reused in some reshape ops and it will impact the inputs when the inputs trying to pull the shape.
    - mul_op and pool2d were updated accordingly per this change
3. fix some corner cases
    - handle shape {0}
    - partial support of lookup_table_grad
4. remove duplicated output
   - this is an optimization
5. Handle reshape2_grad infershape
    Paddle reshape2 add a leading dimension 0 and chop it off in reshape_grad infershape. Ngraph would interpret 0 as 0 element, so it needs special handling in ngraph engine

### DIFF
--- a/paddle/fluid/operators/ngraph/ngraph_engine.cc
+++ b/paddle/fluid/operators/ngraph/ngraph_engine.cc
@@ -38,6 +38,10 @@ namespace operators {
 
 static ngraph::Shape Ddim2Shape(const framework::DDim& dims) {
   ngraph::Shape sp;
+  if (dims.size() == 1 && dims[0] == 0) {
+    sp.emplace_back(0);
+    return sp;
+  }
   for (int i = 0; i < dims.size(); ++i) {
     int k = dims[i];
     k = k == 0 ? 1 : k;
--- a/paddle/fluid/operators/ngraph/ngraph_engine.cc
+++ b/paddle/fluid/operators/ngraph/ngraph_engine.cc
@@ -421,6 +421,15 @@ void NgraphEngine::BuildNgIO(const std::vector<framework::OpDesc*>& ops_desc,
       }
     }
   }
+  // remove output duplicates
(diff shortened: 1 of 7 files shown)


mozga-intel (approved):
LGTM!

paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -626,6 +631,21 @@ void NgraphEngine::Run(const framework::Scope& scope,
     }
   }
 
+  for (auto& op : fused_ops_) {
+    framework::RuntimeContext ctx(op->Inputs(), op->Outputs(), scope_);
+    if (op->Type() == "reshape2_grad") {
```
tensor-tang, line 636: why check reshape2 type in `Run` ? what if we have more ? transpose2 or others?

baojun-nervana: reshape op adds a leading dimension of 0. https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/reshape_op.cc#L261

And chop it off in its grad op, https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/reshape_op.cc#L311.

In paddle the dimension 0 means placeholder or removable dimension, however ngraph will check dimension product and 0 will make ngraph treat the tensor as 0 element. So we have to process reshape2_grad op infershape differently. Doing this in ngraph engine can avoid modifying the original paddle file.

LeoZhao-BJ:
LGTM

tensor-tang (approved):
LGTM

PR #17743, opened 2019-05-30
fix prepare context redundant code problem, optimize executor by cach…

guru4elephant:
…ing create_varaiables

test=develop

optimize executor by caching create_variable results. User only need to open use_program_cache to use this optimization. 

Take Bert model as example:

Before optimization:
image

After optimization：
image

image

This optimization is generalized that all models training can be applied.

### DIFF
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -943,6 +943,8 @@ All parameter, weight, gradient are variables in Paddle.
            })
       .def("prepare_ctx_cache", &Executor::PrepareCtxCache,
            py::call_guard<py::gil_scoped_release>())
+      .def("create_variables", &Executor::CreateVariables,
+           py::call_guard<py::gil_scoped_release>())
       .def("run", [](Executor &self, const ProgramDesc &prog, Scope *scope,
                      int block_id, bool create_local_scope, bool create_vars,
                      const std::vector<std::string> &fetch_vars) {
--- a/python/paddle/fluid/executor.py
+++ b/python/paddle/fluid/executor.py
@@ -366,6 +366,12 @@ class Executor(object):
         self._default_executor = core.Executor(p)
         self._closed = False
 
+    def _get_var_cache(self, program_cache_key):
+        return self.var_caches.get(program_cache_key, None)
+
+    def _get_scope_cache(self, program_cache_key):
+        return self.scope_caches.get(program_cache_key, None)
+
     def _get_ctx_cache(self, program_cache_key):
         return self.ctx_caches.get(program_cache_key, None)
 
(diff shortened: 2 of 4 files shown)


python/paddle/fluid/executor.py
```diff
@@ -719,7 +740,8 @@ def _run(self, program, exe, feed, fetch_list, feed_var_name,
         if not use_program_cache:
             exe.run(program.desc, scope, 0, True, True, fetch_var_name)
         else:
-            exe.run_cached_prepared_ctx(ctx, scope, True, True, False)
+            #exe.run_cached_prepared_ctx(ctx, scope, True, True, False)
```
gavin1332: remove this line

gavin1332 (changes requested):
a commented line need to be removed

python/paddle/fluid/executor.py
guru4elephant: fixed, thx

gavin1332 (dismissed):

gavin1332 (dismissed):

python/paddle/fluid/executor.py
```diff
@@ -366,6 +366,12 @@ def __init__(self, place):
         self._default_executor = core.Executor(p)
         self._closed = False
 
+    def _get_var_cache(self, program_cache_key):
+        return self.var_caches.get(program_cache_key, None)
```
gavin1332, line 372: missing definition of self.var_caches

python/paddle/fluid/executor.py
```diff
@@ -366,6 +366,12 @@ def __init__(self, place):
...
+
+    def _get_scope_cache(self, program_cache_key):
+        return self.scope_caches.get(program_cache_key, None)
```
gavin1332, line 375: missing definition of self.scope_caches

python/paddle/fluid/executor.py
```diff
@@ -719,7 +744,7 @@ def _run(self, program, exe, feed, fetch_list, feed_var_name,
         if not use_program_cache:
             exe.run(program.desc, scope, 0, True, True, fetch_var_name)
         else:
-            exe.run_cached_prepared_ctx(ctx, scope, True, True, False)
+            exe.run_cached_prepared_ctx(ctx, scope, False, False, False)
```
gavin1332: as executor.run() method allows user inputting None scope, it should be set scope to default when it is None

python/paddle/fluid/executor.py
guru4elephant: fixed. thx

python/paddle/fluid/executor.py
guru4elephant: fixed.

paddle/fluid/pybind/pybind.cc
```diff
@@ -943,6 +943,8 @@ All parameter, weight, gradient are variables in Paddle.
            })
       .def("prepare_ctx_cache", &Executor::PrepareCtxCache,
            py::call_guard<py::gil_scoped_release>())
+      .def("create_variables", &Executor::CreateVariables,
```
jacquesqiao, line 946: CreateVariables has no return value.

gavin1332 (approved):

paddle/fluid/pybind/pybind.cc
guru4elephant: will remove this later, thx

PR #17767, opened 2019-05-31
op lite compile bug fix

lijianshe02:
Replace boost related function with general GetAttr function for moble compatible  test=develop


Superjomn (approved):
LGTM

PR #17768, opened 2019-05-31
Using Smart pointer to optimizer memory usage of dyGraph

JiabinYang:
This PR use smart pointer to hold grads and release the fwd vars. In this way, we can save lots of memory:

Test on [Resnet](https://github.com/PaddlePaddle/models/tree/develop/dygraph/resnet), with Tesla K40m(Mem 11439MiB):

Without smart pointer : max batch size -> 56
With smart pointer: max batch size -> 99

And We test the convergence of several models under `models/dygraph` the result is as follow:

Mnist:
Mnist
ptbRNN:
ptb
reinforcement learning:
rl
transformer:
transformer

### DIFF
--- a/paddle/fluid/imperative/layer.cc
+++ b/paddle/fluid/imperative/layer.cc
@@ -14,6 +14,7 @@
 
 #include "paddle/fluid/imperative/layer.h"
 
+#include <algorithm>
 #include <deque>
 #include <limits>
 #include <map>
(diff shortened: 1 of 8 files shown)


JiabinYang:
This Commit will enhance the performance by 15-20%:

The test result is based on Tesla K40m 

[PTB_model](https://github.com/PaddlePaddle/models/tree/develop/dygraph/pbt_lm):
before: 82.2777 s per 1/10 of all data
after: 62.12738 s per 1/10 of  all data

[Resnet](https://github.com/PaddlePaddle/models/tree/develop/dygraph/resnet).     Batch size 50
before: 1.49s per batch
after: 1.26s per batch

sneaxiy (approved):
LGTM.

PR #17786, opened 2019-06-02
fix logging basicConfig cannot be setting after import paddle 

heavengate:
**Bug:** logging.basicConfig can not be settting successfully after import paddle
**Reason:** logging.basicConfig has been set in paddle python code
**Fix:** add log_helper.py get_logger function to get logger with given name, level and format without setting logging.basicConfig


SunGaofeng (dismissed):
It is really hard work to modify so many files to fix this problem.
However, is it necessary to add some instructions or illustrations in develop guid, in case that someone else add that again? You may suggest the administrator or PMs to add this.

SunGaofeng (commented):
using "grep -n basicConfig -r . "  under  python/paddle, it shows that 17 files include basicConfig, in the pr, 12 of them were fixed, please check again.

heavengate:
> using "grep -n basicConfig -r . " under python/paddle, it shows that 17 files include basicConfig, in the pr, 12 of them were fixed, please check again.

Done for change remain file to use get_logger

qingqing01 (approved):

PR #17787, opened 2019-06-02
Make ParallelExecutor support Windows GPU

wopeizl:
Enable parallel_executor on windows to support single GPU.
Note: nccl is NOT official supported by Nvidia and then paddle has no support to multiple GPU yet.


paddle/fluid/framework/parallel_executor.cc
```diff
@@ -653,11 +656,20 @@ bool ParallelExecutor::EnableParallelGraphExecution(
...
+        exec_strategy.type_ == ExecutionStrategy::ExecutorType::kExperimental) {
       enable_parallel_graph = false;
+    }
+  }
+
+#ifdef WIN32
+  LOG_FIRST_N(WARNING, 1)
+      << "Windows has no support to parallel graph, enable_parallel_graph "
+         "would be forced to false.";
+  enable_parallel_graph = false;
```
chengduoZH, line 669: Why use `LOG_FIRST_N ` here?

paddle/fluid/pybind/pybind.cc
```diff
@@ -1339,6 +1339,14 @@ All parameter, weight, gradient are variables in Paddle.
           "num_trainers",
           [](const BuildStrategy &self) { return self.num_trainers_; },
           [](BuildStrategy &self, int num_trainers) {
+#ifdef WIN32
+            if (std::abs(num_trainers) > 1) {
+              LOG_FIRST_N(WARNING, 1)
+                  << "Windows has no support to distribute mode, "
+                     "num_trainers would be forced to 1.";
+              num_trainers = 1;
```
chengduoZH: Now that Windows doesn't support distribution training, the program should be exited.

wopeizl: ok

paddle/fluid/framework/parallel_executor.cc
wopeizl: use vlog instead

chengduoZH (approved):

PR #17788, opened 2019-06-02
fix: when use the "load model from memory" mode, the RAM occupy is high

NHZlX:
fix two things when using the paddle c++ inference api::

1 . trt cpu memory occupy to high.

2. "load model from memory " occupy to much memory.

### DIFF
--- a/paddle/fluid/inference/analysis/argument.h
+++ b/paddle/fluid/inference/analysis/argument.h
@@ -63,6 +63,16 @@ struct Argument {
   using anakin_max_shape_t = std::map<std::string, std::vector<int>>;
 
   bool Has(const std::string& key) const { return valid_fields_.count(key); }
+  void PartiallyRelease() {
+    if (Has("model_program_path")) {
+      if (Has("model_from_memory") && model_from_memory()) {
+        model_program_path().clear();
+        model_program_path().shrink_to_fit();
+        model_params_path().clear();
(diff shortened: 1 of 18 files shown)


paddle/fluid/inference/tensorrt/engine.h
```diff
@@ -213,6 +219,39 @@ class TensorRTEngine {
...
+                         TRTInt8Calibrator* calibrator = nullptr,
+                         int device_id = 0,
+                         nvinfer1::ILogger& logger = NaiveLogger::Global()) {
+    auto* p = new TensorRTEngine(max_batch, max_workspace, enable_int8,
+                                 calibrator, device_id, logger);
+    engines_[name].reset(p);
+    return p;
+  }
+
+  void DeleteALl() {
```
Superjomn: All

paddle/fluid/inference/analysis/argument.h
```diff
@@ -63,6 +63,16 @@ struct Argument {
   using anakin_max_shape_t = std::map<std::string, std::vector<int>>;
 
   bool Has(const std::string& key) const { return valid_fields_.count(key); }
+  void ReleasePart() {
```
Superjomn: PartiallyRelease()

Shixiaowei02 (approved):

PR #17795, opened 2019-06-03
use the bj as default address instead of cdn test=develop

wopeizl:
The cdn.bcebos.com can be unstable randomly for unknown reason, change it to bj.bcebos.com can fix that.


junjun315 (approved):
LGTM

PR #17801, opened 2019-06-03
make omp thread num default 1 after inference run

luotao1:
```
bool AnalysisPredictor::ZeroCopyRun() {
   paddle::platform::SetNumThreads(config_.cpu_math_library_num_threads());
   executor_->Run();
   ...
}
```
We `SetNumThreads` before ` executor_->Run();` in both `AnalysisPredictor::ZeroCopyRun()` and `AnalysisPredictor::Run()`. 

- Thus, for example, if `config_.cpu_math_library_num_threads()=8`, then, after we call `ZeroCopyRun()`, the omp threads number become 8 in environment. 
- In some deployment service, the change of omp threads to 8 will cause omp threads confict and affect performance with other modules. 
- So we  recover the cpu_math_library_num_threads to 1 to avoid thread conflict when integrating it into deployment service.


Superjomn (approved):
LGTM

PR #17804, opened 2019-06-03
fix api doc in slice op, test=develop

jerrywgz:
Fixed API doc. In slice op, input 'axes' must have the same length with 'starts' and 'ends', and it can not be omitted.
image
image


xsrobin (approved):
LGTM

XiaoguangHu01 (approved):

shanyi15 (approved):

PR #17814, opened 2019-06-04
Added unit test for QAT FP32 & INT8 comparison

wojtuss:
test=develop

This PR contains:
- fix for MKL-DNN `pool2d` op kernel reusing,
- unit test for QAT FP32 & INT8 inference output comparison.

The unit test is enabled for the following QAT pretrained models:
* ResNet50
* ResNet101
* GoogleNet
* MobileNetV1
* MobileNetV2
* VGG16
* VGG19
The unit test has `FLAGS_use_mkldnn` set to `true`.

The test is not enabled on CI as it fails on SKX machines due to s8s8s32 gemm issue in MKL-DNN: https://github.com/intel/mkl-dnn/issues/476. The issue is not present on CLX machines.
It can be run manually, via e.g.
```
ctest -R test_qat_int8_resnet50_mkldnn
```

This test requires the `TransformForMkldnnPass` from the PR https://github.com/PaddlePaddle/Paddle/pull/17819.


sfraczek (dismissed):
LGTM

wojtuss:
@luotao1 , the unusual build `PR_CI_Document_Preview (Compile)` timeouts again and again. The timeout seems not related to the PR contents. Could you please help resolve it?

luotao1:
`PR_CI_Document_Preview (Compile)` is not `Required` (you can see it doesn't have `Required` tag), thus, you can ignore it.

luotao1:
Does #17884 resolve the fail on SKX machines?

wojtuss:
> Does #17884 resolve the fail on SKX machines?

@luotao1 , no, the accuracy on SKX is better, but still not within 0.01 difference range

wojtuss:
@luotao1 , the tests `test_qat_int8_<model>` are not supposed to be run on CI (on the SKX like machines they fail), but should be available for running by hand. How can I disable them for CI only?

luotao1:
@wojtuss How about add a cmake flag like
https://github.com/PaddlePaddle/Paddle/blob/993c703bcc082e7a2b91d643471e93399159ec57/python/paddle/fluid/contrib/slim/tests/CMakeLists.txt#L32-L34

luotao1:
If you don't add any unit-tests for QAT FP32&INT8 comparison, how could we ensure the correctness of related codes?
> but still not within 0.01 difference range

Could the difference range become larger than 0.01 for above unit-tests on our CI machines?

wojtuss:
@luotao1 
Following your suggestions, I would:
1. increase the accuracy diff threshold to 0.1, so the tests could pass on CI (the `--acc_diff_threshold` option still allows to set the threshold by hand),
2. add the `WITH_SLIM_MKLDNN_FULL_TEST` flag check wrapping all but two test cases, i.e. two test cases (googlenet and mobilenetv1) would be run without this flag.
What do you think about it?

luotao1:
> increase the accuracy diff threshold to 0.1

Is 0.1 too big? maybe 0.02?

> so the tests could pass on CI (the --acc_diff_threshold option still allows to set the threshold by hand),

I think it's OK.

> add the `WITH_SLIM_MKLDNN_FULL_TEST` flag check wrapping all but two test cases, i.e. two test cases (googlenet and mobilenetv1) would be run without this flag.

We need to test all the QAT unit-tests here. For slim tests, we have the similar C++ unit-tests in INT8v2.

Besides, since #17833 and #17819 are merged, please remove the related files in this PR.

wojtuss:
@luotao1 
1. The 0.1 threshold is the minimum for the tests to pass on the small dataset.
2. We can't use the `quantized_accuracy`, as the test is in python, and not in C-API.
3. Do you mean not to use the `WITH_SLIM_MKLDNN_FULL_TEST` at all?

luotao1:
> Do you mean not to use the `WITH_SLIM_MKLDNN_FULL_TEST` at all

For slim tests, since CI runs slower than INT8 V2, we add `WITH_SLIM_MKLDNN_FULL_TEST` at first. If slim tests run faster, you can remove this flag. Is #17903 related with it?

wojtuss:
These tests last about 2x longer than C-API IN8v2. I will leave only two without this flag, OK?

No. https://github.com/PaddlePaddle/Paddle/issues/17903 is not related to this test as we run.

luotao1:
If you use omp_threads=4, it still 2x longer than C-API IN8v2?

wojtuss:
yes, because in C-API optimized models are being run, in QAT the models are not optimized using fuses

luotao1:
Got it. But I see not all the unit-tests cause a lot of time. Leaving 2 is small. How about leaving unit-tests less than 100 sec?
```
154/159 Test #661: test_qat_int8_googlenet_mkldnn ..................***Failed   32.57 sec
157/158 Test #659: test_qat_int8_resnet50_mkldnn .............***Failed   85.77 sec
157/158 Test #660: test_qat_int8_resnet101_mkldnn ..............   Passed  153.29 sec
[10:47:24]158/158 Test #664: test_qat_int8_vgg16_mkldnn ..................   Passed  165.27 sec
155/159 Test #662: test_qat_int8_mobilenetv1_mkldnn .............***Failed   46.00 sec
153/158 Test #663: test_qat_int8_mobilenetv2_mkldnn ..........***Failed   56.79 sec
159/159 Test #665: test_qat_int8_vgg19_mkldnn ......................   Passed  182.97 sec
```

wojtuss:
OK

luotao1:
If you want to use `WITH_SLIM_MKLDNN_FULL_TEST`, how about change to `WITH_INT8_MKLDNN_FULL_TEST`?

wojtuss:
> If you want to use `WITH_SLIM_MKLDNN_FULL_TEST`, how about change to `WITH_INT8_MKLDNN_FULL_TEST`?

It is up to you. I do not know which is more suitable here. I have set `WITH_SLIM_MKLDNN_FULL_TEST` for now. OK?

luotao1:
That's OK.

wojtuss:
@luotao1 ,
The tests `test_qat_int8_<model>` fail on CI due to missing directories with models:
```
[15:56:17]	Traceback (most recent call last):
[15:56:17]	  File "/paddle/python/paddle/fluid/contrib/slim/tests/qat_int8_comparison.py", line 272, in test_graph_transformation
[15:56:17]	    transform_to_int8=False)
[15:56:17]	  File "/paddle/python/paddle/fluid/contrib/slim/tests/qat_int8_comparison.py", line 163, in _predict
[15:56:17]	    model_path, exe, 'model', 'params')
[15:56:17]	  File "/paddle/build/python/paddle/fluid/io.py", line 1166, in load_inference_model
[15:56:17]	    raise ValueError("There is no directory named '%s'", dirname)
[15:56:17]	ValueError: ("There is no directory named '%s'", '/root/.cache/inference_demo/qat/GoogleNet/model')
```
However, locally the tests have no problem with the directories. Could you please help resolve the issue on CI?
The paths we are using are defined in `python/paddle/fluid/contrib/slim/tests/CMakeLists.txt`.

luotao1 (approved):
LGTM

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
```diff
@@ -11,6 +11,18 @@ function(inference_analysis_python_api_int8_test target model_dir data_dir filen
              --batch_size 50)
 endfunction()
 
+function(inference_qat_int8_test target model_dir data_dir test_script use_mkldnn)
+    py_test(${target} SRCS ${test_script}
+            ENVS FLAGS_OMP_NUM_THREADS=${CPU_NUM_THREADS_ON_CI}
+                 OMP_NUM_THREADS=${CPU_NUM_THREADS_ON_CI}
+                 FLAGS_use_mkldnn=${use_mkldnn}
```
luotao1, line 18: why use `FLAGS_OMP_NUM_THREADS` and `OMP_NUM_THREADS` at the same time? 
@wojtuss

PR #17819, opened 2019-06-04
QAT int8 MKL-DNN transformation pass

bingyanghuang:
This PR is to convert QuantizationFreezePass generated IrGraph to MKL-DNN support INT8
runnable IrGraph. 
Following transformations have been done in this pass:
<ol>
   1. Convert int8 range weights with float type, (which generated by the QuantizationFreezePass), to fp32 range weights with float dtype by the corresponding scales. <br>
    2. Create the new conv2d op with the converted weights and link its output to fake_dequantize_abs_max output and set conv2d's attribute "force_fp32_output" as true <br>
    3. Transform fake_quantize_xx to quantize op <br>
    4. Remove fake_dequantize_abs_max op <br>
</ol>
Unit test is testing based on the Graph already applied QuantizationTransformPass and QuantizationFreezePass, it will check if:
<ol>
    1. conv2d's output is rightly linked to the fake_dequantize op's output <br>
    2. conv2d's weights has been converted to fp32 range, they are not integer any more <br>
    3. check the graph locally to make sure that op's type have been transformed as we expected. <br>
</ol>
The latest accuracy of full ImageNet validation  set on Cascade Lake (new generation Xeon) is as below:

image


sfraczek (changes requested):
We should have checking somehwere if mkldnn is enabled (i.e. FLAGS_use_mkldnn==1). Currently this transformation happens even if FLAGS_use_mkldnn=false when I run test written by @wojtuss https://github.com/PaddlePaddle/Paddle/pull/17814.

(This transformation sets attribute `use_mkldnn=true` for some operators which causes execution of mkldnn operators)

python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # Based on the QAT scale value to calculate the scales for MKL-DNN INT8 conv2d
+        scale_in = self.s8_max / self.InScale[output_name]
+        scale_w = []
+        scale_w.append(self.max_range[output_name] / self.s8_max)
+
+        conv_op_node.set_attr("Scale_weights", scale_w)
+        conv_op_node.set_attr("Scale_in", scale_in)
+        conv_op_node.set_attr("Scale_out", 1.0)
+        conv_op_node.set_attr("use_mkldnn", 1)
```
wojtuss, line 165: The `use_mkldnn` attribute value should be copied from the original `conv2d` op.

wojtuss: On the other hand, this should be called only with MKL-DNN, so it can be left as it is.

sfraczek (dismissed):
LGTM. the control whether to run this or not based on use_mkldnn can be added somewhere else

bingyanghuang:
@luotao1 @wzzju This PR has been reviewed inside intel. And has been passed CI once. Could you help review it?

python/paddle/fluid/contrib/slim/tests/test_transform_to_mkldnn_pass.py
```diff
@@ -0,0 +1,194 @@
...
+import six
+import paddle
+from paddle.fluid.framework import IrGraph
+from paddle.fluid.contrib.slim.quantization import QuantizationFreezePass
+from paddle.fluid.contrib.slim.quantization import QuantizationTransformPass
+from paddle.fluid.contrib.slim.quantization import TransformForMkldnnPass
+from paddle.fluid import core
+
+os.environ["CUDA_VISIBLE_DEVICES"] = "0"
+os.environ["CPU_NUM"] = "1"
```
luotao1: Why use these two flags?

python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
```diff
@@ -0,0 +1,227 @@
...
+from .... import core
+from ....framework import IrGraph
+from ....framework import IrNode
+
+__all__ = ['TransformForMkldnnPass']
+
+
+class TransformForMkldnnPass(object):
+    def __init__(self, scope=None, place=None):
+        """
```
luotao1, line 40: These comments are not for `__init__` function, they are for `class TransformForMkldnnPass`. Thus, move them after line 23?

luotao1 (commented):
Review is not finished.

python/paddle/fluid/contrib/slim/tests/test_transform_to_mkldnn_pass.py
bingyanghuang: In this test, we need to train a model for several iterations to get the weights value. And then converted to int8 and check if it is converted rightly. When we train this model, we need to set "CPU_NUM". This train model idea is following below example: https://github.com/PaddlePaddle/Paddle/blob/c1379bf2387175937be325aeabf2887c31c38c22/python/paddle/fluid/contrib/slim/tests/test_quantization_pass.py#L295

python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
bingyanghuang: Agree. Thanks

bingyanghuang: Done

luotao1 (approved):
LGTM! @wzzju If you have any suggestions, please feel free to comment it.

Issue #17611, opened 2019-05-24
exception occurs when do BERT training on CPU with transpose mkldnn op enabled


LeoZhao-BJ:
I met a issue related with mkldnn when I tried to enable transpose mkldnn op in ParallelExecutor for BERT training. 

throw exception in TransposeMKLDNNHandler::  AcquireTranspose()
```
PADDLE_ENFORCE((transpose_p != nullptr) || (is_reusing_ == false),
                   "Fail to find convolution primitive in device context");
```

For current implementation in Compute(), a hash key is used for create TransposeMKLDNNHandler, and this key is generated by
```
const std::string key = platform::TransposeMKLDNNHandler::GetHash(
        nchw_tz, axis, ctx.op().Output(framework::GradVarName("X")));
```
This method is ok for common case in Executor or NativeExecutor since the only instance for one op, and even for multi-instances inference case, SetMkldnnThreadID() is designed for identifying op.
But for ParallelExecutor case, it is specific, it duplicates program execution by user defined instance number, it looks like you have multiple GPU cards in one machine. Let's come to the root cause for my case, for ParallelExecutor on CPU, no place to SetMkldnnThreadID() to identify different mkldnn op instances on same program, simply speaking, the key generation method is not enough anymore.
For my current temporary solution, I add kernel instance pointer 'this' into hash key, so it can be simplified with just 'this', and then it works.
I think it is not a issue just on Transpose, it may be a generic issue for all mkldnn ops on ParallelExecutor.

LeoZhao-BJ:
@jczaja

luotao1:
We meet the same problem on content_dnn model.  We use parallel_executor for training. 
- We enable transpose2/transpose2_grad mkldnn op for one thread training, it runs OK.
- We enable transpose2/transpose2_grad mkldnn op for 16 threads training, it runs Crash.
- We disable transpose2/transpose2_grad mkldnn op for 16 threads training, it runs OK.

Thus, we guess is there something wrong in transpose mkldnn op?

> For my current temporary solution, I add kernel instance pointer 'this' into hash key, so it can be simplified with just 'this', and then it works

@LeoZhao-Intel Could you give the `git diff` for your update? And we can test it for content_dnn at the same time.

LeoZhao-BJ:
A temporary fix, just for test. 

```
[...]

PR #17826, opened 2019-06-04
enable mkldnn primitive reuse for platform reorder

LeoZhao-BJ:
Use mkldnn handler to reuse primitives in reordering to improve performance.
Key is generated by:
input_dims + input_mkldnn_format + output_mkldnn_format + input_mkldnn_type

test=develop

### DIFF
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -13,11 +13,13 @@
 // limitations under the License.
 
 #include "paddle/fluid/framework/data_layout_transform.h"
+#include <string>
 #include <vector>
 
 #include "paddle/fluid/operators/math/math_function.h"
 #ifdef PADDLE_WITH_MKLDNN
 #include "paddle/fluid/platform/mkldnn_helper.h"
+#include "paddle/fluid/platform/mkldnn_reuse.h"
 #endif
 
 namespace paddle {
@@ -145,7 +147,6 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
   memory::data_type in_type = ToMKLDNNDataType(in.type());
   PADDLE_ENFORCE(in_type != memory::data_type::data_undef,
                  "Input tensor type is not supported: %s", in.type());
-  memory::data_type out_type = in_type;
 
   auto in_format = platform::MKLDNNFormatForSize(in_tz.size(), in.format());
   auto out_format =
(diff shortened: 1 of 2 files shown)


LeoZhao-BJ:
@jianhang-liu @luotao1 
This patch is for detect model performance improvement

LeoZhao-BJ:
This is data before change (without this PR and bilinear_interp  omp optimization):
image
This is data after change (code base includes bilinear_interp  omp optimization):
image

bilinear_interp and split op execution time are reduced due to reordering time counted into op time and reuse shorting this time.

LeoZhao-BJ:
@jczaja @yihuaxu @kbinias for code review

jczaja (approved):
LGTM

paddle/fluid/framework/data_layout_transform.cc
```diff
@@ -156,14 +157,21 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
 
   if (in_format != out_format) {
     void* in_data = GetDataFromTensor(in, in_type);
-    auto out_data = out->mutable_data(expected_kernel_type.place_, in.type());
+    const std::string key = platform::ReorderMKLDNNHandler::GetHash(
+        in_tz, in_format, out_format, std::to_string(in_type));
```
ghost, line 161: Is this "key" good enough to identify two different reorder? Assume two reorder OP in one graph have exactly same dim/format/type for in and out. Then they will use same "key" value. 
We should use something unique, e.g. var name of output variable

LeoZhao-BJ: Despite of race condition, the key is enough, this primitive can be reused for different input/out variables since it will update internal mkldnn memory data handle(buffer ptr) for each call.  see https://github.com/PaddlePaddle/Paddle/pull/17826/files#diff-a60a81d3c97ec5a7cf7cf3d7e1aa9adbR428

But for multi-instance case, it may introduce race condition issue as Jacek is working on another Issue #17611. 
For this line, in_data was added into key to reduce potential race condition.

luotao1:
I test this PR on other internal face models, it seems no speedup.

jczaja:
@luotao1 Speedup introduced by that one will depend alot on model design eg. Those changes will speedup model that is having lots of transitions from MKL-DNN tensors to Paddle Tensors. If model is entirly Paddle ops based or MKL-DNN based then speedup may not be noticible. Moreoever due to specific design of MKL-DNN Transpose (output of MKL-DNN Transpose is paddle tensor, not MKL-DNN tensor)op also models like BERT (I tested performance impact of reorder some time ago) will not benefit from this one.

PR #17834, opened 2019-06-04
Add mkldnn int8 mul-op kernel

Minghui-Intel:
Add MKLDNN INT8 Mul-Op Kernel for CPU place. This new kernel support: 
1. s8 type input multiplied with fp32 type weights;
2. u8 type input multiplied with fp32 type weights;


paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+using mkldnn::inner_product_forward;
+using mkldnn::memory;
+using mkldnn::primitive;
+using mkldnn::prop_kind;
+using mkldnn::stream;
+using platform::GetMKLDNNFormat;
+using platform::MKLDNNDeviceContext;
+using platform::to_void_cast;
+
+template <typename T, typename K>
```
Sand3r-: I do believe that for nonquantized mul primitive, only a single type argument is necessary.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+    weights_ = TransposeWeights(&y_matrix);
+
+    auto dst_desc = CreateMemDescriptor(output, memory::format::any);
+
+    fc_ = CreateFcPrimitive(*input_, *weights_, dst_desc, output, ctx);
+    return *fc_;
+  }
+
+ protected:
+  void UpdateDataPointers(const ExecutionContext& ctx, Tensor* out,
```
Sand3r-: I do believe that y input ptr should updated as well.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+
+    if (fc_) {
+      UpdateDataPointers(ctx, output, &x_matrix);
+      return *fc_;
+    }
+
+    auto src_desc =
+        CreateMemDescriptor(&x_matrix, x_dt, mkldnn::memory::format::nc);
+    input_ = CreateMemory(src_desc, &x_matrix);
+    weights_ = TransposeWeights(&y_matrix);
```
Sand3r-: Why is the second input to mul called "weights"?

Sand3r-: If necessary, the second input "y" to mul should have its data transposed before the fc from lines  68-71 is returned. Otherwise if the data from y change, they are treated as if they were transposed in returned `*fc_` primitive, while in fact they weren't.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+    const auto src_desc = src_memory.get_primitive_desc().desc();
+
+    auto fc_prim_desc = CreateFcPrimDesc(src_desc, weights_desc, dst_desc);
+    output_ = this->CreateDstMemory(fc_prim_desc, ctx, output);
+
+    return inner_product_forward(fc_prim_desc, src_memory, weights_memory,
+                                 *output_);
+  }
+
+  mkldnn::inner_product_forward::primitive_desc CreateFcPrimDesc(
```
Sand3r-: I am doubtful whether this function is necessary.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+                                               weights_desc, dst_desc);
+
+    return inner_product_forward::primitive_desc(fc_desc, engine_);
+  }
+
+  mkldnn::inner_product_forward::primitive_desc CreateFcPrimDesc(
+      const mkldnn::memory::desc& input_desc,
+      const mkldnn::memory::desc& weights_desc,
+      const mkldnn::memory::desc& dst_desc,
+      const mkldnn::primitive_attr& fc_attr) {
```
Sand3r-: This function should be moved to Quantized Factory

Sand3r- (commented):
Currently, QuantMulPrimitiveFactory in two places uses the same mechanisms to figure out the output type (checking force_fp32_output and so on). Perhaps the output type could be determined upfront in GetPrimitiveFactory and passed as a template argument when creating a factory.

MulPrimitiveFactory and QuantMulPrimitiveFactory  share quite amount of common functionalities. If the time allows I suggest combining them into a single class.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+ public:
+  using MulPrimitiveFactory<T, K>::MulPrimitiveFactory;
+
+  virtual inner_product_forward CreateMulPrimitive(
+      const Tensor* input, const Tensor* weights, Tensor* output,
+      const ExecutionContext& ctx) {
+    /* check data format and reorder if need */
+    int x_num_col_dims = ctx.Attr<int>("x_num_col_dims");
+    int y_num_col_dims = ctx.Attr<int>("y_num_col_dims");
+    auto scale_weight = ctx.Attr<std::vector<float>>("scale_y");
```
Sand3r-: I just want to remind that this is not a fully-connected layer where we have an input and weights, but mul with x and y inputs that are subject to matrix multiplication. It might be helpful to adjust the namings accordingly.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+
+    auto output_dim = output->dims();
+    if (output_dim.size() != 2) {
+      (*output).Resize({input->dims()[0], weights->dims()[1]});
+    }
+
+    if (this->fc_) {
+      this->UpdateDataPointers(ctx, output, input);
+      return *(this->fc_);
+    }
```
Sand3r-, line 223: Once again - this part should come after updating of weights, or otherwise the input data from y won't be transposed and quantized.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+    auto dst_desc = this->CreateMemDescriptor(dims, type, memory::format::oi);
+    // MOD: return Reorder(src_desc, dst_desc, weights->data<T>());
+    return Reorder(src_desc, dst_desc, weights->data<K>());
+  }
+
+  mkldnn::memory UpdateWeights(mkldnn::memory weights,
+                               const std::vector<float>& scale_weight) {
+    auto dims = weights.get_primitive_desc().desc().data.dims;
+    auto ndims = weights.get_primitive_desc().desc().data.ndims;
+    auto weights_dims = std::vector<int>(dims, dims + ndims);
```
Sand3r-: Wouldn't it be sufficient here to just make a copy of weights desc and update the datatype of the destination memory descriptor ?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+                                                  Tensor* output,
+                                                  const ExecutionContext& ctx) {
+    const auto weights_desc = weights_memory.get_primitive_desc().desc();
+    const auto src_desc = src_memory.get_primitive_desc().desc();
+
+    mkldnn::primitive_attr fc_attr;
+
+    bool force_fp32_output = ctx.Attr<bool>("force_fp32_output");
+    auto scale_weights_data = ctx.Attr<std::vector<float>>("scale_y");
+    auto scale_in_data = ctx.Attr<float>("scale_x");
```
Sand3r-: Why does y input use a multichannel quantization while x only uses a single scale?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+
+    auto z_dim = z->dims();
+    z->set_layout(DataLayout::kMKLDNN);
+    z->set_format(z_dim.size() != 2
+                      ? platform::MKLDNNFormatForSize(z_dim.size(), x->format())
+                      : z->format());
+
+    if (z_dim.size() != 2) {
+      z->Resize(z_dim);
+    }
```
Sand3r-, line 414: Is it possible that output is not 2 dimensional?

paddle/fluid/operators/mul_op.h
```diff
@@ -110,95 +112,5 @@ class MulGradKernel : public framework::OpKernel<T> {
   }
 };
 
-template <typename DeviceContext, typename T>
-class MulDoubleGradKernel : public framework::OpKernel<T> {
```
Sand3r-: Why was this operator completely removed?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+    return platform::MKLDNNMemDesc(dims, platform::MKLDNNGetDataType<T>(),
+                                   format);
+  }
+
+  mkldnn::memory::desc CreateMemDescriptor(const std::vector<int>& dims,
+                                           memory::data_type type,
+                                           memory::format format) {
+    return platform::MKLDNNMemDesc(dims, type, format);
+  }
+
```
lidanqing-vv, line 114: In the above two CreateMemDescriptor. What about
```
mkldnn::memory::desc CreateMemDescriptor(const std::vector<int>& dims,
                                           memory::data_type type = platform::MKLDNNGetDataType<T>(),
                                           memory::format format)
````
or 
```
mkldnn::memory::desc CreateMemDescriptor(const std::vector<int>& dims,
                                           boost::optional<memory::format&> type,
                                           memory::format format)
```
and then pass `boost::none` as a parameter.  boost::optional is also used in [mkldnn_reuse.h](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/platform/mkldnn_reuse.h)

So one CreateMemDescriptor could be removed

File: paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc [hunk 13]
```diff
@@ -0,0 +1,497 @@
+/* Copyright (c) 2016 PaddlePaddle Authors. All Rights Reserved.
```
lidanqing-vv: Newly created files (created this year) will be `Copyright (c) 2019` and  modifying old files will keep old date, which we have confirmed with luotao

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,497 @@
...
+    return CreateMemDescriptor(dims, format);
+  }
+
+  mkldnn::memory::desc CreateMemDescriptor(const Tensor* tensor,
+                                           memory::data_type type,
+                                           memory::format format) {
+    auto dims = framework::vectorize2int(tensor->dims());
+    return CreateMemDescriptor(dims, type, format);
+  }
+
```
lidanqing-vv, line 129: Here we may also consider `boost::optional`.  Because I just feel less overloading functions will be easier to read.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: 1. the second input to mul called "weights" because of mul op implement share part of the code with fc op

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: for inference, y to mul will not change(just like weights for fc op)

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+#include "paddle/fluid/operators/mul_op.h"
+#include "paddle/fluid/platform/mkldnn_helper.h"
+#include "paddle/fluid/platform/mkldnn_reuse.h"
+
+namespace paddle {
+namespace operators {
+
+using framework::DataLayout;
+using Tensor = framework::Tensor;
+using framework::DataLayout;
```
sfraczek: This was already declared 2 lines above

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, I'll try to remove some unnecessary overloaded functions

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay~

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: this function is reserved for <fp32, fp32> mul

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: since y to mul won't change, so it's unnecessary to update y

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Sand3r-: According to documentation http://paddlepaddle.org/documentation/docs/zh/1.3/api/layers.html#mul
y input is just a regular tensor. 
Otherwise, what would be a point of having 2 operators (mul, fc) doing exactly the same thing?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, i will try to simplify the implementation of this function.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Sand3r-: The function for fp32 mul can be found here:
https://github.com/PaddlePaddle/Paddle/pull/17834/files/718ba9a24e31e93164efc70b523d35f7f9fb35e5#diff-11cb8bfd67488f87b2007c8ba7e91e44R198
This one contains a `bias_desc` which is unrelated to mul.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+using framework::DDim;
+using framework::ExecutionContext;
+using framework::Tensor;
```
sfraczek, line 29: already declared above

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+    dev_ctx.SetBlob(key, prim_creator);
+  }
+
+  return prim_creator;
+}
+
+template <typename T, typename K>
+class MulMKLDNNKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& ctx) const override {
```
sfraczek: don't need `framework::`

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+
+using framework::DataLayout;
+using Tensor = framework::Tensor;
+using framework::DataLayout;
+using framework::DDim;
+using framework::ExecutionContext;
+using framework::Tensor;
+using mkldnn::inner_product_forward;
+using mkldnn::memory;
+using mkldnn::primitive;
```
sfraczek: do you use this?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+
+template <typename T, typename K>
+class MulMKLDNNKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& ctx) const override {
+    PADDLE_ENFORCE(paddle::platform::is_cpu_place(ctx.GetPlace()),
+                   "It must use CPUPlace.");
+
+    auto& dev_ctx =
+        ctx.template device_context<paddle::platform::MKLDNNDeviceContext>();
```
sfraczek: you have `using platform::MKLDNNDeviceContext;` so you could remove paddle::platform

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+  virtual ~MulPrimitiveFactory() {}
+
+  virtual inner_product_forward CreateMulPrimitive(
+      const Tensor* input, const Tensor* weights, Tensor* output,
+      const ExecutionContext& ctx) {
+    /* check format and reorder if need */
+    int x_num_col_dims = ctx.Attr<int>("x_num_col_dims");
+    int y_num_col_dims = ctx.Attr<int>("y_num_col_dims");
+
+    memory::data_type x_dt = paddle::framework::ToMKLDNNDataType(input->type());
```
sfraczek: I think you don't need `paddle::` here

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+    memory::data_type y_dt =
+        paddle::framework::ToMKLDNNDataType(weights->type());
```
sfraczek: I think you don't need `paddle::` here

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+  }
+
+  return prim_creator;
+}
+
+template <typename T, typename K>
+class MulMKLDNNKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& ctx) const override {
+    PADDLE_ENFORCE(paddle::platform::is_cpu_place(ctx.GetPlace()),
```
sfraczek: I think you could remove `paddle::`, this is in paddle the namespace

sfraczek (changes requested):
small quick cosmetic review

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, I'll fix it

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, I'll fix it

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, I'll fix it

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, I'll fix it

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: as we discussed on skype, we need an independent MUL op because in QAT model MUL won't be fused to a FC op.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: yes, it's unnecessary, I will remove it

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: QuantMulPrimitiveFactory will use some member methods of MulPrimitiveFactory, so I think it'd better letting two template classes share the same type arguments

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: sure, i'll move it to Quantized Factory

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: no, the output should always be 2 dims, will remove it

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: I think that's how multichannel quantization designed.

paddle/fluid/operators/mul_op.h
Minghui-Intel: that's because int8 mul is only used for inference

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: sure, i will change it.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay~

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay~

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay~

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: <using mkldnn::primitive> is not used, will remove it

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, will adjust it

Minghui-Intel: I tried to modify the variable name from  "input" to "x" and "weight" to "y", but since that some code is shared from FC OP，so reserved the original variable name that in the shared part.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: done

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: done

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: I tried to simplify the code, it results in an exception while running the unit test,  tried to figure it out the reason but still don't know why for yet, since the deadline is coming, so I kept the original code for now.

Minghui-Intel: done

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: done

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: will refine the code as bug fix, keep the original code for now.

Minghui-Intel: done

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: will refine the code as bug fix, keep the original code for now.

Minghui-Intel: done

luotao1:
please use `start a review` to reply comments, see https://github.com/PaddlePaddle/Paddle/pull/17634#issuecomment-496386004

Minghui-Intel:
> please use `start a review` to reply comments, see [#17634 (comment)](https://github.com/PaddlePaddle/Paddle/pull/17634#issuecomment-496386004)

Sure, but there is no pending conversation for this PR, I'll pay attention to that next time. thanks for reminding.

paddle/fluid/operators/mul_op.cc
```diff
@@ -114,6 +144,23 @@ class MulOpMaker : public framework::OpProtoAndCheckerMaker {
...
+                                "Only used with MKL-DNN INT8")
+        .SetDefault({1.0f});
+    AddAttr<float>("scale_out",
+                   "scale_out to be used for int8 output data."
+                   "Only used with MKL-DNN INT8")
+        .SetDefault(1.0f);
+    AddAttr<bool>(
+        "force_fp32_output",
+        "(bool, default false) Force quantize kernel output FP32, only "
+        "used in quantized MKL-DNN.")
```
luotao1, line 162: Is  `scale_x`, `scale_y`, `scale_out` and `force_fp32_output` all needed?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+                   ops::MulMKLDNNKernel<uint8_t, float>);
+
+/** Unregister <fp32, fp32> MUL OP
+REGISTER_OP_KERNEL_WITH_CUSTOM_TYPE(mul, MKLDNN, ::paddle::platform::CPUPlace,
+                                    FP32, ops::kMULMKLDNNFP32,
+                                    ops::MulMKLDNNKernel<float, float>);
+
+REGISTER_OP_KERNEL(mul, MKLDNN, ::paddle::platform::CPUPlace,
+                   ops::MulMKLDNNKernel<float, float>);
+**/
```
luotao1: remove 530-537.

paddle/fluid/operators/mul_op.h
```diff
@@ -24,6 +24,9 @@ namespace operators {
 
 using Tensor = framework::Tensor;
 
+constexpr int kMULMKLDNNFP32 = 1;
```
luotao1: Do we need `kMULMKLDNNFP32`? It's not used in this PR.

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
```diff
@@ -0,0 +1,193 @@
+# Copyright (c) 2019 PaddlePaddle Authors. All Rights Reserved.
```
luotao1: move int8 related ut from `test_mul_mkldnn_op.py` to `test_mul_int8_mkldnn_op.py`

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Sand3r-: Since as you told me mul is just an fc without bias, it makes sense to have multichannel quantization. Please discard my initial comment.

paddle/fluid/operators/mul_op.h
Sand3r-: Yeah, but that's grad code for general case which includes fp32.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,537 @@
...
+namespace ops = paddle::operators;
+REGISTER_OP_KERNEL_WITH_CUSTOM_TYPE(mul, MKLDNN, ::paddle::platform::CPUPlace,
+                                    U8, ops::kMULMKLDNNINT8,
+                                    ops::MulMKLDNNKernel<uint8_t, float>);
+
+REGISTER_OP_KERNEL_WITH_CUSTOM_TYPE(mul, MKLDNN, ::paddle::platform::CPUPlace,
+                                    S8, ops::kMULMKLDNNINT8,
+                                    ops::MulMKLDNNKernel<int8_t, float>);
+
+REGISTER_OP_KERNEL(mul, MKLDNN, ::paddle::platform::CPUPlace,
```
luotao1, line 432: what's the difference between line 519 and line527?

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
```diff
@@ -0,0 +1,193 @@
...
+
+        self.outputs = {'Out': output}
+
+
+'''
+ test case for fp32 * fp32
+'''
+
+
+class TestMKLDNNMulOpFP32(TestMKLDNNMulOpS8S8):
```
luotao1: why FP32 inherit from `TestMKLDNNMulOpS8S8`?

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
```diff
@@ -0,0 +1,193 @@
...
+            "scale_out": self.scale_out,
+            "force_fp32_output": self.force_fp32,
+        }
+
+    def init_data(self):
+        self.scale_x = 1.0
+        self.scale_y = [1.0]
+        self.scale_out = 1.0
+
+        A_data_u7 = np.array([[127, 125, 54, 126], [88, 122, 78, 44],
```
luotao1: why not use `np.random.random` for the data?

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
```diff
@@ -0,0 +1,193 @@
...
+            (self.scale_x * self.scale_y[0])
+        output = np.round(output * scale_output_shift).astype(self.dsttype)
+
+        self.inputs = {'X': A_data_u8, 'Y': B_data_s8}
+
+        self.outputs = {'Out': output}
+
+
+'''
+ test case for  s8 * u7
```
luotao1: Do we need test s8*u7?

luotao1: what's u7?

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
```diff
@@ -0,0 +1,193 @@
...
+import unittest
+import paddle.fluid.core as core
+import numpy as np
+from paddle.fluid.tests.unittests.test_mul_op import TestMulOp, TestMulOp2
+'''
+ test case for s8 * s8
+'''
+
+
+class TestMKLDNNMulOpS8S8(TestMulOp):
```
luotao1: There are a lot of duplicated codes in ut.

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
```diff
@@ -0,0 +1,193 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+import unittest
+import paddle.fluid.core as core
+import numpy as np
+from paddle.fluid.tests.unittests.test_mul_op import TestMulOp, TestMulOp2
```
luotao1: I don't see any unit-test inherit from TestMulOp2.
Please see the unit tests in `test_conv2d_int8_mkldnn.py`

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
```diff
@@ -0,0 +1,193 @@
...
+'''
+ test case for s8 * s8
```
luotao1: There  should be s8*s8, u8*u8, s8*u8, u8*s8 four types, and each type test `TestMulOp` and `TestMulOp2`?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: okay, thanks.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: It seems like there has to be default kernel type registered by `REGISTER_OP_KERNEL()` when using `MKLDNN` or otherwise build would fail.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: done

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
Minghui-Intel: done

paddle/fluid/operators/mul_op.h
Minghui-Intel: no, we don't use it for now since we unregister <fp32, fp32> type kernel. `kMULMKLDNNFP32` is reserved for future use. have deleted it.

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
Minghui-Intel: well, it's because this set of numbers can reproduce the problem of output value mismatch caused by overflow truncation on SKL.  To pass CI, I changed `A_data_u8 = np.array([[252, 136, 54, 160], [88, 132, 78, 44],` to `A_data_u8 = np.array([[152, 136, 54, 160], [88, 132, 78, 44],` in TestMKLDNNMulOpS8U8.

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
Minghui-Intel: yeah,  it's no need to inherit from `TestMKLDNNMulOpS8S8`. Have fixed it.

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
Minghui-Intel: have refined it. pls have another review.

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
Minghui-Intel: have added it(didn't inherit from TestMulOp2, just implement the test case designed by TestMulOP2).

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
Minghui-Intel: As `input y` is always be quantized to s8, so actually, we just need u8s8, s8s8, of course, each type both with `TestMulOp` and `TestMulOp2`.

python/paddle/fluid/tests/unittests/mkldnn/test_mul_mkldnn_op.py
Minghui-Intel: s8u7 is an additional(or double confirm) test to check for overflow truncation issue on SKL.
```
x * y in MKLDNN is actually compute like 2 * (x/2 * y), so I think it's necessary to confirm (x/2 * y) wouldn't lead to overflow.
```
but it's has deleted in new UT.

Minghui-Intel: u7 is `x/2`.  but this case has been instead of new UT case, pls have another review.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
sfraczek: have you decided not to change this or not yet?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
sfraczek: have you decided not to change this or not yet?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
sfraczek: have you decided not to change this or not yet?

paddle/fluid/operators/mul_op.cc
bingyanghuang: `scale_x`, `scale_y` and `force_fp32_output` are needed for QAT, same as in conv2d. Besides `scale_out` is needed for int8v2.

luotao1: Why int8v2 use the different attribute from QAT?

bingyanghuang: int8v2 will use all the attributes mentioned above. `scale_out` is used to convert s32 output -> s8 .

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
yihuaxu: Done! please refer to bb64cb8

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
yihuaxu: Done! please refer to bb64cb8

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
yihuaxu: Done! Please refer to ac7de0e

sfraczek (dismissed):

Minghui-Intel (commented):
1. [done] rename ut;
2. [done] refine ut code;
pls have another review.

Minghui-Intel:
@luotao1 new commit has done the variable name changing(renamed the variables related to FC Op).  please have another review. thanks.

Minghui-Intel:
Accuracy data of QAT models with INT8 MUL OP:
---

n/a | top1 | diff | top5 | diff
-- | -- | -- | -- | --
googlenet | 0.70386 | -0.00014 | 0.89464 | -0.00004
mobilenet_v1 | 0.70842 | -0.00018 | 0.8957 | -0.0001
mobilenet_v2 | 0.72094 | -0.00006 | 0.90704 | -0.00046
resnet101 | 0.77466 | -0.00024 | 0.9365 | -0.0001
resnet50 | 0.76618 | -0.00032 | 0.93082 | 0.00012
vgg16 | 0.72694 | -0.00036 | 0.91092 | -0.00018
vgg19 | 0.7337 | -0.0002 | 0.91414 | 0.00014

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+      if (is_y) {
+        x_tmp.mutable_data<K>(ctx.GetPlace(), data->memory_size());
+
+        auto src_mem =
+            memory({src_mdesc, engine_}, to_void_cast<K>(data->data<K>()));
+        auto dst_mem =
+            memory({dst_mdesc, engine_}, to_void_cast<K>(x_tmp.data<K>()));
+
+        auto reorder = mkldnn::reorder(src_mem, dst_mem);
+        stream(stream::kind::eager).submit({reorder}).wait();
```
Sand3r-: I think you could've just used Reorder function here.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+      } else {
+        x_tmp.mutable_data<T>(ctx.GetPlace(), data->memory_size());
+
+        auto src_mem =
+            memory({src_mdesc, engine_}, to_void_cast<T>(data->data<T>()));
+        auto dst_mem =
+            memory({dst_mdesc, engine_}, to_void_cast<T>(x_tmp.data<T>()));
+
+        auto reorder = mkldnn::reorder(src_mem, dst_mem);
+        stream(stream::kind::eager).submit({reorder}).wait();
```
Sand3r-: ...and here as well.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+    return CreateMemDescriptor(dims, type, format);
+  }
+
+  memory CreateMemory(const memory::desc &desc, const Tensor *tensor,
+                      bool is_y) {
+    if (is_y)
+      return CreateMemory(desc, tensor->data<K>());
+    else
+      return CreateMemory(desc, tensor->data<T>());
+  }
```
Sand3r-, line 156: The `is_y` parameter is unnecessary here. It'd be enough to templatize the function with datatype that would be passed to tensor->data<T>() function.

```
  template <typename DT>
  memory CreateMemory(const memory::desc &desc, const Tensor *tensor) {
      return CreateMemory(desc, tensor->data<DT>());
  }
```

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+    auto dims = framework::vectorize2int(data->dims());
+    auto src_mdesc = CreateMemDescriptor(dims, type, src_fmt);
+
+    if ((data->dims().size() == 4 &&
+         src_fmt != (dst_fmt = memory::format::nchw)) ||
+        (data->dims().size() == 5 &&
+         dst_fmt != (dst_fmt = memory::format::ncdhw))) {
+      auto dst_mdesc = CreateMemDescriptor(dims, type, dst_fmt);
+
+      if (is_y) {
```
Sand3r-: The only way these two branches differ is the choice of type parameter. This can be solved by templatizing the `UpdateDataFormat` function so that it accepts the data type and uses it for all the `to_void_cast` or `data->data` functions.

```
template <typename DT>
Tensor UpdateDataFormat(const Tensor *data, int num_col_dims,
                          memory::data_type type, const ExecutionContext &ctx)
```
This makes the code shorter and more maintainable.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+      const Tensor *x_input, const Tensor *y_input, Tensor *output,
+      const ExecutionContext &ctx) {
+    /* check data format and reorder if need */
+    int x_num_col_dims = ctx.Attr<int>("x_num_col_dims");
+    int y_num_col_dims = ctx.Attr<int>("y_num_col_dims");
+    auto scale_y = ctx.Attr<std::vector<float>>("scale_y");
+
+    memory::data_type x_dt = framework::ToMKLDNNDataType(x_input->type());
+    memory::data_type y_dt = framework::ToMKLDNNDataType(y_input->type());
+    auto x_matrix = this->UpdateDataFormat(x_input, false, /*is_y*/
```
Sand3r-: Please refer to the relevant comment in `MulPrimitiveFactory` on replacing `is_y` with a function template parameter.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+    if (this->mul_) {
+      UpdateDataPointers(ctx, output, &x_matrix);
+      return *(this->mul_);
+    }
+
+    auto src_desc =
+        this->CreateMemDescriptor(&x_matrix, x_dt, memory::format::nc);
+    this->x_input_ = this->CreateMemory(src_desc, &x_matrix, false /*is_y*/);
+
+    if (std::is_same<K, float>::value) {
```
Sand3r-: More of a question rather than a suggestion here: When could weights be of type float instead of an int8 one?

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+
+    bool force_fp32_output = ctx.Attr<bool>("force_fp32_output");
+    memory::data_type dst_dt;
+    if (force_fp32_output) {
+      dst_dt = framework::ToMKLDNNDataType(
+          framework::DataTypeTrait<float>::DataType);
+    } else {
+      dst_dt = framework::ToMKLDNNDataType(
+          framework::DataTypeTrait<int8_t>::DataType);
+    }
```
Sand3r-: This can be simplified to:
```
memory::data_type dst_dt = force_fp32_output ? MKLDNNGetDataType<float>() : MKLDNNGetDataType<int8_t>();
```
I think you'd need to add an appropriate overload in `mkldnn_helper.h` for `int8_t` type though, since its not yet available.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+  boost::optional<memory> output_;
+  boost::optional<inner_product_forward> mul_;
+};  // namespace operators
+
+template <typename T, typename K>
+class QuantMulPrimitiveFactory : public MulPrimitiveFactory<T, K> {
+ public:
+  using MulPrimitiveFactory<T, K>::MulPrimitiveFactory;
+
+  virtual inner_product_forward CreateMulPrimitive(
```
Sand3r-, line 202: This function has gotten quite lengthy now. You may consider dividing it into functions such as:
`CreateInputMemories()`, `CreateXMemory()`, `CreateYMemory()`, `CreateOutputMemory()` so that their outputs can be fed into `CreateMulPrimitive()`. Nothing crucial, but it would improve the readability of the code.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+    auto reorder_pd = mkldnn::reorder::primitive_desc(
+        src_mem.get_primitive_desc(), dst_mem.get_primitive_desc(), attr);
+
+    auto reorder = mkldnn::reorder(reorder_pd, src_mem, dst_mem);
+    stream(stream::kind::eager).submit({reorder}).wait();
+
+    return dst_mem;
+  }
+
+  memory TransposeInputY(const Tensor *input_y, memory::data_type type) {
```
Sand3r-: This function can be safely removed. As it doesn't make any use of `type` parameter, an equivalent function inherited from `MulPrimitiveFactory` can be used.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+  const std::string key = GetHash(input_x, input_y, ctx.op().Output("Out"));
+
+  auto prim_creator =
+      std::static_pointer_cast<MulPrimitiveFactory<T, K>>(dev_ctx.GetBlob(key));
+  if (prim_creator == nullptr) {
+    bool enable_quant =
+        std::is_same<T, int8_t>::value || std::is_same<T, uint8_t>::value;
+    if (enable_quant) {
+      prim_creator =
+          std::make_shared<QuantMulPrimitiveFactory<T, K>>(mkldnn_engine);
```
Sand3r-: I have noticed you have used `ctx.Attr<bool>("force_fp32_output")` in your code. Perhaps it'd be easier to check this function here, just once, and pass the information about output datatype through template parameter to the `QuantMulPrimitiveFactory` factory so that there is much less hassle with checking what the output dt should be. The code would be shorter that way as well.  That's just something to consider.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+    std::vector<float> output_shift_scale(count);
+    for (int i = 0; i < count; i++) {
+      if (scale_y_data[i] == 0.0)
+        output_shift_scale[i] = scale_out_data;
+      else
+        output_shift_scale[i] =
+            scale_out_data / (scale_x_data * scale_y_data[i]);
+    }
+    int mul_mask = is_multi_channel ? 1 : 0;
+    mul_attr.set_output_scales(mul_mask, output_shift_scale);
```
Sand3r-, line 293: Please consider extracting the creation of `mkldnn::primitive_attr mul_attr` to an external function. It would make `CreateMulPrimitive` much shorter and easier to analyse.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
```diff
@@ -0,0 +1,518 @@
...
+    out->set_layout(DataLayout::kMKLDNN);
+    out->set_format(out->format());
+  }
+};
+
+}  // namespace operators
+}  // namespace paddle
+
+namespace ops = paddle::operators;
+REGISTER_OP_KERNEL_WITH_CUSTOM_TYPE(mul, MKLDNN, ::paddle::platform::CPUPlace,
```
Sand3r-, line 424: I think it'd be worth to add op registration for plain `fp32` as well.

Sand3r- (commented):
@Minghui-Intel thank you so much for the changes you have introduced the last time. Please take your time to acknowledge yourself with the suggestions I have enclosed in the comments below.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: I have refined the code to make this function shorter.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: yeah...good suggestion, followed.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: yeah, I think so if the MUL Op wouldn't be used as regular MUL OP, but for now, fp32 x fp32 hasn't been tested in models, so we plan to register plain fp32 in future when it's necessary.

paddle/fluid/operators/mkldnn/mul_mkldnn_op.cc
Minghui-Intel: I tried to use 3 template parameters and it does make the code shorter, but in this way, the code seems to be more difficult to understand. so I still use 2 template parameters for now.

Sand3r- (dismissed):
@Minghui-Intel, thank you for your hard work on this feature. I'm glad you found the suggestions useful, thank you for applying the 3 template parameters to quantization factory as well.
The code to be merged is maintainable, easy to read and passes all CIs. In general, it LGTM.

luotao1:
```
[06:47:18]	[Step 1/1] + HAS_CONST_CAST=const_cast
[06:47:18]	[Step 1/1] + '[' const_cast ']'
[06:47:18]	[Step 1/1] + '[' 17834 '!=' '' ']'
[06:47:18]	[Step 1/1] ++ curl -H 'Authorization: token 5ae37427a95d9e51fb0ea3c71121b75d5c2ebefa' 'https://api.github.com/repos/PaddlePaddle/Paddle/pulls/17834/reviews?per_page=10000'
[06:47:18]	[Step 1/1] ++ python /paddle/tools/check_pr_approval.py 1 21351065 3048612 46782768 30176695 12538138 6836917 32832641
[06:47:18]	[Step 1/1]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[06:47:18]	[Step 1/1]                                  Dload  Upload   Total   Spent    Left  Speed
[06:47:18]	[Step 1/1] 
[06:47:19]	[Step 1/1]   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[06:47:20]	[Step 1/1]   0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
[06:47:20]	[Step 1/1]   0  100k    0   139    0     0     67      0  0:25:33  0:00:02  0:25:31    67
[06:47:20]	[Step 1/1] 100  100k  100  100k    0     0  38150      0  0:00:02  0:00:02 --:--:-- 38160
[06:47:20]	[Step 1/1] current pr 17834 got approvals: FALSE
```
Please do not use `const_cast` in the codes.

luotao1 (approved):
LGTM

PR #17835, opened 2019-06-04
Fix 3 bugs for CI Cuda9/10 tasks

zhhsplendid:
The detail of 3 bugs:

1. Unit test test_imperative_se_resnext always fail on Cuda9/10.
The reason may be the large memory cost so set "Exclusive" for the
task.

2. tools/manylinux1/Dockerfile cannot build Cuda9 Cuda10 image on
CI. Fixed it and tested manually. Will test on CI after this PR is
merged.

3. tools/manylinux1/build_scripts/install_nccl2.sh doesn't have
Cuda10 version. Added it

test=develop


chengduoZH (dismissed):
LGTM

chengduoZH (dismissed):

PR #17837, opened 2019-06-04
Use Python C-API to speed up dygraph trace

sneaxiy:
Use Python C-API to speed up Python code in `tracer.py`.

This PR speeds up ptb model 6%.


JiabinYang (approved):
LGTM

PR #17845, opened 2019-06-04
Cache compiled function instead in ngraph engine

baojun-nervana:
This PR includes the followings:
1. cached compiled ngraph function instead function, this will have some performance benefit to avoid recompile each time
2. Added a basic cache management by clearing the cache when it reaches the capacity cap (set is as 5 for now); this is necessary for BERT model training
3. Simplified some logic and made the code concise; Original it contains a cache off mode for debug purpose, it was replaced by forcing cache clearing.

### DIFF
--- a/paddle/fluid/operators/ngraph/ngraph_engine.cc
+++ b/paddle/fluid/operators/ngraph/ngraph_engine.cc
@@ -465,12 +465,12 @@ void NgraphEngine::RunInferShape() {
   }
 }
 
-void NgraphEngine::BuildNgFunction(const framework::ExecutionContext& ctx) {
+std::shared_ptr<ngraph::Function> NgraphEngine::BuildNgFunction(
+    const framework::ExecutionContext& ctx) {
   Prepare(ctx);
   RunInferShape();
   GetNgInputShape();
   BuildNgNodes();
-  ngraph_function_ = nullptr;
   ngraph::NodeVector func_outputs;
   ngraph::ParameterVector func_inputs;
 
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -485,99 +485,105 @@ void NgraphEngine::BuildNgFunction(const framework::ExecutionContext& ctx) {
...
+    ++it_tensor;
+  }
+  t_in_cache_.clear();
 }
 
 void NgraphEngine::GetNgFunction(const framework::ExecutionContext& ctx) {
   auto interval = ctx.Attr<std::vector<int>>("interval");
   std::string engine_key = ctx.Attr<std::string>("engine_key");
+
+  // set to flase, to debug cache or recompile everytime.
```
LeoZhao-BJ, line 519: so use_cache = false is just for debug purpose, no functional support for non-cache mode anymore, right?
or simply saying, even use_cache = false, bridge will still do cache internally, and clear cache before next time execution.

paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -644,8 +650,7 @@ void NgraphEngine::Run(const framework::Scope& scope,
     }
   }
 
-  auto handle = backend_->compile(ng_func);
-  handle->call_with_validate(t_out, *p_t_in);
+  ng_handle->call(t_out, *p_t_in);
```
LeoZhao-BJ, line 673: in this way, it can really improve performance.

paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -485,99 +485,105 @@ void NgraphEngine::BuildNgFunction(const framework::ExecutionContext& ctx) {
...
+  engine_cache.clear();
+  auto it_tensor = t_in_cache_.begin();
+  while (it_tensor != t_in_cache_.end()) {
+    auto t_vec = it_tensor->second;
+    for (auto t_in : t_vec) {
+      t_in.reset();
+    }
+    ++it_tensor;
+  }
+  t_in_cache_.clear();
```
LeoZhao-BJ, line 512: those lines change gonna fix memory leak issue?

paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -485,99 +485,105 @@ void NgraphEngine::BuildNgFunction(const framework::ExecutionContext& ctx) {
...
+  if (!use_cache) ClearNgCache();
+
+  this->func_cache_key_ = "";
+  for (int i = 0; i < static_cast<int>(feed_vars.size()); ++i) {
+    auto* var = scope_.FindVar(feed_vars[i]);
+    if (var && var->IsType<framework::LoDTensor>()) {
+      auto* tensor_pd = GetLoDTensorOrSelectedRowsValueFromVar(*var);
+      auto dims = tensor_pd->dims();
+      for (int j = 0; j < dims.size(); ++j) {
+        func_cache_key_ += std::to_string(dims[j]);
```
LeoZhao-BJ, line 530: For single instance case, it is enough, but for multi-instance case, this key generation condition may not be sufficient .

paddle/fluid/operators/ngraph/ngraph_engine.cc
baojun-nervana: yes

paddle/fluid/operators/ngraph/ngraph_engine.cc
baojun-nervana: yes, this simplify the cache off mode

paddle/fluid/operators/ngraph/ngraph_engine.cc
baojun-nervana: we will update per multi-instance need when we get there

paddle/fluid/operators/ngraph/ngraph_engine.cc
baojun-nervana: yes, it is faster

mozga-intel (approved):
LGTM!

LeoZhao-BJ (approved):

tensor-tang (approved):
LGTM Thanks @baojun-nervana @mozga-intel @LeoZhao-Intel

PR #17856, opened 2019-06-05
fix bug: the lod_tensor_to_array op will aplly a new var but not release when doing inference.

NHZlX:
The lod_tensor_to_array op will aplly a new var but not release when doing inference.

So the memory will continue to grow when doing the inference.

### DIFF
--- a/paddle/fluid/operators/lod_tensor_to_array_op.cc
+++ b/paddle/fluid/operators/lod_tensor_to_array_op.cc
@@ -131,9 +131,7 @@ class LoDTensorToArrayOp : public framework::OperatorBase {
       }
     }
 
-    auto &outputs = *const_cast<framework::Scope &>(scope)
-                         .Var()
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/lod_tensor_to_array_op.cc
```diff
@@ -131,9 +131,7 @@ class LoDTensorToArrayOp : public framework::OperatorBase {
       }
     }
 
-    auto &outputs = *const_cast<framework::Scope &>(scope)
-                         .Var()
-                         ->GetMutable<std::map<size_t, framework::Tensor>>();
+    std::map<size_t, framework::Tensor> outputs;
```
chengduoZH, line 134: The tensor of outputs is shared from `out[i]` by calling `out[i].Slice`, it doesn't matter that place in stack space.

chengduoZH (approved):

PR #17868, opened 2019-06-05
test=develop, add dygraph_not_support and refine ocr

JiabinYang:
This PR add `dygraph_not_support ` decorator in dygraph mode

### DIFF
--- a/python/paddle/fluid/dygraph/base.py
+++ b/python/paddle/fluid/dygraph/base.py
@@ -22,6 +22,7 @@ from .tracer import Tracer
 __all__ = [
     'enabled',
     'no_grad',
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/base.py
```diff
@@ -52,6 +62,7 @@ def __impl__(*args, **kwargs):
 
 
 no_grad = wrap_decorator(_no_grad_)
+dygraph_not_support = wrap_decorator(_dygraph_not_support_)
```
sneaxiy: Too long name. Users have to write something like `dygraph.dygraph_not_support`. Please simplify it to be `not_supported`.

PR #17878, opened 2019-06-06
Remove attribute in Allocator::Allocate

sneaxiy:
`Allocator::Attr` is not used and has been intrusive to many modules of Paddle.

We remove it to clean codes.


chengduoZH (approved):
LGTM

PR #17881, opened 2019-06-06
 Add trainable_statist attr for bn in dygraph

tink2123:
Add trainable_statist attr for bn in dygraph, if set trainable_statist True, is equal to is_test=False


phlrain (approved):
LGTM

PR #17886, opened 2019-06-06
add target assign, sigmoid focal loss, detection output operations for supporting retinanet 

FlyingQianMM:
test=develop
1. retinanet_target_assign_op
   This operator can be, for given the Intersection-over-Union (IoU) overlap between anchors and ground truth boxes, to assign classification and regression targets to each anchor, these target labels are used for training retinanet. No other sampling strategy is adopted to filter negative anchors because all negative labels along with positive labels contribute to the classfication objective(focal loss). 

2. sigmoid_focal_loss_op
  Focal loss is used to address the foreground-background class imbalance existed on the training phase of one-stage detectors. This operator computes the sigmoid value for each element in the input tensor, after which focal loss is measured.

3. retinanet_detection_out
  This operation is to get the detection results by performing following steps: (1)Divide the input bounding box predictions into box predictions perFPN level, according to that the box number of the i-th FPN level is 4**(MAX_LEVEL-i) times of that of the MAX_LEVEL-th FPN level. (2) Decode top-scoring bounding box predictions per FPN level according to the anchor boxes. (3)Merge top predictions from all levels and applying multi-class non maximum suppression (NMS) on them to get the final detections.


luotao1:
It's better to split into three PRs, i.e one Op one PR.

PR #17890, opened 2019-06-06
Update the Anakin interfaces for content-dnn and MLU

Shixiaowei02:
1. Modified the obsolete interfaces of Anakin.
2. Support the interfaces of the MLU hardware and modify the cmake files.

The content-dnn and MLU architectures are supported after this pull request is merged.


fc500110 (approved):
LGTM

PR #17570, opened 2019-05-22
Improve mobilenetv2 INT8 performance by using INT8 relu as post-op

lidanqing-vv:
This PR improved mobilenetv2 INT8 performance with good accuracy.

We have to use relu instead of brelu as the post-op in INT8 conv2d kernel , because INT8 brelu as a post-op is not enabled in mkldnn v01.8. I add TODO and comments of what will be changed when v0.20 is enabled.

The performance of mobilenetv2 with this PR is follows:

INT8/FP32 | Top1 Accuracy |  Performance
| -- |-- | --
FP32 | 71.90% | X
INT8 | 71.43% | 1.92 X

test machine: Intel(R) Core(TM) i9-7940X CPU @ 3.10GHz, 14 Cores
paddle_num_threads 14

Performance of all the int8 models on CLX and SKX will be delivered by this Friday.

PR #17897, opened 2019-06-06
refactor the function ConvFwdPrimitiveDesc

lidanqing-intel:
refractor the function ConvFwdPrimitiveDesc
as required in [PR17570](https://github.com/PaddlePaddle/Paddle/pull/17570) 

Maybe there are more places that could be refactored but I will work on them later.

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -460,24 +460,17 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
       // TODO(lidanqing): We use relu post-op instead of brelu post-op cause
       // mkldnn v0.18 does not support INT8 brelu post-op. Use code in /**/ when
       // v0.20 is enabled
+      boost::optional<memory::desc> bias_md(boost::none);
       if (bias) {
         bias_tz = paddle::framework::vectorize2int(bias->dims());
-        auto bias_md = platform::MKLDNNMemDesc(bias_tz, memory::data_type::s32,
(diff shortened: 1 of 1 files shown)


lidanqing-vv:
@wojtuss  @luotao1

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -473,10 +473,10 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
 
       } else {
         conv_pd = ConvFwdPrimitiveDesc(
-            src_md, weights_md, dst_md, strides, paddings, mkldnn_engine,
-            fuse_relu || fuse_brelu /*fuse_relu*/, fuse_residual_conn,
-            false /*fuse_brelu*/, fuse_brelu_threshold, output_shift_scale,
-            sum_scale, is_test);
+            src_md, weights_md, boost::none, dst_md, strides, paddings,
```
wojtuss: What about defining the `bias_md = boost::none` before `if (bias) {...}` block, and calling the `ConvFwdPrimitiveDesc()` in a single place?

lidanqing-vv: > What about defining the `bias_md = boost::none` before `if (bias) {...}` block, and calling the `ConvFwdPrimitiveDesc()` in a single place?

After trying `boost::optional<memory::desc> bias_md{none_t}` and then call in single place. We found we could not pass reference as parameter. As we discussed it may copy big memory.
Later we decided to use `memory::desc* bias_md_p = nullptr;`

PR #17908, opened 2019-06-06
Fix/program doc

JiabinYang:
This PR add some comment for `fluid.Program.clone()`


python/paddle/fluid/framework.py
```diff
@@ -2756,6 +2756,15 @@ class Program(object):
     it will contains nested block.
```
xsrobin: contains->contain

python/paddle/fluid/framework.py
```diff
@@ -2756,6 +2756,15 @@ class Program(object):
...
     Please reference the framework.proto for details.
```
xsrobin, line 2759: reference->refer to

python/paddle/fluid/framework.py
```diff
@@ -2756,6 +2756,15 @@ class Program(object):
...
 
+    A set of Program usually contains startup program and main program.
```
xsrobin: contains->contain

python/paddle/fluid/framework.py
JiabinYang: Changed, but these are not related to my PR

python/paddle/fluid/framework.py
JiabinYang: Changed, but these are not related to my PR

PR #17911, opened 2019-06-07
Feature/buffer_shared_inplace

sneaxiy:
This PR works for:

- refactoring the inplace pass to fix the issues that users have to set `var.persistable = True` before using inplace pass.
- turning on `build_strategy.enable_inplace = True` by default.
- enabling caching pointers of `Variable` in local execution scope.
- fixing the bug of global scope sharing in two different ParallelExecutors.

[CE Debug](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=51510&buildTypeId=PaddleModesl_ModelDebug_Model11Build&tab=buildLog) has passed except `flowers_32_gpu_memory` of vgg16 model. But I found that this kpi value in CE Debug may be wrong, because:
- This PR only changes `BuildStrategy` in `ParallelExecutor`, which does not have any effect on `Executor`. VGG16 model in CE Debug uses `Executor` but not `ParallelExecutor`.
- I have tried another PR (#18065) which does nothing to do with memory optimization, but it also fails in [CE Debug](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=51827&buildTypeId=PaddleModesl_ModelDebug_Model11Build&tab=buildLog) for the same reason.

```
[15:35:31][Step 3/3] Evaluate [df81b131a6f1cc17aab44619362c17df44c37e19] failed!
[15:35:31][Step 3/3] The details:
[15:35:31][Step 3/3] task: model_ce_vgg16
[15:35:31][Step 3/3] passed:  False
[15:35:31][Step 3/3] infos [cifar10_128_train_speed] pass
[15:35:31][Step 3/3] [cifar10_128_gpu_memory] pass
[15:35:31][Step 3/3] [flowers_32_train_speed] pass
[15:35:31][Step 3/3] [flowers_32_gpu_memory] failed, diff ratio: 0.11795102224219277 larger than 0.1.
[15:35:31][Step 3/3] kpis keys ['cifar10_128_train_speed', 'cifar10_128_gpu_memory', 'flowers_32_train_speed', 'flowers_32_gpu_memory']
[15:35:31][Step 3/3] kpis values [[[1703.047607421875]], [[2332]], [[111.2959976196289]], [[9952]]]
[15:35:48][Step 3/3] model_ce_vgg16 
```


paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
```diff
@@ -59,39 +64,53 @@ FeedFetchList ScopeBufferedSSAGraphExecutor::Run(
...
     for (auto &info : var_infos_) {
-      if (scope->FindVar(info.name_) != nullptr) {
-        continue;
-      }
       if (info.persistable_) {  // Persistable
+        auto var = scope->FindVar(info.name_);
+        if (var != nullptr) {
+          VLOG(2) << info.name_ << " is found in parent scope, skipped";
+          InitializeVariable(var, info.type_);
+          continue;
```
chengduoZH, line 109: 1. If `var` is not nullptr, this var may in the `scope` or the `scope`‘s parent.
2. Now that `var` is not nullptr, why do you call `InitializeVariable` here?

paddle/fluid/framework/scope.cc
```diff
@@ -200,6 +200,21 @@ Variable* Scope::FindVarLocally(const std::string& name) const {
...
+    SCOPE_VARS_WRITER_LOCK
+    for (auto iter = vars_.begin(); iter != vars_.end();) {
+      if (vars.count(iter->second.get()) != 0) {
+        ++iter;
+      } else {
+        vars_.erase(iter++);
+      }
+    }
+  }
+}
```
chengduoZH, line 212: I prefer that not add `ClearWithPreserve` here, because scope provides `EraseVars` and `DropKids`, and `ClearWithPreserve` can be replaced by `EraseVars` and `DropKids`. **We should ensure that the function of the interface of scope is orthogonal.**

paddle/fluid/framework/details/scope_buffered_ssa_graph_executor.cc
sneaxiy: 1. I have changed the log message. Thanks!
2. I have removed `InitializeVariable` when `info.name_` is found in `scope`.

sneaxiy: 1. I have changed the log message. Thanks!
2. I have removed `InitializeVariable` when `info.name_` is found in `scope`.

paddle/fluid/framework/scope.cc
sneaxiy: I agree. But `Scope` does not have a method like `EraseVarsExcept`. How about add this method?

paddle/fluid/framework/ir/memory_optimize_pass/memory_reuse_pass.h
```diff
@@ -0,0 +1,128 @@
...
+ *    NoNeedBufferVarsInference.
+ *
+ * Considering the drawbacks of the former renaming strategy, we design a
+ * novel memory-optimize pass to fix these issues. Whether in-place is
+ * performed can be decided during run-time. ShareTensorBufferOpHandle
+ * would only share tensor buffers between in/out, never rename variable,
+ * and not change dim and LoD of variable. If users want to fetch a certain
+ * variable, we can skip in-place during run-time.
+ *
+ * The only concern on speed performance may be: there are too many
```
zhhsplendid, line 67: Great explanation comment!

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_inplace_op_pass.cc
```diff
@@ -0,0 +1,160 @@
...
+    auto in_to_outs =
+        OpInfoMap::Instance().Get(op_type).infer_inplace_(*op_desc, use_cuda);
+    for (auto &pair : in_to_outs) {
+      auto &in_param = pair.first;
+      auto &in_args = op_desc->Input(in_param);
+      if (in_args.empty()) {
+        VLOG(4) << "Cannot inplace because Input(" << in_param
+                << ") is empty in " << op_type;
+        continue;
+      }
```
chengduoZH, line 92: Why does this happen? The input is empty but the input and output could be inplace.

chengduoZH (approved):

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_inplace_op_pass.cc
sneaxiy: Not quite sure. Just to avoid some possible corner cases.

PR #17924, opened 2019-06-09
Fix FLAGS_fuse_parameter_memory_size unit from Bytes to MBytes.

gongweibao:
We use MBytes common instead of Bytes.

### DIFF
--- a/paddle/fluid/framework/ir/alloc_continuous_space_for_grad_pass.cc
+++ b/paddle/fluid/framework/ir/alloc_continuous_space_for_grad_pass.cc
@@ -23,8 +23,8 @@
 #include "paddle/fluid/framework/ir/graph_helper.h"
 #include "paddle/fluid/framework/op_registry.h"
 
-DEFINE_uint64(fuse_parameter_memory_size, 0,  // Bytes
-              "fuse_parameter_memory_size is up limited memory size "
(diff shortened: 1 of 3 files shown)


paddle/fluid/framework/ir/alloc_continuous_space_for_grad_pass.cc
```diff
@@ -230,15 +228,16 @@ class AllocContinuousSpaceForGradPass : public ir::Pass {
       }
       VLOG(10) << out.str()
                << ", group size:" << group_grads_params->at(i).size()
-               << ", group memory size:" << gps_size;
+               << ", group memory size:"
+               << static_cast<double>(gps_size) / 1048576.0 << "(MB)";
```
chengduoZH, line 232: Please don't use `1048576.0 ` directely.
```
static double kMB = 1048576.0;
```

chengduoZH (approved):

paddle/fluid/framework/ir/alloc_continuous_space_for_grad_pass.cc
gongweibao: Thanks! Will polish later.

xsrobin (commented):
LGTM

PR #17925, opened 2019-06-09
Light mem reuse strategy for inference.

NHZlX:
The existing dynamic memory reuse strategy has a lot of problems： One of the more important things is that it is not applicable to many general models. I  improved this and applied a simple strategy.


Superjomn (approved):
LGTM

PR #17934, opened 2019-06-10
refine GetExpectedKernelType in concat, test=develop

jerrywgz:
Fix GetExpectedKernelType, the original method is to check whether input variable is initialized, but it can not include all cases when inputs have None. This pr checks whether tensor is initialized and numel is non-zero.

### DIFF
--- a/paddle/fluid/operators/concat_op.cc
+++ b/paddle/fluid/operators/concat_op.cc
@@ -23,7 +23,7 @@ limitations under the License. */
 
 namespace paddle {
 namespace operators {
-using framework::Tensor;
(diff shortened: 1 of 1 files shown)


python/requirements.txt
```diff
@@ -13,3 +13,4 @@ funcsigs
 pyyaml
 decorator
 prettytable
+cython
```
luotao1: why add `cython` here?

jerrywgz: useless, has been deleted

qingqing01 (approved):
LGTM.

PR #17976, opened 2019-06-11
Cherry-pick: fix random CI failure.

zhhsplendid:
In some tests, SWEs used "print('xxx').format('xxx')". The syntax
is only supported in Python2, not python3. However, since those
lines are related to data download, if the CI machines already have
the data, it passes CI tests. That causes random failure.

Also temporary disable a unit test. The test will be fixed under high priority


sneaxiy (approved):
LGTM. Please fix TestCudnnCTCOp asap.

kolinwei (approved):

PR #17995, opened 2019-06-11
cherrypick commit 40885c225b61 into v1.5

SunGaofeng:
This is for unfold op, which has been merged into develop. Now cherry pick it into v1.5.


SunGaofeng:
False because mistake to pr on develop, but it should be v1.5

Issue #17903, opened 2019-06-06
SetNumThreads(1) breaks slim multi-threading


sfraczek:
Hi,

https://github.com/PaddlePaddle/Paddle/blob/e2c1b7c354c9ca6720228f1b55bfc4c6a6f82292/paddle/fluid/inference/api/analysis_predictor.cc#L240

this change breaks int8 slim tests. Multi-threading works until int8 warm-up phase and after that, everything runs on 1 thread only.
The test series such as test_analyzer_int8_resnet50 are not affected.

Is this change necessary?
Any suggestions how we can fix it?

luotao1:
> Is this change necessary

Yes, we only want to use multi-thread inside `AnalysisPredictor::ZeroCopyRun()` and `AnalysisPredictor::Run()`. You can see the reason in #17801.

> this change breaks int8 slim tests, Multi-threading works until int8 warm-up phase and after that, everything runs on 1 thread only.

How about making `mkldnn_quantizer_config` support `cpu_math_library_num_threads`?

sfraczek:
Should we set number of threads from python before and after calling exe->Run?

sfraczek:
We never hit that `  paddle::platform::SetNumThreads(config_.cpu_math_library_num_threads());` line again from python after it was set to 1 at the end of call. It seems python is calling different method later.

luotao1:
> It seems python is calling different method later

If python is calling a different method, #17801 will not affect it. Could you find the method python called at first?

> Should we set number of threads from python before and after calling exe->Run

Could you give `git diff` of your change?

sfraczek:
I don't have a diff - I mean here (edit ~https://github.com/PaddlePaddle/Paddle/pull/17814/files#diff-79b68fb950ff79f6c979a5b3774ca8f7R199~ https://github.com/PaddlePaddle/Paddle/pull/17634/files?file-filters%5B%5D=.cc&file-filters%5B%5D=.h&file-filters%5B%5D=.py&owned-by%5B%5D=#diff-c06e05c8a775f42a2c811269a7d9e88fR140) this calls Executor and it runs on 1 thread.

I see two options here:
1. surround with pybind calls to setnumthreads like you did here: https://github.com/PaddlePaddle/Paddle/pull/17801/files#diff-859bde2f1a0236fbd9b5ac0c67026fb7R590
2. move the setnumthreads calls to inside executor->run somehow

luotao1:
[...]

PR #18009, opened 2019-06-11
fix slim multithreading issue

sfraczek:
With this fix the slim int8 mkldnn quantization test (https://github.com/PaddlePaddle/Paddle/pull/17634) runs on multiple threads.
This related issue: https://github.com/PaddlePaddle/Paddle/issues/17903.

@zhaify, @bingyanghuang - Do you think its ok that I renamed `FLAGS_OMP_NUM_THREADS` -> `FLAGS_paddle_num_threads` in CMake? There is such a flag defined already in Paddle so I wanted to reuse it since it has the same use. I didn't find OMP_NUM_THREADS flag.

### DIFF
--- a/paddle/fluid/pybind/inference_api.cc
+++ b/paddle/fluid/pybind/inference_api.cc
@@ -23,7 +23,6 @@
 #include <vector>
 #include "paddle/fluid/inference/api/analysis_predictor.h"
 #include "paddle/fluid/inference/api/paddle_inference_api.h"
-#include "paddle/fluid/platform/cpu_helper.h"
 
 namespace py = pybind11;
 
@@ -47,7 +46,6 @@ static void BindNativeConfig(py::module *m);
 static void BindNativePredictor(py::module *m);
 static void BindAnalysisConfig(py::module *m);
 static void BindAnalysisPredictor(py::module *m);
-static void BindPaddlePlatform(py::module *m);
 
 #ifdef PADDLE_WITH_MKLDNN
 static void BindMkldnnQuantizerConfig(py::module *m);
@@ -63,7 +61,6 @@ void BindInferenceApi(py::module *m) {
   BindNativePredictor(m);
   BindAnalysisConfig(m);
   BindAnalysisPredictor(m);
-  BindPaddlePlatform(m);
 #ifdef PADDLE_WITH_MKLDNN
   BindMkldnnQuantizerConfig(m);
 #endif
@@ -313,10 +310,6 @@ void BindAnalysisPredictor(py::module *m) {
            py::return_value_policy::reference)
       .def("SaveOptimModel", &AnalysisPredictor::SaveOptimModel,
            py::arg("dir"));
-}
-
-void BindPaddlePlatform(py::module *m) {
-  m->def("set_omp_num_threads", &platform::SetNumThreads);
 }
 
 }  // namespace pybind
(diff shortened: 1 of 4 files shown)


paddle/fluid/pybind/pybind.cc
```diff
@@ -949,8 +951,10 @@ All parameter, weight, gradient are variables in Paddle.
                      int block_id, bool create_local_scope, bool create_vars,
                      const std::vector<std::string> &fetch_vars) {
         pybind11::gil_scoped_release release;
+        paddle::platform::SetNumThreads(FLAGS_paddle_num_threads);
```
wojtuss: Using the flag `FLAGS_paddle_num_threads` in every call to the `run` method seems inappropriate to me.
What about adding a pybind to the `SetNumThreads()` method, and calling it in the test before executor's `run`?
@luotao1 , what do you think about such an idea?

luotao1: > What about adding a pybind to the SetNumThreads() method, and calling it in the test before executor's run

I think it is a nice idea.

sfraczek: OK I've changed that

paddle/fluid/pybind/inference_api.cc
```diff
@@ -312,5 +315,9 @@ void BindAnalysisPredictor(py::module *m) {
            py::arg("dir"));
 }
 
+void BindPaddlePlatform(py::module *m) {
+  m->def("set_omp_num_threads", &platform::SetNumThreads);
```
luotao1: `set_omp_num_threads`->`set_num_threads`, since openblas don't use openmp.

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ string(REPLACE ".py" "" TEST_OPS "${TEST_OPS}")
 
 function(inference_analysis_python_api_int8_test target model_dir data_dir filename)
     py_test(${target} SRCS ${filename}
-        ENVS FLAGS_OMP_NUM_THREADS=${CPU_NUM_THREADS_ON_CI}
+        ENVS OMP_NUM_THREADS=${CPU_NUM_THREADS_ON_CI}
```
luotao1: `OMP_NUM_THREADS`->`CPU_NUM_THREADS` more readable?

paddle/fluid/pybind/inference_api.cc
```diff
@@ -46,6 +47,7 @@ static void BindNativeConfig(py::module *m);
 static void BindNativePredictor(py::module *m);
 static void BindAnalysisConfig(py::module *m);
 static void BindAnalysisPredictor(py::module *m);
+static void BindPaddlePlatform(py::module *m);
```
luotao1: Why not change in `paddle/fluid/pybind/pybind.cc` directly? This function is not used for inference only.

sfraczek: ok

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
sfraczek: ok

paddle/fluid/pybind/inference_api.cc
sfraczek: ok

luotao1:
please trigger the CI

luotao1 (approved):
LGTM!

luotao1:
Please cherry-pick to release/1.5 @sfraczek

PR #18015, opened 2019-06-12
modify the access level of anakin engine

Shixiaowei02:
Modify the access level of base class from public to protected.

test=develop


fc500110 (approved):
LGTM

PR #18023, opened 2019-06-12
Add distributions of normal and uniform

LielinJiang:
resolve part of #17049

image
image
image
image
image
image
image


python/paddle/fluid/layers/distributions.py
```diff
@@ -0,0 +1,380 @@
...
+
+          uniform = Uniform(low, high)
+
+          sample = uniform.sample([2, 3])
+          entropy = uniform.entropy()
+          lp = uniform.log_prob(values)
+
+    """
+
+    def __init__(self, low, high):
```
qingqing01, line 176: Need to add comments for low, high.  You can refer layers.fc for doc format.

python/paddle/fluid/layers/distributions.py
```diff
@@ -0,0 +1,380 @@
...
+          normal = Normal(loc, scale)
+          other_normal = Normal(other_loc, other_scale)
+
+          sample = normal.sample([2, 3])
+          entropy = normal.entropy()
+          lp = normal.log_prob(values)
+          kl = normal.kl_divergence(other_normal)
+    """
+
+    def __init__(self, loc, scale):
```
qingqing01, line 313: Whether need dtype as the input, since hard code float32 is used in many places.

xsrobin (approved):

qingqing01 (approved):

PR #18025, opened 2019-06-12
Fix reinitialized ncclid error!

gongweibao:
When we use `share_vars_from`  for PE, nccluniqueid will be reinitailized。So we change the ncclcommunicator to be a persistable var.


gongweibao:
TODO: https://github.com/PaddlePaddle/Paddle/issues/18088

chengduoZH (approved):

PR #17952, opened 2019-06-10
Fix scatter and gather op when has duplicate index

wawltor:
Fix #https://github.com/PaddlePaddle/Paddle/issues/17509
The scatter op has a calc bug when the indices has same index, the scatter op use overwrite mode to calculate the same index, fix this bug by using the accumulate mode to calculate the same index.At the same time, the gather op has the same bug when the op calc the grad. And we use the lib of open-blas and eigen to optimize the time cost in accumulate mode.

PR #18038, opened 2019-06-12
Fix gather and scatter op has same index bug  cherry-pick from #17952 

wawltor:
cherry-pick from #17952
The scatter op has a calc bug when the indices has same index, the scatter op use overwrite mode to calculate the same index, fix this bug by using the accumulate mode to calculate the same index.At the same time, the gather op has the same bug when the op calc the grad. And we use the lib of open-blas and eigen to optimize the time cost in accumulate mode.


luotao1 (approved):

xsrobin (approved):
LGTM

PR #18049, opened 2019-06-13
add print log for unittest of distributed training

guru4elephant:
test_dist_xx unittest fails with timeout error frequently these days, I have to add more print log on the unittest because it is very hard to replicate the timeout error on local machine.
test=develop


python/paddle/fluid/tests/unittests/test_async_ssa_graph_executor_mnist.py
```diff
@@ -137,6 +142,7 @@ def train(use_cuda, thread_num, cpu_num):
         main_program=main_program,
         build_strategy=build_strategy,
         exec_strategy=exec_strategy)
+    print("declare parallel executor done.")
```
jacquesqiao, line 145: 这些日志是干啥的

jacquesqiao (dismissed):

python/paddle/fluid/tests/unittests/test_async_ssa_graph_executor_mnist.py
guru4elephant: 主要想定位timeout的具体位置

PR #18072, opened 2019-06-13
Sync Dockerfile change of PR#17889

zhhsplendid:
Jian Tang made change on latest-dev Dockerfile, so sync the change
in the cuda9/10 Dockerfile

test=develop


tensor-tang (approved):
LGTM

PR #18077, opened 2019-06-13
reuse C-API INT8 unit test application

wojtuss:
With this patch, the unit test application binary for image classification INT8 inference is being reused for all the 7 models instead of building 7 separate binaries for each model.

test=develop


sfraczek (commented):
Could you please use the same mechanism for fp32 tests?
This would mean chaning file `test_analyzer_resnet50_tester.cc` -> `test_analyzer_image_classification_tester.cc` and using it for targets: `test_analyzer_resnet50`, `test_analyzer_googlenet`, `test_analyzer_mobilenet_depthwise_conv` in lines 150-160 of paddle/fluid/inference/tests/api/CMakeLists.txt

sfraczek (approved):
LGTM

luotao1:
please cherry-pick to release/1.5 with #18111

PR #18084, opened 2019-06-14
Refine unittest log

guru4elephant:
test_dist_xx unittest fails with timeout error frequently these days, I have to add more print log on the unittest because it is very hard to replicate the timeout error on local machine.
test=develop


raindrops2sea (approved):
debug完成后请对临时log进行清理整理。

PR #18107, opened 2019-06-14
Fix slice op shape=-1 bug

phlrain:
修复slice 再遇到shape=-1情况的bug
已经增加desrease dim的功能


paddle/fluid/operators/slice_op.cc
```diff
@@ -39,21 +39,49 @@ class SliceOp : public framework::OperatorWithKernel {
...
-      out_dims[axes[i]] = end - start;
+      if (dim_value > 0) {
+        start = starts[i] < 0 ? (starts[i] + dim_value) : starts[i];
+        end = ends[i] < 0 ? (ends[i] + dim_value) : ends[i];
+        start = std::max(start, 0);
+        end = std::max(end, 0);
+        start = std::min(start, dim_value);
+        end = std::min(end, dim_value);
+        start = std::min(start, end);
+        PADDLE_ENFORCE_GT(end, start, "end should greater than start");
```
JiabinYang, line 57: should this check before line 56, and in this way maybe we don't need line 56?

paddle/fluid/operators/slice_op.h
```diff
@@ -130,6 +162,36 @@ class SliceGradKernel : public framework::OpKernel<T> {
...
+        for (int i = 0; i < out_dims.size(); ++i) {
+          while (true) {
+            if (vec_origin_out_shape[step_in] == -1) {
+              vec_origin_out_shape[step_in] = out_dims[i];
+              break;
+            }
+
+            step_in++;
+          }
+        }
```
JiabinYang, line 181: how about change a way to implement this it looks weird to using an `while(true)` loop

python/paddle/fluid/framework.py
```diff
@@ -787,6 +787,7 @@ def _sliceAndConcatVar(self, item, axis):
                 return self._cloneVar(True)
             start, stop, step = self._slice_indices(item, self.shape[axis])
             if step == 1:
+                print("22", start)
```
JiabinYang: delete this?

paddle/fluid/operators/slice_op.h
```diff
@@ -55,17 +55,49 @@ class SliceKernel : public framework::OpKernel<T> {
...
```
JiabinYang, line 81: can we change a way to impl this it looks weird with a while(true)

JiabinYang (commented):
some comments

JiabinYang (approved):
LGTM

PR #18108, opened 2019-06-14
Fix py_reader iterable bug

sneaxiy:
The previous async read mode would cause data overrwritten. This PR fixes it.


chengduoZH (dismissed):

zhhsplendid (approved):
LGTM

PR #18114, opened 2019-06-15
Fix getitems slice bug (#18053)

phlrain:
* fix get items slice bug; test=develop

* fix unique_name bug; test=develop


JiabinYang (approved):
GTM

PR #18131, opened 2019-06-17
simplify dataset and executor

xyzhou-puck:
I slightly modify two python file in order to improve the usability of PaddlePaddle:

1. in paddle.dataset.uci_housing, I make the load_data function be able to return train/test data, by which way, users can directly work with that raw data.

2. in paddle.fluid.executor, I give the place a default value (CPUPlace()), so that the user (who does not need GPU) no longer needs to worry about the place anymore.


python/paddle/fluid/executor.py
```diff
@@ -357,7 +357,7 @@ class Executor(object):
 
     """
 
-    def __init__(self, place):
+    def __init__(self, place=fluid.CPUPlace()):
```
kuke, line 360: Please also update the doc

PR #18111, opened 2019-06-14
unify FP32 vs. INT8 comparison tests output

wojtuss:
This patch unifies output format of all the three FP32 & INT8 inference comparison test groups:

- `test_analyzer_int8_*`
- `test_slim_int8_*`
- `test_qat_int8_*`

Added summary sections are helpful also for our validation team in parsing test results from the output.

test=develop

PR #18145, opened 2019-06-17
Cherry pick or  #18077 and  #18111

wojtuss:
https://github.com/PaddlePaddle/Paddle/pull/18077 - reuse C-API INT8 unit test application
https://github.com/PaddlePaddle/Paddle/pull/18111 - unify FP32 vs. INT8 comparison tests output


luotao1 (approved):
LGTM

PR #18152, opened 2019-06-17
fix type error of std::pow in sigmoid_focal_loss_op.cu and sigmoid_focal_loss_op.h

FlyingQianMM:
test=develop

fix type error of std::pow in sigmoid_focal_loss_op.cu and sigmoid_focal_loss_op.h


heavengate (approved):
LGTM

PR #18155, opened 2019-06-17
fix bug in fleet interface, test=develop

a6802739:
在fleet的init函数中初始化_role_maker。


jacquesqiao:
fixed by https://github.com/PaddlePaddle/Paddle/pull/18137

PR #18182, opened 2019-06-18
Reset DeviceContext after quantization warmup

Sand3r-:
Once the quantization warmup is perfromed using fp32 operators, some of the cached MKL-DNN primitvies are incorrectly reused as int8 operators. This PR clears device context cache after the execution of  warmup, so that the aformentioned doesn't occur.

### DIFF
--- a/paddle/fluid/inference/api/mkldnn_quantizer.cc
+++ b/paddle/fluid/inference/api/mkldnn_quantizer.cc
@@ -355,6 +355,13 @@ AnalysisPredictor::MkldnnQuantizer::Histogram(
   return std::make_pair(std::move(hist), std::move(bin_width));
 }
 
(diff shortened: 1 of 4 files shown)


wojtuss (dismissed):
LGTM

luotao1:
Is this PR need for release/1.5 or develop branch?

wojtuss:
> Is this PR need for release/1.5 or develop branch?

@luotao1 
This fix is important, but not critical for release 1.5. All the current tests work fine without it, however, when working on FC INT8 kernel (not to be released in 1.5) we found the reusing problem in the tests and this patch fixes it.

luotao1:
> but not critical for release 1.5

Got it. Could you remove the milestone for this PR?

wojtuss (approved):
LGTM

PR #18186, opened 2019-06-18
Fix Pooling output scale

Sand3r-:
Pooling's output scale should be equal to its input scale, since the operator itself doesn't perform any scaling. This PR ensures such behaviour.

### DIFF
--- a/paddle/fluid/inference/api/mkldnn_quantizer.cc
+++ b/paddle/fluid/inference/api/mkldnn_quantizer.cc
@@ -54,7 +54,7 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
         for (auto const& conn : connections) {
           for (const auto& var_name : conn.second) {
             // skip if scale already computed
-            if (scales_.find(var_name) != scales_.end()) return;
+            if (scales_.find(var_name) != scales_.end()) continue;
 
             auto* var = predictor_.sub_scope_->FindVar(var_name);
             PADDLE_ENFORCE(var, "%s is not in the scope", var_name);
@@ -64,6 +64,7 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
 
             // force unsigned type if already know it
             bool is_unsigned = false;
+            bool compute_scale = true;
             if (is_output) {
               if (op->Type() == "conv2d") {
                 // output of conv2d with relu must be unsigned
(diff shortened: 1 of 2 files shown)


paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -62,24 +62,35 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
...
+                // output of ops with unsigned input must be unsigned
+                is_unsigned = true;
+                for (auto input_var_name : op->Input("X")) {
+                  PADDLE_ENFORCE(
+                      scales_.find(input_var_name) != scales_.end(),
+                      "Input scales must be calculated before the "
+                      "output scales to infer if output is unsigned.");
+                  is_unsigned = is_unsigned && scales_[input_var_name].first;
+                }
+              } else if (op->Type() == "pool2d") {
```
wojtuss: Actually, the same situation concerns all the operators that do not handle different input and output scales, i.e. `transpose2`, `reshape2` and `concat`. Please extend this PR to cover also those operators.

Sand3r-: This issue has been addressed in the latest commit. Feel free to review. 😄

paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -62,24 +64,41 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
...
+                      scales_.find(input_var_name) != scales_.end(),
+                      "Input scales must be calculated before the "
+                      "output scales to infer if output is unsigned.");
+                  is_unsigned = is_unsigned && scales_[input_var_name].first;
+                  min = std::min(
+                      min, scales_[input_var_name].second.data<double>()[0]);
+                }
+                auto scale_tensor = CreateScaleTensor();
+                scale_tensor.data<double>()[0] = min;
+                scales_[var_name] = {is_unsigned, scale_tensor};
```
wojtuss, line 103: Scale calculation should be skipped after the output scale is set here.

paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -62,24 +64,41 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
...
+                // output of ops with unsigned input must be unsigned
+                is_unsigned = true;
+                double min = std::numeric_limits<double>::max();
+                for (auto input_var_name : op->Input("X")) {
+                  PADDLE_ENFORCE(
+                      scales_.find(input_var_name) != scales_.end(),
+                      "Input scales must be calculated before the "
+                      "output scales to infer if output is unsigned.");
+                  is_unsigned = is_unsigned && scales_[input_var_name].first;
+                  min = std::min(
```
wojtuss: Now, the minimum scale is being calculated twice, here and in `QuantizeInputs()` method (`cpu_quantize_pass.cc`).
IMO, calculating minimum scale as well as updating all inputs' scales can be performed here.

paddle/fluid/inference/api/mkldnn_quantizer.cc
Sand3r-: Good point, I'll add a missing `continue` keyword.

paddle/fluid/inference/api/mkldnn_quantizer.cc
Sand3r-: Unfortunately I cannot do it here, because that would overwrite the out_scales of convolutions that use the concat input variables as output vars. I'll move finding min mechanism to quantizer though.

wojtuss: You are right!

wojtuss (approved):

PR #18207, opened 2019-06-19
Use OpenMP to optimize im2col

yihuaxu:
According to the performance issue of im2col, to use OpenMP optimize it.


luotao1:
It seems 5x slower than before.

ghost:
Close without merge.

PR #18214, opened 2019-06-19
Enable MKL-DNN for slim FP32 vs. INT8 tests

wojtuss:
With this patch, MKL-DNN is enabled for `test_slim_int8_*` tests via `FLAGS_use_mkldnn=true` env variable, i.e. the FP32 part of the test runs with MKL-DNN when `FLAGS_use_mkldnn=true`, the INT8 part runs with MKL-DNN either way.

Also, a test case with MKL-DNN is added for MobileNetV1.

test=develop


luotao1 (approved):
LGTM

PR #17965, opened 2019-06-10
[MKL-DNN] Thread-Safety for MKL-DNN reusing Part 1

jczaja:
This PR adress the issue reported in #17611 by making Reusing concept thread safe. Currently only
part of ops were adapted. After sorting out an issue #17960  additional PR for Relu, pooling and INT8 prims will be raised.

@luotao1 I have updated this branch couple of hours ago, so please double check if this PR works for you. If ContentDNN is not working then it means that Relu and other missing ops has to be updated as well. Anyway please share error messages if any.

PR #18081, opened 2019-06-13
create new api to indicate detect thread usage

LeoZhao-BJ:
This PR solves 2 issues:
1. When Predictor.run is called in changed thread, it will make memory leak due to threadid is inserted into key, while each time key is different.
2. For detect model, input dims are dynamic, not fixed, it will make conv/pool/concat mkldnn op memory leak due to each time key is different.

The solve method is to disable cache in this case, we extend EnableMKLDNN in AnaysisConfig, use parameter to control if cache is needed.

related #17611

PR #18217, opened 2019-06-19
[MKL-DNN] Thread ID is added for caching only ParallelExecutor based executions

jczaja:
**This is Work in Progress**

This PR is to limit memory leaks caused by caching MKL-DNN primitives per thread-if as introduced #17965 . Here Parallel Executor still is caching MKL-DNN primitives per thread as there is a thread pool used and threads are reused after finishing tasks to execute next tasks. Change is when using Analysis Predictor then caching is done per mkldnn thread id (As set with setMklDnnThreadID()).

Architectural chart of Key suffix generation for cache

### DIFF
--- a/paddle/fluid/framework/parallel_executor.cc
+++ b/paddle/fluid/framework/parallel_executor.cc
@@ -392,6 +392,22 @@ ParallelExecutor::ParallelExecutor(const std::vector<platform::Place> &places,
     }
 #endif
   }
+
+#if defined(PADDLE_WITH_MKLDNN)
+  // When MKL-DNN is used we are to cache MKL-DNN primitives.
+  // For Parallel Executor this is to be done per std::thread,
+  // so hashing key has to contain Thread ID
+  platform::CPUPlace cpu;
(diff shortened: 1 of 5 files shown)


LeoZhao-BJ:
Firstly, my understanding is, the purpose of this PR is to enable tid in ParallelExecutor and use SetMkldnnThreadID() to enable tid in AnalysisPredictor.  

While there may have few potential issues here:
1. SetMkldnnThreadID is not expose to user, need extend interface, similar work please see PR #18081 
2. If user uses SetMkldnnThreadID, actually the change in SetMkldnnThreadID is meaningless, tid in SetMkldnnThreadID parameter can guarantee distinguishing.
3.  If thread id is not enabled in Face models, it will still have core dump issues due to multi instances of AnalysisPredictor.

LeoZhao-BJ:
BTW, using SetMkldnnThreadID does have some harms, it sets thread_local variable, while when thread is reused, the variable value is stored there, which may make confusion when it is reused by other models, see PR #18081, there is a fix for that.

ghost:
Code will be merged via another PR from Leo.

PR #18223, opened 2019-06-20
fix some bug when merge sparse embedding parameters, test=develop

a6802739:
fix some bug when merge sparse embedding parameters:

1.fix the bug that out_put_var in SaveSelectedRows would be empty string
2.use merge_sparse_lookup_table to replace sum op for load_persistables_for_inference
3.fix the bug in _clone_var_in_block_ when the var is SELECTED_ROWS.


seiriosPlus (approved):
LGTM

PR #18237, opened 2019-06-20
Remove useless functions in dim.h

zhhsplendid:
（Please don't review now as I'm still testing）

As title.

test=develop

### DIFF
--- a/paddle/fluid/framework/dim.h
+++ b/paddle/fluid/framework/dim.h
@@ -45,10 +45,6 @@ class Dim : public Array<int64_t, D> {
   HOSTDEVICE explicit Dim(int64_t head, Args... args)
       : BaseClass(head, args...) {}
 
-  /** Construct a Dim from a linear index and size.  Uses Fortran order
-   * indexing. */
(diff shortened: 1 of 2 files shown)


paddle/fluid/framework/dim.h
```diff
@@ -57,32 +53,6 @@ class Dim : public Array<int64_t, D> {
...
-}  // namespace detail
-
-template <int D>
-HOSTDEVICE Dim<D>::Dim(int64_t idx, const Dim<D>& size) {
-  detail::FortranOrderIndexingConstructorFunctor<0, D, D == 0>::Run(
-      size.Get(), &idx, this->GetMutable());
-}
-
 template <int idx, int D>
 HOSTDEVICE inline int64_t get(const Dim<D>& dim) {
```
sneaxiy: Also remove `get`. I think that nobody would prefer `get` rather than `operator []`.

zhhsplendid:
I close this PR because it is duplicated to JinLe's fix

PR #18255, opened 2019-06-21
Polish backwards optimizer dependency codes and use more default values.

gongweibao:
1. User can set `use_hierarchical_allreduce=True` and code background can select the needed parameters automatically.
2. Fix a bug of launch.py that it should use started_port all.
3. Set  `enable_backward_optimizer_deps` default value to true.
4. Add the hierarchical allreduce unit test.


paddle/fluid/framework/details/build_strategy.cc
```diff
@@ -351,6 +353,12 @@ ir::Graph *BuildStrategy::Apply(ir::Graph *graph,
     } else if (pass->Type() == "mkldnn_placement_pass") {
       pass->Set("mkldnn_enabled_op_types",
                 new std::unordered_set<std::string>(mkldnn_enabled_op_types_));
+    } else if (pass->Type() == "backward_optimizer_op_deps_pass") {
+      if (!use_cuda) {
+        LOG(WARNING) << "backward_optimizer_op_deps_pass is only supported on "
+                        "GPU, skipped.";
```
chengduoZH: Not need to add `LOG(WARNING) << "back ...` here.

gongweibao: Done.

chengduoZH (approved):

PR #18260, opened 2019-06-21
improve the hint message of memory optimize, test=develop

liupluswei:
use six.print_ instead of logging.warn to improve the ambiguous "warning" message


sneaxiy (approved):
LGTM.

PR #18282, opened 2019-06-24
[Cherry-pick #18041] Optimize fused_elewise_activation_grad op.

Xreki:
Cherry-pick https://github.com/PaddlePaddle/Paddle/pull/18041.
Also fix the bug in `fused_elewise_activeation` op for the case when `save_intermedia_out` is false.


phlrain (approved):
LGTM

PR #18287, opened 2019-06-24
Update alloc_continuous_space_for_grad_pass

chengduoZH:
Set persistable attributed for all the same var descs.
The graph may have the same name node. For example, a parameter is the input of operator and it also is the output of optimizer. If the program doesn't set them all, GC will delete the "not persistable variables" by their name which is not what we want.


gongweibao (dismissed):
LGTM

sneaxiy (approved):
LGTM.

PR #18295, opened 2019-06-24
Fix default value of fluid.memory_optimize

chengduoZH:
The gradients of OP_ROLE_VAR should not be renamed.


gongweibao (approved):
LGTM

luotao1 (approved):

python/paddle/fluid/transpiler/memory_optimization_transpiler.py
```diff
@@ -497,7 +497,7 @@ def memory_optimize(input_program,
                     skip_opt_set=None,
                     print_log=False,
                     level=0,
-                    skip_grads=False):
+                    skip_grads=True):
```
qingqing01, line 500: There is no comments for this arg. And why change to True?

xsrobin (approved):

python/paddle/fluid/transpiler/memory_optimization_transpiler.py
chengduoZH: `memory_optimize` reduce the memory to use by renaming the var's name, in some case, we don't want the name of parameters' gradient to be renamed, otherwise, we could not find the gradient of the parameter.

PR #17910, opened 2019-06-07
QAT int8 MKL-DNN transformation pass with MUL 

bingyanghuang:
This PR is to enable the mul op in mkldnn quantization pass in slim. It is based on PR #17819 .

PR #18322, opened 2019-06-25
QAT int8 MKL-DNN transformation pass with MUL

bingyanghuang:
This is the new PR of PR #17910.
This PR is to enable the mul op in mkldnn quantization pass in slim. It works with PR #17834 for QAT support.


bingyanghuang:
@Sand3r- I've followed your comments in PR #17910. 
Is there any other way of obtaining the mul and conv weights rather than by searching for their names directly?
Maybe it can be found by searching `persistable` variable.

python/paddle/fluid/contrib/slim/tests/test_quantization_mkldnn_pass.py
```diff
@@ -162,11 +160,15 @@ def mkldnn_based_freeze_graph(self,
                             activation_quant_type + '_' + weight_quant_type,
                             marked_nodes)
         mkldnn_program = test_graph.to_program()
-        w_mkldnn = np.array(scope.find_var('conv2d_1.w_0').get_tensor())
+
+        # Check the transformation weights of conv2d and mul
+        conv_w_mkldnn = np.array(scope.find_var('conv2d_1.w_0').get_tensor())
+        mul_w_mkldnn = np.array(scope.find_var('fc_0.w_0').get_tensor())
```
Sand3r-, line 166: I still think its unsafe to refer directly to the variable names here, but currently I don't posses enough knowledge on how this could be done differently. We should probably think how to adjust it in the future.

Sand3r- (approved):
LGTM

luotao1:
please update since #17834 merge @bingyanghuang

luotao1:
How about trigger either `Vgg16` or `Vgg19` QAT unit-test in this PR?

PR #18323, opened 2019-06-25
Add install check for multigpu

JiabinYang:
This PR will add multi gpu test


python/paddle/fluid/install_check.py
```diff
@@ -24,11 +25,35 @@
...
+    env = os.environ
+    device_list = []
+    if env.get('CUDA_VISIBLE_DEVICES') is not None:
+        cuda_devices = env['CUDA_VISIBLE_DEVICES']
+        if core.get_cuda_device_count() == 0:
+            logging.warning(
+                "No CUDA Device Found! But your are using GPU version Paddle Fluid"
+            )
+            device_list = []
+        elif len(cuda_devices) == 1:
```
sneaxiy: What if there are more than 10 cards in one machine? It may be `CUDA_VISIBLE_DEVICES=10`.

sneaxiy: This function can be simplify by `return list(range(core.get_cuda_device_count()))`.

PR #18334, opened 2019-06-26
Update reduce config

chengduoZH:
Update reduce config. 
If you set build_strategy.reduce with 'Reduce', the number of places should be 1, the PE will also work.


Xreki (approved):
LGTM

PR #18354, opened 2019-06-26
Fix Bug-prone code of PE

chengduoZH:
1. Fix Bug-prone code of PE.
2. Add drop_local_exe_scopes in compiler.py.


sneaxiy (approved):
LGTM.

PR #18383, opened 2019-06-27
fix py-cpuinfo mac random fail

### LINKED ISSUE #18348: Mac安装报错，ModuleNotFoundError: No module named 'paddle.fluid.core_noavx'
涉及python27 python36 python37
经排查是cpuinfo包返回值不稳定，导致没有走avx分支，noavx没有编译报错。
复现代码
image

image
image
出错代码位置
core.py 
image

tensor-tang:
fix #18348


junjun315 (dismissed):

Dockerfile
```diff
@@ -92,17 +92,17 @@ RUN localedef -i en_US -f UTF-8 en_US.UTF-8
 # specify sphinx version as 1.5.6 and remove -U option for [pip install -U
 # sphinx-rtd-theme] since -U option will cause sphinx being updated to newest
 # version(1.7.1 for now), which causes building documentation failed.
```
luotao1, line 94: please add a comment why `py-cpuinfo==4.0.0` in next PR.

tensor-tang: please check the issue and the reference tells

luotao1: It is more easier for developers to find comments in the codes than use `git blame` to find in the PRs or related issues.

tensor-tang: no need add all comments in our source code, this is not some kind of note or our bug for developers , we only blame when we found errors. Just like other packages with specific version we did not need to add comment one by one.
And the commit description already tells the simple reason.

junjun315 (approved):

PR #18386, opened 2019-06-27
Call the test_slim_int8_* tests through absolute paths

wojtuss:
With this patch, running the `test_slim_int8_*` tests with
```
ctest -R test_slim_int8_googlenet -V
```
will print the test command with an absolute path to the test, so the test command can be called directly from the current directory without having to go to the `build/python/paddle/fluid/contrib/slim/tests/` directory first.

test=develop


sfraczek (approved):
LGTM!

PR #18388, opened 2019-06-27
[cherry-pick] Simplify multi_box_head API in detection.py and remove assign op.

qingqing01:
Simplify multi_box_head API in detection.py and remove assign op.


heavengate (approved):
LGTM

PR #18390, opened 2019-06-28
add MultiSlotStringDataGenerator for speedup of string based user inp…

guru4elephant:
test=develop

This PR aims to speed up DataGenerator when users' input is string based array. 
For example, 
```
class SyntheticData(MultiSlotDataGenerator):
    def generate_sample(self, line):
        def data_iter():
            for i in range(10000):
                yield ("words", [1, 2, 3, 4]), ("label", [0])

        return data_iter


class SyntheticStringData(MultiSlotStringDataGenerator):
    def generate_sample(self, line):
        def data_iter():
            for i in range(10000):
                yield ("words", ["1", "2", "3", "4"], ("label", ["0"]))
```
If a user's input data is from text file, and training data can be generated without numerical preprocessing on input data, a faster way to generate data is to inherit MultiSlotStringDataGenerator. In real world application, for example, a ctr task with hundreds of feature slots, MultiSlotStringDataGenerator can speedup IO by 100%.

### DIFF
--- a/python/paddle/fluid/incubate/data_generator/__init__.py
+++ b/python/paddle/fluid/incubate/data_generator/__init__.py
@@ -265,7 +265,8 @@ class MultiSlotStringDataGenerator(DataGenerator):
         '''
         if not isinstance(line, list) and not isinstance(line, tuple):
             raise ValueError(
-                "the output of process() must be in list or tuple type")
+                "the output of process() must be in list or tuple type"
+                "Examples: [('words', ['1926', '08', '17']), ('label', ['1'])]")
         output = ""
         for index, item in enumerate(line):
             name, elements = item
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/incubate/data_generator/__init__.py
```diff
@@ -235,6 +235,49 @@ def local_iter():
...
+
+        Args:
+            line(str): the output of the process() function rewritten by user.
+
+        Returns:
+            Return a string data that can be read directly by the MultiSlotDataFeed.
+        '''
+        if not isinstance(line, list) and not isinstance(line, tuple):
+            raise ValueError(
+                "the output of process() must be in list or tuple type")
```
seiriosPlus: add an example such as "[("words", ["1926", "08", "17"]), ("label", ["1"])]" in ValueError？

guru4elephant: added. thx

seiriosPlus (approved):
LGTM

PR #18392, opened 2019-06-28
supports collective training with programs

gavin1332:
test=develop

1. Since allreduce op has 4 reduce types, We split these four reduce types into four ops;
2. We also refined the collective op code, e.g. we separated the collective op kernel into CPUKernel and CUDAKernel, and remove the device specified DeviceContext parameter in template as we already knew the target DeviceContext;
3. We remove the newly added Collective op role to reduce the complexity of program and graph analysis.
4. Append a new ParamAttr 'distributed' for distributed parameter specification, which means no grad allreduced for these parameters;


LLMHao (dismissed):

python/paddle/fluid/param_attr.py
```diff
@@ -65,14 +65,16 @@ def __init__(self,
                  regularizer=None,
                  trainable=True,
                  gradient_clip=None,
-                 do_model_average=False):
+                 do_model_average=False,
+                 distributed=False):
```
guru4elephant: do you have example for distributed=True？
I wonder is it possible to inference the value of distributed through op type?

python/paddle/fluid/transpiler/collective.py
```diff
@@ -173,8 +185,8 @@ class GradAllReduce(Collective):
     '''
     '''
 
-    def __init__(self):
-        Collective.__init__(self)
+    def __init__(self, nrings=2):
```
guru4elephant, line 179: why 2?

guru4elephant (commented):
Please take care of the distributed arguments in ParamAttr

guru4elephant (commented):
and also please remove shard_index_op in this PR

python/paddle/fluid/transpiler/collective.py
gavin1332: The minimum number for parallel comms/streams. As it is no harm for paralleling collective communication in GradAllReduce mode, we prefer this value as the default to 1 which refers to no parallel at all.

python/paddle/fluid/param_attr.py
gavin1332: solution is found, so that we will remove this attribute from ParamAttr. And as it is in the Distributed FC domain, we will handle this parameter in the next pr.

gavin1332:
> and also please remove shard_index_op in this PR

as the shard_index_op is also in Distributed FC domain, we have removed it.

paddle/fluid/framework/op_proto_maker.cc
```diff
@@ -75,6 +75,10 @@ void OpProtoAndCheckerMaker::operator()(proto::OpProto* proto,
            static_cast<int>(OpRole::kOptimize) |
                static_cast<int>(OpRole::kLRSched),
            static_cast<int>(OpRole::kCollective),
+           static_cast<int>(OpRole::kCollective) |
+               static_cast<int>(OpRole::kBackward),
+           static_cast<int>(OpRole::kCollective) |
+               static_cast<int>(OpRole::kOptimize),
```
chengduoZH: Op role will increase the complexity of the Graph and Program analysis. I do not recommend adding new o prole.

chengduoZH:
>>  Since allreduce op has 4 reduce types, the type of 'sum' has corresponding gradient calculation, but the others, 'max', 'min', 'prod' do not. So we split these four reduce types into four ops;

If the operator in all reduce is sum, does it mean that all reduce is used for gradient aggregation? What logic is this?

>> As collective ops could be used in backward and optimization phase,

Why? This is an unreasonable assumption.

paddle/fluid/framework/op_proto_maker.cc
gavin1332: i agree with u, and i will try to remove newly added op roles. However, as collective ops swap data among trainers, their behaviors is different from backward and optimize ops more or less, i will create a new pr to discuss this topic if necessary. 3ks.

gavin1332:
> If the operator in all reduce is sum, does it mean that all reduce is used for gradient aggregation? What logic is this?

we are trying to introduce a model parallel strategy to train extreme large classification problem in face recognition, which has up to 10 million classes and the size of the last fc parameter is beyond the GPU memory. So that we have to separate parameter into multiple cards and call collective ops in the forward phase besides the gradient aggregation.

gavin1332:
> > As collective ops could be used in backward and optimization phase,
> 
> Why? This is an unreasonable assumption.

recent research has been done for algorithms to accelerate deep learning training, namely [LocalSGD](https://arxiv.org/abs/1808.07217), which allreduce and average the parameters in the optimization phase instead of allreducing the gradient in the backward phase, which our assumption is based on.

paddle/fluid/framework/op_proto_maker.cc
gavin1332: done

guru4elephant (dismissed):
LGTM. Should add collective op backward ut if possible.

guru4elephant (approved):
LGTM

chengduoZH (approved):

PR #18393, opened 2019-06-28
Fix potential mkldnn concat/pool/conv kernel issues

LeoZhao-BJ:
1. some key generation method is not aligned with PR#17965
2. enlarge ptr lifetime to avoid memory release if SetBlob fails
   otherwise it will get core dump.

test=develop


luotao1 (approved):
LGTM

PR #18372, opened 2019-06-27
[DO NOT MERGE] detect model test2 for dynamic shape

luotao1:
combine #18331 and #18285 
@jianhang-liu @LeoZhao-Intel

PR #18428, opened 2019-07-01
use static variable to do cache instead of thread local in thread frequent switching case

LeoZhao-BJ:
This case is special, NativeExecutor Predictor.run is called in a thread frequent switching case, for mkldnn, TransferScope cache creation is unavoidable. While current cache mechanism is using thread local variable to store cache, if execution thread is changed every iteration, it will always create new memory which makes big memory leak.

The solution is to use a global static variable to handle this case by explicitly indicating from user. Here I reuse SetMkldnnThreadid.

fix https://github.com/PaddlePaddle/Paddle/pull/18372#issuecomment-506929844


paddle/fluid/framework/transfer_scope_cache.cc
```diff
@@ -17,14 +17,31 @@
 namespace paddle {
 namespace framework {
 
+static std::unordered_map<size_t, Scope*>* static_transfer_data_cache = nullptr;
+static std::unordered_set<Scope*>* static_transfer_scope_cache = nullptr;
+
 std::unordered_map<size_t, Scope*>& global_transfer_data_cache() {
-  thread_local auto* x = new std::unordered_map<size_t, Scope*>;
-  return *x;
+  // if get_cur_thread_id() == -1, means not use thread local method to do cache
```
luotao1: Add more comments `platform::get_cur_thread_id() == -1` is only used in MKLDNN.
OR: 
```
#ifdef PADDLE_WITH_CUDA
    thread_local auto* x = new std::unordered_map<size_t, Scope*>;
    return *x;
#else
#ifdef PADDLE_WITH_MKLDNN
   if (platform::get_cur_thread_id() == -1) {
      if (!static_transfer_data_cache)
         static_transfer_data_cache = new std::unordered_map<size_t, Scope*>;
   }
#endif
#endif
```
?

LeoZhao-BJ: good idea, to be updated.

LeoZhao-BJ: it is not special mkldnn issue, but this solution only works with mkldnn api.

LeoZhao-BJ:
Using static variable solution and API is to make it simple, otherwise if we use single globe structure, it makes more complicated, it need thread sync etc, and efficiency will drop.

luotao1:
@LeoZhao-Intel, Discussed with @Superjomn :
1) Do you have some comments for "-1", or do you have some unit-test to ensure this "-1"?
2) Do you need add lock for `static_transfer_scope_cache`? When multi-instance use "-1" mode, it will fail on this.

LeoZhao-BJ:
> @LeoZhao-Intel, Discussed with @Superjomn :
> 
> 1. Do you have some comments for "-1", or do you have some unit-test to ensure this "-1"?

-1 is just specific for mkldnn single instance but with thread frequent switching, so far we don't have unit-test.

> 2. Do you need add lock for `static_transfer_scope_cache`? When multi-instance use "-1" mode, it will fail on this.

In mkldnn single instance but with thread frequent switching, it only supports one instance, yes, it will fail if with multi-instances. We do have plan to refine this solution in the future.

jczaja:
@luotao1 , @Superjomn Why GPU execution needs to store TransferScope in TLS, why each thread has to keep its own copy of scopes and data caches eg. what is wrong with having only one global data and scope cache?

luotao1:
> Why GPU execution needs to store TransferScope in TLS? why each thread has to keep its own copy of scopes and data caches eg. what is wrong with having only one global data and scope cache?

@jczaja CUDADeviceContext is thread_local, i.e. one GPU-card have one CUDADeviceContext.

paddle/fluid/framework/transfer_scope_cache.cc
```diff
@@ -18,42 +18,62 @@ namespace paddle {
...
-    return *static_transfer_data_cache;
-  } else {
-#endif
-    thread_local auto* x = new std::unordered_map<size_t, Scope*>;
-    return *x;
-#ifdef PADDLE_WITH_MKLDNN
+  size_t tid = static_cast<size_t>(platform::get_cur_thread_id());
+
+  // if tid != 0, means there is specific mkldnn tid setting from user.
+  if (tid != 0) {
```
luotao1: How about use CONST value to describe "0" or "-1" mode, such as `constexpr char kEnableCacheRuntimeContext` etc. You can put these CONST value somewhere.

paddle/fluid/framework/transfer_scope_cache.cc
```diff
@@ -17,12 +17,61 @@
...
+using transfer_scope_cache_map = std::unordered_set<Scope*>;
+static std::unordered_map<size_t, transfer_data_cache_map*>
+    static_transfer_data_caches;
+static std::unordered_map<size_t, transfer_scope_cache_map*>
+    static_transfer_scope_caches;
+#endif
+
 std::unordered_map<size_t, Scope*>& global_transfer_data_cache() {
+#ifdef PADDLE_WITH_MKLDNN
+  size_t tid = static_cast<size_t>(platform::get_cur_thread_id());
```
luotao1: get_cur_thread_id->get_cur_mkldnn_session_id

paddle/fluid/framework/transfer_scope_cache.cc
```diff
@@ -17,12 +17,61 @@
...
+
+    auto map_it = static_transfer_scope_caches.find(tid);
+    if (map_it == static_transfer_scope_caches.end()) {
+      auto* x = new transfer_scope_cache_map;
+      static_transfer_scope_caches[tid] = x;
+      return *x;
+    } else {
+      return *static_transfer_scope_caches[tid];
+    }
+  }
```
luotao1, line 73: I will help you add a unit-test here.

LeoZhao-BJ:
I guess, it is similar with mkldnn cache, to support multi-instance, thread_local is used to simply the implementation.

paddle/fluid/framework/transfer_scope_cache.cc
LeoZhao-BJ: new PR #18453 is submitted

paddle/fluid/framework/transfer_scope_cache.cc
LeoZhao-BJ: change in PR #18453 .

LeoZhao-BJ: solved

paddle/fluid/framework/transfer_scope_cache.cc
LeoZhao-BJ: solved

paddle/fluid/inference/tests/api/analyzer_bert_tester.cc
```diff
@@ -251,6 +251,7 @@ TEST(Analyzer_bert, transfer_scope_cache) {
     threads.emplace_back([&, i]() {
       std::getline(fin, line);
       ParseLine(line, &input);
+      if (is_static) platform::set_cur_mkldnn_session_id(1);
```
luotao1, line 257: Please add a TODO here, we will hide directly use of `platform::set_cur_mkldnn_session_id(1)` later.

paddle/fluid/inference/tests/api/analyzer_bert_tester.cc
```diff
@@ -264,9 +265,21 @@ TEST(Analyzer_bert, transfer_scope_cache) {
...
   // paddle::framework::global_transfer_data_cache() are thread_local,
   // their pointer should be different among different thread id.
-  PADDLE_ENFORCE(global_transfer_scope_cache.size(), threads_num);
-  PADDLE_ENFORCE(global_transfer_data_cache.size(), threads_num);
+  if (is_static) {
+    PADDLE_ENFORCE(global_transfer_scope_cache.size(), 1);
+    PADDLE_ENFORCE(global_transfer_data_cache.size(), 1);
+  } else {
+    PADDLE_ENFORCE(global_transfer_scope_cache.size(), threads_num);
+    PADDLE_ENFORCE(global_transfer_data_cache.size(), threads_num);
```
luotao1, line 281: line 265-267 is the comment for line 272 and 273. please add comments for  line269-270

paddle/fluid/inference/tests/api/analyzer_bert_tester.cc
LeoZhao-BJ: done

paddle/fluid/inference/tests/api/analyzer_bert_tester.cc
LeoZhao-BJ: done

luotao1 (dismissed):
LGTM

luotao1:
```
[14:29:15]	/home/teamcity/work/e84e6e698a3f913d/paddle/fluid/inference/tests/api/analyzer_bert_tester.cc:256:32: error: no member named 'set_cur_mkldnn_session_id' in namespace 'paddle::platform'
[14:29:15]	      if (is_static) platform::set_cur_mkldnn_session_id(1);
[14:29:15]	                     ~~~~~~~~~~^
```
In mac_ci

paddle/fluid/inference/tests/api/analyzer_bert_tester.cc
```diff
@@ -261,12 +264,26 @@ TEST(Analyzer_bert, transfer_scope_cache) {
...
     std::vector<PaddleTensor>().swap(input);
   }
-  // Since paddle::framework::global_transfer_scope_cache() and
-  // paddle::framework::global_transfer_data_cache() are thread_local,
-  // their pointer should be different among different thread id.
-  PADDLE_ENFORCE(global_transfer_scope_cache.size(), threads_num);
-  PADDLE_ENFORCE(global_transfer_data_cache.size(), threads_num);
+  if (is_static) {
+    // Use static method to do transfer_scope_cache() instead of thread_local
+    // so paddle::framework::global_transfer_data_cache() should be same
```
luotao1: should be same->should always be 1

LeoZhao-BJ: done

luotao1 (approved):
LGTM

luotao1:
@LeoZhao-Intel Could you revert this PR since #18578 merge and there is no random fail CI on MKLDNN?

LeoZhao-BJ:
> @LeoZhao-Intel Could you revert this PR since #18578 merge and there is no random fail CI on MKLDNN?

ok, will revert it soon.

LeoZhao-BJ:
see new PR #18879 for revert,

PR #18441, opened 2019-07-02
make fleet support mpi job submit directly

guru4elephant:
test=develop
support qsub to submit mpi job, depend on mpi4py. 
```bash
mpirun -npernode 2 python trainer.py
```
will startup 2 process one for trainer, the other one for server.
A user has to use MPISymetricRoleMaker to initialize fleet.


gavin1332 (approved):
LGTM

PR #18451, opened 2019-07-02
test=develop, fix docker with paddle nccl problem

JiabinYang:
This PR will fix docker with paddle installed crash into nccl issue


yma-admin:
这个问题是什么？具体影响是？

PR #18454, opened 2019-07-02
add shape_blob for cache mkldnn primitive

luotao1:
### previous:
The previous cache mkldnn primitive are two-level:
https://github.com/PaddlePaddle/Paddle/blob/840969327271159ef46917508f6c50a004ba9e62/paddle/fluid/platform/device_context.h#L381-L382
```
BlobMap = Map<cur_thread_id, KeyBlob>
KeyBlob  = Map<blob_name, blob>
```
### This PR
This PR implements the three-level cache framework, i.e insert a `ShapeBlob` cache between KeyBlob and BlobMap for cache dynamic input shape.
```
BlobMap = Map<cur_thread_id, ShapeBlob>
ShapeBlob = Map<cur_input_shape_str, KeyBlob>
KeyBlob  = Map<blob_name, blob>
```
In this PR, `cur_input_shape_str` is default null string now. Cache mkldnn primitive based on dynamic input shape is not taken effect. (TODO in next PRs), i.e. the same as before.

### TODO works:
- The inference interface to trigger dynamic input shape setting.
- Cache cleaning strategy for dynamic input shape.


luotao1:
@LeoZhao-Intel Please take a review!

LeoZhao-BJ:
How about combine tid with shape as first level key? given these 2 variable are all thread local, and should be ready before run(), it may save once search and data structure. 
or do you think it is hard to find all shapes belonging to one tid if combine together? if this cache clearing is global basis, I think that's fine to combine, but this clearing mechanism is just tid basis, then that's fine to seperate.

luotao1:
> if this cache clearing is global basis, I think that's fine to combine, but this clearing mechanism is just tid basis, then that's fine to separate.

Is the clearing mechanism must tid basis, if we want to support multi-instance clearing mechanism?

luotao1:
@jczaja Could you take a review for this PR, thanks very much!

jczaja (approved):
LGTM

My understanding is that some cache clearing mechanics based on cur shape will follow those changes?

luotao1:
Yes, you can refer to https://github.com/PaddlePaddle/Paddle/pull/18372/files#diff-d5f6c6a79232fe26298554b79ea1108fR455
```
if ((tid == -1) && (sBlob->size() > cur_input_shape_cache_size)) {
      VLOG(2) << "tid=" << tid
              << ", remove all head blob of shape: " << sBlob->begin()->first;
      sBlob->erase(sBlob->begin()->first);
    }
    pBlob = std::shared_ptr<KeyBlob>(new KeyBlob());
    (*sBlob)[cur_input_shape_str] = pBlob;
}
```
Will remove the first shape_blob when its size is larger than `cur_input_shape_cache_size`.

PR #18459, opened 2019-07-02
Refactor for Thread Check

hutuxian:
move the thread-check code from train_from_dataset to a single function
add ut for the thread check function
test=develop


gavin1332 (approved):
LGTM

PR #18469, opened 2019-07-02
Optimize LayerNorm grad operator with intrinsic function

yihuaxu:
According to the training status of bert model, the intrinsic function was used to optimize the operator and to improve the performance of layer norm grad operator. it can speed up about 3x (avx2 instruction set) on bert training model. 

**Platform**: Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz
**Model Path**: [benchmark/NeuralMachineTranslation/BERT/fluid/train/chinese_L-12_H-768_A-12](https://bert-models.bj.bcebos.com/chinese_L-12_H-768_A-12.tar.gz)
**Batch Size**: 32
**CPU NUM**: 20
**OMP_THREADS_NUM**: 4
**Command**: CPU_NUM=20 OMP_NUM_THREADS=4 python -u run_classifier.py --task_name XNLI --use_cuda false --do_train true --do_val true --do_test true --batch_size 32 --in_tokens False --init_pretraining_params  ../../chinese_L-12_H-768_A-12/params --data_dir ../../data --vocab_path ../../chinese_L-12_H-768_A-12/vocab.txt --checkpoints ../../save --save_steps 1000 --weight_decay 0.01 --warmup_proportion 0.1 --validation_steps 1000 --epoch 2 --max_seq_len 128 --bert_config_path ../../chinese_L-12_H-768_A-12/bert_config.json --learning_rate 5e-5 --skip_steps 10 --random_seed 1
**Data Source**: [benchmark/NeuralMachineTranslation/BERT/fluid/train/data/XNLI-MT-1.0.zip](https://bert-data.bj.bcebos.com/XNLI-MT-1.0.zip)
                      [benchmark/NeuralMachineTranslation/BERT/fluid/train/data/XNLI-1.0.zip](https://bert-data.bj.bcebos.com/XNLI-1.0.zip)

The following is the comparison with the different scenarios.
image

If you try to enable avx512 part, please use the below patch.

diff --git a/cmake/configure.cmake b/cmake/configure.cmake
index 5f7b4a4..d2ee77b 100644
--- a/cmake/configure.cmake
+++ b/cmake/configure.cmake
@@ -28,7 +28,10 @@ if(NOT WITH_PROFILER)
     add_definitions(-DPADDLE_DISABLE_PROFILER)
 endif(NOT WITH_PROFILER)

-if(WITH_AVX AND AVX_FOUND)
+if(WITH_AVX AND AVX512F_FOUND)
+set(SIMD_FLAG ${AVX512F_FLAG})
+add_definitions(-DPADDLE_WITH_AVX)
+elseif(WITH_AVX AND AVX_FOUND)
     set(SIMD_FLAG ${AVX_FLAG})
     add_definitions(-DPADDLE_WITH_AVX)
 elseif(SSE3_FOUND)


yihuaxu:
@jczaja Please help me review this PR.

jczaja:
@yihuaxu 
1. In ..jit/intrinsics/layer_norm.cc there seems to be very similar functions: LayerNormGrad_impl_avx 
and LayerNormGrad_impl_avx512 eg. they differ by registersize used eg. YMM and ZMM. Could it be possible to merge them into one eg. make one template based function that depending on size given eg. 128, 256, 512 would generate specialzed code for each intruction set?
2. Due to lack of my expertise in PaddlePaddle JIT infrastrucutre, could @tensor-tang  please take a look at this review as well?

yihuaxu:
> @yihuaxu
> 
> **1. In ..jit/intrinsics/layer_norm.cc there seems to be very similar functions: LayerNormGrad_impl_avx**
>    **and LayerNormGrad_impl_avx512 eg. they differ by registersize used eg. YMM and ZMM. Could it be possible to merge them into one eg. make one template based function that depending on size given eg. 128, 256, 512 would generate specialzed code for each intruction set?**
As the difference of avx and avx512 intrinsic function name, if we merge two function as one, it will result in it be exist more macros(such as #ifdef, #else and #endif) and show the function ugly. and if we use template to import "256" and its like as one parameters, it still can not replace "__m256" and "__m512" dynamically except we use marco segment.
> **2. Due to lack of my expertise in PaddlePaddle JIT infrastrucutre, could @tensor-tang  please take a look at this review as well?**
I will invite Tensor-tang to review this PR.

yihuaxu:
@tensor-tang Could you help me review this PR? thank you very much.

yihuaxu:
Now I have two directions to reduce the code size. one is use the marco according the size(such as 256 or 512), another is use jit via xbyak. @tensor-tang do you prefer it?

tensor-tang:
Sorry for the later reply, 

>  eg. make one template based function that depending on size given eg. 128, 256, 512 would generate specialzed code for each intruction set

I thought template should be fine as well.

bingyanghuang:
Close this PR because:

1. This PR is too old to maintain

2. We are working on layer_norm MKL-DNN FP32 support

3. If this layer_norm grad needed in the future, will create the new PR.

PR #18477, opened 2019-07-03
Set the outputs' lod to the first step_input's lod in DynamicRNN.

Xreki:
As title. But when there are many step_inputs and outputs, step_input[0]'s lod may not be the needed lod of outputs[i]. I'm doubted about this PR.


Xreki:
The bug is fixed in #19144 , so close this PR.

PR #18479, opened 2019-07-03
core remove package pycpuinfo

tensor-tang:
删除对python包py-cpuinfo的依赖
1. 减轻框架对第三方python的依赖。
2. 该包在MAC上判断AVX flag不太稳定。

### DIFF
--- a/python/paddle/fluid/core.py
+++ b/python/paddle/fluid/core.py
@@ -16,7 +16,6 @@ from __future__ import print_function
 
 import sys
 import os
-from cpuinfo import get_cpu_info
 
 core_suffix = 'so'
 if os.name == 'nt':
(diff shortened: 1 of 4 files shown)


tensor-tang:
@tianshuo78520a Please help to review this PR

junjun315 (approved):
checked： `python/requirements.txt`

python/paddle/fluid/core.py
```diff
@@ -57,27 +56,112 @@
...
+                    'sysctl machdep.cpu.leaf7_features | grep -i avx').read(
+                    ) != ''
+            except Exception as e:
+                sys.stderr.write(
+                    'Can not get the AVX flag from machdep.cpu.leaf7_features.\n'
+                    'The original error is: %s\n' %
+                    cpt.get_exception_message(e))
+        return has_avx
+    elif sysstr == 'windows':
+        import ctypes
```
JiabinYang, line 96: will this package be an dependency?

JiabinYang (approved):
LGTM

python/paddle/fluid/core.py
tensor-tang: It seems in `The Python Standard Library` https://docs.python.org/3.6/library/ctypes.html

python/paddle/fluid/core.py
```diff
@@ -57,27 +56,112 @@
...
+        ONE_PAGE = ctypes.c_size_t(0x1000)
 
-if has_avx:
+        def asm_func(code_str, restype=ctypes.c_uint32, argtypes=()):
```
wopeizl, line 99: not sure whether this would violate the DAP on windows

tensor-tang: It should be fine ctypes calls the windll, and protect the memory https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualprotect

PR #18512, opened 2019-07-05
Add sequence_topk_pooling Op

Aurelius84:
- **sequence_topk_pooling**:

Example:
```python
import numpy as np
from paddle.fluid import layers
x_lod_tensor = layers.data(name='x', shape=[1], lod_level=1)
out = layers.sequence_topk_pooling(input=x_lod_tensor, topk=3, channel_num=5)
```

sequence_top_pooling


paddle/fluid/operators/sequence_ops/sequence_topk_pooling_op.cc
```diff
@@ -0,0 +1,131 @@
...
+    PADDLE_ENFORCE(ctx->HasInput("X"), "The input X should not be null.");
+
+    ctx->ShareDim("X", /*->*/ framework::GradVarName("X"));
+    ctx->ShareLoD("X", /*->*/ framework::GradVarName("X"));
+  }
+
+ protected:
+  framework::OpKernelType GetExpectedKernelType(
+      const framework::ExecutionContext& ctx) const override {
+    // return framework::OpKernelType(
```
lanxianghit: please remove the useless commented codes, here and other place

paddle/fluid/operators/sequence_ops/sequence_topk_pooling_op.cc
```diff
@@ -0,0 +1,131 @@
...
+    AddInput("X",
+             "(LoDTensor) The variable-length input of SequenceTopkPoolingOp");
+    AddOutput(
+        "Out",
+        "(Tensor) The output of SequenceTopkPoolingOp does not contain LoD "
+        "infomation.");
+    AddOutput("pos", "(Tensor<int>) store the topk index ").AsIntermediate();
+    AddAttr<int>("topk", "topk attr");
+    AddAttr<int>("channel_num", "channel number");
+    AddAttr<bool>("is_test",
```
lanxianghit: 'is_test' seems useless? please remove it if so.

paddle/fluid/operators/sequence_ops/sequence_topk_pooling_op.cc
Aurelius84: Fixed. Thanks for your review.

PR #18513, opened 2019-07-05
add mkldnn shapeblob cache clear strategy

luotao1:
### This PR
1. add mkldnn shapeblob cache clear strategy: When `shapeblob.size() == mkldnn_input_shape_cache_size`, erase the first one of shapeblob.
2. add a unit-test `TestMkldnnCacheClear` to ensure this cache clear strategy
   - If not use cache clear strategy, `shapeblob.size()` always = 1.
   - If use cache clear strategy, `shapeblob.size() = mkldnn_input_shape_cache_size`.
3. remove unused `platform::get_cur_input_shape_str` function, and add `size_t MKLDNNDeviceContext::GetShapeBlobSize(int mkldnn_session_id)` to get the ShapeBlob size by mkldnn_session_id.

### TODO
1. explicit following settings such as `platform::set_cur_input_shape_str(ss.str());` and `platform::set_cur_mkldnn_session_id(platform::kMKLDNNSessionID_CacheClearing);`, will be deprecated after enhance `config.EnableMKLDNN()` interface.
2. for global function  such as `platform::set_cur_input_shape_str(ss.str());`, `platform::set_cur_mkldnn_session_id(platform::kMKLDNNSessionID_CacheClearing);` etc, make them become Class function of `MKLDNNDeviceContext`.

### DIFF
--- a/paddle/fluid/platform/device_context.cc
+++ b/paddle/fluid/platform/device_context.cc
@@ -407,9 +407,9 @@ thread_local size_t cur_mkldnn_session_id = kMKLDNNSessionID_Default;
 // - For fixed-shape, it's a null string in default.
 // - For dynamic-shape, it's user specific.
 thread_local std::string cur_input_shape_str = "";
-// the cache size of different input shapes for MKLDNN.
+// the cache capacity of different input shapes for MKLDNN.
 // Default 1 means fixed input shape, not dynamic shape.
-thread_local int cur_input_shape_cache_size = 1;
+thread_local int cur_input_shape_cache_capacity = 1;
 }  // namespace
 
 void set_cur_mkldnn_session_id(size_t sid) { cur_mkldnn_session_id = sid; }
@@ -417,18 +417,18 @@ size_t get_cur_mkldnn_session_id(void) { return cur_mkldnn_session_id; }
 void set_cur_input_shape_str(std::string input_shape_str) {
   cur_input_shape_str = input_shape_str;
 }
-void set_cur_input_shape_cache_size(int input_shape_cache_size) {
-  cur_input_shape_cache_size = input_shape_cache_size;
+void set_cur_input_shape_cache_capacity(int input_shape_cache_capacity) {
+  cur_input_shape_cache_capacity = input_shape_cache_capacity;
 }
 
 void MKLDNNDeviceContext::ResetBlobMap() const { p_blobmap_->clear(); }
 
-size_t MKLDNNDeviceContext::GetShapeBlobSize(int mkldnn_session_id) const {
+size_t MKLDNNDeviceContext::GetShapeBlobSize() const {
   BlobMap* pMap = p_blobmap_.get();
-  auto map_it = pMap->find(mkldnn_session_id);
+  auto map_it = pMap->find(cur_mkldnn_session_id);
   if (map_it == pMap->end()) {
-    LOG(FATAL) << "MKLDNNDeviceContext don't find mkldnn_session_id : "
-               << mkldnn_session_id;
+    LOG(FATAL) << "MKLDNNDeviceContext don't find cur_mkldnn_session_id : "
+               << cur_mkldnn_session_id;
   }
   return map_it->second->size();
 }
@@ -439,18 +439,18 @@ void MKLDNNDeviceContext::SetBlob(const std::string& name,
   std::shared_ptr<ShapeBlob> sBlob = nullptr;
   std::shared_ptr<KeyBlob> pBlob = nullptr;
 
-  int tid = platform::get_cur_mkldnn_session_id();
+  int sid = platform::get_cur_mkldnn_session_id();
 
   std::lock_guard<std::mutex> lock(*p_mutex_);
 
-  // Find ShapeBlob for current thread
-  auto map_it = pMap->find(tid);
+  // Find ShapeBlob for current mkldnn session id.
+  auto map_it = pMap->find(sid);
 
   if (map_it == pMap->end()) {
     // 1st time to set blob in current thread
     sBlob = std::shared_ptr<ShapeBlob>(new ShapeBlob());
-    (*pMap)[tid] = sBlob;
-    VLOG(2) << "SetBlob: tid=" << tid << ", add new tid\n";
+    (*pMap)[sid] = sBlob;
+    VLOG(2) << "SetBlob: sid=" << sid << ", add new sid\n";
   } else {
     sBlob = map_it->second;
   }
@@ -459,12 +459,13 @@ void MKLDNNDeviceContext::SetBlob(const std::string& name,
   auto key_it = sBlob->find(cur_input_shape_str);
 
   if (key_it == sBlob->end()) {
-    // In cache clearing mode, cur_input_shape_cache_size defines max pblob
-    // capacity
-    if ((tid == kMKLDNNSessionID_CacheClearing) &&
-        (sBlob->size() == static_cast<size_t>(cur_input_shape_cache_size))) {
-      VLOG(2) << "tid=" << tid
-              << ", remove all head blob of shape: " << sBlob->begin()->first;
+    // In cache clearing mode, cur_input_shape_cache_capacity defines
+    // max pblob capacity
+    if ((sid == kMKLDNNSessionID_CacheClearing) &&
+        (sBlob->size() ==
+         static_cast<size_t>(cur_input_shape_cache_capacity))) {
+      VLOG(2) << "sid=" << sid
+              << ", remove all blobs of shape: " << sBlob->begin()->first;
       sBlob->erase(sBlob->begin()->first);
     }
     pBlob = std::shared_ptr<KeyBlob>(new KeyBlob());
@@ -480,7 +481,7 @@ void MKLDNNDeviceContext::SetBlob(const std::string& name,
   } else {
     blob_it->second = data;  // set data to existing blob
   }
-  VLOG(2) << "SetBlob: tid=" << tid << ", add blob=" << name << "\n";
+  VLOG(2) << "SetBlob: sid=" << sid << ", add blob=" << name << "\n";
   // lock will be automatically released when out of scope
   return;
 }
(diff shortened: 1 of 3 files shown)


luotao1:
@LeoZhao-Intel Please take a review!

paddle/fluid/platform/device_context.cc
```diff
@@ -407,17 +407,32 @@ thread_local size_t cur_mkldnn_session_id = kMKLDNNSessionID_Default;
 // - For fixed-shape, it's a null string in default.
 // - For dynamic-shape, it's user specific.
 thread_local std::string cur_input_shape_str = "";
+// the cache size of different input shapes for MKLDNN.
+// Default 1 means fixed input shape, not dynamic shape.
+thread_local int cur_input_shape_cache_size = 1;
```
LeoZhao-BJ: how about cur_input_shape_cache_capacity instead of size?  size here is a kind of real data size, while it is for max cache size in my understanding.

paddle/fluid/platform/device_context.cc
```diff
@@ -407,17 +407,32 @@ thread_local size_t cur_mkldnn_session_id = kMKLDNNSessionID_Default;
...
 }  // namespace
 
 void set_cur_mkldnn_session_id(size_t sid) { cur_mkldnn_session_id = sid; }
 size_t get_cur_mkldnn_session_id(void) { return cur_mkldnn_session_id; }
 void set_cur_input_shape_str(std::string input_shape_str) {
   cur_input_shape_str = input_shape_str;
 }
-std::string get_cur_input_shape_str(void) { return cur_input_shape_str; }
+void set_cur_input_shape_cache_size(int input_shape_cache_size) {
```
LeoZhao-BJ: set_cur_input_shape_cache_capacity?

paddle/fluid/platform/device_context.cc
```diff
@@ -407,17 +407,32 @@ thread_local size_t cur_mkldnn_session_id = kMKLDNNSessionID_Default;
...
+  cur_input_shape_cache_size = input_shape_cache_size;
+}
 
 void MKLDNNDeviceContext::ResetBlobMap() const { p_blobmap_->clear(); }
 
+size_t MKLDNNDeviceContext::GetShapeBlobSize(int mkldnn_session_id) const {
```
LeoZhao-BJ: why need a explicit input parameter mkldnn_session_id? can we just get it from thread_local? seems there is no case for user getting shapeblobsize for other session id with different id in thread_local

paddle/fluid/platform/device_context.cc
```diff
@@ -441,10 +456,17 @@ void MKLDNNDeviceContext::SetBlob(const std::string& name,
   }
 
   // Find KeyBlob for current input shape
-  std::string cur_input_shape_str = platform::get_cur_input_shape_str();
   auto key_it = sBlob->find(cur_input_shape_str);
 
   if (key_it == sBlob->end()) {
+    // In cache clearing mode, cur_input_shape_cache_size defines max pblob
+    // capacity
+    if ((tid == kMKLDNNSessionID_CacheClearing) &&
```
LeoZhao-BJ: better rename tid to sid (session id) to align with kMKLDNNSessionID_xxx

paddle/fluid/platform/device_context.h
```diff
@@ -389,14 +389,14 @@ using ShapeBlob = std::unordered_map<std::string, std::shared_ptr<KeyBlob>>;
 using BlobMap = std::unordered_map<int, std::shared_ptr<ShapeBlob>>;
 
 // default mkldnn session id
-constexpr size_t kMKLDNNSessionID_Default = 0;
+constexpr int kMKLDNNSessionID_Default = 0;
```
LeoZhao-BJ: why change to int? size_t is used for aligning with std::hash(this_thread::gettid()), otherwise it may need a static_cast

paddle/fluid/platform/device_context.cc
```diff
@@ -441,10 +456,17 @@ void MKLDNNDeviceContext::SetBlob(const std::string& name,
...
+        (sBlob->size() == static_cast<size_t>(cur_input_shape_cache_size))) {
+      VLOG(2) << "tid=" << tid
+              << ", remove all head blob of shape: " << sBlob->begin()->first;
+      sBlob->erase(sBlob->begin()->first);
```
LeoZhao-BJ, line 470: why here remove sBlob->begin()->first? seems it is wrong,  should be erase(sBlob->begin()),  sBlob->begin()->first is string.
BTW, it doesn't really mean removing the head or first one because sBlob is std::unordered_map and its index method is not same with vector or queue,.

paddle/fluid/platform/device_context.cc
luotao1: Got it. I will change.

paddle/fluid/platform/device_context.cc
luotao1: Got it. I will change.

paddle/fluid/platform/device_context.cc
luotao1: Got it. I will change.

paddle/fluid/platform/device_context.h
luotao1: Got it. I will change.

paddle/fluid/platform/device_context.cc
luotao1: > can we just get it from thread_local

Could you give more details on how to get it from thread_local?

> why need a explicit input parameter mkldnn_session_id? 

if we don't have `mkldnn_session_id` parameter, line428 `pMap->find(mkldnn_session_id)` could not find. Or do you mean there is only one session_id in `pMap`?

LeoZhao-BJ: mkldnn_session_id is already stored in cur_mkldnn_session_id, it is a thread_local variable, so if user call it in same thread with predictor.run, it can easily get id from cur_mkldnn_session_id.

LeoZhao-BJ: if user call getblobsize in same thread with predictor.run, then code is like:
```
  BlobMap* pMap = p_blobmap_.get();
  auto map_it = pMap->find(cur_mkldnn_session_id);
  if (map_it == pMap->end()) {
    LOG(FATAL) << "MKLDNNDeviceContext don't find mkldnn_session_id : "
               << mkldnn_session_id;
  }
  return map_it->second->size();
}
```
If we still want user to call this function in other threads, then this parameter is necessary.

paddle/fluid/platform/device_context.cc
luotao1: map could erase by key, see: https://www.geeksforgeeks.org/map-erase-function-in-c-stl/

>  it doesn't really mean removing the head or first one because sBlob is 

Got it. I will change the LOG. For code, since we can erase any one of the sBlob, it runs successfully.

paddle/fluid/platform/device_context.cc
luotao1: Got it. I will change it.

paddle/fluid/platform/device_context.cc
LeoZhao-BJ: you are right, there are 3 erase functions.

luotao1:
All done, please review again @LeoZhao-Intel

paddle/fluid/platform/device_context.cc
```diff
@@ -459,12 +459,13 @@ void MKLDNNDeviceContext::SetBlob(const std::string& name,
...
-    // In cache clearing mode, cur_input_shape_cache_size defines max pblob
-    // capacity
-    if ((tid == kMKLDNNSessionID_CacheClearing) &&
-        (sBlob->size() == static_cast<size_t>(cur_input_shape_cache_size))) {
-      VLOG(2) << "tid=" << tid
-              << ", remove all head blob of shape: " << sBlob->begin()->first;
+    // In cache clearing mode, cur_input_shape_cache_capacity defines
+    // max pblob capacity
+    if ((sid == kMKLDNNSessionID_CacheClearing) &&
+        (sBlob->size() ==
```
LeoZhao-BJ: do you think">=" is better  here ?

luotao1: For batchsize=1, when `==`, it will clean. Thus `>` does not work. 
But for batchsize >1, I think `>=` is better. I will change it.

luotao1: when `==`, it will clean. Thus `>` does not work.
What's the case when `>` works?

LeoZhao-BJ: so far I still not see cases for >, but from code level, it looks safer to use >=.

luotao1: Got it.

LeoZhao-BJ:
LGTM

luotao1:
@jczaja Please take a review!

paddle/fluid/platform/device_context.cc
```diff
@@ -407,44 +407,67 @@ thread_local size_t cur_mkldnn_session_id = kMKLDNNSessionID_Default;
...
-std::string get_cur_input_shape_str(void) { return cur_input_shape_str; }
+void set_cur_input_shape_cache_capacity(int input_shape_cache_capacity) {
+  cur_input_shape_cache_capacity = input_shape_cache_capacity;
+}
 
 void MKLDNNDeviceContext::ResetBlobMap() const { p_blobmap_->clear(); }
 
+size_t MKLDNNDeviceContext::GetShapeBlobSize() const {
+  BlobMap* pMap = p_blobmap_.get();
+  auto map_it = pMap->find(cur_mkldnn_session_id);
```
jczaja, line 429: currently it is only for UT purposes? Perhaps it would be safer to guard whole function with critical section , the same mutex as for SetBlob and GetBlob. In particular if there is plan that to use GetShapeBlobSize with parallel executor. Apart from that LGTM

luotao1: Got it. You are right.

luotao1:
@LeoZhao-Intel @jianhang-liu The number of `analyzer_mm_dnn` hang increases in this PR after adding a unit-test here.
http://ci.paddlepaddle.org/viewType.html?buildTypeId=Paddle_PrCi&branch_Paddle=pull%2F18513&tab=buildTypeStatusDiv
Is the reason of https://github.com/PaddlePaddle/Paddle/pull/18513#discussion_r300788047?

luotao1:
@jczaja @LeoZhao-Intel I add the mutex, and `analyzer_mm_dnn` don't hang when I repeat 3 times.
http://ci.paddlepaddle.org/viewType.html?buildTypeId=Paddle_PrCi&branch_Paddle=pull%2F18513&tab=buildTypeStatusDiv
image
Please review again!

LeoZhao-BJ:
LGTM

ghost (approved):
Please consider whether we really need VLOG(2) considering those log are just detail information of cache usage.

luotao1:
> Please consider whether we really need VLOG(2) considering those logs are just detail information of cache usage.

Since those logs are used by MKLDNN only, what's your opinion about it? @LeoZhao-Intel @jczaja

LeoZhao-BJ:
I am fine with VLOG(2), but not sure if there is a clear rule for log level definition.

PR #18528, opened 2019-07-05
test=release/1.5, cherry-pick hide not_support for dygraph

JiabinYang:
This PR cherry-picked from develop and to hide `not_support` api cause it should not be used by user

### DIFF
--- a/python/paddle/fluid/clip.py
+++ b/python/paddle/fluid/clip.py
@@ -21,7 +21,7 @@ import functools
 from . import layers
 from . import framework
 from . import core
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/clip.py
```diff
@@ -21,7 +21,7 @@
 from . import layers
 from . import framework
 from . import core
-from .dygraph import not_support
+from .dygraph import _not_support
```
junjun315: should be from: .dygraph.base

PR #18529, opened 2019-07-05
Fix topk cannot handle 1D vector bug

zhaoyuchen2018:
[cherry-pick] Add path to handle 1D vector(#18466)


tensor-tang (approved):
The title should be more clear that this is a cherry-pick.

PR #18532, opened 2019-07-05
Inference: fix mask rcnn model diff, optim  memory usage, memory leak.

NHZlX:
1. fix mask rcnn 
2. add interface for setting optim_cache_dir(eg: when in trt int8 mode, and load model from memory, there should be a interface for setting the trt calibration table data dir)


paddle/fluid/operators/controlflow/conditional_block_infer_op.cc
```diff
@@ -0,0 +1,74 @@
...
+      scope.DeleteScope(scopes->front());
+    }
+  }
+};
+
+}  // namespace operators
+}  // namespace paddle
+
+namespace ops = paddle::operators;
+REGISTER_OPERATOR(conditional_block_infer, ops::ConditionalBlockInferOp,
```
Superjomn, line 72: 这个 op 在哪儿替换呢？

NHZlX: 有两个文件给漏掉了，已经添加。

Superjomn (approved):
LGTM

PR #18533, opened 2019-07-06
upgrade collective fleet api

guru4elephant:
test=develop
```python
from paddle.fluid.incubate.fleet.collective import fleet, DistributedStrategy
import paddle.fluid.incubate.fleet.base.role_maker as role_maker

# total loss definination

role = role_maker.PaddleCloudRoleMaker(is_collective=True)
fleet.init(role)
strategy = DistributedStrategy()
optimizer = fluid.optimizer.SGD(learning_rate=0.001)
opts, param_and_grads = optimizer.minimize(total_loss)

```

### DIFF
--- a/python/paddle/fluid/incubate/fleet/collective/__init__.py
+++ b/python/paddle/fluid/incubate/fleet/collective/__init__.py
@@ -34,6 +34,7 @@ class DistributedStrategy(object):
         self.h_allreduce = False
 
     def build(self):
+        self.strategy_map = {}
         # make sure we set single precision config True
         if self.use_fp32 and self.use_fp16:
             self.use_fp16 = False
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/incubate/fleet/collective/__init__.py
```diff
@@ -231,32 +165,51 @@ def minimize(self,
...
+        t.transpile(
+            trainer_id=fleet.worker_index(),
+            trainers=fleet.worker_endpoints(),
+            current_endpoint=fleet.worker_endpoints()[fleet.worker_index()],
+            startup_program=startup_program,
+            program=loss.block.program)
+        return opts, param_and_grads
+
+
+class FP32SGDOptimizer(CollectiveOpBasedOptimizer):
```
gavin1332, line 191: GradAllreduce is not only suitable for SGD but most gradient-based optimization, isn't it?

gavin1332 (commented):
GradAllreduce is not only suitable for SGD but most other gradient-based optimizations, isn't it?

python/paddle/fluid/incubate/fleet/collective/__init__.py
guru4elephant: of course. It is only a symbol for inner usage and is not visible to users. SGDOptimizer is not meant to be the sgd optimizer. Similarly, I don't think localsgd only works with sgd optimizer.

gavin1332: That's OK

gavin1332 (approved):
LGTM

PR #18542, opened 2019-07-08
Support Bitmain Anakin

Shixiaowei02:
1. update anakin apis of bitmain arch.
2. fix definition bug in cmake.
test=develop


fc500110 (approved):
LGTM

PR #18549, opened 2019-07-08
enhance config.EnableMKLDNN api for mkldnn cache clear strategy

luotao1:
1. enhance config.EnableMKLDNN api for mkldnn cache clear strategy: `EnableMKLDNN(int mkldnn_input_shape_cache_capacity = 0)`
2. simplify the `TEST(Analyzer_MM_DNN, mkldnn_cache_clear)` with the enhancement api, and add output compare between no cache strategy and using cache strategy.

### DIFF
--- a/paddle/fluid/inference/api/analysis_predictor.cc
+++ b/paddle/fluid/inference/api/analysis_predictor.cc
@@ -360,10 +360,10 @@ void AnalysisPredictor::PrepareArgument() {
   argument_.SetStaticMemoryOptimForceUpdate(
       config_.static_memory_optim_force_update_);
   argument_.SetModelFromMemory(config_.model_from_memory_);
-  argument_.SetEngineOptInfo(config_.engine_opt_info_);
   // Analyze inference_program
   argument_.SetUseAnakin(config_.anakin_engine_enabled());
   argument_.SetPredictorID(predictor_id_);
+  argument_.SetOptimCacheDir(config_.opt_cache_dir_);
   if (!config_.model_dir().empty()) {
     argument_.SetModelDir(config_.model_dir());
   } else {
--- a/paddle/fluid/inference/api/analysis_predictor.cc
+++ b/paddle/fluid/inference/api/analysis_predictor.cc
@@ -185,10 +185,37 @@ bool AnalysisPredictor::PrepareExecutor() {
   return true;
 }
 
+void AnalysisPredictor::MkldnnPreRun(const std::vector<PaddleTensor> &inputs) {
+#ifdef PADDLE_WITH_MKLDNN
+  VLOG(2) << "AnalysisPredictor::Run get_cur_mkldnn_session_id="
+          << platform::get_cur_mkldnn_session_id();
+  // In cache clearing mode.
+  if (config_.mkldnn_input_shape_cache_capacity_ > 0) {
+    VLOG(2) << "In mkldnn cache clear mode.";
+    platform::set_cur_mkldnn_session_id(
+        platform::kMKLDNNSessionID_CacheClearing);
+    platform::set_cur_input_shape_cache_capacity(
+        config_.mkldnn_input_shape_cache_capacity_);
+    // Set current_input_shape for caching dynamic shape.
+    std::stringstream ss;
+    for (size_t i = 0; i < inputs.size(); ++i) {
+      for (size_t j = 0; j < inputs[i].shape.size(); ++j) {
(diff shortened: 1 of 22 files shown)


paddle/fluid/inference/api/analysis_predictor.cc
```diff
@@ -185,10 +185,35 @@ bool AnalysisPredictor::PrepareExecutor() {
...
+  }
+}
+
 bool AnalysisPredictor::Run(const std::vector<PaddleTensor> &inputs,
                             std::vector<PaddleTensor> *output_data,
                             int batch_size) {
   paddle::platform::SetNumThreads(config_.cpu_math_library_num_threads());
+#ifdef PADDLE_WITH_MKLDNN
+  if (config_.use_mkldnn_) MkldnnPreRun(inputs);
+#endif
```
luotao1, line 228: Compared with #18372, the reason of don't use `MkldnnPostRun` is:  if reset the `mkldnn_session_id` to 0, the unit-test `dev_ctx->GetShapeBlobSize()` could not get the correct shape_blob size.

paddle/fluid/platform/device_context.cc
```diff
@@ -462,7 +462,8 @@ void MKLDNNDeviceContext::SetBlob(const std::string& name,
   if (key_it == sBlob->end()) {
     // In cache clearing mode, cur_input_shape_cache_capacity defines
     // max pblob capacity
-    if ((sid == kMKLDNNSessionID_CacheClearing) &&
+    if ((static_cast<size_t>(sid) == kMKLDNNSessionID_CacheClearing) &&
+        sBlob->size() &&
```
luotao1, line 466: Enhance it for `cur_input_shape_cache_capacity=1` and `sBlob.size()==0`

luotao1:
@LeoZhao-Intel @jczaja Please take a review!

paddle/fluid/inference/api/analysis_predictor.cc
LeoZhao-BJ: There may be a corner case, e.g. if thread is reuse with pool, in last execution, instance X config.mkldnn_input_shape_cache_capacity_ is set >0, then thread A is set thread local cache capacity and this variable is not cleared after execution, but when this thread A is reused by another instance B with config.mkldnn_input_shape_cache_capacity_ = 0, it will hit wrong branch.

paddle/fluid/pybind/inference_api.cc
```diff
@@ -250,7 +250,8 @@ void BindAnalysisConfig(py::module *m) {
       .def("tensorrt_engine_enabled", &AnalysisConfig::tensorrt_engine_enabled)
       .def("switch_ir_debug", &AnalysisConfig::SwitchIrDebug,
            py::arg("x") = true)
-      .def("enable_mkldnn", &AnalysisConfig::EnableMKLDNN)
+      .def("enable_mkldnn", &AnalysisConfig::EnableMKLDNN,
+           py::arg("mkldnn_input_shape_cache_capacity") = 0)
       .def("mkldnn_enabled", &AnalysisConfig::mkldnn_enabled)
       .def("set_cpu_math_library_num_threads",
            &AnalysisConfig::SetCpuMathLibraryNumThreads)
```
LeoZhao-BJ: there may be another failure in CI, see in my PR #18081.   https://github.com/PaddlePaddle/Paddle/pull/18081/files?file-filters%5B%5D=.py#diff-876ea1bc109973488c161a657f79812fR74  , but it may be fixed in your PR.

paddle/fluid/inference/api/analysis_predictor.cc
LeoZhao-BJ: so suggest to do like below in MkldnnPreRun:

```
  if (config_.mkldnn_input_shape_cache_capacity_ > 0) {
    VLOG(2) << "In mkldnn cache clear mode.";
    platform::set_cur_mkldnn_session_id(
        platform::kMKLDNNSessionID_CacheClearing);
    platform::set_cur_input_shape_cache_capacity(
        config_.mkldnn_input_shape_cache_capacity_);
  }

  // Set current_input_shape .
  std::stringstream ss;
  for (size_t i = 0; i < inputs.size(); ++i) {
    for (size_t j = 0; j < inputs[i].shape.size(); ++j) {
      ss << inputs[i].shape[j] << "-";
    }
  }
  VLOG(2) << "Set input shape=" << ss.str();
  platform::set_cur_input_shape_str(ss.str());
```

PR #18568, opened 2019-07-09
Remove reshape2_grad infershape special handling 

baojun-nervana:
There was a reshape2_grad infershape special handling in ngraph engine as it has a dimension element of 0. This PR is to remove that special handling. reshape, transpose and gather ops are updated accordingly.

### DIFF
--- a/paddle/fluid/operators/ngraph/ngraph_engine.cc
+++ b/paddle/fluid/operators/ngraph/ngraph_engine.cc
@@ -38,14 +38,8 @@ namespace operators {
 
 static ngraph::Shape Ddim2Shape(const framework::DDim& dims) {
   ngraph::Shape sp;
-  if (dims.size() == 1 && dims[0] == 0) {
-    sp.emplace_back(0);
-    return sp;
-  }
   for (int i = 0; i < dims.size(); ++i) {
-    int k = dims[i];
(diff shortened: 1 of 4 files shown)


mozga-intel (approved):
LGTM!

tensor-tang (commented):
Could you add some unit test to cover the case?

baojun-nervana:
> Could you add some unit test to cover the case?

The dim 0 is an"always-do", so it is covered by current UT. 

https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/reshape_op.cc#L321
https://github.com/PaddlePaddle/Paddle/blob/7a4fb0c1b7575482a59faebac9789bdb9423ddb3/paddle/fluid/operators/ngraph/ops/transpose_op.h#L55

tensor-tang:
> special handling in ngraph engine as it has a dimension element of 0.

so you mean just change(not add or remove) the way handing the dims of 0?

baojun-nervana:
> > special handling in ngraph engine as it has a dimension element of 0.
> 
> so you mean just change(not add or remove) the way handing the dims of 0?

Yes, the special handling of reshape2_grad infershape was removed in this PR

tensor-tang (approved):
LGTM

PR #18570, opened 2019-07-10
Extend Matmul to support matrix multiplication with multiple heads

czhu15:
With the support of multiple head, the multiplication of two big matrixes is split into multiplication of several (head_number) small matrixes. e.g. if Mat A is [3, 24] and Mat B is [24, 4], when multiple A and B with head_number as 4, Mat A will be split as 4 matrix of [3, 6] and Mat B will be 4 matrix of [6, 4]. The result of final matrix will be 4 matrix of [3, 4], i.e. [3, 16].

Use this new extension, we can avoid head split/merge in Transformer.

first part of #16342


bingyanghuang:
@yihuaxu Please help review this PR. @wojtuss could you help assign someone in your team to help review this PR?

yihuaxu:
> @yihuaxu Please help review this PR. @wojtuss could you help assign someone in your team to help review this PR?

LGTM

paddle/fluid/operators/matmul_op.cc
```diff
@@ -60,7 +60,18 @@ class MatMulKernel : public framework::OpKernel<T> {
     auto mat_dim_b = math::CreateMatrixDescriptor(
         ColumnMatrixFromVector(y.dims()), 0, context.Attr<bool>("transpose_Y"));
     auto scale = static_cast<T>(context.Attr<float>("alpha"));
+
+#if defined(PADDLE_WITH_MKLML) && !defined(PADDLE_WITH_CUDA)
+    int head_number = context.Attr<int>("head_number");
+    if (1 == head_number) {
+      blas.MatMul(x, mat_dim_a, y, mat_dim_b, scale, out, T(0));
+    } else {
+      blas.MatMulWithHead(x, mat_dim_a, y, mat_dim_b, scale, head_number, out,
```
wojtuss, line 69: What about adding `if (head_number > 1)` here, without `else` and without `#else`?

paddle/fluid/operators/matmul_op.cc
```diff
@@ -295,16 +306,33 @@ class MatMulOp : public framework::OperatorWithKernel {
                      mat_dim_x.batch_size_ == 0 || mat_dim_y.batch_size_ == 0);
     }
     std::vector<int64_t> dim_out;
+#ifdef PADDLE_WITH_MKLML
+    int head_number = context->Attrs().Get<int>("head_number");
+    PADDLE_ENFORCE_GE(head_number, 1);
+    PADDLE_ENFORCE_LE(head_number, mat_dim_x.width_);
+#endif
```
wojtuss, line 316: Here, with something like:
```
    int64_t dim_out_y = head_number * mat_dim_y.width_;
#else
    int64_t dim_out_y = mat_dim_y.width_;
#endif
```
you could avoid multiple `#ifdef`s below, like this
```
    dim_out[dim_out.size() - 1] = dim_out_y;
```

paddle/fluid/operators/matmul_op.cc
```diff
@@ -360,13 +392,19 @@ Examples without transpose:
...
+
 The behavior is designed to be similar to the `numpy.matmul` function.
 The differences are:
 - When the rank of the input data is less than or equal to 3, it
   is similar to the `numpy.matmul` function.
 - When the rank of the input is greater than 3, the rank of X and
   Y must be equal, and the first `rank - 2` dimensions must be equal.
 - We add `transpose_X` and `transpose_Y` flags.
+- We add `head_number` attribute, which is used to multiple two matrixes head
+  by head, and eventually concatenates the output of several(head_number)
```
wojtuss: Missing space: `several (head_number)`.

paddle/fluid/operators/matmul_op.cc
```diff
@@ -360,13 +392,19 @@ Examples without transpose:
 - X: [B, M, K], Y: [B, K, N] => Out: [B, M, N]
 - X: [B, ..., M, K], Y: [B, ..., K, N] => Out: [B, ..., M, N]
 
+Example of matrix multiple with head_number of H
```
wojtuss: Maybe: `Example of matrix multiplication with head number H` ?

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
```diff
@@ -0,0 +1,161 @@
...
+                              Y,
+                              head_number=1,
+                              transpose_X=False,
+                              transpose_Y=False):
+    """Reference forward implementation using np.matmul."""
+    # np.matmul does not support the transpose flags, so we manually
+    # transpose X and Y appropriately.
+    if transpose_X:
+        if X.ndim == 1:
+            X = X.reshape((X.size, 1))
```
wojtuss: Is it necessary to reshape 1-dimensional arrays?

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
```diff
@@ -0,0 +1,161 @@
...
+        elif X.ndim == 2:
```
wojtuss: This case is also handled by the `else` case.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
```diff
@@ -0,0 +1,161 @@
...
+            X = X.T
+        else:
+            dim = [i for i in range(len(X.shape))]
```
wojtuss: The `else` case could be simplified a liitle bit:
```
dim = np.arange(X.dim)
dim[[-1, -2]] = dim[[-2, -1]]
X = np.transpose(X, tuple(dim))
```

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
```diff
@@ -0,0 +1,161 @@
...
+            dim[-1], dim[len(X.shape) - 2] = dim[len(X.shape) - 2], dim[-1]
+            X = np.transpose(X, tuple(dim))
+    if transpose_Y:
+        if Y.ndim == 1:
+            Y = Y.reshape((1, Y.size))
+        else:
+            dim = [i for i in range(len(Y.shape))]
```
wojtuss: IMO, this duplicated code lends itself for wrapping into a small function.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
```diff
@@ -0,0 +1,161 @@
...
+        # Everywhere else, we are compatible with np.matmul.
+        Out = np.array([Out], dtype="float32")
+    return Out
+
+
+# Generator for multiple head
+class GeneratorMulHead(object):
+    def setUp(self):
+        self.op_type = "matmul"
+        X = np.random.random(self.shape_X).astype("float32")
```
wojtuss, line 119: Using random values (without setting seed) makes the test non-repeatable.

paddle/fluid/operators/math/blas.h
```diff
@@ -112,6 +112,15 @@ class Blas {
 
   template <typename T>
   void GEMM_FREE(T* data) const;
+
+#if !defined(PADDLE_WITH_CUDA)
```
wojtuss, line 116: Sometimes `#if !defined(PADDLE_WITH_CUDA)` and sometimes `#ifdef PADDLE_WITH_MKLML` are used for `MatMulWithHead`.

paddle/fluid/operators/matmul_op.cc
czhu15: Yes. You are right. In normal if head_number > 1, this means we should call MatMulWithHead(). But  in current design, the attribute of  head_number only exist for MKLML and non-CUDA environments. So i have to use macro to limit this judgement.

paddle/fluid/operators/math/blas.h
czhu15: You are right. I will make the macro consistent for multiple head.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: Thanks for your comments. Here I re-used the original python/paddle/fluid/tests/unittests/test_matmul_op.py code.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: Thanks for your comments. Here I re-used the original python/paddle/fluid/tests/unittests/test_matmul_op.py code.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: Thanks for your comments. Here I re-used the original python/paddle/fluid/tests/unittests/test_matmul_op.py code.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: Thanks for your comments. Here I re-used the original python/paddle/fluid/tests/unittests/test_matmul_op.py code.

paddle/fluid/operators/matmul_op.cc
czhu15: thanks. will be fixed.

paddle/fluid/operators/matmul_op.cc
czhu15: thanks. will be fixed.

paddle/fluid/operators/matmul_op.cc
czhu15: Good idea. Your way makes code much cleaner :)
Will update it accordingly.

paddle/fluid/operators/matmul_op.cc
czhu15: Yes. You are right. In normal if head_number > 1, this means we should call MatMulWithHead(). But in current design, the attribute of head_number only exist for MKLML and non-CUDA environments. So i have to use macro to limit this judgement.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
wojtuss: OK, but please consider writing better code than in the original test :-)

wojtuss: Feel free also to improve the original test.

paddle/fluid/operators/matmul_op.cc
wojtuss: Sorry, I meant leaving out the `else` only and keeping the `#else`.

czhu15: Sorry still not quite get your idea on how to remove the else and keep the #else. 
Even using `if (head_number > 1)`, we have to consider the case when `head_number == 1` (this could be the default case when`defined(PADDLE_WITH_MKLML) && !defined(PADDLE_WITH_CUDA)` is true but multiple head is not used. so we still need an else in this case.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: Hi wojtuss,

After digging into the original code, I think I found the reason why this logic (reshape 1 dim. arrays) exists. Below is some code snip I did to evaluate the output with and without reshape when Y need be transposed.
```python
import numpy as np
X = np.array([1, 2, 3])

Y = np.array([4, 5, 6])    # assume we need transpose Y
Y_T = Y.reshape((Y.size, 1))

Out = np.matmul(X, Y)      # if no reshape
Out_T = np.matmul(X, Y_T)  # with reshape

print(Out)                 # 32
print(Out_T)               # [32]
```
If no reshape, the output is a scalar, which is not supported by paddle according to the comments from line 79, test_matmul_op.py
```C
        # We do not support 0-dimensional Tensors (scalars). So where
        # np.matmul outputs a scalar, we must convert to a Tensor of
        # shape (1, ) instead.
```
BTW, reshaping 1 dim array if it need be transposed seems natural. With reshaping, the original Y (array([4, 5, 6])) was converted to array([[4],
       [5],
       [6]])
This more matches what I was taught during school:) though np.array can handle this case correctly (but it broke the scalar limitation I mentioned above).
So I prefer to keep this logic as the original implementation.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: you are right. will remove this `elif `case.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: agree your code looks more concise. will update it accordingly, and for Y also.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: agree. will extra these codes into a small function.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
czhu15: I would prefer to keep same test method as common matmul test cases.

paddle/fluid/operators/matmul_op.cc
wojtuss: Right, I mixed it up, sorry.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
wojtuss: I understand. Here, I do not consider it to be a big problem, just wanted to make sure you are aware of it.

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
wojtuss: I know that from the scholar point of view a 1-dimensional array can be transposed, but that is perfectly handled by numpy and you do not need to do that manually.
As for scalar outputs, the `if` branch in the line 108 (https://github.com/PaddlePaddle/Paddle/pull/18570/files#diff-5a8f116278fbc29c2462c4827a5e75e6R108) is written exactly to handle the case of 0-dimentional outputs.

czhu15: OK. will remove the reshaping of 1-dim array.

paddle/fluid/operators/math/blas_impl.h
```diff
@@ -627,6 +662,67 @@ void Blas<DeviceContext>::MatMul(const framework::Tensor &mat_a,
         dim_a.stride_, dim_b.stride_);
   }
 }
+
+#if defined(PADDLE_WITH_MKLML) && !defined(PADDLE_WITH_CUDA)
```
wojtuss, line 666: This `#if` is still inconsistent with `#ifdef` for `MatMulWithHead()` declaration.

paddle/fluid/operators/matmul_op.cc
```diff
@@ -60,7 +60,18 @@ class MatMulKernel : public framework::OpKernel<T> {
     auto mat_dim_b = math::CreateMatrixDescriptor(
         ColumnMatrixFromVector(y.dims()), 0, context.Attr<bool>("transpose_Y"));
     auto scale = static_cast<T>(context.Attr<float>("alpha"));
+
+#if defined(PADDLE_WITH_MKLML) && !defined(PADDLE_WITH_CUDA)
```
wojtuss, line 64: This #if is still inconsistent with #ifdef for MatMulWithHead() declaration.

paddle/fluid/operators/math/blas.h
wojtuss: This `#if` is still inconsistent with other `#if`s for code with `MatMulWithHead()`.

czhu15: there is already a `#ifdef PADDLE_WITH_MKLML` outside the definition of `MatMulWithHead`, like below:
```C
#ifdef PADDLE_WITH_MKLML
...

#if !defined(PADDLE_WITH_CUDA)
  template <typename T>
  void MatMulWithHead(const framework::Tensor& mat_a,
                      const MatDescriptor& dim_a,
                      const framework::Tensor& mat_b,
                      const MatDescriptor& dim_b, T alpha, int head_number,
                      framework::Tensor* mat_out, T beta) const;
#endif
#endif
```
so they are consistent with other code:)

paddle/fluid/operators/math/blas_impl.h
czhu15: there is already a `#ifdef PADDLE_WITH_MKLML` outside the definition of `MatMulWithHead`, like below:
```C
#ifdef PADDLE_WITH_MKLML
...

#if !defined(PADDLE_WITH_CUDA)
  template <typename T>
  void MatMulWithHead(const framework::Tensor& mat_a,
                      const MatDescriptor& dim_a,
                      const framework::Tensor& mat_b,
                      const MatDescriptor& dim_b, T alpha, int head_number,
                      framework::Tensor* mat_out, T beta) const;
#endif
#endif
```
so they are consistent with other code:)

paddle/fluid/operators/matmul_op.cc
czhu15: there is already a `#ifdef PADDLE_WITH_MKLML` outside the definition of `MatMulWithHead`, like below:
```C
#ifdef PADDLE_WITH_MKLML
...

#if !defined(PADDLE_WITH_CUDA)
  template <typename T>
  void MatMulWithHead(const framework::Tensor& mat_a,
                      const MatDescriptor& dim_a,
                      const framework::Tensor& mat_b,
                      const MatDescriptor& dim_b, T alpha, int head_number,
                      framework::Tensor* mat_out, T beta) const;
#endif
#endif
```
so they are consistent with other code:)

paddle/fluid/operators/math/blas.h
wojtuss: OK, that's fine then. Thanks!

paddle/fluid/operators/math/blas_impl.h
wojtuss: OK

paddle/fluid/operators/matmul_op.cc
wojtuss: OK

wojtuss (approved):
LGTM

python/paddle/fluid/tests/unittests/test_matmul_op_with_head.py
```diff
@@ -0,0 +1,158 @@
...
+    M = 3
+    N = 4
+    K = 24
+    if (dim_X == 1 and transpose_X) or (dim_Y == 1 and transpose_Y):
+        K = 1
+    if dim_X == 1:
+        if transpose_X:
+            shape_X = [M]
+        else:
+            shape_X = [K]
```
luotao1, line 33: reduce 4 lines to 1 line.
`shape_X = [M] if  transpose_X else [K]`
You can update in your next PR

luotao1 (approved):
LGTM

Issue #17275, opened 2019-05-08
Fix depthwise_conv kernel to ensure it has same effect on GPU and CPU


JiabinYang:
问题： 在CPU和GPU上 `depthwise_conv`的行为不一致，导致在GPU上训练的模型在CPU下预测出现diff

PR #18582, opened 2019-07-10
Fix depthwise conv gpu kernel bug

### LINKED ISSUE #2509: Use std::variant instead of boost::variant
It seems that the difference between std::variant and boost::variant https://stackoverflow.com/a/40204093/724872 shouldn't affect our code.

If we can do so, It seems that we don't have to rely on boost.

LielinJiang:
resolve #2509，#17275


qingqing01 (commented):
Please add more unit testing

LielinJiang:
> Please add more unit testing

add depthwise conv3x3, 5x5, 7x7 with different attribute.

qingqing01 (approved):

JiabinYang (approved):
Nice Catch!

PR #18596, opened 2019-07-11
init new amp

Jie-Fang:
In the previous mixed precision module, we have to cast input to fp16 first, and store fp16 and fp32 weights in the startup program, and cast fp32 to fp16 weights after updating. Thanks to the black/white list, we don't need to consider master weight copy and the input can maintain fp32. Because we will rewrite the program and insert appropriate cast ops for fp16 execution which will cause inserting cast op for fp16 gradients converting to fp32 in the backward procedure, and updating will occur on fp32 weight. Another advantage is that we don't need to write other functions for saving and loading pretrained fp32 model. Because our parameters are always kept in fp32.

### DIFF
--- a/python/paddle/fluid/contrib/mixed_precision/decorator.py
+++ b/python/paddle/fluid/contrib/mixed_precision/decorator.py
@@ -17,7 +17,6 @@ from ... import default_startup_program
 from ... import layers
 from ... import unique_name
 from . import fp16_utils
-from .fp16_utils import create_master_params_grads, master_param_to_train_param
 from .fp16_utils import update_loss_scaling, rewrite_program
 from .fp16_lists import AutoMixedPrecisionLists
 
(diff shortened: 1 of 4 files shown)


Jie-Fang:
I tested the some perf between old amp version and this new amp version.

The number of cast op after applying amp

| Model     | old amp | new amp |
| --------- | ------- | ------- |
| ResNet50  | 541     | 114     |
| VGG16     | 71      | 68      |
| GoogleNet | 292     | 288     |

The speed of FP16 and FP32 (sec/step)

| Model     | batch size | old amp                | new amp                |
| --------- | ---------- | ---------------------- | ---------------------- |
| ResNet50  | 128        | FP16: 0.32  FP32: 0.39 | FP16: 0.21  FP32: 0.39 |
| VGG16     | 64         | FP16: 0.16  FP32: 0.29 | FP16: 0.16  FP32: 0.29 |
| GoogleNet | 128        | FP16: 0.11  FP32: 0.12 | FP16: 0.11  FP32: 0.12 |

Because ResNet50 have many batchnorm layer, thus it can reduce the number of cast op dramatically which enables faster speed in FP16.

BTW, the new amp can decrease more GPU memory. For example, The reason for setting batchsize 128 for ResNet50 is that the old amp will exceed GPU memory if setting batchsize 256. While the new amp can set batchsize 256.

gongweibao (approved):
LGTM

PR #18607, opened 2019-07-12
Enhance multiclass nms

jerrywgz:
Add one more Output(Index) for multiclass nms and it returns when return_index is True.
Index is the absolute index of selected bbox cross batches.
If the index is used to gather other attribute such as age, one needs to reshape the input(N, M, 1) to (N * M, 1) as first.


jerrywgz:
API doc preview:
detection_output
image
image



multiclass_nms
image
image

python/paddle/fluid/layers/detection.py
```diff
@@ -489,21 +490,29 @@ class number, M is number of bounding boxes. For each category
...
+        If all images have not detected results, LoD will be set to {1}, and 
+        output tensor only contains one value, which is -1.
+        (After version 1.3, when no boxes detected, the lod is changed
+        from {0} to {1}.)
+         
+        Index: Only return when return_index is True. A 2-D LoDTensor with 
+        shape [No, 1] represents the selected index. The index is the absolute 
+        value cross batches. No is the same number as Out. If the index is 
+        used to gather other attribute such as age, one needs to reshape the 
+        input(N, M, 1) to (N * M, 1) as first.
```
qingqing01: 1. If the index is used to gather ...  最后这个注释不是个通用的注释，觉得可以不写。
2. 告诉用户 类型是 int

python/paddle/fluid/layers/detection.py
```diff
@@ -2731,19 +2747,28 @@ class number
...
+        boxes for all images, lod will be set to {1} and Out only contains one 
+        value which is -1.
+        (After version 1.3, when no boxes detected, the lod is changed 
+        from {0} to {1}) 
+
+        Index: Only return when return_index is True. A 2-D LoDTensor with 
+        shape [No, 1] represents the selected index. The index is the absolute 
+        value cross batches. No is the same number as Out. If the index is 
+        used to gather other attribute such as age, one needs to reshape the 
+        input(N, M, 1) to (N * M, 1) as first.
```
qingqing01: 同上

python/paddle/fluid/layers/detection.py
jerrywgz: Done

python/paddle/fluid/layers/detection.py
jerrywgz: Done

paddle/fluid/API.spec
```diff
@@ -376,7 +376,7 @@ paddle.fluid.layers.density_prior_box (ArgSpec(args=['input', 'image', 'densitie
...
 paddle.fluid.layers.bipartite_match (ArgSpec(args=['dist_matrix', 'match_type', 'dist_threshold', 'name'], varargs=None, keywords=None, defaults=(None, None, None)), ('document', '080ce0d54d3f1950ad5a3a8e5ae529e9'))
 paddle.fluid.layers.target_assign (ArgSpec(args=['input', 'matched_indices', 'negative_indices', 'mismatch_value', 'name'], varargs=None, keywords=None, defaults=(None, None, None)), ('document', 'e9685f32d21bec8c013626c0254502c5'))
-paddle.fluid.layers.detection_output (ArgSpec(args=['loc', 'scores', 'prior_box', 'prior_box_var', 'background_label', 'nms_threshold', 'nms_top_k', 'keep_top_k', 'score_threshold', 'nms_eta'], varargs=None, keywords=None, defaults=(0, 0.3, 400, 200, 0.01, 1.0)), ('document', 'efae414c1137c7944d6174dd08c5347a'))
+paddle.fluid.layers.detection_output (ArgSpec(args=['loc', 'scores', 'prior_box', 'prior_box_var', 'background_label', 'nms_threshold', 'nms_top_k', 'keep_top_k', 'score_threshold', 'nms_eta', 'return_index'], varargs=None, keywords=None, defaults=(0, 0.3, 400, 200, 0.01, 1.0, False)), ('document', '664cbfde565e8a3d3c195a27711c0050'))
```
Superjomn, line 379: Do not change the definition of the existing operators, or the different version of training and inference model will be incompatible.

You can add another op called  multi_nms2, and that will add version information to this Op, the latter trained model will use nms2_op, not nms_op; while the old training will save model base on nms_op. 
Both the nms_op and nms2_op are compatible with the new inference, and for the old inference, the nms2 in the newly trained model will remind the users that a new version of inference lib is needed.

Superjomn (changes requested):
Do not change the definition of the existing operators, or the different version of training and inference model will be incompatible.

You can add another op called  multi_nms2, and that will add version information to this Op, the latter trained model will use nms2_op, not nms_op; while the old training will save model base on nms_op. 
Both the nms_op and nms2_op are compatible with the new inference, and for the old inference, the nms2 in the newly trained model will remind the users that a new version of inference lib is needed.

Superjomn (approved):

xsrobin (approved):

jerrywgz:
改动已在https://github.com/PaddlePaddle/Paddle/pull/19228 中合入

PR #18614, opened 2019-07-12
Downgrade gcc to 4.8

JiabinYang:
This PR will downgrade gcc to 4.8 for latest-dev, we test it on these case:
image

And this downgrade will cause link error on Debug mode which I fixed in PR #18558 
Some question has already been answered in PR #18462


Dockerfile
```diff
@@ -166,4 +195,4 @@ RUN mkdir /var/run/sshd
 RUN echo 'root:root' | chpasswd
 RUN sed -ri 's/^PermitRootLogin\s+.*/PermitRootLogin yes/' /etc/ssh/sshd_config
 RUN sed -ri 's/UsePAM yes/#UsePAM yes/g' /etc/ssh/sshd_config
-EXPOSE 22
+EXPOSE 22
```
zhhsplendid: A small format issue, add end of line at last line. Other things LGTM

JiabinYang: Done, thx

PR #18622, opened 2019-07-12
Fix memory leak in test

kbinias:
This PR fixes the memory leak in the module that initiates all unit tests.

The fix is not critical because the system will deallocate the memory when exiting the function, but limits the number of errors in the AddressSanitizer report: `==56712==ERROR: LeakSanitizer: detected memory leaks`

### DIFF
--- a/paddle/testing/paddle_gtest_main.cc
+++ b/paddle/testing/paddle_gtest_main.cc
@@ -85,5 +85,12 @@ int main(int argc, char** argv) {
   char** new_argv_address = new_argv.data();
   google::ParseCommandLineFlags(&new_argc, &new_argv_address, false);
   paddle::framework::InitDevices(true);
+
+  for (int i = argc; i < new_argv.size(); ++i) {
+    if (new_argv[i]) {
+      free(new_argv[i]);
+    }
+  }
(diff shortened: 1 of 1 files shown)


jczaja (dismissed):
LGTM

paddle/testing/paddle_gtest_main.cc
```diff
@@ -61,29 +61,30 @@ int main(int argc, char** argv) {
...
 
+  std::string env_string = "--tryfromenv=";
   if (envs.size() > 0) {
-    std::string env_string = "--tryfromenv=";
     for (auto t : envs) {
       env_string += t + ",";
     }
     env_string = env_string.substr(0, env_string.length() - 1);
-    new_argv.push_back(strdup(env_string.c_str()));
+    new_argv.push_back(const_cast<char*>(env_string.c_str()));
```
luotao1: It is not recommended to use `const_cast`, could you use another way?

kbinias: I prepared other solution (without const_cast)

luotao1 (approved):
LGTM

Issue #18624, opened 2019-07-12
Typing out the specific error types of calling CUDA API


zhwesky2010:
Error log is currently only prompting the call to the CUDA API to fail. You can't see the detailed error information. It's not good for debug. Optimize it as follows:
Typing out the specific error types of calling CUDA API and prompting users to go to nvidia's official website to see the error value explanation.

wopeizl:
感谢建议，我们会在易用性中加以改进

PR #18625, opened 2019-07-12
resolve: #18624

zhwesky2010:
Error log is currently only prompting the call to the CUDA API to fail. You can't see the detailed error information. It's not good for debug. Optimize it as follows:
Typing out the specific error types of calling CUDA API and prompting users to go to nvidia's official website to see the error value explanation.
For example: 
cudaGetDeviceProperties failed in paddle::platform::GetCUDAComputeCapability! 
Error Type ID = 1, Please see detail in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g3f51e3575c2178246db0a94a430e0038

### DIFF
--- a/paddle/fluid/platform/gpu_info.cc
+++ b/paddle/fluid/platform/gpu_info.cc
@@ -111,7 +111,8 @@ int GetCUDAComputeCapability(int id) {
   auto e = cudaGetDeviceProperties(&device_prop, id);
   int a = 10;
   char c[] = "helloworld";
-  PADDLE_ENFORCE(cudaGetDeviceProperties(&device_prop, id), a, c);
+  string str = "helloworld";
+  PADDLE_ENFORCE(cudaGetDeviceProperties(&device_prop, id), a, c, str);
   return device_prop.major * 10 + device_prop.minor;
 }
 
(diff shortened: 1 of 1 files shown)


zhouwei25:
Error log is currently only prompting the call to the CUDA API to fail. You can't see the detailed error information. It's not good for debug. Optimize it as follows:
Typing out the specific error types of calling CUDA API and prompting users to go to nvidia's official website to see the error value explanation.
For Example: cudaGetDeviceProperties failed in paddle::platform::GetCUDAComputeCapability!  Error Type ID = 1, Please see detail in  https://docs.nvidia.com/cuda/cuda-runtime-api/ group__CUDART__TYPES.html#group__CUDART__TYPES_1g3f51e3575c2178246db0a94a430e0038

paddle/fluid/platform/gpu_info.cc
```diff
@@ -79,6 +79,8 @@ DEFINE_string(selected_gpus, "",
 namespace paddle {
 namespace platform {
 
+inline string ErrorLog(cudaError e) {}
```
liupluswei, line 82: what's the purpose of adding this function?

PR #18626, opened 2019-07-13
list element name error

Yogurt2019:
list element name error in  python/paddle/fluid/contrib/slim/core/config.py


adaxi123:
@Yogurt2019 Please add test=develop in each commit info for triggering the CI.
image

PR #18628, opened 2019-07-14
increase timeout again

guru4elephant:
testcase in develop branch can not detect the speed change during development. if our executor has 10-20% slow down, it is more likely to have timeout during test. So just increase the timeout value
test=develop


sandyhouse (approved):
LGTM

PR #18634, opened 2019-07-15
Fix the compiling issue with AVX512F macro

yihuaxu:
When enabled the AVX512 flags, it shows the compile error.  Now we provided this PR for it.

The below patch is to enable AVX512F flag to verify this PR.
```
diff --git a/cmake/configure.cmake b/cmake/configure.cmake
index 5f7b4a4..d2ee77b 100644
--- a/cmake/configure.cmake
+++ b/cmake/configure.cmake
@@ -28,7 +28,10 @@ if(NOT WITH_PROFILER)
add_definitions(-DPADDLE_DISABLE_PROFILER)
endif(NOT WITH_PROFILER)

-if(WITH_AVX AND AVX_FOUND)
+if(WITH_AVX AND AVX512F_FOUND)
+set(SIMD_FLAG ${AVX512F_FLAG})
+add_definitions(-DPADDLE_WITH_AVX)
+elseif(WITH_AVX AND AVX_FOUND)
set(SIMD_FLAG ${AVX_FLAG})
add_definitions(-DPADDLE_WITH_AVX)
elseif(SSE3_FOUND)
```

### DIFF
--- a/paddle/fluid/operators/jit/more/intrinsic/crf_decoding.cc
+++ b/paddle/fluid/operators/jit/more/intrinsic/crf_decoding.cc
@@ -46,7 +46,7 @@ void CRFDecoding(const int seq_len, const float* x, const float* w,
     x_content = _mm512_loadu_ps(x + i_offset);
     alpha_content = _mm512_add_ps(w_content, x_content);
     // Save the alpha value.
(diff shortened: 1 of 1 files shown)


tensor-tang (approved):
LGTM

PR #18638, opened 2019-07-15
Support for distributed classification training

gavin1332:
test=develop


gavin1332:
minimum functional code supports distributed classification training.
There are some files we have not checked in, for their name and location have not been decided yet, including an optimizer and a front-end api named dist_classify

PR #18655, opened 2019-07-16
Change select_rows add from math::scatter::MergeAdd to SelectedRowsAd…

JesseyXujin:
…dTo() on select_rows_function.h.

From profile test, after change it, the compute speed is 20 times faster than before.
test=develop
Before change:
profile2
After change:
profile1

### DIFF
--- a/paddle/fluid/operators/sum_op.h
+++ b/paddle/fluid/operators/sum_op.h
@@ -16,6 +16,7 @@ limitations under the License. */
 #include "paddle/fluid/framework/op_registry.h"
 #include "paddle/fluid/operators/math/math_function.h"
 #include "paddle/fluid/operators/math/selected_rows_functor.h"
+#include "paddle/fluid/platform/profiler.h"
 
 namespace paddle {
 namespace operators {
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/sum_op.h
```diff
@@ -37,9 +37,8 @@ void SelectedRowsCompute(const framework::ExecutionContext &context) {
     return;
   }
 
-  std::vector<const paddle::framework::SelectedRows *> inputs;
+  std::vector<paddle::framework::SelectedRows *> inputs;
```
seiriosPlus, line 40: why not `const`

paddle/fluid/operators/sum_op.h
```diff
@@ -50,34 +49,52 @@ void SelectedRowsCompute(const framework::ExecutionContext &context) {
     for (size_t i = 1; i < in_vars.size(); ++i) {
       auto &in = in_vars[i]->Get<SelectedRows>();
       if (in.rows().size() > 0) {
-        inputs.push_back(&in);
+          temp_in0.set_height(in.height());
```
seiriosPlus, line 52: 这样写， inplace是不是就失效了？

Issue #17654, opened 2019-05-26
[MKL-DNN] Add Leaky-relu support


jczaja:
Leaky-relu is supported by MKL-DNN , but not supported by PaddlePaddle MKL-DNN integration.
Adding that once could improve performance of Face model (reduce number of reorders + hopefully more efficent mkl-dnn leaky relu implementation)

Scope of task:
1) Add interface to integration for exposing mkl-dnn leaky relu
2) Add Pass : conv + leaky_relu -> conv_with_post_op_leaky_relu

bingyanghuang:
@jczaja Could you help close this PR. Since #17535 has been closed too. Thanks.

luotao1:
I guess this issue is another thing different from #17535. @bingyanghuang

jczaja:
This one is separate thing its execution will be beneficial for performance of face model, but it does not have anything in common with #17535

jczaja:
Executed by #18656 and merged

Issue #17960, opened 2019-06-10
Customize OpMaker for MKL-DNN builds


jczaja:
Hi,

I want to solve following problem:
Activations of paddle (Grad ops) are using inputs: Out, Out@GRAD   while  mkl-dnn grad primitives  activations need: X  (input signal to fwd op)  . eg. MKL-DNN grad activation ops are expecting input of forward to be available.
So far we used SetBlob to have it transfered, but this is problematic for parallel executor.
Hence, please advice how can I modify paddle so that In activation grad ops (for mkl-dnn builds), X (input signal to fwd op) is available in grad op ?  In other words :
ExecutionContext::OperatorBase::inputs_  should have:
{["Out"] = vector<> , ["Out@GRAD"] = vector<>, **["X"]= vector<>**}

Please advice on how to do so if it is possible?

jczaja:
@reyoung , @tensor-tang , @luotao1 Could you please advice on this issue?

ghost:
@jczaja Could you please rewrite the description for this issue? We (Luo Tao, Leo, and me) all have difficulty to understand it. But I told Luo Tao what you are asking based on our previous discussion. Currently we use SetBlob to save info (e.g. pd/workspace_memory) in FWD OP and then use GetBlob to retrieved those saved info which is needed in BWD OP. For example, we save pooling pd in Pool OP and retrieve it in Pool Grad OP. 
However, this is not a good solution. The best way to pass info between FWD OP and BWD OP are through output tensors of FWD OP. This is because by design all input/output tensors of FWD OP will be passed to BWD OP as inputs.

Based on above description, could you rewrite your question again? And please add:
1. An example 
2. Why this issue is mandatory for "activation" op? why SetBlob/GetBlob way can't apply to it?

Thanks!

jczaja:
Currently(unless something changed) for parallel executor we use std::thread_id to distinguish caching
for one ParallelExecutor thread from other ParallelExecutor thread. But there is once exception which
is when data is to be transfered from Forward Op to Grad Op then TID is **not** used. This is because
diffrent threads may execute Grad ops and Forward Ops.   For activation mkl-dnn ops flow is like this:

<img src="https://raw.githubusercontent.com/jczaja/Paddle/pictures/images/customize.svg?sanitize=true">

For other ops where MKL-DNN PD is transfered from Forward to Backward this is fine, but here (activation mkldnn) there may be a problem as without TID as part of Hash Key effectively
[...]

PR #18656, opened 2019-07-16
Add LeakyRelu MKLDNN support

grygielski:
LeakyRelu MKLDNN support along with unit tests have been added.

### DIFF
--- a/paddle/fluid/operators/activation_op.cc
+++ b/paddle/fluid/operators/activation_op.cc
@@ -363,6 +363,13 @@ class LeakyReluOpMaker : public framework::OpProtoAndCheckerMaker {
     AddInput("X", "Input of LeakyRelu operator");
     AddOutput("Out", "Output of LeakyRelu operator");
     AddAttr<float>("alpha", "The small negative slope").SetDefault(0.02f);
+    AddAttr<bool>("use_mkldnn",
+                  "(bool, default false) Only used in mkldnn kernel")
+        .SetDefault(false);
+    AddAttr<bool>("is_test",
+                  "(bool, default false) Set to true for inference only, false "
+                  "for training. Some layers may run faster when this is true.")
+        .SetDefault(false);
     AddComment(R"DOC(
 LeakyRelu Activation Operator.
 
@@ -695,6 +702,8 @@ class LeakyReluDoubleGradMaker
     op->SetType("leaky_relu_grad_grad");
     // input1: X
     op->SetInput("X", Input("X"));
+    // input2: Out
+    op->SetInput("Out", Input("Out"));
     // X@GRAD@GRAD: ddx
     op->SetInput("DDX", OutputGrad(framework::GradVarName("X")));
     op->SetAttrMap(Attrs());
--- a/paddle/fluid/operators/activation_op.h
+++ b/paddle/fluid/operators/activation_op.h
@@ -1001,7 +1001,7 @@ struct LeakyReluGradFunctor : public BaseActivationFunctor<T> {
     dx.device(d) = dout * (temp1 + temp2).template cast<T>();
   }
 
-  static constexpr ActBwdOpFwdDeps FwdDeps() { return kDepX; }
+  static constexpr ActBwdOpFwdDeps FwdDeps() { return kDepXOut; }
 };
 
 template <typename T>
--- a/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
@@ -77,8 +77,7 @@ class MKLDNNActivationGradKernel
 
 template <typename T>
 void eltwise_forward(const framework::ExecutionContext &ctx,
-                     mkldnn::algorithm algorithm, const T alpha = 0,
-                     const T beta = 0) {
+                     mkldnn::algorithm algorithm) {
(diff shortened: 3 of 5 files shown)


bingyanghuang:
This PR is to fix #17654

jczaja (approved):
LGTM

Sand3r- (approved):

paddle/fluid/operators/activation_op.cc
```diff
@@ -695,6 +702,8 @@ class LeakyReluDoubleGradMaker
     op->SetType("leaky_relu_grad_grad");
     // input1: X
     op->SetInput("X", Input("X"));
+    // input2: Out
+    op->SetInput("Out", Input("Out"));
```
luotao1, line 706: Why add an Input here?

grygielski: It's because MKLDNN activation function grad functor expects to have input called "Out" to create hash key for caching. Without this change activation grad test crashed.

luotao1: > create hash key for caching

Why should we add this input only for creating hash key, since this op doesn't need this input indeed?

grygielski: Well, I did it as it felt like the fastest way to get around this error but you are right, this Input unneccesary so I will change the way MKLDNN key is generated.

luotao1: Thanks very much! If we change the input, then many trained models before will fail after update the develop branch.

jczaja: @luotao1 Could you please explain why trained model will fail when we add additional input to grad op? This is just leaky relu grad op , so this grad op is not holding any params, so why extending inputs for grad op would cause failure? I'm asking because as described in #17960 I would like to extend all mkl-dnn grad ops to have X as input of grad ops as mkl-dnn backward implementations are requiring X to compute backward/grad.

luotao1: @phlrain Could you help see this problem?

luotao1: @jczaja @grygielski 
Discussed with @phlrain, it's OK to add additional input to grad op as this PR does. But we could not `addinput()` in `OpMaker()`.

luotao1:
Do you have any benchmark result for this mkldnn op on face model in #17654

jczaja:
@luotao1 When we wanted to test performance we noticed that face model no longer works on develop for us #18658 . So until we have demark model fixed we cannot test performance of leaky rely of mkl-dnn

luotao1:
Got it. I will check it tomorrow

paddle/fluid/operators/activation_op.cc
grygielski: Ok, so my sollution is acceptable? Because during key generation we would have to add another cases to check if specific operator grad has only X input what causes additional problems when adding MKLDNN support for further activations.

luotao1: Yes, it is.

luotao1:
Thanks very much for your contribution.
I test it on demark model. leaky_relu speedup from 132ms->38ms on 4 threads, **3.47x speedup**!
- before:
```
Event                        Calls       Total       Min.        Max.        Ave.        Ratio.
thread0::conv2d_transpose    400         3794.74     2.86285     17.8366     9.48685     0.768815
thread0::conv2d              400         899.1       0.729642    3.6068      2.24775     0.182158
thread0::leaky_relu          400         132.44      0.085567    1.10562     0.3311      0.0268324
thread0::batch_norm          300         50.5053     0.1174      0.256459    0.168351    0.0102324
thread0::relu                300         39.7742     0.067462    0.242046    0.132581    0.00805825
thread0::fetch               100         11.2578     0.106524    0.123902    0.112578    0.00228083
thread0::tanh                100         7.4732      0.069471    0.090934    0.074732    0.00151407
thread0::feed                100         0.541868    0.004703    0.006246    0.00541868  0.000109783
```
- after:
```
Event                        Calls       Total       Min.        Max.        Ave.        Ratio.
thread0::conv2d_transpose    400         3809.5      2.86721     17.8962     9.52376     0.786597
thread0::conv2d              400         882.729     0.732503    3.49819     2.20682     0.182269
thread0::batch_norm          300         50.5289     0.119364    0.270264    0.16843     0.0104334
thread0::relu                300         42.0752     0.074177    0.234958    0.140251    0.00868782
thread0::leaky_relu          400         38.5553     0.047176    0.194089    0.0963881   0.007961
thread0::fetch               100         10.8161     0.099839    0.117321    0.108161    0.00223335
thread0::tanh                100         8.28066     0.078006    0.108433    0.0828066   0.00170981
thread0::feed                100         0.526761    0.004705    0.006533    0.00526761  0.000108767
```

luotao1:
https://github.com/PaddlePaddle/Paddle/blob/067ed70f2def19263d33ef792323c439fa482484/paddle/fluid/framework/ir/is_test_pass.cc#L27-L35
Do you need to update the `op_list` in `is_test_pass`?

grygielski:
I guess leaky_relu is already there

luotao1:
> I guess leaky_relu is already there

But I don't see it. @grygielski @jczaja

grygielski:
What about 3rd row, 2nd column in op_list you quoted in your last comment?

luotao1:
Sorry, I miss finding it.

PR #18672, opened 2019-07-17
remove ctr reader, all functions are satisfied in dataset

guru4elephant:
ctr reader and dataset should be equivalent given ctr task and dataset is a more general module that we can do more things with.
test=develop


xsrobin (approved):

luotao1 (approved):
LGTM

PR #18674, opened 2019-07-17
Support int64 hash_size in hash_op 

hutuxian:
Support int64 hash_size attribute in hash op.
add corresponding UT.
test=develop


gavin1332 (approved):
LGTM

PR #18683, opened 2019-07-18
Change to use brpc rdma branch instead of personal branch.

gongweibao:
In the past, we use a personal brpc branch, and it's not allowed. So change it to the original rdma branch.


cmake/external/brpc.cmake
```diff
@@ -40,8 +40,8 @@ ExternalProject_Add(
     extern_brpc
     ${EXTERNAL_PROJECT_LOG_ARGS}
     # TODO(gongwb): change to de newst repo when they changed.
```
luotao1: Could you remove the `TODO` here? Since this TODO is finished in this PR.

gongweibao: Thanks! Done!

PR #18686, opened 2019-07-18
add slot to sparse table

Thunderbrook:
the sparse table has not save the slot of feasign before, so this commit add slot to sparse table in order to trace or debug.


python/paddle/fluid/incubate/fleet/parameter_server/pslib/optimizer_factory.py
```diff
@@ -162,7 +162,7 @@ def _minimize(self,
         opt_info["fleet_desc"] = ps_param
         opt_info["worker_skipped_ops"] = worker_skipped_ops
         opt_info["use_cvm"] = strategy.get("use_cvm", False)
-
+        opt_info["dump_slot"] = True
         for loss in losses:
```
xjqbest, line 170: why dump_slot is default true， and user can not set it from strategy ?  what if a user doesn't need dump_slot

Thunderbrook: done

paddle/fluid/framework/fleet/fleet_wrapper.cc
```diff
@@ -315,7 +323,14 @@ void FleetWrapper::PushSparseVarsWithLabelAsync(
     }
     int len = tensor->numel();
     int64_t* ids = tensor->data<int64_t>();
-
+    push_values->resize(fea_keys.size() + 1);
+    for (auto& t : *push_values) {
+      t.resize(emb_dim + offset + slot_offset);
+    }
```
xjqbest: resize here should be moved out of the loop，because it has nothing to do with the loop ("for (size_t i = 0; i < sparse_key_names.size(); ++i) ")

paddle/fluid/framework/fleet/fleet_wrapper.cc
```diff
@@ -315,7 +323,14 @@ void FleetWrapper::PushSparseVarsWithLabelAsync(
...
+    int slot = 0;
+    if (dump_slot) {
+      slot = boost::lexical_cast<int>(sparse_key_names[i]);
```
xjqbest, line 328: are the slot names always be safe to convert to  int ?  what if user defines a slot named "myslot"

paddle/fluid/framework/fleet/fleet_wrapper.cc
Thunderbrook: done

paddle/fluid/framework/fleet/fleet_wrapper.cc
Thunderbrook: check will be add here

xjqbest (approved):

PR #18690, opened 2019-07-19
supports distributed classification

gavin1332:
minimum functional code support for distributed classification training
design doc: http://agroup.baidu.com/paddlepaddle/md/article/1924796


AUTHORS.md
```diff
@@ -28,6 +28,7 @@
 | lcy-seso | Ying Cao |
 | cjld | Dun Liang |
 | lipeng-unisound | Peng Li |
+| liuyi05 | Yi Liu |
```
guru4elephant: are you gavin?

paddle/fluid/operators/shard_index_op.h
```diff
@@ -0,0 +1,58 @@
...
+namespace operators {
+
+using LoDTensor = framework::LoDTensor;
+template <typename T>
+class ShardIndexCPUKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& context) const override {
+    auto* in = context.Input<LoDTensor>("X");
+    auto* out = context.Output<LoDTensor>("Out");
+    int index_range = context.Attr<int>("index_range");
```
guru4elephant: could you use max_index or some other specific name for this variable? index_range seems to be a value of width, better to use a detailed variable name.

paddle/fluid/operators/shard_index_op.h
```diff
@@ -0,0 +1,58 @@
...
+    int nshards = context.Attr<int>("nshards");
+    int shard_id = context.Attr<int>("shard_id");
+    int ignore_value = context.Attr<int>("ignore_value");
+    PADDLE_ENFORCE_GT(index_range, 0);
+    PADDLE_ENFORCE_GT(nshards, 0);
+    PADDLE_ENFORCE(shard_id >= 0 && shard_id < nshards,
+                   "shard_id(%d) is not in range [0, %d)", shard_id, nshards);
+
+    int shard_range = index_range / nshards;
```
guru4elephant: same as above. shard_width can be better? shard_range can be width of a shared, it also can be how many shards we have.

guru4elephant (commented):
I think range is a concept with at least two values, e.g. [0, 10], how about to change index_range to max_index so that we can understand it easily.

AUTHORS.md
gavin1332: 囧，忙晕了

gavin1332: done

paddle/fluid/operators/shard_index_op.h
gavin1332: the attribute name "index_range" is ambiguous indeed, and we need another proper name. In most of cases, users have the variable preserving **number** of indices, so "max_index" attributed requires user manually subtract 1 from the variable and we also have to recover it for "shard_size" calculation later. So I change the attribute "index_range" to "index_num" as a detailed name, which denotes the  number of indices precisely.

paddle/fluid/operators/shard_index_op.h
gavin1332: have been renamed to "shard_size".

guru4elephant (approved):
LGTM, please write some technique documents so that it can be easy to use.

xsrobin (approved):

sneaxiy (approved):
LGTM.

sandyhouse (approved):
LGTM

paddle/fluid/operators/shard_index_op.cu
```diff
@@ -0,0 +1,77 @@
...
+
+template <typename T>
+__global__ void ShardIndexInner(const T* in_data, T* out_data,
+                                const int64_t numel, const int index_num,
+                                const int nshards, const int shard_id,
+                                const int ignore_value) {
+  int shard_size = index_num / nshards;
+  int idx = blockIdx.x * blockDim.x + threadIdx.x;
+  if (idx < numel) {
+    assert(in_data[idx] >= 0 && in_data[idx] < index_num);
```
chengduoZH, line 32: Why do you use assert here? do you check whether it works?

gavin1332: I want just make sure the input is in the valid range

gavin1332:
TODO：replace raw `assert` by `PADDLE_ASSERT_MSG`

chengduoZH (approved):

hutuxian (approved):

guoshengCS (approved):
LGTM

PR #18702, opened 2019-07-19
Add python API for appending LoD level

wanghaoshuang:
Make lod reset op support for append lod level


python/paddle/fluid/layers/nn.py
```diff
@@ -7060,8 +7061,60 @@ def lod_reset(x, y=None, target_lod=None):
...
+            level: [1, 1, 1, 1, 1, 1, 1]
+
+            then we get a 2-level LoDTensor:
+                x.lod =  [[ 2, 3, 1 ], [1, 1, 1, 1, 1, 1]]
+                x.data = [[1.0], [2.0], [3.0], [4.0], [5.0], [6.0]]
+                x.dims = [6, 1]
 
+    Args:
+        x (Variable): Input variable which could be a tensor or LoDTensor.
+        level (list|tuple): The LoD level to be appended into LoD of x..
```
heavengate: 多了个.

python/paddle/fluid/layers/nn.py
```diff
@@ -7060,8 +7061,60 @@ def lod_reset(x, y=None, target_lod=None):
...
+
+    Returns:
+        Variable: Output variable with new LoD.
```
heavengate: new LoD level ?

python/paddle/fluid/layers/nn.py
wanghaoshuang: Thx. Fixed.

python/paddle/fluid/layers/nn.py
wanghaoshuang: Thx. Fixed.

Issue #18708, opened 2019-07-19
PR #18656 add unnecessary input in leaky_relu_grad op


sneaxiy:
PR #18656 adds unnecessary input in leaky_relu_grad op. Grad ops of activation ops would never need both forward input `X` and forward output `Out` as their inputs, because `Out` can be always calculated by `X` in backward pass.

This changing would make the existing GPU memory optimization strategy not work. Please fix it asap. @grygielski .

related #18707

bingyanghuang:
@jczaja Please solve this issue with high priority.

PR #18723, opened 2019-07-22
Revert "Add LeakyRelu MKLDNN support (#18656)"

### LINKED ISSUE #18707: PR #18656 add unnecessary input in leaky_relu_grad op
PR #18656 adds unnecessary input in leaky_relu_grad op. Grad ops of activation ops would never need forward input `X` and forward output `Out` as their inputs, because `Out` can be always calculated by `X` in backward pass.

This changing would make the existing GPU memory reuse strategy not work. Please fix it asap. @grygielski .

luotao1:
fix #18707 and #18708 
image
After merge 18656, the gpu memory usage of yoloV3 become larger. yoloV3 contains leakyRelu op, thus, this PR reverts #18656 at first. @jczaja @grygielski


sneaxiy (approved):
LGTM.

PR #18725, opened 2019-07-22
Fix build strategy doc

chengduoZH:
Fix build strategy doc error.

For `GradientScaleStrategy.Customized`, there has a use case in [the transformer model](https://github.com/PaddlePaddle/models/blob/develop/PaddleNLP/neural_machine_translation/transformer/train.py#L533-L537).
```
    # Since the token number differs among devices, customize gradient scale to
    # use token average cost among multi-devices. and the gradient scale is
    # `1 / token_number` for average cost.
    # build_strategy.gradient_scale_strategy = fluid.BuildStrategy.GradientScaleStrategy.Customized
```

### DIFF
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -1283,12 +1283,13 @@ All parameter, weight, gradient are variables in Paddle.
             PADDLE_ENFORCE(!self.IsFinalized(), "BuildStrategy is finlaized.");
             self.reduce_ = strategy;
           },
-          R"DOC(The type is STR, there are two reduce strategies in ParallelExecutor,
-                'AllReduce' and 'Reduce'. If you want that all the parameters'
-                optimization are done on all devices independently, you should choose 'AllReduce';
-                if you choose 'Reduce', all the parameters' optimization will be evenly distributed
-                to different devices, and then broadcast the optimized parameter to other devices.
-                In some models, `Reduce` is faster. Default 'AllReduce'.
+          R"DOC(The type is fluid.BuildStrategy.ReduceStrategy, there are two reduce
+                strategies in ParallelExecutor, AllReduce and Reduce. If you want
+                that all the parameters' optimization are done on all devices independently,
+                you should choose AllReduce; if you choose Reduce, all the parameters'
+                optimization will be evenly distributed to different devices, and then
+                broadcast the optimized parameter to other devices.
+                Default 'AllReduce'.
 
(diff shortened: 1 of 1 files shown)


paddle/fluid/pybind/pybind.cc
```diff
@@ -1302,21 +1303,62 @@ All parameter, weight, gradient are variables in Paddle.
...
+                        loss = fluid.layers.mean(hidden)
+                        fluid.optimizer.SGD(learning_rate=0.01).minimize(loss)
+
+                        fluid.default_startup_program().random_seed=1
+                        exe.run(fluid.default_startup_program())
+
                         build_strategy = fluid.BuildStrategy()
-                        build_strategy.gradient_scale_strategy = True
+                        build_strategy.gradient_scale_strategy = \
+                                 fluid.BuildStrategy.GradientScaleStrategy.Customized
```
chenwhql, line 1349: Why use `Customized` in example? is this more commonly used?

chengduoZH: Because it is a bit complicated, most people don't know how to use it.

chenwhql (approved):
LGTM

paddle/fluid/pybind/pybind.cc
```diff
@@ -1302,21 +1303,62 @@ All parameter, weight, gradient are variables in Paddle.
...
-          R"DOC(The type is STR, there are three ways of defining :math:`loss@grad` in
-                ParallelExecutor, 'CoeffNumDevice', 'One' and 'Customized'. By default,
-                ParallelExecutor sets the :math:`loss@grad` according to the number of devices.
-                If you want to customize :math:`loss@grad`, you can choose 'Customized'.
-                Default 'CoeffNumDevice'.
+          R"DOC(The type is fluid.BuildStrategy.GradientScaleStrategy, there are three
+                ways of defining :math:`loss@grad` in ParallelExecutor, CoeffNumDevice,
+                One and Customized. By default, ParallelExecutor sets the :math:`loss@grad`
+                according to the number of devices. If you want to customize :math:`loss@grad`,
+                you can choose Customized. Default 'CoeffNumDevice'.
```
lanxianghit, line 1314: I think we should tell the users what the action is in each way of the strategis.

chengduoZH: Where do you think the description is not clear?

The functionality of each option has been described here, and an example shows how to use the Customized pattern.

PR #18741, opened 2019-07-23
Open fuse optimization ops

chengduoZH:
The PR's work includes: 
- [x] Open fuse_optimization_ops, currently. If this model uses `SGD`, `Momentum` or `Adam` and the type of gradients is not sparse, those optimization ops will be fused. Note that if the model has dense gradients and sparse gradients, only the optimization ops which are used to update the parameters whose gradients are dense will be fused. 
- [x] Polish build_strategy code. The appending operations are reorganized, for example, all the op fusion passes are placed in `AppendOpFusePasses`, all memory optimization passes are placed in `AppendMemoryOptimizePasses`.
- [x] Fix sync_batch_norm_pass bug. For some op nodes, they don't have op_desc, for example, all_reduce_op node, the `node->Op()` is nullptr.


luotao1:
Could you separate into three-small PRs which are easier to review?
> Fix sync_batch_norm_pass bug. For some op nodes, they don't have op_desc, for example, all_reduce_op node, the node->Op() is nullptr.

For example, could you describe the bug in details? What's the wrong result of this bug? And if you fix it, could you add unit-test to guarantee the mistake update in the later PRs from others.

paddle/fluid/framework/details/build_strategy.cc
```diff
@@ -47,212 +47,195 @@ class ParallelExecutorPassBuilder : public ir::PassBuilder {
...
-    if (strategy_.enable_sequential_execution_) {
-      VLOG(1) << "Add sequential_execution_pass";
-      AppendPass("sequential_execution_pass");
+  void ResolveOptionConfliction() {
+    // Specifies the restrictions between different pass.
+    if (strategy_.enable_parallel_graph_) {
+      VLOG_IF(3, strategy_.fuse_all_optimizer_ops_)
+          << "Currently, fuse_all_optimizer_ops doesn't works under "
+             "parallel_graph.";
+      strategy_.fuse_all_optimizer_ops_ = false;
```
gongweibao, line 88: If we changed the user's config, there should be LOG(WARNING) and the other under this line is the same problem.

paddle/fluid/framework/details/build_strategy.cc
```diff
@@ -47,212 +47,195 @@ class ParallelExecutorPassBuilder : public ir::PassBuilder {
...
+      auto viz_pass = AppendPass(pass_name);
+      const std::string graph_path = string::Sprintf(
+          "%s%s", strategy_.debug_graphviz_path_.c_str(), debug_file_suffix);
+      viz_pass->Set<std::string>(ir::kGraphvizPath,
+                                 new std::string(graph_path));
+    }
+  }
+
+  void AppendPassWithCheck(bool append_pass, const std::string &pass_name) {
+    if (append_pass) {
```
gongweibao, line 217: The pass used for optimize graph is important info for user to debug and why delete the `VLOG(1)`?

paddle/fluid/framework/details/build_strategy.cc
```diff
@@ -47,212 +47,195 @@ class ParallelExecutorPassBuilder : public ir::PassBuilder {
...
     multi_devices_pass->SetNotOwned<const BuildStrategy>("strategy",
                                                          &strategy_);
   }
 
+  void AppendPrintGraphPass(const std::string &pass_name,
+                            const std::string &debug_file_suffix) {
+    if (!strategy_.debug_graphviz_path_.empty()) {
+      auto viz_pass = AppendPass(pass_name);
+      const std::string graph_path = string::Sprintf(
+          "%s%s", strategy_.debug_graphviz_path_.c_str(), debug_file_suffix);
```
gongweibao, line 210: `debug_graphviz_path_`  should be renamed?

paddle/fluid/framework/ir/pass_builder.cc
```diff
@@ -13,12 +13,15 @@ See the License for the specific language governing permissions and
...
 #include "paddle/fluid/framework/ir/pass_builder.h"
+#include <memory>
+#include <utility>
 
 namespace paddle {
 namespace framework {
 namespace ir {
 
 std::shared_ptr<Pass> PassBuilder::AppendPass(const std::string& pass_type) {
+  VLOG(3) << "Append " << pass_type;
```
gongweibao, line 24: VLOG(1)?

paddle/fluid/framework/ir/pass.h
```diff
@@ -104,6 +108,10 @@ class Pass {
     LOG(FATAL) << "Calling virtual Pass not implemented.";
   }
 
+  // Some Pass must be placed before this Pass, and some
+  // Pass must be placed after this Pass.
+  virtual void CheckPrevPass() const {}
```
gongweibao, line 113: Who implements it?

paddle/fluid/framework/ir/pass.cc
```diff
@@ -40,6 +41,10 @@ Graph* Pass::Apply(Graph* graph) const {
   PADDLE_ENFORCE(VarDescIsConsistency(*graph),
                  "The VarDescs of persistable variable are not consistency.");
   applied_ = true;
+  if (!graph->Has(kPassRecorder)) {
```
gongweibao, line 45: No one use it also?

gongweibao (changes requested):

paddle/fluid/framework/ir/pass.cc
chengduoZH: Maybe, I will fix this in the next [PR](https://github.com/PaddlePaddle/Paddle/pull/18833).

paddle/fluid/framework/ir/pass.h
chengduoZH: The derived class of Pass will implement this function if the Pass must be placed after other specified Pass.

paddle/fluid/framework/details/build_strategy.cc
chengduoZH: I agree with you, so we should recognize whether this option is enabled by the user or not. I will fix this in the next PR https://github.com/PaddlePaddle/Paddle/pull/18833.

paddle/fluid/framework/details/build_strategy.cc
chengduoZH: Right.

paddle/fluid/framework/ir/sync_batch_norm_pass.cc
```diff
@@ -26,7 +26,7 @@ class SyncBatchNormPass : public Pass {
   void ApplyImpl(ir::Graph *graph) const override {
     VLOG(3) << "Use synchronous batch norm";
     for (const Node *n : graph->Nodes()) {
-      if (n->IsOp()) {
+      if (n->IsOp() && n->Op()) {
```
Xreki, line 29: Is there a common method to fix this? In fact, all passes should add this check to avoid crash.

python/paddle/fluid/tests/unittests/test_buffer_shared_inplace_pass.py
```diff
@@ -171,6 +175,16 @@ def test_multi_card_fetch_var(self):
 class CPUInplaceTest(InplaceTestBase):
     def initParameter(self):
         self.use_cuda = False
+        self.fuse_all_optimizer_ops = False
+
+
+class CUDAInplaceTestWithFuseOptimizationOps(InplaceTestBase):
+    def initParameter(self):
+        self.use_cuda = True
+        # FIXME(zcd): if training is in CPU, fuse_all_optimizer_ops
```
Xreki: Why? There should be no difference between CPU and GPU... And you mean, we can't set this option to `True` when training on CPU?

paddle/fluid/framework/details/build_strategy.h
```diff
@@ -88,7 +88,7 @@ struct BuildStrategy {
   bool fuse_elewise_add_act_ops_{false};
   // Fuse_all_optimizer_ops and fuse_all_reduce_ops require that gradients
   // should not be sparse types
-  bool fuse_all_optimizer_ops_{false};
+  bool fuse_all_optimizer_ops_{true};
```
Xreki, line 91: As you say, there are some differences for CPU, I don't think it is suitable to set the default value to true.

python/paddle/fluid/tests/unittests/test_buffer_shared_inplace_pass.py
chengduoZH: >  if training is in CPU, fuse_all_optimizer_ops may cause some diff with the original model, the difference is about 1e-8. This may be related to byte alignment.

paddle/fluid/framework/details/build_strategy.h
chengduoZH: This effect is very small, and I have run the models of CE and CI, all the tests are passed.

paddle/fluid/framework/ir/sync_batch_norm_pass.cc
chengduoZH: We should throw an exception in the `Node.Op()` if this Node doesn't have OpDesc. And so, the developer will know the error info.

paddle/fluid/framework/details/build_strategy.h
gongweibao: If we set `fuse_all_optimizer_ops_` default true, how much will the performance speedup?

chengduoZH: About 5%, in the transformer base model.

gongweibao (dismissed):
LGTM

gongweibao (approved):
LGTM

paddle/fluid/framework/details/build_strategy.h
luotao1: This will make pyramid_dnn training process fail without any error.

PR #15939, opened 2019-02-26
enable cpplint, remove go_fmt

luotao1:
- enable `cpplint` on travis_ci
- remove `paddle/scripts/cpplint.py`, we use `pip install cpplint` on both local machine and travis_ci
- remove go_fmt since fluid version doesn't have go codes

TODO: `cpplint` on travis_ci runs in multi-thread, and it will print several same logs when there are errors.
related #15907 #15876

PR #18744, opened 2019-07-23
remove unused cmake file

luotao1:
remove unused cmake file
- since we have `cmake/external/glog.cmake` and `cmake/external/gflags.cmake`, this PR remove `FindGlog.cmake` and `FindGflags.cmake`
- since #15939 remove the check of `go`format, this PR remove `go` related cmake files.


Xreki (approved):
LGTM

PR #18750, opened 2019-07-23
Fix config description error in cuda_profiler function document

chenwhql:
image
The config parameter in doc is duplicate, we should change the second `gpustarttimestamp` to `gpuendtimestamp`, just as shown in the figure.


Superjomn (approved):
LGTM

lanxianghit (approved):
Please change a better title

PR #18762, opened 2019-07-23
Add LeakyReLU MKLDNN support

grygielski:
Add LeakyReLU MKLDNN support v2. This version of code uses additional attribute added to activation grad operator `op->SetAttr("forward_input_X", Input("X").back());`. I tried different things but it looks like matching operator and its gradient is only possible through this unique string got from Input() function. Is this additional Attr also a problem for you? @luotao1 @sneaxiy


sneaxiy (commented):
Adding attributes seems too strange. I do not recommend to do so. How about your view @luotao1 on this PR? Moreover, I am not sure that whether the memory optimization strategy of inference would be conflict with variable names in attributes. Memory optimization strategy of training has discarded the way of changing variable names, and does inference strategy rely on  changing variable names? Please take a look @Superjomn.

grygielski:
Do you have any other idea how to connect activation operator and its grad during execution? I've tried different things and it seems like in the current state of the project it is the only reasonable way. Things would be easier if activation op grads would be normalized (e.g. every grad op would take X as a Input). Take a look at this attribute as a pointer indicating forward function of specific instance of grad op. It seems ok for me as its just a string attribute (like other operators have tons of them as flags or other stuff) which does not influence any current functionality of operator at all. (and if it does, put me right).

paddle/fluid/operators/activation_op.cc
```diff
@@ -83,6 +83,7 @@ class ActivationGradOpDescMaker : public framework::SingleGradOpDescMaker {
     op->SetInput(framework::GradVarName("Out"), OutputGrad("Out"));
     op->SetOutput(framework::GradVarName("X"), InputGrad("X"));
     op->SetAttrMap(Attrs());
+    op->SetAttr("forward_input_X", Input("X").back());
```
luotao1: Do you mean
```
if (op->HasAttr("use_mkldnn"))
   op->SetAttr("forward_input_X", Input("X").back());
```
by https://github.com/PaddlePaddle/Paddle/issues/18788#issuecomment-514599133 @jczaja

grygielski: More like:
```cpp
if (executing by mkldnn) {
    op->SetInput("X", Input("X"));
}
````

Comment (luotao1) [reply paddle/fluid/operators/activation_op.cc]: why `SetInput("X", Input("X"))` here? not `op->SetAttr("forward_input_X", Input("X").back());`?

Comment (Sand3r-) [reply paddle/fluid/operators/activation_op.cc]: So essentially
```C++
if (op->HasAttr("use_mkldnn") && op->Attr<bool>(use_mkldnn))
   op->SetAttr("forward_input_X", Input("X").back());
```

luotao1: @sneaxiy How do you think about https://github.com/PaddlePaddle/Paddle/pull/18762#discussion_r306772070. I think it doesn't affect Memory optimization strategy in both training and inference.

Sand3r-: I just realised that since `use_mkldnn` is set during MKLDNN Placement Pass, the `use_mkldnn` attribute will return false here. We need to find another way of determining if MKLDNN is to be enabled here.

grygielski: @luotao1 `SetInput()` instead of `SetAttr()` because you didn't like additional attributes and it carries exact same information. As long as we use mkldnn not gpu, we won't mess with gpu memory strategy and it's also easier for us to reference it as Input.

jczaja: As this is training via python, how about using FLAGS_use_mkldnn value to determine if we should add another input or not ?

luotao1:
@sneaxiy and I both confused why you add attr `forward_input_X` in `leak_relu_grad`, we see the grad computing use `X` only, do you want to add `Out` in fact? @jczaja @Sand3r- 

I translate to Chinese and @jianhang-liu could you help translate the answer to us? We may have something misunderstanding.
为什么leak_relu_grad要加forward_input_X，代码里反向计算要用的就是X，也就是说X本来就是输入里？是不是想加Out？

jczaja:
@luotao1 For this PR either X or Out is fine. For #17960 We need X. So we prefer adding X to MKL-DNN based workloads for activations.

grygielski:
I guess, the main problem is we do not understand each other well in case of this problem. I prepared some diagrams to show you our problem and make a common ground. By "Solution we are proposing now" I meant the one we discussed in previous comments, not the one in this PR. @luotao1 @jczaja  @Sand3r- @sneaxiy 

mkldnnops

sneaxiy:
Good explanations! I have been more clear about what you want to do. Thank a lot!

How about list the problems we face currently or in the future:

- MKLDNN forward op wants to cache a key-value pair when running, where the key is generated by forward input "X". The cached pair is used to store some primitive things used in MLDNN internally.

- MKLDNN backward op would extract the MKLDNN primitive cache using the key generated by forward input "X". Therefore, it is the first reason why you want to add "X" as additional input of MKLDNN backward op.

- What's more critcal when making parallelization is that: the key of cache is also generated by thread id too. As a result, MKLDNN forward op and backward op have to run in the same thread, otherwise data overwritting by other threads may happen. In order to guarantee that MKLDNN forward op and backward op run in the same thread, you have to find the matching forward/backward op pairs beforehand, so that they can be scheduled in the same thread during run-time. The matching method is based on whether input "X" of forward op is the same as input "X" of backward op. Therefore, it is the second reason why you want to add "X" as additional input of MKLDNN backward op as a common solution.

Is my understanding above right? If so, we can discuss furthermore on how to solve this problem. If not, please let me know. @grygielski

sneaxiy:
Good explanations! I have been more clear about what you want to do. Thanks a lot!

How about list the problems we face currently or in the future:

- HOW IT IS NOW: MKLDNN forward op wants to cache a key-value pair when running, where the key is generated by forward output "Out" and the value is forward input "X". This cache is used to query MKLDNN primitives.

- HOW IT IS NOW: MKLDNN backward op would extract the MKLDNN primitives in the way: generate key by forward output "Out" -> obtain forward input "X" by the generated key -> obtain MKLDNN primitive by forward input "X".

- The query of MKLDNN primitives can be simplified: query MKLDNN primitive directly by forward input "X". Therefore, it is the first reason why you want to add "X" as additional input/attribute of MKLDNN backward op.

- What's more critcal when making parallelization is that: the key of cache(or key of MKLDNN primitives) is also generated by thread id too. As a result, MKLDNN forward op and backward op have to run in the same thread, otherwise data overwritting by other threads may happen. In order to guarantee that MKLDNN forward op and backward op run in the same thread, you have to find the matching forward/backward op pairs beforehand, so that they can be scheduled in the same thread during run-time. The matching method is based on whether input "X" of forward op is the same as input/attribute "X" of backward op. Therefore, it is the second reason why you want to add "X" as additional input/attribute of MKLDNN backward op as a common solution.

Is my understanding above right? If so, we can discuss furthermore on how to solve this problem. If not, please let me know. @grygielski

grygielski:
1. Exactly
2. The problem is that we can't be sure if we should cache value of forward Input("X") by X or Out key because different grad operations have different inputs. In this situation we can't get X input needed by MKLDNN grad primitive from cache since we don't have access to specific key in some grad operators.
3. Exactly
4. Multi threading problem exists in a way you described but we can get rid of it just by adding Input("X") to every Grad Operation which uses MKLDNN. This is because we won't be using caching forward input X as we will have access to it in grad directly.

Thanks a lot for involvement. Let me know if it is all clear for you now. @sneaxiy @luotao1

grygielski:
I pushed code for diagram section: "Solution we are proposing now"

grygielski:
I would be pleased if you could run test on current commit and check if this version is neutral for gpu implementation. @sneaxiy @luotao1

jczaja (approved):
LGTM

@sneaxiy , @luotao1 Please test those changes on your GPU workload so we know that GPU memory strategy works as expected.

luotao1:
@grygielski @jczaja Sorry for the late reply, we have some different internal options. We will have some discussion ASAP.
- @sneaxiy think we should not add any input/output in an Op.
- but @phlrain think we could add some input as this PR does.

From @sneaxiy opinion, you can try:
- Define `DECLARE_NO_NEED_BUFFER_VARS_INFERENCE(class_name，no_need_variable)`。For example, define`DECLARE_NO_NEED_BUFFER_VARS_INFERENCE(LeakyReluGradOpNoNeedBufferVarInference, "X")`, means `leaky_relu_op` don't need X's buffer。
- Register above class into Op, i.e `REGISTER_OPERATOR()`. For example, `REGISTER_OPERATOR(leak_relu_grad, LeakReluGradOpNoNeedBufferVarInference....)`

luotao1:
如果我在没有打开MKLDNN的情况下搭了个网络，然后把program序列化存成文件。然后下次，我load这个文件，把训练program从文件里反序列化回来，然后打开MKLDNN训练，就会出现MKLDNN op找不到X的问题

If we train a model when `use_mkldnn=false` and save a training program_desc, then we load the training program_desc and continually train when `use_mkldnn=true`. Could `mkldnn_op` find input `X`?@phlrain

grygielski:
@luotao1 @sneaxiy Another solution came to my mind. If all operators with grad dependent on "Out" are inplace operators. Can't we just use Input("X") as input to every grad op since Input("X") carries same value as Output("Out") does (as it is inplace operation, this is described in https://github.com/PaddlePaddle/Paddle/issues/17960). This solution would normalize every grad operator to use Input("X") without any major changes. Is there something problematic with such solution? @jczaja

luotao1:
Discussed with @phlrain, this usage https://github.com/PaddlePaddle/Paddle/pull/18762#issuecomment-515361165 is almost nonexistent in practice. Thus, this PR is OK to be merged.

PR #18781, opened 2019-07-24
Change select_rows add from math::scatter::MergeAdd to SelectedRowsAd…

JesseyXujin:
…dTo() on select_rows_function.h.

From profile test, after change it, the compute speed is 20 times faster than before.
test=develop
Before change:
profile2
After change:
profile1


paddle/fluid/operators/sum_op.h
```diff
@@ -47,37 +47,56 @@ void SelectedRowsCompute(const framework::ExecutionContext &context) {
...
                           temp_in0.mutable_value());
     inputs.push_back(&temp_in0);
+    SelectedRows temp_in1;
     for (size_t i = 1; i < in_vars.size(); ++i) {
       auto &in = in_vars[i]->Get<SelectedRows>();
       if (in.rows().size() > 0) {
-        inputs.push_back(&in);
+        temp_in1.set_height(in.height());
+        temp_in1.set_rows(in.rows());
+        framework::TensorCopy(in.value(), in.place(), context.device_context(),
```
sneaxiy: Why copy? It seems wrong. `temp_in1` would be overwritten multiple times.

seiriosPlus:
没有加 test=develop？

PR #18831, opened 2019-07-26
fix several security bugs reported by security team

liupluswei:
fix several null pointer dereference and invalid mem operation issues：

1. avoid using memcpy with 0 length which may cause undifined behaviour in api.cc
2. check potenail nullptr in gpc.cc
3. avoid using memcmp with 0 length in piece.cc


phlrain (approved):
LGTM

PR #18833, opened 2019-07-26
Open fuse broadcast option

chengduoZH:
In Reduce mode, if the broadcast ops, which are used to sync the parameters between GPUs, are fused, the training speed increased by 15.2%(**261.22 img/sec -> 301.17 img/sec**). The env is V100, CUDA9, nccl2.4.2, GPUs: 4. The model is `SE_ResNeXt50_32x4d` and batch_size: 256.

- [x] Recognize whether this option is enabled by the user or not.
- [x] Use VLOG(1) in build_strategy.


paddle/fluid/framework/details/broadcast_op_handle.cc
```diff
@@ -126,6 +126,9 @@ void BroadcastOpHandle::BroadcastOneVar(
             &VariableVisitor::GetMutableTensor(out_var));
       }
     });
+    for (auto &p : places_) {
```
gongweibao, line 129: Why wait?

gongweibao (commented):
Just one comment.

paddle/fluid/framework/details/broadcast_op_handle.cc
chengduoZH: In Reduce mode, all the updating of parameters are placed in separate GPUs, so we should use broadcast the new parameter from one GPU to others, but the GPU is an async device, it means that when the calling returned on CPU side, the tasks in the GPU are not finished, this will induce, in the next iteration, the parameters in some GPUs may be the old value but not the updated.

gongweibao (approved):
LGTM

PR #18735, opened 2019-07-23
Feature/mem opt pass refactor

sneaxiy:
This PR refactors legacy C++ memory optimization strategy, which is based on changing variable names. The new strategy uses `ShareTensorBufferOpHandle` to perform memory reuse during run-time instead of changing variable names.

This is designed to fix the issue that users have to set `var.persistable = True` before using this memory optimization strategy.

**Sorry that: since this PR is too long, I would remove the legacy C++ memory optimization codes in next PR.**

I have tried to turn on `build_strategy.memory_optimize = True` by default on CE Debug. All tasks have passed. Please see [here](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=54043&buildTypeId=PaddleModesl_ModelDebug_Model11Build&tab=buildResultsDiv).

PR #18834, opened 2019-07-26
Remove legacy C++ memory optimization codes

sneaxiy:
As title.

This PR also enables `build_strategy.memory_optimize` when gc is disabled by default.

We have refactored in-place and memory-optimize passes in #17911 and #18735, which would never rename variable names. We do not need the legacy strategies any more.


paddle/fluid/framework/parallel_executor.cc
```diff
@@ -252,7 +252,21 @@ ir::Graph *ParallelExecutorPrivate::ApplyMemoryOptimizePass(ir::Graph *graph) {
     VLOG(10) << "buffer_shared_inplace_pass Applied";
   }
 
-  if (build_strategy_.memory_optimize_) {
+  /**
```
zhhsplendid, line 255: The comment logic doesn't match the code. In the comment you said 
"If gc is enabled, BuildStrategy.memory_optimize = False.
If gc is disabled, BuildStrategy.memory_optimize = True.
...
gc + memory_optimize is worse than gc only."

So your code logic should probably be without "if (!build_strategy_.memory_optimize_)" 
The if condition makes it possible for user to set both
build_strategy_.memory_optimize_ and gc to True.

Please consider to modify the comment or the code.

paddle/fluid/framework/parallel_executor.cc
```diff
@@ -313,6 +327,9 @@ ir::Graph *ParallelExecutorPrivate::ApplyMemoryOptimizePass(ir::Graph *graph) {
     eager_deletion_pass->SetNotOwned(ir::kAllPlaces, &places_);
     graph = eager_deletion_pass->Apply(graph);
     VLOG(10) << "EagerDeletionPass Applied";
+    LOG(INFO) << "Garbage collection strategy is enabled, when "
+              << "FLAGS_eager_delete_tensor_gb="
```
zhhsplendid: If you have time, add places around '='

such as "FLAGS_eager_delete_tensor_gb = "

paddle/fluid/framework/parallel_executor.cc
```diff
@@ -313,6 +327,9 @@ ir::Graph *ParallelExecutorPrivate::ApplyMemoryOptimizePass(ir::Graph *graph) {
...
+              << GetEagerDeletionThreshold() / (static_cast<int64_t>(1) << 30);
```
zhhsplendid: Can we write like this?
<< (GetEagerDeletionThreshold() >> 30);

paddle/fluid/framework/parallel_executor.cc
```diff
@@ -252,7 +252,21 @@ ir::Graph *ParallelExecutorPrivate::ApplyMemoryOptimizePass(ir::Graph *graph) {
     VLOG(10) << "buffer_shared_inplace_pass Applied";
   }
 
-  if (build_strategy_.memory_optimize_) {
+  /**
+   * NOTE(zjl): If BuildStrategy.memory_optimize = None in Python,
```
luotao1: `NOTE(zjl)`->`TODO(zengjinle)`? It's easier to find who write these with the full name.

luotao1 (commented):
LGTM for remove the codes.

paddle/fluid/framework/parallel_executor.cc
sneaxiy: Discussed offline. User can forcely turn on `memory_optimize` by setting True. The comment is right.

paddle/fluid/framework/parallel_executor.cc
sneaxiy: Done.

paddle/fluid/framework/parallel_executor.cc
sneaxiy: Done.

paddle/fluid/framework/parallel_executor.cc
sneaxiy: Done.

lanxianghit (approved):

xsrobin (approved):

PR #18840, opened 2019-07-26
Add warning info If CPU_NUM is not set

chengduoZH:
relate https://github.com/PaddlePaddle/Paddle/issues/18665

If the CPU_NUM is not set, we should print a warning to inform the user to set CPU_NUM if they want to make the program faster.
The warning info is:
```
!!! The CPU_NUM is not specified, you should set CPU_NUM in the environment variable list.
CPU_NUM indicates that how many CPUPlace are used in the current task.
And if this parameter are set as N (equal to the number of physical CPU core) the program may be faster.

export CPU_NUM=38 # for example, set CPU_NUM as number of physical CPU core which is 38.

!!! The default number of CPU_NUM=1.
```

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -85,11 +85,14 @@ def _current_expected_place():
 
 def _cpu_num():
     if "CPU_NUM" not in os.environ.keys():
-        sys.stderr.write(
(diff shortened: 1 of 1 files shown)


luotao1 (approved):
LGTM

luotao1:
image
我们以后会加强下这段输出么，比如有颜色显示。 @wopeizl

PR #18849, opened 2019-07-26
[Cherry-pick] fix affine_channel no_need buffer bug

sneaxiy:
Cherry-pick #18844.

### DIFF
--- a/paddle/fluid/operators/affine_channel_op.cc
+++ b/paddle/fluid/operators/affine_channel_op.cc
@@ -295,10 +295,10 @@ class AffineChannelNoNeedBufferVarsInference
   using framework::NoNeedBufferVarsInference::NoNeedBufferVarsInference;
 
  private:
-  inline bool HasInput(const std::string& name) const {
-    auto& inputs = Inputs();
-    auto iter = inputs.find(name);
-    if (iter == inputs.end() || iter->second.empty()) {
+  inline bool HasOutput(const std::string& name) const {
+    auto& outputs = Outputs();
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/affine_channel_op.cc
```diff
@@ -295,20 +295,20 @@ class AffineChannelNoNeedBufferVarsInference
...
     }
   }
 
  public:
-  std::unordered_set<std::string> operator()() const {
-    if (!HasInput(framework::GradVarName("Scale")) &&
-        !HasInput(framework::GradVarName("Bias"))) {
+  std::unordered_set<std::string> operator()() const override {
+    if (!HasOutput(framework::GradVarName("Scale")) &&
+        !HasOutput(framework::GradVarName("Bias"))) {
```
zhhsplendid, line 311: Since, Out = Scale*X + Bias, do we have to have X's data when computing  Bias@Grad?

Can we write:

if (!HasOutput(framework::GradVarName("Scale"))) {
    return {"X"};
}

zhhsplendid: Same comment as develop branch.

zhhsplendid: Discussed offline, the OP calculates Scale@Grad and Bias@Grad at the same time. In other words, the two Grads must occur at the same time. The PR LGTM

zhhsplendid (approved):
LGTM

PR #18867, opened 2019-07-29
Split test_buffer_shared_memory_reuse_pass.py to two files

chengduoZH:
Split test_buffer_shared_memory_reuse_pass.py to two files, otherwise, the time consuming of test_buffer_shared_memory_reuse_pass.py  is too long, and it results in this unit test timeouts.


sneaxiy (approved):
LGTM

Issue #18788, opened 2019-07-24
[MKL-DNN] Add Leaky-relu support and conv+leaky_relu fuse pass


grygielski:
Add LeakyReLU mkldnn support and Conv+LeakyReLU mkldnn pass to improve performance of Face model. This is reopening of #17654 due to PR reversion.

Scope of task:
1. Add mkl-dnn leaky_relu support
2. Add Pass: conv + leaky_relu -> conv_leaky_relu_mkldnn_fuse_pass

grygielski:
I created PR that adds leaky_relu mkldnn support but uses additional attribute added to activation op grad (pointer to corresponding forward activation op): https://github.com/PaddlePaddle/Paddle/pull/18762. Let me know if you accept this universal solution. If not I will just add conditional cases to check for specific operators whose grads takes only X as input. This solution does not add any more parameters to current operators but leads to more complicated development of mkldnn activation operators support. @luotao1

grygielski:
#18762

luotao1:
I agree with `Adding attributes seems too strange` from @sneaxiy  in https://github.com/PaddlePaddle/Paddle/pull/18762#pullrequestreview-265794216.

> If not I will just add conditional cases to check for specific operators whose grads takes only X as input. This solution does not add any more parameters to current operators but leads to more complicated development of mkldnn activation operators support.

Could you easily describe how to do this check? And `more complicated development` will add a lot of codes? 

@jczaja How do you see about it?

grygielski:
To solve this issue for now I would just use something like this in eltwise_forward() and similar one in eltwise_grad() in activation_mkldnn_op.cc
```cpp
if (ctx.op().Type() == "leaky_relu" || Other operators with grad X input) {
  std::string key = ... + ctx.op().Input("X"));
} else {
  std::string key = ... + ctx.op().Output("Out"));
}
```
[...]

PR #18875, opened 2019-07-29
Add Conv+LeakyReLU MKLDNN fuse pass

grygielski:
As mentioned in https://github.com/PaddlePaddle/Paddle/issues/18788, I add conv + leaky_relu fuse pass to accelerate internal face model. However, solution is temporary and I will work on creating one, universal conv+activation fuse pass so additional conv parameters will be cut.


grygielski:
@luotao1 Can you help with this failing CI?

jczaja:
@grygielski PR_CI_Document_Preview does not have "Required" label next to it , so it is not mandatory to have it passed.

@luotao1 Could you please test Face model performance with this PR and share your findings ?

paddle/fluid/operators/conv_op.cc
```diff
@@ -209,6 +209,12 @@ void Conv2DOpMaker::Make() {
       .SetDefault(false);
   AddAttr<bool>("fuse_relu", "(bool, default false) Only used in mkldnn kernel")
       .SetDefault(false);
+  AddAttr<bool>("fuse_leaky_relu",
```
jczaja, line 212: Since Relu is leaky relu with alpha=0 , it would be good not to introduce  attribute "fuse_leaky_relu" and use "fuse_relu" along with "fuse_leaky_relu_alpha" to perform leaky relu and relu.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -690,8 +702,9 @@ class ConvMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
     // as well as attributes of primitive to be created
     // This name will be used as key when saving info into device context
     const std::string key = platform::ConvMKLDNNHandler::GetHash(
-        src_tz, weights_tz, fuse_relu, fuse_brelu, strides, paddings, dilations,
-        groups, ctx.op().Input("Input") + ctx.op().Input("Filter"));
+        src_tz, weights_tz, fuse_relu, fuse_leaky_relu, fuse_brelu, strides,
```
jczaja, line 705: Fuses are available for Inference only, so technically speaking fuse_leaky_relu should always be "false" here , hence no point of reading and declaring it here.

jczaja (approved):
Some minor thing to improve (this PR or next one) , but in general LGTM

paddle/fluid/operators/conv_op.cc
grygielski: That's a good point. Since I'm currently working on rewriting whole conv+activation pass creation strategy, I won't change it in this PR. Its main point is checking if it speeds up face model.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
grygielski: I will take this into account in final solution, thanks!

paddle/fluid/operators/conv_op.cc
luotao1: Please remove the attr in this PR. Thanks very much!
The reason is that if you add the attr here, then all the models trained with the develop commit will have above attrs in there program_desc.

luotao1:
> Could you please test Face model performance with this PR and share your findings 

I will test it ASAP. But could you solve the error when you test the face model?

paddle/fluid/operators/conv_op.cc
grygielski: If I will finish my general conv+activation pass by tommorow I will just close this PR and go with updated one since there is no point in merging old one then. However, I really would like to know if it works on face model before I put major changes.

jczaja:
@luotao1 I will take a look on Face Model crash that we see on our side

paddle/fluid/operators/conv_op.cc
luotao1:
```
I0731 15:21:47.283124 25431 helper.h:326] ====== batch latency: 49.1396ms, number of samples: 1, sample latency: 49.1396ms, fps: 20.3502, data type: float ======
W0731 15:21:47.284312 25431 profiler.cc:89] CUDA CUPTI is not enabled

------------------------->     Profiling Report     <-------------------------

Place: CPU
Time unit: ms
Sorted by total time in descending order in the same thread

Event                        Calls       Total       Min.        Max.        Ave.        Ratio.
thread0::conv2d_transpose    400         3886.4      2.79948     45.771      9.71599     0.79294
thread0::conv2d              400         886.211     0.748841    4.3301      2.21553     0.180813
thread0::batch_norm          300         51.3999     0.120103    1.15955     0.171333    0.0104871
thread0::relu                300         40.7469     0.071005    0.297146    0.135823    0.00831358
thread0::load_combine        1           12.8519     12.8519     12.8519     12.8519     0.00262217
thread0::tanh                100         12.3476     0.11473     0.309273    0.123476    0.00251927
thread0::fetch               100         10.7564     0.098268    0.300906    0.107564    0.00219462
thread0::feed                100         0.54181     0.004738    0.006375    0.0054181   0.000110545
```
It seems no speedup on Model level.

grygielski:
I'm closing this one since newer version of code is submitted in PR #18942

PR #18578, opened 2019-07-10
not use transferscope cache in cpu case

LeoZhao-BJ:
For naiveexecutor on CPU model, do not use transferscope cache, and release new created transferscope at the end of op RunImpl() to avoid memory leak.

This PR is also trying to fix https://github.com/PaddlePaddle/Paddle/issues/15032#issuecomment-509959284

PR #18879, opened 2019-07-29
Revert "use static variable to do cache instead of thread local in thread frequent switching case

LeoZhao-BJ:
Revert (#18428)

This reverts commit ce38bb5341b6904e060a3bf0b934f811ae2bab4b.

This PR was to fix memory leak in transfer scope caching case for CPU, but new fix #18578 merge to disable transfer scope caching for CPU, and there is no random fail CI on MKLDNN, so 
revert this PR.


luotao1 (approved):
LGTM

PR #18888, opened 2019-07-30
Optimize the error report information when loadcombine fail to open m…

wopeizl:
Optimize the error report information when load combine fail to open model files test=develop


junjun315 (approved):
LGTM

PR #18911, opened 2019-07-30
[MKL-DNN] Removing  transfering X from FWD to GRAD via device context

jczaja:
Changes disscused here are removing transfering of Tensor X from FWD activation op to backward activation op via device context (this was not working properly for parallel executor based trainings).
Instead for mkl-dnn activations Tensor X is an input , hence we take needed info from X tensor directly : x->format(), x_data and we recreate mkl-dnn memory object : src_memory based on those two transfered quantities.

### DIFF
--- a/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
@@ -103,24 +103,8 @@ void eltwise_forward(const framework::ExecutionContext &ctx,
 
   bool is_test = ctx.Attr<bool>("is_test");
 
-  std::string key = platform::MKLDNNHandler::GetHash(
-      src_tz, std::to_string(algorithm) + std::to_string(alpha) +
-                  std::to_string(beta) + ctx.op().Input("X"));
-
-  // TODO(jczaja): Make it Thread safe
-  // save input data and layout to be referred in backward path
-  const std::string key_src_data = key + "@eltwise_fwd_src_data";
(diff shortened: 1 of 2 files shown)


jczaja:
@grygielski Please review this PR

grygielski:
LGTM

paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
```diff
@@ -169,47 +153,43 @@ void eltwise_grad(const framework::ExecutionContext &ctx,
...
   auto diff_y_format =
       diff_dst_tz.size() == 2 ? mkldnn::memory::format::nc : diff_y->format();
 
   auto diff_dst_md = platform::MKLDNNMemDesc(
       diff_dst_tz, platform::MKLDNNGetDataType<T>(), diff_y_format);
 
   std::string key = platform::MKLDNNHandler::GetHash(
-      diff_dst_tz, std::to_string(algorithm) + std::to_string(alpha) +
-                       std::to_string(beta) + ctx.op().Input("X"));
+      diff_dst_tz, std::to_string(algorithm) + std::to_string(src_format) +
```
kbinias: What do you think about moving the code responsible for creating the key for one function ?

jczaja: That's a good suggestions. I have updated changes

jczaja:
@luotao1 Could you please advice on failure of PR_Windows_CI (there is checkout failure) ?

kbinias (approved):
LGTM

PR #18919, opened 2019-07-31
Update CompiledProgram

chengduoZH:
This PR's work includes:
1. Add arguments(`build_strategy` ) for `CompiledProgram`.
2. Add check when saving the program.

After talking with the core team, we decide that if the `program` that passed to `executor.run` is a `CompiledProgram`, the framework will use `ParallelExecutor` to run this program, otherwise, the framework will use `Executor` to run this program.


python/paddle/fluid/compiler.py
```diff
@@ -193,17 +217,15 @@ def with_data_parallel(self,
...
-            if not isinstance(places, (list, tuple)):
-                places = [places]
-            self._places = places
-        else:
-            self._places = None
-        self._build_strategy.is_distribution = _is_pserver_mode(self._program)
+        self._places = places
+
+        if _has_backward_op(self._graph):
+            assert self._loss_name is not None, "The loss_name should be set here."
```
chengduoZH, line 232: For training, the default loss@GRAD should be `1/dev_count`, this is set in ParallelExecutor, and this process needs to use `loss_name `, otherwise, the loss@GRAD is 1.

python/paddle/fluid/tests/unittests/parallel_executor_test_base.py
```diff
@@ -76,7 +76,7 @@ def run_executor(exe, binary, feed, fetch_list):
                 build_strategy=build_strategy,
                 exec_strategy=exec_strategy)
         else:
-            binary = compiler.CompiledProgram(main)
```
chengduoZH, line 91: In origin implementation, the `binary` is executed by `executor` as a `program`, but now, the `compiled program` is executed by `parallel executor`.

python/paddle/fluid/compiler.py
```diff
@@ -101,9 +109,13 @@ class CompiledProgram(object):
...
             further optimizations. Note: graph is only supported when compiled
             with with_data_parallel option.
+        build_strategy(BuildStrategy): build_strategy is used to
+            build the graph with the specified options.
+            For more information, please refer to fluid.BuildStrategy.
+            Default None.
     """
 
-    def __init__(self, program_or_graph):
+    def __init__(self, program_or_graph, build_strategy=None):
```
chenwhql, line 118: the general use case for `CompiledProgram` is like `cp = fluid.CompiledProgram().with_data_parallel()`, now the `build_strategy` is the argument of `CompiledProgram.__init__` and `with_data_parallel`  at the same time, in the extreme case, is there a way of writing like `p = fluid.CompiledProgram(build_strategy).with_data_parallel(build_strategy)`?

python/paddle/fluid/compiler.py
```diff
@@ -238,10 +252,13 @@ def with_inference_optimize(self, config):
     def _with_distributed(self):
         raise NotImplementedError()
 
-    def _compile_data_parallel(self, use_cuda=False, scope=None):
+    def _compile_data_parallel(self, use_cuda=False, scope=None, places=None):
```
chenwhql: why not move `places` to front position, we declared the `places` arg can be `None`, but in the function, we use `assert places` limit it cannot be `None`, it's a bit awkward.

python/paddle/fluid/compiler.py
```diff
@@ -335,13 +357,28 @@ def _compile(self, scope, place):
...
+
+        if self._is_inference:
             self._executor = self._compile_inference()
         else:
-            p = _place_obj(self._place)
-            self._executor = core.Executor(p)
+            if self._is_data_parallel:
+                self._places = self._get_places(self._place, self._places)
+            else:
+                self._places = [self._place]
```
chenwhql, line 376: in this branch, `_is_data_parallel` is still `False`, then compiled, return to `program._compile()` in executor.py, is it get into the else branch? it will invoke `_run_program` and execute program by `Executor`, the `_compile_data_parallel` here is meaningless?

python/paddle/fluid/compiler.py
```diff
@@ -122,6 +134,11 @@ def __init__(self, program_or_graph):
         self._compiled = False
         self._is_data_parallel = False
         self._is_inference = False
+        self._loss_name = None
+        self._share_vars_from = None
+        self._places = None
+        self._build_strategy = build_strategy
+        self._exec_strategy = None
 
     def with_data_parallel(self,
```
chenwhql, line 143: I think we can set the `is_data_parallel`, `is_inference` to the arguments of `CompiledProgram`, now if user just use `CompiledProgram` but not invoke `with_data_parallel`, the program is still executed by `Executor`, In this case, the program is not `compiled` in fact.

python/paddle/fluid/compiler.py
chengduoZH: In this case, the `build_strategy` of `with_data_parallel()` will be overwritten if users pass `build_strategy` when creating `CompiledProgram`.

python/paddle/fluid/compiler.py
chengduoZH: > I think we can set the is_data_parallel, is_inference to the arguments of CompiledProgram

I don't think this is better.

> now if user just use CompiledProgram but not invoke with_data_parallel, the program is still executed by Executor, In this case, the program is not compiled in fact.

Please read this introduction: https://github.com/PaddlePaddle/Paddle/pull/18919#issue-302769160

python/paddle/fluid/compiler.py
chengduoZH: Please read this introduction first: https://github.com/PaddlePaddle/Paddle/pull/18919#issue-302769160

chenwhql (dismissed):
LGTM

python/paddle/fluid/compiler.py
```diff
@@ -101,9 +109,13 @@ class CompiledProgram(object):
             (potentially optimized before), it will be directly used for
             further optimizations. Note: graph is only supported when compiled
             with with_data_parallel option.
+        build_strategy(BuildStrategy): build_strategy is used to
+            build the graph with the specified options.
+            For more information, please refer to fluid.BuildStrategy.
```
lanxianghit, line 114: 给一个fluid.BuildStrategy官网文档的链接吧

python/paddle/fluid/compiler.py
lanxianghit: what about rm the ‘build_strategy’ from with_data_parallel？

python/paddle/fluid/compiler.py
```diff
@@ -172,9 +189,11 @@ def with_data_parallel(self,
         Args:
             loss_name (str): The loss name must set in training. Default None.
             build_strategy(BuildStrategy): build_strategy is used to
-                build the graph so it can run on multiple devices/cores with
-                optimized topology.
+                build the graph with the specified options.
                 For more information, please refer to fluid.BuildStrategy.
+                Note that, if you set build_strategy in the argument list when
+                creating CompiledProgram and calling with_data_parallel,
+                the build_strategy in CompiledProgram will be overwritten by the latter.
```
lanxianghit, line 196: 在多个地方都可以设置build_strategy 并且有优先级的问题，可能会对用户造成困惑，能否考虑只保留一个设置的入口？

python/paddle/fluid/compiler.py
```diff
@@ -172,9 +189,11 @@ def with_data_parallel(self,
         Args:
             loss_name (str): The loss name must set in training. Default None.
             build_strategy(BuildStrategy): build_strategy is used to
-                build the graph so it can run on multiple devices/cores with
-                optimized topology.
+                build the graph with the specified options.
                 For more information, please refer to fluid.BuildStrategy.
+                Note that, if you set build_strategy in the argument list when
```
lanxianghit, line 194: 这里最好也能给一个fluid.BuildStrategy官网文档的链接

python/paddle/fluid/compiler.py
chengduoZH: Considering the compatibility problem, the original interface parameters can not be modified.

chengduoZH: Currently, the build_strategy can be set during creating CompiledProgram or calling with_data_parallel, and it may be confusing, but in the long run, we should set up build_strategy only when creating CompiledProgram, and exec_strategy should be deprecated.

python/paddle/fluid/compiler.py
chengduoZH: 我也觉应该有链接，但是不止是fluid.BuildStrategy，其他也应该有，比如exec_strategy，program，executor等，不清楚文档方面是否可以统一做一下?

python/paddle/fluid/compiler.py
chengduoZH: 这里有解释 https://github.com/PaddlePaddle/Paddle/pull/18919/files#r315989245

lanxianghit (approved):

guoshengCS (approved):

XiaoguangHu01 (approved):

PR #18928, opened 2019-07-31
replace cifar dataset url

tink2123:
To speed up the dataset download ，replace dataset url to bce
* cifar
* imdb
* imikolov
* movielens


kuke (dismissed):
LGTM

PR #18940, opened 2019-07-31
fix g_param shape mismatch in WeightNormParamAttr

SunGaofeng:
1, fix g_param shape mismatch between startup_program and main_program
2, fix 'dim' attributes from int to list(int), when append reduce_sum op

### DIFF
--- a/python/paddle/fluid/layer_helper_base.py
+++ b/python/paddle/fluid/layer_helper_base.py
@@ -244,6 +244,8 @@ class LayerHelperBase(object):
             out=g_param,
             dim=attr.dim,
             block=self.startup_program.global_block())
+
+        # keep g_param shape to be consistent with that in main_program
         __reshape_op(
             g_param,
             g_param_shape,
--- a/python/paddle/fluid/layer_helper_base.py
+++ b/python/paddle/fluid/layer_helper_base.py
@@ -177,19 +177,24 @@ class LayerHelperBase(object):
             elif dim == 0:
                 out_shape = [x.shape[0]] + [1] * (len(x.shape) - 1)
                 reshape = __reshape_op(x, shape=[x.shape[0], -1], block=block)
-                norm = __norm_op(reshape, dim=1, block=block)
+                norm = __norm_op(reshape, dim=[1], block=block)
                 __reshape_op(norm, out=out, shape=out_shape, block=block)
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/layer_helper_base.py
```diff
@@ -177,19 +177,24 @@ def __norm_except_dim(x,
...
                 transpose = __transpose_op(x, perm, block=block)
-                norm = __norm_op(transpose, dim=0, block=block)
-                __transpose_op(norm, perm, out=out, block=block)
+                out_shape = [transpose.shape[0]] + [1] * (len(transpose.shape) -
+                                                          1)
+                reshape = __reshape_op(
+                    transpose, shape=[transpose.shape[0], -1], block=block)
+                norm = __norm_op(reshape, dim=[1], block=block)
+                reshape2 = __reshape_op(norm, shape=out_shape, block=block)
+                __transpose_op(reshape2, perm, out=out, block=block)
```
guoshengCS, line 197: Since the shape of output is a vector, maybe we can remove this line `__transpose_op(reshape2, perm, out=out, block=block)` by redefining `out_shape` in `reshape2` as `[1...1, transpose.shape[0], 1...1]` .

python/paddle/fluid/param_attr.py
```diff
@@ -180,14 +180,14 @@ def _to_kwargs(self, with_initializer=False):
...
-    in a neural network that decouples the length of those weight vectors from
+    in a neural network that decouples the magnitude of those weight vectors from
     their direction. Weight Norm has been implemented as discussed in this
     paper: `Weight Normalization: A Simple Reparameterization to Accelerate
     Training of Deep Neural Networks
     <https://arxiv.org/pdf/1602.07868.pdf>`_.
 
     Args:
-        dim(list): The parameter's name. Default None.
+        dim(int): Dimension over which to compute the norm. Default None.
```
guoshengCS, line 190: Might it be the  Dimension except which

SunGaofeng: this is referred to pytorch, which said 'dim (int, optional): dimension over which to compute the norm
', just keep consistent with other framework

guoshengCS (approved):

heavengate (approved):

PR #18942, opened 2019-07-31
Add generalized Conv+Activation MKLDNN fuse pass creation

grygielski:
I changed the way conv+activation mkldnn fuse passes are added. Instead adding separate files for every activation you just create specific class instance with overridden conv_type() and activation_type() functions and register pass with this class. This also cuts separate bool attributes from conv operator and introduces string fuse_activation parameter which takes name of fused activation and 2 float attributes (fuse_alpha and fuse_beta) to pass parameters to fused activation function.


paddle/fluid/framework/ir/mkldnn/conv_elementwise_add_mkldnn_fuse_pass.cc
```diff
@@ -179,8 +182,12 @@ void ResidualConnectionMKLDNNFusePass::ProjectionFuseHandle::operator()(
...
   }
 
-  auto fuse_relu = HasAttribute<bool>(*residual_conv_op, "fuse_relu");
-  if (fuse_relu && *fuse_relu) return;
+  std::string fuse_activation =
+      residual_conv_op->Op()->HasAttr("fuse_activation")
+          ? boost::get<std::string>(
+                residual_conv_op->Op()->GetAttr("fuse_activation"))
+          : "";
+  if (fuse_activation == "relu" || fuse_activation == "relu6") return;
```
Sand3r-: Since the code here and in lines 112-116 is roughly the same, please consider moving it to static function
``` C++
static bool HasFusedActivation(OpDesc* conv_op) {
  std::string fuse_activation =
      conv_op->Op()->HasAttr("fuse_activation")
          ? boost::get<std::string>(
                conv_op->Op()->GetAttr("fuse_activation"))
          : "";
   return (fuse_activation == "relu" || fuse_activation == "relu6");
}
```
So that later on you can just use `if (HasFusedActivation(conv_op/residual_conv_op) return;`

paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -68,10 +68,13 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
...
-                is_unsigned = (op->HasAttr("fuse_relu") &&
-                               boost::get<bool>(op->GetAttr("fuse_relu"))) ||
-                              (op->HasAttr("fuse_brelu") &&
-                               boost::get<bool>(op->GetAttr("fuse_brelu")));
+                std::string fuse_activation =
+                    op->HasAttr("fuse_activation")
+                        ? boost::get<std::string>(
+                              op->GetAttr("fuse_activation"))
+                        : "";
+                is_unsigned =
```
Sand3r-, line 73: Please consider introducing method in Op or OpDesc

```C++
template <typename T>
T GetAttrIfExists(std::string AttrName)
{
   T result;
    if(HasAttr(AttrName) {
        result = GetAttr(AttrName)
   }
   return result;
}
```

since the pattern of
```
std::string fuse_activation =
                    op->HasAttr("fuse_activation")
                        ? boost::get<std::string>(
                              op->GetAttr("fuse_activation"))
                        : "";
```
which is often recurring, could be simplified to `std::string fuse_activation = op_->GetAttrIfExists("fuse_activation")`

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -230,9 +230,8 @@ class MKLDNNHandler {
     AppendKey(key, std::to_string(groups));
     AppendKey(key, std::to_string(srcdt));
     AppendKey(key, std::to_string(format));
-    AppendKey(key, std::to_string(relu));
+    // AppendKey(key, fuse_activation);
```
Sand3r-: Please consider removing this AppendKey if it was commented out

Sand3r- (commented):
@grygielski Thank you so much for the changes introduced. Please refer to the suggestions provided below.

paddle/fluid/inference/api/mkldnn_quantizer.cc
grygielski: Great idea! This function will take use in many places.

paddle/fluid/platform/mkldnn_reuse.h
grygielski: I forgot to uncomment it out after debugging, thanks.

paddle/fluid/framework/ir/mkldnn/conv_elementwise_add_mkldnn_fuse_pass.cc
grygielski: Will do it. What is more I think it would be better to check if any of activation function is fused, not only these two. Thanks!

grygielski:
@luotao1 According to @Sand3r- advice I've added following code to op_desc.h:
```cpp
template <typename T>
  T GetAttrIfExists(const std::string &name) const {
    T result;
    if (HasAttr(name)) {
      result = boost::get<T>(GetAttr(name));
    }
    return result;
  }
```
and i guess I need Your permition to do so since CI_CPU_Py2 fails with following error:
`You must have one RD (XiaoguangHu01,chengduoZH,Xreki,luotao1,sneaxiy,tensor-tang) approval for the api change! paddle/fluid/framework/op_desc.h for the management reason of the underlying code for fluid.`

grygielski:
I'm having troubles with passing some of test. PR_CI_Coverage failed only with 'test_buffer_shared_memory_reuse_pass' timeout which has been already fixed but PR_CI keeps failing with some int8 test either with '-0.0<0.0' error or timeout. The thing is that exactly same tests passed in different CIs. I even tried to recreate same errors on our machine with Broadwell E5-2699 as it is similar to what is used to execute CIs. I did not get any errors and received following time data: image It's hard for me to try fixing it because when I rerun CI here I have to wait about 10h or longer and I still don't know why it sometimes fails. Could you advice me on that @luotao1 ?

luotao1:
1. You can ignore `PR_CI` which is not required now.
2. I see the fail in `PR_CI_Coverage (Paddle)`: Python unit-test fails.
```
[20:46:51]	[Step 1/1] Assert Python Diff Coverage
[20:46:51]	[Step 1/1] expected >= 90.0 %, actual 0.0 %, failed
```
image
http://ci.paddlepaddle.org/viewLog.html?buildId=137676&buildTypeId=Paddle_PrCiCoverage&tab=report_project7_Python_Diff_Coverage
`Inference_transpiler` is already deprecated, please do not add codes in it or you should add complete test for it. 
3. #19033 add CI new rule: `You must have Dianhai (Owner of Paddle Repo) approval for change 20+ files or add than 1000+ lines of content`. Please split your PR if possible. @bingyanghuang Please inform this new rule to Intel friends.

grygielski:
Thank you a lot! I will split it on functional and cleanup PRs. I will close this PR when everything will be fine and merged.

grygielski:
Closing this one since divided parts (https://github.com/PaddlePaddle/Paddle/pull/19072 and https://github.com/PaddlePaddle/Paddle/pull/19237) were merged.

PR #18953, opened 2019-08-01
Integrate distributed classification

gavin1332:


python/paddle/fluid/framework.py
```diff
@@ -691,6 +691,35 @@ def _set_error_clip(self, error_clip):
         """
         self.error_clip = error_clip
 
+    def _set_info(self, key, value):
+        """
+        Set some key-value information associated to this variable.
+
+        Args:
+            key(str): The key of the information entry.
+            value(object): The calue of the information entry.
```
sandyhouse: (typo) The **value** of the information entry.

python/paddle/fluid/layers/collective.py
```diff
@@ -178,3 +183,223 @@ def _c_sync_comm_stream(x, ring_id):
...
+        # avoid distributed parameter allreduce gradients
+        weight.is_distributed = True
+        # avoid distributed parameter broadcasting in startup program
+        default_startup_program().global_block().vars[
+            weight.name].is_distributed = True
+
+        bias = None
+        if use_bias:
+            bias = helper.create_parameter(
+                shape=shard_dim, dtype=dtype, is_bias=True)
```
sandyhouse: Should shape be in_dim if transpose_weight is True?

python/paddle/fluid/framework.py
gavin1332: done

python/paddle/fluid/layers/collective.py
gavin1332: not necessary

Issue #18703, opened 2019-07-19
动态图Conv2DTranspose bias_attr, act参数不生效


DDDivano:
https://www.paddlepaddle.org.cn/documentation/docs/en/1.5/api/dygraph.html#conv2dtranspose

data = np.ones(shape=(1, 1, 2, 2), dtype=np.float32)
data = to_variable(data)
conv2dtrans = fluid.dygraph.Conv2DTranspose("conv2dtrans", 1, filter_size=2,
                                                    param_attr=fluid.initializer.ConstantInitializer(value=1),
                                                    bias_attr=fluid.initializer.ConstantInitializer(value=1),
                                                    act="sigmoid")
res = conv2dtrans(data)
print(res)

![image](https://user-images.githubusercontent.com/10734244/61526195-8826ee80-aa4c-11e9-8912-3a78b89247e1.png)

PR #18968, opened 2019-08-01
fix con2d transpose bias by create and init it in build_once 

JiabinYang:
This PR will fix con2d transpose bias by create in build_once instead of doing it in forward


JiabinYang:
> This PR will fix con2d transpose bias by create in build_once instead of doing it in forward

And Then we fix problem of  #18774 #18703 #19200 , GRUUnit original mode has no effect and add sample weight part for NCE model

PR #18836, opened 2019-07-26
Open gc by default

sneaxiy:
Enable GC when `PADDLE_ON_INFERENCE` is off.

This PR also fixes bug of GC when using conditional_block op.

See [CE Debug](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=54474&buildTypeId=PaddleModesl_ModelDebug_Model11Build&tab=buildLog) here.

There are 3 models in CE Debug failed.
```
[22:30:44][Step 3/3] task: model_cycle_gan
[22:30:44][Step 3/3] passed:  False
[22:30:44][Step 3/3] infos [d_train_cost] pass
[22:30:44][Step 3/3] [g_train_cost] pass
[22:30:44][Step 3/3] [duration] failed, diff ratio: 0.06092244084538961 larger than 0.05.
[22:30:44][Step 3/3] kpis keys ['d_train_cost', 'g_train_cost', 'duration']
[22:30:44][Step 3/3] kpis values [[[2.12036180496]], [[12.1869783401]], [[1.05452570915]]]
[22:30:44][Step 3/3] task: model_icnet
[22:30:44][Step 3/3] passed:  False
[22:30:44][Step 3/3] infos [train_cost] pass
[22:30:44][Step 3/3] [train_duration] failed, diff ratio: 0.04116670226811877 larger than 0.02.
[22:30:44][Step 3/3] kpis keys ['train_cost', 'train_duration']
[22:30:44][Step 3/3] kpis values [[[2.1491]], [[41.469499]]]
[22:30:44][Step 3/3] task: model_ocr_recognition
[22:30:44][Step 3/3] passed:  False
[22:30:44][Step 3/3] infos [train_cost] pass
[22:30:44][Step 3/3] [train_duration] failed, diff ratio: 0.06440406463978306 larger than 0.06.
[22:30:44][Step 3/3] kpis keys ['train_cost', 'train_duration']
[22:30:44][Step 3/3] kpis values [[[58.981466]], [[46.207872]]]
[22:31:02][Step 3/3] model_cycle_gan model_icnet model_ocr_recognition 
```

- model_cycle_gan: CE Debug uses legacy CycleGAN code in [release/1.1](https://github.com/PaddlePaddle/models/tree/release/1.1) branch, which uses Executor and there is no cache. In [develop](https://github.com/PaddlePaddle/models/tree/develop) branch of models, GC is enabled in CycleGAN model. 
[...]

PR #18983, opened 2019-08-02
Try to deprecate unstable python memory optimize

sneaxiy:
We have enabled [inplace](https://github.com/PaddlePaddle/Paddle/pull/17911) and [GC](https://github.com/PaddlePaddle/Paddle/pull/18836) by default. We do not need the legacy `fluid.memory_optimize` interface,  since this API is not stable.

After this PR, `fluid.memory_optimize` does not take any effect on the program.

CE Debug has passed. See [here](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=54897&buildTypeId=PaddleModesl_ModelDebug_Model11Build&tab=buildLog). `dcgan_d_train_cost` fails because of the previous PR not this one.


chengduoZH (dismissed):
LGTM

python/paddle/fluid/tests/unittests/parallel_executor_test_base.py
```diff
@@ -68,14 +67,11 @@ def run_executor(exe, binary, feed, fetch_list):
             loss = method(use_feed=feed_dict is not None)
             # NOTE(zjl): memory_optimize/inplace pass would not require 
             # that loss.persistable = True 
-            loss.persistable = memory_opt
+            loss.persistable = False
```
chengduoZH: Why do you set `False` here?

chengduoZH (dismissed):
LGTM

python/paddle/fluid/tests/unittests/parallel_executor_test_base.py
sneaxiy: Because all memory optimization strategies do not need `loss.persistable = True`. I set `loss.persistable = False` intentionally to test whether the memory optimization strategy is right.

luotao1 (approved):
LGTM

python/paddle/fluid/tests/unittests/test_deprecated_memory_optimize_interfaces.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2019 PaddlePaddle Authors. All Rights Reserved.
```
luotao1, line 1: 请问什么时候会完全删除？包括删除`test_deprecated_memory_optimize_interfaces.py`?

sneaxiy: 接口壳子应该得保留一段时间，预计1.6会作为缓冲期，1.6以后接口可直接删除。

PR #18988, opened 2019-08-02
Replace Relu with bounded Relu in MobileNetV2 quantization mkldnn

wozna:
It removes the workaround from  PR #17570 

In MKL-DNN 0.18 there was no support for bounded ReLU post-op. Hence, when adding support for MobileNetV2 support we decided to use ReLU instead of ReLU6 (bounded relu). 
As the MKL-DNN version is upgraded to 0.20 now and it supports bounded ReLU post-op, the workaround can be removed and bounded ReLU can be used. 

test=develop


wozna:
@Sand3r-  I am asking for a review

bingyanghuang:
@lidanqing-intel could you help review this PR?

lidanqing-vv:
Looks good to me~
    Later we may consider to call `DequantizeOutput`  after ` conv_op->Op()->SetAttr("fuse_brelu_threshold", scale_out * threshold)` , not before, because `DequantizeOutput` uses `conv_op`.  But anyway conv_op is passed by pointer so it works well.

PR #18990, opened 2019-08-03
make listen and server as exclusive run

guru4elephant:
Make test_listen_and_serv_op run exclusively, or it will affect other distribution unit test randomly.


gongweibao (approved):
LGTM

PR #19004, opened 2019-08-05
support debug each output of each ins

Thunderbrook:
In debug mode, we need to know each output, so in this commit support to dump each output tensor of each instance into afs.


python/paddle/fluid/incubate/fleet/parameter_server/pslib/__init__.py
```diff
@@ -485,7 +501,7 @@ def minimize(self,
                           parameter_list,
                           no_grad_set,
                           self._strategy)
-
+        opt_info["mpi_rank"] = fleet._role_maker._get_rank()
```
xjqbest, line 577: maybe trainer_rank or trainer_id is better，role_maker can have many types (mpi or paddlecloud, etc.)  and  all of them can get_rank

paddle/fluid/framework/downpour_worker.cc
```diff
@@ -72,6 +74,85 @@ void DownpourWorker::Initialize(const TrainerDesc& desc) {
...
+  }
+  need_dump_field_ = false;
+  if (dump_fields_.size() != 0 && dump_fields_path_ != "") {
+    need_dump_field_ = true;
+  }
+  mpi_rank_ = desc.mpi_rank() / 2;
+}
+
+template <typename T>
+std::string print_lod_tensor_type(LoDTensor* tensor, int64_t start,
```
xjqbest: code style of function name is wrong,  so as other functions

Thunderbrook: done

yaoxuefeng6 (approved):
LGTM

xjqbest (approved):

PR #19016, opened 2019-08-05
fix QueueDataset queue size

xjqbest:
fix QueueDataset queue size，set queue size =  batch size * 100, to avoid too many instances in channel  when training is much slower than reading data.


yaoxuefeng6 (commented):
LGTM

Thunderbrook (approved):

PR #19020, opened 2019-08-06
Fix memory_reuse_pass memory_size calculation error

sneaxiy:
The original memory size calculation method is wrong. It should multiply size of data type.


chengduoZH (approved):
LGTM

PR #19030, opened 2019-08-06
Fix memory overwriting of tensors returned by executor

zhiqiu:
The original fetch behavior in executor.run() creates a persistable variable named 'fetch', and returns reference of the tensors held in 'fetch'; thus when two program executed, memory overwriting of tensors returned may appear when executors run with`return_numpy=False`.

This PR adds a unittest `(test_executor_return_tensor_not_overwriting.py)` to test memory overwriting behavior, which fails on original code,   and it fixes the bug by returning different tensors in executor.run().

Plus, this PR fixes wrong usage of ParallelExecutor in op_test.py. It seems the unittests of operators never run in parallel mode.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_executor_return_tensor_not_overwriting.py
+++ b/python/paddle/fluid/tests/unittests/test_executor_return_tensor_not_overwriting.py
@@ -16,10 +16,11 @@ from __future__ import print_function
 import unittest
 import numpy as np
 import paddle.fluid.core as core
+import paddle.fluid as fluid
 from op_test import OpTest
 
 
-class TestFetchList(OpTest):
+class TestExecutorReturnTensorNotOverwritingWithOptest(OpTest):
     def setUp(self):
         pass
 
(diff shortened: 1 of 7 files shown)


python/paddle/fluid/tests/unittests/test_fetch_list.py
```diff
@@ -0,0 +1,70 @@
+#  Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
sneaxiy: This filename is too simple to know what it works for. Please rename the filename. By the way, should we add unittests independent of op_test, which is only used to test whether `exe.run(return_numpy=False)` is right?

zhiqiu: Done, thanks!

python/paddle/fluid/tests/unittests/test_executor_return_tensor_not_overwriting.py
```diff
@@ -0,0 +1,116 @@
...
+        }
+        self.outputs = {'Out': self.out}
+        self.op_type = "elementwise_mul"
+        self.dtype = np.float32
+        outs, fetch_list = self._calc_output(place, parallel=parallel)
+        return outs
+
+    def test_executor_run_twice(self):
+        places = [fluid.CPUPlace()]
+        if fluid.is_compiled_with_cuda() and fluid.core.op_support_gpu(
```
sneaxiy: Remove `op_support_gpu`. It is not necessary. Leaving `fluid.is_compiled_with_cuda()` is ok.

sneaxiy (approved):
Excellent! LGTM.

python/paddle/fluid/tests/unittests/test_executor_return_tensor_not_overwriting.py
zhiqiu: Done, thanks.

PR #19032, opened 2019-08-06
Refine inplace inference registry

sneaxiy:
This PR use a macro to unify `InplaceOpInference` definition. This is designed to refine codes and support extensions in the future (i.e., to support SelectedRows, we should add parameter inside `InplaceOpInference::operator()`).


paddle/fluid/operators/batch_norm_op.cc
```diff
@@ -598,36 +598,13 @@ std::unique_ptr<framework::OpDesc> BatchNormGradMaker::Apply() const {
   return std::unique_ptr<framework::OpDesc>(op);
 }
 
-class BatchNormInplaceInToOut : public framework::InplaceOpInference {
```
zhhsplendid, line 601: Delete but no DECLARE_INPLACE_OP_INFERER?

zhhsplendid (approved):

paddle/fluid/operators/batch_norm_op.cc
sneaxiy: This is cleaning unused codes.

PR #19046, opened 2019-08-06
Fix the inference library compilation bug on windows

zhwesky2010:
Fix the inference library compilation bug on windows. The current bug is that most of files under the inference library path are missed, which is caused by the logic error of the code. Therefore, the code of copying files has been overwritten.

### DIFF
--- a/cmake/inference_lib.cmake
+++ b/cmake/inference_lib.cmake
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 # make package for paddle fluid shared and static library
+
+ set(COPY_SCRIPT_DIR ${PADDLE_SOURCE_DIR}/cmake)
 function(copy TARGET)
     set(options "")
     set(oneValueArgs "")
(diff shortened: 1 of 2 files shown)


cmake/inference_lib.cmake
```diff
@@ -26,42 +28,14 @@ function(copy TARGET)
...
-                foreach (src_file ${src_files})
-                    add_custom_command(TARGET ${TARGET} PRE_BUILD
-                            COMMAND ${CMAKE_COMMAND} -E copy "${src_file}" "${dst}"
-                            COMMENT "copying ${src_file} -> ${dst}")
-                endforeach ()
-            endif()
-        else (WIN32) # not windows
+        if (WIN32)   #windows
+            add_custom_command(TARGET ${TARGET} POST_BUILD
+                    COMMAND python ${COPY_SCRIPT_DIR}/copyfile.py ${src} ${dst})
```
wopeizl, line 37: 不能直接用python，需要用cmake传入的python变量，否则有些环境会出错

zhwesky2010: OK，我改一下

PR #19071, opened 2019-08-08
datafeed support reading to cuda place directly.

hutuxian:
DataFeed support reading to cuda place directly.
Previously, only CPUPlace is supported. So we add a place field in DataFeed to denote which place it will feed data to.


gavin1332 (dismissed):
LGTM

gavin1332 (approved):
LGTM

PR #19079, opened 2019-08-09
Remove unused inplace act codes

sneaxiy:
`IsInplace` is not used now, because we have inplace pass. Remove these codes.


luotao1 (approved):
LGTM

PR #19101, opened 2019-08-09
Enhance OpTest to check the consistency of operators when using and not using inplace

zhiqiu:
This PR enhances OpTest to check the consistency of operators when using and not using `inplace`.
The `inplace` check is done imperceptibly inside `OpTest.check_output_with_place` with almost no modification of original test cases.

For most ops, the `inplace` check use `np.array_equal`  to ensure identical outputs. And for few ops which can not ensure computational consistency, such as `group_norm` which use `atomicAdd`, the `inplace` check use `np.allclose`  to ensure small difference by setting  `check_output(xx, inplace_atol = 1e-4)`.

Plus, the original `inplace` implementation of `sum` can not pass `inplace` check because of different computational order , which is fixed in this PR.

### DIFF
--- a/paddle/fluid/framework/op_info.h
+++ b/paddle/fluid/framework/op_info.h
@@ -79,9 +79,13 @@ struct OpInfo {
   }
 
   // some op has no grad_op_maker, can't use GradOpMaker() ?
-  bool HasGradOpMaker() const { return grad_op_maker_ != NULL ? true : false; }
+  bool HasGradOpMaker() const {
+    return grad_op_maker_ != nullptr ? true : false;
+  }
 
-  bool HasInferInplace() const { return infer_inplace_ != NULL ? true : false; }
+  bool HasInferInplace() const {
+    return infer_inplace_ != nullptr ? true : false;
+  }
 
   const OpAttrChecker* Checker() const { return checker_; }
 
--- a/paddle/fluid/framework/op_info.h
+++ b/paddle/fluid/framework/op_info.h
@@ -129,8 +129,6 @@ class OpInfoMap {
 
   std::vector<std::string> GetUseDefaultGradOpDescMakerOps() const;
 
-  std::vector<std::string> GetHasInferInplaceOps() const;
-
  private:
   OpInfoMap() = default;
   std::unordered_map<std::string, OpInfo> map_;
--- a/paddle/fluid/framework/op_info.h
+++ b/paddle/fluid/framework/op_info.h
@@ -120,6 +120,8 @@ class OpInfoMap {
 
   std::vector<std::string> GetUseDefaultGradOpDescMakerOps() const;
 
+  std::vector<std::string> GetHasInferInplaceOps() const;
+
  private:
   OpInfoMap() = default;
   std::unordered_map<std::string, OpInfo> map_;
--- a/paddle/fluid/framework/op_info.h
+++ b/paddle/fluid/framework/op_info.h
@@ -78,7 +78,7 @@ struct OpInfo {
     return grad_op_maker_;
   }
 
-  // some op has no grad_op_maker, can't use GradOpMaker() ?
+  // some op has no grad_op_maker, add check before use GradOpMaker()
   bool HasGradOpMaker() const {
     return grad_op_maker_ != nullptr ? true : false;
   }
--- a/paddle/fluid/framework/op_info.h
+++ b/paddle/fluid/framework/op_info.h
@@ -78,6 +78,11 @@ struct OpInfo {
     return grad_op_maker_;
   }
 
+  // some op has no grad_op_maker, can't use GradOpMaker() ?
+  bool HasGradOpMaker() const { return grad_op_maker_ != NULL ? true : false; }
+
+  bool HasInferInplace() const { return infer_inplace_ != NULL ? true : false; }
+
   const OpAttrChecker* Checker() const { return checker_; }
 
   const InferNoNeedBufferVarsFN& NoNeedBufferVarsInferer() const {
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -176,6 +176,9 @@ PYBIND11_MODULE(core_noavx, m) {
   m.def("_get_use_default_grad_op_desc_maker_ops",
         [] { return OpInfoMap::Instance().GetUseDefaultGradOpDescMakerOps(); });
 
+  m.def("_get_has_infer_inplace_ops",
+        [] { return OpInfoMap::Instance().GetHasInferInplaceOps(); });
+
   // NOTE(zjl): ctest would load environment variables at the beginning even
   // though we have not `import paddle.fluid as fluid`. So we add this API
   // to enable eager deletion mode in unittest.
@@ -490,10 +493,17 @@ PYBIND11_MODULE(core_noavx, m) {
            Returns:
                out (Tensor): new Tensor(NOT LoDTensor).
            )DOC")
-      .def("__str__", [](const LoDTensor &self) {
-        std::stringstream ostr;
-        ostr << self;
-        return ostr.str();
+      .def("__str__",
+           [](const LoDTensor &self) {
+             std::stringstream ostr;
+             ostr << self;
+             return ostr.str();
+           })
+      .def("_copy", [](const LoDTensor &self) {
+        LoDTensor dst;
+        TensorCopySync(self, self.place(), &dst);
+        dst.set_lod(self.lod());
+        return dst;
       });
 
   py::class_<SelectedRows>(m, "SelectedRows")
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -86,6 +86,7 @@ limitations under the License. */
 DEFINE_bool(reader_queue_speed_test_mode, false,
             "If set true, the queue.pop will only get data from queue but not "
             "remove the data from queue for speed testing");
+DECLARE_bool(use_mkldnn);
 
 // disable auto conversion to list in Python
 PYBIND11_MAKE_OPAQUE(paddle::framework::LoDTensorArray);
@@ -734,13 +735,12 @@ All parameter, weight, gradient are variables in Paddle.
         return std::make_pair(grad_op_desc_ptrs, grad_to_var);
       });
   m.def("has_grad_op_maker", [](const std::string op_type) {
-    auto op_info = framework::OpInfoMap::Instance().GetNullable(op_type);
-    return op_info == nullptr ? false : op_info->HasGradOpMaker();
+    return framework::OpInfoMap::Instance().Get(op_type).HasGradOpMaker();
   });
   m.def("has_infer_inplace", [](const std::string op_type) {
-    auto op_info = framework::OpInfoMap::Instance().GetNullable(op_type);
-    return op_info == nullptr ? false : op_info->HasInferInplace();
+    return framework::OpInfoMap::Instance().Get(op_type).HasInferInplace();
   });
+  m.def("get_flags_use_mkldnn", []() { return FLAGS_use_mkldnn; });
 
   m.def("prune", [](const ProgramDesc &origin,
                     const std::vector<std::array<size_t, 2>> &targets) {
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -737,10 +737,12 @@ All parameter, weight, gradient are variables in Paddle.
         return std::make_pair(grad_op_desc_ptrs, grad_to_var);
       });
   m.def("_has_grad_op_maker", [](const std::string op_type) {
-    return framework::OpInfoMap::Instance().Get(op_type).HasGradOpMaker();
+    auto op_info = framework::OpInfoMap::Instance().GetNullable(op_type);
+    return op_info == nullptr ? false : op_info->HasGradOpMaker();
   });
   m.def("_has_infer_inplace", [](const std::string op_type) {
-    return framework::OpInfoMap::Instance().Get(op_type).HasInferInplace();
+    auto op_info = framework::OpInfoMap::Instance().GetNullable(op_type);
+    return op_info == nullptr ? false : op_info->HasInferInplace();
   });
 
   m.def("prune", [](const ProgramDesc &origin,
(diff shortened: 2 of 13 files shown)


paddle/fluid/framework/op_info.h
```diff
@@ -70,6 +70,11 @@ struct OpInfo {
     return grad_op_maker_;
   }
 
+  // some op has no grad_op_maker, can't use GradOpMaker() ?
+  bool HasGradOpMaker() const { return grad_op_maker_ != NULL ? true : false; }
```
sneaxiy: Use `nullptr` instead. This would be something more C++ 11 style.

paddle/fluid/framework/op_info.h
```diff
@@ -70,6 +70,11 @@ struct OpInfo {
...
+
+  bool HasInferInplace() const { return infer_inplace_ != NULL ? true : false; }
```
sneaxiy: Same above. Use `nullptr` instead.

paddle/fluid/pybind/pybind.cc
```diff
@@ -490,10 +493,17 @@ PYBIND11_MODULE(core_noavx, m) {
            Returns:
                out (Tensor): new Tensor(NOT LoDTensor).
            )DOC")
-      .def("__str__", [](const LoDTensor &self) {
-        std::stringstream ostr;
-        ostr << self;
-        return ostr.str();
+      .def("__str__",
```
sneaxiy, line 494: Why need this method?

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +357,132 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
         # fetch_list = map(block.var, fetch_list)
         if not isinstance(fetch_list[0], fluid.framework.Variable):
             fetch_list = list(map(block.var, fetch_list))
+
+        if enable_inplace is not None:
+            build_strategy = fluid.BuildStrategy()
+            build_strategy.enable_inplace = enable_inplace
+            build_strategy.memory_optimize = False
```
sneaxiy: Why set `build_strategy.memory_optimize = False`?

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +357,132 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
...
+
+        # get grad_op 
+        if not core._has_grad_op_maker(op.desc.type()):
+            return
+        grad_op_desc_list, op_grad_to_var = core.get_grad_op_desc(op.desc,
+                                                                  set(), [])
+        # has grad_op_maker but no grad_op ?
+        if not grad_op_desc_list:
+            return
+        grad_op_desc = grad_op_desc_list[0]
```
sneaxiy: Why only get `grad_op_desc_list[0]`? You would get all grad op descs.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +357,132 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
...
+            for arg in grad_op_desc.input_arg_names():
+                if arg in feed_map.keys():
+                    grad_feed_map[arg] = feed_map[arg]._copy()
+                else:
+                    forward_var_name = op_grad_to_var.get(arg, None)
+                    if forward_var_name is None:
+                        forward_var_name = arg
+                    for i, out in enumerate(fetch_list):
+                        if out.name == forward_var_name:
+                            grad_feed_map[arg] = forward_outs[i]._copy()
```
sneaxiy: Please `_copy` with place. Re-design the pybind method `_copy` with adding `place` parameter.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +357,132 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
...
+
+            exe = Executor(place)
+
+            build_strategy = fluid.BuildStrategy()
+            build_strategy.enable_inplace = enable_inplace
+            build_strategy.memory_optimize = False
```
sneaxiy: Why set `build_strategy.memory_optimize = False`?

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +357,132 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
...
+            compiled_program = fluid.CompiledProgram(
+                grad_program).with_data_parallel(
+                    build_strategy=build_strategy, places=place)
+            outs = exe.run(compiled_program,
+                           feed=grad_feed_map,
+                           fetch_list=grad_fetch_list,
+                           return_numpy=True)
```
sneaxiy: Why `return_numpy=False` when checking forward and `return_numpy=True` when checking backward?

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +357,132 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
...
+            compiled_prog = fluid.CompiledProgram(program).with_data_parallel(
+                build_strategy=build_strategy, places=place)
+            program = compiled_prog
+
         executor = Executor(place)
         outs = executor.run(program,
                             feed=feed_map,
                             fetch_list=fetch_list,
                             return_numpy=False)
         return outs, fetch_list
```
sneaxiy, line 385: I think that we should fetch all forward outputs instead of `fetch_list` which is not including `no_check_set`? So that we can make backward inputs to check the accuracy of backward inplace?

paddle/fluid/pybind/pybind.cc
zhiqiu: Actually, method “\_\_str\_\_”  existed before this PR.

paddle/fluid/framework/op_info.h
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Explicitly disable the memory_optimize strategy, it can be removed.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Exactly, done.

paddle/fluid/framework/op_info.h
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: It will be removed, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Set both to `return_numpy=False`, thanks.

paddle/fluid/pybind/pybind.cc
```diff
@@ -719,6 +733,15 @@ All parameter, weight, gradient are variables in Paddle.
                        [](std::unique_ptr<OpDesc> &p) { return p.release(); });
         return std::make_pair(grad_op_desc_ptrs, grad_to_var);
       });
+  m.def("has_grad_op_maker", [](const std::string op_type) {
+    auto op_info = framework::OpInfoMap::Instance().GetNullable(op_type);
+    return op_info == nullptr ? false : op_info->HasGradOpMaker();
```
sneaxiy: Why Nullable? Do not hide errors...

paddle/fluid/pybind/pybind.cc
```diff
@@ -719,6 +733,15 @@ All parameter, weight, gradient are variables in Paddle.
...
+  });
+  m.def("has_infer_inplace", [](const std::string op_type) {
+    auto op_info = framework::OpInfoMap::Instance().GetNullable(op_type);
```
sneaxiy: Why Nullable? Do not hide errors...

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +368,172 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
         # fetch_list = map(block.var, fetch_list)
         if not isinstance(fetch_list[0], fluid.framework.Variable):
             fetch_list = list(map(block.var, fetch_list))
+
+        if enable_inplace is not None:
+            build_strategy = fluid.BuildStrategy()
+            build_strategy.enable_inplace = enable_inplace
+            build_strategy.memory_optimize = False
```
sneaxiy: Why set `build_strategy.memory_optimize = False`?

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -325,6 +335,13 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
         outputs = self._get_outputs(block)
         feed_map = self.feed_var(inputs, place)
 
+        if for_inplace_grad_test is not None:
+            # set persistable for variables without buffer (shape contains 0 like shape = [0,2,5], holder_ is NULL), like XShape
```
sneaxiy: Add comments on `XShape`, something like `XShape in reshape2 op`? Many developers do not know what is `XShape`.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +368,172 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
...
+            return
+        grad_op_desc_list, op_grad_to_var = core.get_grad_op_desc(op.desc,
+                                                                  set(), [])
+        # has grad_op_maker but no grad_op ?
+        if not grad_op_desc_list:
+            return
+
+        for i, grad_op_desc in enumerate(grad_op_desc_list):
+            # grad_op can not inplace
+            if not fluid.core.has_infer_inplace(grad_op_desc.type()):
```
sneaxiy, line 443: Should we check whether there is any grad op whose `has_infer_inplace` is True? For example, if there are two grad ops, i.e., `op1` and `op2`, and `has_infer_inplace(op1)` is false but `has_infer_inplace(op2)` is true, we should still check inplace cases.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -351,13 +368,172 @@ def _calc_output(self, place, parallel=False, no_check_set=None, loss=None):
...
+                forward_var = block.vars.get(forward_var_name)
+                assert forward_var is not None, "{} cannot be found".format(
+                    forward_var_name)
+                grad_var = grad_block.create_var(
+                    name=arg,
+                    dtype=forward_var.dtype,
+                    shape=forward_var.shape,
+                    type=forward_var.type,
+                    persistable=False)
+                # set persistable for variables without buffer (shape contains 0 like shape = [0,2,5], holder_ is NULL), like XShape
```
sneaxiy: Also, please explain what is `XShape`. No one knows it.

python/paddle/fluid/tests/unittests/test_group_norm_op.py
```diff
@@ -44,6 +44,7 @@ def setUp(self):
         self.shape = (2, 4, 3, 3)
         self.attrs = {'epsilon': 1e-5, 'groups': 2}
         self.compare_between_place = False
+        self.inplace_atol = 1e-5
```
sneaxiy: Why use `self.inplace_atol` without adding a parameter inside `OpTest.check_output()`? Seems a bit strange.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -464,6 +640,20 @@ def find_actual(target_name, fetch_list):
...
                             str(place) + " in dygraph mode")
 
+        if self.inplace_atol is not None:
+            warnings.warn(
+                "By default, inplace_atol should not be set, please check it")
+        self.check_inplace_output_with_place(place, no_check_set=no_check_set)
+
+        # TODO(zhiqiu): enhance inplace_grad test for ops (sum and activation) using mkldnn
+        if (hasattr(self, 'attrs') and self.attrs.get('use_mkldnn') is not None
+            ) or os.getenv('FLAGS_use_mkldnn') is not None:
```
sneaxiy: `FLAGS_use_mkldnn` may be False. Not suitable to check `os.getenv('FLAGS_use_mkldnn') is not None`.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Yes, change `return` to `continue` to solve this, thanks.

paddle/fluid/pybind/pybind.cc
zhiqiu: Done, thanks.

paddle/fluid/pybind/pybind.cc
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Removed, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/test_group_norm_op.py
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/op_test.py
zhiqiu: Exactly, I add a function `get_flags_use_mkldnn` to get `FLAGS_use_mkldnn` in `pybind.cc`, thanks.

sneaxiy (approved):
Excellent! Good job! LGTM for in-place `OpTest` framework!

PR #19117, opened 2019-08-10
Remove book_memory_optimization directory

sneaxiy:
This directory is unnecessary now, since we have deprecated python interface `paddle.fluid.memory_optimize` currently since #18983


luotao1 (approved):
LGTM

PR #19143, opened 2019-08-12
Move python reader api to fluid.io module

sneaxiy:
This PR works for moving all Python reader APIs to `fluid.io` module. Also, this PR  removes APIs inside `paddle/reader/creator.py`, because these APIs are not important and seldomly used.

For example, users can use `fluid.io.batch` instead of `paddle.batch` from now on. For compatibility, users can use `paddle.batch` as well.


lanxianghit (approved):
LGTM

kuke (approved):
LGTM

PR #19155, opened 2019-08-12
Update ngraph engine for multiple threading

baojun-nervana:
This is update ngraph to handle multiple threading safe; prepare to enable parallel executor
* Removed a few static class members which can cause issues for parallelization
* changed backed_ to be per instance
   - used a weak pointer to hold the backend to keep it alive in desired lifetime; otherwise it may run into a case it will be destroyed too early
  - a mutex lock was used to keep it thread safe - this is subject to optimization which we are looking at.

CC. @pawelpiotrowicz @LeoZhao-Intel

### DIFF
--- a/paddle/fluid/operators/ngraph/ngraph_engine.cc
+++ b/paddle/fluid/operators/ngraph/ngraph_engine.cc
@@ -439,14 +439,14 @@ std::shared_ptr<ngraph::Function> NgraphEngine::BuildNgFunction(
   ngraph::ParameterVector func_inputs;
 
   for (auto& vo : var_out_) {
-    PADDLE_ENFORCE(var_node_map_->count(vo),
-                   "Cannot find vo %s in var_node_map_", vo);
+    PADDLE_ENFORCE_GT(var_node_map_->count(vo), 0,
+                      "Cannot find vo %s in var_node_map_", vo);
     func_outputs.emplace_back(var_node_map_->at(vo));
   }
 
   for (auto& vi : var_in_) {
-    PADDLE_ENFORCE(var_node_map_->count(vi),
-                   "Cannot find vi %s in var_node_map_", vi);
+    PADDLE_ENFORCE_GT(var_node_map_->count(vi), 0,
+                      "Cannot find vi %s in var_node_map_", vi);
     std::shared_ptr<ngraph::op::Parameter> prm =
         std::dynamic_pointer_cast<ngraph::op::Parameter>(
             var_in_node_map_->at(vi));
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/ngraph/ngraph_engine.cc
```diff
@@ -232,31 +229,32 @@ NgraphEngine::NgraphEngine(const framework::Scope& scope,
   var_node_map_ = std::make_shared<
       std::unordered_map<std::string, std::shared_ptr<ngraph::Node>>>();
 
+  std::lock_guard<std::mutex> lock(ng_mutex_);
+
+  if (!wp_backend_.lock()) {
+    try {
+      VLOG(3) << "ngraph creating CPU  backend.";
+      backend_ = ngraph::runtime::Backend::create("CPU");
```
LeoZhao-BJ, line 237: here, is it possible to create cpu backend for multiple times and keep multiple instances?
why not keep backend as static and keep one instance? since only ngraph function and executable are useful for next iteration execution, backend is for creating or removing.

original definition is static shared_ptr, there is no explicit releasing, so when it is released early? 

Reason to keep static shared_ptr is on SCR, backend may only have one instance, and device is used exclusively, then it is still possible to support PE, but if use multiple backends, it becomes difficult.

paddle/fluid/operators/ngraph/ngraph_engine.h
```diff
@@ -127,9 +129,7 @@ class NgraphEngine {
 
   void Run(const framework::Scope& scope, const platform::Place& place) const;
 
-  static bool is_training;
-  static const framework::BlockDesc* p_bdesc;
-  static std::vector<std::string> feed_vars, fetch_vars;
+  static std::vector<std::string> feed_vars;
```
LeoZhao-BJ, line 132: why only keep feed vars as static?

paddle/fluid/operators/ngraph/ngraph_engine.cc
pawelpiotrowicz: @LeoZhao-Intel 

> but if use multiple backends, it becomes difficult.

This code has one backend , Baojun sharing reference (single instance) via weak_ptr to others. First execution of this part of code  is creating a backend. If all ng-engine die , local shared_ptr backend_ are destroyed and backend is deleted, when the last one die.

LeoZhao-BJ: if backend changed to non-static, then each NGraphEngine will have its own shared_ptr instance, and NGraphEngine instance is created for each op kernel execution, and use wp_backend to transfer backend btw multiple ngraphengine instances. This is the idea, right ?
```
template <typename DeviceContext, typename T>
class NgraphEngineKernel : public framework::OpKernel<T> {
 public:
  void Compute(const framework::ExecutionContext& ctx) const override {
    auto& scope = ctx.scope();
    auto place = ctx.GetPlace();

    NgraphEngine ngraph_engine(scope, place, ctx);
    ngraph_engine.Run(scope, place);
  }
};
``` 
It is possible in PE that multiple kernel instance running in parallel. I agree to use single backend, and expect to keep single backend not multiple.

LeoZhao-BJ: one corner case, for only one ngraphengine op, whether backend is created and destroy for each iteration?

LeoZhao-BJ: ok, got answer, then it will use cached ctx

paddle/fluid/operators/ngraph/ngraph_engine.h
baojun-nervana: It is unlikely to cause issue, so keep it for now. It can be removed if desired.

pawelpiotrowicz (approved):
LGTM

luotao1 (approved):
LGTM for paddle_enforce

PR #19175, opened 2019-08-13
Smooth label cuda

wangchaochaohu:
https://github.com/PaddlePaddle/benchmark/issues/148
using cuda to realize the label smooth kernel

using transformer-big model to profile get the average perfoemance:
Single OP time cost:3.51607----------->2.39707(ms)


kuke (approved):
LGTM

PR #19176, opened 2019-08-13
fix default value of fleet desc

xjqbest:
* fix default value of fleet desc,  default values are same with jingpai
* print log when save model


yaoxuefeng6 (approved):
LGTM

PR #19177, opened 2019-08-13
Support Tensor input with padding for warpctc op

vincentXiyu:
As described in the title.


vincentXiyu:
屏幕快照 2019-08-13 下午5 28 55

vincentXiyu:
屏幕快照 2019-08-13 下午5 29 03

PR #19181, opened 2019-08-13
Fix compatibility issue of fluid.io.save_vars on windows platform

nepeplwu:
Fix the problem of saving variables that has `/` in name when using fluid.io.save_vars under windows platform [#19180 ](https://github.com/PaddlePaddle/Paddle/issues/19180).

The reason for this problem is that Windows' default file separator is `\`, and fluid.io.save_vars does not normalize the save path. As a result, both `\` and `/` separators appear in the save path.

593AF7049523271945AAD77537268820


phlrain (approved):
LGTM

PR #19187, opened 2019-08-13
modify fast_install.sh about mac installation, test=develop

JepsonWong:
modify checkMacPython2 and checkMacPython3 function to ensure python version used for paddle installation is EQ/GT than 2.7.15 on MacOS; modify checkMacPaddleVersion function to ensure paddle version is release version, because paddle don't have develop version on MacOS.


JiabinYang (approved):
LGTM

PR #19190, opened 2019-08-13
Fix the inference library compilation bug on windows 

zhwesky2010:
Fix the inference library compilation bug on windows. The current bug is that most of files under the inference library path are missed, which is caused by the logic error of the code. Therefore, the code of copying files has been overwritten.


wopeizl (approved):
LGTM

Issue #19110, opened 2019-08-09
Clean up for UT coverage


lidanqing-vv:
MKL-DNN related ops UT states.

After the PR19011 is merged, the states will be
![image](https://user-images.githubusercontent.com/47151829/62776724-b62bab80-baab-11e9-9043-b6a4278cdd35.png)

![image](https://user-images.githubusercontent.com/47151829/62776967-51248580-baac-11e9-8114-b8fe84f7ad47.png)

In which,
1. the `group conv transpose` in `conv_tranpose_mkldnn_op.cc` should be removed because it is not used in any models
2. `in_place` should be removed in `sum_mkldnn_op.cc` because this op is not registered and never used too. (Above 2 clean-ups I need to confirm with @wojtuss  He is on vocation now.)
3. mul_mkldnn_op.cc is hard to cover more because of virtual functions. Improvements should be done first in `mul_mkldnn_op.cc` and then UT will follow.

4. In `elementwise_mul_mkldnn_op.cc`, the condition `if (is_x_format_correct && is_y_format_correct && are_dims_divisable && is_avx512_enabled)` will meet when MKLDNN choose the format by itself, which means one more op like conv should appear before elementwise_mul_mkldnn_op.cc then this condition will be met. I am working on this now.

lidanqing-vv:
PR19191 is last PR for this issue. The conclusion is that for elementwise_mul_mkldnn, coverage is dependent on machines.

PR #19191, opened 2019-08-13
Add elementwise_mul_mkldnn UT with [conv + elt_mul + conv] 

lidanqing-intel:
**Goal is:**
1. to cover the `is_avx512_enabled=true` branch in `elementwise_mul_mkldnn_op.cc`
2. and also not fail on avx2 machines

**What is done:**

Old tests (which is skipped now) hard-coded the format. This new test did not set mkldnn-format and the format will be chosen by itself. The new test made `conv+elementwise_mul+conv` ops block and avoid mkldnn-format settings.

**Current status:**
1. This `is_avx512_enabled=true` branch is covered and the test passed on my own avx512 machine.
2. Expecting it to pass CI avx2 machine. It should be because I did not set any mkldnn-format.

Related issue: #16764  and #19110

### DIFF
--- a/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
+++ b/python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
@@ -84,7 +84,7 @@ class TestElementwiseMulMKLDNNOp_Integrated_With_Convs(ElementwiseMulOp):
             for name in ground_truth:
                 block.create_var(
                     name=name, dtype="float32", shape=ground_truth[name].shape)
-            bn_op = block.append_op(
+            conv2d_op = block.append_op(
                 type="conv2d",
                 inputs={
                     "Input": block.var('input'),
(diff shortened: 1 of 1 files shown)


luotao1:
I test it on AVX2 machine, it runs OK.

python/paddle/fluid/tests/unittests/mkldnn/test_elementwise_mul_mkldnn_op.py
```diff
@@ -19,6 +19,137 @@
...
+            "elt_mul_output": self.elt_mul_output,
+            "conv_output2": self.conv_output2,
+        }
+        program = fluid.Program()
+        with fluid.program_guard(program):
+            block = program.global_block()
+            for name in ground_truth:
+                block.create_var(
+                    name=name, dtype="float32", shape=ground_truth[name].shape)
+            bn_op = block.append_op(
```
wojtuss: Please, rename `bn_op` into `conv2d_op`

PR #19207, opened 2019-08-14
cherry-pick PR#18671 , PR#19190 and PR#19225 into Release/1.5

zhwesky2010:
There are two changes：
1. 报错信息修改为：举例：cudaGetDeviceProperties failed in paddle::platform::GetCUDAComputeCapability!
error code = 1, Please see detail in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g3f51e3575c2178246db0a94a430e0038

2. Fix the inference library compilation bug on windows. The current bug is that most of files under the inference library path are missed. So, on windows platform, we are using a python script to copy filles/folders when build inference lib.

### DIFF
--- a/cmake/copyfile.py
+++ b/cmake/copyfile.py
@@ -1,0 +1,44 @@
+#   Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


liupluswei (dismissed):
LGTM

liupluswei (approved):
LGTM

PR #19213, opened 2019-08-14
Fix BUG: Mask RCNN inference diff When using AnalysisPredictor.

NHZlX:
There are two factors that lead to errors in results.
1. in merge_lod_tensor op

```
if (in_true.IsInitialized()) {
      rank = in_true.dims().size();
      in_dims = framework::slice_ddim(in_true.dims(), 1, rank);
    } else {
      rank = in_false.dims().size();
      in_dims = framework::slice_ddim(in_false.dims(), 1, rank);
    }
}
```
The code enters different logic by determining whether the is_true variable is initialized or not.
In mask rcnn model, if there is object in input samples, the in_true var is always not be initialized. 
If there is no object in input sample, the in_var is initialized, then the object in  following samples will not be detected any more (Different with the process of the inference, the training process will clear all vars after each batch). 

2. when the memory optim func open, there is diff in the result. 
the lod_reset op leads to the problem.


luotao1 (approved):
LGTM for paddle_enforce

Shixiaowei02 (approved):

PR #19218, opened 2019-08-14
modify paddle_build.sh for Paddle python3 image generation, test=develop

JepsonWong:
modify gen_dockerfile function to support Paddle python3 version image generation.


JiabinYang (approved):
LGTM

PR #19240, opened 2019-08-16
Add explanation of capacity to PyReader Document

zhhsplendid:
Add document explanation of capacity to PyReader.


zhhsplendid:
close this PR because Fei Guo has modified it

PR #19243, opened 2019-08-16
Make PADDLE_ENFORCE_EQ support types that cannot be converted to std::string

sneaxiy:
Previously, `PADDLE_ENFORCE_EQ` cannot support types that cannot convert to `std::string`. This PR use C++ SFINAE feature to support all types.

- If the type can be converted to `std::string`, the error message is the same with previous one, such as `x:3 != y:2`.
- If the type cannot be converted to `std::string`, the error message is `x != y`.

### DIFF
--- a/paddle/fluid/platform/enforce.h
+++ b/paddle/fluid/platform/enforce.h
@@ -27,6 +27,7 @@ limitations under the License. */
 #endif  // PADDLE_WITH_CUDA
 
 #include <iomanip>
+#include <iostream>
 #include <memory>
 #include <sstream>
 #include <stdexcept>
@@ -307,7 +308,7 @@ DEFINE_CUDA_STATUS_TYPE(ncclResult_t, ncclSuccess);
   do {                                                                         \
     throw ::paddle::platform::EOFException("There is no next data.", __FILE__, \
                                            __LINE__);                          \
-  } while (false)
+  } while (0)
 
 /*
  * Some enforce helpers here, usage:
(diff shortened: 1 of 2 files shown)


paddle/fluid/platform/enforce.h
```diff
@@ -366,28 +367,72 @@ using CommonType1 = typename std::add_lvalue_reference<
...
+  static NoType Check(...) {
+    return 0;
+  }
+
+ public:
+  static constexpr bool kValue =
+      std::is_same<YesType, decltype(Check<T>(std::cout))>::value;
+};
+
+template <bool kCanToString /* = true */>
```
zhhsplendid, line 393: Two questions:

1. Can we change "struct" to "class"? Use a struct only for passive objects that carry data; everything else is a class. Ref: 
https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes 

2. After we get "constexpr bool __kCanToString__", are we able to use "if" clause without template class and function?

paddle/fluid/platform/enforce_test.cc
```diff
@@ -296,3 +298,60 @@ TEST(enforce, cuda_success) {
...
+TEST(enforce, cannot_to_string_type) {
+  static_assert(
+      !paddle::platform::details::CanToString<CannotToStringType>::kValue,
+      "CannotToStringType must not be converted to string");
+  static_assert(paddle::platform::details::CanToString<int>::kValue,
+                "int can be converted to string");
+  CannotToStringType obj1(3), obj2(4), obj3(3);
+  PADDLE_ENFORCE_NE(obj1, obj2, "Object 1 is not equal to Object 2");
+  PADDLE_ENFORCE_EQ(obj1, obj3, "Object 1 is equal to Object 3");
+  std::string msg = "Object 1 is not equal to Object 2";
```
zhhsplendid: msg at line 326 is same message as the one in line 324. Consider to move line 326 before line 324 and use it in PADDLE_ENFORCE_NE

paddle/fluid/platform/enforce_test.cc
```diff
@@ -296,3 +298,60 @@ TEST(enforce, cuda_success) {
...
+  try {
+    PADDLE_ENFORCE_EQ(obj1, obj2, msg);
+  } catch (paddle::platform::EnforceNotMet& error) {
+    std::string ex_msg = error.what();
+    LOG(INFO) << ex_msg;
+    EXPECT_TRUE(ex_msg.find(msg) != std::string::npos);
+    EXPECT_TRUE(ex_msg.find("obj1 != obj2") != std::string::npos);
```
zhhsplendid: Can we add "Expected obj1 == obj2" and "but received obj1 != obj2" in the check of exception message so that:

1. Code reader will be more clear
2. To prevent future developer from making error (since "expected" case and "but" case both have the two messages)

paddle/fluid/platform/enforce.h
sneaxiy: 1. Done.

2. We can only use template classes. Consider the following codes using "if" clause:
```C++
std::string Convert(const char *expression, const T& value, bool can_to_string) {
    if (can_to_string) {
       // Compilation error if T cannot be converted to std::string!
       return expression + std::string(":") + string::to_string(value);
    } else {
       return expression;
    } 
}
```
The codes above would cause compilation error!

paddle/fluid/platform/enforce_test.cc
sneaxiy: Done.

paddle/fluid/platform/enforce_test.cc
sneaxiy: Done.

paddle/fluid/platform/enforce.h
zhhsplendid: Discussed offline, I meant something like #if macro. Sorry for the confusion. We found a case that we can use #if, but we are afraid of Windows supports different C++ standard ... so we won't change here.

zhhsplendid (commented):
LGTM now, but Sneaxiy would like Tao to review and approve it. So I didn't approve it.

luotao1 (approved):
LGTM！Excellent!

PR #19258, opened 2019-08-16
distribute launch : add use_paddlecloud argument 

danleifeng:
Add use_paddlecloud argument in launch.py. 
Aims to automatically get cluster_ip  infos  while using paddlecloud.


python/paddle/distributed/launch.py
```diff
@@ -140,6 +143,13 @@ def start_procs(args):
     current_node_ip = args.node_ip
     node_ips = [x.strip() for x in args.cluster_node_ips.split(',')]
     node_id = node_ips.index(current_node_ip)
+    if args.use_paddlecloud:
+        # solve PADDLE_TRAINERS=null when using single node.
+        # you can automatically get ip info while using paddlecloud.
+        current_node_ip = os.getenv("POD_IP", "127.0.0.1")
```
gongweibao, line 150: 1. It's used only when running on paddlecloud multi nodes. And
2. these arguments should be checked with arguments from users.

PR #19269, opened 2019-08-18
Unset unittests http_proxy env to avoid timeout.

gongweibao:
If http_proxy is set, the dist unit test will timeout, so unset it.


guru4elephant (approved):
LGTM

PR #19271, opened 2019-08-18
Fix sequence mask in dygraph

kuke:
One fix:

- The wrong assertion that cannot use `sequence_mask` op in dygraph, for this op has nothing to do `LoDTensor`, not like other sequence ops.


kuke:
Revert the change in `dg.DataParallel` because of the potential risks.

chengduoZH (approved):
LGTM

PR #19273, opened 2019-08-19
distribute launch : add use_paddlecloud argument

danleifeng:
Add use_paddlecloud argument in launch.py.
Aims to automatically get cluster_ip infos while using paddlecloud.


gongweibao (approved):
LGTM

PR #19277, opened 2019-08-19
Add use_ngraph option to image classification unit tests

wojtuss:
This patch adds an option to enable NGraph in unit tests for image classification FP32&INT8 inference.

test=develop


wojtuss:
INT8 is not working on nGraph yet. Another solution is needed.

Issue #19275, opened 2019-08-19
elementwise_mul_grad 1.5比1.4慢很多


sneaxiy:
以下代码`use_constant`为True和False时，性能差异很大。
```
import paddle.fluid as fluid
from paddle.fluid.backward import append_backward
import numpy as np
import time

C = 3
batch_size = 32
scale = 224

use_constant = True
print('use_constant {}'.format(use_constant))

shape = [batch_size, C, scale, scale]

place = fluid.CUDAPlace(0)

image = fluid.layers.data(name='image', shape=shape, dtype='float32', append_batch_size=False)
image.stop_gradient = False

w = float(3.0)

if not use_constant:
    w = fluid.layers.create_global_var(
        shape=[scale, scale],
        value=w,
        dtype='float32',
        persistable=True)

out = image * w 

loss = fluid.layers.reduce_mean(out)

append_backward(loss)

for op in fluid.default_main_program().global_block().ops:
    print(op.type)

exe = fluid.Executor(place)
exe.run(fluid.default_startup_program())

compiled_prog = fluid.CompiledProgram(fluid.default_main_program())

in_tensor = fluid.LoDTensor()
in_tensor.set(np.random.random(size=shape).astype('float32'), place)

start_t = time.time()
for _ in range(1000): 
    exe.run(compiled_prog, feed={image.name: in_tensor}, fetch_list=[out])
end_t = time.time()

print('time cost: {}'.format(end_t - start_t))
```

| <br> | Paddle 1.4 | Paddle 1.5|
|---|---|---|
| use_constant = False |10.821|15.4534|
| use_constant = True |12.2862|67.80492|

Related to #19023.

怀疑是[FastElemwiseGradBroadcast1CUDAKernel](https://github.com/PaddlePaddle/Paddle/blob/release/1.5/paddle/fluid/operators/elementwise/elementwise_op_function.h#L360)导致的问题。

sneaxiy:
develop分支代码将[FastElemwiseGradBroadcast1CUDAKernel](https://github.com/PaddlePaddle/Paddle/blob/release/1.5/paddle/fluid/operators/elementwise/elementwise_op_function.h#L360)改为[ElemwiseGradBroadcast1CUDAKernel](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/elementwise/elementwise_op_function.h#L326)后，速度从67.7126891613降为16.8340370655。

[...]

PR #19278, opened 2019-08-19
Fix elementwise performance poor issue

zhaoyuchen2018:
For small case use 1D block is better than 2D block.

Refer to this issue: https://github.com/PaddlePaddle/Paddle/issues/19275


sneaxiy (approved):
LGTM.

PR #19285, opened 2019-08-19
replace part of PADDLE_ASSERT to PADDLE_ENFORCE

luotao1:
In general, it's recommended to use `PADDLE_ENFORCE` for error_mesage. However, in CUDA device kernel or HOSTDEVICE function, we must use  `PADDLE_ASSERT` now.
https://github.com/PaddlePaddle/Paddle/blob/c2063217e7458cfa97a16874b6a6982448259304/paddle/fluid/platform/assert.h#L40
This PR：
- replace part of `PADDLE_ASSERT` to `PADDLE_ENFORCE`, and later we will see how to unify `PADDLE_ASSERT` and `PADDLE_ENFORCE`, which is so confused for usage now.
- remove unused `fallback_alloc_size_` in `system_allocator.cc` and add unit-test for `CUDAPinnedAllocator`, to pass the `PR_CI_COVERAGE`

### DIFF
--- a/paddle/fluid/memory/detail/memory_block.cc
+++ b/paddle/fluid/memory/detail/memory_block.cc
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 #include "paddle/fluid/memory/detail/memory_block.h"
-#include "paddle/fluid/platform/assert.h"
+#include "paddle/fluid/platform/enforce.h"
 
 namespace paddle {
 namespace memory {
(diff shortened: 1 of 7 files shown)


paddle/fluid/memory/detail/memory_block.cc
```diff
@@ -61,7 +61,7 @@ MemoryBlock* MemoryBlock::right_buddy(const MetadataCache& cache) const {
 
 void MemoryBlock::split(MetadataCache* cache, size_t size) {
   // make sure the split fits
-  PADDLE_ASSERT(total_size(*cache) >= size);
```
sneaxiy, line 64: 这些地方可能原本确实是想用`PADDLE_ASSERT`，以减少Release编包时的运行效率？

luotao1: 1. `PADDLE_ENFORCE`本身的运行效率可以忽略不计，`total_size(*cache)`的效率可以等merge后跑benchmark CE看下。
2. 由于assert在release下会被跳过，那这里的报错信息在whl包里还需要么？

PR #19307, opened 2019-08-20
Remove is_mem_optimized in Program

sneaxiy:
`Program._is_mem_optimized` is a very fancy field that is not important to framework itself. This PR removes this field for clean codes.


luotao1 (approved):
LGTM

Issue #18531, opened 2019-07-05
[MKL-DNN] Refactor Acquire API of mkl-dnn integration


jczaja:
It would be good to refactor Acquire API of mkl-dnn:
- Handler should have forward_pd and backward_pd  not like it is now that all inheriting Handler Types are defining its own attributes. 
- Store PD of Fwd and BWD not to rely on PD given by argument of Acquire API
- All AcquireXXXPrimitiveDesc should be called internally by Handler ,as there is no need to have it exposed to op
- descriptors should only be created when they are used eg. when primitives/memory objects are to be created , not reused
- Move Key genereation inside Handler , not to keep it in a body of Compute method

This task should reduce some  redundancy in mkl-dnn acquire api

Issue added to track this task

Sand3r-:
@jczaja is refactoring finished, and can the issue be closed now?

jczaja:
No, It is work in progress

jczaja:
done (several PRs)

PR #19312, opened 2019-08-20
[MKL-DNN] Refactoring Softmax 

jczaja:
This PR is introducing modification to move creation of MKL-DNN memory descriptors inside AcquireAPI rather than have it exposed in Compute() method. This is only for softmax mkldnn op, once we agree on this change all others op will be modified.

@grygielski Please review

### DIFF
--- a/paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
@@ -34,47 +34,61 @@ using platform::to_void_cast;
 
 class SoftmaxMKLDNNHandler : public platform::MKLDNNHandler {
  public:
-  SoftmaxMKLDNNHandler(const platform::MKLDNNDeviceContext& dev_ctx,
+  SoftmaxMKLDNNHandler(const std::vector<int>& dims,
+                       framework::proto::VarType::Type vtype,
+                       const mkldnn::memory::data_type dtype,
+                       const mkldnn::memory::format fmt,
+                       const platform::MKLDNNDeviceContext& dev_ctx,
                        mkldnn::engine engine, const std::string& base_key)
-      : platform::MKLDNNHandler(dev_ctx, engine, base_key) {}
-
-  SoftmaxMKLDNNHandler(
-      std::shared_ptr<mkldnn::softmax_forward::primitive_desc> softmax_pd,
-      std::shared_ptr<mkldnn::softmax_backward::primitive_desc> softmax_bwd_pd,
-      const platform::MKLDNNDeviceContext& dev_ctx, mkldnn::engine engine,
-      const std::string& base_key)
       : platform::MKLDNNHandler(dev_ctx, engine, base_key),
-        softmax_pd_(softmax_pd),
-        softmax_bwd_pd_(softmax_bwd_pd) {
+        dims_(dims),
+        vtype_(vtype),
+        dtype_(dtype),
+        fmt_(fmt) {}
+
+  SoftmaxMKLDNNHandler(const std::vector<int>& dims,
+                       framework::proto::VarType::Type vtype,
+                       const mkldnn::memory::data_type dtype,
+                       const mkldnn::memory::format fmt,
+                       const mkldnn::memory::format diff_fmt,
+                       const platform::MKLDNNDeviceContext& dev_ctx,
+                       mkldnn::engine engine, const std::string& base_key)
+      : platform::MKLDNNHandler(dev_ctx, engine, base_key),
+        dims_(dims),
+        vtype_(vtype),
+        dtype_(dtype),
+        fmt_(fmt),
+        diff_fmt_(diff_fmt) {
     // If we are in Grad operatgor then update a key with BWD suffix to
     // distinguish from FWD memory primitives
+    // Key_common will allow to access FWD_PD from cache
     key_ += "-BWD";
   }
 
-  std::shared_ptr<softmax_forward::primitive_desc>
-  AcquireSoftmaxPrimitiveDescriptor(const softmax_forward::desc& softmax_desc,
-                                    const mkldnn::engine& engine) {
-    // Softmax PD has to be passed to Grad op that
-    // may be executed by diffrent thread, hence
-    // for that one we use key that does not contain TID
-    const std::string key_softmax_pd = key_common_ + "@softmax_pd";
+  // TODO(jczaja): Once fwd_pd_ are moved to MKLDNNHandler then this function
+  // should be moved as well eg. SoftmaxMKLDNNHandler -> MKLDNNHandler<softmax_>
+  std::shared_ptr<mkldnn::memory> AcquireSrcMemory(void* ptr) {
+    return this->AcquireMemory(dims_, dtype_, fmt_, ptr, "@user_src_mem_p");
+  }
 
-    softmax_pd_ = std::static_pointer_cast<softmax_forward::primitive_desc>(
-        dev_ctx_.GetBlob(key_softmax_pd));
-    if (softmax_pd_ == nullptr) {
-      static std::mutex acquire_barrier;
-      std::lock_guard<std::mutex> block_threads_until_finish_this_job(
-          acquire_barrier);
-      softmax_pd_ = std::static_pointer_cast<softmax_forward::primitive_desc>(
-          dev_ctx_.GetBlob(key_softmax_pd));
-      if (softmax_pd_ == nullptr) {
-        softmax_pd_.reset(
-            new softmax_forward::primitive_desc(softmax_desc, engine));
-        dev_ctx_.SetBlob(key_softmax_pd, softmax_pd_);
-      }
-    }
+  std::shared_ptr<mkldnn::memory> AcquireDstMemory(void* ptr) {
+    return this->AcquireMemory(dims_, dtype_, fmt_, ptr, "@user_dst_mem_p");
+  }
 
-    return softmax_pd_;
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
```diff
@@ -34,47 +34,61 @@ using platform::to_void_cast;
 
 class SoftmaxMKLDNNHandler : public platform::MKLDNNHandler {
  public:
-  SoftmaxMKLDNNHandler(const platform::MKLDNNDeviceContext& dev_ctx,
+  SoftmaxMKLDNNHandler(const std::vector<int>& dims,
+                       framework::proto::VarType::Type vtype,
+                       const mkldnn::memory::data_type dtype,
```
Sand3r-: As I see you save variable and data type for usage of memory allocation with mutable_data and for memory descriptor creation in MKL-DNN. However mutable_data allows providing a template type as an argument to initialize tensor. For MKL-DNN you can use `platform::MKLDNNGetDataType<T>()` to convert a type to MKL-DNN's data type. Please consider templatizing this handler to handle those scenarios in this way.

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
```diff
@@ -34,47 +34,61 @@ using platform::to_void_cast;
...
+                       const mkldnn::memory::format fmt,
+                       const platform::MKLDNNDeviceContext& dev_ctx,
+                       mkldnn::engine engine, const std::string& base_key)
+      : platform::MKLDNNHandler(dev_ctx, engine, base_key),
+        dims_(dims),
```
Sand3r-, line 43: Perhaps something as common as dimensions and formats could be moved to the generic platform::MKLDNNHandler class. Such action would require more forethought though. Since some operators require several inputs, perhaps they could be stored in two vectors, or one vector containing dims, format pairs.

Sand3r- (commented):
@jczaja thank you so much for making effort to unify the handlers and prepare them for MKL-DNN 1.0 upgrade. Please refer to the comments attached.

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
```diff
@@ -34,47 +34,61 @@ using platform::to_void_cast;
...
-                                    const mkldnn::engine& engine) {
-    // Softmax PD has to be passed to Grad op that
-    // may be executed by diffrent thread, hence
-    // for that one we use key that does not contain TID
-    const std::string key_softmax_pd = key_common_ + "@softmax_pd";
+  // TODO(jczaja): Once fwd_pd_ are moved to MKLDNNHandler then this function
+  // should be moved as well eg. SoftmaxMKLDNNHandler -> MKLDNNHandler<softmax_>
+  std::shared_ptr<mkldnn::memory> AcquireSrcMemory(void* ptr) {
+    return this->AcquireMemory(dims_, dtype_, fmt_, ptr, "@user_src_mem_p");
+  }
```
grygielski, line 66: I think we should get memory descriptors from mkldnn::query functions invoked on primitive descriptions instead of creating new memory descriptors with dims_, dtype_ and fmt_.

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
```diff
@@ -34,47 +34,61 @@ using platform::to_void_cast;
...
+  std::shared_ptr<mkldnn::memory> AcquireDiffSrcMemory(void* ptr) {
+    return this->AcquireMemory(dims_, dtype_, diff_fmt_, ptr,
+                               "@user_diff_src_mem_p");
+  }
 
-    return softmax_pd_;
+  std::shared_ptr<mkldnn::memory> AcquireDstMemoryFromPrimitive(void* ptr) {
+    this->AcquireSoftmaxPrimitiveDescriptor();
+    return this->AcquireMemoryFromPrimitive(fwd_pd_->dst_primitive_desc(), ptr,
+                                            "@dst_mem_p");
```
grygielski, line 86: Can't we move AcquireSoftmaxPrimitiveDescriptor to constructor so we won't have to store unnecessary data like dims_, dtype_and so on?

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
```diff
@@ -154,21 +211,15 @@ class SoftmaxMKLDNNKernel : public paddle::framework::OpKernel<T> {
...
-        {softmax_tz}, platform::MKLDNNGetDataType<T>(), memory::format::nc);
-    // Normalization is made after innermost dimension eg. C out of NC
-    auto softmax_desc = softmax_forward::desc(prop_kind::forward_scoring,
-                                              softmax_md, 1 /*dim: C*/);
-
-    auto softmax_pd =
-        handler.AcquireSoftmaxPrimitiveDescriptor(softmax_desc, mkldnn_engine);
+    SoftmaxMKLDNNHandler handler(
+        softmax_tz, input->type(), platform::MKLDNNGetDataType<T>(),
+        mkldnn::memory::format::nc, dev_ctx, mkldnn_engine, key);
```
grygielski: I think we should consider hiding key creation inside AcquireAPI too.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -119,6 +119,25 @@ class MKLDNNHandler {
...
+
+      mem_p = std::make_shared<mkldnn::memory>(
+          mkldnn::memory::primitive_desc{md, engine_}, ptr);
+      dev_ctx_.SetBlob(local_key, mem_p);
+    } else {
+      mem_p->set_data_handle(ptr);
+    }
+    return mem_p;
+  }
+
```
grygielski, line 140: As said above, I think we shouldn't create memory descriptor from parameters like that: `auto md = mkldnn::memory::desc(dims, dtype, fmt);` but use `fwd_pd_->src_desc()` or others instead. This will be possible once we create (or acquire from GetBlob) primitive descriptor in constructor or just before creating memory for primitives.

grygielski (commented):
@jczaja  Thanks for your contribution into unifying AcquireAPI and moving functionality inside class functions so it is easier to use and develop. However, I have some suggestions so we can make it even better, especially in perspective of mkldnn 1.0 update.

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
jczaja: In this code we create Memory object for Tensor data as it comes to MKL-DNN op. This not necessery is the same layout/format as indicated by computational primitive 's PD .

grygielski: I understand that user_src_md is not always the same as src_md but in this case you create softmax primitive descriptor this way:
`auto md = mkldnn::memory::desc(dims_, dtype_, fmt_);
auto softmax_desc = softmax_forward::desc(prop_kind::forward_scoring, md, 1 /*dim: C*/);` 
So mkldnn::query such as `fwd_pd_->src_desc()` = `mkldnn::memory::desc(dims_, dtype_, fmt_)`

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
jczaja: That is interesting idea, but still we need dims_, dtype_ etc. for calling Acquire{Src|Dst..}Memory. I'm not planning to do modification you suggest on this PR, but I hope we can disscuss design offline and perhaps do some changes in next PRs.

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
jczaja: That is actually very good idea to be implemented. I have added it to the list in relevant refactoring issue #18531 so it can be done later on.

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
grygielski: Ok, sounds good to me!

paddle/fluid/platform/mkldnn_reuse.h
jczaja: To certain degree this can be done . I will modify this PR to replace AcquireDstMemory with AcquireDstMemoryFromPrimitive like here: https://github.com/PaddlePaddle/Paddle/blob/85914f7a88f748a3883bfd09df810e910cce3089/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc#L120-L121

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
jczaja: Interesting thought. I hope direction whether to do it or now will become clear once 
we move further with refactoring.

paddle/fluid/platform/mkldnn_reuse.h
jczaja: On second thought. Some more work is to be done here, so I created separated issue for that #19566

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
jczaja: done

grygielski (approved):
We agreed to discuss my suggestions offline and implement them in further PRs so
LGTM

Sand3r- (approved):
@jczaja thank you so much for incorporating proposed changes. Looks Good To Me.
Please ensure that the CIs pass.

PR #19318, opened 2019-08-21
change get hash key inputName to tensor pointer; test=develop

phlrain:
### 背景
在动态图模式下，OperatorBase在每一个op中，都会被构造和析构，OperatorBase中保存了两个成员变量，VariableNameMap inputs_, outputs_, 这两个成员变量的构造和析构花费了很多时间，大约有30%的c++端的开销。

在后续版本中，input_, output_ 成员变量以及相关的函数 Input(), Output()会被移除。

这个PR改动了MKLDNN相关的部分，想请Intel的人看下改动是否可以。

### Background:
In dygraph mode, `OperatorBase` will be constructed and destructed in each Op. About 30% time cost on the construction and destruction of `VariableNameMap inputs_`  and `VariableNameMap ouputs_`.

Thus, `Input()`/`Output()` interface, and `input_`/`output_` variable will be removed from OperatorBase later.

This PR modify the MKLDNN related codes, @jczaja @jianhang-liu Could you please review it?

### DIFF
--- a/paddle/fluid/operators/mkldnn/lrn_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/lrn_mkldnn_op.cc
@@ -63,7 +63,8 @@ class LRNMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
         dims, platform::MKLDNNGetDataType<T>(), x->format());
 
     const std::string key = platform::LRNMKLDNNHandler::GetHash(
-        dims, n, alpha, beta, k, x->format(), ctx.op().Output("Out"));
+        dims, n, alpha, beta, k, x->format(),
+        std::to_string(*reinterpret_cast<int*> out));
 
     platform::LRNMKLDNNHandler handler(ctx.Attr<bool>("is_test"), dev_ctx,
                                        mkldnn_engine, key);
@@ -121,8 +122,10 @@ class LRNMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
 
     auto dims = paddle::framework::vectorize2int(x->dims());
 
+    auto orig_out = ctx.Input<Tensor>("Out");
     const std::string key = platform::LRNMKLDNNHandler::GetHash(
-        dims, n, alpha, beta, k, x->format(), ctx.op().Input("Out"));
+        dims, n, alpha, beta, k, x->format(),
+        std::to_string(*reinterpret_cast<int*> orig_out));
 
     platform::LRNMKLDNNHandler handler(false, dev_ctx, mkldnn_engine, key);
 


luotao1:
@LeoZhao-Intel Could you please review this PR~

LeoZhao-BJ:
This change may not work for mkldnn, since out pointer may always be different for each iteration. I remember we disabled transferscope cache for cpu with PR #18578, it means each iteration, the out tensor may be re-allocated and pointer may change, so it can't match hash key initial goal, that's sharing and finding.

luotao1:
@LeoZhao-Intel Could you give some suggestion if this PR doesn't meet mkldnn demands?

LeoZhao-BJ:
If it is just for inference, then it is easy, even without output, it is still ok, but for training, it may be used to find relationship btw fwd and bwd, as Jacek raised issue before, we may need to find a new way to connect fwd and bwd.

paddle/fluid/operators/mkldnn/lrn_mkldnn_op.cc
```diff
@@ -63,7 +63,8 @@ class LRNMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
         dims, platform::MKLDNNGetDataType<T>(), x->format());
 
     const std::string key = platform::LRNMKLDNNHandler::GetHash(
-        dims, n, alpha, beta, k, x->format(), ctx.op().Output("Out"));
+        dims, n, alpha, beta, k, x->format(),
+        std::to_string(*reinterpret_cast<int*> out));
```
jczaja, line 67: Added code is basically extracting first four bytes(integer) of Tensor object . First four bytes in Tensor for MKL-DNN is:
https://github.com/PaddlePaddle/Paddle/blob/561232c25a0af9fb408f086faf6f443fa0140d1c/paddle/fluid/framework/tensor.h#L57

Which describe what mkl-dnn format is used for data arrangement in given Tensor. In most cases it will be value of 40 (nchw8c) or 41 (nchw16c). Using this value does not help as it will be almost always value of 40 or 41. 

I understand that this change is more like to show some idea

jczaja:
As @LeoZhao-Intel mentioned there may be problem with having training working eg. FWD and BWD. It happens that for multi-threaded execution one thread may execute forward and other thread may execute backward and still forward need to pass some data to backward. if pointer of tensor is used then address may be different in FWD of Tensor X and Grad of Tensor X as they are executed by different threads (to be confirmed).

We worked on different way of passing data from FWD op to Grad op. For MKL-DNN activations
to pass data from FWD op to Grad op we make X to be available for activation grad op:
https://github.com/PaddlePaddle/Paddle/blob/561232c25a0af9fb408f086faf6f443fa0140d1c/paddle/fluid/operators/activation_op.cc#L75-L80
As data needed in mkl-dnn Grad op can be taken from X. That way we do not have to use cache for transfering data from fwd to grad ops (still be need cache for caching data to be used in following iterations). To have this working for LRN and all ops we need to extend this idea with
creating Tensors to be used only by MKL-DNN ops for communication. In other words
when mkl-dnn execution does happen (listing above) we  create Tensor that will serve as a way of transfering data from FWD to Grad. Data we need to transfer are: mkl-dnn primitive descriptor and mkl-dnn workspace. Please suggest how to do it.

paddle/fluid/operators/mkldnn/lrn_mkldnn_op.cc
luotao1: > I understand that this change is more like to show some idea

Yes, it is only to show that we will remove the `Input` and `Output` interface later.

LeoZhao-BJ:
I am not sure if it can work if we add op name into hash? e.g. fc12.

Generally speaking, in original design, this kind of mkldnn primitive can be reused among multiple instances of op same with dimension, e.g. in one graph, it has 3 fc ops with same in/out size, then this primitives can be reused for these 3 fc ops execution.

New approach is by adding op name to hash, each op will have its primitive despite of same size, it will consume more memory, but the good thing is it may work well with fwd/bwd since:

1. For bwd op, the output of fwd op is the input of bwd op, then we can find fwd op from input, and from fwd op name, we can find mkldnn primitive. 
2. We can ignore dimension, since in current implementation, there is new cache index for shape, see https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/platform/device_context.cc#L411

Just a draft idea, for your reference.

PR #19332, opened 2019-08-21
 Remove constraint that last dimension is forced to be 1 in huber_loss op

Aurelius84:
**before：**
+ the input(x) and input(y) must be 2-D tensor with shape [batch_size, 1]
```python
import paddle.fluid as fluid

x = fluid.layers.data(name='x', shape=[13], dtype='float32')
predict = fluid.layers.fc(input=x, size=1)
label = fluid.layers.data(name='label', shape=[1], dtype='float32')
loss = fluid.layers.huber_loss(input=predict, label=label, delta=1.0)

place = fluid.CPUPlace()
x_data = np.random.rand(10, 13).astype("float32")
label_data = np.random.random(size=(10, 1)).astype('float32')
exe = fluid.Executor(place)
ret = exe.run(feed={'x': x_data, 'label': label_data}, 
              fetch_list=[loss], return_numpy=False)
```

**after：**
+ the input(x) and input(y) can be tensor with shape [batch_size, 1] or [batch_size]
+ API interface call by `fluid.loss.huber_loss` instead of `fluid.layers.huber_loss`
```python
batch_size = 10
label = fluid.layers.data(name='label', shape=[batch_size],append_batch_size=False, dtype='float32')
...
loss = fluid.loss.huber_loss(input=predict, label=label, delta=1.0)
...
label_data = np.random.random(size=(10)).astype('float32')
```

### DIFF
--- a/paddle/fluid/operators/loss/huber_loss_v2_op.cc
+++ b/paddle/fluid/operators/loss/huber_loss_v2_op.cc
@@ -1,0 +1,149 @@
+/* Copyright (c) 2016 PaddlePaddle Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */
+
+#include "paddle/fluid/operators/loss/huber_loss_v2_op.h"
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace paddle {
+namespace operators {
+
+class HuberLossV2Op : public framework::OperatorWithKernel {
+ public:
+  using framework::OperatorWithKernel::OperatorWithKernel;
+
+  void InferShape(framework::InferShapeContext* ctx) const override {
+    PADDLE_ENFORCE(ctx->HasInput("X"), "Input(X) must be initialized.");
+    PADDLE_ENFORCE(ctx->HasInput("Y"), "Input(Y) must be initialized.");
+
+    auto x_dims = ctx->GetInputDim("X");
+    auto y_dims = ctx->GetInputDim("Y");
+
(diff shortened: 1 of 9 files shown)


paddle/fluid/operators/loss/huber_loss_v2_op.cc
```diff
@@ -0,0 +1,148 @@
...
+                      "Input(Y) must be initialized.");
+
+    auto x_dims = ctx->GetInputDim("X");
+    auto y_dims = ctx->GetInputDim("Y");
+
+    PADDLE_ENFORCE_LE(x_dims.size(), 2,
+                      "The rank of Input(X) should less than or equal to 2.");
+    PADDLE_ENFORCE_LE(y_dims.size(), 2,
+                      "The rank of Input(Y) should less than or equal to 2.");
+    if (ctx->IsRuntime()) {
```
phlrain, line 40: PADDLE_ENFORCE_EQ(x_dims, y_dims, "Shape of X and Y should be same");
product 一致，有可能shape不一致

paddle/fluid/operators/loss/huber_loss_v2_op.cc
```diff
@@ -0,0 +1,148 @@
...
+  using framework::SingleGradOpDescMaker::SingleGradOpDescMaker;
+
+ protected:
+  std::unique_ptr<framework::OpDesc> Apply() const override {
+    std::unique_ptr<framework::OpDesc> op(new framework::OpDesc());
+    op->SetType("huber_loss_v2_grad");
+    op->SetInput("Residual", Output("Residual"));
+    op->SetInput(framework::GradVarName("Out"), OutputGrad("Out"));
+    op->SetOutput(framework::GradVarName("X"), InputGrad("X"));
+    op->SetOutput(framework::GradVarName("Y"), InputGrad("Y"));
```
phlrain, line 127: Y 好像是没有梯度回传的

paddle/fluid/operators/loss/huber_loss_v2_op.cc
Aurelius84: ok，可以添加一个检查x.dim[0] == y.dim[0]

paddle/fluid/operators/loss/huber_loss_v2_op.cc
Aurelius84: y不一定是label吧，也可能是其他的吧

PR #19346, opened 2019-08-22
A a pass to enable the use of cudnn

Xreki:
In some pre-saved inference model, the attr `use_cudnn` is set to `false`, but using cudnn may get better performance, so add an interface to enable this.


luotao1 (approved):
LGTM！！

PR #19349, opened 2019-08-22
Fix bug of getting bool Flags from os.environ

zhiqiu:
This PR fix a bug of getting bool Flags from os.environ. For example, original code use 
`use_ngraph = bool(os.getenv("FLAGS_use_ngraph", False))` 
`use_mkldnn = bool(os.getenv("FLAGS_use_mkldnn", False))`
to get bool Flags from os.environ, but when user set `export FLAGS_use_mkldnn=False`, the code becomes `use_mkldnn = bool('False')` which returns True, and it is wrong.

Since paddle use `gflags` to handle commandline flags in C++, to keep the consistency of flags processing logic between Python and C++, this PR create methods in `pybind.cc` to get the `Flags_xx` processed by C++.


sneaxiy (approved):
LGTM.

PR #19368, opened 2019-08-22
Add optimizer state io

DogHeadMoon:
Add save/load state dict to make optimizers to save/load their states
Add learning_rate() to access learning rate value without waiting
Add a unit test file to test save/load the optimizers' states


iclementine (dismissed):
LGTM！

PR #19372, opened 2019-08-23
Update MLU batch size inputs supports.

Shixiaowei02:
We need to update these codes because the API update of Anakin.

test=develop


NHZlX (approved):
LGTM

PR #19376, opened 2019-08-23
Fix arg do_model_average in param_attr

kuke:
Before fix, users cannot set the status of `do_model_average` because it is not passed to the parameter correctly.


wanghaoshuang (dismissed):
LGTM

wanghaoshuang (approved):

guoshengCS (approved):

PR #19379, opened 2019-08-23
fix multi-thread exec of trt, test=release/1.5

Shixiaowei02:
[Cherry-pick of commit e1eca83, test=release/1.5]

The problem is only triggered by TensorRT and Clone(), and the user feedbacks that using version 1.4 can avoid this problem. There are two reasons for the final confirmation:

TensorRT uses the uniqueness of the molecular graph of the predictor_id area. Predictor::Clone() will change the predictor_id value, but the logic of TensorRT subgraph passes the predictor_id by value when Pass, and the Clone() does not re-trigger Pass, so the uniqueness fails.
Prior to commit ae576f3, the Deserialize() call to the TensorRT subgraph occurred at tensorrt_engine_op.h, once for each cloned-predictor. After this commit, the call occurred at tensorrt_subgraph_pass.cc, only once in ir_pass of master-predictor.
This pull request fixed the bug.

### DIFF
--- a/paddle/fluid/inference/tensorrt/engine.cc
+++ b/paddle/fluid/inference/tensorrt/engine.cc
@@ -35,8 +35,15 @@ void TensorRTEngine::Build(const DescType &paddle_model) {
 void TensorRTEngine::Execute(int batch_size, std::vector<void *> *buffers,
                              cudaStream_t stream) {
   freshDeviceId();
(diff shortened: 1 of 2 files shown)


NHZlX (dismissed):
LGTM

lijianshe02 (approved):

PR #19383, opened 2019-08-23
Fix a shape bug in anakin-nvgpu, test=release/1.5

Shixiaowei02:
Fixed a size error in a call in the anakin interface that could cause the prediction to fail.

test=release/1.5


NHZlX (approved):
LGTM

PR #19385, opened 2019-08-23
Add the support of padding_idx attribute

yihuaxu:
As PyramidDNN model's requirement, it need the support of padding_idx attribute. Now implement it.


yihuaxu:
it has been already merged in PR#193333

PR #18823, opened 2019-07-25
Remove recordio interfaces

sneaxiy:
RecordIO interfaces are not used and maintained any more. This PR removes them.

`read_file` and `double_buffer` use legacy recordio APIs as examples. This PR changes them to py_reader.

24231b38002ff6d9dfed7e3cf7221adc

e68e7c6c8f912e437b7879046764d07c

PR #19387, opened 2019-08-23
Remove recordio convert in dataset

sneaxiy:
RecordIO is not supported since #18823. This PR removes the rest of recordio interfaces inside dataset.


chengduoZH (approved):
LGTM

PR #19397, opened 2019-08-23
Fix train error when test_program.clone is executed after optimizer.minimize

chenwhql:
Because the `optimizer.minimize` will append backward part for the main_program, when the `test_program.clone` is executed after `optimizer.minimize`, the backward part will also be cloned into test_program, which would make the test data also participate train process, thus bring the illusion of high accuracy. So we recommend users to place the `test_program.clone` before `optimizer.minimize`.

But if users don't know this rule, it's easy to make mistakes. In order to solve this program, I try to prune the backward part of network when the `test_program.clone` is executed after `optimizer.minimize`, the result is just used for test, but not for inference. 

If users place the `test_program.clone` before `optimizer.minimize`, no new operations were introduce.

image


python/paddle/fluid/framework.py
```diff
@@ -3192,7 +3192,13 @@ def network(is_test):
         The two code snippets above will generate and print same programs.
         """
         if for_test:
-            p = self._inference_optimize(prune_read_op=False)
+            if self._appending_grad_times > 0:
```
chenwhql, line 3237: in order to check whether the program contains backward part, the flag `_appending_grad_times` can be used to judge here, but it also can judge by other way, like

```
def _has_backward_op(self):
    op_role_key = core.op_proto_and_checker_maker.kOpRoleAttrName()
    backward = int(core.op_proto_and_checker_maker.OpRole.Backward)
    for op in reversed(self.global_block().ops):
        if int(op.all_attrs()[op_role_key]) == backward
            return true
    return false
```

which is better?

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -0,0 +1,67 @@
...
+            loss_data_in_orig, = exe.run(test_program_orig,
+                                         feed={'X': input_x,
+                                               'Y': input_y},
+                                         fetch_list=[avg_loss.name])
+            loss_data_in_prune, = exe.run(test_program_prune,
+                                          feed={'X': input_x,
+                                                'Y': input_y},
+                                          fetch_list=[avg_loss.name])
+
+            self.assertEqual(loss_data_in_orig, loss_data_in_prune)
```
chenwhql: Is there a better way to determine whether two programs are same? Because the order of vars and attrs of ops will be changed by minimize, the order of vars and op attrs in pruned program isn't same as the origin forward program,  the string comparing cannot be used here directly.

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -0,0 +1,99 @@
...
+            place = core.CPUPlace()
+            exe = fluid.Executor(place)
+            exe.run(fluid.default_startup_program())
+
+            loss_data_orig, = exe.run(test_prog_orig,
+                                      feed=feed_dict,
+                                      fetch_list=[loss.name])
+            loss_data_prune, = exe.run(test_prog_prune,
+                                       feed=feed_dict,
+                                       fetch_list=[loss.name])
```
chengduoZH: You should check whether the operators of `test_prog_orig ` and `test_prog_prune ` are the same.

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -0,0 +1,99 @@
...
+
+class TestProgramPruneBackward(unittest.TestCase):
+    def check_prune_correctness(self, method, feed_dict, optimizer):
+        with self.program_scope_guard():
+            loss = method(use_feed=False)
+
+            main_program = fluid.default_main_program()
+            test_prog_orig = main_program.clone(for_test=True)
+            optimizer(learning_rate=0.001).minimize(loss)
+            test_prog_prune = main_program.clone(for_test=True)
```
chengduoZH: Maybe we should make the model more complicated, for example, add regularization ops.

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -0,0 +1,138 @@
...
+            block_a = program_a.blocks[idx]
+            block_b = program_b.blocks[idx]
+            self.assertEqual(len(block_a.ops), len(block_b.ops))
+            # self.assertEqual(len(block_a.vars), len(block_b.vars))
+            for op_idx in range(len(block_a.ops)):
+                self.assertEqual(block_a.ops[op_idx].type,
+                                 block_b.ops[op_idx].type)
+            for var_key in list(block_a.vars.keys()):
+                if block_a.vars[var_key].type == core.VarDesc.VarType.RAW:
+                    continue
```
chenwhql: in the `resnet`, there are three vars that are lost after `prune`.
```
name: "kCUDNNBwdFilterAlgoCache"
type {
  type: RAW
}
persistable: true

name: "kCUDNNBwdDataAlgoCache"
type {
  type: RAW
}
persistable: true

name: "kCUDNNFwdAlgoCache"
type {
  type: RAW
}
persistable: true
```
These vars are independent in scope, not associated with Op.

related PR: [#14286](https://github.com/PaddlePaddle/Paddle/pull/14286)

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
chenwhql: Done

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
chenwhql: Done

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -0,0 +1,190 @@
...
+    lstm_max_tanh = fluid.layers.tanh(lstm_max)
+    fc1 = fluid.layers.fc(input=lstm_max_tanh, size=hid_dim2, act='tanh')
+    prediction = fluid.layers.fc(input=fc1, size=class_dim, act='softmax')
+    cost = fluid.layers.cross_entropy(input=prediction, label=label)
+    avg_cost = fluid.layers.mean(x=cost)
+    return avg_cost
+
+
+class TestProgramPruneBackward(unittest.TestCase):
+    def program_compare(self, program_a, program_b):
```
chenwhql, line 56: Is there a better way to determine whether two programs are same? Because the order of vars and attrs of ops will be changed by minimize, the order of vars and op attrs in pruned program isn't same as the origin forward program, the string comparing cannot be used here directly.

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -0,0 +1,190 @@
...
+        exe = fluid.Executor(place)
+        exe.run(fluid.default_startup_program())
+
+        loss_data_orig, = exe.run(test_prog_orig,
+                                  feed=feed_dict,
+                                  fetch_list=[loss.name])
+        loss_data_prune, = exe.run(test_prog_prune,
+                                   feed=feed_dict,
+                                   fetch_list=[loss.name])
+        self.assertEqual(loss_data_orig, loss_data_prune)
```
chengduoZH, line 95: You should run `test_prog_prune` first here.
Because, if the pruning function has a bug, which induces to the `test_prog_prune` has backward ops and optimization ops, the parameters of this model may be changed.

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def _init_fc_data(self):
+        np.random.seed(5)
+        img = np.random.random(size=[32, 784]).astype(np.float32)
+        label = np.ones(shape=[32, 1], dtype='int64')
+        return img, label
```
chengduoZH: You can use `init_data` directly, which is in `simple_net.py`.

python/paddle/fluid/framework.py
chengduoZH: I am not sure, maybe both ok here.

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
chenwhql: Done

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
chenwhql: Done

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
chenwhql: these three variables are useless now, delete them in PR [#19644 ](https://github.com/PaddlePaddle/Paddle/pull/19644)

chengduoZH (approved):
LGTM

lanxianghit:
描述配图中有type：‘foreword’ -> 'forward'; 'backeward' -> 'backward'

Superjomn (approved):
LGTM

chenwhql:
> 描述配图中有type：‘foreword’ -> 'forward'; 'backeward' -> 'backward'

已修改

lanxianghit (approved):

PR #19405, opened 2019-08-25
Enable the arg trainable in static graph mode

kuke:
Current problem is that the grad and optimizer's params would still be created even though the parameter is not trainable, which is not efficient in memory and computation.


kuke:
Close this pull request because of https://github.com/PaddlePaddle/Paddle/pull/19410

PR #19408, opened 2019-08-26
Disable GC in test_parallel_exe_seresnext

chengduoZH:
test_parallel_exe_seresnext randomly failed in CI system, I guess that this it is because we open GC default.


sneaxiy (approved):
LGTM.

PR #19409, opened 2019-08-26
Add retry_allocator for gpu

sneaxiy:
This PR adds `RetryAllocator` for CUDA allocator. This is designed for fault tolerance if there are multiple threads want to allocate GPU memory in the same time.

See [CE Debug](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=56975&tab=buildResultsDiv&buildTypeId=PaddleModesl_ModelDebug_Model11Build) here. Although `model_image_classification` fails in `test_acc_top5_card4`, I found that it is the original error of CE Debug, since there is [another task](http://ce.paddlepaddle.org:8080/viewLog.html?buildId=56696&buildTypeId=PaddleModesl_BuildModels&tab=buildResultsDiv) also failing.


paddle/fluid/memory/allocation/allocator_facade.cc
```diff
@@ -37,7 +37,7 @@
 #endif
 
 DEFINE_int64(
-    gpu_allocator_retry_time, 0,
+    gpu_allocator_retry_time, 100,
```
chengduoZH, line 40: Why do you set 100 here?

paddle/fluid/memory/allocation/cuda_allocator.cc
```diff
@@ -36,7 +37,7 @@ Allocation* CUDAAllocator::AllocateImpl(size_t size) {
   void* ptr;
   auto status = cudaMalloc(&ptr, size);
   if (UNLIKELY(status != cudaSuccess)) {
-    throw BadAlloc(string::Sprintf(
+    PADDLE_THROW_BAD_ALLOC(string::Sprintf(
```
chengduoZH: Why do you use `string::Sprintf` here?

paddle/fluid/memory/allocation/allocator_facade.cc
sneaxiy: I am not sure what this value should be. TensorFlow sets it to be 10 seconds, but I think it is too large...

paddle/fluid/memory/allocation/cuda_allocator.cc
sneaxiy: `string::Sprintf` can be removed here. Done.

paddle/fluid/memory/detail/system_allocator.cc
```diff
@@ -118,19 +119,19 @@ void* GPUAllocator::Alloc(size_t* index, size_t size) {
...
-                 << FLAGS_initial_gpu_memory_in_mb
-                 << ". Current FLAGS_reallocate_gpu_memory_in_mb value is "
-                 << FLAGS_reallocate_gpu_memory_in_mb;
-    return nullptr;
+    PADDLE_THROW_BAD_ALLOC(
+        "Cannot malloc " + std::to_string(size / 1024.0 / 1024.0) +
+        " MB GPU memory. Please shrink "
+        "FLAGS_fraction_of_gpu_memory_to_use or "
+        "FLAGS_initial_gpu_memory_in_mb or "
+        "FLAGS_reallocate_gpu_memory_in_mb"
```
zhhsplendid, line 126: No space here, "in_mb" will concat with "environment" next line

paddle/fluid/platform/enforce.h
```diff
@@ -66,25 +66,59 @@ inline std::string demangle(std::string name) {
 inline std::string demangle(std::string name) { return name; }
 #endif
 
+template <typename StrType>
+inline std::string GetTraceBackString(StrType&& what, const char* f, int l) {
```
zhhsplendid: "f", "l" is not good parameter names. In my understanding, you print trace at [file:line] Could you change "f" to "file" or "file_name", "l" to "line" or "line_number" or something like that.

paddle/fluid/memory/allocation/allocator_facade.cc
```diff
@@ -145,6 +153,25 @@ class AllocatorFacadePrivate {
...
+      PADDLE_ENFORCE_EQ(pair.second->IsAllocThreadSafe(), true);
+    }
+  }
+
+  void WrapCUDARetryAllocator(size_t retry_time) {
+    PADDLE_ENFORCE_GT(retry_time, 0, "Retry time must be larger than 0");
+    for (auto& pair : allocators_) {
+      if (!platform::is_gpu_place(pair.first)) continue;
+
+      pair.second = std::make_shared<RetryAllocator>(pair.second, retry_time);
```
zhhsplendid: Make it shorter without "continue" logic:

if (platform::is_gpu_place(pair.first)) {
    pair.second = std::make_shared<RetryAllocator>(pair.second, retry_time);
}

paddle/fluid/memory/allocation/retry_allocator.cc
```diff
@@ -31,29 +53,29 @@ Allocation* RetryAllocator::AllocateImpl(size_t size) {
   // But it would add lock even when allocation success at the first time
   try {
     return alloc_func();
-  } catch (BadAlloc& bad_alloc) {
+  } catch (BadAlloc&) {
```
zhhsplendid, line 60: Consider adding some logs when catching BadAlloc. Same as while loop beneath. It may be easier to debug in the future.

chengduoZH (dismissed):
LGTM

paddle/fluid/memory/detail/system_allocator.cc
sneaxiy: Discuss offline. Maybe we should change the scripts in benchmark repo first. This PR may not change this log.

paddle/fluid/platform/enforce.h
sneaxiy: Done.

paddle/fluid/memory/allocation/allocator_facade.cc
sneaxiy: Done.

paddle/fluid/memory/allocation/retry_allocator.cc
sneaxiy: Done.

paddle/fluid/memory/allocation/allocator.h
```diff
@@ -19,21 +19,21 @@
...
 
 namespace paddle {
 namespace memory {
 namespace allocation {
 
 // Exception when `Alloc`/`AllocShared` failed
-class BadAlloc : public std::exception {
- public:
-  inline explicit BadAlloc(std::string msg) : msg_(std::move(msg)) {}
+struct BadAlloc : public std::exception {
```
zhhsplendid, line 30: I found that there are not only one place using f, l ... including old code ... please just modify the f -> file, l -> line at newly added code. You don't have to change old code.

sneaxiy: Done. I have changed the old codes too.

zhhsplendid (dismissed):
LGTM

paddle/fluid/memory/allocation/retry_allocator_test.cc
```diff
@@ -93,6 +97,50 @@ TEST(RetryAllocator, RetryAllocator) {
   }
 }
 
+class DummyAllocator : public Allocator {
+ public:
+  bool IsAllocThreadSafe() const override { return true; }
+
+ protected:
+  Allocation *AllocateImpl(size_t size) override {
```
zhhsplendid, line 105: Great dependency injection for test! Nothing to change. Just applause.

paddle/fluid/memory/allocation/retry_allocator_test.cc
```diff
@@ -93,6 +97,50 @@ TEST(RetryAllocator, RetryAllocator) {
...
+    PADDLE_THROW_BAD_ALLOC("Always BadAlloc");
+  }
+
+  void FreeImpl(Allocation *) override {}
+};
+
+TEST(RetryAllocator, RetryAllocatorLastAllocFailure) {
+  {
+    RetryAllocator allocator(std::make_shared<DummyAllocator>(), 10);
+    try {
```
zhhsplendid, line 116: Very tiny issue. You can decide whether to change or not.

Usually, actual values of variables (such as "10" here, "true/false" in some code) can decrease readability of code. A software engineer has to look at API if he/she doesn't know the second parameter of RetryAllocator is "retry_ms".

some people use
"size_t retry_ms=10; // may also be const at head of file
RetryAllocator allocator(std::make_shared<DummyAllocator>(), retry_ms)" 
or "RetryAllocator allocator(std::make_shared<DummyAllocator>(), /* retry_ms = */ 10)"

However, it depends on programmer's background knowledge whether to add it. For example, in the next line, "allocator.Allocate(100)", since "malloc" is very common in C/C++, so we don't have to add that and SWE will have sense it's byte size.

In summary, consider adding readability of code for future developer, but not too much. In this case. I suggest adding "/* retry_ms = */ 10", but keep "allocator.Allocate(100)".

zhhsplendid (dismissed):

paddle/fluid/memory/allocation/retry_allocator_test.cc
sneaxiy: Thanks!

paddle/fluid/memory/allocation/retry_allocator_test.cc
sneaxiy: Done.

chengduoZH (approved):
LGTM

PR #19410, opened 2019-08-26
Remove unnecessary op when trainable is false

chengduoZH:
If the parameter is not trainable, the parameter should not have a gradient and any optimization operators.

TODO: remove the backward ops which are not used to generate the gradients of trainable parameters.

### DIFF
--- a/python/paddle/fluid/optimizer.py
+++ b/python/paddle/fluid/optimizer.py
@@ -588,6 +588,13 @@ class Optimizer(object):
             tuple: (optimize_ops, params_grads) which are, list of operators appended;
             and list of (param, grad) Variables pair for optimization.
         """
+        assert isinstance(loss, Variable), "The loss should be an Variable."
+        parameters = loss.block.program.global_block().all_parameters()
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/optimizer.py
```diff
@@ -587,6 +588,12 @@ def minimize(self,
             tuple: (optimize_ops, params_grads) which are, list of operators appended;
             and list of (param, grad) Variables pair for optimization.
         """
+        assert isinstance(loss, Variable), "The loss should be an Variable."
+        parameters = loss.block.program.global_block().all_parameters()
+        param_no_trainable = set(
+            [param.name for param in parameters if param.trainable is False])
+        if no_grad_set is None: no_grad_set = set()
+        no_grad_set.update(param_no_trainable)
```
chengduoZH: If the parameter is no trainable, it should not have a gradient.

sneaxiy (approved):
LGTM.

kuke (approved):
LGTM

PR #19296, opened 2019-08-20
support Tensor input for row conv op

LDOUBLEV:

PR #19412, opened 2019-08-26
Support Tensor input for row conv op

LDOUBLEV:
Support tensor input for row_conv_op.  The input of row_conv_op can be tensor or lodtensor, and no other extra input is used.  When input is tensor, output is also tensor with padding.


qingqing01 (approved):
Has reviewed in https://github.com/PaddlePaddle/Paddle/pull/19296

paddle/fluid/operators/row_conv_op.cc
```diff
@@ -1,5 +1,4 @@
 /* Copyright (c) 2016 PaddlePaddle Authors. All Rights Reserved.
-
```
heavengate, line 2: 这行不用删

Superjomn (approved):
LGTM

paddle/fluid/operators/row_conv_op.cu
```diff
@@ -348,9 +369,27 @@ class RowConvGradKernel<platform::CUDADeviceContext, T>
...
+    if (is_tensor) {
+      for (int i = 0; i < batch_size + 1; i++) {
+        batch_indices[i] = i * timesteps;
+      }
+      input_dim = X->dims()[2];
+    } else {
+      batch_indices = X->lod()[0];
+      input_dim = X->dims()[1];
+    }
+    // int input_dim = X->dims()[1];
```
heavengate, line 392: detele

frankwhzhang (approved):
LGTM API.

paddle/fluid/operators/row_conv_op.cc
LDOUBLEV: OK

PR #19415, opened 2019-08-26
Remove open_files codes

sneaxiy:
RecordIO is not supported since #18823. This PR removes the rest of related codes in `open_files`.


luotao1 (commented):
There are still recordio codes which don't be deleted now
image

PR #19443, opened 2019-08-27
adapte fleet api for localsgd and support nccl comm configuration in …

gavin1332:
…executor

test=develop


python/paddle/fluid/transpiler/collective.py
```diff
@@ -278,10 +278,12 @@ def _transpile_startup_program(self):
         Collective._transpile_startup_program(self)
 
         block = self.startup_program.global_block()
+        non_dist_params = []
         for param in block.iter_parameters():
```
gavin1332, line 282: fix failure in python35 ut. For the ordered set in block  should not be modified in iteration.

sandyhouse (approved):
LGTM

PR #19447, opened 2019-08-27
Add persistable variables initialized check in exe.run to polish the error message of run(startup_porgram) before optimizer.minimize

chenwhql:
This PR wants to solve the `run(startup_program) ` position sensitive program. When `run(startup_program) ` is executed before `optimizer.minimize`, because the parameters added by `optimizer` are not initialized, which will cause run failed. The error message is like:

image

this error occur to `RunTimeInferShape` phase. the program doesn't check whether the tensor in `LearningRate` variable is initialized, and get the dim of variable used to compare directly, which is the reason that this error messgae is not easy to understand. 

So this PR add persistable Input variables check in exe.run(main_program), the result after change is like：

image

This method is not very elegant, maybe there is a better way to solve this problem.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -3581,6 +3581,13 @@ class Program(object):
             for each_var in list(each_block.vars.values()):
                 yield each_var
 
+    def is_startup_program(self):
+        for each_block in self.blocks:
+            for each_var in list(each_block.vars.values()):
+                if not each_var.persistable:
+                    return False
+        return True
+
 
 class Parameter(Variable):
     """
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/framework.py
```diff
@@ -2867,6 +2873,10 @@ def __init__(self):
         # appending gradients times
         self._appending_grad_times = 0
 
+        # when program is executed by exe.run, it will be set False
+        # When new op or var was append to program, it will be set True
+        self._add_new_elements = True
```
chenwhql, line 2878: I think adding flag here is not very elegant, but I have not thought of improving efficiency in a better way.

sneaxiy (changes requested):
Very hard-coding...

PR #19468, opened 2019-08-27
Delete pserver complete file before executor running.

gongweibao:
1. Delete pserver complete file before executor running.
2. Add more detail log of unit test context to find unit test error.


gavin1332 (approved):
LGTM

PR #19477, opened 2019-08-27
add detach for Variable in dygraph, test=develop

JepsonWong:
Add detach for Variable in dygraph.
This interface returns a new Variable (stop_gradient=True), detached from the current graph.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -34,8 +34,6 @@ from .proto import framework_pb2
 from . import core
 from . import unique_name
 
-import paddle
-
 __all__ = [
     'Program',
     'default_startup_program',
@@ -546,14 +544,23 @@ class Variable(object):
             self.is_data = is_data
 
     def detach(self):
-        new_var = self._cloneVar()
-        self.block.append_op(
-            type="assign",
-            inputs={'X': [self]},
-            outputs={'Out': [new_var]},
-            attrs={'device': 0},
-            stop_gradient=True)
-        return new_var
+        """
+        Returns a new Variable, detached from the current graph.
+        
+        Returns:
+            Variable: The detached Variable.
+        """
+        if in_dygraph_mode():
+            new_var = self._cloneVar()
+            self.block.append_op(
+                type="assign",
+                inputs={'X': [self]},
+                outputs={'Out': [new_var]},
+                attrs={'device': 0},
+                stop_gradient=True)
+            return new_var
+        else:
+            raise Exception("static graph model DO NOT supprt detach")
 
     def numpy(self):
         new_ivar = self._ivar._copy_to(core.CPUPlace(), True)
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -34,6 +34,8 @@ from .proto import framework_pb2
 from . import core
 from . import unique_name
 
+import paddle
+
 __all__ = [
     'Program',
     'default_startup_program',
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/framework.py
```diff
@@ -34,6 +34,8 @@
 from . import core
 from . import unique_name
 
+import paddle
```
JiabinYang: DO NOT import paddle here, this may cause circle deps

python/paddle/fluid/framework.py
```diff
@@ -543,6 +545,16 @@ def __init__(self,
             self._stop_gradient = stop_gradient
             self.is_data = is_data
 
+    def detach(self):
+        new_var = self._cloneVar()
```
JiabinYang: use `in_dygraph_mode()` to make sure this will not be used in static graph mode, and throw error when user use it. What's more, add Doc
for this API just like other apis

python/paddle/fluid/tests/unittests/test_detach.py
```diff
@@ -0,0 +1,180 @@
...
+                initializer=fluid.initializer.Constant(9.0))
+            fc2_b_param_attrs = fluid.ParamAttr(
+                initializer=fluid.initializer.Constant(10.0))
+            fc2 = FC("fc",
+                     1,
+                     num_flatten_dims=1,
+                     param_attr=fc2_w_param_attrs,
+                     bias_attr=fc2_b_param_attrs)
+            data = to_variable(data)
+            conv = fc(data)
```
JiabinYang: why do u call this conv? change name to fc?

python/paddle/fluid/tests/unittests/test_detach.py
```diff
@@ -0,0 +1,180 @@
...
+            dim = dim * index_shape
+        return dim
+
+    def test_NoDetachMulti_DetachMulti(self):
+        array_no_detach_multi = self.no_detach_multi()
+        array_detach_multi = self.detach_multi()
+
+        array_no_detach_multi_dim = self.get_Dim(array_no_detach_multi)
+
+        array_no_detach_multi = array_no_detach_multi.reshape(
```
JiabinYang: no need to do this, why not get numpy value and use np.array_equalI() to assert the value equality

JiabinYang (commented):
Some comments

python/paddle/fluid/framework.py
JepsonWong: done

python/paddle/fluid/framework.py
JepsonWong: done

python/paddle/fluid/tests/unittests/test_detach.py
JepsonWong: done

python/paddle/fluid/tests/unittests/test_detach.py
JepsonWong: done

python/paddle/fluid/tests/unittests/test_detach.py
```diff
@@ -0,0 +1,162 @@
...
+        array_detach_multi = self.detach_multi()
+        assert np.array_equal(array_no_detach_single, array_detach_multi)
+
+    def test_detach_exception(self):
+        x = fluid.layers.data(name="a", shape=[3, 4], dtype='float32')
+        y = fluid.layers.fc(input=x, size=10, bias_attr=True)
+        try:
+            y_detach = y.detach()
+        except Exception as e:
+            print(e)
```
JiabinYang: This print seems has no effect for ut, better option is raise an specific exception and catch it here, or do some string match to make sure it raises a predictable exception

JiabinYang (commented):
some comments

python/paddle/fluid/tests/unittests/test_detach.py
JepsonWong: done

JiabinYang (approved):
LGTM

luotao1 (approved):
LGTM for framework.py

PR #19484, opened 2019-08-28
Add stop_gradient in range_api

baiyfbupt:
Do stop_gradient in range() api 's output
fix http://newicafe.baidu.com/issue/NLPP-34/show?from=page


wanghaoshuang (approved):
LGTM

PR #19487, opened 2019-08-28
Fix softmax segmentation fault in AVX

sneaxiy:
`_mm256_load_ps` would raise `SIGSEG` when the address is not aligned, which makes softmax op segmentation fault when compiled with debug mode. This PR fixes it by using `_mm256_loadu_ps `.

8b2fc171ab8248a42954fec5dce84127


tensor-tang (approved):
LGTM

PR #19491, opened 2019-08-28
test=release/1.5, fix multiple Layers parameter missing error

JiabinYang:
This PR cherry-picked the fix of multiple Layers parameter missing error in dygraph mode，the original one is #18968


python/paddle/fluid/dygraph/nn.py
```diff
@@ -1374,13 +1375,20 @@ def _build_once(self, input):
                 shape=param_shape,
                 dtype=self._dtype,
                 default_initializer=Constant(1.0))
+        else:
+            if self._param_attr:
+                logging.warn("param_attr are only avaliable with scale is True")
```
xsrobin, line 1380: available，记得全文替换

PR #19503, opened 2019-08-28
Fix mean_iou api misleading example

baiyfbupt:
Fix mean_iou api misleading example
predict and label should have save shape
fix http://newicafe.baidu.com/issue/DLTP-2253/show


baiyfbupt:
mean_iou

wanghaoshuang (approved):

heavengate (approved):

PR #19506, opened 2019-08-28
Change backward_guard to optimize_guard to maximize the allreduce overlap.

gongweibao:
1. Change backward_guard to optimize_guard to maximize the allreduce overlap.
If we put `cast` operator in backward stage， it will wait for allredudce completion, and since paddle has only one compute stream, the stream will hang for waiting. So we need to put cast operator in optimize stage to avoid this.
image

2. Add more logs to locate potential problem.

### DIFF
--- a/python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
+++ b/python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
@@ -80,8 +80,11 @@ def create_master_params_grads(params_grads, main_prog, startup_prog,
         A list of master parameters and gradients. 
     """
     master_params_grads = []
-    with main_prog._backward_role_guard():
-        for p, g in params_grads:
+    #with main_prog._backward_role_guard():
+    #    for p, g in params_grads:
+    for p, g in params_grads:
+        # create master parameters
+        with main_prog._optimized_guard([p, g]):
             # create master parameters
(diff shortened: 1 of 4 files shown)


guru4elephant (commented):
I think you can explain the reason why this pr can maximize the allreduce overlap with some figures or examples.

gongweibao:
Done.Thanks!

kuke (commented):
LGTM for the optimize_guard

python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
```diff
@@ -80,8 +80,9 @@ def create_master_params_grads(params_grads, main_prog, startup_prog,
         A list of master parameters and gradients. 
     """
     master_params_grads = []
-    with main_prog._backward_role_guard():
-        for p, g in params_grads:
+    for p, g in params_grads:
+        # create master parameters
+        with main_prog._optimized_guard([p, g]):
```
gavin1332, line 85: what's the difference while training with pe? As the op orders in the front-end have no effect on the dependency.

gongweibao: It doesn't change the dependency, but delay op's running to avoid waiting.

gavin1332 (approved):
LGTM

PR #19508, opened 2019-08-28
Add the support the int64 data type of `scatter_op`  input Index(#18804)

wawltor:
support the scatter_op when the input Index data type is int64, when the Index is big, the int32 can not be hold the index.


kuke (approved):
LGTM

PR #19509, opened 2019-08-28
Update elementwise double grad to save gpu memory

zhiqiu:
This PR includes:

- update `elementwise add/sub/div/mul double grad op`, add `inplace` and `NoNeedBufferVars` to save memory
-  update `activation double grad op`, add `inplace` to save memory 
-  update computations in `activation op`, remove unnecessary `eval()` function in `eigen` expressions, since `eval()` leads to temporary tensors
-  add c++ unittests for `elementwise double grad op` to cover branches that are hard to cover by python unittests
-  update `OpTest`  to support `inplace` test for double grad op


sneaxiy (approved):
LGTM.

PR #19510, opened 2019-08-28
[Work in progress] MKL-DNN 1.0 update

grygielski:
**Final version target:** Release 1.7
**Current state:** First functional version has been made. Its performance was tested on AVX2 and AVX512 machines (containing 6248 machine) and compared to current develop branch. Performance of FP32 is mostly the same but Int8 has improved a little (~5%).
**Significant changes:** This version does not contain any significant changes touching PaddlePaddle core. There are mostly cosmetic and MKL-DNN core changes.
**Future plans:** We will maintain and resolve conflicts if needed. Refactoring of some mkldnn ops are also in future plans.

This PR exceeds files and code lines limit but we cannot divide it because whole mkl-dnn infrastructure has been rebuilt.


grygielski:
@luotao1 We posted this PR to let You know about current state of this topic and answer questions if there are any.

luotao1:
> This PR exceeds files and code lines limit but we cannot divide it because whole mkl-dnn infrastructure has been rebuilt.

Hope you could divide it, not just for files or code lines. If this new PR has some CE/benchmark/Memory problems, you should revert it at all. And it's hard for you to track and debug it.

I have some findings just for this PR:
- There are a lot of `mkldnn::memory::format`->`mkldnn::memory::format_tag`, could you write it in one file? It could reduce a lot of update lines.
```
using mkldnn_memory_format = mkldnn::memory::format;
```
- `paddle::framework::vectorize2int(tensor->dims())`->`paddle::framework::vectorize(tensor->dims())`, could you replace it at first?

grygielski:
With `using mkldnn_memory_format = mkldnn::memory::format;` we can reduce lines changed but not number of files because of other changes in them but sure, I can do that to prepare code for 1.0 update.

We cannot change vectorize2int to vectorize since mkldnn v0.2 uses int in its primitives and v1.0 uses int64_t. There are other places where this function is used so can't change its behavior also. However, there is possibility to wok around this by creating separate function calling vectorize and use it in mkldnn ops (as a separate PR) but it will generate additional, redundant code.

@luotao1 Additionaly could You please test this version on Your machine for some models in spare time? I would like to know if everything works just fine on Baidu's machines.

paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
```diff
@@ -96,10 +95,10 @@ void eltwise_forward(const framework::ExecutionContext &ctx,
       x->dims().size() == 2 || x->dims().size() == 3 || x->dims().size() == 4,
       "Input dim must be with 2, 3 or 4");
 
-  std::vector<int> src_tz = framework::vectorize2int(x->dims());
+  std::vector<int64_t> src_tz = paddle::framework::vectorize(x->dims());
```
luotao1, line 98: How about?
```
auto src_tz = paddle::framework::vectorize(x->dims());
```
Maybe we can unify `vectorize` and `vectorize2int` at first.

luotao1:
> Additionaly could You please test this version on Your machine for some models in spare time? I would like to know if everything works just fine on Baidu's machines.

It's OK.

paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
grygielski: Code You proposed does work for v1.0 but will crash in v0.2 because vectorize() returns int64_t vector and mkldnn v0.2 kernel expects int vector. We can't unify this as cuda kernels also use vectorize2int() function and vectorize() is used in some other places not maintained by us. Improvents like using auto will be made during code cleanup and refactor in the near future.

Sand3r-: What if we have template parametrised the vectorize function, so that by default it returns int64_t but it can be chosen through a template parameter what data type we want our dimensions to be returned in?

luotao1: @Sand3r- I think it's OK. Could you create a separate PR to do this?

grygielski: I'll do that

grygielski: Done in https://github.com/PaddlePaddle/Paddle/pull/19611

luotao1:
Could you update this PR since there are a lot of conflicts here?

grygielski:
Is it important for You to have them resolved right now? This update is scheduled for 1.7 version so we are now preparing ground for that by developing series of mkldnn handler update PRs in order to clean a lot of code. But this will affect almost every mkldnn operator and make conflicts again. If it's not needed by You right now, I would wait till we stabilize mkldnn operator changes. @luotao1

luotao1:
Got it.

grygielski:
Closing this one since there is newer version of this PR in https://github.com/PaddlePaddle/Paddle/pull/20162

PR #19515, opened 2019-08-28
Some ngraph op and unittest fix

baojun-nervana:
- fix some edge cases of ngraph op
- disable inplace_grad for ngraph as it is not supported


mozga-intel (dismissed):
LGTM!

luotao1 (approved):
LGTM

PR #19519, opened 2019-08-29
Remove reset recordio usage

sneaxiy:
RecordIO is not supported any more. This PR removes all of the rest RecordIO usage in Paddle codes.


luotao1 (approved):
LGTM

PR #19529, opened 2019-08-29
remove unused assert.h

luotao1:
In general, it's recommended to use `PADDLE_ENFORCE` for error_mesage. However, in CUDA device kernel or HOSTDEVICE function, we must use `PADDLE_ASSERT` now.

This PR remove the unused `assert.h`, and is the next of #19285. 
TODO: 
1. remove `PADDLE_ASSERT`, remain only `PADDLE_ASSERT_MSG`, i.e, we should write error message in CUDA device kernel.
2. unify `PADDLE_ASSERT_MSG` into `PADDLE_ENFORCE` to meet the demands of CUDA device kernel.


sneaxiy (approved):
LGTM.

PR #19064, opened 2019-08-08
Use sparse matrix to implement fused emb_seq_pool operator

yihuaxu:
To implement fused emb_seq_pool operator with sparse matrix will be save the cost of  memory loading/transferring. It is helpful to optimize this operator.

PR #19533, opened 2019-08-29
fix the bug that mklkl URL in windows doesn't support mkl_CSRMM

zhwesky2010:
The PR(#19064) use sparse matrix to implement fused emb_seq_pool operator, and Update the URL of mklml library. But, CSRMM of the new mklml library dosen't support no-Linux at present.
So we let the library of mklml stay the same.


liupluswei (approved):
LGTM

PR #19543, opened 2019-08-29
[Do not merge] Modify conv INT8 scales functions, getWeightsTz and getDstType

lidanqing-intel:
@wojtuss Please take a look at function GetDstType, do you think this way would be better?

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -29,34 +29,12 @@ using mkldnn::stream;
 using platform::to_void_cast;
 using platform::GetMKLDNNFormat;
 
-inline void GetWeightsTz(std::vector<int>& weights_tz, int groups,  // NOLINT
-                         bool is_conv3d) {
+inline void GetWeightsTz(std::vector<int>& weights_tz, int groups) {  // NOLINT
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -69,22 +47,58 @@ inline mkldnn::memory::format GetWeightsFormat(mkldnn::memory::format format,
...
+  }
+}
+
 static mkldnn::memory::data_type GetDstType(bool is_int8,
                                             bool force_fp32_output,
                                             std::string fuse_activation,
                                             bool fuse_residual_conn,
                                             const Tensor* residual_param) {
   auto dst_dt = mkldnn::memory::data_type::f32;  // uint8_t, int8_t, float
   if (is_int8) {
```
wojtuss: I would turn it into:
```
if (is_int8 && !force_fp32_output) {
   if (fuse_residual_conn && residual_param) {
      ...
   } else {
      ...
   }
}
```

lidanqing-vv: Yes, that would be better. I changed.

lidanqing-vv:
Will reopen when I finish removing "else (keys exist)" branch

PR #18905, opened 2019-07-30
Add conv dequant squash

wozna:
This squash is based on PR https://github.com/PaddlePaddle/Paddle/pull/18754
This squash change chain with conv op. and dequantize op. to conv with forced float output. 
Thanks to this you will avoid additional conversion to int8. 

In mobilenet ssd there are 12 pattern like this and the use of this squash improves accuracy 
(on clx machine):
    - Before squash:  acc_fp32 = 0.7733,  acc_int8=0.77 
    - After that squash: acc_fp32 = 0.7733,  acc_int8=0.7719 

Unfortunately, after a few tests using mkl-dnn and int8 on clx machine, fps decreased:
    - Before squash: avg fps  = 151.634
    - After squash: avg fps = 146.8749
Is's almost 3%. 
That  could be some kind of option if accuracy is more important than performance.

PR #19545, opened 2019-08-29
Fix conv2d+dequantize squash for residual fusion

wozna:
It fix recently merged PR #18905
The squash `conv2d+dequantize` must be skipped if the `conv2d` op has the attribute `fuse_residual_connection` set to `true`, because residual fusion does not support force output with fp32.

There is no change in number of squashed pairs in mobilenet_ssd.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_squash_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_squash_pass.cc
@@ -177,7 +177,12 @@ void CPUQuantizeSquashPass::ConvDequantSquash(Graph* graph) const {
     GET_IR_NODE_FROM_SUBGRAPH(dequant_out, dequant_out, conv_dequant_pattern);
 
     // if conv2d has one output
-    if (conv_out->outputs.size() == 1) {
+    // and there is no fuse residual connection
+    // because residual fusion does not support force output with fp32
+    if (conv_out->outputs.size() == 1 &&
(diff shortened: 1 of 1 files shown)


paddle/fluid/framework/ir/mkldnn/cpu_quantize_squash_pass.cc
```diff
@@ -177,7 +177,12 @@ void CPUQuantizeSquashPass::ConvDequantSquash(Graph* graph) const {
     GET_IR_NODE_FROM_SUBGRAPH(dequant_out, dequant_out, conv_dequant_pattern);
 
     // if conv2d has one output
-    if (conv_out->outputs.size() == 1) {
+    // and there is no fuse residual connection
+    // because residual fusion does not support force output with fp32
+    if (conv_out->outputs.size() == 1 &&
+        !(conv_op->Op()->HasAttr("fuse_residual_connection") &&
```
Sand3r-: Is it necessary to perform conv out number check?

Also, you can replace the second condition with `!(conv_op->GetAttrIfExists("fuse_residual_connection"))`

wozna: Yes, we should check conv out number if its one or check if every op after conv is dequantize. 
Right now there is no model that have that, but its good to avoid such situation. 
image

Sand3r-:
@luotao1 This fix is required for QAT resnet50 to not fail on `cpu_quantize_squash_pass`.

PR #19557, opened 2019-08-30
add:GetInputTensorShape, ZeroCopyTensor support uint8; fix:profile, trt shared param bugs

cryoco:
-- added GetInputTensorShape() method for AnalysisPredictor.
-- added uint8 support for ZeroCopyTensor.
-- changed way of enabling profiling from setting GFLAGS to specifying AnalysisConfig options, which fixed the GFLAGS not working bug.(currently only supports AnalysisPredictor)
-- fixed 2 bugs of TensorRT, which happens when 2 ops share the same weight parameter:
    1. weight_map might be set with the same key multiple times, which causes PADDLE_ENFORCE not met error;
image
    2. some parameters are marked as repetitive and deleted after subgraph pass to reduce GPU memory, but if some of them are shared by multiple ops, they might still be needed in fluid, which causes fatal error.
image

-- added unittests for the features above.


luotao1:
Maybe it's better to separate this PR into 5 small PRs.
- added GetInputTensorShape() method for AnalysisPredictor.
- added uint8 support for ZeroCopyTensor.
- changed way of enabling profiling from setting GFLAGS to specifying AnalysisConfig options
- fix bug1
- fix bug2

Then, it's easy for developers to see each change of the feature or the bug. Now we could not see which part is to fix bug1, which part to fix bug2, etc.

For fix bug, you should paste the bug error message. Only description of `weight_map might be set with the same key multiple times, which causes PADDLE_ENFORCE not met error` is not enough.

PR #19562, opened 2019-08-30
Remove constraint that last dimension is forced to be 1 in huber_loss op

Aurelius84:
**before：**
+ the `label` must be 2-D tensor with shape [batch_size, 1]
```python
import paddle.fluid as fluid

x = fluid.layers.data(name='x', shape=[13], dtype='float32')
predict = fluid.layers.fc(input=x, size=1)
label = fluid.layers.data(name='label', shape=[1], dtype='float32')
loss = fluid.layers.huber_loss(input=predict, label=label, delta=1.0)

place = fluid.CPUPlace()
x_data = np.random.rand(10, 13).astype("float32")
label_data = np.random.random(size=(10, 1)).astype('float32')
exe = fluid.Executor(place)
ret = exe.run(feed={'x': x_data, 'label': label_data}, 
              fetch_list=[loss], return_numpy=False)
```

**after：**
+ the `label` can be tensor with shape [batch_size, 1] or [batch_size]
```python
batch_size = 10
label = fluid.layers.data(name='label', shape=[batch_size],append_batch_size=False, dtype='float32')
...
loss = fluid.layers.huber_loss(input=predict, label=label, delta=1.0)
...
label_data = np.random.random(size=(10)).astype('float32')
```

### DIFF
--- a/paddle/fluid/operators/huber_loss_op.cc
+++ b/paddle/fluid/operators/huber_loss_op.cc
@@ -25,27 +25,32 @@ class HuberLossOp : public framework::OperatorWithKernel {
   using framework::OperatorWithKernel::OperatorWithKernel;
 
   void InferShape(framework::InferShapeContext* ctx) const override {
-    PADDLE_ENFORCE(ctx->HasInput("X"), "Input(X) must be initialized.");
-    PADDLE_ENFORCE(ctx->HasInput("Y"), "Input(Y) must be initialized.");
+    PADDLE_ENFORCE_EQ(ctx->HasInput("X"), true,
+                      "Input(X) must be initialized.");
+    PADDLE_ENFORCE_EQ(ctx->HasInput("Y"), true,
+                      "Input(Y) must be initialized.");
 
     auto x_dims = ctx->GetInputDim("X");
     auto y_dims = ctx->GetInputDim("Y");
+    int rank = x_dims.size();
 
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/huber_loss_op.cc
```diff
@@ -25,27 +25,34 @@ class HuberLossOp : public framework::OperatorWithKernel {
...
+    if (rank == y_dims.size()) {
+      PADDLE_ENFORCE_EQ(y_dims[rank - 1], 1U,
+                        "The last dimension of Input(Y) should be equal to 1.");
+    } else {
+      PADDLE_ENFORCE_EQ(rank, y_dims.size() + 1,
+                        "The rank of Input(X) should be equal to "
+                        "the rank of Input(Y) plus 1.");
+    }
     if (ctx->IsRuntime() ||
         (framework::product(x_dims) > 0 && framework::product(y_dims) > 0)) {
```
phlrain: product 有风险，如果shape中有两个-1， 会有问题

phlrain: 建议使用 contain_unknown_dim

Aurelius84: 好的，已改成contain_unknown_dim进行判断

phlrain (approved):
LGTM

PR #19568, opened 2019-08-30
using MKLDNNMemoryFormat = mkldnn::memory::format changes

grygielski:
Suggested in https://github.com/PaddlePaddle/Paddle/pull/19510# changes to make it easier to switch to MKLDNN v1.0 in the future. I didn't change it in Tensor class because we plan to apply some changes there anyway and I was close to 20 file limit in this PR.

### DIFF
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -18,7 +18,6 @@
 
 #include "paddle/fluid/operators/math/math_function.h"
 #ifdef PADDLE_WITH_MKLDNN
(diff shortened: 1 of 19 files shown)


luotao1:
Please update `PADDLE_ENFORCE` to pass the CI.

grygielski:
@luotao1 I updated every PADDLE_ENFORCE that appeared in my PR but I'm still getting CI error. Should I update every single one in modified files to make it pass?

Sand3r- (approved):
GJ @grygielski ! Pity that you had to update so many paddle enforces along with introducing the memory format string replacements. In any case, a work was well done. LGTM.

PR #19576, opened 2019-09-02
fix the warning caused by mistach arguments of flags.cmake

liupluswei:
fix the warning message in cmake casued by the mismatch arguments between if and endif in flags.cmake
image
image


JiabinYang (approved):
LGTM

PR #19581, opened 2019-09-02
Fix parallel compilation error of allocator

sneaxiy:
`system_allocator.cc` includes `place.h` but does not depend on its target, which makes compilation errors. This PR fixes it.
cb8a6da2b0eba1e2bfaa684550d4114b


luotao1 (approved):
LGTM

PR #19419, opened 2019-08-26
add padding in linear_chain_crf op

JesseyXujin:
Add padding  linear_chain_crf op. 
See changes in : https://github.com/PaddlePaddle/Paddle/pull/19040

PR #19583, opened 2019-09-02
add padding in linear_chain_crf op

JesseyXujin:
add padding in linear_chain_crf op
The history see: https://github.com/PaddlePaddle/Paddle/pull/19419


wanghaoshuang (approved):
LGTM on API.

kuke (approved):

PR #19585, opened 2019-09-02
Remove reshape transpose in attention module for bert optimization

bingyanghuang:
This PR is to refactor the PR #16342, one PR #18570 for matmul kernel has been merged. This PR contains:

- Pass for modifying the graph

- A fix for the merged matmul kernel for supporting last dimension value is not equal

> This PR can be run with the ERNIE, and there is a little accuracy diff with this PASS are under fixing. 

**Pass for graph modification:**

- Before
image

- After
image

### DIFF
--- a/paddle/fluid/inference/api/paddle_pass_builder.cc
+++ b/paddle/fluid/inference/api/paddle_pass_builder.cc
@@ -153,18 +153,18 @@ CpuPassStrategy::CpuPassStrategy() : PassStrategy({}) {
                   // "seqpool_concat_fuse_pass",    //
                   "seqpool_cvm_concat_fuse_pass",  //
                   // "embedding_fc_lstm_fuse_pass", //
-                  "fc_lstm_fuse_pass",             //
-                  "mul_lstm_fuse_pass",            //
-                  "fc_gru_fuse_pass",              //
-                  "mul_gru_fuse_pass",             //
-                  "seq_concat_fc_fuse_pass",       //
-                  "fc_fuse_pass",                  //
-                  "repeated_fc_relu_fuse_pass",    //
-                  "squared_mat_sub_fuse_pass",     //
-                  "conv_bn_fuse_pass",             //
-                  "conv_eltwiseadd_bn_fuse_pass",  //
-                  "is_test_pass",                  //
-                  "remove_reshape_transpose_pass", //
+                  "fc_lstm_fuse_pass",              //
+                  "mul_lstm_fuse_pass",             //
+                  "fc_gru_fuse_pass",               //
+                  "mul_gru_fuse_pass",              //
+                  "seq_concat_fc_fuse_pass",        //
+                  "fc_fuse_pass",                   //
+                  "repeated_fc_relu_fuse_pass",     //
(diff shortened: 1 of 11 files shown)


bingyanghuang:
Tested on Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz, with MKL, single thread

- Command line
` ./paddle/fluid/inference/tests/api/test_analyzer_bert --infer_model=third_party/inference_demo/bert_emb128/model --infer_data=third_party/inference_demo/bert_emb128/data.txt --gtest_filter=Analyzer_bert.profilel --paddle_num_threads=1 --num_threads=1 --repeat=1 --batch_size=1`

| Bert-128 | Before(ms) | After(ms) |
| :---         |     :---:      |          ---: |
| MKL  | 37.620     | 26.333    |
| MKL_DNN    | 42.244       | 33.342     |

paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -146,17 +146,18 @@ CpuPassStrategy::CpuPassStrategy() : PassStrategy({}) {
...
+                  "fc_gru_fuse_pass",               //
+                  "mul_gru_fuse_pass",              //
+                  "seq_concat_fc_fuse_pass",        //
+                  "fc_fuse_pass",                   //
+                  "repeated_fc_relu_fuse_pass",     //
+                  "squared_mat_sub_fuse_pass",      //
+                  "conv_bn_fuse_pass",              //
+                  "conv_eltwiseadd_bn_fuse_pass",   //
+                  "is_test_pass",                   //
+                  "remove_reshape_transpose_pass",  //
```
tensor-tang, line 167: same here https://github.com/PaddlePaddle/Paddle/pull/16342#discussion_r317476841

tensor-tang: BTW, this PR is doing a big fuse right?
So the name of `remove_reshape_transpose_pass` seems be not fair then, it's just say removing these two ops.

bingyanghuang: This PR is not the big fuse. As described in Brian's document, this pass is served to remove the redundant reshape and transpose ops in the BERT.

tensor-tang: Did not get the point if do not do some big fuse, then how can just remove transpose and reshape.

tensor-tang:
Add from the doc here https://github.com/PaddlePaddle/Paddle/pull/16342#issue-263117199
I found one question below
image

Where is the transpose did of `K`, set transpose true when sgemm?

bingyanghuang:
> Add from the doc here [#16342 (comment)](https://github.com/PaddlePaddle/Paddle/pull/16342#issue-263117199)
> I found one question below
> image
> 
> Where is the transpose did of `K`, set transpose true when sgemm?

It's set with inputs of matmul: tanspose_x and transpose_y. Sgemm just use these two inputs rather than change them.
image

bingyanghuang:
This PR  depend on PR #19708.

PR #19591, opened 2019-09-02
fix instag op

yaoxuefeng6:
There's a bug in Instag Op.

When 2nd input (instag) provides multiple tags for 1 ins.  'Out of range' error might come out.

for example,

Input 1 (input) has 8 ins.

Input 2 (instag) has more than 8 tags, but divided into 8 parts by LOD info. every part matches one ins from input 1.

if some ins has one tag, and some ins has more, it might trigger the bug.

I've check twice the logic of this op and modify some implementations.


paddle/fluid/operators/filter_by_instag_op.h
```diff
@@ -73,22 +73,21 @@ class FilterByInstagKernel : public framework::OpKernel<T> {
     } else {
       x1_lods = context.Input<LoDTensor>("Ins")->lod()[0];
     }
-
+    // std::cout << "X1 lod: "<< x1_lods.size() << std::endl;
+    // std::cout << "X2 lod: "<< x2_lods.size() << std::endl;
```
xjqbest: remove unnecessary lines

xjqbest (approved):
add some comment may be better,  this op is a little complicated....

PR #19606, opened 2019-09-03
Remove constraint that last dimension is forced to be 1 in cross_entropy

Aurelius84:
+ Remove constraint that last dimension is forced to be 1 of `label`
+ Keep shape of output loss  same with `label` shape
+ Remove unnecessary dims check in `GradKernel InferShape`.

### DIFF
--- a/paddle/fluid/operators/cross_entropy_op.h
+++ b/paddle/fluid/operators/cross_entropy_op.h
@@ -37,15 +37,18 @@ class CrossEntropyOpKernel : public framework::OpKernel<T> {
     int rank = x->dims().size();
     auto label_dims = labels->dims();
     Tensor x_2d = framework::ReshapeToMatrix(*x, rank - 1);
-    Tensor labels_2d;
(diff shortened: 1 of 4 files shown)


paddle/fluid/operators/cross_entropy_op.h
```diff
@@ -35,9 +35,20 @@ class CrossEntropyOpKernel : public framework::OpKernel<T> {
...
-    Tensor y_2d = framework::ReshapeToMatrix(*y, rank - 1);
+    Tensor labels_2d, y_2d;
+    if (label_dims.size() < rank) {
+      labels_2d.ShareDataWith(*labels);
+      labels_2d.Resize({framework::product(label_dims), 1});
+
+      y_2d.ShareDataWith(*y);
+      y_2d.Resize({framework::product(y->dims()), 1});
+
+    } else {
```
lanxianghit, line 48: else部分是否包括label_dims.size()  > rank的情况？

phlrain (approved):

PR #19609, opened 2019-09-03
no use

xjqbest:
no use


Thunderbrook (approved):

JiabinYang (changes requested):
Better to find out why instead of just try catch it?

PR #19610, opened 2019-09-03
optimize the GPU inplementation of stack_op

Xreki:
stack_op's function is the same as concat, except the output's dims. We can call concat functor to optimize it. The current code is implemented by thrust which may generate the calling of  `cudaMalloc`, `cudaFree` and `Wait`. See the timelien of ernie (https://github.com/PaddlePaddle/benchmark/issues/165#issuecomment-521134541):

image


jeng1220:
@NHZlX ,
looks like this patch can improve ERNIE performance, doesn't? Why did no one review this PR?

Xreki:
@jeng1220 stack op is optimized by https://github.com/PaddlePaddle/Paddle/pull/22142 .

jeng1220:
@Xreki,
So we can close this PR, right?

Xreki:
@jeng1220 Yes, I'll close it.

PR #19611, opened 2019-09-03
paddle::framework::vectorize() templatization

grygielski:
Suggested in https://github.com/PaddlePaddle/Paddle/pull/19510 changes to `paddle::framework::vectorize()` function. Following changes has allowed to get rid of `vectorize2int() `function from mkldnn ops.

@luotao1 Check if this modification is ok for You and let me know if I should swap `vectorize2int()` to `vectorize<int>()` in other files. They aren't maintained by us but since it's purely cosmetic change I think I can do it. This would require 2 more PR due to file limit.


luotao1:
LGTM! Thanks very much for your effort!

jczaja (approved):
LGTM

grygielski:
@luotao1 Could you tell me if I should change other `vectorize2int` to `vectorize<int>` in not mkldnn related files as well (in next PRs)?

luotao1:
If you are convenient you can change all other `vectorize2int` to `vectorize<int>`, otherwise, we will change it ourselves.

PR #19612, opened 2019-09-03
Enable inplace support for some ops

sneaxiy:
This PR enables inplace support to `clip`, `softmax`, `squeeze`, `unsqueeze` and `affine_channel` to save memory usage.


python/paddle/fluid/tests/unittests/test_affine_channel_op.py
```diff
@@ -109,8 +109,8 @@ def test_check_grad_stopgrad_dscale_dbias(self):
 
 class TestAffineChannelNHWCLargeShape(TestAffineChannelNCHWLargeShape):
     def init_test_case(self):
-        self.shape = [64, 32, 32, 512]
```
sneaxiy, line 112: I change this because in-place OpTest framework would run forward and backward to check whether in-place is correct. This unittest runs too slow when `self.shape = [64, 32, 32, 512]`, so I decrease the shape to accelerate it.

zhhsplendid (approved):
LGTM

Issue #19566, opened 2019-08-30
[MKL-DNN] Further clean softmax mkl-dnn op


jczaja:
softmax op works on flattened dims of Tensor. For this purpose Two input & two output tensors are created to share allocation , but diffrent in dims. 

This is not needed, since dims for mkl-dnn primitives are stored in Memory descriptor, we can have one input and one output and just use diffrent memory descriptors (2D for actual computation) and the other MD to be registered in output

jczaja:
fix in #19615

PR #19615, opened 2019-09-03
[MKL-DNN] Further refactoring of softmax 

jczaja:
Those changes are refactoring further softmax mkldnn op . It should be easier to maintain and develop now
also it does fix issue #19566 and prepares code for soon to come mkl-dnn 1.0 integration

@grygielski Please review

### DIFF
--- a/paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
@@ -38,52 +38,69 @@ class SoftmaxMKLDNNHandler : public platform::MKLDNNHandler {
   SoftmaxMKLDNNHandler(const std::vector<int>& dims,
                        const MKLDNNMemoryFormat fmt,
                        const platform::MKLDNNDeviceContext& dev_ctx,
-                       mkldnn::engine engine, const std::string& base_key)
-      : platform::MKLDNNHandler(dev_ctx, engine, base_key),
-        dims_(dims),
-        fmt_(fmt) {}
+                       platform::Place cpu_place, const std::string& uniq_name)
+      : platform::MKLDNNHandler(dev_ctx, dev_ctx.GetEngine(),
+                                platform::GetHash(dims, uniq_name)),
+        place_(cpu_place),
+        fwd_pd_(nullptr),
+        bwd_pd_(nullptr) {
+    this->AcquireSoftmaxPrimitiveDescriptor(dims, fmt);
+  }
 
   SoftmaxMKLDNNHandler(const std::vector<int>& dims,
                        const MKLDNNMemoryFormat fmt,
                        const MKLDNNMemoryFormat diff_fmt,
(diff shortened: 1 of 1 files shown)


grygielski (dismissed):
Thanks for introducing changes we discussed about! LGTM

paddle/fluid/operators/mkldnn/softmax_mkldnn_op.cc
```diff
@@ -38,52 +38,72 @@ class SoftmaxMKLDNNHandler : public platform::MKLDNNHandler {
...
 
   // TODO(jczaja): Once fwd_pd_ are moved to MKLDNNHandler then this function
   // should be moved as well eg. SoftmaxMKLDNNHandler -> MKLDNNHandler<softmax_>
-  std::shared_ptr<mkldnn::memory> AcquireSrcMemory(void* ptr) {
-    return this->AcquireMemory(dims_, platform::MKLDNNGetDataType<T>(), fmt_,
-                               ptr, "@user_src_mem_p");
+  std::shared_ptr<mkldnn::memory> AcquireSrcMemory(const Tensor* input) {
+    const T* input_data = input->data<T>();
+    return this->AcquireMemoryFromPrimitive(fwd_pd_->src_primitive_desc(),
+                                            const_cast<T*>(input_data),
```
luotao1: please do not use `const_cast`
http://ci.paddlepaddle.org/viewLog.html?buildId=158687&buildTypeId=YYTest_PrCiCpuPy2&tab=buildLog

jczaja: @luotao1 MKL-DNN interface does not accept const objects , so that is why const_cast is needed at some point. What solution would you recommend to achieve similar effect?

luotao1: Maybe you can see https://stackoverflow.com/questions/10498292/c-why-is-const-cast-evil

jczaja: That was interesting thread to read. There is one interesting fragment:
> The use of const_cast for any reason other than adapting to (old) libraries where the interfaces have non-const pointers/references but the implementations don't modify the arguments is wrong and dangerous.

This describe my use case eg. MKL-DNN does not have const interface. I asked MKL-DNN team to consider extending API to support const arguments, but due to specific MKL-DNN design this extension is unachievable to implement. So at the end we need to use const_cast. Please consider approving it.

jczaja: I just realized that not all _const_cast_ I had where needed and that we have a dedicated mechanism in paddle to deal with such a situation , so I updated PR.

grygielski (dismissed):
LGTM

luotao1:
please resolve the conflicts

jczaja:
@luotao1 Conflicts were resolved

luotao1 (approved):
LGTM

PR #19620, opened 2019-09-04
Add the support for mixed-precision configuration in fleet apis (collective mode).

sandyhouse:
Add the support for FP16 in fleet apis, collective mode only.
How to use:
```
     dist_strategy = DistributedStrategy()
     dist_strategy.use_mixed_precision = True
     dist_strategy.loss_scaling = args.loss_scaling
     optimizer = fluid.optimizer.Adam(...)
     optimizer = fleet.distributed_optimizer(optimizer, strategy=dist_strategy)
     optimizer.minimize(avg_cost)
```


python/paddle/fluid/tests/unittests/dist_mnist_fp16.py
```diff
@@ -0,0 +1,121 @@
...
+            initializer=fluid.initializer.Constant(value=0.01)))
+    return predict
+
+
+class TestDistMnist2x2(TestDistRunnerBase):
+    def get_model(self, batch_size=2, use_dgc=False, dist_strategy=None):
+        # Input data
+        images = fluid.layers.data(name='pixel', shape=[1, 28, 28], dtype=DTYPE)
+        label = fluid.layers.data(name='label', shape=[1], dtype='int64')
+        cast = fluid.layers.cast(images, "float16")
```
gongweibao, line 81: It's need not to cast inputs now. You can check the newest code of decrate.

python/paddle/fluid/tests/unittests/dist_mnist_fp16.py
```diff
@@ -0,0 +1,121 @@
...
+
+DTYPE = "float32"
+paddle.dataset.mnist.fetch()
+
+# Fix seed for test
+fluid.default_startup_program().random_seed = 1
+fluid.default_main_program().random_seed = 1
+
+
+def cnn_model(data):
```
gongweibao, line 42: Can we import cnn_model  function from dist_mnist.py?

python/paddle/fluid/tests/unittests/dist_mnist_fp16.py
```diff
@@ -0,0 +1,121 @@
...
+        input=conv_pool_2,
+        size=SIZE,
+        act="softmax",
+        param_attr=fluid.param_attr.ParamAttr(
+            initializer=fluid.initializer.Constant(value=0.01)))
+    return predict
+
+
+class TestDistMnist2x2(TestDistRunnerBase):
+    def get_model(self, batch_size=2, use_dgc=False, dist_strategy=None):
```
gongweibao, line 77: duplicate code with dist_mnist.

gongweibao (changes requested):
It needs not to cast input now. So the code can be simplified.

gongweibao:
I don't think it's a good idea to put mixed-precision decorator into the fleet API since it needs more options.
Just trait it as a common Optimizer and the user will call it outside the fleet API?

gavin1332:
> I don't think it's a good idea to put mixed-precision decorator into the fleet API since it needs more options.
> Just trait it as a common Optimizer and the user will call it outside the fleet API?

usually, there has to be an optimizer outside the fleet api, like SGD. Do you mean to wrap another optimizer over SGD outside?

gavin1332:
i have been thinking these days that what is the boundary of fleet api? or rather, what the rule to decide one function be inside the fleet api or not?

gongweibao:
https://github.com/PaddlePaddle/Paddle/pull/19682

PR #19624, opened 2019-09-04
fix scope lock bug on infer

tensor-tang:
According to https://github.com/PaddlePaddle/Paddle/blob/4cfe432cad1f6ec91cfd2add6cbbdf17f9645bec/paddle/fluid/framework/scope.cc#L36-L61

so, this should be a bug when on_infer=ON.

The problem raised:

image

image


luotao1 (approved):
LGTM

PR #19627, opened 2019-09-04
paddle::framework::vectorize() templatization [PART2]

luotao1:
Suggested in https://github.com/PaddlePaddle/Paddle/pull/19510 changes to `paddle::framework::vectorize()` function. Following changes has allowed to get rid of `vectorize2int() `function from mkldnn ops.

TODO： remove `std::vector<int> vectorize2int(const DDim& ddim)` in next PR due to file limits.


luotao1:
@grygielski @jczaja Please take a review!

paddle/fluid/operators/fused/fusion_conv_inception_op.cu
```diff
@@ -61,7 +61,7 @@ class CUDNNConvInceptionFusionOpKernel : public framework::OpKernel<T> {
     T* temp_data = temp_outs[0]->mutable_data<T>(input->dims(), ctx.GetPlace());
 
     DataLayout layout = DataLayout::kNCHW;
-    std::vector<int> in_dim = framework::vectorize2int(input->dims());
+    std::vector<int> in_dim = framework::vectorize<int>(input->dims());
```
grygielski, line 64: I'm not sure if You want to dive so much into changes but some lines might be improved by using `auto` keyword

grygielski (dismissed):
Please look at my comment, but: LGTM

paddle/fluid/operators/fused/fusion_conv_inception_op.cu
luotao1: I use one `sed` command to do this change, so I don't use `auto` keyword.

sneaxiy (approved):
LGTM.

PR #19638, opened 2019-09-05
Reduce thread num of retry_allocator_test

sneaxiy:
`retry_allocator_test` fails randomly in CI. Maybe it is because there are too many threads to allocate the same memory size. This PR reduces the thread number from 8 to 4.


luotao1 (approved):
LGTM

PR #19676, opened 2019-09-06
Make leaky relu inplacable

sneaxiy:
The previous implementation of `leaky_relu_grad` relies on forward input `X`, but it can be refined by relying on forward output `Out`, which makes `leaky_relu` inplacable in forward part. This helps to save memory usages of models which use `leaky_relu` operator.


chengduoZH (approved):
LGTM

PR #19680, opened 2019-09-06
Refine tensor.mutable_data to free memory before reallocate

sneaxiy:
`Tensor::mutable_data` would try to allocate new Allocation before release the old one. This PR make it free memory first and then re-allocate to save memory.


zhhsplendid (dismissed):
LGTM

zhhsplendid (approved):
LGTM

PR #19689, opened 2019-09-06
Upgrade ngraph to support mkldnn v1.0 

baojun-nervana:
The upgraded version is compatible with mkldnn v1.0.
It uses NGRAPH_USE_LEGACY_MKLDNN to switch between v0.x and v1.x. It currently set to be TRUE to use v0.x. When MKLDNN is upgraded to v1.x, it need to set to FALSE.
https://github.com/PaddlePaddle/Paddle/blob/6a28f7e7933d92cc19891ef9726896a94b05a1c4/cmake/external/ngraph.cmake#L79

CC. @grygielski @LeoZhao-Intel @mozga-intel

### DIFF
--- a/cmake/external/ngraph.cmake
+++ b/cmake/external/ngraph.cmake
@@ -37,7 +37,7 @@ INCLUDE(GNUInstallDirs)
 INCLUDE(ExternalProject)
 
 SET(NGRAPH_PROJECT         "extern_ngraph")
-SET(NGRAPH_GIT_TAG         "v0.24.0-rc.2")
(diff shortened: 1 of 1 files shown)


LeoZhao-BJ:
what is the version of 0.x with nGraph commit?

baojun-nervana:
> what is the version of 0.x with nGraph commit?

nGraph share mkldnn with DO.

LeoZhao-BJ:
LGTM

luotao1 (approved):

PR #19692, opened 2019-09-06
Add document annotations for FLAGS that need to be open to external developers

XiaoguangHu01:
Add document annotations for FLAGS that need to be open to external developers


lanxianghit (approved):
LGTM

PR #19708, opened 2019-09-09
add support of matmul with multiple head even different width and height

czhu15:
Original matmul with multiple head supports only the mat_a.width == mat_b.height,
in that case, mat_b will be horizontally split. In this patch, we extend the
support when mat_a.width != mat_b.height but mat_a.width/head_number == mat_b.height,
in this case, mab_b will be vertically split.

One example is A is [3, 8], B is [2, 16], head_number is 4. In this
case, A will be split as [3, 2], B will be (vertically) split as
[2, 4]. The final result will be 4 matrix of 4 matrix of [3,4], i.e. [3, 16]

test=develop

### DIFF
--- a/paddle/fluid/operators/math/blas.h
+++ b/paddle/fluid/operators/math/blas.h
@@ -125,7 +125,8 @@ class Blas {
                       const MatDescriptor& dim_a,
                       const framework::Tensor& mat_b,
                       const MatDescriptor& dim_b, T alpha, int head_number,
-                      framework::Tensor* mat_out, T beta) const;
+                      bool mat_y_split_vertical, framework::Tensor* mat_out,
+                      T beta) const;
 #endif
 #endif
 
@@ -194,9 +195,10 @@ class Blas {
 #if defined(PADDLE_WITH_MKLML) && !defined(PADDLE_WITH_CUDA)
   template <typename T>
   void BatchedGEMMWithHead(CBLAS_TRANSPOSE transA, CBLAS_TRANSPOSE transB,
-                           int M, int N, int K, T alpha, const T* A, const T* B,
-                           T beta, T* C, int batchCount, int64_t strideA,
-                           int64_t strideB, int64_t head_number) const;
+                           int W1, int H1, int W2, int H2, T alpha, const T* A,
+                           const T* B, T beta, T* C, int batchCount,
+                           int64_t strideA, int64_t strideB,
+                           int64_t head_number, bool split_b_vertical) const;
 #endif
 
   template <typename T>
--- a/paddle/fluid/operators/math/blas.h
+++ b/paddle/fluid/operators/math/blas.h
@@ -125,8 +125,8 @@ class Blas {
                       const MatDescriptor& dim_a,
                       const framework::Tensor& mat_b,
                       const MatDescriptor& dim_b, T alpha, int head_number,
-                      bool mat_y_split_vertical, framework::Tensor* mat_out,
-                      T beta) const;
+                      framework::Tensor* mat_out, T beta,
+                      bool mat_y_split_vertical) const;
 #endif
 #endif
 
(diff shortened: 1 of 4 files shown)


bingyanghuang:
@Sand3r- Could you help review this PR ?

bingyanghuang:
@Sand3r- @LeoZhao-Intel @yihuaxu Could you help review this PR?

paddle/fluid/operators/math/blas_impl.h
```diff
@@ -690,51 +725,93 @@ void Blas<DeviceContext>::MatMul(const framework::Tensor &mat_a,
...
 template <typename T>
 void Blas<DeviceContext>::MatMulWithHead(
     const framework::Tensor &mat_a, const MatDescriptor &dim_a,
     const framework::Tensor &mat_b, const MatDescriptor &dim_b, T alpha,
-    int head_number, framework::Tensor *mat_out, T beta) const {
-  PADDLE_ENFORCE_EQ(dim_a.width_, dim_b.height_);
+    int head_number, bool mat_b_split_vertical, framework::Tensor *mat_out,
+    T beta) const {
   PADDLE_ENFORCE_EQ(dim_a.width_ % head_number, 0);
   PADDLE_ENFORCE_GE(head_number, 1);
```
LeoZhao-BJ, line 741: what's the check on dim_b.width when "mat_b_split_vertical==true"? e.g.

PADDLE_ENFORCE_EQ(dim_b.width_ % head_number, 0);

paddle/fluid/operators/math/blas.h
```diff
@@ -125,7 +125,8 @@ class Blas {
                       const MatDescriptor& dim_a,
                       const framework::Tensor& mat_b,
                       const MatDescriptor& dim_b, T alpha, int head_number,
-                      framework::Tensor* mat_out, T beta) const;
+                      bool mat_y_split_vertical, framework::Tensor* mat_out,
```
LeoZhao-BJ: why not extend mat_y_split_vertical at the end ? and make mat_y_split_vertical=false as default.

paddle/fluid/operators/math/blas_impl.h
```diff
@@ -690,51 +725,93 @@ void Blas<DeviceContext>::MatMul(const framework::Tensor &mat_a,
...
+                        true);
+
+      this->template BatchedGEMMWithHead<T>(
+          transA, transB, dim_a.width_, dim_a.height_, dim_b.width_,
+          dim_b.height_, alpha, mat_a.data<T>(), mat_b.data<T>(), beta,
+          mat_out->data<T>(),
+          dim_a.batch_size_ == 0 ? dim_b.batch_size_ : dim_a.batch_size_,
+          dim_a.stride_, dim_b.stride_, head_number, mat_b_split_vertical);
+    }
   }
```
LeoZhao-BJ, line 804: can this part code be refined to avoid duplication?

paddle/fluid/operators/math/blas.h
czhu15: Will add mat_y_split_vertical at the end to keep it consistent with defination of BatchedGEMMWithHead. But to make it false as default lead to compiling error. I googled it and seems it is not easy to fix it.
https://stackoverflow.com/questions/4050202/why-default-argument-cannot-be-specified-for-an-explicit-template-specialization
since in our case, both true and false were equally been used. so I keep that parameter as non default.

paddle/fluid/operators/math/blas_impl.h
czhu15: you are right. will add that assert in the code when "mat_b_split_vertical==true"

paddle/fluid/operators/math/blas_impl.h
czhu15: will refactor the code to remove the duplication in the new submit.

paddle/fluid/operators/math/blas_impl.h
czhu15: done

paddle/fluid/operators/math/blas_impl.h
czhu15: done

LeoZhao-BJ:
LGTM

luotao1 (approved):
LGTM

PR #19714, opened 2019-09-09
Optimize amp for multi-gpu to enable FP16 gradients transfer across gpus

Jie-Fang:
For the original amp, all the gradients are FP32. Thus in multi-gpus training, FP32 gradients are transfered across gpus which may cause communication overhead. In the amp, we will obtain FP16 gradients of some ops automatically and we can transfer them instead to reduce the overhead.


python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
```diff
@@ -178,15 +184,15 @@ def rewrite_program(main_prog, amp_lists):
                         # this in_var isn't the output of other op
                         if in_var.op is None:
                             continue
-                        if in_var.op is op:
+                        elif in_var.op is op:
```
gongweibao, line 198: Why elif?

python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
```diff
@@ -130,6 +130,13 @@ def _insert_cast_op(block, op, idx, src_dtype, dest_dtype):
 
 
 def find_true_prev_op(ops, var_name):
+    """
+    Find the true prev op that outputs var_name variable.
+
+    Args:
+        ops (list): A list of ops.
+        var_name (string): Variable name.
+    """
```
gongweibao, line 140: I think we should assert that there should be only one op which generates this var.

gongweibao (changes requested):

python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
gongweibao: Just for safe!

python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
Jie-Fang: It's the same. I think elif may be better

gongweibao:
Can you give the performance comparison of Bert Large and Resnet50 with the old?
We should approve that float16 allreduce can improve communication.

Jie-Fang:
Yes, I think so. But I need ResNet50 multi-gpu version. The script in the models/PaddleCV/images_classification seems not work. It's OK for me to fix the bugs , but it'll takes more time. It's better you have correct script.

python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
Jie-Fang: I'll fix it

gongweibao:
I see.
But I think this pr should test on multi-GPUs to prevent the program bugs. Can you try this?
https://github.com/PaddlePaddle/Fleet/tree/develop/benchmark/collective/resnet

I will test the performance on multi-gpu machines.

gongweibao:
I will test it! Thanks!

kuke (approved):
LGTM

gongweibao (approved):
LGTM

wanghaoshuang (approved):

hutuxian (approved):

PR #19716, opened 2019-09-09
Remove constraint that last dimension is forced to be 1 by adding one_hot_v2

Aurelius84:
**before：**
+ the `input` label must be tensor with last dimension forced to be 1, such as shape [batch_size, 1]
```python
import paddle.fluid as fluid

label = fluid.layers.data(name="label", shape=[1], dtype="int32")
one_hot_label = fluid.layers.one_hot(input=label, depth=10)

place = fluid.CPUPlace()
label_data = np.random.rand(10, 1).astype("int32")
exe = fluid.Executor(place)
ret = exe.run(feed={ 'label': label_data}, 
              fetch_list=[one_hot_label], return_numpy=False)
```

**after：**
+ the `input` label can be tensor with shape [batch_size]
+ use `fluid.input.one_hot` instead of `fluid.layers.one_hot`
+ **Attention:** if input shape is [batch_size, 1], the output shape will be [batch_size, 1, depth]
```python
batch_size = 10
label = fluid.layers.data(name='label', shape=[batch_size],append_batch_size=False, dtype='int32')
...
one_hot_label = fluid.input.one_hot(input=label, depth=10)
...
label_data = np.random.rand(batch_size).astype("int32")
```


phlrain (dismissed):
LGTM

phlrain (approved):
LGTM

guoshengCS (approved):

luotao1 (approved):
LGTM for init.py

heavengate (approved):

PR #19719, opened 2019-09-09
add kernel for fill_op, test=develop

JepsonWong:
Add kernel for fill_op.

- The FillOp class originally inherited the framework::OperatorBase class, so it cannot be used in dygraph mode.

- By modifying the FillOp class to inherit the framework::OperatorWithKernel class, this op can be used in the dygraph mode.


paddle/fluid/operators/fill_op.h
```diff
@@ -0,0 +1,76 @@
...
+
+  framework::LoDTensor *tensor_;
+  const std::vector<float> &value_;
+};
+
+template <typename T>
+class FillKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const paddle::framework::ExecutionContext &ctx) const override {
+    auto *out = ctx.Output<framework::LoDTensor>("Out");
```
JiabinYang: `detail::Ref` can check if outs is NULL and throw a `PADDLE_ENFORCE` with info, if you wanna to use pointer here, better to do the same check to avoid segment fault with no info

python/paddle/fluid/tests/unittests/test_fill_op.py
```diff
@@ -18,23 +18,75 @@
...
+            dtype=int(core.VarDesc.VarType.FP64),
+            force_cpu=f_cpu,
+            Out='Out')
+        fill_op.run(scope, place)
+
+        # get result from Out
+        result_array = np.array(out)
+        full_array = np.array(val, 'float64')
+
+        self.assertTrue(np.allclose(result_array, full_array, 1e-5, False))
```
JiabinYang: `all close` is enough why not `array_equal` ?

paddle/fluid/operators/fill_op.cu.cc
```diff
@@ -0,0 +1,20 @@
+/* Copyright (c) 2019 PaddlePaddle Authors. All Rights Reserved.
```
JiabinYang, line 1: why not .cu?

JiabinYang (commented):
some comments

python/paddle/fluid/tests/unittests/test_fill_op.py
JepsonWong: done

paddle/fluid/operators/fill_op.h
JepsonWong: done

JiabinYang (approved):
LGTM

PR #19727, opened 2019-09-10
Refine math_op_patch.py when the other_var is a scalar

sneaxiy:
If the other operand in `elementwise_add/sub/mul/div` method of `Variable` is a scalar, the operation can be simplified by using `scale` op. This would save memory usages and speed up the model.


zhhsplendid (approved):
LGTM

PR #19730, opened 2019-09-10
paddle::framework::vectorize() templatization [PART4]

luotao1:
Suggested in https://github.com/PaddlePaddle/Paddle/pull/19510 changes to `paddle::framework::vectorize()` function. 

- This PR remove `std::vector<int> vectorize2int(const DDim& ddim)`  in `ddim.h`
- When updating all `framework::vectorize2int` to `framework::vectorize<int>`, we find that `warpctc_cudnn_op.cu.cc` uses the old `framework::vectorize2int` interface. Discussed with @wanghaoshuang, since warpctc-cudnn implementation always has accuracy-diff in both python2.7 and python3.5, and it will not be maintained later, this PR remove it.


sneaxiy (approved):
LGTM.

wanghaoshuang (approved):

heavengate (approved):

PR #19749, opened 2019-09-10
Delete local execution scopes

chengduoZH:
Add scope_buffered_monitor to monitored the local execution scopes.


sneaxiy (commented):
As a user, using a flag to control the memory leak sizes has no differences with setting `exec_strategy.num_iteration_per_drop_scope = xxx`. It is also a burden for users.

sneaxiy (commented):
BTW, for large models which occupy lots of GPU memory, limitation of 1GB is even too large.

chengduoZH:
> As a user, using a flag to control the memory leak sizes has no differences with setting exec_strategy.num_iteration_per_drop_scope = xxx. It is also a burden for users.

It is different in fact,  this flag focuses on the sub-scope of local execution scope, but not variables of local execution scope, those variables can be collected by GC now. For some models, if the operators in this graph(converted from a model) are run in different places, there have sub-scope in local execution scope, if not, there may not have sub-scope in local execution scope.

paddle/fluid/framework/details/scope_buffered_monitor.cc
```diff
@@ -0,0 +1,202 @@
...
+  for (size_t scope_id = 0; scope_id < local_exec_scopes_.size(); ++scope_id) {
+    for (auto &scope : post_local_exec_scopes_.at(scope_id)) {
+      if (!pre_local_exec_scopes_.at(scope_id).count(scope)) {
+        incr_local_exec_scopes.at(scope_id).insert(scope);
+      }
+    }
+
+    if (incr_local_exec_scopes.at(scope_id).size() &&
+        FLAGS_local_exe_sub_scope_limit > 0) {
+      LOG_FIRST_N(WARNING, 1)
```
sneaxiy: Do we need to add this warning to users? Users may not know what it means.

paddle/fluid/platform/flags.cc
```diff
@@ -427,3 +427,17 @@ DEFINE_uint64(reallocate_gpu_memory_in_mb, 0ul,
...
+ * Scope related FLAG
+ * Name: local_exe_sub_scope_limit
+ * Since Version: 1.6.0
+ * Value Range: double, default=256 (MB)
+ * Example:
+ * Note:
+ */
+DEFINE_double(local_exe_sub_scope_limit, 256.0,  // MBytes
+              "The memory up limit of sub-scopes of local execution scope for "
+              "each GPUPlace. If you don't need to limit the memory, "
```
sneaxiy: How about "each CUDAPlace"?

sneaxiy (dismissed):
LGTM for `ScopeBufferedMonitor`.

lanxianghit:
FLAG的用户使用说明文档在哪里？

chengduoZH:
@lanxianghit 
> FLAG的用户使用说明文档在哪里？

这里 https://github.com/PaddlePaddle/Paddle/pull/19749/files#diff-441c4a3926cd9492a8c825c8e718e9b3R431

paddle/fluid/platform/flags.cc
chengduoZH: Ok, I will fix this in next PR.

paddle/fluid/framework/details/scope_buffered_monitor.cc
chengduoZH: Ok, I will fix this in next PR.

lanxianghit (approved):
LGTM

paddle/fluid/platform/flags.cc
chengduoZH: Done

paddle/fluid/framework/details/scope_buffered_monitor.cc
chengduoZH: Done.

PR #19753, opened 2019-09-11
remove trainer desc test in windows temporarily

Thunderbrook:
the unitest of trainer desc will random fail in some env, so this commit remove this unitest temporarily.
image


luotao1:
不应该在cmake里写一下？
```
if(WIN32)
    list(REMOVE_ITEM TEST_OPS test_trainer_desc)
endif()
```
https://github.com/PaddlePaddle/Paddle/blob/bb4f8dee83e2e7d0ff84bada0a1bbfc330fe541d/python/paddle/fluid/tests/unittests/CMakeLists.txt#L38-L40

luotao1 (approved):

PR #19757, opened 2019-09-11
Feature/auto prune in dygraph

JiabinYang:
This PR will support auto prune in dygraph mode


python/paddle/fluid/framework.py
```diff
@@ -456,12 +456,13 @@ def __init__(self,
         if in_dygraph_mode():
             # record vars in tracer rather than blocks
             self._ivar = kwargs.get("ivar", None)
+            self.stop_gradient_ = kwargs.get("stop_gradient", True)
```
chengduoZH, line 459: `stop_gradient` is in the argument list and the default value is `False`, so the `kwargs.get("stop_gradient", True)` is always `True` here.
So user can not set `stop_gradient` when creating `Variable`, is this your intention?

JiabinYang: yes, user should not set stop_gradient while creating Variable，since it's not legal for user to create Variable with its constructor

PR #19772, opened 2019-09-11
Add prune_backward function to cover complicated test_program.clone situation

chenwhql:
To solve the test_program.clone position sensitive program, I call `Program._prune` in `Program.clone` when user clone main_program after `Optimizer.minimize` for test, but `Program._prune(loss_op)` cannot cover the situation that user append metric op(e.g. accuracy) after pretiction result. 

So I write `prune_forward` function to solve this problem, it is more efficient than using original `_prune` function.

related PR: https://github.com/PaddlePaddle/Paddle/pull/19397

### DIFF
--- a/paddle/fluid/framework/prune.cc
+++ b/paddle/fluid/framework/prune.cc
@@ -17,10 +17,17 @@ limitations under the License. */
 #include <glog/logging.h>
 
 #include <algorithm>
+#include <memory>
 #include <set>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
+
+#include "paddle/fluid/framework/block_desc.h"
+#include "paddle/fluid/framework/op_desc.h"
+#include "paddle/fluid/framework/op_proto_maker.h"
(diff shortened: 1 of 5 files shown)


paddle/fluid/framework/prune.cc
```diff
@@ -61,14 +90,32 @@ bool HasSubBlock(const proto::OpDesc& op_desc) {
...
+void AppendOpInputVarNames(const proto::OpDesc& op_desc,
+                           std::unordered_set<std::string>* vars_set) {
+  for (auto& var : op_desc.inputs()) {
+    for (auto& arg : var.arguments()) {
+      vars_set->emplace(arg);
+    }
+  }
+}
+
+void AppendOpOutputputVarNames(const proto::OpDesc& op_desc,
```
chengduoZH: `AppendOpOutputputVarNames` -> `AppendOpOutputVarNames`

chenwhql: Done

chengduoZH (approved):

PR #19780, opened 2019-09-12
Reduce default value of maximum cudnn workspace size

sneaxiy:
The default value of maximum cudnn workspace size is 4096MB, which is too large for most models. This PR reduces the default value to be 512MB. This makes maximum batch size of ResNet50/ResNet101 models increase about 10%.


zhhsplendid (approved):
LGTM

PR #19794, opened 2019-09-12
Fix retry allocator bug

sneaxiy:
The previous implementation of `RetryAllocator` would not clear the error flag of cuda, which is returned by `cudaGetLastError()`. It would make the following calls of `cudaGetLastError() == cudaSuccess`  failure, even though allocation retry is successful. This PR clears the error flag of cuda by calling `cudaGetLastError()` before throwing `BadAlloc` exception.


zhhsplendid (approved):
LGTM

PR #19800, opened 2019-09-14
fix softmax precision error axis!=-1.

heavengate:
**fix precision bugs in softmax when axis!=-1**
shifted_logits calculate max values only along axis.


JesseyXujin (approved):

kuke (approved):
LGTM

PR #19801, opened 2019-09-14
Add MKLDNNhandler templatized class

grygielski:
This PR is next in series of refactoring, templatizing and preparing code for easier MKLDNN v1.0 update.

It adds basic version of templatized MKLDNNhandler class that allows to remove common code from most of mkldnn operator handlers. MKLDNNhandler class is now duplicated by its template version MKLDNNhanlderT but it will be deleted once every operator will be refactored.

Future work: Move more functionality from child classes to MKDLNNhandler class. Use MKLDNN query function to get memory primitive descriptors, this will generalize interface and allow to move more functions to base class.


jczaja (approved):
LGTM

PR #19805, opened 2019-09-16
Fix warning info of build_strategy

chengduoZH:
- [x] Polish the warning info of build_strategy.
- [x] Fix bug of all_reduce_deps_pass.

### DIFF
--- a/paddle/fluid/framework/ir/multi_devices_graph_pass/all_reduce_deps_pass.cc
+++ b/paddle/fluid/framework/ir/multi_devices_graph_pass/all_reduce_deps_pass.cc
@@ -131,7 +131,7 @@ class AllReduceDepsPass : public ir::Pass {
            auto right_in_vars =
                details::DynamicCast<details::VarHandle>(right->Inputs());
            PADDLE_ENFORCE_GT(left_in_vars.size(), 0);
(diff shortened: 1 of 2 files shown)


paddle/fluid/framework/ir/multi_devices_graph_pass/all_reduce_deps_pass.cc
```diff
@@ -131,7 +131,7 @@ class AllReduceDepsPass : public ir::Pass {
            auto right_in_vars =
                details::DynamicCast<details::VarHandle>(right->Inputs());
            PADDLE_ENFORCE_GT(left_in_vars.size(), 0);
-           PADDLE_ENFORCE_EQ(left_in_vars.size(), right_in_vars.size());
```
chengduoZH, line 134: The size of `left_in_vars ` and `right_in_vars ` may be different when the `fuse_all_reduce_ops` is enabled.

gongweibao: I was surprised that we didn’t encounter this bug before.

gongweibao (approved):
LGTM

PR #19814, opened 2019-09-16
Refine executor_gc_helper codes to follow google coding style

sneaxiy:
`OperatorBase` would not be changed inside `executor_gc_helper.cc`, so we should change `OperatorBase *` to `const OperatorBase *` to follow [google coding style](https://google.github.io/styleguide/cppguide.html#Reference_Arguments).


zhhsplendid (approved):
LGTM

PR #19821, opened 2019-09-16
Fix example error of Variable and Operator

chengduoZH:
Fix example error of Variable and Operator.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -653,18 +653,6 @@ class Variable(object):
 
     __repr__ = __str__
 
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/framework.py
```diff
@@ -642,18 +653,6 @@ def to_string(self, throw_on_error, with_details=False):
 
     __repr__ = __str__
 
-    def set_desc(self, input):
```
chengduoZH, line 645: This API is not used, and it is not in API.spec.

Xreki (approved):
LGTM

PR #19824, opened 2019-09-16
Support dispensable student_loss in PaddleSlim distillation

baiyfbupt:
Make 'loss' dispensable in PaddleSlim distillation process.


wanghaoshuang (approved):
LGTM

PR #19825, opened 2019-09-16
Fix BUGS: paddle-TRT repeatedly sets weight_map and overdeletes repetitive_params

cryoco:
-- fixed 2 bugs of paddle TensorRT, which happens when 2 ops share the same weight parameter:
1. weight_map might be set with the same key multiple times, which causes function GetWeightCPUData() to throw an error(happens in retinanet);
image
2. some parameters are marked as repetitive and deleted after subgraph pass to reduce GPU memory, but if some of them are shared by multiple ops, they might still be needed in fluid, which causes fatal error(happens in cascade_rcnn and faster_rcnn).
image


NHZlX (approved):
LGTM

PR #19826, opened 2019-09-16
Enhance OpTest to support double grad inplace check

zhiqiu:
This PR updates OpTest to support double grad `inplace` check.
After this, OpTest will check `inplace`  for forward, grad and double grad  op automatically if op has registered grad and double grad.


sneaxiy (approved):
Excellent!

PR #19827, opened 2019-09-16
Fix gc bug in controlflow ops

sneaxiy:
Some control-flow ops may be nested inside other control-flow ops, so that they cannot be found in block 0 when gc is enabled. This PR traverse all blocks to find out all control-flow ops to avoid this bug.


zhhsplendid (approved):
LGTM

PR #19353, opened 2019-08-22
Python infer api update and add unit test

FrostML:
python api support numpy and add unit test. 
moreover, repair the test fail occurred in test_slim_int8_googlenet and test_slim_int8_mobilenet
all updates are base on PR:https://github.com/PaddlePaddle/Paddle/pull/19009

PR #19831, opened 2019-09-17
Cherry-pick #18817 and #19353. Python inference api update and add unittest

FrostML:
Cherry-pick #18817 and #19353. Python inference api update and add unittest


fc500110 (approved):
LGTM

PR #19833, opened 2019-09-17
Fix leaky_relu op when alpha is zero

sneaxiy:
When alpha = 0, the implementation of LeakyRelu is wrong, since it uses `out >= 0` to judge whether `x >= 0`. This PR fixes it by using `out > 0` instead of `out >= 0`.


zhhsplendid (approved):
LGTM

PR #19835, opened 2019-09-17
[Bug fix] Disable memory reuse on feeded variables

sneaxiy:
It is better that memory reuse passes do not reuse the variable that users feeds, because users may feed them to another network, which may cause calculation errors. This PR disables memory reuse on feeded variables.

For example:

```Python
tensor = fluid.LoDTensor()
... 

# If tensor is reused by other variables, its data would be changed.
exe.run(program1, feed={'image': tensor})

# The following training may be wrong if tensor is reused in the previous network.
# However, users may not know tensor has been changed.
exe.run(program2, feed={'image': tensor})
```

Although memory reuse is disabled in feeded variables, garbage collection still works on them.

### DIFF
--- a/paddle/fluid/framework/ir/memory_optimize_pass/memory_optimization_var_info.h
+++ b/paddle/fluid/framework/ir/memory_optimize_pass/memory_optimization_var_info.h
@@ -35,7 +35,11 @@ class MemOptVarInfo {
     return ref_cnt_ == 1 || (runtime_ref_cnt_.fetch_sub(1) == 1);
   }
 
-  void ResetRuntimeRefCnt() { runtime_ref_cnt_ = ref_cnt_; }
+  void ResetRuntimeRefCnt() {
+    if (ref_cnt_ != 1) {
+      runtime_ref_cnt_ = ref_cnt_;
+    }
+  }
 
   void SetRefCnt(size_t ref_cnt) {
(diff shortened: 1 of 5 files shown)


paddle/fluid/framework/ir/memory_optimize_pass/memory_optimization_var_info.h
```diff
@@ -44,17 +48,30 @@ class MemOptVarInfo {
...
+
+  bool IsSkippedMemoryReuse() const {
+    return skip_memory_reuse_ || skip_all_memory_optimization_;
+  }
 
   const std::string &Name() const { return name_; }
 
  private:
   std::string name_;
   size_t ref_cnt_;
```
zhhsplendid, line 85: Please add comment for the difference of these two variables.

zhhsplendid: I meant, ref_cnt_ and runtime_ref_cnt_

sneaxiy: Done.

zhhsplendid: Great!

zhhsplendid (approved):
LGTM

PR #19838, opened 2019-09-17
warning when user save a inference model which contains auc op 

DogHeadMoon:
Add warning  when user save a inference model which contains auc op
related to #18826

When saving a inference model, if there is a auc op in this model, paddle
 should remind user to set auc states to zeros.

### DIFF
--- a/python/paddle/fluid/io.py
+++ b/python/paddle/fluid/io.py
@@ -1031,6 +1031,15 @@ def save_inference_model(dirname,
 
     main_program = _get_valid_program(main_program)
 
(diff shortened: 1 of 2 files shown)


luotao1:
描述里不要贴内部卡片编号，请将卡片里的内容直接复制到描述中。

DogHeadMoon:
> 描述里不要贴内部卡片编号，请将卡片里的内容直接复制到描述中。

已更正

phlrain (approved):
LGTM

PR #19843, opened 2019-09-18
Refine reallocate of cudnn workspace

sneaxiy:
Cudnn workspace can be freed before re-allocation to save memory usages of models.


zhhsplendid (approved):
LGTM

PR #19845, opened 2019-09-18
Add lock to cudnn handle calls

sneaxiy:
`cudnnHandle_t` is not thread-safe. Please see https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html and https://github.com/pytorch/pytorch/blob/master/caffe2/core/cudnn_wrappers.h. 

This PR adds lock to cudnn function calls.


zhhsplendid (approved):
LGTM

zhhsplendid:
Please also run our CE benchmark to see whether models work correctly before merging into develop.

PR #19847, opened 2019-09-18
Refactor Inplace test in OpTest and check grad inplace even if forward has no inplace

zhiqiu:
Before this PR, OpTest stops  `inplace` checking for grad op when forward op has no `inplace`, and stops `inplace` checking for double_grad op when grad op has no `inplace` (e.g., elementwise_div_grad_grad),  and so on.

This PR fixes that problem and refactors `inplace test` code for better modularity and understanding.


sneaxiy (dismissed):
LGTM.

sneaxiy (approved):
Excellent!

PR #19857, opened 2019-09-18
Avoid treating broadcast as initialization operation

ForFishes:
When the mode is collective mode, an error of two initializations will occur. The broadcast should be treated as a non-initialized operation here. Think of broadcast as a special case here.
MacHi 2019-09-18 17-31-10

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_avoid_broadcast.py
+++ b/python/paddle/fluid/tests/unittests/test_avoid_broadcast.py
@@ -1,0 +1,50 @@
+#   Copyright (c) 2019 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/tests/unittests/test_avoid_broadcast.py
```diff
@@ -0,0 +1,50 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+import unittest
+import paddle.fluid as fluid
+
+
+class Test_Avoid_Broadcast(unittest.TestCase):
```
gavin1332: the class name is illegal

gavin1332 (changes requested):
The class name of TestAvoidBroadcast is illegal

gavin1332 (approved):
LGTM

luotao1 (approved):
LGTM for framework.py

PR #19863, opened 2019-09-18
Change fleet._origin_program to be the program before transpiled.

gongweibao:
Now, fleet._origin_program is a transpiled program and it's a bug. It should be the program which is not changed.


gavin1332 (approved):
LGTM

PR #19864, opened 2019-09-18
Add TRT input shape check between model and runtime

cryoco:
Added shape check (the runtime input shape set by user should be consistent with the input shape in model description) for TRT engine, because TRT 5 or lower version doesn't support dynamic shapes of inputs.

Before this, when a user set the input shapes wrong, TRT engine wouldn't throw an error properly, but just went down while running, showing that some ops' dims are not matched, which might cause confusion to users.

Now the error message can be thrown correctly, guiding users to change their input shapes to be consistent with the model.

### DIFF
--- a/paddle/fluid/operators/tensorrt/tensorrt_engine_op.h
+++ b/paddle/fluid/operators/tensorrt/tensorrt_engine_op.h
@@ -212,9 +212,11 @@ class TensorRTEngineOp : public framework::OperatorBase {
                                                i_shape.end());
         std::vector<int64_t> runtime_input_shape(t_shape.begin() + 1,
                                                  t_shape.end());
-        PADDLE_ENFORCE_EQ(model_input_shape == model_input_shape, true,
-                          "input shapes not consistent with the model. TRT5 "
-                          "does not support dynamic shapes.");
+        PADDLE_ENFORCE_EQ(model_input_shape == runtime_input_shape, true,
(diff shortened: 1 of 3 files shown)


paddle/fluid/operators/tensorrt/tensorrt_engine_op.h
```diff
@@ -203,7 +204,19 @@ class TensorRTEngineOp : public framework::OperatorBase {
...
-      auto t_shape = framework::vectorize(t.dims());
+      auto t_shape = framework::vectorize<int64_t>(t.dims());
+      // check if the input shapes are consistent with model.
+      if (HasAttr(x + "_shape")) {
+        std::vector<int64_t> i_shape = Attr<std::vector<int64_t>>(x + "_shape");
+        std::vector<int64_t> model_input_shape(i_shape.begin() + 1,
+                                               i_shape.end());
+        std::vector<int64_t> runtime_input_shape(t_shape.begin() + 1,
+                                                 t_shape.end());
+        PADDLE_ENFORCE_EQ(model_input_shape == model_input_shape, true,
```
NHZlX: model_input_shape == runtime_input_shape

cryoco: Done, Thanks.

NHZlX (approved):
LGTM

PR #19865, opened 2019-09-18
Set states of recurrent op as dependent vars in prune

zhhsplendid:
This PR will fix the save/load inference model problem of RNN models.

The reason of the bug is that save_inferenc_model will prune OPs that doesn't contribute to Output. But in recurrent_op, States are not Output, OPs refers States will be pruned. 

This fix adds States of recurrent_op as dependent var so that OPs referring States won't be pruned.

### DIFF
--- a/paddle/fluid/framework/prune.cc
+++ b/paddle/fluid/framework/prune.cc
@@ -35,6 +35,10 @@ namespace framework {
 const char kFeedOpType[] = "feed";
 const char kFetchOpType[] = "fetch";
 
+const char kRecurrent[] = "recurrent";
+const char kStates[] = "states";
+const char kExStates[] = "ex_states";
+
 bool HasDependentInputVar(
     const proto::OpDesc& op_desc,
     const std::unordered_set<std::string>& dependent_vars) {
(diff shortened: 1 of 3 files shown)


paddle/fluid/framework/prune.cc
```diff
@@ -189,6 +194,20 @@ void prune_impl(const proto::ProgramDesc& input, proto::ProgramDesc* output,
             }
           }
         }
+
+        // Recurrent op's states are also dependent vars
+        if (op->type() == kRecurrent) {
```
sneaxiy, line 199: A bit hard code. Maybe we should find a better way later after this PR merges.

zhhsplendid: I don't think it is hard code here ... because due to early design, States are Attrs of Recurrent OP. There is no way I can avoid adding States as dependent vars. This is the only way I can fix the bug on current design framework.

As we have discussed, changing Attrs to Inputs/Outputs is not encouraged in PaddlePaddle and it may bring new bugs. We can rewrite RNNs in the future and discard early design, but now it's too risky for release 1.6 ...

In addition, I'm also suffering from the early design and TU CAO it.

sneaxiy (commented):
LGTM.

NHZlX (approved):
LGTM

PR #19866, opened 2019-09-18
Fix SplitLoDTensor bug when batch_size = 0

zhiqiu:
Original SplitLodTensor raises FPE (Float Point Exception) when the batch_size of given LoDTensor is 0.
This PR fixes that bug.


sneaxiy (approved):
LGTM.

PR #19867, opened 2019-09-18
Add template functions for Acquire primitive/primitive_desc

grygielski:
This PR is next in series of refactoring, templatizing and preparing code for easier MKLDNN v1.0 update.

It adds variadic template version of acquire primitive and primitive descriptor function for either forward and backward operators. This will allow deleting ~100 lines for every operator during refactoring and lead to easier implementation for further operators. After replacing mkldnn_format to memory descriptor in Tensor, code will be even cleaner (we will get rid of creating memory descriptors in handler constructor).

Future work: Test possibility of using MKLDNN queries instead of primitive_desc functions to get memory descriptors (this will lead to more generalized code). Further operators refactoring.


jczaja (approved):
LGTM

PR #19868, opened 2019-09-18
Fix readers bug

sneaxiy:
This PR works for:
- Fix bug of `paddle.dataset.cifar.train10` when cycle=True.
- Fix hang bug when `multiprocess_reader` raises exceptions.


zhhsplendid (approved):
LGTM

PR #19870, opened 2019-09-18
Fix download.py empty line

sneaxiy:
`paddle.dataset.xxx` apis would print an extra blank line even if the file has existed in local machine. This PR removes this blank line.


zhhsplendid (approved):
LGTM

PR #19871, opened 2019-09-19
remove the useless warning for user to avoid confuse test=develop

wopeizl:
用户提出warning提示：The number of graph should be only one, but the current graph has 4 sub_graphs. 不重要，建议删除。


lanxianghit:
请添加PR描述

lanxianghit (approved):
LGTM

PR #19889, opened 2019-09-19
Fix reduce and broadcast to avoid multi-stream error

sneaxiy:
Some operators would release memories before gpu kernel ends. NCCL Reduce and Broadcast works on another stream different from operators on one device. 

Currently, the execution order is: wait input generated  -> allocate memory -> call ncclReduce/ncclBroadcast.

But the allocated memory may be used in the compute CUDA stream.

Therefore, the right execution order should be: allocate memory -> wait input generated -> call ncclReduce/ncclBroadcast.


chengduoZH (approved):
LGTM

PR #19894, opened 2019-09-19
[PaddleSlim]Enhence compressor api in PaddleSlim

wanghaoshuang:
1. Support customize eval function instead of eval program.
2. Fix loading checkpoint in quantization strategy.
3. Support saving eval model when saving checkpoint.
test=develop


python/paddle/fluid/contrib/slim/core/compressor.py
```diff
@@ -518,7 +557,7 @@ def run(self):
             else:
                 context.optimize_graph = context.train_graph
 
-        context, self.strategies = self._load_checkpoint(context)
+        context, strategies = self._load_checkpoint(context)
```
wanghaoshuang: ?

PR #19907, opened 2019-09-20
Add support for other axes in MKLDNN softmax op

grygielski:
This PR adds support for other axes in MKLDNN softmax op. These changes will allow to use MKLDNN softmax op in wider range of models.


jczaja (approved):
LGTM

PR #19918, opened 2019-09-21
move tree_conv to fluid.contrib.layers

luotao1:
**before：**

```python
import paddle.fluid as fluid

tree_conv = fluid.layers.tree_conv
```

**after：**

```python
import paddle.fluid as fluid

tree_conv = fluid.contrib.layers.tree_conv
````


lanxianghit (approved):
LGTM

heavengate (approved):
LG API

PR #19923, opened 2019-09-23
Add py_reader combination unittest

sneaxiy:
Add a unittest to test whether two PyReader objects can be used through `zip`.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_py_reader_combination.py
+++ b/python/paddle/fluid/tests/unittests/test_py_reader_combination.py
@@ -1,0 +1,100 @@
+# Copyright (c) 2019 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import paddle.fluid as fluid
+import unittest
+import numpy as np
+
+
+class TestPyReaderCombination(unittest.TestCase):
+    def setUp(self):
+        self.n1 = 10
+        self.n2 = 20
+        self.batch_size = 2
+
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/tests/unittests/test_py_reader_combination.py
```diff
@@ -0,0 +1,100 @@
...
+class TestPyReaderCombination(unittest.TestCase):
+    def setUp(self):
+        self.n1 = 10
+        self.n2 = 20
+        self.batch_size = 2
+
+    def create_reader(self, batch_num):
+        def __impl__():
+            for _ in range(batch_num):
+                image = np.random.uniform(
```
zhhsplendid, line 29: Set fixed np.random.seed at the head of the file. Because it helps SWEs to debug when unit test error is reproducible.

python/paddle/fluid/tests/unittests/test_py_reader_combination.py
```diff
@@ -0,0 +1,100 @@
...
+                    low=-1, high=1, size=[batch_num, 784]).astype('float32')
+                label = np.random.random_integers(
+                    low=0, high=9, size=[batch_num, 1]).astype('int64')
+                yield image, label
+
+        return __impl__
+
+    def assertFeedVarEqual(self, d1, d2):
```
zhhsplendid: Not good for two input parameters with different types having similar name.

Two optional suggestions:

1. rename d1, d2, like np_data, tensor_data.

2. keep the name d1, d2, but they have to be both numpy data. Do the "np.array" out of "assertFeedVarEqual"

python/paddle/fluid/tests/unittests/test_py_reader_combination.py
sneaxiy: It is not necessary to fix random seed of numpy. Because we call `fluid.io.cache` to cache the reader data for the first time. If there is any error caused by numeric differences, it must be a bug.

python/paddle/fluid/tests/unittests/test_py_reader_combination.py
sneaxiy: Done.

python/paddle/fluid/tests/unittests/test_py_reader_combination.py
zhhsplendid: Discussed offline, the io.cache provides same function as fixing random numbers. So this PR LGTM.

zhhsplendid (approved):
LGTM

PR #19926, opened 2019-09-23
refine ctc align op with padding 

slf12:
refine ctc align op with padding, when input is padding, add input length which is length is every input sequence  and output length which is length of every output length

image
image
image


wanghaoshuang (approved):

kuke (approved):
LGTM

PR #19937, opened 2019-09-23
Fix huber loss op attr type bug

sneaxiy:
Op attribute cannot be a template argument. This PR fixes huber_loss bug.


zhhsplendid (approved):
LGTM

PR #19945, opened 2019-09-24
Fix allocator unittests

sneaxiy:
Do not know why, but `cudaMalloc` would cause seg fault when allocating -1UL on some machines.


zhhsplendid (approved):
LGTM

PR #19953, opened 2019-09-24
Fix cuda dev_ctx allocator cmake dependency

sneaxiy:
Move `CUDADeviceContextAllocator` codes to a header file and include it inside `device_context.cc` to solve cyclic dependency between `malloc` and `device_context`.


zhhsplendid:
I think we can find a more elegant solution in the future. But for 1.6, it fixes the random compilation error so I will approve this PR.

zhhsplendid (approved):
LGTM

sneaxiy:
@zhhsplendid Yes. I agree with you.

PR #19961, opened 2019-09-24
Polish multi-process warning info.

chengduoZH:
In multi-process training mode, if one process exit because of some error, other processes may be in hang status, at that time, if we rerun the training script, the process may print `worker: xx is not ready, will retry after 3 seconds...`. Users may be confusing about this and don't know how to deal with that.  In that situation, the user should kill those processes that in hang status manually, and this solution is added to the warning info.


JiabinYang (approved):
LGTM

PR #9017, opened 2018-03-13
Add fp16 mul op support and bind paddle fp16 to numpy fp16

kexinzhao:
fix #8816 
fix #9021

numpy float16 is internally represented as numpy.uint16. Hence, we create a binding via the helper of the uint16_t type.

PR #19964, opened 2019-09-24
Update Tensor.set() to support float16

zhiqiu:
Original implementation of Tensor.set() use template specialization to support different dtype, like int, float. It has some problems:

- It expose 24 APIs in total.

- Since float16 is not a builtin type in C++, Paddle uses uint16_t to 'represent' float16
in [PR9017](https://github.com/PaddlePaddle/Paddle/pull/9017). As a result, numpy.float16 is set to float32, and uint16 is set to float16.  This is not convenient and may cause errors if users don't know.
image

This PR did the followings.

- Registers `paddle::platform::float16` as `numpy.float16`, so users can feed float16 tensor directly, no need to view as uint16. For compatibility, keeps uint16-float16 hacking temporarily.

- Reduces the  number of set() API to 3, 

- Updates some unit tests (test_npair_loss_op.py, test_square_error_cost.py, test_mse_loss.py), which feed Variable to program.

### DIFF
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -400,55 +400,9 @@ PYBIND11_MODULE(core_noavx, m) {
              return reinterpret_cast<uintptr_t>(self.mutable_data(place, type));
            })
       .def("_clear", &Tensor::clear)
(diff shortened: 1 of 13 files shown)


sneaxiy (approved):
LGTM.

PR #19966, opened 2019-09-24
Disable fuse_all_optimizer_ops

chengduoZH:
Temporally disable fuse_all_optimizer_ops, because in some case, the parameter's gradients of the model are in different Place, for example, g0 and g1 are in CPU but g2 and g3 are in GPU0, this is because the generation op of g0 and g1 only have CPU kernel, in this case, the current implementation of fuse_all_optimizer_ops can not deal with it.


chenwhql (approved):
LGTM

PR #19967, opened 2019-09-24
Fix buddy_allocator_test

sneaxiy:
buddy_allocator_test would consume 10G GPU memory, which makes this ut fails in some machines with little GPU memory.


JiabinYang (approved):
LGTM

PR #19969, opened 2019-09-24
FIx C++ inference BUG: When open memory optim and enable trt subgraph at the same time, there is a bug

NHZlX:
1. When open memory optim and enable trt subgraph at the same time, there is a bug:

```
0924 07:44:34.093153  3990 naive_executor.cc:37] NaiveExecutor init with scope 0x272f1d0
---  skip [feed], feed -> image
terminate called after throwing an instance of 'paddle::platform::EnforceNotMet'
  what():  Attribute 'subgraph' is required! at [/paddle/paddle_anakin_many_linux2/paddle/fluid/framework/attribute.h:302]
PaddlePaddle Call Stacks: 
0       0x7fb070e7b201p std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > paddle::platform::GetTraceBackString<char const*>(char const*&&, char const*, int) + 1505
```

2. clean memory  optimize code


paddle/fluid/inference/analysis/argument.h
```diff
@@ -196,9 +196,7 @@ struct Argument {
 
   // Memory optimized related.
   DECL_ARGUMENT_FIELD(enable_memory_optim, EnableMemoryOptim, bool);
-  DECL_ARGUMENT_FIELD(static_memory_optim, StaticMemoryOptim, bool);
-  DECL_ARGUMENT_FIELD(static_memory_optim_force_update,
-                      StaticMemoryOptimForceUpdate, bool);
+
```
Superjomn, line 199: 在预测文档上也做下更新

Superjomn (approved):
LGTM

paddle/fluid/inference/analysis/argument.h
NHZlX: 好的

PR #19997, opened 2019-09-25
Remove constraint that last dimension is forced to be 1 in rank_loss

Aurelius84:
**before:**
+ The `shape` of `label`, `left`, `right` must be `[batch_size, 1]`. 
+ The last dimension of `shape` must be 1.

```python
import paddle.fluid as fluid
label = fluid.layers.data(name="label", shape=[5, 1], append_batch_size=False, dtype="float32")
left = fluid.layers.data(name="left", shape=[5, 1], append_batch_size=False, dtype="float32")
right = fluid.layers.data(name="right", shape=[5, 1], append_batch_size=False dtype="float32")

out = fluid.layers.rank_loss(label, left, right)
```

**after:**
+ The `shape` of `label`, `left`, `right` can  be `[batch_size, 1]` or `[batch_size]`
+ Modify `PADDLE_ENFORCE` into `PADDLE_ENFORCE_EQ/LT`

```python
import paddle.fluid as fluid
# shape can be [batch_size, 1] or [batch_size]
label = fluid.layers.data(name="label", shape=[5], append_batch_size=False, dtype="float32")
# shape can be [batch_size, 1] or [batch_size]
left = fluid.layers.data(name="left", shape=[5], append_batch_size=False, dtype="float32")
# shape can be [batch_size, 1] or [batch_size]
right = fluid.layers.data(name="right", shape=[5, 1], append_batch_size=False dtype="float32")

out = fluid.layers.rank_loss(label, left, right)
```

### DIFF
--- a/paddle/fluid/operators/rank_loss_op.cc
+++ b/paddle/fluid/operators/rank_loss_op.cc
@@ -37,14 +37,10 @@ class RankLossOp : public framework::OperatorWithKernel {
     auto label_dims = ctx->GetInputDim("Label");
     auto left_dims = ctx->GetInputDim("Left");
     auto right_dims = ctx->GetInputDim("Right");
-
-    PADDLE_ENFORCE_EQ(label_dims[0], left_dims[0],
-                      "The first dimension of Input(Label) and Input(Left) "
-                      "must have the same value.");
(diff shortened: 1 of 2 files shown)


phlrain (approved):
LGTM

PR #20008, opened 2019-09-25
Fix math_op_patch division between integers

sneaxiy:
The previous implementation of `a / 7` would be converted to `scale(a, 1.0/7)`. But inside `scale` op, `1.0/7` would be cast to be `0` when `a` is a Tensor of integers, which loses accuracy.

This PR is a hot fix of math_op_path division between integers.


zhhsplendid (approved):
LGTM

PR #20016, opened 2019-09-26
Add dtype for coalesce_tensor_op

chengduoZH:
Add dtype in coalesce_tensor_op, because the output data type may be different from the data type of the input tensors.


gongweibao (dismissed):
LGTM

gongweibao (approved):
LGTM

PR #19589, opened 2019-09-02
add feed_var_names to Prune interface

mapingshuo:
Description of the requirement：https://github.com/PaddlePaddle/Paddle/issues/19362

Usage of the paddle::framework:: Prune() interface:
```cpp
proto::ProgramDesc pruned_desc;
Prune(original_program_desc, feeded_var_names, &pruned_desc);
```
`feeded_var_names` are Variables that we feed data in the inference program.
`original_program_desc` is the original program desc before pruning.
`pruned_desc` is the pruned program desc.

Usage of save_inference_model() interface remains the same. 

详细中文文档
----------------------------------
本Pull Request修复了prune接口无法无法指定feed_var的bug，修改后，其python端接口的使用方法为：
```python
    pruned_program = program._prune(feeded_var_names, targets)
```
其中，`feeded_var_names`是起始Variables的名字；targets为终止Variables，或者产生终止Variables的Operators。一个简单的例子可以说明本接口的使用方法：

```python
import paddle.fluid as fluid

x = fluid.layers.data(name='x', shape=[2], dtype='float32')
label = fluid.layers.data(name="label", shape=[1], dtype="int64")

# define net here
y = fluid.layers.fc(input=[x], size=2, act="softmax")
loss = fluid.layers.cross_entropy(input=y, label=label)
loss = fluid.layers.mean(x=loss)

sgd = fluid.optimizer.SGD(learning_rate=0.01)
sgd.minimize(loss)

with open("main_program", "w") as f:
    f.write(str(fluid.default_main_program()))

pruned_program = fluid.default_main_program()._prune(
                        feeded_var_names=[y.name, label.name],
                        targets = [loss])

with open("pruned_main_program", "w") as f:
    f.write(str(pruned_program))    
    
```
打印结果显示，裁剪后的模型的Operators只剩下`cross_entropy2`和`mean`，其余无关Operators和Variables都被裁剪掉了。

PR #20022, opened 2019-09-26
add a new interface _prune_with_input

mapingshuo:
framework._prune() interface is modified by this PR: https://github.com/PaddlePaddle/Paddle/pull/19589. Some users gave feedback that the old usage as below does not work now:

```
new_prog = old_prog._prune(target_vars)
```

This PR adds a new API named _prune_with_input() and restore _prune() interface, that is, only use targets as it's input.


sneaxiy (approved):
LGTM for framework.py

PR #20024, opened 2019-09-26
Update API of paddle.fluid.data

zhhsplendid:
We found the both the old data and the new data layer doesn't support output type other than LoDTensor. So we remove 'type' parameter from the new paddle.fluid.data

image


sneaxiy (dismissed):
LGTM.

heavengate (dismissed):
LG API

sneaxiy (dismissed):

heavengate (approved):

wanghaoshuang (approved):

PR #20051, opened 2019-09-27
Change distribute tests to serial test=develop

gongweibao:
Distribute unit tests can't run parallelly since they need more CPU resource. One distributed unit test's resource may equal to 4 local unit test.


wopeizl (approved):
LTGM

PR #20076, opened 2019-09-29
Fix op_compatiable_compile_error

sneaxiy:
`op_compatible_info` should depend on `proto_desc`. Otherwise, compilation would be wrong.
http://ci.paddlepaddle.org/viewLog.html?tab=buildLog&buildTypeId=Paddle_PrCiPython35Night&buildId=178789

image


luotao1 (approved):
LGTM

PR #20108, opened 2019-09-29
RecomputeOptimizer: rm unused ckpt and sort ckpt

mapingshuo:
RecomputeOptimizer has some bugs, this PR fixes them:

- some checkpoints are not used in the graph and should not be a checkpoint (This PR remove them from checkpoints list)

- If the checkpoints are not in order (the order they are generated), RecomputeOptimizer won't work correctly. (This PR sorts them)


guru4elephant (approved):
LGTM

sneaxiy (approved):

PR #20116, opened 2019-09-29
mv two function in conv op for good code style

liym27:
1. Delete PadFuntion, include padding.h instead. 

2.move function(IsSymmetricPadding) from conv_cudnn_op.cu to padding.h


lanxianghit (approved):
LGTM

PR #20127, opened 2019-09-30
refine sequence_softmax grad maker

sneaxiy:
SequenceSoftmaxGradOp does not need data buffer of `X`. This PR refines the memory of this op.


zhhsplendid (approved):
LGTM

PR #20167, opened 2019-10-05
Add Multihead matmul pass

zhaoyuchen2018:
multihead matmul pass for ernie opt
fuse some nodes together，improve performance


wangchaochaohu (approved):
LGTM

PR #20174, opened 2019-10-07
[Bug-fix][1.6] Improve QAT accuracy

Sand3r-:
Provides a fix for accuracy drop obtained with QAT models by enabling uint8 variable type on conv-relu outputs.
Also enables executing VGG16 and VGG19 by:
* Adding additional way of fetching scales (by associating tensor with its tensor scale by the *.scale suffix)
* Fixing bug with empty list of disabled op ids
Restores fake-quant mul to achieve best accuracy.

### Resnet50
|                 | top1 acc | top 5 acc |
|-----------------|----------|-----------|
| Original QAT    | 0.7655   | 0.9304    |
| Transformed QAT | 0.7653   | 0.9298    |
| Diff            | 0.0002   | 0.0006    |

### Mobilenet V1
|                 | top1 acc | top 5 acc |
|-----------------|----------|-----------|
| Original QAT    | 0.7077   | 0.8954    |
| Transformed QAT | 0.7076   | 0.8943    |
| Diff            | 0.0001   | 0.0011    |

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
+++ b/python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
@@ -381,13 +381,13 @@ class FakeQAT2MkldnnINT8PerfPass(object):
                 next_op = op_out.outputs[0]
                 if next_op.name() not in self._mul_ops:
                     self._remove_fake_quantize(graph, op)
-                else:
-                    quant_op = self._transform_to_quantize_mkldnn(graph, op)
-                    self._transform_to_mul_mkldnn(graph, next_op, quant_op)
 
         for op in graph.all_op_nodes():
             if op.name() in self._fake_dequantize_types:
-                self._remove_fake_dequantize(graph, op)
+                op_in = graph._find_node_by_name(op.inputs, op.input("X")[0])
+                prev_op = op_in.inputs[0]
+                if prev_op.name() not in self._mul_ops:
+                    self._remove_fake_dequantize(graph, op)
         return graph
 
     def _remove_fake_quantize(self, graph, op):
--- a/python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
+++ b/python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
@@ -378,13 +378,13 @@ class FakeQAT2MkldnnINT8PerfPass(object):
                 next_op = op_out.outputs[0]
                 if next_op.name() not in self._mul_ops:
                     self._remove_fake_quantize(graph, op)
+                else:
+                    quant_op = self._transform_to_quantize_mkldnn(graph, op)
+                    self._transform_to_mul_mkldnn(graph, next_op, quant_op)
 
         for op in graph.all_op_nodes():
             if op.name() in self._fake_dequantize_types:
-                op_in = graph._find_node_by_name(op.inputs, op.input("X")[0])
-                prev_op = op_in.inputs[0]
-                if prev_op.name() not in self._mul_ops:
-                    self._remove_fake_dequantize(graph, op)
+                self._remove_fake_dequantize(graph, op)
         return graph
 
     def _remove_fake_quantize(self, graph, op):
--- a/python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
+++ b/python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
@@ -470,9 +470,8 @@ class FakeQAT2MkldnnINT8PerfPass(object):
     def _apply_pass(self, graph, pass_name, attrs=None, attr_values=None):
         ir_pass = core.get_pass(pass_name)
         cpp_graph = graph.graph
-        if cpp_graph.has('__param_scope__'):
-            cpp_graph.erase('__param_scope__')
-        cpp_graph.set_not_owned('__param_scope__', self._scope)
+        if not cpp_graph.has('__param_scope__'):
+            cpp_graph.set_not_owned('__param_scope__', self._scope)
         if attrs:
             assert attr_values and len(attrs) == len(
                 attr_values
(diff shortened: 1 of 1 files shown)


Sand3r-:
For some reason the CI fails on `test_parallel_executor_seresnext_with_reduce_gpu` due to accuracy diff. I'm not sure if there is anything related to it, what has been proposed in this PR.

luotao1:
Please create PR to develop first, and after merge it to develop, then cherry-pick to release/1.6.

luotao1:
You can ignore the `test_parallel_executor_seresnext_with_reduce_gpu` random fail.

Sand3r-:
@luotao1 Thank you for the information. I have rebased on top of develop, and changed the target branch in this PR to it.

luotao1:
Please update the code format to pass the travis_ci

wzzju:
> Provides a fix for accuracy drop obtained with QAT models by enabling uint8 variable type on conv-relu outputs.
> Also enables executing VGG16 and VGG19 by:
> 
> * Adding additional way of fetching scales (by associating tensor with its tensor scale by the *.scale suffix)
> * Fixing bug with empty list of disabled op ids
>   Restores fake-quant mul to achieve best accuracy.
> 
> ## Accuracy comparison
> ### Resnet50
> top1 acc	top 5 acc
> Original QAT	0.7655	0.9304
> Transformed QAT	0.7653	0.9298
> Diff	0.0002	0.0006
> ### Mobilenet V1
> top1 acc	top 5 acc
> Original QAT	0.7077	0.8954
> Transformed QAT	0.7076	0.8943
> Diff	0.0001	0.0011

1. What is the original bug? Could you please give more information about this bug?
2. How do you resolve this bug? Could you please give more details?
3. Could you please explain what is `enabling uint8 variable type on conv-relu outputs` ? Because QAT doesn't provide the uint8 scale, how do you enable uint8 variable type on conv-relu outputs?

Sand3r-:
@wzzju 
> What is the original bug? Could you please give more information about this bug?

Certainly. Based on parameters such as `fuse_residual_connection` and `fuse_activation` int8 convolution kernel makes a decision what should the output datatype be. Since in many cases 'relu' activation was fused with convolution, the output was set to be uint8, however the signed scale has been assigned to it.

> How do you resolve this bug? Could you please give more details?

To resolve this bug, I have changed the information about the scale in `self._var_quant_scales` map that the scale provided to the `cpu_quantize_pass` should be rescaled to work with uint8 data. This is accomplished by changing the first value of the pair (bool unsigned, Tensor scale) to true, as done in line https://github.com/PaddlePaddle/Paddle/pull/20174/files#diff-eba5e4d0f5420f0702b898665a44dc56R571

To shed more light on  how this done: `cpu_quantize_pass` accepts the `VarQuantScales` map where the scales provided are "normalized". Based on the `unsigned` parameter they are then either multiplied by max_s8 (127) or max_u8 (255) so that later on they can be used in int8 computation kernels without further modifications.


> Could you please explain what is enabling uint8 variable type on conv-relu outputs ? Because QAT doesn't provide the uint8 scale, how do you enable uint8 variable type on conv-relu outputs?
As described above, the conv (with fused relu) output scales provided by QAT are appropriately rescaled.

python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
```diff
@@ -557,13 +560,16 @@ def _transform_to_quantize_mkldnn(self, graph, op_node):
...
-        op_node.set_attr("force_fp32_output", True)
+    def _update_conv_relu_scales(self, graph):
+        for op in graph.all_op_nodes():
+            if op.name() in self._conv_ops:
+                out_name = op.output("Output")[0]
+                if out_name in self._var_quant_scales and \
+                   op.op().attr("fuse_activation") == 'relu' and \
+                   op.op().attr("fuse_residual_connection") == False:
+                    _, tensor = self._var_quant_scales[out_name]
+                    self._var_quant_scales[out_name] = (True, tensor)
```
wzzju, line 571: Here, you just set the first flag as `True`, but don't change the scale value. Is it right?

Sand3r-: Yes, here I pass the normalized scale. The actual rescaling happens here:
https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc#L46-L47

`scale_to_one` is what we've passed to the quantizer.

python/paddle/fluid/contrib/slim/quantization/quantization_mkldnn_pass.py
```diff
@@ -350,6 +351,8 @@ def _gather_scales(self, graph):
                 use_unsigned_int = False
                 self._var_quant_scales[input_name] = (use_unsigned_int,
                                                       lod_tensor)
+                self._var_quant_scales[scale_name.replace(".scale", "")] = (
```
wzzju, line 354: What is this prepare for? Are `input_name` and `scale_name.replace(".scale", "")` different？

Sand3r-: I wish they were the same, but when we encounter the follwing case:
[op]->[fake_dequantize]->[fake_quantize]->[another_op] 
as it is in, for example, VGG16, the tensor that enters `fake_quantize_moving_average_abs_max` has a `.dequantized` suffix.
image
In this example `pool2d_1.tmp_0.dequantized` enters `fake_quantize_moving_average_abs_max`.

Sand3r-: The reason for which we need the clear names without any *.dequantized, or *.scale suffixes, is that after the graph is cleared off of all the fake-(de)quantize operators, the tensors with suffixes disappear, and `cpu_quantize_pass` cannot find appropriate scales for tensors with normal names, such as `pool2d_1.tmp_0`.

wzzju: A very clear explanation. :thumbsup:

wzzju (approved):
LGTM.

PR #20180, opened 2019-10-07
add dll to inference capi

FrostML:
The c-api is not support dynamic link library (dll). This PR adds dll compilation to inference c-api, and fixes the compilation error on Windows.


fc500110 (approved):
LGTM

PR #20185, opened 2019-10-08
Remove some DefaultGradOpDescMaker

sneaxiy:
`DefaultGradOpDescMaker` is not necessary for some ops, because some grad ops do not need some forward inputs or outputs as its input. This PR removes them to save memory usage of model.


luotao1 (approved):
LGTM!

PR #20209, opened 2019-10-08
fix windows compilation issue when compile with VS2015

liupluswei:
fix compilation issue when compile with VS2015


ceci3 (approved):
LGTM

PR #20241, opened 2019-10-08
[MKL-DNN] Added mkl-dnn cache clearing when creating Executor instance

jczaja:
This PR is introducing mkl-dnn cache clearing when Executor is created. This change will allow to reduce length of hashing keys , as MKL-DNN cache will become independent among unit tests. 
Executor constructor by default is clearing cache unless parametr clear_mkldnn_cache is set to false.
Disabling clearing of cache is used when LoadModel() is called. Reasoning behind this is that
for multithreaded scenario each thread may load model for its own prediction and at that situation
cache should be maintained (not-cleared).

**Performance impact**: Depending on how much (relatively towards all ops instances) mkl-dnn ops are used Fp32 mkl-dnn models after change are about upto ~2% faster

@grygielski , @Sand3r- Please review

### DIFF
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -91,15 +91,14 @@ ExecutorPrepareContext::~ExecutorPrepareContext() {
   VLOG(5) << "destroy ExecutorPrepareContext";
 }
 
-Executor::Executor(const platform::Place& place, bool clear_mkldnn_cache)
-    : place_(place), clear_mkldnn_cache_(clear_mkldnn_cache) {}
+Executor::Executor(const platform::Place& place) : place_(place) {}
 
 Executor::~Executor() {
 #ifdef PADDLE_WITH_MKLDNN
   // Clear mkl-dnn cache, unless explicitly
   // (as set in constructor) marked not to do so
   // this is needed to have mkl-dnn unit tests working
-  if ((clear_mkldnn_cache_) && (platform::is_cpu_place(place_))) {
+  if (platform::is_cpu_place(place_)) {
     platform::DeviceContextPool& pool = platform::DeviceContextPool::Instance();
     platform::MKLDNNDeviceContext* dev_ctx =
         (platform::MKLDNNDeviceContext*)pool.Get(place_);
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -91,7 +91,17 @@ ExecutorPrepareContext::~ExecutorPrepareContext() {
   VLOG(5) << "destroy ExecutorPrepareContext";
 }
 
-Executor::Executor(const platform::Place& place) : place_(place) {}
+Executor::Executor(const platform::Place& place, bool clear_mkldnn_cache)
+    : place_(place) {
+#ifdef PADDLE_WITH_MKLDNN
+  if ((clear_mkldnn_cache) && (platform::is_cpu_place(place))) {
+    platform::DeviceContextPool& pool = platform::DeviceContextPool::Instance();
+    platform::MKLDNNDeviceContext* dev_ctx =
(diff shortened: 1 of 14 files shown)


jczaja:
@luotao1  In your multi-threaded inference scenario ( hundreds of threads to do inference, but one at once actually doing prediction ) It each of Threads creating Predictor on its own or are they sharing Once created predictor?

grygielski (dismissed):
Great job! Maybe in the future we will be able to unify key creation into something like (dims, data_type, unique_string) and get rid of template versions. LGTM

luotao1:
> each of Threads creating Predictor on its own or are they sharing Once created predictor?

You can see the analyzer_detect_tester unit-test in [here](https://github.com/PaddlePaddle/Paddle/compare/develop...luotao1:detect_model_test?expand=1), it seems they sharing once created predictor.

grygielski (dismissed):
LGTM

paddle/fluid/framework/executor.cc
```diff
@@ -91,7 +91,22 @@ ExecutorPrepareContext::~ExecutorPrepareContext() {
   VLOG(5) << "destroy ExecutorPrepareContext";
 }
 
-Executor::Executor(const platform::Place& place) : place_(place) {}
+Executor::Executor(const platform::Place& place, bool clear_mkldnn_cache)
+    : place_(place), clear_mkldnn_cache_(clear_mkldnn_cache) {}
```
luotao1: @LeoZhao-Intel @jianhang-liu Could you review this PR, it's not elegant to add  `clear_mkldnn_cache` in `Executor` directly, do you have any suggestion?

jczaja:
@LeoZhao-Intel , @jianhang-liu Any suggestions, thoughts on this PR?

paddle/fluid/framework/executor.cc
jczaja: @luotao1  This `clear_mkldnn_cache` was added to prevent clearing cache when we have mulitple predictors e.g. mulitple threads , each thread is having its own predictor (and each predictor's IR Loads Model). If you tell me that this is not needed/used scenario then I can remove `clear_mkldnn_cache` param and clearing cache will always happen in destructor of Executor (this is fine when one predictor , shared by multiple threads).

luotao1: > To restore FP32 performance (or even make it a bit better) the only option now is to merge: #20241 . This PR
is having TID in key , but remove many of other attributes that were creating a hashing key. https://github.com/PaddlePaddle/Paddle/issues/20532#issuecomment-542310539

how about remove `clear_mkldnn_cache` param this time to fix FP32 regression ASAP for release/1.6? @jczaja @LeoZhao-Intel @jianhang-liu

ghost: @jczaja I also don't think adding mkldnn stuff in general "executor" is a good idea. Shall we simply remove this. @LeoZhao-Intel What's your suggestion?

jczaja: I will remove clear_mkldnn_cache param for Executor and that way destructor of Executor will call clearing of mkl-dnn cache regardless any condition. I will update this PR today EOD.

jczaja: @luotao1 I have updated PR. If this PR in acceptable then please test your internal workloads in particular I would like you to test performance of your multi-threaded inference scenario (Multiple threads each doing inference) as I could not test it locally.

luotao1: I test it on https://github.com/PaddlePaddle/Paddle/pull/18372#issuecomment-542992219 multi-thread detect model on the dynamic shape. It seems OK before and after this PR.

luotao1 (approved):

paddle/fluid/framework/executor.cc
jczaja: @luotao1 thank you

PR #20271, opened 2019-10-09
Fix none pull

zhhsplendid:
Test whether windows CI would fail after pull develop


zhhsplendid:
I'm going to close it because another PR has done same thing

PR #20277, opened 2019-10-09
polish en doc fluid.data in lstms

songyouwei:
1. `fluid.layers.data` -> `fluid.data`
2. use `int64` for `fluid.data` (feeding into `fluid.layers.embedding`)


songyouwei:
open at #20383

PR #20306, opened 2019-10-09
fix fuse_reduce_op quantization bug 

slf12:
fix fuse_reduce_op quantization bug, because build_strategy.fuse_all_reduce_ops's default value  changes to 'True' from v1.6.


wanghaoshuang (dismissed):

baiyfbupt (approved):
LGTM

PR #20331, opened 2019-10-09
Refine py_reader exit

sneaxiy:
`py_reader` would raise `EOFException` when the user-defined reader raises exceptions. The proper way is to terminate the training process when the user-defined reader raises exceptions. This PR refines the exit way.


paddle/fluid/pybind/reader_py.cc
```diff
@@ -139,19 +144,44 @@ class MultiDeviceFeedReader {
...
+        try {
+          readers_[i]->ReadNext(&ret_[i]);
+          return ret_[i].empty() ? Status::kEOF : Status::kSuccess;
+        } catch (...) {
+          exceptions_[i] = std::current_exception();
+          return Status::kException;
+        }
       });
     }
   }
```
zhhsplendid, line 156: Is it return for a void function?

sneaxiy: No. The return value is captured in `futures_[i]` but not the return value of `ReadAsync`.

zhhsplendid: Sorry, I was reading incorrectly :-)

zhhsplendid (approved):
LGTM

PR #20351, opened 2019-10-10
[cherry-pick]add input type, dtype and shape check for reshape op

liym27:
add input type and dtype check for reshape op. 
enhance shape error messages for reshape op.
chrry-pick [PR20099](https://github.com/PaddlePaddle/Paddle/pull/20099)


luotao1 (approved):
LGTM

PR #20370, opened 2019-10-10
remove incorrect new in c style

FrostML:
In C language, there should be a malloc, not new. The "new" in C api file is put in the c_api.h file by mistake by me. Remove incorrect new in C style for inference c api.


wopeizl (approved):
LGTM

PR #20421, opened 2019-10-10
Set batch norm and data norm argument 'do_model_average_for_mean_and_var' default as True

LielinJiang:
if argument 'do_model_average_for_mean_and_var' is False, api ExponentialMovingAverage cat not work at mean and variance parameter.


kuke (approved):
LGTM

heavengate (approved):
LG API

PR #20424, opened 2019-10-10
fix api fluid.cuda_pinned_places

songyouwei:
bug fix : `core.cuda_pinned_places` does not exists

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -264,7 +264,7 @@ def cuda_pinned_places(device_count=None):
         "Not compiled with CUDA"
     if device_count is None:
         device_count = _cpu_num()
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/framework.py
```diff
@@ -264,7 +264,7 @@ def cuda_pinned_places(device_count=None):
         "Not compiled with CUDA"
     if device_count is None:
         device_count = _cpu_num()
```
chengduoZH: `device_count = _cpu_num()` -> `device_count = len(_cuda_idx())`

songyouwei: `_cuda_idx` -> `_cuda_ids`

chengduoZH (approved):

PR #20427, opened 2019-10-10
Fleet: special case: strategy is None

mapingshuo:
We need to deal with a special case that distribute strategy is None, when initializing an instance of DistributedOptimizer class.


guru4elephant (approved):
LGTM

PR #20499, opened 2019-10-11
get shape from var desc to avoid run startup program

phlrain:
get shape from var desc to avoid run startup program


seiriosPlus (approved):
LGTM

PR #20529, opened 2019-10-11
[error-message]fix reshape input(x) error check on float16.

liym27:
input(x) of reshape support float16 in GPU.
image


luotao1 (approved):
LGTM

PR #20579, opened 2019-10-12
record the data_size of PD_ZeroCopyData's data

FrostML:
problem: no data_size to tell user how many data is in PD_ZeroCopyData
record the data_size of PD_ZeroCopyData's data


fc500110 (approved):
LGTM

PR #20581, opened 2019-10-12
Add config option in AnalysisConfig to disable GLOG INFO

cryoco:
**Problem:** The LOG(INFO) and VLOG outputs in paddle inference contains information of models(such as OP or OP fusion), which might compromise the confidentiality of the models.

**Solution:** Added DisableGlogInfo() interface in AnalysisConfig. Now one can call this function if the inner structure of his/her inference models must be strictly confidential, and no LOG(INFO) or VLOG logs will ever show during inference.


NHZlX (approved):
LGTM

PR #20612, opened 2019-10-14
fix the PD_ZeroCopyPredictorRun output problem

FrostML:
fix the PD_ZeroCopyPredictorRun output problem which is unexpected string
and add some checks and logs for users
and fix the cmakelists problem


fc500110 (approved):
LGTM

PR #20617, opened 2019-10-14
fix dgc test and bug when not set trainers_endpoints_

wangxicoding:
1. DGC will fail when fuse is enabled. Disable fuse to fix the test coverage problem.
2. Fix bug when not set trainers_endpoints DGC get nranks=0.


python/paddle/fluid/incubate/fleet/collective/__init__.py
```diff
@@ -271,7 +271,6 @@ def _try_to_compile(self, startup_program, main_program):
         node_num = self._node_num()
         assert node_num >= 1, "nccl2 node_num must >= 1, now:{}" % node_num
 
-        self._strategy.fuse_all_reduce_ops = True
```
wangxicoding, line 274: This line of code forces fuse_all_reduce_ops = True, regardless of whether the user has previously set it or not.

gongweibao (approved):
LGTM

PR #20619, opened 2019-10-14
fix small bug of error message for elementwise_add and elementwise_mul

danleifeng:
fix small bug in #20283 

edit type check message in nn.py


chenwhql (approved):
LGTM

PR #20636, opened 2019-10-15
Support fp16 in GPU impl of fused_elemwise_activation_op.

qingqing01:
When setting:

    build_strategy = fluid.BuildStrategy()
    build_strategy.fuse_elewise_add_act_ops = True

fused_elemwise_activation_op will used in some model, but it not support fp16 now.

- Support fp16 in GPU impl of fused_elemwise_activation_op.
- Add unit testing.

PR #20655, opened 2019-10-16
[cherry-pick] Support fp16 in GPU impl of fused_elemwise_activation_op. (#20636)

qingqing01:
* Support fp16 in fused_elemwise_activation_op.
* Fix unit testing in ONLY-CPU mode.


chengduoZH (approved):
LGTM

PR #20662, opened 2019-10-16
Make conv workspace size configurable

sneaxiy:
The original implementation of conv workspace size cannot be higher than 512MB. This PR fixes the bug by controlling by `FLAGS_conv_workspace_size_limit`.


zhhsplendid (approved):
LGTM

PR #20665, opened 2019-10-16
improve the performance of infer capi in PD_PredictorRun

FrostML:
improve the performance of infer capi in PD_PredictorRun. 
Since the predictor will be create every time Run is called, so use a map to store the predictor already created to avoid performance decrease.


fc500110 (approved):
LGTM

PR #20044, opened 2019-09-27
Add IndicateVarDataType interface to block tensor is not initialized problem in OP GetExceptedKernelType

chenwhql:
In some cases, paddle will throw errors like `holder_ should not be null. Tensor not initialized yet when Tensor::type() is called.`. This error is caused by the direct use of the tensor in the `Variable` without initialization, but this error message is not user friendly.

By analyzing issues, I found that `GetExceptedKernelType` is an important source of this problem. Our developers get the tensor type of input variable but don't check if the tensor was initialized. So I think we can add a funciton to deal with this problem.

If Op doesn't overwrite the `GetExceptedKernelType` function, use the `OperatorWithKernel::GetExpectedKernelType` function, it wil check if tensor was initialized, so I add the `IndicateVarDataType` which is just did what we should have done.

Then the interface `GetDataTypeOfVar` can be removed, and we should add a rule that developers must use this `IndicateVarDataType` interface in Ops' `GetExceptedKernelType`.

### The error cases:

https://github.com/PaddlePaddle/Paddle/issues/19522

image

https://github.com/PaddlePaddle/Paddle/issues/18432

image

### This PR's results:

- original error message:

```
holder_ should not be null
Tensor not initialized yet when Tensor::type() is called. at [/paddle/paddle/fluid/framework/tensor.h:139]
```

- new error message:

```
The Tensor in the save Op's Input Variable X(mean_0.tmp_0) is not initialized. at [/work/paddle/paddle/fluid/framework/operator.cc:1162]
  [operator < save > error]
```

## Use Cases

1. `GetDataTypeOfVar(ctx.InputVar("X")) -> IndicateVarDataType(ctx, "X")`

2. `ctx.Input<Tensor>("X")->type() -> IndicateVarDataType(ctx, "X")`

### Code
```
import paddle.fluid as fluid

place = fluid.CPUPlace()
exe = fluid.Executor(place)

x = fluid.layers.data(name='X', shape=[13], dtype='float32')
y = fluid.layers.data(name='Y', shape=[1], dtype='float32')
y_ = fluid.layers.fc(input=x, size=1, act=None)
loss = fluid.layers.square_error_cost(input=y_, label=y)
avg_loss = fluid.layers.mean(loss)
avg_loss.persistable = True

fluid.optimizer.SGD(learning_rate=0.01).minimize(avg_loss)

exe.run(fluid.default_startup_program())
fluid.io.save_inference_model(dirname="./save_op_test",
[...]

PR #20668, opened 2019-10-16
Replace risky GetInputType method with secure IndicateVarDataType interface

chenwhql:
In many operators, developers use `ctx.Input<Tensor>("X")->type()` to get Tensor type directly, then use this type infer the OpKernelType, but in this phase, The Tensor in Variable may not be initialized yet, which will cause paddle throw errors like `holder_ should not be null. Tensor not initialized yet when Tensor::type() is called.`, this error message makes users confused.

So I defined a secure interface `IndicateVarDataType` to replace the old implementation, this PR replace all risky code in paddle that has this problem.

Related PR: https://github.com/PaddlePaddle/Paddle/pull/20044

CE Verification passed: http://yq01-paddle-online.epc.baidu.com:8098/commit/details?commit=1acb9ceada9b8c120f48fca9def48d4ccff60098

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -48,16 +48,6 @@ std::vector<std::tuple<platform::Place, LibraryType>> kKernelPriority = {
     std::make_tuple(platform::CPUPlace(), LibraryType::kPlain),
 };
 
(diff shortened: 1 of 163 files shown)


luotao1 (approved):
LGTM

lanxianghit (approved):

raindrops2sea (approved):

PR #20672, opened 2019-10-16
Polish some codes

sneaxiy:
Polish codes of `SingleOpInplaceInToOut` and remove `BaseActivationFunctor<T>::Inplace()` method which is not used.


luotao1 (approved):
LGTM

PR #20676, opened 2019-10-16
Refine reduce codes to save compiling time and binary size

sneaxiy:
`TensorReduce` uses too many template functions, which makes compilation slow and the binary size larger. This PR refines it.

The testing make command on my machine is:
```
cmake .. -DWITH_GPU=ON -DWITH_TESTING=OFF -DWITH_DISTRIBUTE=ON -DWITH_MKL=ON -DWITH_AVX=ON -DWITH_NGRAPH=ON -DCMAKE_BUILD_TYPE=Release && make -j
```

The size of `core_avx.so` decreases from 533MB to 486MB, about 8.8% lower.


luotao1 (approved):
LGTM

PR #20695, opened 2019-10-17
Fix op run log when memory optimization strategy is enabled

sneaxiy:
When memory optimization strategy is enabled, data buffer of some tensors may be collected or reused in another thread, which causes the thread printing op running log raises error.

This PR fixes this bug. After this PR, the tensors collected before op runs would be printed as `unknown_dtype`.

```
I1017 09:26:49.364127 43469 operator.cc:162] CPUPlace Op(elementwise_add_grad), inputs:{Out@GRAD[Out@GRAD:float[2, 3, 4]({})], Y[Y:unknown_dtype[1, 1]({})]}, outputs:{X@GRAD[X@GRAD:[0]({})], Y@GRAD[Y@GRAD:[0]({})]}.
```


gongweibao (approved):
LGTM

PR #20697, opened 2019-10-17
alter the capi of PD_PredictorRun to provide proper function

FrostML:
Alter the c-api of PD_PredictorRun() to provide proper function. 
Since the use of (int\*\* size) is not proper, and there is no need to provide a pointer's pointer. So fix it to (int \*). Modify the way to pass parameter out_size in function and get a pass-by-parameter value. 
The familiar reason to fix the parameter output_data. There is a problem for (PD_Tensor\* output_data) when pass a PD_Tensor\* array. So fix this to PD_Tensor\*\*.


fc500110 (approved):
LGTM

PR #20699, opened 2019-10-17
Fix tensor_util_test random failure

sneaxiy:
`tensor_util_test` creates another stream to copy data, which may cause random data failure, conflicting with the default compute stream.


chengduoZH (approved):
LGTM

PR #20705, opened 2019-10-18
[cherry-pick] c api update in PD_PredictorRun

FrostML:
improve the performance of capi in PD_PredictorRun (#20665) to make predictor in a map to avoid create everytime run is called. 
alter the capi of PD_PredictorRun to provide proper function (#20697 ) to alter the way to pass value out_size.


fc500110 (approved):
LGTM

PR #20718, opened 2019-10-18
remove the warning issue test=develop

wopeizl:
Suppress some warnings for gcc8.

### DIFF
--- a/cmake/flags.cmake
+++ b/cmake/flags.cmake
@@ -167,6 +167,8 @@ if(NOT APPLE)
                 -Wno-error=cast-function-type # Warning in boost gcc 8.2
                 -Wno-error=parentheses # Warning in boost gcc 8.2
                 -Wno-error=catch-value # Warning in boost gcc 8.2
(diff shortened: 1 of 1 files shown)


cmake/flags.cmake
```diff
@@ -167,6 +167,8 @@ if(NOT APPLE)
                 -Wno-error=cast-function-type # Warning in boost gcc 8.2
                 -Wno-error=parentheses # Warning in boost gcc 8.2
                 -Wno-error=catch-value # Warning in boost gcc 8.2
+                -Wno-error=nonnull-compare # Warning in boost gcc 8.2
```
JiabinYang, line 170: these waring may be important for other files, better to add them back on other files except `boost` related files

JiabinYang (approved):
LGTM, better to change follow comments

cmake/flags.cmake
luotao1: 请问这个能加在boost.cmake里面么？不加在flags.cmake

PR #20733, opened 2019-10-19
Bug Fix: Paddle-TRT cannot handle adaptive pooling in pool2d op converter and "num" attribute in split op converter

cryoco:
**Problem:** 
1、Paddle-TRT had no support for adaptive attribute for pool2d op, thus when it is included in models like pspnet, there will be diff.
2、Paddle-TRT had no support for `num` attribute in split op, so when a split op with `num` but not `section` attribute set, TRT split op converter will throw an EnforceNotMet error.
**Solution:** 
1、Added adaptive attribute in pool2d trt plugin and extended avgpool plugin to pool plugin. Now paddle-TRT can handle pool op with adaptive attribute properly, and is able to fully support pspnet.
2、Added support for `num` attribute in split op. Now TRT can handle split ops with either `num` or `section` properly.


NHZlX (approved):
LGTM

PR #20748, opened 2019-10-21
add load numpy.ndarray from disk

DogHeadMoon:
For the saved model parameters *.pdparams, we have provided a new interface to load those parameters as python dictionary.

### DIFF
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -318,16 +318,10 @@ PYBIND11_MODULE(core_noavx, m) {
     for (auto &load_tensor : map_load_tensor) {
       const auto &dims = load_tensor.second->dims();
       auto dtype = load_tensor.second->type();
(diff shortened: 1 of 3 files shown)


paddle/fluid/pybind/pybind.cc
```diff
@@ -310,6 +310,53 @@ PYBIND11_MODULE(core_noavx, m) {
...
+        case proto::VarType::FP32:
+          np_type = py::dtype::of<float>();
+          break;
+        case proto::VarType::FP64:
+          np_type = py::dtype::of<double>();
+          break;
+        default:
+          PADDLE_THROW("Not support type %d", dtype);
+      }
+      map_output.emplace(
```
phlrain: TensorToPyArray 我们框架中有一个这样的接口，把tensor转成一个numpy的ndarray

DogHeadMoon: 已改为基于该接口进行调用

PR #20755, opened 2019-10-21
Cmake_generotor support has been added to enable multi-version VS sup…

zhwesky2010:
Protobuf and xxhash do not support multiple versions of VS because the cmake_generotor of main project has not been passed in. It will cause the compilation to be fail on a windows machines that have multi VS.
This PR fix these problem by passing in cmake_generator.


liupluswei (approved):
LGTM

PR #20758, opened 2019-10-21
Fix DGC algorithm flow to make it the same as paper

wangxicoding:
At present, the parameter update on DGC is different from the paper. The paper has already made momentum correction in the local and finally uses SGD to update the parameters. While, DGC is using Momentum SGD now, so it needs to fix the algorithm flow of DGC to make it the same as paper.
image


paddle/fluid/operators/optimizers/momentum_op.h
```diff
@@ -400,5 +400,52 @@ class MomentumOpKernel : public framework::OpKernel<T> {
   }
 };
 
+class MomentumOpMaker : public framework::OpProtoAndCheckerMaker {
```
gongweibao: Declare here and need not to define here?

python/paddle/fluid/tests/unittests/test_dist_mnist_dgc_nccl.py
```diff
@@ -33,7 +33,10 @@ def test_dist_train(self):
         if fluid.core.is_compiled_with_cuda():
             self.check_with_place(
                 "dist_mnist.py",
-                delta=1e-5,
+                # FIXME(wangxi): DGC may need a new unit test because its algo flow
```
gongweibao: How to assert when DGC be disabled?

paddle/fluid/operators/optimizers/dgc_momentum_op.h
```diff
@@ -0,0 +1,59 @@
...
+      VLOG(10) << " so use momentum optimizer";
+      return _momentum_op_kernel->Compute(context);
+    }
+
+    VLOG(10) << " so use sgd optimizer";
+    return _sgd_op_kernel->Compute(context);
+  }
+
+ private:
+  std::unique_ptr<MomentumOpKernel<DeviceContext, T>> _momentum_op_kernel;
```
gongweibao, line 54: 👍🏻

gongweibao (changes requested):

paddle/fluid/operators/optimizers/momentum_op.h
wangxicoding: Done. Define class declare func in .h，define func in .cc

python/paddle/fluid/tests/unittests/test_dist_mnist_dgc_nccl.py
wangxicoding: Done. Use log files to judge whether DGC is disabled.

gongweibao (approved):
LGTM

PR #20763, opened 2019-10-22
remove assert statement to support sqeeze op in drgraph

JepsonWong:
remove assert statement in squeeze function (python/paddle/fluid/layers/nn.py) to support sqeeze op in drgraph.


JiabinYang (approved):
LGTM

PR #20775, opened 2019-10-22
[Dygraph to static graph]JIT/Trace

sneaxiy:
Add `fluid.dygraph.jit.trace` API to trace the operators run in dygraph mode, so that it can be converted to a `ProgramDesc`.

The usage is:

```Python
import paddle.fluid.dygraph.jit as jit

outs, program = jit.trace(module=module, inputs=[image, label], 
                          feed_names=['image', 'label'], fetch_names=['loss'])
```

The `outs` are the output of `module(*inputs)` and program is the converted program with the given feed names and fetch names. 

**This API is not public yet in this PR because we should discuss the detailed API style later.**


python/paddle/fluid/dygraph/jit.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def extract_vars(inputs):
+    result_list = []
+    _extract_vars(inputs, result_list)
+    return result_list
+
+
+@dygraph_only
+def trace(module, inputs, feed_names=None, fetch_names=None):
```
JiabinYang, line 47: maybe we need some doc and sample code for this method

python/paddle/fluid/dygraph/jit.py
```diff
@@ -0,0 +1,81 @@
...
+def _extract_vars(inputs, result_list):
+    if isinstance(inputs, Variable):
+        result_list.append(inputs._ivar)
+
+    if isinstance(inputs, (list, tuple)):
+        for var in inputs:
+            _extract_vars(var, result_list)
+
+
+def extract_vars(inputs):
```
JiabinYang, line 40: Maybe there is no need for user to access to this method

JiabinYang (commented):
some comments

JiabinYang (approved):
LGTM

python/paddle/fluid/dygraph/jit.py
sneaxiy: Yes, we would not expose this method to users.

python/paddle/fluid/dygraph/jit.py
sneaxiy: Thanks! I would add more docs and sample codes in next PR!

PR #20781, opened 2019-10-22
fix bug in reshape: shape of input can contain more than one -1.

liym27:
**fix bug in reshape when the shape of input can contain more than one -1.**

### before:
run code:
```
input = fluid.data(name='x', shape=[1, None, None, 255], dtype="float32")
fluid.layers.reshape(input, shape=[1, 26, 26, 3, 85])
```
bug like this:
image


jiangjiajun:
在实际模型上测试已无问题

phlrain (approved):

PR #20798, opened 2019-10-23
update worker_num for MPISymetricRoleMaker

guru4elephant:
test=develop
worker_num should be notified to servers so that they can do communication things.


mapingshuo (approved):
LGTM

PR #20806, opened 2019-10-24
[1.6] fix bug for elementwise op cuda kernel grid_size when using fp16.

danleifeng:
when using fp16 mode and size=1,  cuda kernel will not run correctly.

```
 dim3 gird_size = dim3(
        (size / 2 + PADDLE_CUDA_THREAD_SIZE - 1) / PADDLE_CUDA_THREAD_SIZE, 1);
```


kuke (approved):
LGTM

PR #20809, opened 2019-10-24
Integration of third_party compilation structure

zhwesky2010:
Integration of third_party compilation structure. Add new file third_party.cmake.

The new file third_party.cmake accomplish uniform compilation control of third_party.

This file(third_party.cmake) can achieve the following functions:

- 1.Control whether the submodule of third_party should be compiled through flag;
- 2.Uniformly compile error message of third_party;
- 3.Accomplish the uniform compilation of third_party under cross-platform (Windows/ Linux/MAC);

### DIFF
--- a/cmake/third_party.cmake
+++ b/cmake/third_party.cmake
@@ -1,0 +1,233 @@
+# Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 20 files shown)


cmake/third_party.cmake
```diff
@@ -0,0 +1,233 @@
...
+    endif()
+    
+    if(WITH_MKLML)
+        MESSAGE(WARNING
+            "Mac is not supported with MKLML in Paddle yet. Force WITH_MKLML=OFF.")
+        set(WITH_MKLML OFF CACHE STRING "Disable MKLML package in MacOS" FORCE)
+    endif()
+endif()
+
+if(WIN32)
```
liupluswei, line 47: WITH_DISTRIBUTE is also used for non-third party logic, I prefer to remain the below code in the main CMakeLists.txt

PR #20852, opened 2019-10-28
strided_slice perforamnce improvement

wangchaochaohu:
add if branch to reduce time according strides parameter. if strides is positive, we don't need to do reverse.


zhaoyuchen2018 (approved):
LGTM

PR #20854, opened 2019-10-28
Lazy initialization of allocators

sneaxiy:
The former strategy would fail in allocator initialization even if the network only runs on one GPU card. This PR refines it.

### DIFF
--- a/paddle/fluid/memory/allocation/naive_best_fit_allocator.cc
+++ b/paddle/fluid/memory/allocation/naive_best_fit_allocator.cc
@@ -101,25 +101,38 @@ size_t Used<platform::CPUPlace>(const platform::CPUPlace &place) {
 }
 
 #ifdef PADDLE_WITH_CUDA
-BuddyAllocator *GetGPUBuddyAllocator(int gpu_id) {
-  static std::once_flag init_flag;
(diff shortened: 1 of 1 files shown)


paddle/fluid/memory/allocation/naive_best_fit_allocator.cc
```diff
@@ -101,25 +101,38 @@ size_t Used<platform::CPUPlace>(const platform::CPUPlace &place) {
...
+    }
+  }
+
+  static GPUBuddyAllocatorList *CreateNewInstance() {
+    return new GPUBuddyAllocatorList();
+  }
+
+ public:
+  static GPUBuddyAllocatorList *Instance() {
+    static auto *instance = CreateNewInstance();
```
zhhsplendid, line 121: Can we just write

"static auto *instance = new GPUBuddyAllocatorList()"

and discard the CreateNewInstance private method?

zhhsplendid (approved):

paddle/fluid/memory/allocation/naive_best_fit_allocator.cc
sneaxiy: Not quite sure whether `new` expression in static variable is thread-safe.

PR #20858, opened 2019-10-28
Better TensorRT support

jeng1220:
1. Add new search path for TensorRT at tensorrt.cmake. In NVIDIA official docker image, TensorRT headers are located at `/usr/include/x86_64-linux-gnu` and TensorRT libraries are located at `/usr/lib/x86_64-linux-gnu`, so using `-DTENSORRT_ROOT` will fail to detect TensorRT.

2. Add better debug message. There is no debug/warning message to tell developer that TensorRT is failed to be detected.

3. Fix the bug of detection of TensorRT version. In later version of TensorRT (e.g. v6), `NV_TENSORRT_MAJOR` is defined at `NvInferVersion.h` instead of `NvInfer.h`, so add compatibility fix.


jeng1220:
@NHZlX ,
Could you take a look?

jeng1220:
test=develop

jeng1220:
It seems that I don't have permission to see the result of `Paddle-CI` , so I don't why it failed to pass Paddle-CI. @NHZlX , Could you help to take a look?

PR #20860, opened 2019-10-28
Fix Paddle Cloud role maker

MrChengmo:
- Fix Paddle Cloud Role Maker
- Add Error information in distribute optimizer when user hasn't call fleet.init(role)
- change rpc_retry_times default value，from 0->3


seiriosPlus (approved):
LGTM

PR #20861, opened 2019-10-29
Fix py_reader combination ut

sneaxiy:
`test_py_reader_combination.py` may fail if the thread does not finish after python interpreter exits. Do not know why it happens. This PR is a hot fix of the unittest.


JiabinYang (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #20869, opened 2019-10-29
turn on the flag:ON_INFER to build inference library, test=develop

zhwesky2010:
Turn on the flag: ON_INFER to build inference library. Now it is turned off by defalut, which will make C++ inference library useless.


luotao1:
看了下失败的单测，主要是Parallel Executor和分布式相关的，这些单测应该在`ON_INFER`打开时不跑的。

PR #20875, opened 2019-10-29
Add custom black variable name set in amp interface.

gongweibao:
1. Some variables feed into executor may be very large to use float16, so we need the interface to disable them.
2. If one variable is converted to the destination type, duplicated conversion should be avoid.

### DIFF
--- a/python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
+++ b/python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
@@ -94,21 +94,24 @@ def _insert_cast_op(block, op, idx, src_dtype, dest_dtype):
             if in_var.type not in valid_types:
                 continue
             if in_var.dtype == src_dtype:
-                out_var = block.create_var(
-                    name=in_var.name + \
-                            '.cast_' + _dtype_to_str(dest_dtype),
-                    dtype=dest_dtype,
-                    persistable=False,
-                    stop_gradient=False)
-                block._insert_op(
-                    idx,
-                    type="cast",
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
```diff
@@ -155,6 +159,18 @@ def find_true_prev_op(ops, cur_op, var_name):
...
+def _is_in_black_varnames(op, amp_lists):
+    for in_name in op.input_arg_names:
+        if in_name in amp_lists.black_varnames:
+            return True
+
+    for out_name in op.output_arg_names:
+        if out_name in amp_lists.black_varnames:
+            return True
+
+    return False
```
Jie-Fang, line 171: This is a special situation. Could you give me an specific case. In the Transformer, which op has these vars? In my opinion, users should get var name to customize amp list, it's a little inconvenient. Maybe this special situation can be categorized to the general case. Thus it's better you can give an example.
BTW, this method may solve the demand that we can not only make some kinds of op types, but also some exact ops to be casted in Object Detection models.

gongweibao: https://github.com/PaddlePaddle/Fleet/blob/develop/benchmark/collective/transformer/train.py#L213

gongweibao: src_slf_attn_bias, trg_slf_attn_bias, trg_src_attn_bias

PR #20876, opened 2019-10-29
fix select_rows mergeadd bug, test=develop

frankwhzhang:
fix select_rows mergeadd bug, use actual memory size instead of defined memory.


seiriosPlus (approved):
LGTM

PR #20882, opened 2019-10-29
Add c++ global current tracer for dygraph

zhiqiu:
In dygraph mode, `tracer` is used for running ops. 
This PR, 

- adds and keeps global current tracer in c++, so we can use the global tracer without passing it from python when moving some logics from python to C++, in the future.

- adds `_expected_place` and `_train_mode` to C++ Tracer and pybind to python, which is set during corresponding Python `guard`.


paddle/fluid/imperative/tracer.cc
```diff
@@ -12,13 +12,20 @@
...
 #include "paddle/fluid/imperative/tracer.h"
+#include <iostream>
 #include <unordered_set>
 #include <utility>
 #include "paddle/fluid/platform/profiler.h"
-
 namespace paddle {
 namespace imperative {
 
+static std::shared_ptr<Tracer> current_tracer(nullptr);
```
sneaxiy: Try to use `g_` prefix for global variable.

zhiqiu: OK, thanks.

paddle/fluid/imperative/tracer.h
```diff
@@ -94,5 +94,9 @@ class Tracer {
   std::unique_ptr<UniqueNameGenerator> generator_;
 };
 
+// To access static variable current_tracer
+std::shared_ptr<Tracer>& GetCurrentTracer();
```
sneaxiy: Try to return non-const pointer instead of non-const reference.

zhiqiu: Done, thanks.

sneaxiy (approved):

PR #20894, opened 2019-10-30
Refine pe when exception raises

sneaxiy:
`ExceptionHolder` has lock inside. But when thread number is 1, it is not necessary to use the lock of `ExceptionHolder`. This PR refines it.


chenwhql (approved):
LGTM

PR #20899, opened 2019-10-30
add dequantize_abs_max op and modify lookup_table op

slf12:
add dequantize_abs_max op  that input is int8 type and output is float32 type and modify lookup_table op to support int8 type


paddle/fluid/operators/lookup_table_op.h
```diff
@@ -122,8 +121,8 @@ class LookupTableKernel : public framework::OpKernel<T> {
             PADDLE_ENFORCE_GE(
                 id_index, 0, "the input key should be exists. But received %d.",
                 id_index);
-            blas.VCOPY(row_width, table + id_index * row_width,
-                       output + i * row_width);
+            memcpy(output + i * row_width, table + id_index * row_width,
+                   row_width * sizeof(T));
```
qingqing01: 这里还是依据type判断下，非int8时，用blas.VCOPY，int8时用std::memcpy

qingqing01 (approved):

paddle/fluid/operators/lookup_table_op.h
slf12: done

PR #20912, opened 2019-10-30
Compatible int32 and int64 for attr in concat/split/unsqueeze. 

liym27:
When attribute supports Tensor or list with Tensor, the data type of Tensor can be int32 or int64. Now it is compatible with int32 and int64.


guoshengCS (approved):
LGTM

lanxianghit (approved):

PR #20913, opened 2019-10-30
simplify master+patch，remove ins when size != merge_size or has conflict slot

xjqbest:
* remove duplicate code and duplicate config of master+patch
* drop all ins which has conflict slot or size < merge_size 

note: 
 * user only need to set merge size，if ins num of same id is not equal to merge size, just drop these ins
* user must make sure master data and patch data has no same slot whose feasigns are both non-zero, otherwise these ins will be dropped. (slot list should still be the same of both master and patch)


fuyinno4 (dismissed):

yaoxuefeng6 (approved):
lgtm

PR #20935, opened 2019-10-31
And Enforce in fuse pass for DGC doesn't support fuse for now

wangxicoding:
DGC doesn't support fuse for now, add PADDLE_ENFORCE_NE in fuse_all_reduce_op_pass.cc to judge if use DGC and fuse same time。


gongweibao (approved):
LGTM

Issue #20749, opened 2019-10-21
paddle.dataset.sentiment 数据集api报错


DDDivano:
linux下
![image](https://user-images.githubusercontent.com/10734244/67183092-35761d80-f413-11e9-836e-3a62dfb9ea00.png)
mac下
![image](https://user-images.githubusercontent.com/10734244/67183121-4cb50b00-f413-11e9-8869-acdbaa936e48.png)

sandyhouse:
解决方案可参考：https://www.cnblogs.com/wsygdb/p/7902956.html和https://www.imooc.com/qadetail/89863
后面修复一下。

sandyhouse:
fixed by pr #20942

PR #20942, opened 2019-10-31
fix the compatiable problem between PY2 and PY3 (issue#20749)

sandyhouse:
Fix issue[#20749](https://github.com/PaddlePaddle/Paddle/issues/20749)
the compatiable problem between PY2 and PY3

### DIFF
--- a/python/paddle/dataset/sentiment.py
+++ b/python/paddle/dataset/sentiment.py
@@ -28,6 +28,9 @@ from itertools import chain
 
 import nltk
 from nltk.corpus import movie_reviews
+import ssl
+ssl._create_default_https_context = ssl._create_unverified_context
+from functools import cmp_to_key
 
 import paddle.dataset.common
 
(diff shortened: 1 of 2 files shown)


gavin1332 (approved):
LGTM

python/paddle/dataset/sentiment.py
```diff
@@ -28,6 +28,9 @@
 
 import nltk
 from nltk.corpus import movie_reviews
+import ssl
+ssl._create_default_https_context = ssl._create_unverified_context
```
luotao1, line 32: Why should you add `ssl` here? It is not used in this PR, and if you add this line, we should `pip install ssl` before development. @sandyhouse

sandyhouse: add ssl to fix the issue#20749. I will launch a new pr to add ssl to requirements.txt.

luotao1: You should add `ssl` into `python/requirments.txt` for the usage.

sandyhouse: Right. I have opened the [pr#21669](https://github.com/PaddlePaddle/Paddle/pull/21669) to do that.

PR #20949, opened 2019-10-31
fix the computation for dx (grad for x) for prelu operation.

sandyhouse:
For prelu:
f(x) = x if x > 0 else _a_*x, where _a_ is a learnable parameter.
so df(x) = 1 if x > 0 else _a_.

Todo：
- Replace the macro definition of CUDA_NUM_THREADS with the constant PADDLE_CUDA_NUM_THREADS;
- Remove the _PADDLE_ prefix of the PADDLE_GET_BLOCKS function or define the function as a global function.

### DIFF
--- a/paddle/fluid/operators/math/prelu.cu
+++ b/paddle/fluid/operators/math/prelu.cu
@@ -18,7 +18,7 @@ namespace paddle {
 namespace operators {
 namespace math {
 
-static constexpr int PD_CUDA_NUM_THREADS = 1024;
+#define CUDA_NUM_THREADS 1024
 
 // CUDA: grid stride looping
 #define CUDA_KERNEL_LOOP(i, n)                                 \
@@ -26,7 +26,7 @@ static constexpr int PD_CUDA_NUM_THREADS = 1024;
        i += blockDim.x * gridDim.x)
 
 inline static int PADDLE_GET_BLOCKS(const int N) {
-  return (N + PD_CUDA_NUM_THREADS - 1) / PD_CUDA_NUM_THREADS;
+  return (N + CUDA_NUM_THREADS - 1) / CUDA_NUM_THREADS;
 }
 
 template <typename T>
@@ -74,7 +74,7 @@ void PreluChannelWiseDirectCUDAFunctor<T>::operator()(
   size_t plane_size = input_shape[2] * input_shape[3];
   size_t spatial_size = input_shape[1] * plane_size;
   size_t numel = input_shape[0] * spatial_size;
-  PReluChannelWiseKernel<<<PADDLE_GET_BLOCKS(numel), PD_CUDA_NUM_THREADS, 0,
+  PReluChannelWiseKernel<<<PADDLE_GET_BLOCKS(numel), CUDA_NUM_THREADS, 0,
                            stream>>>(input, alpha, output, input_shape[1],
                                      plane_size, numel);
 }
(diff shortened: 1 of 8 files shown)


paddle/fluid/operators/math/prelu.cu
```diff
@@ -18,120 +18,88 @@ namespace paddle {
...
-  size_t offset = blockIdx.x * spatial_size;
-  const T *in = input + offset;
-  T *out = output + offset;
-  T scale = alpha[blockIdx.x % channel];
-
-  for (size_t i = threadIdx.x; i < spatial_size; i += blockDim.x) {
-    T x = in[i];
-    out[i] = (x > 0) ? x : scale * x;
+                                       T *output, size_t channel_num,
+                                       size_t plane_size, size_t numel) {
```
gavin1332, line 35: be aware this is in case of BCHW layout. i learn from gongweibao that the BHWC layout is more fit for FP16, so it is better to discuss with him in the fureture.

paddle/fluid/operators/math/prelu.cu
```diff
@@ -18,120 +18,88 @@ namespace paddle {
 namespace operators {
 namespace math {
 
-static const int CUDA_NUM_THREADS = 1024;
-static const int CUDA_MAX_NUM_BLOCKS = 65535;
-inline static int GET_NUM_BLOCKS(const int N) {
-  return (N + CUDA_NUM_THREADS - 1) / CUDA_NUM_THREADS;
+static constexpr int PD_CUDA_NUM_THREADS = 1024;
```
gavin1332: according to the code style of paddle, static variable should be declared in s_xxx format. In this condition, “#define CUDA_NUM_THREADS 1024” looks better

gavin1332: Have you check whether there is a global variable to define this `CUDA_NUM_THREADS`，if there is, you'd better to use the global one for consistence . if not, `CUDA_NUM_THREADS` becomes a local configuration, which is not proper for `PD_` or `PADDLE_` prefix.

paddle/fluid/operators/prelu_op.cu
```diff
@@ -59,71 +67,47 @@ class CUDAPReluKernel : public framework::OpKernel<T> {
...
-    T x = x_ptr[offset + i];
-    T dy = dy_ptr[offset + i];
-    T alpha = alpha_func(alpha_ptr, channel, spatial_size, i);
-    if (dx_ptr != nullptr) dx_ptr[offset + i] = (y > 0) ? dy : alpha * dy;
-    if (dalpha_ptr != nullptr) dalpha_ptr[offset + i] = (x > 0) ? 0 : x * dy;
+__global__ void PReluOpGradKernel(const T* x_ptr, const T* y_ptr,
+                                  const T* alpha_ptr, const T* dy_ptr,
+                                  T* dx_ptr, T* dalpha_ptr, size_t channel_num,
+                                  size_t plane_size, size_t spatial_size,
+                                  size_t numel, PRELU_MODE mode) {
```
gavin1332, line 77: why useing one single kernel function for BP but 3 separate functions for FF?

gavin1332: just a discussion, need your argument

gavin1332 (commented):
some comments is given for further discussion.

paddle/fluid/operators/math/prelu.cu
sandyhouse: OK, I will have a discussion with weibao next week.

paddle/fluid/operators/math/prelu.cu
sandyhouse: Done. And there is a global variable called PADDLE_CUDA_NUM_THREADS which is set 512. But i think 1024 is better value, so I use the #define instructive statement to set the value to 1024.

paddle/fluid/operators/prelu_op.cu
sandyhouse: Because other files use the separate functions for FF, e.g., paddle/fluid/inference/tensorrt/plugin/prelu_op_plugin.cu.

gavin1332 (approved):
LGTM

PR #20959, opened 2019-11-01
Update tensor.set() to supports array list

zhiqiu:
This PR clears unused code of tensor.set() and updates [PR19964](https://github.com/PaddlePaddle/Paddle/pull/19964), make tensor.set() support array list.
image


sneaxiy (approved):
LGTM.

PR #20983, opened 2019-11-04
Support numpy bridge (enabled by default in dygraph mode)

songyouwei:
NumPy Bridge is used for **zero-copy creation of a Tensor with a NumPy array**.
`zero_copy` is enabled by default when place is a `CPUPlace` and is in dygraph mode.

The created Tensor and NumPy array will **share their underlying memory locations** (only when working with `CPUPlace`), and change the numpy array will also change the corresponding Tensor.

**Example:**
```python
import numpy as np
import paddle.fluid as fluid

data_np = np.array([[2, 3, 1]]).astype('float32')
with fluid.dygraph.guard(fluid.CPUPlace()):
    var = fluid.dygraph.to_variable(data_np, zero_copy=True)
    data_np[0][0] = 4
    var[0][0].numpy()[0] == 4  # True

```


paddle/fluid/pybind/tensor_py.h
```diff
@@ -64,6 +70,22 @@ namespace pybind {
 
 namespace details {
 
+class PYBIND11_HIDDEN NumpyAllocation : public memory::Allocation {
+ public:
+  explicit NumpyAllocation(PyObject *arr, size_t size,
+                           paddle::platform::Place place)
+      : Allocation(PyArray_DATA(reinterpret_cast<PyArrayObject *>(arr)), size,
+                   place),
+        arr_(arr) {}
```
phlrain: 析构的时候 有decref，构造的时候需要increase 么

songyouwei: 这里NumpyAllocation只hold一个PyObject的裸指针，传入之前需要incref：
https://github.com/PaddlePaddle/Paddle/blob/07f055feae6d703dbecab0a708fae58c7e1d6a89/paddle/fluid/pybind/tensor_py.h#L176

phlrain (dismissed):

heavengate (dismissed):

wanghaoshuang (dismissed):

paddle/fluid/framework/tensor.h
```diff
@@ -168,6 +168,8 @@ class Tensor {
 
   void ResetHolder(std::shared_ptr<memory::Allocation> holder);
 
+  void ResetType(const proto::VarType::Type type) { type_ = type; }
```
sneaxiy: Try to add a function like `void ResetHolderWithType(std::shared_ptr<memory::Allocation> holder, proto::VarType::Type type)` instead of reseting data type alone. Reseting data type alone is meaningless.

In this PR, you would call `ResetHolder` and `ResetType` in the meantime, but it is not guaranteed in the future development.

paddle/fluid/pybind/pybind.cc
```diff
@@ -458,11 +458,12 @@ PYBIND11_MODULE(core_noavx, m) {
...
       .def("_clear", &Tensor::clear)
       .def("set", SetTensorFromPyArray<paddle::platform::CPUPlace>,
-           py::arg("array"), py::arg("place"))
+           py::arg("array"), py::arg("place"), py::arg("zero_copy") = false)
       .def("set", SetTensorFromPyArray<paddle::platform::CUDAPlace>,
-           py::arg("array"), py::arg("place"))
+           py::arg("array"), py::arg("place"), py::arg("zero_copy") = false)
       .def("set", SetTensorFromPyArray<paddle::platform::CUDAPinnedPlace>,
-           py::arg("array"), py::arg("place"), R"DOC(
+           py::arg("array"), py::arg("place"), py::arg("zero_copy") = false,
```
sneaxiy, line 465: No doc for `zero_copy` while `Tensor.set` is a public API?

paddle/fluid/pybind/tensor_py.h
```diff
@@ -13,7 +13,13 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 #pragma once
+
+#ifndef NPY_NO_DEPRECATED_API
```
sneaxiy: Why add this? Try to add some comments on code?

paddle/fluid/pybind/tensor_py.h
```diff
@@ -64,6 +70,22 @@ namespace pybind {
 
 namespace details {
 
+class PYBIND11_HIDDEN NumpyAllocation : public memory::Allocation {
+ public:
+  explicit NumpyAllocation(PyObject *arr, size_t size,
```
sneaxiy: The design of `NumpyAllocation` can be improved. There are a few problems in the current implementation:

- There is no mechanism to check whether `PyObject *arr` is exactly an object of `PyArrayObject`. You use `reinterpret_cast` below, but it is not type safe.
- There is not necessary to parse `arr` and `size` in the same time. Information of `size` can be calculated from `arr` itself.
- Increase reference count outside `NumpyAllocation` and decrease reference count inside `NumpyAllocation` is not good.
- Check the `place` is `CPUPlace` exactly or just remove this parameter.
- `explicit` can be removed when there is not 1 parameter alone.

paddle/fluid/pybind/tensor_py.h
```diff
@@ -64,6 +70,22 @@ namespace pybind {
...
+        arr_(arr) {}
+  ~NumpyAllocation() override {
+    py::gil_scoped_acquire gil;
+    Py_XDECREF(arr_);
+  }
+
+ private:
+  PyObject *arr_;
+};
+
```
sneaxiy, line 92: Recommended better design for `NumpyAllocation`:

```C++
class NumpyAllocation : public Allocation {
 public:
  explicit NumpyAllocation(const py::array &arr) : 
        Allocation(arr.data(), arr.bytes(), platform::CPUPlace()), 
        arr_(arr.ptr()) {
       PADDLE_ENFORCE_NOT_NULL(arr_);
       PADDLE_ENFORCE_NE(arr, Py_None);
       Py_INCREF(arr_);
  }

   ~NumpyAllocation() {
       py::gil_scoped_acquire gil;
       Py_DECREF(arr_);
   }

 private:
   PyObject *arr_;
};
```

paddle/fluid/pybind/tensor_py.h
```diff
@@ -141,19 +163,28 @@ template <typename T, typename P>
...
   for (decltype(array.ndim()) i = 0; i < array.ndim(); ++i) {
     dims.push_back(static_cast<int>(array.shape()[i]));
   }
   self->Resize(framework::make_ddim(dims));
-  auto dst = self->mutable_data<T>(place);
 
   if (paddle::platform::is_cpu_place(place)) {
-    std::memcpy(dst, array.data(), array.nbytes());
+    if (zero_copy) {
+      array.inc_ref();
```
sneaxiy: See the comments above and change this part of codes.

paddle/fluid/pybind/tensor_py.h
```diff
@@ -13,7 +13,13 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 #pragma once
+
+#ifndef NPY_NO_DEPRECATED_API
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+#endif
+
 #include <Python.h>
+#include <numpy/arrayobject.h>
```
sneaxiy: If the constructor of `NumpyAllocation` only accepts `py::array` as its unique parameter, please remove this header.

paddle/fluid/pybind/tensor_py.h
songyouwei: done

paddle/fluid/pybind/tensor_py.h
songyouwei: removed (since we use pybind interface now)

paddle/fluid/pybind/tensor_py.h
```diff
@@ -64,6 +65,31 @@ namespace pybind {
...
+                                      "numpy array cannot be nullptr"));
+    PADDLE_ENFORCE_NE(
+        arr_, Py_None,
+        platform::errors::PreconditionNotMet(
+            "The underlying PyObject pointer of numpy array cannot be None"));
+    Py_INCREF(arr_);
+  }
+  ~NumpyAllocation() override {
+    py::gil_scoped_acquire gil;
+    Py_XDECREF(arr_);
```
sneaxiy: Not quite important: since `arr_` is not `nullptr`, `Py_DECREF` is enough. 

`Py_XDECREF(arr_)` is equal to `if (arr_ != nullptr) Py_DECREF(arr_);`

python/paddle/fluid/framework.py
```diff
@@ -4752,9 +4752,11 @@ def _dygraph_guard(tracer):
     global _dygraph_tracer_
     tmp_trace = _dygraph_tracer_
     _dygraph_tracer_ = tracer
+    core._switch_tracer(tracer)
 
     yield
 
+    core._switch_tracer(None)
```
sneaxiy: Why switch to `None` instead of `tmp_trace`?

python/paddle/fluid/dygraph/base.py
```diff
@@ -138,7 +138,6 @@ def guard(place=None):
     train = framework.Program()
     startup = framework.Program()
     tracer = Tracer()
-    core._switch_tracer(tracer)
```
sneaxiy, line 141: Why remove this line?

python/paddle/fluid/framework.py
songyouwei: a mistake, updated to `tmp_trace`

paddle/fluid/pybind/tensor_py.h
songyouwei: done

python/paddle/fluid/dygraph/base.py
songyouwei: moved to `_dygraph_guard`

sneaxiy (approved):
LGTM.

phlrain (approved):

saxon-zh (approved):
LGTM

PR #20988, opened 2019-11-04
refine code for code reuse 

wangchaochaohu:
we need to add same function to the fuse_conv pass, In order to reuse the function code, we refine the function code to the cudnn_layer_helper.h


zhaoyuchen2018 (approved):
LGTM

PR #20995, opened 2019-11-04
Add Amsgrad optimizer

jinxing94:
Amsgrad optimizer is a variant of the Adam optimizer. It is proposed in the ICLR 2018 paper titled [on the convergence of Adam and beyond](https://openreview.net/pdf?id=ryQu7f-RZ) .  

The paper argues that the exponential moving average used in Adam optimizer may cause the model fail to converge to an optimal solution, even in some simple convex optimization settings.  Amsgrad is proposed to address the issue by endowing the algorithm with “long-term memory” of past gradients.

Both tensorflow and pytorch have added the amsgrad optimizer in their official APIs.  See the amsgrad option in the Adam APIs of tensorflow and pytorch for reference.


paddle/fluid/operators/optimizers/amsgrad_op.cc
```diff
@@ -0,0 +1,165 @@
+/* Copyright (c) 2016 PaddlePaddle Authors. All Rights Reserved.
```
shippingwang: 2016->2019

paddle/fluid/operators/optimizers/amsgrad_op.cc
```diff
@@ -0,0 +1,165 @@
...
+                     "when not zero, if param row size is larger then "
+                     "min_row_size_to_use_multithread and "
+                     "inner_op_parallelism is larger then 0, sparse update "
+                     "will run in multithread mode")
+        .SetDefault(1000);
+
+    AddComment(R"DOC(
+Amsgrad Optimizer.
+
+This implements the Amsgrad optimizer from Section 2 of the Amsgrad
```
shippingwang: same as Adam??

shippingwang (changes requested):
*haven't finished the review yet

paddle/fluid/operators/optimizers/amsgrad_op.cc
jinxing94: Thanks for pointing this. I have changed it in a new commit.

paddle/fluid/operators/optimizers/amsgrad_op.cc
jinxing94: No. Thanks for pointing the typo.  I have updated it in a new commit.

PR #21012, opened 2019-11-05
Fix dgc buffer illegal & reuse velocity

wangxicoding:
1. Fix DGC gather buffer illegal problem by alloc buffer in dgc_op, avoid multi-stream bug in memory::Alloc
2. Reuse velocity in dgc_op & dgc_momentum_op, save memory & reserve global momentum information.

### DIFF
--- a/paddle/fluid/framework/details/sparse_all_reduce_op_handle.cc
+++ b/paddle/fluid/framework/details/sparse_all_reduce_op_handle.cc
@@ -137,10 +137,6 @@ void SparseAllReduceOpHandle::RunImplEncoded() {
     // dgc use ncclAllGather to get all the encoded data
     // so the buffer need nranks.
     int buf_size = nranks_ * encode_size;
-    // auto tmp_ious_data = memory::Alloc(place, buf_size);
-    // void *gather_buff = reinterpret_cast<void *>(tmp_ious_data->ptr());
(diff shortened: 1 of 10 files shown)


paddle/fluid/framework/details/sparse_all_reduce_op_handle.cc
```diff
@@ -135,9 +137,11 @@ void SparseAllReduceOpHandle::RunImplEncoded() {
     // dgc use ncclAllGather to get all the encoded data
     // so the buffer need nranks.
     int buf_size = nranks_ * encode_size;
-    auto tmp_ious_data = memory::Alloc(place, buf_size);
-    void *gather_buff = reinterpret_cast<void *>(tmp_ious_data->ptr());
-    allocations.emplace_back(std::move(tmp_ious_data));
+    // auto tmp_ious_data = memory::Alloc(place, buf_size);
```
gongweibao: clean up these codes!

gongweibao (changes requested):
Small needs fix!

paddle/fluid/framework/details/sparse_all_reduce_op_handle.cc
wangxicoding: Done.

gongweibao (approved):
LGTM

PR #21016, opened 2019-11-05
Add select_input_op and select_output_op

zhhsplendid:
Add select_input_op and select_output_op, which is useful in control flow.


paddle/fluid/operators/merge_selected_var_op.cc
```diff
@@ -0,0 +1,160 @@
...
+See the License for the specific language governing permissions and
+limitations under the License. */
+
+#include "paddle/fluid/framework/op_registry.h"
+#include "paddle/fluid/memory/memcpy.h"
+
+namespace paddle {
+namespace operators {
+
+class MergeSelectedVarOp : public framework::OperatorBase {
```
sneaxiy: Seems that `merge_selected_var` and `split_selected_var` ops are more like something `select` instead of `merge` and `split`.

paddle/fluid/operators/merge_selected_var_op.cc
```diff
@@ -0,0 +1,160 @@
...
+ public:
+  MergeSelectedVarOp(const std::string &type,
+                     const framework::VariableNameMap &inputs,
+                     const framework::VariableNameMap &outputs,
+                     const framework::AttributeMap &attrs)
+      : OperatorBase(type, inputs, outputs, attrs) {}
+
+ private:
+  void RunImpl(const framework::Scope &scope,
+               const platform::Place &dev_place) const override {
```
sneaxiy: The implementation of `MergeSelectedVarOp` and `SplitSelectedVarOp` is nearly the same. Try to simplify the code (also consider the code of `AssignOp`. Some codes can be reused).

paddle/fluid/operators/merge_selected_var_op.cc
```diff
@@ -0,0 +1,160 @@
...
+    const framework::Variable *selected_x =
+        scope.FindVar(x_names[output_branch]);
+
+    if (selected_x->IsType<framework::LoDTensor>()) {
+      auto &to_copy = selected_x->Get<framework::LoDTensor>();
+      auto *out =
+          scope.FindVar(Output("Out"))->GetMutable<framework::LoDTensor>();
+      framework::TensorCopy(to_copy, to_copy.place(), dev_ctx, out);
+      out->set_lod(to_copy.lod());
+    } else if (selected_x->IsType<framework::SelectedRows>()) {
```
sneaxiy: I think that it is not necessary to support `SelectedRows`.

paddle/fluid/operators/split_selected_var_op.cc
```diff
@@ -0,0 +1,153 @@
...
+              "The output can contains multiple variables. The output of "
+              "selected branch will be same as input. We do nothing for "
+              "variables in other branch")
+        .AsDuplicable();
+    // TODO(huihuangzheng): decide whether to add support for lod level
+    // Because this op is blocking whole control flow. I am implementing MVP
+    // (minimal viable product) here.
+    AddComment(R"DOC(
+Split input variable into one output branch. The mask is an integer tensor to
+specify which output branch should copy the input.
```
sneaxiy: As discussed offline, we should add some unittests to check whether it is necessary to set the other non-selected var to be zero later.

paddle/fluid/operators/merge_selected_var_op.cc
zhhsplendid: Done. In addition, thank you for these! I found AssignFunctor in assign op can be reused!

Created a file named "select_op_helper.h" and put GetBranchNumber which converts Mask tensor to int.

Moved AssignFunctor from assign_op.cc to assign_op.h and use it by including assign_op.h

paddle/fluid/operators/merge_selected_var_op.cc
zhhsplendid: Done, renamed to SelectInput, SelectOutput. Also renamed python API

paddle/fluid/operators/merge_selected_var_op.cc
zhhsplendid: Deleted

paddle/fluid/operators/split_selected_var_op.cc
zhhsplendid: Yes, I will do that in follow-up PRs and you will be one of reviewers :-)

sneaxiy (dismissed):
LGTM.

sneaxiy (approved):

PR #21031, opened 2019-11-06
Refine Tensor method

sneaxiy:
Change the parameter of some methods in Tensor to be reference instead of copy to avoid deep copy.


Xreki (approved):
LGTM

luotao1 (approved):
LGTM

PR #21041, opened 2019-11-06
Add Asypadding for conv fusion.

zhaoyuchen2018:
test=develop

reference: pr/20042


paddle/fluid/operators/fused/conv_fusion_op.cu
```diff
@@ -33,6 +36,44 @@ using framework::AlgorithmsCache;
 template <typename T>
 using ScalingParamType = typename platform::CudnnDataType<T>::ScalingParamType;
 
+static inline bool IsSymmetricPadding(const std::vector<int>& paddings,
```
liym27: `IsSymmetricPadding` and `PadFunction` have been defined in  file **padding.h**,  better to  call them in [**padding.h**](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/padding.h).

paddle/fluid/operators/fused/conv_fusion_op.cc
```diff
@@ -90,7 +111,8 @@ class Conv2DFusionOpInferShape : public framework::InferShapeBase {
       std::vector<framework::DDim> oshapes;
       oshapes.reserve(channels.size());
       for (size_t i = 0; i < channels.size(); ++i) {
-        oshapes.push_back({oshape[0], channels[i], oshape[2], oshape[3]});
+        oshapes.push_back(
+            {output_shape[0], channels[i], output_shape[2], output_shape[3]});
       }
       ctx->SetOutputsDim("Outputs", oshapes);
```
liym27, line 166: The computational logic of the above code is the same as [`ConvOp::InferShape`](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/conv_op.cc#L33), and there are more checks on parameters in `ConvOp::InferShape` .
Better to use `ConvOp::InferShape`.

paddle/fluid/operators/fused/conv_fusion_op.cu
zhaoyuchen2018: done

paddle/fluid/operators/fused/conv_fusion_op.cc
zhaoyuchen2018: when compute split channels need output dims, cannot call conv::infershape directly.

liym27:
LGTM

wangchaochaohu (approved):
LGTM

PR #21050, opened 2019-11-06
Add basic Python Cond Layer

zhhsplendid:
Add basic Python cond layer. Currently we only tested basic forward output

I will add test and probably change cond API in following up PRs

At least I have to add tests:

Complicated input/output
Complicated network inside branches
Backward among different branches
Different optimizers and losses inside different branches


python/paddle/fluid/layers/control_flow.py
```diff
@@ -1729,6 +1730,69 @@ def complete(self):
             })
 
 
+def copy_scope_var_to_global(var, layer_helper):
+    if var is None:
+        return None
+    global_var = layer_helper.create_global_variable(
```
sneaxiy: How about create variable inside the outer block instead of global block? Think about nested control flow ops.

zhhsplendid: Good catch! Thank you a lot for finding it which can cause buy in nested control flow. It's me who didn't consider well :-)

zhhsplendid: Done.

sneaxiy (approved):
LGTM.

lanxianghit (approved):

heavengate (approved):

Issue #16948, opened 2019-04-17
remove cases for using "x_data_format" and "y_data_format" in test_elementwise_mul_mkldnn_op


LeoZhao-BJ:
As one of the following tasks for Issue #16764, raise issue to track changes for removing cases for using "x_data_format" and "y_data_format" in test_elementwise_mul_mkldnn_op.

Sand3r-:
As I've mentioned in #16764 - removal of these attributes will make it impossible to test AVX512 code path, because currently there is no way of setting the data format of input Tensors in python. Hence, the AVX512 case would never get executed.

luotao1:
How about use model level to test these codes, you can refer to `se_resnext50` in 
https://github.com/PaddlePaddle/Paddle/blob/de3b70a101dabd48e0659ddb4df4651b882aeafa/paddle/fluid/inference/api/demo_ci/run.sh#L47

LeoZhao-BJ:
@Sand3r-  Let's sync offline and discuss a better solution to handle this case.

Sand3r-:
After discussing the possible solutions with @LeoZhao-Intel the following solutions for testing the elementwise_mul without the attributes have been proposed:
#### 1. Making it possible to set data format in python tensor.
**Advantages:**
- Simple code
- Code path tested

**Disadvantages:**
- Interfering with python, c-api bindings
- Enabling user to possibly set incorrect data format for their data
- MKL-DNN will soon remove the possibility to freely set, get data formats, so this feature would soon have to be removed
#### 2. Putting a dummy mkldnn convolution layer before elementwise_mul. The convolution would perform a conversion of data from nchw to nchw16c and feed it to eltwise_mul so that it can execute the avx512 branch.
**Advantages:**
- Code path can be tested

**Disadvantages:**
- Hacky solution, hard to understand for those who don't know the internals of conv kernel and MKL-DNN
- Can fail if MKL-DNN format selection will change
#### 3. Implementing the tests for mkldnn elementwise_mul in C++.
**Advanatages:**
- Formats can be easily set through available interface.
- Reorders make the code cleaner and easier to read
- It's easier to detect CPU features (thorugh `platform::MayIUse` or direct xbyak solution), so that nchw16c can be tested on avx512 machines only and nchw8c on avx2 (once it will be implemented).

**Disadvantages:**
- Code for performing reference computations of elementwise multiplication needs to be handcrafted (i. e. goodness of numpy can no longer be benefited from)

[...]

PR #20965, opened 2019-11-01
[WIP] Add 3-dimension ncw format and use naive version

lidanqing-intel:
Related to #20842  ernie large model issue.
Problem: Slow latency. We are trying to use 3-dimension MKLDNN mul.

Hi, sorry, I forgot I was on my own machine i9 and got slow latency. On 6148, `num_threads=10` latency is as follow. The latency with `num_threads=1` is in later comments.

```
I1101 10:09:51.044067  5401 analysis_predictor.cc:474] ======= optimize end =======
I1101 10:09:51.050020  5401 inference.cc:213] Load 10 samples from /home/lidanqing/data/test_ds_10
I1101 10:09:58.605739  5401 inference.cc:353] Run 10 samples, average latency: 755.564 ms per sample.
I1101 10:09:58.605875  5401 inference.cc:358] Run 5 samples, average latency [exclude 5 warmup steps]: 559.852
ms per sample.
```

PR #17104, opened 2019-04-25
[JIT][AVX2] Elementwise Mul

Sand3r-:
This PR introduces several things:
- It extends existing Op Testing suite in C++ to cover Functional testing
- AVX2 JIT kernel for NCHW8C x NC is introduced and tests for it are provideed
- MKL-DNN Elementwise Mul has been rewritten and simplified (it's easier to analyse when checking the whole file than by diff changes now) and now support both AVX512 and AVX2 instruction sets
- `x_format` and `y_format` attrs have been removed from Elementwise Operator, resolves #16948
- The Python UT for elementwise mul has been removed since it doesn't expose enough API to test this Op
- Instead, a C++ UT has been introduced that tests AVX512 and AVX2 implementations when the runtime architecture supports it.

PR #21061, opened 2019-11-06
Refactor MKL-DNN ElementwiseMul

### LINKED ISSUE #20842: Ernie-large模型MKLDNN库报错
- PaddlePaddle版本：develop
commit id：
屏幕快照 2019-10-28 16 10 23

- CPU型号: Intel(R) Xeon(R) Gold 6148 CPU

-  使用Docker镜像：hub.baidubce.com/paddlepaddle/paddle:latest-dev

- 模型信息
[模型：Ernie-large](https://github.com/PaddlePaddle/ERNIE)

- 错误信息：
打开MKLDNN库，报错。
关掉MKLDNN库，错误消失。

- 错误详细信息：
70ecb5ee1c2a3b8ce1edb3798508143e

 复现步骤：

/benchmark/Inference/c++/ernie/路径下：
mkdir build
cd build
cmake -DUSE_GPU=OFF -DPADDLE_ROOT=XXX ..
make

Sand3r-:
This PR fixes #20842.
The manual fallback to native Paddle kernel is removed, now that happens through GetExpetectedKernelType mechanism.
That means:
GetExpectedKernelType of ElementwiseMul checks if:
- x input is of format nChw16c
- y input is of format nc
- channels are divisable by 16
And only all those conditions are true, returns an MKL-DNN kernel. Otherwise the native Paddle kernel is returned. If that happens, all the necessary tensor format conversions that were taking place in the kernel are now delegated to the code in `data_layout_transform.cc`

Also it removes x_data_format and y_data_format attributes resolving #16948.


lidanqing-vv:
This PR is a more general solution to "Fallback to naive version" which fixes issue #20842 and PR https://github.com/PaddlePaddle/Paddle/pull/20965 is a manual way to add `ncw` fall back to naive version.

Michal @Sand3r- removed test_elementwise_mul_mkldnn_op.py, I am worried Paddle_coverage_CI may fail. But anyway, the benchmark of this PR is as follow.

Test machine:6148
num_thread=1
Data: **test_data_1000 (1000 lines form the test_ds)**

| Ernie Large        | -DWIHT_MKLDNN=OFF           | -DWIHT_MKLDNN=ON  | Improved by MKL-DNN  |
| ------------- |:-------------:|:-----:|:-----:|
| latency     | 1728.31ms |  1202.17|30.4%|

luotao1:
@Sand3r- @lidanqing-intel Do you mean use this PR instead of #20965 to fix ernie-large bug?

luotao1:
@GaoWei8 Please double-check whether this PR solve the ernie bug, thanks!

paddle/fluid/operators/elementwise/elementwise_mul_op.h
```diff
@@ -17,10 +17,89 @@ limitations under the License. */
...
 namespace paddle {
 namespace operators {
 
+class ElementwiseMulOp : public framework::OperatorWithKernel {
+ public:
+  using framework::OperatorWithKernel::OperatorWithKernel;
+  using Tensor = framework::Tensor;
+
+  void InferShape(framework::InferShapeContext* ctx) const override {
+    PADDLE_ENFORCE(ctx->HasInput("X"),
```
luotao1: PADDLE_ENFORCE is not recommended. Please use PADDLE_ENFORCE_EQ/NE/GT/GE/LT/LE or PADDLE_ENFORCE_NOT_NULL or PADDLE_ENFORCE_CUDA_SUCCESS instead.
http://ci.paddlepaddle.org/viewLog.html?buildId=214325&buildTypeId=YYTest_PrCiCpuPy2&tab=buildLog&branch_YYTest=__myBranchesProvider-myBranches__

Sand3r-: You're right, sorry I have cherry-picked the code from the old PR #17104 which used old PADDLE_ENFORCE system and forgotten to replace it. On my way to do it now.

Sand3r-:
@luotao1 FYI: PR_CI_CPU_Py2 (Compile) test failes due to removal of code containing `const_casts`.

luotao1:
Got it. Please take an internal review at first.

paddle/fluid/operators/elementwise/elementwise_mul_op.h
```diff
@@ -17,10 +17,100 @@ limitations under the License. */
 #include "paddle/fluid/operators/elementwise/elementwise_op_function.cu.h"
 #include "paddle/fluid/operators/elementwise/elementwise_op_function.h"
 #include "paddle/fluid/operators/math/blas.h"
+#include "paddle/fluid/platform/cpu_info.h"
 
 namespace paddle {
 namespace operators {
 
+class ElementwiseMulOp : public framework::OperatorWithKernel {
```
LeoZhao-BJ: why not inherit from ElementwiseOp ? seems InferShape is exactly same with ElementwiseOp 's .

paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
```diff
@@ -82,103 +72,25 @@ class ElementwiseMulMKLDNNKernel : public framework::OpKernel<T> {
...
+    const int64_t n = x_dims[0];
+    const int64_t c = x_dims[1];
+    const int64_t h = x_dims[2];
+    const int64_t w = x_dims[3];
+
+    const int64_t simd_width = 16;
+    auto multiply =
+        jit::KernelFuncs<jit::NCHW16CMulNCTuple<T>, platform::CPUPlace>::Cache()
+            .At(0);
+    ComputeBroadcastedMultiply(x_data, y_data, z_data, n, c, h, w, simd_width,
```
LeoZhao-BJ, line 90: In ComputeBroadcastedMultiply function definition,  "int simd_width", why here input parameter is int64_t ?

const int64_t simd_width = 16;

LeoZhao-BJ (changes requested):

paddle/fluid/operators/elementwise/elementwise_mul_op.h
Sand3r-: Initially I tried doing it, but encountered compile errors that I coudn't fix, so eventually I've implemented it by inheriting from OperatorWithKernel.

paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
Sand3r-: I don't recall a particular reason (probably just to match the types of variables listed above), on x86_64 architecture such parameter would be put in a 64-bit register anyway, so changing it to 32-bit int wouldn't matter much, apart from being compliant with function declaration.

LeoZhao-BJ: int is always 32bit, while int64 is 64bit, so need use "int simd_width =16"

paddle/fluid/operators/elementwise/elementwise_mul_op.h
LeoZhao-BJ: what's the compile error? better to inherit from ElementwiseOp to avoid code duplication.

Sand3r-: > better to inherit from ElementwiseOp to avoid code duplication

Totally agree.

The error can be found here:
https://pastebin.com/aQ9DTd5d

Sand3r-: I have found a solution, but it introduces changes to elementwise_op.h.
https://pastebin.com/jHDqErwp

paddle/fluid/operators/elementwise/elementwise_mul_op.h
```diff
@@ -22,56 +23,15 @@ limitations under the License. */
 namespace paddle {
 namespace operators {
 
-class ElementwiseMulOp : public framework::OperatorWithKernel {
+class ElementwiseMulOp : public ElementwiseOp {
  public:
-  using framework::OperatorWithKernel::OperatorWithKernel;
```
LeoZhao-BJ: You can just try 
```
using ElementwiseOp::ElementwiseOp;
```
and no need to extend ElementwiseOp construct func.

paddle/fluid/operators/elementwise/mkldnn/elementwise_mul_mkldnn_op.cc
Sand3r-: Whether int is 16, 32 or 64-bit is actually implementation specific, but I have modified it in recent commit as to match the function signature to which it is passed.

paddle/fluid/operators/elementwise/elementwise_mul_op.h
Sand3r-: Yup, that makes it compile just fine. I didn't know that one, thank you so much for your suggestion, will appear in the next commit.

LeoZhao-BJ:
LGTM !

GaoWei8:
> @GaoWei8 Please double-check whether this PR solve the ernie bug, thanks!

This PR can solve the ernie bug. The performance testing will be add later.

luotao1 (approved):
LGTM

paddle/fluid/operators/elementwise/elementwise_op.h
```diff
@@ -120,21 +120,6 @@ class ElementwiseOpMaker : public framework::OpProtoAndCheckerMaker {
         .EqualGreaterThan(-1);
     AddAttr<bool>("use_mkldnn", "(bool, default false). Used by MKLDNN.")
         .SetDefault(false);
-    AddAttr<std::string>(
```
luotao1, line 133: Please do not remove the attribute!
https://www.paddlepaddle.org.cn/documentation/docs/en/advanced_usage/development/new_op/op_notes_en.html#op-compatibility-issue
image

luotao1 (changes requested):
You may update the notes in `x_data_format` and `y_data_format`, said `legacy xxx`.

paddle/fluid/operators/elementwise/elementwise_op.h
Sand3r-: Well, that means that #16948 can be closed, right? I thought this issue is still relevant since it was left opened.

luotao1: I think after merge this PR, #16948 could be closed, i.e. you can remove the unit-test ？

lidanqing-vv: unit-test 已经删除了哈。因为那个unit test 是构建三个op的block，这边认为原来的UT算集成测试，不是单元测试了。而集成测试Resnext 模型本身就是了。

lidanqing-vv: > I think after merge this PR, #16948 could be closed, i.e. you can remove the unit-test ？

unit-test 已经删除了哈。因为那个unit test 是构建三个op的block，这边认为原来的UT算集成测试，不是单元测试了。而集成测试Resnext 模型本身就是了。

Sand3r-:
> You may update the notes in `x_data_format` and `y_data_format`, said `legacy xxx`.

It's done.

Sand3r-:
@luotao1 
I cannot achieve coverage in CI, because it's runs on an AVX2 machine.
```
[16:01:17] :	 [Step 1/1] -- Performing Test AVX2_FOUND
[16:01:17] :	 [Step 1/1] -- Performing Test AVX2_FOUND - Success
[16:01:17] :	 [Step 1/1] -- Performing Test AVX512F_FOUND
[16:01:17] :	 [Step 1/1] -- Performing Test AVX512F_FOUND - Failed
```

Hence, it cannot enter the AVX512 kernel due to mechanism of choosing `elementwise_mul_mkldnn_op` kernel described here:
https://github.com/PaddlePaddle/Paddle/blob/94f6ef26570158906a185f42d76cad8e6d9174e9/paddle/fluid/operators/elementwise/elementwise_mul_op.h#L55-L57
And therefore mark the lines of the kernel as tested by UT.

luotao1:
image
Got it. But the above function is not related to avx512, and it doesn't have unit-test as well.

Sand3r-:
Yup it is not, because:
https://github.com/PaddlePaddle/Paddle/blob/94f6ef26570158906a185f42d76cad8e6d9174e9/paddle/fluid/operators/elementwise/elementwise_mul_op.h#L55-L57
since the first part of boolean expression is false (`platform::MayIUse(platform::avx512f)`) then the part on the right side of `&&` is not evaluated either.

luotao1:
Thus, is it easy to add a unit-test for `AreDimsAndFormatCorrect`? And then, I will ask QA to get the exemption for other codes.

Sand3r-:
@luotao1 It required some tinkering (mainly to create ExecutionContext from scratch, I couldn't find anybody else that had to do that anywhere in the code before me), but the UT is ready 😃.

luotao1 (approved):
LGTM

luotao1:

PR #21069, opened 2019-11-07
add dlpack to imdb demo

mapingshuo:
add dlpack to cmake when compiling IMDB train demo.

If it is not added, a Compiling Error will occur.


guru4elephant (approved):
LGTM

PR #20315, opened 2019-10-09
refactor fast nan-inf debug

xyzhou-puck:
add unit-test for PR #19942

https://github.com/PaddlePaddle/Paddle/pull/19942

PR #21076, opened 2019-11-07
Rewrite check nan inf tools

wangxicoding:
modify from https://github.com/PaddlePaddle/Paddle/pull/20819
reference ideas of sum from https://github.com/PaddlePaddle/Paddle/pull/20315

1. Principle of this PR：
>  nan - nan = nan, inf - inf = nan, num - num = 0,
    nan + nan = nan, inf + inf = inf, nan + 0 = nan,
    inf + 0 = inf, nan + inf = nan, 0 + 0 = 0

    for(value:values): sum += (value-value)
    only check sum is ok

2. Speed. Test env: Resnet50, 2cards v100 16G. 
>    No check 270 images/s.  Original tools,  0.25 images/s.
    commit2: Use struct transform 128or256B, 110 images/s. Struct only transform 64B, 189 images/s.
    commit4: Use cudaMemcpyAsync in every step, 223 images/s.
    commit11: cudaMemcpyAsync in start, now is 229 images/s.

    No check 270 images/s, original tools 0.25 images/s, this tools 229 images/s, 916 times faster than before.

3. Functional. Support fp16、fp32、fp64.
Print the numel of nan,inf,num & which op, tensor NAN INF.
image
Can use environment variable to skip op or op_role or op_var
>  PADDLE_INF_NAN_SKIP_OP="op0,op1,op2"
  PADDLE_INF_NAN_SKIP_ROLE="role1,role2,role3"
  PADDLE_INF_NAN_SKIP_VAR="op0:var0,op0:var1,op1:var0"
More feature will add if future.

#### Practical results until now.
1. Fix INF bug of softmax_cross_entropy_op (https://github.com/PaddlePaddle/Paddle/pull/21165)
We use this tools find the INF bug of softmax_cross_entropy_op and solved it.
nan
2. Add coalesce_tensor to white_list.
Default, coalesce_tensor op will not init tensor, and it's not a mistake. So need to add this op to white_list.
fuse_nan
3. Debug fp16 models.
max(fp16)=65500.
In this models, in one step this op do value /= 0.7 > 65500,  then value=inf.
So add this op of this models into fp16 black_list.
fp16_nan


paddle/fluid/framework/details/var_utils.cu
```diff
@@ -0,0 +1,110 @@
...
+
+  auto* dev_ctx = reinterpret_cast<platform::CUDADeviceContext*>(
+      platform::DeviceContextPool::Instance().Get(tensor_.place()));
+
+  std::string debug_str = "[op=" + op_type_ + "] [tensor=" + var_name_ + "]";
+  auto debug_tensor = paddle::memory::Alloc(*dev_ctx, debug_str.length() + 1);
+  char* debug_ptr = reinterpret_cast<char*>(debug_tensor->ptr());
+
+  PADDLE_ENFORCE_CUDA_SUCCESS(
+      cudaMemcpyAsync(debug_ptr, debug_str.c_str(), debug_str.length() + 1,
```
wangxicoding: debug_str会析构，内存空间可能被重写覆盖，或许还是需要按之前commit的方式作为参数传入，不过会降低速度https://github.com/PaddlePaddle/Paddle/pull/21076/commits/1e862206bd7431d5d9c82d040f4a024928b39308#diff-50a4c82163a9ec5b4346a5750b73536fR36-R39
或者debug_str与op、tensor绑定

wangxicoding:
> 1. All-reduce-op handle should be checked
> 2. Inputs should be checked just like outputs because some input is feed from outside or from a saved model.
> 3. And op_role filter is needed. For fp16, Nan and inf in grad are normal, and it can process such problems.
>    So this tool should support check only forward op or minimize op and etc by op_role.

Completed the third point, the rest is ready to be added later.

paddle/fluid/framework/details/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ endif()
 
 
 if(WITH_GPU)
+    nv_library(nan_inf_utils SRCS nan_inf_utils_detail.cc nan_inf_utils_detail.cu DEPS framework_proto scope place)
```
gongweibao, line 25: nan_inf_utils_detail.cu?

paddle/fluid/framework/details/nan_inf_utils_detail.cc
```diff
@@ -0,0 +1,320 @@
...
+   * framework::OpRole::kBackward     = 0x0001,
+   * framework::OpRole::kOptimize     = 0x0002,
+   * framework::OpRole::kRPC          = 0x0004,
+   * framework::OpRole::kDist         = 0x0008,
+   * framework::OpRole::kLRSched      = 0x0010,
+   * framework::OpRole::kLoss         = 0x0100,
+   * framework::OpRole::kNotSpecified = 0x1000,
+   */
+  static const std::unordered_map<std::string, int> _role_str2int = {
+      {"forward", FORWARD}, /* kForward=0, can't filter */
```
gongweibao, line 49: 👍🏻

paddle/fluid/framework/details/nan_inf_utils_detail.cc
```diff
@@ -0,0 +1,320 @@
...
+void CheckNanInf<paddle::platform::float16>(
+    const paddle::platform::float16* value, const size_t numel, int print_num,
+    const std::string& op_type, const std::string& var_name) {
+  float sum = 0.0f;
+#pragma omp parallel for reduction(+ : sum)
+  for (size_t i = 0; i < numel; ++i) {
+    sum += static_cast<float>(value[i] - value[i]);
+  }
+
+  if (std::isnan(sum) || std::isinf(sum)) {
```
gongweibao, line 204: Interesting!

PR #21081, opened 2019-11-08
Delete test resnet50 in post train quantization 

juncaipeng:
Delete test resnet50 in post train quantization  to avoid timeout error for ci.


wzzju (approved):
LGTM

PR #21084, opened 2019-11-08
Add dependency for error_codes.proto

chenwhql:
Add dependency for error_codes.proto to fix inference CI compile faild.
image
http://ci.paddlepaddle.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=215415


luotao1 (approved):
LGTM

PR #21085, opened 2019-11-08
Use 2 cards for hallreduce unit test.

gongweibao:
The hierarchical allreduce uses 4 cards, but we only have two. So I try to load two processed on one card.


gavin1332 (approved):
LGTM

PR #20597, opened 2019-10-14
Add NHWC datalayout support on conv2d without data transformation.

gongweibao:
PR Description:

Previous conv2d only supports NCHW computation even though data format is NHWC. Transforming data format from NHWC to NCHW will be applied internally. But on Tensor Cores, conv2d is faster with NHWC format when the data is fp16, it will improve perf when we use amp.

This pr implements NHWC computation when the data is fp16 and device has Tensor Cores. The perf results are listed below:

|              | FP32 | FP16 NCHW   | FP16 NHWC   |
|--------------|------|-------------|-------------|
| ResNet50     | 0.40 | 0.195(2x)   | 0.16(2.2x)  | 
| VGG16        | 0.77 | 0.55(1.4x)  | 0.32(2.4x)  |
| Inception V4 | 0.55 | 0.30(1.83x) | 0.27(2x)    |   
| GoogleNet    | 0.12 | 0.1(1.2x)   | 0.075(1.6x) |

PR #21090, opened 2019-11-08
nhwc optimization for batchnorm

Jie-Fang:
Optimize NHWC data format for batchnorm.

Use new cudnn batchnorm api for NHWC. Combine this with NHWC optimization for conv2d, see [pr20597](https://github.com/PaddlePaddle/Paddle/pull/20597), ResNet50 improves about 15%-20% performace.


gongweibao (changes requested):
Need a UT

paddle/fluid/operators/batch_norm_op.h
```diff
@@ -39,24 +41,91 @@ template <typename T>
...
+    auto& dev_ctx = context.template device_context<DeviceContext>();
+    std::vector<int> axis{0, 2, 3, 4, 1};
+    math::Transpose<DeviceContext, T, 5> trans5;
+    trans5(dev_ctx, *input, transformed_input, axis);
+
+  } else if (dim == 2) {
+    auto& dev_ctx = context.template device_context<DeviceContext>();
+    std::vector<int> axis{0, 2, 3, 1};
+    math::Transpose<DeviceContext, T, 4> trans4;
+    trans4(dev_ctx, *input, transformed_input, axis);
```
gongweibao, line 120: else?

paddle/fluid/operators/batch_norm_op.h
```diff
@@ -39,24 +41,91 @@ template <typename T>
...
+                                const Tensor* input,
+                                Tensor* transformed_input) {
+  int dim = input->dims().size() - 2;
+  if (dim == 3) {
+    auto& dev_ctx = context.template device_context<DeviceContext>();
+    std::vector<int> axis{0, 4, 1, 2, 3};
+    math::Transpose<DeviceContext, T, 5> trans5;
+    trans5(dev_ctx, *input, transformed_input, axis);
+
+  } else if (dim == 2) {
```
gongweibao, line 93: else?

paddle/fluid/operators/batch_norm_op.cu
```diff
@@ -395,20 +531,94 @@ class BatchNormGradKernel<platform::CUDADeviceContext, T>
...
+              transformed_d_x.template mutable_data<T>(ctx.GetPlace()),
+              bn_param_desc_, scale->template data<BatchNormParamType<T>>(),
+              d_scale->template mutable_data<BatchNormParamType<T>>(
+                  ctx.GetPlace()),
+              d_bias->template mutable_data<BatchNormParamType<T>>(
+                  ctx.GetPlace()),
+              epsilon, saved_mean_data, saved_var_data));
+        }
+
+        if (data_layout_str == "NHWC" && compute_format == "NCHW") {
```
gongweibao: else?

gongweibao (changes requested):

paddle/fluid/operators/batch_norm_op.cu
Jie-Fang: else do nothing. Because when data_layout_str == "NHWC" && compute_format == "NCHW", we transform the data_format before, after computing, we should revert the data_format

Jie-Fang: else do nothing. Because when data_layout_str == "NHWC" && compute_format == "NCHW", we transform the data_format before, after computing, we should revert the data_format

paddle/fluid/operators/batch_norm_op.h
Jie-Fang: Input dims has checked by BatchNormOp::InferShape. We can ensure input dims is valid. Thus Else branch does nothing. Batchnorm op supports 2,3,4,5 dims. I add new transform logic for inputs that have 3 dims.

paddle/fluid/operators/batch_norm_op.h
Jie-Fang: Same as I mentioned before.

paddle/fluid/operators/batch_norm_op.cu
```diff
@@ -56,12 +56,37 @@ class BatchNormKernel<platform::CUDADeviceContext, T>
...
+    const bool fast_nhwc_batch_norm =
+        is_test ||
+        (dtype == CUDNN_DATA_HALF && FLAGS_cudnn_batchnorm_spatial_persistent);
+
+    auto compute_format =
+        fast_nhwc_batch_norm && data_layout_str == "NHWC" ? "NHWC" : "NCHW";
+
+    Tensor transformed_x(x->type());
+    Tensor transformed_y(y->type());
+    if (data_layout_str == "NHWC" && compute_format == "NCHW" &&
```
gongweibao: Why not use a const string?

paddle/fluid/operators/batch_norm_op.cu
```diff
@@ -90,7 +115,7 @@ class BatchNormKernel<platform::CUDADeviceContext, T>
     VLOG(3) << "Setting descriptors.";
     std::vector<int> dims;
     std::vector<int> strides;
-    if (data_layout == DataLayout::kNCHW) {
+    if (compute_format == "NCHW") {
```
gongweibao: Use a constant string?

paddle/fluid/operators/batch_norm_op.cu
```diff
@@ -158,23 +184,103 @@ class BatchNormKernel<platform::CUDADeviceContext, T>
...
+          size_t reserve_space_size = 0;
+          void *reserve_space_ptr = nullptr;
+          void *workspace_ptr = nullptr;
+          Tensor workspace_tensor;
+          // Create reserve space and workspace for batch norm.
+          // Create tensor for each batchnorm op, it will be used in the
+          // backward. Thus this tensor shouldn't be temp.
+          auto it = ctx.op().Inputs().find("X");
+          auto input_var_name = it->second[0];
+          Tensor *reserve_space_tensor =
```
gongweibao: Here is an example: https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/conv_cudnn_op.cu#L130

gongweibao (changes requested):

cmake/coveralls.cmake
```diff
@@ -62,8 +62,8 @@ function(code_coverage _COVERAGE_SRCS _COVERALLS_UPLOAD _CMAKE_SCRIPT_PATH)
 endfunction()
 
 if(WITH_COVERAGE)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O0 -fprofile-arcs -ftest-coverage")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O0 -fprofile-arcs -ftest-coverage")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O0 -fprofile-arcs -ftest-coverage -fno-inline -fno-inline-small-functions -fno-default-inline")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O0 -fprofile-arcs -ftest-coverage -fno-inline -fno-inline-small-functions -fno-default-inline")
```
sneaxiy: Why set these flags?

python/paddle/fluid/layers/nn.py
```diff
@@ -2523,6 +2523,12 @@ def get_decay_momentum(momentum_init, decay_steps, decay_rate):
     check_type_and_dtype(input, 'input', Variable,
                          ['float16', 'float32', 'float64'], 'batch_norm')
     dtype = helper.input_dtype()
+
+    has_reserve_space = False
+    if data_layout == 'NHWC' and os.environ.get(
```
sneaxiy: `os.environ.get('FLAGS_cudnn_batchnorm_spatial_persistent')` would return a string or bool. So if `FLAGS_cudnn_batchnorm_spatial_persistent=false`, this condition would return true before string `"false"` is true in Python.

cmake/coveralls.cmake
gongweibao: To fix the coverage potential bug.

python/paddle/fluid/layers/nn.py
gongweibao: Done.Thanks.

cmake/coveralls.cmake
gongweibao: Done.

sneaxiy (approved):
LG op.

gongweibao (approved):
LGTM

PR #21091, opened 2019-11-08
Add friendly dygraph trace API

sneaxiy:
This PR add `TracedLayer` interface to convert the dygraph model to static graph model.

Would expose this API and add doc in the next PR.


python/paddle/fluid/dygraph/jit.py
```diff
@@ -95,7 +100,7 @@ def forward(self, input):
                 layer = ExampleLayer("example_layer")
                 in_np = np.random.random([2, 3]).astype('float32')
                 in_var = to_variable(in_np)
-                out, program = jit.trace(layer, inputs=[in_var],
+                out, program, _, _ = jit._trace(layer, inputs=[in_var],
```
JiabinYang, line 103: better not show a private interface to user?

JiabinYang (commented):
just one comment

python/paddle/fluid/dygraph/jit.py
sneaxiy: Yes, I agree. `_trace` is not inside `__all__` of `jit.py`. So it is not a public interface. The only public interface is `TracedLayer`.

JiabinYang (approved):
LGTM

PR #21092, opened 2019-11-08
Remove useless code/message of openblas and fix the previous hard-code

zhwesky2010:
Fix some problems of and **cblas.cmake** **openblas.cmake** file:

**openblas.cmake** has these problem: 
 - 1. The previous error message no longer applies to the current version and is incorrect;
 - 2. Some useless code doesn't work, which should be removed;
 a. include_directories(${THIRD_PARTY_PATH}/install) should not be here.
 b. the import library(openblas) has never been used and have no meaning.
 c. add_custom_command for extern_openblas have no meaning, because the library has been linked to cblas.
 - 3. No indentation in the code;

**cblas.cmake** has these problem:
- The execution sequence of finding and building cblas is moved into **cbals.cmake**, and **openblas.cmake** is only an external project and no longer contains other code.

### DIFF
--- a/cmake/cblas.cmake
+++ b/cmake/cblas.cmake
@@ -1,3 +1,17 @@
+# Copyright (c) 2016 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


liupluswei (approved):
LGTM

PR #21099, opened 2019-11-09
Optimizer memcpy when _rand_len=16 and Remove dataCopy in GradKernel

Aurelius84:
Optimized some code related  data copy to slightly increase speed.

Apply testing cost time in each epoch while training with same data:
**before : after = 526.8s : 497s**


luotao1 (approved):
LGTM

PR #21105, opened 2019-11-11
Remove so many logs of memory optimization in parallel_executor.cc

sneaxiy:
Use `LOG_FIRST_N(INFO, 1)` instead of `LOG(INFO)` to avoid too many logs when too many `ParallelExecutor` creates.


chenwhql (approved):
LGTM

PR #21114, opened 2019-11-11
Add control flow api: case

liym27:
- ### Add a control flow API: `case`
```
case(pred_fn_pairs, default=None, name=None)
```
> #### `case` works like an `if-elif-elif-else` chain.
- ### API document
- Chinese Docment: https://github.com/PaddlePaddle/FluidDoc/pull/1629

- English docment:
image

### DIFF
--- a/python/paddle/fluid/layers/control_flow.py
+++ b/python/paddle/fluid/layers/control_flow.py
@@ -1798,14 +1798,15 @@ def cond(pred, true_fn=None, false_fn=None, name=None):
 
 
 def _error_message(what, arg_name, op_name, right_value, error_value):
-    raise TypeError(
-        "{what} of '{arg_name}' in Op({op_name}) "
-        "must be {right_value}, but received: {error_value}.".format(
-            what=what,
-            arg_name=arg_name,
-            op_name=op_name,
-            right_value=right_value,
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/layers/control_flow.py
```diff
@@ -1797,6 +1797,131 @@ def cond(pred, true_fn=None, false_fn=None, name=None):
     return merged_output
 
 
+def _error_message(what, arg_name, op_name, right_value, error_value):
```
zhhsplendid, line 1801: I saw you raise TypeError in _error_message and also raise at where you use _error_message, that's not good.

Two options:

Don't raise TypeError in _error_message, but return the error message as a string in _error_message

Remove "raise TypeError" here, rename "_error_message" to raise_error_with_message or some other name properly.

liym27: Thanks. Done according to option 1.

zhhsplendid (dismissed):
LGTM now

heavengate (dismissed):
LG API

lanxianghit (dismissed):

lanxianghit (approved):

heavengate (approved):

PR #21118, opened 2019-11-11
Fix the error of init variable in StaticRNN when stop_gradient=ON

gfwm2013:
Fix the error of init variable in StaticRNN when stop_gradient=ON

### DIFF
--- a/paddle/fluid/operators/recurrent_op.cc
+++ b/paddle/fluid/operators/recurrent_op.cc
@@ -636,10 +636,6 @@ class RecurrentGradOpShapeInference : public framework::InferShapeBase {
 
     // In some case the kInitialStates is empty.
     if (ctx->HasInputs(RecurrentBase::kInitialStates)) {
-      PADDLE_ENFORCE_EQ(ctx->HasOutputs(framework::GradVarName(
-                            RecurrentBase::kInitialStates)),
-                        true, "The output of(%s) should not be empty.",
-                        framework::GradVarName(RecurrentBase::kInitialStates));
       ctx->SetOutputsDim(framework::GradVarName(RecurrentBase::kInitialStates),
                          ctx->GetInputsDim(RecurrentBase::kInitialStates));
     }
--- a/paddle/fluid/operators/recurrent_op.cc
+++ b/paddle/fluid/operators/recurrent_op.cc
@@ -635,7 +635,9 @@ class RecurrentGradOpShapeInference : public framework::InferShapeBase {
                       RecurrentBase::kOutputs);
 
     // In some case the kInitialStates is empty.
-    if (ctx->HasInputs(RecurrentBase::kInitialStates)) {
+    if (ctx->HasInputs(RecurrentBase::kInitialStates) &&
+        ctx->HasOutputs(
+            framework::GradVarName(RecurrentBase::kInitialStates))) {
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/recurrent_op.cc
```diff
@@ -636,10 +636,6 @@ class RecurrentGradOpShapeInference : public framework::InferShapeBase {
 
     // In some case the kInitialStates is empty.
     if (ctx->HasInputs(RecurrentBase::kInitialStates)) {
-      PADDLE_ENFORCE_EQ(ctx->HasOutputs(framework::GradVarName(
-                            RecurrentBase::kInitialStates)),
-                        true, "The output of(%s) should not be empty.",
-                        framework::GradVarName(RecurrentBase::kInitialStates));
       ctx->SetOutputsDim(framework::GradVarName(RecurrentBase::kInitialStates),
```
zhhsplendid, line 641: Although I asked you to delete this PADDLE_ENFORCE_EQ line, you should also read some lines above and below. The lines below are Potential dangerous code.

If there is no GradVarName, you shouldn't SetOutputsDim for the grad of InitialStates. You should change the "if" condition to

if (ctx->HasInputs(RecurrentBase::kInitialStates) && ctx->HasOutputs(framework::GradVarName(RecurrentBase::kInitialStates))  {
    SetOutputsDim ...
}

python/paddle/fluid/tests/unittests/test_recurrent_op.py
```diff
@@ -124,6 +124,7 @@ def setup_program(self):
     def setUp(self):
         self.setup_program()
         self.data_field = {"x", "h_boot"}
```
zhhsplendid: Their names are not good ... if you are a future developer who saw those lines, you will ask what's data_field, what's fetch_data_field? Why are they same in most cases? Also, if fetch_data_field is used for fetching, why do we still use data_field in "def forward(self)"....

I would suggest to change the name "data_field" to "feed_data_field"
"fetch_data_field" to "grad_data_field". In old test cases, you should write "self.grad_data_field = self.feed_data_field"

In the test case you added, you should comment "init doesn't have gradient"

python/paddle/fluid/tests/unittests/test_recurrent_op.py
```diff
@@ -602,5 +607,76 @@ def dot_attention(query, memory):
         return rnn()
 
 
+class RecurrentOpTest3(RecurrentOpTest1):
```
zhhsplendid: I guess you were copying the old test class name. In fact, var1, var2, class1, class2 are very bad class names ...... you should do the best from new code.

You can write "RecurrentOpStopGradientTest"

paddle/fluid/operators/recurrent_op.cc
gfwm2013: Done.

python/paddle/fluid/tests/unittests/test_recurrent_op.py
gfwm2013: Done.

python/paddle/fluid/tests/unittests/test_recurrent_op.py
gfwm2013: Done.

zhhsplendid (approved):
LGTM

PR #21131, opened 2019-11-12
remove build_once & name_scope

songyouwei:
## Motivation

In dygraph, most `Parameter`s of `Layer`s are created when the `__call__` is first invoked, thus part of the shape of the `Parameter`s can be derived from the input, which is similar to paddle's static graph API.
This design pattern introduces many problems. The model depends on data, when a complex model (a subclass of layer) is inited, there are no parameters in it, and we can't load the previously saved checkpoint before a `__call__`. It also brings troubles to the initialization of custom Parameters.
The creation of `Layer` also requires the `name_scope` parameter, which is unnecessary and troublesome for users.

This PR removes `build_once` function and `name_scope` parameter.

### `fluid.dygraph.Layer`
1. `name_scope` parameter is not required in `Layer`'s `__init__`
2. simplify `create_parameter` api, only `shape` is required

### `__init__` of `fluid.dygraph.nn.*`
1. For all subclasses of `Layer`: `name_scope`s are removed, all missing optional `dtype` parameters are added.
1. `Conv2D`, `Conv2DTranspose`, `Conv3D` and `Conv3DTranspose`: add `num_channels` (int): The number of channels in the input image.
1. `LayerNorm`: remove `begin_norm_axis`, add `normalized_shape` (int or list or tuple): Input shape from an expected input of size :math:`[*, normalized_shape[0], normalized_shape[1], ..., normalized_shape[-1]]`. If it is a single integer, this module will normalize over the last dimension which is expected to be of that specific size.
1. `NCE`: add `dim` (int): Dimension of input (possibly embedding dim).
1. `PRelu`: add `input_shape`(list or tuple, optional): The shape of input. This parameter is required when mode is not "all". Default: None.
1. `BilinearTensorProduct`: remove `size`. add `input1_dim` (int): The dimension of each first input. add `input2_dim` (int): The dimension of each second input.
1. `GroupNorm`: add `channels` (int): The number of channels of input.
1. `SpectralNorm`: add `weight_shape` (list or tuple): The shape of weight parameter.
1. `TreeConv`: add `feature_size` (int): last dimension of nodes_vector.

### For existing subclasses of `Layer`:
1. remove `build_once` function, and move `Parameter` creations into `__init__`.
2. avoid positional arguments for `Layer.create_parameter` function, and only `shape` is required.

### For constructions of `Layer`'s subclasses in dygraph models:
1. remove `name_scope` parameter of `__init__`
2. add possible new parameters of `__init__` (these parameters are input shape related)

#### Example (Conv2D)
**before**
```python
import paddle.fluid as fluid
import numpy as np
data = np.random.uniform(-1, 1, [10, 3, 32, 32]).astype('float32')
with fluid.dygraph.guard():
    data = fluid.to_variable(data)
    conv2d = fluid.Conv2D("conv2d", 2, 3)
    conv = conv2d(data)
```

**after**
```python
import paddle.fluid as fluid
import numpy as np
data = np.random.uniform(-1, 1, [10, 3, 32, 32]).astype('float32')
with fluid.dygraph.guard():
    data = fluid.to_variable(data)
    conv2d = fluid.Conv2D(3, 2, 3)  # "conv2d" removed, and 3 is num_channels
    conv = conv2d(data)
```

### DIFF
--- a/python/paddle/fluid/dygraph/layers.py
+++ b/python/paddle/fluid/dygraph/layers.py
@@ -36,7 +36,7 @@ class Layer(core.Layer):
     Parameters:
         name_scope (str, optional): prefix name used by the layer to name parameters.
             If prefix is "my_layer", parameter name in MyLayer
-            can be "my_layer_0.w_n", where w is the parameter
+            can be "mylayer_0.w_n", where w is the parameter
             base name and n is an unique suffix auto-generated.
             If None, prefix name will be lower cased class name. Default: None.
         dtype(str or core.VarDesc.VarType, optional): data type of this parameter.
@@ -51,7 +51,11 @@ class Layer(core.Layer):
     def __init__(self, name_scope=None, dtype=core.VarDesc.VarType.FP32):
         if name_scope is None:
             name_scope = self.__class__.__name__.lower()
-        self._full_name = unique_name.generate(name_scope)
+            self._full_name = unique_name.generate(name_scope)
+        else:
+            # TODO: remove name_scope parameter and all hard-coded usages
+            self._full_name = unique_name.generate(name_scope + "/" +
+                                                   self.__class__.__name__)
         self._helper = LayerObjectHelper(self._full_name)
         self._built = False
         self._dtype = dtype
--- a/python/paddle/fluid/dygraph/layers.py
+++ b/python/paddle/fluid/dygraph/layers.py
@@ -34,10 +34,11 @@ class Layer(core.Layer):
     """Dynamic graph Layer based on OOD, includes the parameters of the layer, the structure of the forward graph and so on.
 
     Parameters:
-        name_scope (str): prefix name used by the layer to name parameters.
-            If prefix is "my_model/layer_1", parameter name in MyLayer
-            can be "my_model/layer_1/MyLayer/w_n", where w is the parameter
+        name_scope (str, optional): prefix name used by the layer to name parameters.
+            If prefix is "my_layer", parameter name in MyLayer
+            can be "my_layer_0.w_n", where w is the parameter
             base name and n is an unique suffix auto-generated.
+            If None, prefix name will be lower cased class name. Default: None.
         dtype(str or core.VarDesc.VarType, optional): data type of this parameter.
                 If set str, it can be "bool",  "float16", "float32", "float64",
                 "int8", "int16", "int32", "int64", "uint8" or "uint16".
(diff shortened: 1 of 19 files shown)


python/paddle/fluid/dygraph/layers.py
```diff
@@ -47,17 +48,18 @@ class Layer(core.Layer):
         None
     """
 
-    def __init__(self, name_scope, dtype=core.VarDesc.VarType.FP32):
-        self._full_name = unique_name.generate(name_scope + "/" +
-                                               self.__class__.__name__)
+    def __init__(self, name_scope=None, dtype=core.VarDesc.VarType.FP32):
```
phlrain, line 51: 这个地方保留 name_scope，是为了兼容性的考虑么

python/paddle/fluid/dygraph/nn.py
```diff
@@ -177,25 +178,21 @@ def __init__(self,
...
-        # if (self._num_channels == self._groups and
-        #         num_filters % self._num_channels == 0 and not self._use_cudnn):
-        #     self._l_type = 'depthwise_conv2d'
-        # else:
-        # TODO(jiabin): recover the usage of depthwise_conv2d when it's
-        #  kernel fixed https://github.com/PaddlePaddle/Paddle/issues/17275
-        self._l_type = 'conv2d'
+        if (self._num_channels == self._groups and
+                num_filters % self._num_channels == 0 and not self._use_cudnn):
+            self._l_type = 'depthwise_conv2d'
```
phlrain, line 184: 这个地方 depthwise_conv2d 打开，确认会不会有bug，之前是关闭的状态

python/paddle/fluid/dygraph/layers.py
songyouwei: 是的，后面改完了应该移除

python/paddle/fluid/dygraph/nn.py
songyouwei: 应该已经在 PR #18582 中Fix了

lanxianghit:
PR描述中请补充对用户使用方式改变的示例，可以找个典型OP，说明改之前和之后用法的区别，相应的影响是什么，是否有需要用户注意的地方

songyouwei:
> PR描述中请补充对用户使用方式改变的示例，可以找个典型OP，说明改之前和之后用法的区别，相应的影响是什么，是否有需要用户注意的地方

songyouwei:
@lanxianghit 已补充说明和示例

python/paddle/fluid/dygraph/layers.py
iclementine: 要不要把 dtype 一起去了，反正有可能有混合精度的 Layer。以及考虑到以后可能有以下的写法

```python
class Model(Layer):
    def __init__(self, encoder, decoder):
```
model 对于已经创建出来的 encoder 和 decoder 的 dtype 是没有控制力的，所以要不在基类中去掉它？

lanxianghit (approved):

saxon-zh (approved):

PR #21144, opened 2019-11-12
Add relative error measure when (value > 1)

grygielski:
This PR resolves problem with tests failing when compared values are large in magnitude because checking for absolute error between them is often wrong due to floating point calculation precision.

### DIFF
--- a/paddle/fluid/inference/tests/api/tester_helper.h
+++ b/paddle/fluid/inference/tests/api/tester_helper.h
@@ -92,6 +92,14 @@ void PrintConfig(const PaddlePredictor::Config *config, bool use_analysis) {
   LOG(INFO) << analysis_config->ToNativeConfig();
 }
 
+void CheckError(float data_ref, float data) {
+  if (std::abs(data_ref) > 1) {
+    CHECK_LE(std::abs((data_ref - data) / data_ref), FLAGS_accuracy);
+  } else {
+    CHECK_LE(std::abs(data_ref - data), FLAGS_accuracy);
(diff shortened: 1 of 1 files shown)


grygielski:
@luotao1 @bingyanghuang @Sand3r- Please review and share your thoughts on this solution. I used same accuracy threshold for both relative and absolute errors.

paddle/fluid/inference/tests/api/tester_helper.h
```diff
@@ -169,7 +174,12 @@ void CompareResult(const std::vector<PaddleTensor> &outputs,
         float *pdata_ref = ref_out.data<float>(&place, &ref_size);
         EXPECT_EQ(size, ref_size);
         for (size_t j = 0; j < size; ++j) {
-          CHECK_LE(std::abs(pdata_ref[j] - pdata[j]), FLAGS_accuracy);
+          if (std::abs(pdata_ref[j]) > 1) {
+            CHECK_LE(std::abs((pdata_ref[j] - pdata[j]) / pdata_ref[j]),
+                     FLAGS_accuracy);
+          } else {
+            CHECK_LE(std::abs(pdata_ref[j] - pdata[j]), FLAGS_accuracy);
+          }
```
luotao1: Could you create a function to do this? It seems duplicated codes.

grygielski: Sure, I'll do that

Sand3r- (approved):
LGTM

PR #21159, opened 2019-11-13
Remove some temporary variables

yihuaxu:
For performance requirement, now we remove some temporary tensors for fused_emb_seq_pool operator.


luotao1:
Close due to no speedup on PaddleCloud.

PR #21169, opened 2019-11-13
add unused input vars check for OpWithKernel, test=develop

zhiqiu:
In Paddle, there are several operators that registered some variables as inputs while never use them during computation in kernel. For example, `instance_norm_grad` registered `kernel` in `InstanceNormGradMaker`, which is not used in `InstanceNormGradKernel`.
This could have two problems,
1. Unnecessary memory consumption, especially GPU memory. Paddle uses Garbage Collection(GC) to save GPU memory, and GC relies on reference count calculation. When variables are registered as inputs, their reference count increased, thus these unused variables' memory can not be released.
2. This makes the code confusing.

In order to reduce unused variables, this PR, 
1. Count variable usage in `InputVar`, `MultiInput`, `MultiInputVar `, compare them with op's inputs registered, and report unused variables during operator running.
2. Add `unused_var_check` to CI job. (PR_CI_Coverage)
3. Keep a white list which contains the existing operators that have unused variables. These operators should be fixed in the future.

See [wiki](https://github.com/PaddlePaddle/Paddle/wiki/OP-Should-Not-Have-Unused-Input) for more details.


paddle/scripts/paddle_build.sh
```diff
@@ -436,8 +436,9 @@ EOF
         fi
 
         # TODO: jiabin need to refine this part when these tests fixed on mac
+        # NOTE(cql): enable unused_var_check for MAC CI
```
luotao1: 438行可以去掉了吧。439行请写具体一点，cql用完整的github名字，不然无法对应。

zhiqiu: 好的

paddle/fluid/framework/operator.cc
```diff
@@ -428,6 +429,8 @@ bool ExecutionContext::HasOutput(const std::string& name) const {
 }
 
 const Variable* ExecutionContext::InputVar(const std::string& name) const {
+  LogVarUsageIfUnusedVarCheckEnabled(name);
```
sneaxiy, line 439: Call this function with `if (FLAGS_enable_unused_var_check)`?

paddle/fluid/framework/operator.cc
```diff
@@ -457,6 +460,8 @@ const Tensor* ExecutionContext::Input<Tensor>(const std::string& name) const {
 template <>
 const std::vector<const Tensor*> ExecutionContext::MultiInput<Tensor>(
     const std::string& name) const {
+  LogVarUsageIfUnusedVarCheckEnabled(name);
```
sneaxiy, line 470: Call this function with `if (FLAGS_enable_unused_var_check)`?

paddle/fluid/framework/operator.h
```diff
@@ -268,6 +269,8 @@ class ExecutionContext {
 
   const std::vector<const Variable*> MultiInputVar(
       const std::string& name) const {
+    LogVarUsageIfUnusedVarCheckEnabled(name);
```
sneaxiy, line 294: Same above.

paddle/fluid/framework/operator.h
```diff
@@ -298,6 +301,8 @@ class ExecutionContext {
 
   template <typename T>
   const std::vector<const T*> MultiInput(const std::string& name) const {
+    LogVarUsageIfUnusedVarCheckEnabled(name);
```
sneaxiy, line 336: Same above.

paddle/fluid/framework/operator.h
```diff
@@ -349,6 +354,7 @@ class ExecutionContext {
 
   //! Get actual name vector for this input.
   const std::vector<std::string>& Inputs(const std::string& name) const {
+    LogVarUsageIfUnusedVarCheckEnabled(name);
```
sneaxiy: Same above.

paddle/fluid/framework/unused_var_check.cc
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+  if (!unsed_input_var_names.empty()) {
+    std::string err_msg = "Operator " + op.Type() + " has input(s) not uesed: ";
+    for (auto &in_var_name : unsed_input_var_names) {
+      err_msg += in_var_name;
+      err_msg += ", ";
+    }
+    err_msg +=
+        "please make sure it(them) is(are) needed. If not, remove it(them) "
```
sneaxiy: Typo: `(them)` -> `(they)`?

paddle/fluid/framework/unused_var_check.cc
```diff
@@ -0,0 +1,125 @@
...
+        "from inputs; if yes, register NoNeedBufferVars or add the operator to "
```
sneaxiy: register NoNeedBufferVars or remove them in the GradOpMaker or add the operator to white list in unused_var_check.cc?

paddle/fluid/framework/unused_var_check.cc
```diff
@@ -0,0 +1,125 @@
...
+        "white list in unused_var_check.cc.";
+    // VLOG(1) << err_msg;
```
sneaxiy: Remove unused comments.

paddle/fluid/framework/unused_var_check.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+
+#include <string>
+#include <unordered_set>
+#include "paddle/fluid/framework/operator.h"
+
+DECLARE_bool(enable_unused_var_check);
+DECLARE_bool(use_mkldnn);
```
sneaxiy: Not important: `DECLARE_xxx` here is not necessary in header file. You can move them to `.cc`.

paddle/fluid/operators/flatten_op.cc
```diff
@@ -234,6 +234,8 @@ DECLARE_INPLACE_OP_INFERER(FlattenOpInplaceInToOut, {"X", "Out"});
 DECLARE_INPLACE_OP_INFERER(FlattenGradInplaceinToOut,
                            {framework::GradVarName("Out"),
                             framework::GradVarName("X")});
+DECLARE_NO_NEED_BUFFER_VARS_INFERENCE(FlattenGradNoNeedBufferVarsInference,
```
sneaxiy, line 253: Why no need `Out`? Maybe it is `X` that is not needed?

paddle/fluid/operators/squeeze_op.cc
```diff
@@ -274,7 +274,8 @@ DECLARE_INPLACE_OP_INFERER(SequeezeInplaceInferer, {"X", "Out"});
 DECLARE_INPLACE_OP_INFERER(SequeezeGradInplaceInferer,
                            {framework::GradVarName("Out"),
                             framework::GradVarName("X")});
-
+DECLARE_NO_NEED_BUFFER_VARS_INFERENCE(SqueezeGradNoNeedBufferVarsInference,
```
sneaxiy, line 293: Same above.

paddle/fluid/framework/operator.cc
zhiqiu: Thanks. To reduce code duplication,  I will put the `if`  statement inside `LogVarUsageIfUnusedVarCheckEnabled`.

paddle/fluid/framework/unused_var_check.cc
zhiqiu: Done, thanks.

paddle/fluid/framework/unused_var_check.cc
zhiqiu: Discussed offline. I use `if` to discussed failure of unused_var_check on two case, 

1. If the `input var` is not need in computation, the only fix suggestion is "remove them in the inputs."
2. If the `input var` is really need in computation but failed on check (due to some reasons, like different kernels, conditional usage, or only needing dims), the fix suggestion is "register NoNeedBufferVars or add the operator to white list in unused_var_check.cc".

paddle/fluid/framework/unused_var_check.cc
zhiqiu: Done, thanks.

paddle/fluid/framework/operator.cc
zhiqiu: Thanks. To reduce code duplication, I will put the if statement inside LogVarUsageIfUnusedVarCheckEnabled.

paddle/fluid/framework/operator.h
zhiqiu: Done, thanks.

paddle/fluid/framework/operator.h
zhiqiu: Done, thanks.

paddle/fluid/framework/operator.h
zhiqiu: Done, thanks.

paddle/fluid/framework/unused_var_check.h
zhiqiu: I remove `DECLARE_bool(use_mkldnn)` since it is not used, and keep `DECLARE_bool(enable_unused_var_check)` because some `.cc` file will include this header file and use the `FLAGS`.

tools/check_api_approvals.sh
```diff
@@ -89,6 +90,9 @@ for API_FILE in ${API_FILES[*]}; do
       elif [ "${API_FILE}" == "paddle/fluid/operators/distributed/send_recv.proto.in" ];then
           echo_line="You must have one RD (gongweibao or seiriosPlus) approval for the paddle/fluid/operators/distributed/send_recv.proto.in, which manages the environment variables.\n"
           check_approval 1 10721757 5442383
+      elif [ "${API_FILE}" == "paddle/fluid/framework/unused_var_check.cc" ];then
+          echo_line="You must have one RD (zhiqiu (Recommend) , sneaxiy or luotao1) approval for the paddle/fluid/framework/unused_var_check.cc, which manages the white list of operators that have unused input variables.\n"
```
luotao1: You should have a wiki to tell developers what's the problem, and how to fix this problem, otherwise, they don't understand it, and will directly add to the write list.

luotao1 (dismissed):
LGTM

sneaxiy (dismissed):
LGTM with 2 things to do in the next PR:

1. Remove `InitThreadLocalUsedVarNameSet` in `unused_var_check.h` since it is not used and not defined.

2. Add custom grad op maker for `squeeze` and `flatten` to remove `Out` dependency.

sneaxiy (approved):
Excellent!

luotao1 (approved):
LGTM

PR #21189, opened 2019-11-14
Fix current block in math_patch_op

zhhsplendid:
Fix the var in math_patch_op doesn't return correct current block, which can cause + - * / error in control flow ops.

This PR also passed CE test:

http://ce.paddlepaddle.org:8080/viewLog.html?buildId=71796&tab=buildResultsDiv&buildTypeId=PaddleModesl_ModelDebug_Model11Build


sneaxiy (approved):
LGTM.

PR #21192, opened 2019-11-14
[MKL-DNN] Support for NHWC data_format of Pool,  LRN , conv2d and conv2d transpose ops (FWD only)

jczaja:
This PR is adding for NHWC data_format support to Pool and LRN MKL-DNN  Ops (FWD only) as disscused in #20964 .  

It was only tested on implemented UTs and some more testing is needed hence it is **WIP**


jczaja:
Functionality presented here was merged by other PRs

PR #21197, opened 2019-11-15
add the framework support for distfc

sandyhouse:
For distfc (distributed classification for large scale classfication), we have to set some key-value information for the loss variable, so we add the corresponding internal _set_info and _get_info methods in framework.py.

In addition, we fix the implementation of shard_index_op so that the number of classes shouldn't be an integral multiplication of number of ranks.


gavin1332 (approved):
LGTM

luotao1 (approved):
LGTM for the update on framework.py

PR #21206, opened 2019-11-15
use prefetch to load next mem into cache

LeoZhao-BJ:
memcpy usage in hash_embedding_ff of Pyramid DNN is a typical random memory access case, which makes cache miss frequently. Before each memcpy, it uses a hash func to calculate next position for mem read, this position is not continuous, so each time L1 cache need refresh and access cache line to read next mem block, it costs time, and make memcpy perf bad.

The good thing is the hashed position is predictable, we can use XXH32() to predict next position in last memcpy, by using prefetch(), we can let system preload next mem into cache during current memcpy, it saves time and reduce cache miss rate.
   
test=develop

### DIFF
--- a/paddle/fluid/operators/pyramid_hash_op.cc
+++ b/paddle/fluid/operators/pyramid_hash_op.cc
@@ -172,12 +172,8 @@ class CPUPyramidHashOPKernel : public framework::OpKernel<T> {
 
       unsigned int pos3 =
           XXH32(hash_id, len * sizeof(T), j + 2 * _rand_len) % _space_len;
-      if (_rand_len == 16) {
-        memcpy(top_pos + j, const_cast<float*>(weights + pos1), 16 * sizeof(T));
-      } else {
-        memcpy(top_pos + j, const_cast<float*>(weights + pos1),
-               _rand_len * sizeof(T));
-      }
+      memcpy(top_pos + j, const_cast<float*>(weights + pos1),
+             _rand_len * sizeof(T));
       pos1 = pos2;
       pos2 = pos3;
     }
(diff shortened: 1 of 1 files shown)


LeoZhao-BJ:
https://github.com/PaddlePaddle/benchmark/issues/151

luotao1:
I test it on pyramid_dnn training, see https://github.com/PaddlePaddle/benchmark/issues/151#issuecomment-552296437 for details. 

|-|single thread on 6148 (s/epoch)|24 threads on 6148(s/epoch)|
|---|---|---|
|before this PR|381|495|
|after this PR|364(+4.5%)|485.7(+1.8%)|
|remove copy 16 float|367.6|484.7|

paddle/fluid/operators/pyramid_hash_op.cc
```diff
@@ -161,14 +161,25 @@ class CPUPyramidHashOPKernel : public framework::OpKernel<T> {
...
+      unsigned int pos3 =
+          XXH32(hash_id, len * sizeof(T), j + 2 * _rand_len) % _space_len;
       if (_rand_len == 16) {
-        memcpy(top_pos + j, const_cast<float*>(weights + pos), 16 * sizeof(T));
+        memcpy(top_pos + j, const_cast<float*>(weights + pos1), 16 * sizeof(T));
       } else {
-        memcpy(top_pos + j, const_cast<float*>(weights + pos),
+        memcpy(top_pos + j, const_cast<float*>(weights + pos1),
                _rand_len * sizeof(T));
       }
```
luotao1: The benchmark result is in https://github.com/PaddlePaddle/Paddle/pull/21206#issuecomment-557553820
It seems line175-177 could be removed

LeoZhao-BJ: agree.

luotao1 (approved):
LGTM

luotao1:
Please update the PR description for more details.

PR #21212, opened 2019-11-17
fix fs_client_param bug

xjqbest:
* fix fs_client_param bug， user can set this config through fleet_desc_file or fleet config
* test=develop


yaoxuefeng6 (approved):
LGTM

PR #21222, opened 2019-11-18
Change GCC version to be 8.2 in Dockerfile.GCC8

sneaxiy:
Change GCC version in `Dockerfile.GCC8` from gcc 8.3 to gcc 8.2.


luotao1 (approved):
LGTM

PR #21231, opened 2019-11-19
[Bug fix] paddle-TRT set duplicate weight name into cpu

cryoco:
**Problem**: previously added suffix number is mixed with earlier suffix number added in ir passes, which may cause conflict in very rare situation.
e.g.: `conv_bn_fuse/eltwise_add_in/10129` is both the 9th `conv_bn_fuse/eltwise_add_in/1012` and the 29th `conv_bn_fuse/eltwise_add_in/101`
**Solution**: added splitter "__" between weight name and suffix number to avoid conflicts.


NHZlX (approved):
LGTM

PR #21240, opened 2019-11-19
Fix the CAPI ZeroCopy shape error and reuse the code to get output

FrostML:
Fix the CAPI ZeroCopy shape error and reconstruct the method to get output.

* Before this fix, the shape data return to the user is wrong number and the shape size is wrong, too. Now, it not just assigns the value to shape, but use memmove to actually state a memory for shape. And problem solved.

* Reconstruct the method to obtain the predictor's output to simplify the code and make it can be reused by different output type and increase the file's unit tests coverages.

* Fix the unit tests' template to float. Since the output of typeid(T).name() is different from Linux and Windows, the unit tests may not be suitable for both platform.


fc500110 (approved):
LGTM

PR #21242, opened 2019-11-19
Feature/BufferSharedIdentityInplaceOpPass

sneaxiy:
# Background
This PR enhances in-place strategy further.

Some operators do not change the input data when in-place is performed. These operators (we call them "identity ops") include: 
- reshape, reshape_grad and reshape_grad_grad
- squeeze and squeeze_grad
- unsqueeze and unsqueeze_grad
- flatten and flatten_grad
- assign and its grad (the grad of assign is assign itself)
- ...

Suppose that `X` is input of op1 and op2, where op1 is an identity op and op2 is a non-identity op.
In-place can be performed in op1 safely because op1 would not change the data of `X` even though in-place is performed. This PR adds `BufferSharedIdentityInplaceOpPass` to enable this kind of in-place strategy.

For-example,
```
x2 = fluid.layers.reshape(x1, ...)
x3 = non_identity_op(x1)
```

Although x1 is input of two ops, in-place can be performed safely when running reshape.

# Design of BufferSharedIdentityInplaceOpPass
## 1. Do not consider last lived ops only
Suppose that we have the network like:
```
x2 = fluid.layers.reshape(x1, ...)
x3 = op2(x2)
x4 = op3(x1, x3)
```
It is obvious that the last lived ops of `x1` are `[op3]` only (because `reshape` is strictly before `op3` in the graph). However, since the last version of `x1` would be only read in `reshape` and `op3`, `x1` can be still identity inplaced in this case. Therefore, in the implementation of this PR, we would check whether the last lived ops of `x1` only read `x1`. If so, we would scan all ops that read `x1` to find out the identity inplace ops instead of only scanning last lived ops of `x1`.

## 2. Non-branched identity inplace and branched identity inplace
There are two kinds of identity inplace reuse:
- Non-branched identity inplace: input `X` may be only reused by one output var, such as:
```
X -> reshape -> Y1 -> squeeze -> Y2 -> op1 -> ...
```
In-place can be performed in `reshape` and `squeeze` no matter `op1` is a non-identity inplace op (i.e, `relu`) or not.

- Branched identity inplace: input `X` may be reused by many output vars. Branched non-identity inplace is not allowed, since these ops would change data of input, but branched identity inplace is allowed. For example,
```
       -> reshape -> Y1 -> op1 -> ...
X -  |
       -> squeeze -> Y2 -> op2 -> ...
```
If op1 is an inplace `relu`, inplace of `reshape` should not happen! Therefore, when we record the identity inplace ops, we should also know whether the leaf vars of the branched inplace tree would be non-identity inplaced or not. In the implementation of this PR, we would record all the leaf var nodes that are non-identity inplaced and prune them when branched identity inplace happens.

## 3. Mark some leaf vars to be non-reusable to avoid further reuse error
There are two cases when the leaf vars should not be reused further (i.e, reused inside `BufferSharedCrossOpMemoryReusePass`):
- Branched identity inplace happens. If any leaf var is reused by another vars, calculation result may be wrong.
- If the last lived ops of `X` is not a subset of all identity inplace ops. In this case, the last lived ops of `X` may read the data of `X` after its data is changed by other memory reuse process. Calculation result may be wrong too.


paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
```diff
@@ -0,0 +1,354 @@
...
+      parent[out_var] = in_var;
+    }
+  }
+  return parent;
+}
+
+static void ShrinkIdentityInplacedOps(
+    IdentityInplaceVarInfo *info,
+    const std::unordered_set<VarHandle *> &reused_vars) {
+  if (reused_vars.empty()) return;
```
zhhsplendid: NIT: Not sure whether it is in public Google code style. When I was in Google, we must write {} around "if, while, for" even if it is just one statement.

For example, here you can write
if (reused_vars.empty()) {
     return;
}

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
```diff
@@ -0,0 +1,354 @@
...
+ *  `output_var` is the output of 2 different ops, `output_var`s in these 2 ops
+ *  must have different version, which is guaranteed by SSA graph. Therefore,
+ *  `output_var` must be the unique output of unique op. Condition 2 is proved.
+ */
+using IdentityInplaceVarInfo = std::unordered_map<
+    VarHandle * /*input_var*/,
+    std::vector<std::pair<VarHandle * /*output_var*/,
+                          ComputationOpHandle * /*identity op*/>>>;
+
+static std::unordered_map<VarHandle *, VarHandle *> BuildParent(
```
zhhsplendid: Would you mind adding  a simple comment above "BuildParent"

As I understand when I read the code, here is

"Building directional graph using adjacency map. The returned map mapping a Var to its Parent Var". Then it is easier to read.

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
```diff
@@ -0,0 +1,354 @@
...
+    const IdentityInplaceVarInfo &info) {
+  std::unordered_map<VarHandle *, VarHandle *> parent;
+
+  for (auto &pair : info) {
+    PADDLE_ENFORCE_EQ(pair.second.empty(), false,
+                      platform::errors::InvalidArgument(
+                          "Var info list cannot be empty, this may be a bug"));
+
+    auto *in_var = pair.first;
+    parent[in_var];
```
zhiqiu: What's it used for？

paddle/fluid/framework/ir/memory_optimize_pass/memory_reuse_pass.cc
```diff
@@ -41,6 +41,15 @@ void MemoryReusePass::ApplyImpl(Graph *graph) const {
     pinned_var_set_ = &graph->Get<details::PinnedVars>(details::kPinnedVars);
   }
 
+  skip_reuse_vars_ = &Get<std::vector<SkipReuseVars>>(kSkipReuseVars);
+  if (skip_reuse_vars_->empty()) {
+    skip_reuse_vars_->resize(all_vars_->size());
+  } else {
+    PADDLE_ENFORCE_EQ(
+        skip_reuse_vars_->size(), all_vars_->size(),
+        platform::errors::InvalidArgument("Size not match, this may be a bug"));
```
zhiqiu, line 54: Would you mind add some comments on why skip_reuse_vars_.size() should equal all_vars_->size() or empty?

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
```diff
@@ -0,0 +1,354 @@
...
+
+    if (failure) continue;
+
+    std::unordered_set<ComputationOpHandle *> candidate_ops;
+    for (auto *op_node : in_var->Node()->outputs) {
+      auto *op = dynamic_cast<ComputationOpHandle *>(
+          &(op_node->Wrapper<OpHandleBase>()));
+      if (op == nullptr) {
+        candidate_ops.clear();
+        break;
```
zhiqiu, line 320: A little question, why use break here?

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
zhhsplendid: It is put "in_var" as key in parent if it is not in parent

zhiqiu: I got it, while why we need this, i.e. insert a `key` without `value`?

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
sneaxiy: Thanks a lot. I have changed several lines. Please check it.

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
sneaxiy: Yes, you are right. This line is used to adding `nullptr` parent for root vars. It is useful because we can distinguish whether the `var` is inside the forest or not to avoid probable bug.

paddle/fluid/framework/ir/memory_optimize_pass/memory_reuse_pass.cc
sneaxiy: Thanks a lot! I have added some comments in codes. `all_vars_->size()` returns the place number inside `ParallelExecutor`, so `skip_reuse_vars_` must be not initialized (empty) or have the same place number as `all_vars_`.

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
sneaxiy: Thanks a lot. Comments are added.

paddle/fluid/framework/ir/memory_optimize_pass/buffer_shared_identity_inplace_op_pass.cc
sneaxiy: Now we require that all last live ops that read `in_var` must be `ComputationOpHandle` but not something like `AllReduceOpHandle`. This is something like what `ReferenceCountPass` does.

zhiqiu: Got it, thanks.

Issue #21251, opened 2019-11-20
使用scatter_nd, scatter_nd_add报错


Angus07:
我输入的tensor分别是index , update, ref ，
实际结构为：
index [3200, 1600, 2]
update[3200, 1600 ]
ref [3200, 18000]
打印结果如下
index name: "stack_2.tmp_0"
type {
  type: LOD_TENSOR
  lod_tensor {
    tensor {
      data_type: INT32
      dims: -1
      dims: 1600
      dims: 2
    }
    lod_level: 0
  }
}

update name: "reshape2_51.tmp_0"
type {
  type: LOD_TENSOR
  lod_tensor {
    tensor {
      data_type: FP32
      dims: -1
      dims: 1600
    }
    lod_level: 0
  }
}
persistable: false

ref name: "tmp_53"
type {
  type: LOD_TENSOR
  lod_tensor {
    tensor {
      data_type: FP32
      dims: -1
      dims: 18000
    }
    lod_level: 0
  }
}

 运行 output = fluid.layers.scatter_nd_add(ref, index,  update)然后输出一个output tensor，在使用reshape 或者 print 这个 tensor的时候出错了，提示这个tensor是bool，
reshape时错误如下：
![image](https://user-images.githubusercontent.com/4252318/69207346-63718d80-0b8a-11ea-8d4c-b6060b68f916.png)
Print时错误如下：
attn_dists2 Traceback (most recent call last):
  File "./run_seq2seq.py", line 358, in <module>
    main(args)
  File "./run_seq2seq.py", line 107, in main
    ernie_config=ernie_config)
  File "/mnt/du/chenliangyu/ernie-gen/ernie-gen/ernie-gen-master/finetune/seq2seq.py", line 213, in create_model
    fc_out = cal_logit(src_ids, enc_out, tgt_pos, args, ernie_config, weights, gens)
  File "/mnt/du/chenliangyu/ernie-gen/ernie-gen/ernie-gen-master/finetune/seq2seq.py", line 157, in cal_logit
    print("attn_dists2", attn_dists_projected)
  File "/root/du/chenliangyu/install/paddle_release_home/python/lib/python2.7/site-packages/paddle/fluid/framework.py", line 977, in __str__
    return self.to_string(True)
  File "/root/du/chenliangyu/install/paddle_release_home/python/lib/python2.7/site-packages/paddle/fluid/framework.py", line 1020, in to_string
    res_str = _debug_string_(proto, throw_on_error)
[...]

PR #21257, opened 2019-11-20
Fix the bug of scatter_nd/scatter_nd_add

ForFishes:
It can't obtain the exact `dtype`, so it should add `input_param_name` to resolve the bug.
the issue link: https://github.com/PaddlePaddle/Paddle/issues/21251


xyzhou-puck (dismissed):
LGTM.

gavin1332 (approved):
LGTM

PR #21278, opened 2019-11-21
Fix dgc accuracy by mv regularization to local

wangxicoding:
1. Regularization is part of loss, but its derivative equation is certain. So paddle add weight decay to grad before optimizer after allreduce. Because weight decay is part of grad, so we need to move it to the local before local grad accumulates. Otherwise, the model is very likely overfitting, harm test convergence. In resnet50 imagnet dataset, DGC loss 2% test accuracy before we apply this change because of overfitting.

2. In paddle, loss has divided by nranks. Because dgc_op is before allreduce, so local regular_coeff need div nranks too. But now we multi grad with nranks in dgc_op, in that case, regular_coeff don't need to /nranks, can prevent precision loss. For coeff often equal with 1e-4, if nranks=32, coeff/nranks will be 3.125e-6, the numerical accuracy of coeff/nranks will be too low. In resnet50 imagenet dataset, DGC loss 0.6% test accuracy before we multi grad with nranks because of precision loss.

After the above repair, DGC accuracy fully matches the baseline in resnet50 imagenet dataset.

### DIFF
--- a/paddle/fluid/operators/dgc_op.h
+++ b/paddle/fluid/operators/dgc_op.h
@@ -75,6 +75,10 @@ class DGCOpKernel : public framework::OpKernel<T> {
     // need to /nranks, can prevent precision loss. For coeff often equal
     // with 1e-4, if nranks=32, coeff/nranks will be 3.125e-6, the numerical
     // accuracy of coeff/nranks will be too low.
+    PADDLE_ENFORCE_EQ(regular_type >= 0 && regular_type <= 2, true,
+                      platform::errors::InvalidArgument(
+                          "DGC only support one of None|L1Decay|L2Decay "
+                          "Regularization for now."));
     if (regular_type == 0) {
       grad_out_e.device(eigen_ctx) = (1.0 * nranks) * g_e;
     } else if (regular_type == 1) {
@@ -84,9 +88,6 @@ class DGCOpKernel : public framework::OpKernel<T> {
     } else if (regular_type == 2) {
       // L2Decay. grad = grad + coeff * param
       grad_out_e.device(eigen_ctx) = (1.0 * nranks) * g_e + regular_coeff * p_e;
-    } else {
-      PADDLE_ENFORCE_EQ(regular_type >= 0 && regular_type <= 2, true,
-                        "Regularization must be one of None|L1Decay|L2Decay");
     }
 
     // current step
--- a/paddle/fluid/operators/dgc_op.h
+++ b/paddle/fluid/operators/dgc_op.h
@@ -43,6 +43,8 @@ class DGCOpKernel : public framework::OpKernel<T> {
     auto v = ctx.Input<framework::Tensor>("V");
     auto g = ctx.Input<framework::Tensor>("Grad");
 
+    auto grad_out = ctx.Output<framework::Tensor>("Grad_out");
+
     // attrs
     float m = ctx.Attr<float>("m");
     bool use_nesterov = ctx.Attr<bool>("use_nesterov");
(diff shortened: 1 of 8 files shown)


paddle/fluid/operators/dgc_op.h
```diff
@@ -91,19 +125,17 @@ class DGCOpKernel : public framework::OpKernel<T> {
...
-    auto g_e = framework::EigenVector<T>::Flatten(*g);
-    auto& dev_ctx = ctx.template device_context<DeviceContext>();
-    auto& eigen_ctx = *dev_ctx.eigen_device();
 
-    if (static_cast<int>(*current_step) ==
-        static_cast<int>(rampup_begin_step)) {
-      // calc local momentum from global momentum
-      u_out_e.device(eigen_ctx) = (1.0 / nranks) * u_e;
-    }
+    // calc local momentum from global momentum
```
gongweibao: Clean up this ?

paddle/fluid/operators/optimizers/dgc_momentum_op.cc
```diff
@@ -47,6 +52,10 @@ class DGCMomentumOpMaker : public MomentumOpMaker {
  public:
   void Make() override {
     AddInput("current_step", "(Tensor) Current step.");
+    AddInput("nranks", "(Tensor) nranks.");
```
gongweibao: Add more comment here.

paddle/fluid/operators/optimizers/dgc_momentum_op.h
```diff
@@ -38,6 +38,22 @@ class DGCMomentumKernel : public framework::OpKernel<T> {
     auto current_step_tensor = context.Input<framework::Tensor>("current_step");
     auto* current_step = current_step_tensor->data<T>();
 
+    // nranks
+    auto nranks_tensor = context.Input<framework::Tensor>("nranks");
+    const int nranks = static_cast<const int>(*nranks_tensor->data<float>());
+    PADDLE_ENFORCE_GT(nranks, 1, "DGC is not useful when num_trainers <= 1");
```
gongweibao: PADDLE_ENFORCE_GT(nranks, 1, "DGC is not useful when num_trainers <= 1, but input %d", nranks);

python/paddle/fluid/optimizer.py
```diff
@@ -1191,6 +1215,37 @@ def _dgc_op(self, param_var, clip_var, grad_var, u_var, v_var, k_var,
         dgc_op._set_attr(op_maker.kOpRoleVarAttrName(),
                          [param_var.name, grad_var.name])
 
+    def apply_gradients(self, params_grads):
```
gongweibao: Is it better to modify the base class function, and add the DGC contition there:
```
not_dgc_params_grads = []
        dgc_params_grads = []
        for param, grad in params_grads:
            if not self._is_use_dgc(param, grad):
                not_dgc_params_grads.append((param, grad))
            else:
                dgc_params_grads.append((param, grad))

        # DGC clip and regularization in local
        not_dgc_params_grads = append_gradient_clip_ops(not_dgc_params_grads)

        # Add regularization if any
        not_dgc_params_grads = append_regularization_ops(not_dgc_params_grads,
                                                         self.regularization)
```

gongweibao (changes requested):

gongweibao (dismissed):
LGTM

paddle/fluid/operators/optimizers/dgc_momentum_op.cc
wangxicoding: Done.

paddle/fluid/operators/optimizers/dgc_momentum_op.h
wangxicoding: Done.

Boyan-Liu (approved):
Done.

gongweibao (approved):
LGTM

PR #21279, opened 2019-11-21
cuda kernel refine for mod in elementwise_op

wangchaochaohu:
in cuda kernel, mod operation is cost time longer, so wo use minus and multiple to replace this operation


zhaoyuchen2018 (commented):
有没有测过能提升多少？

PR #21284, opened 2019-11-21
Fix the crash issue when scale or bias was null-pointer.

yihuaxu:
when run "ctest -R test_layers -V" command, the crash issue will be exist. now we use these patch to fix the crash issue.

reslove #21053


luotao1 (approved):
LGTM! Thanks very much! Please cherry-pick to release/1.6

PR #21285, opened 2019-11-21
Add global value getter setter

sneaxiy:
This PR provides a gentle and common way to read C++ global variables in Python.
From now on, we can use `fluid.core.globals()` as a dict to get the C++ variables by its name.

For example, when we read a C++ variable named `X`, we can use:
```Python
x_value = fluid.core.globals()["X"]
```

When we write a C++ variable named `X`, we can use:
```Python
fluid.core.globals()["X"] = some_other_value
```

Also support the following methods:
```Python
# find whether "value_name" exists
"value_name" in fluid.core.globals() 

# return all variables names
fluid.core.globals().keys() 

# return "X" value if "X" exists, while return default value if "X" does not exist
fluid.core.globals().get("X", default=False) 
```

### DIFF
--- a/paddle/fluid/pybind/CMakeLists.txt
+++ b/paddle/fluid/pybind/CMakeLists.txt
@@ -15,6 +15,7 @@ set(PYBIND_SRCS
   exception.cc
   protobuf.cc
   const_value.cc
+  global_value_getter_setter.cc
   reader_py.cc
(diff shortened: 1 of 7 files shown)


chenwhql (approved):
LGTM

chenwhql (approved):
LGTM

PR #21299, opened 2019-11-21
test=develop, package the CAPI inference library and third_party

silingtong123:
1. package the CAPI inference library and third_party
2. simplify the third-party in the CAPI inference library


liupluswei (approved):
LGTM

PR #21300, opened 2019-11-21
Support CVM OP Running in GPU

hutuxian:
Previously, CVM OP was only able to run in CPU. This PR implements its GPU kernel.
What's more, we improve the UTs about CVM OP.


gavin1332 (approved):
LGTM

PR #21307, opened 2019-11-22
adapt test_collective_base.py for only two GPU cards available.

gavin1332:
unit tests fail when only two GPU cards are available. Fix it!


guru4elephant (approved):
LGTM

PR #21317, opened 2019-11-22
Fix sync_bn grad maker

sneaxiy:
`BatchNormGradMaker` is defined twice in `batch_norm_op.cc` and `sync_batch_norm_op.cc` respectively, which would make gcc link error. This PR fixes it.


gongweibao (dismissed):
LGTM

gongweibao (approved):
LGTM

PR #21318, opened 2019-11-22
make config option DisableGlogInfo() able to mute all inference logs

cryoco:
Upgrade config option DisableGlogInfo() from muting GLOG info only, to muting ALL inference logs.
Note: If this option is set, paddle inference will produce no log outputs.


NHZlX (approved):
LGTM

PR #21322, opened 2019-11-22
fix the fill_constant op precious problem 

wangchaochaohu:
To solve https://github.com/PaddlePaddle/Paddle/issues/21113, we use the str to pass the value and  we need pay atttention to the use of math::set_constant, the  ``value`` parameter is float and it is fixed, so we discard the usage of math::set_constant in fill_constant OP .

### DIFF
--- a/python/paddle/fluid/layers/tensor.py
+++ b/python/paddle/fluid/layers/tensor.py
@@ -547,7 +547,10 @@ def fill_constant(shape, dtype, value, force_cpu=False, out=None):
                 'fill_constant')
     check_type(shape, 'shape', (Variable, list, tuple), 'fill_constant')
     inputs = {}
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/layers/tensor.py
```diff
@@ -552,6 +552,11 @@ def fill_constant(shape, dtype, value, force_cpu=False, out=None):
         'force_cpu': force_cpu or force_init_on_cpu()
     }
 
+    if convert_dtype(dtype) in ['int64', 'int32']:
+        attrs['str_value'] = str(int(value))
+    else:
+        attrs['str_value'] = str(float(value))
```
zhaoyuchen2018, line 558: bool 类型的也传递float吗？

wangchaochaohu: 对的 原来的逻辑也是这样

PR #21324, opened 2019-11-22
Polish reference count pass

sneaxiy:
Refine `ReferenceCountPass` by:
- not finding downstream `ComputationOpHandles` if some last live ops of a certain var is not `ComputationOpHandle`. This is because in allreduce mode, this case only happens to parameter, and the last lived ops of parameters must be optimizer ops; in reduce mode, this case only happens to parameter too, and the last lived ops of parameter must be optimizer ops (if the gpu card is responsible to update the parameter) or none (if the gpu card is not responsible to update the parameter, so that the last lived ops would be `Broadcast`).
- do not precede to find last live ops. Preceding is wrong, because last lived ops must be the generated op of var if pending ops are not the last lived ops.

### DIFF
--- a/paddle/fluid/framework/ir/memory_optimize_pass/reference_count_pass.cc
+++ b/paddle/fluid/framework/ir/memory_optimize_pass/reference_count_pass.cc
@@ -202,34 +202,7 @@ static bool ShrinkNoNeedBufferVarOpDependency(
   }
 }
 
-/**
- * Find the nearest downstream computation op handle. If the op is a
- * computation op, just return itself.
- */
-static details::ComputationOpHandle *FindNextComputationOpHandleOrReturnItself(
-    details::OpHandleBase *op, size_t scope_idx) {
-  std::queue<details::OpHandleBase *> q;
-  std::unordered_set<details::OpHandleBase *> visited;
-  q.push(op);
-  while (!q.empty()) {
-    auto *op = q.front();
-    q.pop();
-    auto *compute_op = dynamic_cast<details::ComputationOpHandle *>(op);
-    if (compute_op != nullptr && compute_op->GetScopeIdx() == scope_idx) {
-      return compute_op;
-    }
-    for (auto *out_var : op->Outputs()) {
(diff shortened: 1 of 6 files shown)


paddle/fluid/framework/ir/memory_optimize_pass/reference_count_pass.cc
```diff
@@ -344,47 +305,45 @@ void ReferenceCountPass::ApplyImpl(ir::Graph *graph) const {
...
+                << " in scope " << last_ver_var->scope_idx();
+        continue;
+      }
 
-        PADDLE_ENFORCE_EQ(status, LastLiveOpSearchStatus::kSuccess);
-        PADDLE_ENFORCE(!result.empty(), "Last living ops of %s cannot be empty",
-                       var_name);
+      PADDLE_ENFORCE_EQ(status, LastLiveOpSearchStatus::kSuccess,
+                        platform::errors::InvalidArgument(
+                            "status must be success or faiulre"));
```
zhiqiu: Maybe "status must be success here." is more appropriate, since `kFailure ` is already handled before.

paddle/fluid/framework/ir/memory_optimize_pass/reference_count_pass.cc
```diff
@@ -265,24 +223,27 @@ ExtractComputationOpFromLastLivedVar(details::VarHandle *var, size_t scope_idx,
   std::unordered_set<details::ComputationOpHandle *> computation_op;
```
zhiqiu: Maybe `computation_ops` is more appropriate, compared to `candidates`.

sneaxiy: Done.

paddle/fluid/framework/ir/memory_optimize_pass/reference_count_pass.cc
sneaxiy: Done.

zhiqiu (approved):
LGTM.

PR #21250, opened 2019-11-20
add input type and input data type check for Print_op test=develop

lijianshe02:
为Print python api的输入input加类型检查：
- 检查类型是否为Variable
- 检查数据类型是否为float32, float64, int32_t, int64_t, bool
- 在单测中覆盖类型错误和数据类型错误两个异常情况
- 可手动运行示例，看下错误：
```
import paddle.fluid as fluid
import numpy as np
x1 = fluid.create_lod_tensor(np.array([[-1]]), [[1]], fluid.CPUPlace())
fluid.layers.Print(x1)
```
> TypeError: The type of 'input' in fluid.layers.Print must be <class 'paddle.fluid.framework.Variable'>, but received <class 'paddle.fluid.core_avx.LoDTensor'>.

```
import paddle.fluid as fluid
x2 = fluid.layers.data(name='x2', shape=[4], dtype="float16")
fluid.layers.Print(x2)
```
> TypeError: The data type of 'input' in fluid.layers.Print must be ['float32', 'float64', 'int32_t', 'int64_t', 'bool'], but received float16.

PR #21326, opened 2019-11-22
fix Print_op input dtype list error test=develop

lijianshe02:
The data type should be int32 and int64, not  int32_t, int64_t 
related #21250


luotao1 (approved):
LGTM

PR #21340, opened 2019-11-25
Fix Cond Bug for Nested Control Flow

zhhsplendid:
This PR fixes a lot of bugs. Because it touched a lot of appending backward stuff, I also run and pass CE: http://ce.paddlepaddle.org:8080/viewLog.html?buildId=69616&tab=buildResultsDiv&buildTypeId=PaddleModesl_ModelDebug_Model11Build

Detail of bug:

1. Executor will drop kid scopes but in conditional_block_op, we shouldn't drop kid scopes during conditional_block forward op since it can be used during backward. (But we can drop after grad_op)
image

2. Fix assigning inside gradient to outside gradient in conditional_block_op.
image

3. Fix a case that inputs may set stop_gradient and we shouldn't PADDLE_ENFORCE each input has GRAD when inferring shape
image

4. For conditional_block_op, if Cond is also one of Inputs, we should treat it as input so that GRAD of the input can be passed out.
image

5. Delete the Output of grad of condition for conditional_block_op. Cond shouldn't have gradient.
image

6. During appending backward for control flow, the old code didn't find path for sub-block. It was only done for global block and it can cause sub-block backward error.
image

7. Use different @GRAD@RENAME in different block to avoid assigning GRAD values to parent GRAD. Otherwise child scope GRAD may be assigned to parent scope and causes gradient error. For example,
at parent scope, var@GRAD = var@GRAD@RENAME@0 + var@GRAD@RENAME@1, where rename 0 and rename 1 come from conditional_block_op0,1. If conditional_block_op1 is not run, then the correct var@GRAD@RENAME@1 should be 0 but bug exists if the var@GRAD@RENAME@1 is assigned in conditional_block_op0, then var@GRAD@RENAME@1 is not 0 and it is wrong.
image

8. sum_op.cu has bug when input tensors are not initialized. For example, if in_0.numel() == 0 and in_1 is initialized with values, all three cases won't run.
image

Finally we can run nested control flow.


paddle/fluid/framework/executor.cc
```diff
@@ -192,11 +191,12 @@ void Executor::RunFromDataset(std::shared_ptr<TrainerBase> trainer) {
 void Executor::Run(const ProgramDesc& pdesc, Scope* scope, int block_id,
                    bool create_local_scope, bool create_vars,
                    const std::vector<std::string>& skip_ref_cnt_vars,
-                   bool force_disable_gc) {
+                   bool force_disable_gc, bool keep_kid_scopes) {
```
sneaxiy, line 194: Seems that `keep_kids_scopes` is not used?

zhhsplendid: Used at line 199

sneaxiy (approved):
Cool!

paddle/fluid/framework/executor.cc
Xreki: 我印象中，这个函数本来就有这个参数的呢，是不是调用executor的地方都拆成了low_level函数的调用，所以这里这个参数删掉了？

paddle/fluid/framework/scope.cc
```diff
@@ -174,6 +179,13 @@ const Scope* Scope::FindScopeInternal(const Variable* var) const {
   return (parent_ == nullptr) ? nullptr : parent_->FindScope(var);
 }
 
+const Scope* Scope::FindScopeInternal(const std::string& name) const {
+  if (vars_.find(name) != vars_.end()) {
+    return this;
+  }
+  return (parent_ == nullptr) ? nullptr : parent_->FindScope(name);
+}
```
Xreki, line 187: 新增的这两个函数，跟已有的`const Scope* Scope::FindScopeInternal(const Variable* var) const`，除了输入参数类型不同外，二者还有其他的区别吗？能否只保留一个？
另外，这个新增的接口好像只在单测里面用到了。

paddle/fluid/framework/scope_test.cc
```diff
@@ -53,7 +53,9 @@ TEST(Scope, FindScope) {
   Variable* v = s.Var("a");
 
   EXPECT_EQ(&s, s.FindScope(v));
+  EXPECT_EQ(&s, s.FindScope("a"));
```
Xreki, line 56: 我怎么觉得这两个Find是一样的呢？有什么case，二者的返回值会不一样？

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -58,13 +59,15 @@ class ConditionalBlockOp : public ConditionalOp {
...
       framework::Executor exec(dev_place);
       auto *block = Attr<framework::BlockDesc *>("sub_block");
+      VLOG(3) << "Conditional block.idx = " << block->ID()
+              << ", scope = " << &cur_scope;
       auto &skip_vars =
           Attr<std::vector<std::string>>(ConditionalOp::kSkipEagerDeletionVars);
       exec.Run(*block->Program(), &cur_scope, block->ID(), false, true,
-               skip_vars);
+               skip_vars, /* force_disable_gc */ false,
+               /* keep_kid_scopes */ true);
```
Xreki, line 70: 这里设置`keep_kid_scopes = true`，是为了支持嵌套conditional_block的情况？如果不嵌套的话，不会有问题。while_op里面也是设置成true的。

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -92,60 +95,65 @@ class ConditionalBlockGradOp : public ConditionalOp {
...
-      cur_scope.Rename(new_in_grad_name, in_grad_name);
+      framework::Variable *outside_var =
+          parent_scope.FindVar(outside_grad_name);
+      if (outside_var == nullptr) {
+        continue;
+      }
+      platform::DeviceContext *dev_ctx =
+          platform::DeviceContextPool::Instance().Get(place);
+      framework::VisitVarType(*inside_var,
+                              AssignFunctor(outside_var, *dev_ctx));
```
Xreki, line 156: 赞，其实我也不喜欢在op里面插入别的op这种实现。

paddle/fluid/framework/scope_test.cc
zhhsplendid: 返回值没有不一样，只是输入不一样。就是我创建了一个用名字FindScope的函数，方便一点

paddle/fluid/operators/controlflow/conditional_block_op.cc
zhhsplendid: 对的，是为了支持嵌套

paddle/fluid/framework/scope.cc
zhhsplendid: 这个接口除了输入参数不同，没有别的不同。后面我需要用到。。。只是用起来不方便。。如果没有用名字查找的接口，要先auto* var = scope.FindVar(name), 然后再 scope.FindScope(var). 效率上也低一些。。因为用Variable 去 FindScope，Variable 是 map 的值，不是key

paddle/fluid/framework/executor.cc
zhhsplendid: 我不知道别的op背景，我这边conditional_block_op在现存代码还是调用这个接口，然后没法控制keep kid scopes，所以我增加该bool

paddle/fluid/framework/scope.cc
Xreki: 我觉得没有必要同时存在两个这样的接口。如果新增的接口只是用起来方便，建议后续PR将第一个接口删掉。

PR #21347, opened 2019-11-25
[cherry-pick] find lookup table in order & support dump param

Thunderbrook:
support dump param of model into afs
find lookup table in order


yaoxuefeng6 (approved):
LGTM

PR #21359, opened 2019-11-25
Split VarBase from Python Variable  for Dygraph

zhiqiu:
This PR try to split VarBase from python Variable Wrapper, thus Paddle can use VarBase which is exported from C++ directly in dygraph mode.  
The reason to do that includes,

- to decouple `Python side logic` of `dygraph eager execution` with `static graph`,
- to speed up training process by directly accessing C++ VarBase,
- to cooperate with `auto generated OP functions`, which may be introduced in the future.

After this PR, in dygraph mode, all the APIs of `Variable` before are indeed of `VarBase`. For example, 
- `to_variable()` returns a `VarBase` object which has the same interface with Variable,
-  `layers.xx` accepts `VarBase` objects as inputs and returns `VarBase` objects after some operations,
- `create_parameter` returns a `ParamBase` object which is a child object of `VarBase`.

Models evaluation results. (using `ptm` with default params in models repo.)
- Conclusion: 
    - in dygraph, 8% speed up and no effect on astringency; 
    - in static graph, no effcet on speed and astringency.

- Dygraph, before this PR, time:208.4s, ppl:265.4  
image

- Dygraph, After this PR, time:185.0, ppl:267.3
image

- Static Graph, before this PR, time:43.8s, ppl:268.1
image

- Static Graph, after this PR, time:44.1s, ppl:268.7
image


paddle/fluid/imperative/layer.cc
```diff
@@ -226,7 +226,11 @@ std::shared_ptr<VarBase> VarBase::NewVarBase(const platform::Place& dst_place,
 
   // TODO(Jiabin): change this after move unique_name generator to CXX
   auto new_var = std::make_shared<VarBase>(
-      false, "Itmp" + std::to_string(copied_counter_++));
+      true, Name() + std::to_string(copied_counter_++));
+
+  new_var->SetPersistable(Persistable());
```
phlrain: SetPersistable  会不会导致一些存储了 parameter的VarBase， 调用了这个接口，导致生成的VarBase不会被释放， Persistable 这个属性需要继承么

paddle/fluid/pybind/tensor_py.h
```diff
@@ -510,9 +511,26 @@ inline py::array TensorToPyArray(const framework::Tensor &tensor) {
   std::string py_dtype_str = details::TensorDTypeToPyDTypeStr(tensor.type());
 
   if (!is_gpu_tensor) {
-    return py::array(py::buffer_info(
-        const_cast<void *>(tensor_buf_ptr), sizeof_dtype, py_dtype_str,
-        static_cast<size_t>(tensor.dims().size()), py_dims, py_strides));
+    if (!need_deep_copy) {
```
sneaxiy, line 514: Why use `if (need_deep_copy)` to guard the following code? If the tensor is located on CPU, we do not need to copy. `numpy.array()` would deep copy data by default. This change would make the 2 copies when calling `numpy.array(tensor)` in Python.

python/paddle/fluid/framework.py
```diff
@@ -749,16 +954,7 @@ def detach(self):
...
-            new_var = self._cloneVar()
-            self.block.append_op(
-                type="assign",
-                inputs={'X': [self]},
-                outputs={'Out': [new_var]},
-                stop_gradient=True)
-            return new_var
-        else:
-            raise AttributeError("static graph model DO NOT supprt detach")
+        pass
```
sneaxiy, line 957: Raise error or just remove this method?

python/paddle/fluid/framework.py
```diff
@@ -790,12 +986,7 @@ def numpy(self):
                     print(x.numpy())
 
         """
-
-        if not self._ivar.value().get_tensor()._is_initialized():
-            raise ValueError("%s is Empty, Please check if it has no data in" %
-                             self.name)
-        new_ivar = self._ivar._copy_to(core.CPUPlace(), True)
-        return np.array(new_ivar.value().get_tensor())
+        pass
```
sneaxiy, line 989: Same as `detach`.

python/paddle/fluid/framework.py
```diff
@@ -826,25 +1017,7 @@ def set_value(self, value):
...
-        self_tensor_np = np.array(self_tensor)
-
-        assert self_tensor_np.shape == value_np.shape,  \
-                                      "Variable Shape not match, Variable [ {} ] need tensor with shape {} but load set tensor with shape {}".format( self._ivar.name, self_tensor_np.shape, value_np.shape)
-
-        assert self_tensor_np.dtype == value_np.dtype,  \
-                                      "Variable dtype not match, Variable [ {} ] need tensor with dtype {}  but load tensor with dtype {}".format( self._ivar.name, self_tensor_np.dtype, value_np.dtype)
-
-        self_tensor.set(value_np, _current_expected_place())
+        pass
```
sneaxiy, line 1020: Same as `detach`.

python/paddle/fluid/framework.py
```diff
@@ -882,16 +1055,7 @@ def backward(self, backward_strategy=None):
...
-            from .dygraph import BackwardStrategy
-            if backward_strategy is None:
-                backward_strategy = BackwardStrategy()
-                backward_strategy.sort_sum_gradient = False
-
-            self._ivar._run_backward(backward_strategy, _dygraph_tracer())
-        else:
-            raise ValueError(
-                "Variable.backward() is only avaliable in DyGraph mode")
+        pass
```
sneaxiy, line 1058: Same as `detach`.

sneaxiy (commented):
Please consider the whether to unify some APIs that static graph is not supported.

sneaxiy (dismissed):
LGTM. Please consider API design further asap.

paddle/fluid/imperative/layer.cc
zhiqiu: 动态图C++端VarBase是通过shared_ptr管理，persistable应该不会影响释放。

paddle/fluid/pybind/tensor_py.h
zhiqiu: Discussed offline. By default, `need_deep_copy` is False. Thanks a lot.

python/paddle/fluid/framework.py
zhiqiu: Thanks a lot. For generating English docs, this APIs are kept currently, which should be refined together with upgrading APIs.

sneaxiy (approved):
LGTM. Think about API unification further.

phlrain (approved):

lanxianghit:
请在pr描述中增加一些测试结论，比如性能指标是否按照预期取得收益，已有模型是否运行正常等

zhiqiu:
> 请在pr描述中增加一些测试结论，比如性能指标是否按照预期取得收益，已有模型是否运行正常等

好的

raindrops2sea (approved):

lanxianghit (approved):

PR #21364, opened 2019-11-26
Remove some temporary variables in fused_embedding_seq_pool_op

luotao1:
For performance requirements, now we remove some temporary tensors for fused_emb_seq_pool operator.
related with #21159


luotao1:
Close due to no speedup on PaddleCloud.

Issue #18944, opened 2019-07-31
MKLDNNDeviceContext Global Variables and Functions Code Clean 


luotao1:
There are several global variables and functions：
https://github.com/PaddlePaddle/Paddle/blob/fe32879d2abf9efeb2839eba27e8a419852635fc/paddle/fluid/platform/device_context.h#L380-L399
Could they move into `MKLDNNDeviceContext` class? @jczaja @LeoZhao-Intel

jczaja:
I have proposed some encapsulation of those global objects , but mostly they are  thread local storage objects , so It does not seem proper to put them into MKLDNNDeviceContext. Modification is part of 
#21466. Feel free to comment

luotao1:
@jczaja Could you continue the TLS refactoring?

jczaja:
@luotao1 I will work on this task next week

jczaja:
@pawelpiotrowicz Please take a look at that one.

pawelpiotrowicz:
https://github.com/PaddlePaddle/Paddle/pull/23066

jczaja:
merged

PR #21375, opened 2019-11-26
[MKL-DNN] LRN and Pool2d (FWD) NHWC support

jczaja:
This PR is adding support to NHWC Paddle data_format to forward LRN and pool mkl-dnn ops  as requested in #20964. It is subset of #21192 made to comply with max 20 files modifications per PR.

### DIFF
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -127,13 +127,17 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
       "TransDataLayoutFromMKLDNN only supports transform from MKLDNN to "
       "non-MKLDNN");
 
-  innerTransDataLayoutFromMKLDNN(in_layout, out_layout, in, out, place);
+#ifdef PADDLE_WITH_MKLDNN
+  innerTransDataLayoutFromMKLDNN(in_layout,
+                                 paddle::platform::get_cur_paddle_data_layout(),
+                                 in, out, place);
+#endif
 }
 
+#ifdef PADDLE_WITH_MKLDNN
 void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
                                     const Tensor& in, Tensor* out,
                                     platform::Place place) {
-#ifdef PADDLE_WITH_MKLDNN
   PADDLE_ENFORCE_NE(in.format(), MKLDNNMemoryFormat::format_undef,
                     platform::errors::InvalidArgument(
                         "Input tensor format is invalid. Input tensor should "
@@ -185,11 +189,17 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
   } else {
     out->ShareDataWith(in);
   }
+  // For exepected NHWC data format we need to reshape the Output tensor
+  // As MKL-DNN description was in NCHW and paddle is expecting NHWC
+  if (out_layout == DataLayout::kNHWC) {
+    std::rotate(out_tz.begin() + 1, out_tz.begin() + 2, out_tz.end());
+    out->Resize(framework::make_ddim(out_tz));
+  }
   out->set_layout(out_layout);
   // reset format since the out tensor will be feed to non-MKLDNN OPkernel
   out->set_format(MKLDNNMemoryFormat::format_undef);
-#endif
 }
+#endif
 
 }  // namespace framework
 }  // namespace paddle
--- a/paddle/fluid/framework/data_layout_transform.h
+++ b/paddle/fluid/framework/data_layout_transform.h
@@ -36,6 +36,7 @@ inline MKLDNNMemoryFormat ToMKLDNNFormat(const DataLayout& layout) {
     case DataLayout::kNHWC:
       return MKLDNNMemoryFormat::nhwc;
     case DataLayout::kNCHW:
+    case DataLayout::kAnyLayout:
       return MKLDNNMemoryFormat::nchw;
     default:
       PADDLE_THROW("Fail to convert layout %s to MKLDNN format",
@@ -66,15 +67,14 @@ inline MKLDNNDataType ToMKLDNNDataType(proto::VarType::Type type) {
   return MKLDNNDataType::data_undef;
 }
 
+void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
+                                    const Tensor& in, Tensor* out,
+                                    platform::Place place);
 #endif
 
 void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
                                const OpKernelType& expected_kernel_type,
                                const Tensor& in, Tensor* out);
-
-void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
-                                    const Tensor& in, Tensor* out,
-                                    platform::Place place);
 
 std::vector<int> GetAxis(const DataLayout& from, const DataLayout& to);
 
--- a/paddle/fluid/framework/data_layout_transform.h
+++ b/paddle/fluid/framework/data_layout_transform.h
@@ -36,7 +36,6 @@ inline MKLDNNMemoryFormat ToMKLDNNFormat(const DataLayout& layout) {
     case DataLayout::kNHWC:
       return MKLDNNMemoryFormat::nhwc;
     case DataLayout::kNCHW:
-    case DataLayout::kAnyLayout:
       return MKLDNNMemoryFormat::nchw;
     default:
       PADDLE_THROW("Fail to convert layout %s to MKLDNN format",
--- a/paddle/fluid/framework/data_transform.cc
+++ b/paddle/fluid/framework/data_transform.cc
@@ -19,6 +19,7 @@ limitations under the License. */
 #include "paddle/fluid/framework/data_type_transform.h"
 
 #ifdef PADDLE_WITH_MKLDNN
+#include <algorithm>
 #include "paddle/fluid/platform/mkldnn_helper.h"
 #endif
 
@@ -54,8 +55,16 @@ void TransformData(const OpKernelType &expected_kernel_type,
 
         auto out_format = platform::MKLDNNFormatForSize(in.dims().size(),
                                                         ToMKLDNNFormat(lin));
-
         out.ShareDataWith(input_tensor);
+        // For NHWC data we need reshape of tensors as MKL-DNN
+        // is expecting NHWC dims description order
+        if (lin == DataLayout::kNHWC) {
+          auto nchw_dims = paddle::framework::vectorize<int>(out.dims());
+          std::rotate(nchw_dims.begin() + 1, nchw_dims.end() - 1,
+                      nchw_dims.end());
+          out.Resize(framework::make_ddim(nchw_dims));
+          paddle::platform::set_cur_paddle_data_layout(lin);
+        }
         out.set_layout(DataLayout::kMKLDNN);
         out.set_format(out_format);
 #endif
--- a/paddle/fluid/operators/controlflow/fetch_op.cc
+++ b/paddle/fluid/operators/controlflow/fetch_op.cc
@@ -56,16 +56,20 @@ class FetchOp : public framework::OperatorBase {
     // FIXME(yuyang18): Should we assume the fetch operator always generate
     // CPU outputs?
     if (src_item.IsInitialized() && src_item.numel() > 0) {
+#ifdef PADDLE_WITH_MKLDNN
       // Conversion from MKL-DNN to Paddle
       if (src_item.layout() == framework::DataLayout::kMKLDNN) {
         framework::Tensor out;
         framework::innerTransDataLayoutFromMKLDNN(
-            src_item.layout(), framework::DataLayout::kNCHW, src_item, &out,
-            platform::CPUPlace());
+            src_item.layout(), paddle::platform::get_cur_paddle_data_layout(),
+            src_item, &out, platform::CPUPlace());
         TensorCopySync(out, platform::CPUPlace(), &dst_item);
       } else {
         TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
       }
+#else
+      TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
+#endif
     } else {
       // Not copy, if the src tensor is empty.
       dst_item.clear();
--- a/paddle/fluid/operators/conv_op.cc
+++ b/paddle/fluid/operators/conv_op.cc
@@ -155,6 +155,15 @@ framework::OpKernelType ConvOp::GetExpectedKernelType(
 #ifdef PADDLE_WITH_MKLDNN
   if (library == framework::LibraryType::kPlain &&
       platform::CanMKLDNNBeUsed(ctx)) {
+    // TODO(jczaja): Add support for NHWC
+    const std::string data_format = ctx.Attr<std::string>("data_format");
+    PADDLE_ENFORCE_NE(data_format, "NHWC",
+                      platform::errors::Unimplemented(
+                          "Conv MKLDNN does not support NHWC data format yet"));
+    PADDLE_ENFORCE_NE(
+        data_format, "NDHWC",
+        platform::errors::Unimplemented(
+            "Conv MKLDNN does not support NDHWC data format yet"));
     library = framework::LibraryType::kMKLDNN;
     layout = framework::DataLayout::kMKLDNN;
     customized_type_value =
@@ -190,27 +199,6 @@ framework::OpKernelType ConvOp::GetExpectedKernelType(
   }
 #endif
   return type;
-}
-
-framework::OpKernelType ConvOp::GetKernelTypeForVar(
-    const std::string& var_name, const Tensor& tensor,
-    const framework::OpKernelType& expected_kernel_type) const {
-#ifdef PADDLE_WITH_MKLDNN
-  // Only input require reshaping, weights and
-  // bias are having shape in NCHW order
-  if ((var_name == "Input") &&
-      (expected_kernel_type.data_layout_ == framework::DataLayout::kMKLDNN) &&
-      (tensor.layout() != framework::DataLayout::kMKLDNN)) {
-    auto attrs = Attrs();
-    auto ar = paddle::framework::AttrReader(attrs);
-    const std::string data_format = ar.Get<std::string>("data_format");
-    return framework::OpKernelType(expected_kernel_type.data_type_,
-                                   tensor.place(),
-                                   framework::StringToDataLayout(data_format));
-  }
-#endif
-  return framework::OpKernelType(expected_kernel_type.data_type_,
-                                 tensor.place(), tensor.layout());
 }
 
 void Conv2DOpMaker::Make() {
--- a/paddle/fluid/operators/conv_op.cc
+++ b/paddle/fluid/operators/conv_op.cc
@@ -48,11 +48,7 @@ void ConvOp::InferShape(framework::InferShapeContext* ctx) const {
   int groups = ctx->Attrs().Get<int>("groups");
   std::vector<int> dilations = ctx->Attrs().Get<std::vector<int>>("dilations");
   const std::string data_format = ctx->Attrs().Get<std::string>("data_format");
-
-  // MKL-DNN Kernels are using NCHW order of dims description
-  // so we ignore data_format consideration for MKL-DNN kernel
-  const bool channel_last = (this->IsMKLDNNType() == false) &&
-                            (data_format == "NHWC" || data_format == "NDHWC");
+  const bool channel_last = (data_format == "NHWC" || data_format == "NDHWC");
 
   PADDLE_ENFORCE_EQ(
       in_dims.size() == 4 || in_dims.size() == 5, true,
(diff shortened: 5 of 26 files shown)


luotao1:
@LeoZhao-Intel @jianhang-liu Could you have an internal review at first?

zhangting2020 (dismissed):
LGTM

paddle/fluid/framework/data_layout_transform.cc
```diff
@@ -127,13 +127,17 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
       "TransDataLayoutFromMKLDNN only supports transform from MKLDNN to "
       "non-MKLDNN");
 
-  innerTransDataLayoutFromMKLDNN(in_layout, out_layout, in, out, place);
+#ifdef PADDLE_WITH_MKLDNN
+  innerTransDataLayoutFromMKLDNN(in_layout,
+                                 paddle::platform::get_cur_paddle_data_layout(),
+                                 in, out, place);
+#endif
 }
```
luotao1, line 135: Should the macro `#ifdef PADDLE_WITH_MKLDNN` be refined?
line 95-202 should be in one macro `#ifdef PADDLE_WITH_MKLDNN`, but there are three.

paddle/fluid/framework/data_layout_transform.h
```diff
@@ -36,6 +36,7 @@ inline MKLDNNMemoryFormat ToMKLDNNFormat(const DataLayout& layout) {
     case DataLayout::kNHWC:
       return MKLDNNMemoryFormat::nhwc;
     case DataLayout::kNCHW:
+    case DataLayout::kAnyLayout:
```
luotao1: Why line39 here?

paddle/fluid/platform/device_context.cc
```diff
@@ -408,6 +412,14 @@ void set_cur_input_shape_cache_capacity(int input_shape_cache_capacity) {
...
 }
 
+void set_cur_paddle_data_layout(framework::DataLayout dl) {
+  cur_paddle_data_layout = dl;
+}
+
+framework::DataLayout get_cur_paddle_data_layout(void) {
+  return cur_paddle_data_layout;
+}
+
```
luotao1, line 422: There are too many global functions, should line391-421 in Class MKLDNNDeviceContext? See the duplicated issue #18944

paddle/fluid/operators/conv_op.cc
```diff
@@ -48,7 +48,11 @@ void ConvOp::InferShape(framework::InferShapeContext* ctx) const {
   int groups = ctx->Attrs().Get<int>("groups");
   std::vector<int> dilations = ctx->Attrs().Get<std::vector<int>>("dilations");
   const std::string data_format = ctx->Attrs().Get<std::string>("data_format");
-  const bool channel_last = (data_format == "NHWC" || data_format == "NDHWC");
+
+  // MKL-DNN Kernels are using NCHW order of dims description
+  // so we ignore data_format consideration for MKL-DNN kernel
+  const bool channel_last = (this->IsMKLDNNType() == false) &&
+                            (data_format == "NHWC" || data_format == "NDHWC");
```
luotao1: Since this PR only supports LRN and Pool2d (FWD) NHWC, please don't change other Ops.

paddle/fluid/operators/controlflow/fetch_op.cc
```diff
@@ -56,16 +56,20 @@ class FetchOp : public framework::OperatorBase {
...
-            platform::CPUPlace());
+            src_item.layout(), paddle::platform::get_cur_paddle_data_layout(),
+            src_item, &out, platform::CPUPlace());
         TensorCopySync(out, platform::CPUPlace(), &dst_item);
       } else {
         TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
       }
+#else
+      TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
+#endif
```
luotao1, line 72: line72 is duplicated with line68

paddle/fluid/framework/data_layout_transform.cc
LeoZhao-BJ: agree. Even in data_transform.cc::TransformData(), we can use PADDLE_WITH_MKLDNN cover all branches related with MKLDNN, we may use another PR to cover this part.

paddle/fluid/framework/data_layout_transform.h
```diff
@@ -66,16 +67,15 @@ inline MKLDNNDataType ToMKLDNNDataType(proto::VarType::Type type) {
...
 }
 
+void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
+                                    const Tensor& in, Tensor* out,
+                                    platform::Place place);
 #endif
 
 void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
                                const OpKernelType& expected_kernel_type,
                                const Tensor& in, Tensor* out);
```
LeoZhao-BJ, line 77: why not use PADDLE_WITH_MKLDNN cover TransDataLayoutFromMKLDNN?

paddle/fluid/framework/data_transform.cc
```diff
@@ -54,8 +55,16 @@ void TransformData(const OpKernelType &expected_kernel_type,
...
         out.ShareDataWith(input_tensor);
+        // For NHWC data we need reshape of tensors as MKL-DNN
+        // is expecting NHWC dims description order
+        if (lin == DataLayout::kNHWC) {
+          auto nchw_dims = paddle::framework::vectorize<int>(out.dims());
+          std::rotate(nchw_dims.begin() + 1, nchw_dims.end() - 1,
+                      nchw_dims.end());
+          out.Resize(framework::make_ddim(nchw_dims));
+          paddle::platform::set_cur_paddle_data_layout(lin);
+        }
```
LeoZhao-BJ, line 67: you mean even for paddle NHWC input, dims description is still in nchw order? looks weird.

paddle/fluid/platform/device_context.cc
LeoZhao-BJ: we can move all these global variables and functions into Class MKLDNNDeviceContext and set as static firstly, user can still use as global func.

paddle/fluid/platform/device_context.cc
```diff
@@ -397,6 +397,10 @@ thread_local std::string cur_input_shape_str = "";
 // the cache capacity of different input shapes for MKLDNN.
 // Default 1 means fixed input shape, not dynamic shape.
 thread_local int cur_input_shape_cache_capacity = 1;
+// Recently registered data_format. This is needed to
+// know for converting MKL-DNN Tensor to non MKL-DNN
+thread_local paddle::framework::DataLayout cur_paddle_data_layout =
+    paddle::framework::DataLayout::kNCHW;
```
LeoZhao-BJ, line 403: @luotao1 Is it possible to get layout format from PaddleTensor class? if possible, I think we can combine layout info with cur_input_shape_str.

paddle/fluid/framework/data_layout_transform.cc
```diff
@@ -185,11 +189,17 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
   } else {
     out->ShareDataWith(in);
   }
+  // For exepected NHWC data format we need to reshape the Output tensor
+  // As MKL-DNN description was in NCHW and paddle is expecting NHWC
+  if (out_layout == DataLayout::kNHWC) {
+    std::rotate(out_tz.begin() + 1, out_tz.begin() + 2, out_tz.end());
+    out->Resize(framework::make_ddim(out_tz));
+  }
```
LeoZhao-BJ, line 197: can we put this dim description transfer together? I mean I saw similar change in data_transform.cc? can we put together?

paddle/fluid/platform/device_context.cc
luotao1: @LeoZhao-Intel Do you mean here?
https://github.com/PaddlePaddle/Paddle/blob/d5ff79e55ea329396f4c61686a318b7c77520b93/paddle/fluid/framework/tensor.h#L37-L42
Do you have any idea on getting layout format from here?

paddle/fluid/framework/data_layout_transform.cc
jczaja: Here we rotate dims *left* and in data_transform we rotate dims *right* so code is a bit different. As they are not exactly the same If we make function doing such a rotation it is likely to have two separate implementations rather than one unified.

paddle/fluid/framework/data_transform.cc
grygielski: I think comment here is a bit misleading. It should be that MKL-DNN is expecting dimensions in NCHW order. @LeoZhao-Intel this is how MKL-DNN interprets dimensions. It's reasonable since dims are describing data from logical axis view and data layout is just placement of bits in a memory.

paddle/fluid/platform/device_context.cc
LeoZhao-BJ: https://github.com/PaddlePaddle/Paddle/blob/d5ff79e55ea329396f4c61686a318b7c77520b93/paddle/fluid/inference/api/analysis_predictor.cc#L206

What I mean is if we can combine layout with input shape, and for other places, we just read layout by reusing same global variable. The problem is how we can get layout from PaddleTensor.

But I am not sure if this change is for both inference and training, if for both, then my idea is useless.

luotao1: > But I am not sure if this change is for both inference and training

This PR is for training as well.

paddle/fluid/framework/data_layout_transform.cc
LeoZhao-BJ: yes, I know it is different, I mean can we move these kinds of code together, in one file or better in one function. e.g in data_transform.cc::TransformData

paddle/fluid/operators/controlflow/fetch_op.cc
grygielski: I think it's ok. TensorCopySync is still needed in this place even without PADDLE_WITH_MKLDNN defined.

paddle/fluid/operators/lrn_op.cc
```diff
@@ -207,6 +201,24 @@ class LRNOp : public framework::OperatorWithKernel {
...
+      const framework::OpKernelType& expected_kernel_type) const override {
+#ifdef PADDLE_WITH_MKLDNN
+    if ((expected_kernel_type.data_layout_ == framework::DataLayout::kMKLDNN) &&
+        (tensor.layout() != framework::DataLayout::kMKLDNN)) {
+      auto attrs = Attrs();
+      auto ar = paddle::framework::AttrReader(attrs);
+      const std::string data_format = ar.Get<std::string>("data_format");
+      return framework::OpKernelType(
+          expected_kernel_type.data_type_, tensor.place(),
+          framework::StringToDataLayout(data_format));
```
grygielski: Why you need to get tensor layout from "data_format"? Isn't it set in any place based on this attribute?

paddle/fluid/platform/device_context.cc
LeoZhao-BJ: Then not applicable. skip my comments here.

paddle/fluid/operators/pool_op.cc
```diff
@@ -162,6 +159,24 @@ framework::OpKernelType PoolOp::GetExpectedKernelType(
       layout_, library_);
 }
 
+framework::OpKernelType PoolOp::GetKernelTypeForVar(
+    const std::string& var_name, const Tensor& tensor,
+    const framework::OpKernelType& expected_kernel_type) const {
+#ifdef PADDLE_WITH_MKLDNN
+  if ((expected_kernel_type.data_layout_ == framework::DataLayout::kMKLDNN) &&
+      (tensor.layout() != framework::DataLayout::kMKLDNN)) {
+    auto attrs = Attrs();
```
grygielski, line 168: This code is exactly the same as in other NHWC operators. Can't it be moved to OperatorWithKernel class?

grygielski (dismissed):
Please refer to my questions. LGTM

paddle/fluid/operators/controlflow/fetch_op.cc
luotao1: @grygielski How about remove line66-68?

jczaja: That is correct , but when I tried to have this line only once ,then there were two #ifdef PADDLE_WITH_MKLDNN blocks and it was not very readable.

paddle/fluid/operators/conv_op.cc
jczaja: missed that one. OK

paddle/fluid/operators/controlflow/fetch_op.cc
grygielski: Then it will result in completely different code. If you build with PADDLE_WITH_MKLDNN you get:
```
if (src_item.layout() == framework::DataLayout::kMKLDNN) {
    ...
} else {
    TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
}
```
When you delete these lines you will get:
```
if (src_item.layout() == framework::DataLayout::kMKLDNN) {
    ...
}
```
If you want to delete also this `#else` it will result in:
```
if (src_item.layout() == framework::DataLayout::kMKLDNN) {
    ...
}
TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
```

paddle/fluid/platform/device_context.cc
jczaja: Ok, I have to think about it. I would like to do this little refactoring in next PR (NHWC for conv and conv_transpose).

paddle/fluid/operators/controlflow/fetch_op.cc
luotao1: Got it

paddle/fluid/framework/data_layout_transform.cc
jczaja: I agree with @LeoZhao-Intel , I can try to reorganize mentioned code to be in PADDLE_WITH_MKLDNN , but in next part of NHWC support PR (Conv & conv transpose NHWC support)

paddle/fluid/framework/data_layout_transform.h
jczaja: It should be, I will look to that in next PR (NHWC support to conv & conv_transpose)

paddle/fluid/framework/data_transform.cc
jczaja: MKL-DNN description is always NHWC , regardless actual data arrangement . So even if MKL-DNN is NCHW16C , NHWC  it is always NCHW description. Otherwise I would have to this kind of reshaping inside every MKL-DNN op

paddle/fluid/platform/device_context.cc
jczaja: @LeoZhao-Intel , @luotao1 Just one note here , in current design when *data_format=NHWC* then still Tensors layout is *kNCHW*. So:
1) I'm not sure what is current reading of Tensor::layout .
2) What does DataLayout::AnyLayout mean ?

paddle/fluid/operators/lrn_op.cc
jczaja: apprently it is not. When "data_format=NHWC" still Tensors that contain data stored according to that order are having layout set to : kNCHW. The thing is that I do not know current reading of Tensor::layout. In MKL-DNN integration we use kNCHW and kMKLDNN . As for kAnyLayout I'm not sure of it purpose now.

paddle/fluid/operators/pool_op.cc
jczaja: It will be a bit different in conv & conv_transpose :) I will consider making a common code once I have all 4 ops supporting NHWC (next PR).

paddle/fluid/framework/data_layout_transform.h
jczaja: That was a temporary solution introduced to  be consistent with behavior defined here:
https://github.com/PaddlePaddle/Paddle/blob/d1a6e112e60e19fcfdc0ff4ea69a5fdb5c63a1d8/paddle/fluid/framework/data_layout_transform.cc#L147-L148

which is when layout is unspecified e.g. kAnyLayout then assume it is kNCHW.

But I have just replaced this with more proper fix

paddle/fluid/platform/device_context.cc
LeoZhao-BJ: Honestly speaking, I am still a little confused here, we have 2 kinds of format, one is Paddle Tensor layout, the other is MKLDNN layout, how to map them together?

for input Paddle Tensor is kNHWC, what is real value reading from tensor?
what's MKLDNN input and output layout for different input Paddle Tensor layout?

can we  have a table to show clearly?

LeoZhao-BJ:
Honestly speaking, I am still a little confused, we have 2 kinds of format, one is Paddle Tensor layout series, the other is MKLDNN layout series, how to map them together?

My understanding, MKLDNN layer doesn't have direct map with Paddle tensor layout, it obviously need layout format/dims transition explicitly as did in data_layout_transform.cc and data_transform.cc
for example:
```
if (out_layout == DataLayout::kNHWC) {
    std::rotate(out_tz.begin() + 1, out_tz.begin() + 2, out_tz.end());
    out->Resize(framework::make_ddim(out_tz));
  }
```

My questions are:

1. If input Paddle Tensor is kNHWC, what is original dims format in Tensor before MKLDNN op modification? 
2. after MKLDNN op, what's the expected layout/dims format for output?

We  may need 2 tables to show the mapping relationship clearly.
one is original definition for dims format:

Tensor   DataLayout | Dims format
-- | --
kNHWC | nhwc?
kNCHW | nchw?
kAnyLayout | ?
kMKLDNN | ?

the other is for MKLDNN op with different input tensor data layout, what is expected dims format for output tensor:

Input Tensor   DataLayout | Dims format of output Tensor with   DataLayout kMKLDNN
-- | --
kNHWC | ?
kNCHW | ?
kAnyLayout | ?
kMKLDNN | ?

@jczaja  could you fill the tables as you knew or already did as that? we can discuss the gap and unclear place.

jczaja:
@LeoZhao-Intel Thanks for interesting questions.

Here is an example of model(relevant to this PR and next one) with **data_format=NHWC** that is having two Paddle CPU ops (Feed and fetch) and two MKL-DNN ops (conv and pool), hopefully this will shed some light on implementation details:
nhwc


Now to answer your questions:
1) Tensor you are asking about is represented by second object on the upper-left side ( *Tensor input signal*). It does have *layout=kNCHW* and dims format in NHWC order. This is a situation as existed before this PR and I adapted my changes to work with that. Lower branch is covered in next PR to be e.g. Filter data is treated differently than Input.
2) Right after MKL-DNN op execution Tensor's layout is MKL-DNN and dims format is NCHW order.
    When execution it enter into Paddle non MKL-DNN  domain (fetch_op ) then there is reorder to NHWC data arrangement , dims format is NHWC.
   
Tensor   DataLayout | Dims format
-- | --
kNHWC | no idea
kNCHW | nchw/nhwc (depends on Op *data_format*)
kAnyLayout | nchw
kMKLDNN | nchw

As for second table, It assumes that Tensor's layout is decide if kNCHW or kNHWC is used in model, but this does not seem to be true assumption. Information is stored in op's *data_format*. And from MKL-DNN integration point of view it is always kNCHW or kMKLDNN.  There is never kNHWC scenario. Op's *data_format* decide if Tensor (non MKL-DNN)is holding data arrange in NCHW or NHWC not Tensor::layout.

LeoZhao-BJ:
Thanks Jacek for detailed description, seems now NHWC format detection is not by layout from Tensor, it is decided by Tensor input signal or filter data, @luotao1 does it follow Paddle design or any bug in feed op?

luotao1:
> it is decided by Tensor input signal or filter data

It is decided by `data_format`, see conv2d: https://www.paddlepaddle.org.cn/documentation/docs/zh/api_cn/layers_cn/conv2d_cn.html#conv2d
image

> or any bug in feed op?

The `data_format` parameters don't relate with feed op.

jczaja:
@LeoZhao-Intel , Some more details on design are covered in original issue: #20964 . For example
- On distinguish of input signal and params :
https://github.com/PaddlePaddle/Paddle/issues/20964#issuecomment-554390990
- On how to carry information on NCWH / NHWC (Not sure If I got it all proper  from automatic translation:)):
https://github.com/PaddlePaddle/Paddle/issues/20964#issuecomment-549296501

grygielski (approved):
LGTM

jczaja:
@luotao1 To have CI passed your approval is needed (some code in operator.h was added). I have fixed some simpler issues, but decided to leave #18944 and #ifdef PADDLE_WITH_MKLDNN blocks modification for next PR related to NHWC. If you have any more questions or comments then I will be happy to answer or read about it.

luotao1 (approved):
LGTM

luotao1:
>  I have fixed some simpler issues, but decided to leave #18944 and #ifdef PADDLE_WITH_MKLDNN blocks modification for next PR related to NHWC. 

Hope the next PRs, and could you give a detail design doc in FluidDoc after finishing the NHWC supports?

LeoZhao-BJ:
> @LeoZhao-Intel , Some more details on design are covered in original issue: #20964 . For example
> 
> * On distinguish of input signal and params :
>   [#20964 (comment)](https://github.com/PaddlePaddle/Paddle/issues/20964#issuecomment-554390990)
> * On how to carry information on NCWH / NHWC (Not sure If I got it all proper  from automatic translation:)):
>   [#20964 (comment)](https://github.com/PaddlePaddle/Paddle/issues/20964#issuecomment-549296501)

Got the design point, in this case, the layout in op kernel is not from tensor layout, it is set by data_format. and tensor dim format is always nchw. @jczaja  I suggest we may have a design doc for details later, since some information scatter in different places or comments, it is not easy to understand the whole design purpose.

jczaja:
@LeoZhao-Intel When We have FWD and BWD done then I will update/add design document to FluidDoc repo.

LeoZhao-BJ:
LGTM

luotao1:

PR #21392, opened 2019-11-27
Remove ShareDataWith usage in some ops

sneaxiy:
`ShareDataWith` is not allowed to call inside operator to make output shares memory with input. This PR fixes this bug in some operators.


luotao1 (approved):
LGTM

PR #21402, opened 2019-11-27
set dim[0] to -1 if dim[0] < 0 during compiling for c_allgather op

sandyhouse:
set dim[0] to -1 if dim[0] < 0 during compiling for c_allgather op
move the assertion (input.shape[0] must be divisible by nranks) from compiler time to run time


python/paddle/fluid/layers/collective.py
```diff
@@ -134,10 +134,6 @@ def _c_reducescatter(x, nranks, ring_id=0, use_calc_stream=False):
     if not isinstance(x, Variable):
         raise TypeError('x must be a Variable')
 
-    if x.shape[0] % nranks != 0:
```
gavin1332, line 137: keep this check when shape[0] larger than 0

sandyhouse: done.

gavin1332 (dismissed):
LGTM

gavin1332 (approved):
LGTM

PR #21404, opened 2019-11-27
Add Lod information for gather_nd & scatter_nd

ForFishes:
Both `gather_nd` and `scatter_nd` don't save Lod information. When `LoDTensor` is input, they obtain the `Tensor` without Lod.
Issue: https://github.com/PaddlePaddle/Paddle/issues/21369


gavin1332 (approved):
LGTM

PR #21407, opened 2019-11-27
remove kDepXOut for abs_grad op

zhiqiu:
As the title, `abs_grad` only need `x` and `dout` as input, so use `kDepX` instead of `kDepXOut`.


sneaxiy (dismissed):
LGTM.

sneaxiy (approved):

PR #21410, opened 2019-11-27
Eliminate the impact on incremental compilation,test=develop

zhwesky2010:
The copy operation refreshes the time-stamp, so that make does not perform incremental compilation.

This PR fix that, and Incremental compilation will not be affected, and non-incremental compilation can also be accelerated through caching.
fix https://github.com/PaddlePaddle/Paddle/pull/21190#issuecomment-558668368


luotao1:
LGTM

luotao1 (approved):

PR #21422, opened 2019-11-28
add macro to ban windows

hutuxian:
Since nccl cannot be used in windows, so we add macros to remove nccl related code in windows.


gavin1332 (approved):
LGTM

PR #21428, opened 2019-11-28
Add reshape int8 op

wozna:
This PR enables quantization of reshape2 operator. Type int8 is not added in MLKDNN operator implementation, but in global implementation.  
Reshape and transpose ops are quantized only if previous operator or next operator is quantized. Thanks that we avoid situation where there is one int8 operators between two fp32 operators.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
@@ -393,7 +393,23 @@ ProgramDesc BuildProgramDescReshape() {
   SetOp(&prog, "reshape2", "Reshape1", {"b"}, {"c"}, true, true);
   SetOp(&prog, "transpose2", "Transpose2", {"c"}, {"d"}, true, true);
   SetOp(&prog, "reshape2", "Reshape2", {"d"}, {"e"}, true, true);
-  SetOp(&prog, "dropout", "Dropout", {"e"}, {"f"}, true, true);
+  SetOp(&prog, "dropout", "Dropout", {"e"}, {"f"}, true, false);
+
+  return prog;
+}
+
+// a->Transpose->b
+// b->Reshape->c
+// c->Dropout->d
+ProgramDesc BuildProgramDescReshapeBetweenNonQuantizedOp() {
+  ProgramDesc prog;
+  for (auto& v : variable_names_transpose) {
+    prog.MutableBlock(0)->Var(v);
+  }
+
+  SetOp(&prog, "transpose2", "Transpose2", {"a"}, {"b"}, true, false);
+  SetOp(&prog, "reshape2", "Reshape2", {"b"}, {"c"}, true, true);
+  SetOp(&prog, "dropout", "Dropout", {"c"}, {"d"}, true, false);
 
   return prog;
 }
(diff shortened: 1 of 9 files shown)


paddle/fluid/inference/api/mkldnn_quantizer_config.cc
```diff
@@ -42,6 +42,12 @@ MkldnnQuantizerConfig::MkldnnQuantizerConfig() {
   rules_["fc"]["W"] = ScaleAlgo::MAX_CH_T;
   rules_["fc"]["Bias"] = ScaleAlgo::NONE;
   rules_["fc"]["Out"] = ScaleAlgo::KL;
+
+  rules_["reshape2"]["X"] = ScaleAlgo::KL;
+  rules_["reshape2"]["Shape"] = ScaleAlgo::NONE;
+  rules_["reshape2"]["ShapeTensor"] = ScaleAlgo::NONE;
+  rules_["reshape2"]["XShape"] = ScaleAlgo::NONE;
+  rules_["reshape2"]["Out"] = ScaleAlgo::NONE;
```
wojtuss, line 55: A comment explaining why we do not need to calculate the scale for `reshape2`'s output could be helpful here. Could be added also to `transpose2`.

paddle/fluid/operators/reshape_op.cc
```diff
@@ -245,6 +245,13 @@ class ReshapeOpMaker : public framework::OpProtoAndCheckerMaker {
         "It has the lowest priority compare with Input(Shape) and "
         " Input(ShapeTensor).")
         .SetDefault({});
+    /* int8 parameters */
+    AddAttr<bool>("use_quantizer",
+                  "(bool, default false) "
+                  "Set to true for operators that should be quantized and use "
+                  "int8 kernel. "
+                  "Only used on CPU.")
```
wojtuss: I think "Used only on CPU." is more grammatically correct.

paddle/fluid/operators/reshape_op.cc
```diff
@@ -570,15 +580,20 @@ REGISTER_OPERATOR(reshape2_grad_grad, ops::Reshape2DoubleGradOp,
...
 REGISTER_OP_CPU_KERNEL_FUNCTOR(reshape2, float, ops::ReshapeKernel, double,
-                               ops::ReshapeKernel, int, ops::ReshapeKernel,
-                               int64_t, ops::ReshapeKernel);
+                               ops::ReshapeKernel, int8_t, ops::ReshapeKernel,
+                               uint8_t, ops::ReshapeKernel, int,
+                               ops::ReshapeKernel, int64_t, ops::ReshapeKernel);
 REGISTER_OP_CPU_KERNEL_FUNCTOR(reshape2_grad, float, ops::ReshapeGradKernel,
                                double, ops::ReshapeGradKernel, int,
+                               ops::ReshapeGradKernel, int8_t,
+                               ops::ReshapeGradKernel, uint8_t,
```
wojtuss: We do not provide gradient kernels for the `reshape(2)` ops.

paddle/fluid/operators/reshape_op.cc
```diff
@@ -553,9 +560,12 @@ REGISTER_OPERATOR(reshape_grad, ops::ReshapeGradOp,
 
 REGISTER_OP_CPU_KERNEL_FUNCTOR(reshape, float, ops::ReshapeKernel, double,
                                ops::ReshapeKernel, int, ops::ReshapeKernel,
-                               int64_t, ops::ReshapeKernel);
+                               int8_t, ops::ReshapeKernel, uint8_t,
```
wojtuss: I think we do not need the int8 kernels for the `reshape` op. We are interested in quantizing `reshape2` only now.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
```diff
@@ -370,6 +370,122 @@ TEST(CpuQuantizePass, transpose) {
...
+        dequantize_nodes_count++;
+      }
+    }
+  }
+  EXPECT_EQ(transpose_nodes_count, transpose_count);
+  EXPECT_EQ(reshape_nodes_count, reshape_count);
+  EXPECT_EQ(conv_nodes_count, conv_count);
+  EXPECT_EQ(quantize_nodes_count, quant_count);
+  EXPECT_EQ(dequantize_nodes_count, dequant_count);
+  EXPECT_EQ(original_nodes_num + added_nodes_count, current_nodes_num);
```
wojtuss, line 447: Please, verify also that the scales used for quantization of `reshape2`'s input and output are the same.

paddle/fluid/inference/api/mkldnn_quantizer_config.cc
wozna: Done

paddle/fluid/operators/reshape_op.cc
wozna: Done

paddle/fluid/operators/reshape_op.cc
wozna: You're right. Done

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
wozna: Done

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
```diff
@@ -365,11 +364,120 @@ TEST(CpuQuantizePass, transpose) {
...
+        std::cout << op_name << " \n";
+        if (op_name != "Dequantize1") {
+          dequant_scale = boost::get<float>(op->GetAttr("Scale"));
+          EXPECT_EQ(dequant_scale, scale)
+              << "Scale for node '" + op->Type() + "'.";
+        }
+      }
+    }
+  }
+  EXPECT_EQ(dequant_scale, quant_scale);
```
wojtuss: The equality of those scales is already verified in the above `quantize`'s and `dequantize`'s scales check.

wozna: You are right. It is corrected.

PR #21430, opened 2019-11-28
Add explicit conversion to InferNoNeedBufferVarsFN

sneaxiy:
To avoid `InferNoNeedBufferVarsFN` convert to bool implicitly.

For example, the following codes would work previously:
```
void fun(int a) {...}

InferNoNeedBufferVarsFN f;
fun(f); // f is converted to int implicitly.
```

After this PR, the implicit conversion is prohibited. The codes above would not be compiled right.


luotao1 (approved):
LGTM

PR #21444, opened 2019-11-29
[cherry-pick]Fix the crash issue when scale or bias was null-pointer. (#21284)

yihuaxu:
* Fix the crash issue when scale or bias was null-pointer.

* Add the error message for passing CI.

test=release/1.6


bingyanghuang:
cherry-pick #21284

luotao1 (approved):

PR #21393, opened 2019-11-27
NV jetson(nano, tx2, xavier) inference compile support

NHZlX:
This PR supports: 
Paddle can be compiled on nv jetson hardware for inference, which can make full use of the acceleration performance of Paddle-TRT.

When doing cmake, we should turn off some switches like WITH_MKL, WITH_MKLDNN, WITH_ XBYAK, which can not be used for arm architecture.

Use as follows：

```
cmake .. \
	      -DWITH_CONTRIB=OFF \
	      -DWITH_MKL=OFF  \
	      -DWITH_MKLDNN=OFF \
	      -DWITH_TESTING=OFF \
	      -DCMAKE_BUILD_TYPE=Release \
	      -DON_INFER=ON \
	      -DWITH_PYTHON=OFF \
               -DWITH_XBYAK=OFF  \
               -DWITH_NV_JETSON=ON \
```

Errors will be reported during compilation on jetson when using the v0.2.20. 
The Errors are as follows:

```
getarch_2nd.c: In function ‘main’:
getarch_2nd.c:12:35: error: ‘SGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function); did you mean ‘QGEMM_DEFAULT_UNROLL_M’?
     printf("SGEMM_UNROLL_M=%d\n", SGEMM_DEFAULT_UNROLL_M);
                                   ^~~~~~~~~~~~~~~~~~~~~~
                                   QGEMM_DEFAULT_UNROLL_M
getarch_2nd.c:12:35: note: each undeclared identifier is reported only once for each function it appears in
getarch_2nd.c:13:35: error: ‘SGEMM_DEFAULT_UNROLL_N’ undeclared (first use in this function); did you mean ‘SGEMM_DEFAULT_UNROLL_M’?
     printf("SGEMM_UNROLL_N=%d\n", SGEMM_DEFAULT_UNROLL_N);
                                   ^~~~~~~~~~~~~~~~~~~~~~
                                   SGEMM_DEFAULT_UNROLL_M
getarch_2nd.c:14:35: error: ‘DGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function); did you mean ‘SGEMM_DEFAULT_UNROLL_M’?
     printf("DGEMM_UNROLL_M=%d\n", DGEMM_DEFAULT_UNROLL_M);
                                   ^~~~~~~~~~~~~~~~~~~~~~
                                   SGEMM_DEFAULT_UNROLL_M
[...]

PR #21450, opened 2019-11-29
[openblas]: update openblas for nv Jetson support

NHZlX:
We add Jetson support in this PR:
https://github.com/PaddlePaddle/Paddle/pull/21393

Submit this code in advance to test whether upgrading openblas will have an impact.


luotao1 (approved):
LGTM

PR #21451, opened 2019-11-29
Refine errmsg by comparing shape diff in wrong trt input shape errmsg

cryoco:
**Problem**: When users set wrong input shape in paddle-TRT, the error message does not contain the exact input shape difference.
**Solution**: Add exact shape difference in error message, so that users can be aware of which input shape is wrong.
Now the error message summary is like:
`Error: Input shapes are inconsistent with the model. Expect [3, 473, 473] in model description, but got [3, 472, 472] in runtime. TRT 5 or lower version does not support dynamic input shapes. Please check and modify your input shapes.`


NHZlX (dismissed):
LGTM

NHZlX (approved):
LGTM

PR #21454, opened 2019-11-29
Do not quantize for concat in AddQuantDequantPass, test=develop

juncaipeng:
Do not quantize for concat in AddQuantDequantPass, test=develop


wzzju (approved):
LGTM.

PR #21458, opened 2019-11-29
fix the correctness of memcpy profiling result

wangchaochaohu:
we use event to record the GPU time but in the memcpy it record it again. In the device trace, one GPU activity can only relate one event, so it it not correct.we can konw about more information from    
 record_event("GpuMemcpyAsync:CUDAPinned->GPU") like this


sneaxiy (approved):
LGTM.

PR #21426, opened 2019-11-28
Add template version of UpdatePadding

grygielski:
This PR is needed in order to reduce number of files changed in MKL-DNN 1.0 update.

PR #21465, opened 2019-11-29
Fix bug in UpdatePadding for int64_t type

grygielski:
Little fix for https://github.com/PaddlePaddle/Paddle/pull/21426 allowing to use it for int64_t type.


luotao1 (approved):
LGTM

PR #21466, opened 2019-11-29
[MKL-DNN] Conv2d and Conv2d transpose MKL-DNN NHWC support

jczaja:
This PR is a continuation of #21375 to fulfill requirements of #20912 

Changes:
- conv2d and conv2d_transpose MKL-DNN kernels support NHWC (FWD only)
- Some refactoring pointed out #21375 e.g. TransformDataLayoutFromMKLDNN() was surrounded by PADDLE_WITH_MKLDNN .
- Proposal of refactoring of TLS MKL-DNN variables as requested in #18944 

**Note:**
-  There is performance degradation  with this PR when paddle using MKL-DNN is executing **training**
  of convolutional model. **Inference** is fine. This will be fixed when Grad NHWC is supported. If you think that this is not acceptable for time being then please let me know so I can prepare some fast fix.

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -396,8 +396,8 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     // of conv int8 mkl-dnn. Once conv fp32 and conv int8
     // are merged/unified, this will disappear
     std::string key_tid = "";
-    if (platform::get_mkldnn_tls().get_cur_mkldnn_session_id() ==
-        platform::MKLDNNTLS::kMKLDNNSessionID_Default) {
+    if (platform::get_cur_mkldnn_session_id() ==
+        platform::kMKLDNNSessionID_Default) {
       key_tid = "-t:" + platform::ThreadIDasStr();
     }
 
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -224,7 +224,10 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     // gradient computation proper as this op is called directly without
     // fetch op following it , so numercial grad is computed (in python)
     // using block formats which will give wrong results
-    auto chosen_memory_format = is_test ? MKLDNNMemoryFormat::any : src_format;
+    std::string data_format = ctx.Attr<std::string>("data_format");
+    auto chosen_memory_format =
+        is_test ? MKLDNNMemoryFormat::any
+                : platform::data_format_to_memory_format(data_format);
 
     weights_format = MKLDNNMemoryFormat::any;
     // Check the format for user's special output
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -220,9 +220,11 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
      * ('any') which lets a primitive (convolution in this case) choose
      * the memory format preferred for best performance
      */
-    std::string data_format = ctx.Attr<std::string>("data_format");
-    auto chosen_memory_format =
-        platform::data_format_to_memory_format(data_format);
+    // TODO(jczaja): This is workaround to make grad op UT's numerical
+    // gradient computation proper as this op is called directly without
+    // fetch op following it , so numercial grad is computed (in python)
+    // using block formats which will give wrong results
+    auto chosen_memory_format = is_test ? MKLDNNMemoryFormat::any : src_format;
 
     weights_format = MKLDNNMemoryFormat::any;
     // Check the format for user's special output
@@ -394,8 +396,8 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     // of conv int8 mkl-dnn. Once conv fp32 and conv int8
     // are merged/unified, this will disappear
     std::string key_tid = "";
-    if (platform::get_cur_mkldnn_session_id() ==
-        platform::kMKLDNNSessionID_Default) {
+    if (platform::get_mkldnn_tls().get_cur_mkldnn_session_id() ==
+        platform::MKLDNNTLS::kMKLDNNSessionID_Default) {
       key_tid = "-t:" + platform::ThreadIDasStr();
     }
 
@@ -519,9 +521,7 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
       * ('any') which lets a primitive (convolution in this case) choose
       * the memory format preferred for best performance
       */
-      std::string data_format = ctx.Attr<std::string>("data_format");
-      auto chosen_memory_format =
-          platform::data_format_to_memory_format(data_format);
+      auto chosen_memory_format = MKLDNNMemoryFormat::any;
 
       std::vector<int> bias_tz;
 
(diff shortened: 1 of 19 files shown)


luotao1:
@jczaja 
> Proposal of refactoring of TLS MKL-DNN variables as requested in #18944

The coverage check fails on this refactoring. Maybe you can refactor in the next PR.
image

> There is performance degradation with this PR when paddle using MKL-DNN is executing the training
of the convolutional model. The inference is fine. This will be fixed when Grad NHWC is supported. If you think that this is not acceptable for time being then please let me know so I can prepare some fast fix.

It is not acceptable for the time. How much degradation with this PR on the convolutional training model? Does some fast fix to solve this problem?

@LeoZhao-Intel Could you help to review this PR?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -220,9 +220,11 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
      * the memory format preferred for best performance
      */
-    std::string data_format = ctx.Attr<std::string>("data_format");
-    auto chosen_memory_format =
-        platform::data_format_to_memory_format(data_format);
+    // TODO(jczaja): This is workaround to make grad op UT's numerical
+    // gradient computation proper as this op is called directly without
+    // fetch op following it , so numercial grad is computed (in python)
+    // using block formats which will give wrong results
+    auto chosen_memory_format = is_test ? MKLDNNMemoryFormat::any : src_format;
```
LeoZhao-BJ: can you describe a bit of details about this workaround? why change to use "any" format by default?

jczaja: @LeoZhao-Intel before this PR *data_format* was used in convolution to control which implementation will be executed by MKL-DNN eg. Any or NCHW. It could be left that way , but when NHWC is set to data_format in a model then performance goes drastically down.
To avoid performance degradation We no longer restrict MKL-DNN choice of implementation with *data_format* , but we use this information recognize what is the Paddle data format as it comes to model and what should be Paddle data format when MKL-DNN finish its work. 
Hence *choice_memory_format* is be default *any* to be the fastest. Why it is equal to src_format for training? It is because Grad UTs for mkl-dnn convolution are assuming that
convolution mkl-dnn op is computing FWD in NCHW. So as a temporary workaround
I put for training a *src_format* which for models with training is likely to be NCHW or NHWC.
This reduce performance of training. It is to be fixed once I have Unit tests for conv mkl-dnn GRAD working properly with Blocked layout.

@luotao1 I will restore performance of convolution in training & and revert TLS refactor in this PR

luotao1:
Is batch_norm NHWC support a new PR after this PR is merged?

jczaja:
@luotao1 I just checked that *data_format* does not exist in batch norm, but there is *data_layout*  which I haven;t noticed before. So batch_norm NHWC is not implemented yet. I will start doing it this week after maintaining of this PR

jczaja:
@luotao1 I have reverted TLS refactoring and removed partially training performance on MKL-DNN.
Full restoration of training performance is only possible by introducing NHWC support to Grad ops.

luotao1:
> Full restoration of training performance is only possible by introducing NHWC support to Grad ops

Could you tell us why NHWC support to ops will decrease the training performance? And why full restoration will be only possible by introducing NHWC support to Grad ops?

jczaja:
@luotao1 ,
1. As mentioned earlier GRAD UnitTests do not work with "Any" layout (MKL-DNN has to have enforced NCHW implementation). This is because , when numeric gradient is computed , no executor/analyzer is used just direct op execution:
`op->Run()`
At that situation there is no conversion of output to NCHW/NHWC so numeric grad is computed using
block formats so results are incorrect. Currently to have this passed Unit tests were setting *data_format* to NCHW.   When We implement GRAD NHWC then at the same time we make
Unit tests working with "Anylayout" and there is no longer need to enforce NCHW in grad tests. 
Short answer is to have training full performance We have to implement this:
https://github.com/PaddlePaddle/Paddle/issues/20964#issuecomment-559671308

2. Why performance of training was good? Because default value of *data_format* was "AnyLayout" not NCHW or NHWC like it is now. With "AnyLayout" as a value  of *data_format* there was fast execution
of MKL-DNN ops. But after making NCHW a default value of *data_format* performance goes down. I would like to remove old meaning of *data_format* e.g. enforcing MKL-DNN implementation data layout. I have made (this PR) inference working fast, but GRAD ops are using NCHW slow implementation. 

3. I haven't checked actual training performance (our CI does not check training performance) , so my note on performance degradation was a bit theoretical (based on code). In fact, what I wrote should suggest that performance degradation happened when default values *data_format* were changed from _AnyLayout_ to _NCHW_ . So in that case this PR is not introducing  degradation of  performance.

4. To conclude, currently this PR should not decrease performance on any models. If there is poor training performance on some model, it is likely due to changes of default *data_format* value to "NCHW" from "AnyLayout" (this modification is already on develop).

5. If you could test training performance before/after this PR that would be valuable information.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -220,9 +220,14 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
      * the memory format preferred for best performance
      */
+    // TODO(jczaja): This is workaround to make grad op UT's numerical
+    // gradient computation proper as this op is called directly without
+    // fetch op following it , so numercial grad is computed (in python)
+    // using block formats which will give wrong results
     std::string data_format = ctx.Attr<std::string>("data_format");
     auto chosen_memory_format =
-        platform::data_format_to_memory_format(data_format);
+        is_test ? MKLDNNMemoryFormat::any
```
LeoZhao-BJ, line 229: is this correct?  should be like below ?

is_test ? platform::data_format_to_memory_format(data_format): MKLDNNMemoryFormat::any;

LeoZhao-BJ (commented):
Jacek,

Can you describe the status with your latest commit in PR?
all UTs passed? but training performance still not good?

luotao1:
> If you could test training performance before/after this PR that would be valuable information.

@bingyanghuang Could you or Intel Validation Team test the training performance before/after this PR? Our CI/CE doesn't include the MKLDNN test now.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: yes. UnTill GRAD NHWC is working this is to be like that. Do you have any question on that one?

jczaja:
@luotao1 , @LeoZhao-Intel All UT pass and performance should be good. The only problem is that
our internal CI doesn't check training performance , so I was not able to check that. To conclude : no known problems present. My suggestion is to have training performance checked before/after this PR.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: Yes. When there is inference/test then mkl-dnn is free to choose conv implementation (MKLDNNMemoryFormat::any) . When there is a training we check if user enforced *data_format* and if positive then there are two use cases: AnyLayout and NHWC/NCHW .  For AnyLayout again MKL-DNN is free to choose implementation of conv for NCHW/NHWC MKL-DNN will use NCHW/NHWC implementation. This is needed for GRAD Unit tests to pass (This will be improved once GRAD ops NHWC is implemented). As mentioned earlier using data_format to enforce implementation of MKL-DNN is leading to low performance , but this already happened in previous PRs , this PR does not make any additional harm and make sure performance is good for inference.

LeoZhao-BJ:
> @luotao1 , @LeoZhao-Intel All UT pass and performance should be good. The only problem is that
> our internal CI does check training performance , so I was not able to check that. To conclude : no known problems present. My suggestion is to have training performance checked before/after this PR.

Jacek, my understanding on your answer are:
1. Your current PR doesn't make training performance worse, the previous PR made training performance bad, right?
2. To accommodate the fwd op change and pass UT, you did that change as I commented there, https://github.com/PaddlePaddle/Paddle/pull/21466#discussion_r352959790  , right?

jczaja:
@LeoZhao-Intel 
1. Yes. My PR did not make performance worse. Previous PRs **could** make inference/training performance worse (python workloads that are setting *data_format*). @wojtuss Was responsing to one of issues on performance degradation when NHWC was introduced . Perhaps he can point us to relevant issue on gihub (I cannot find it myself).
2. Yes. To make inference performance good and make sure Grad conv UT pass , pointed change was added

LeoZhao-BJ:
LGTM.

grygielski (approved):
LGTM

luotao1 (approved):

luotao1:

Issue #21232, opened 2019-11-19
【MKL】Different output  with different version of libmklml_intel.so


Aurelius84:
I'm training a model on CPU machine with paddle, but the precision (`PNR`) can not converge up to the baseline with same train dataset. 

I found that the result trained by paddle whl with `2014MKL` is better than  the `2019MKL`. 

+ `2019MKL` means to compile paddle with 2019 `libmklml_intel.so`, which is used by default in Paddle.
+ `2014MKL` means to compile paddle with 2014 `libmklml_intel.so`
+ The result of `Baseline` is trained by other framework tools based on 2014MKL.

There are many `Matrix multiplication` in my model, such as `blas.MatMul` or `blas.GEMM`. I found the output of `Matrix multiplication` has different result while using different version `libmklml_intel.so` to compile paddle whl.

I'm not sure how the difference output of `MatMul` or `GEMM` influence the precision. I would appreciate it very much if relevant developers can help to follow this problem.

**See details as follows:**

![image](https://user-images.githubusercontent.com/9301846/69114722-28575780-0ac1-11ea-82f1-9497e7f711e1.png)

| result | Baseline         |  2019MKL |2014MKL |
| ----------------- | ------------ | ------------ | ------ |
| max(PNR)         | 2.4207       | 2.3924      |  **2.4303**  |

### 1. Enviroment & Version
+ CentOS release 6.3 (Final)
+ docker image: `docker.paddlepaddlehub.com/paddle_manylinux_devel:cuda8.0_cudnn7`
+ python: `2.7.15`

### 2. build paddle
+ Compile with 2019 MKL

```bash
1. git clone https://github.com/PaddlePaddle/Paddle.git
2. cd Paddle & make build
3. comopile
export LD_LIBRARY_PATH=/opt/_internal/cpython-2.7.11-ucs4/lib:${LD_LIBRARY_PATH#/opt/_internal/cpython-2.7.11-ucs2/lib:}

[...]

PR #21471, opened 2019-12-02
allow preloading for mklml

LeoZhao-BJ:
check if symbol is preloaded first, then dlsym.

test=develop

### DIFF
--- a/paddle/fluid/platform/dynload/mklml.h
+++ b/paddle/fluid/platform/dynload/mklml.h
@@ -28,18 +28,20 @@ extern void* mklml_dso_handle;
  * (for each function) to dynamic load mklml routine
  * via operator overloading.
  */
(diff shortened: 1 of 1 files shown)


LeoZhao-BJ:
Issue #21232

LeoZhao-BJ:
@luotao1 @yinghu5

LeoZhao-BJ:
one potential issue:
When MKLDNN is ON in Paddle cmake, the paddle framework library is linked with libmklml_intel.so natually, with this PR, LD_PRELOAD still works, but it breaks MKL dlopen/dlsym routine code.

dlsym(RTLD_DEFAULT,xxx) will always return true by finding symbol from global since  libmklml_intel.so is linked already, it will always skip next dlsym(), in theory no issue should happen since the mkl lib is same, but code coverage is different.

PR #21477, opened 2019-12-02
fix bug in example codes of API case and switch_case

liym27:
#### fix bug:
Examples code in api `case` and `switch_case` run failed.


saxon-zh (approved):
LGTM

Boyan-Liu (approved):

PR #21478, opened 2019-12-02
Fix AdamOptimizer and Scale sample code Bug

Aurelius84:
+ Fix python function sample code in AdamOptimizer
+ Fix python function sample code in Scale


heavengate (approved):

Boyan-Liu (approved):
LGTM

PR #21491, opened 2019-12-02
Fix dgc clip value & rampup step

wangxicoding:
1. DGC clip value should be
image
2. rampup should be started after rampup_begin_step. After rampup end, the last sparsity should be used.


gongweibao (approved):
LGTM

PR #21498, opened 2019-12-03
Remove eval() calls in Eigen

sneaxiy:
Eigen `eval()` would allocate temporary memory to store the immediate result, which causes the memory consumption raise. This PR removes them to save memory.

### DIFF
--- a/paddle/fluid/operators/batch_norm_op.cc
+++ b/paddle/fluid/operators/batch_norm_op.cc
@@ -522,7 +522,7 @@ class BatchNormGradKernel<platform::CPUDeviceContext, T>
       EigenVectorArrayMap<T> inv_var_tmp(running_inv_var_data, C);
       ConstEigenVectorArrayMap<T> var_arr(running_variance->data<T>(), C);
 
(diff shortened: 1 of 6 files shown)


chenwhql (approved):
LGTM

zhangting2020:
Eigen's official documentation mentions that calling eval before broadcasting will speed up the computation. However it is removed in this PR. Are there any experiments showing how much the 
memory is saved? And is there any impact on performance? @sneaxiy  @chenwhql

sneaxiy:
@zhangting2020 `eval()` would allocate memory to store the intermediate result. The memory consumption relies on the model. For example, ResNet usually applies activation after convolution/batch normalization. The output of convolution/batch normalization is usually a very large Tensor. 

> Eigen's official documentation mentions that calling eval before broadcasting will speed up the computation.

Broadcast in Eigen is a very slow operation even though the operand of broadcast is a Eigen Tensor (i.e, pure Eigen Tensor like `a`) or Eigen TensorMap (i.e., Eigen operator like `a+1`). I recommend that you should avoid to use broadcast in Eigen. To write more efficient CUDA kernels is usually a easy job if you use GPU.

PR #21504, opened 2019-12-03
Add _use_system_allocator to some op tests

sneaxiy:
Add `_use_system_allocator(True)` in some op tests that do not use `OpTest` framework.


chenwhql (approved):
LGTM

PR #21505, opened 2019-12-03
Fix transforming tensor bug in bwd of batch_norm op and add more tests

Jie-Fang:
1. fix bug of transforming tensor in BatchNormGradKernel::Compute, add x dims constraint. BatchNormKernel::Compute has added, but forgot to add in GradKernel.

2. fix bug in batchnorm unittest, change data_format to data_layout.

3. add more situations on amp unittest in order to cover batchnorm op test.


Jie-Fang:
4. move resize and transform function to transform.h to reduce the code reuse in the batchnorm and conv2d.

python/paddle/fluid/nets.py
```diff
@@ -146,7 +146,8 @@ def img_conv_group(input,
                    conv_batchnorm_drop_rate=0.0,
                    pool_stride=1,
                    pool_type="max",
-                   use_cudnn=True):
+                   use_cudnn=True,
+                   data_format="NCHW"):
```
lanxianghit, line 150: 新增参数data_format的默认值是NCHW，是否与之前的默认行为相符？

gongweibao: 相符。

PR #21512, opened 2019-12-03
[Cherry-pick] set dim[0] to -1 if dim[0] < 0 during compiling for c_allgather op #21402

sandyhouse:
* set dim[0] to -1 if dim[0] < 0 and move assertion to runtime


gavin1332 (approved):
LGTM

PR #21514, opened 2019-12-03
[Cherry-pick] fix the computation for dx (grad for x) for prelu operation. (#20949)

sandyhouse:
* set the default value of alpha for prelu to 0.25, test=develop

* add the call to __syncthreads(), test=develop

* fix the implementation of cpu prelu, test=develop

* repair the implementation of element mode prelu, test=develop

* modify test_prelu_op.py, test=develop


gavin1332 (approved):
LGTM

PR #21517, opened 2019-12-03
remove elementwise x_should_larger_than_y assert in math_op_patch

danleifeng:
1. remove assert to support x< y
2. if other_var.shape[0] == -1,  axis = 0 is not useful and limited, thus remove it.


phlrain (approved):
LGTM

PR #21519, opened 2019-12-03
[Cherry-pick 1.6] Fix dgc clip & rampup step

wangxicoding:
cherry-pick from https://github.com/PaddlePaddle/Paddle/pull/21491
1. DGC clip value should be
image
2. rampup should be started after rampup_begin_step. After rampup end, the last sparsity should be used.


gongweibao (approved):
LGTM

PR #21524, opened 2019-12-03
dygraph_grad_maker supports varbase without grad_var

zhiqiu:
Originally, `dygraph_grad_maker`  raise exception when forward var's grad_var is nullptr.

This PR fixes that by creating a new grad_var if var's grad_var is nullptr  in `dygraph_grad_maker`.


paddle/fluid/imperative/layer.h
```diff
@@ -87,6 +87,15 @@ class VarBase {
 
   const std::shared_ptr<VarBase>& GradVarBase() const { return grad_var_; }
 
+  void ClearGradVarBase() { grad_var_ = nullptr; }
+
+  std::shared_ptr<VarBase>& MutableGradVarBase() {
```
sneaxiy: Why return non-const reference? If you do not want to call `std::shared_ptr<VarBase>::reset` method or assign `std::shared_ptr<VarBase>` to `nullptr`, you need not return a mutable `std::shared_ptr<VarBase>`.

zhiqiu: Changed to const reference, thanks.

sneaxiy (dismissed):
LGTM.

paddle/fluid/imperative/engine.h
```diff
@@ -51,6 +51,12 @@ class Engine {
     grad_ops_[op] = std::move(op_shared);
   }
 
+  const std::unordered_set<VarBase*>& GradVars() const { return grad_vars_; }
```
phlrain, line 54: GradVars never be called

paddle/fluid/imperative/tests/test_tracer.cc
```diff
@@ -296,6 +300,46 @@ TEST(test_tracer, test_expected_place) {
...
+
+  var_pair x_pair = var_pair("X", vb_vector(1, x_in));
+  var_pair y_pair = var_pair("Y", vb_vector(1, y_in));
+  var_pair out_pair = var_pair("Out", vb_vector(1, vout));
+  imperative::NameVarBaseMap ins = {x_pair, y_pair};
+  imperative::NameVarBaseMap outs = {out_pair};
+  framework::AttributeMap mul_attr_map;
+  mul_attr_map["use_mkldnn"] = false;
+  tracer.TraceOp("mul", ins, outs, mul_attr_map, place, true);
+  const auto& out_tensor = vout->Var().Get<framework::LoDTensor>();
```
phlrain, line 342: 这个地方，是不是可以加一个backward的测试

zhiqiu: 是的，我加一下。

paddle/fluid/imperative/engine.h
zhiqiu: Thanks, I will add usage it in `test_tracer.cc ` to check `grad_vars_ ` not empty.

phlrain (approved):

PR #21537, opened 2019-12-04
Refactor fetch handler (#21264)

seiriosPlus:
* fix fetch handler problem and refactor
when a user define FetchHandler class, he or she should initialize a handler
with variable dict. the key of a variable dict is a user defined name,
the value of a variable dict is a Varaible generated from python API.

For each fetching, a user should implement handler function in which
fetched_result_dict will be available and the user can access the fetched value
with user defined keys.


guru4elephant (approved):
LGTM

PR #21538, opened 2019-12-04
let WHTI_XBYAK can be adjusted by -D when cmake,test=develop

zhwesky2010:
let WHTI_XBYAK can be adjusted by -D when cmake,test=develop, now it is ON and can't be changed.


NHZlX (approved):
LGTM

PR #21540, opened 2019-12-04
add approver and remove useless information

silingtong123:
add approver and remove useless information according to @tianshuo78520a's suggestion

fix https://github.com/PaddlePaddle/Paddle/pull/21483 (review)


liupluswei (approved):
LGTM

PR #21541, opened 2019-12-04
Unify the rank of prelu alpha to 4, except for the all mode.

sandyhouse:
Unify the rank of prelu alpha to 4, corresponding to [N, C, H, W], except for the all mode.


gavin1332 (approved):
LGTM

PR #21545, opened 2019-12-04
bugfix: construct a DistributedStrategy instance if the passed one is None

sandyhouse:
If the passed strategy for fleet.distributed_optimizer is None, the program will throw exception.
This pr addressed the problem by construct a DistributedStrategy instance if the passed one is None.


mapingshuo (approved):
LGTM

PR #21561, opened 2019-12-05
Polish op registry codes

sneaxiy:
Refine `op_registry.h` codes by:
- Force that each field of `OpInfo` only registers once.
- Make `infer_shape_` registered when `import paddle.fluid` instead of when calling `OpDesc::InferShape` for the first time.


luotao1 (approved):
LGTM

PR #21564, opened 2019-12-05
Add grad maker assertion

sneaxiy:
For operators with kernels, both static graph grad maker and dygraph grad maker should be registered or not registered in the same time. This PR adds assertion to meet this requirements to avoid unexpected bugs.


chenwhql (approved):
LGTM

PR #21567, opened 2019-12-05
[cherry-pick] construct a DistributedStrategy instance if the passed one is None

sandyhouse:
cherry-pick #21545
Construct a DistributedStrategy instance if the passed one is None which will throw exception.


mapingshuo (approved):
LGTM

PR #21569, opened 2019-12-05
Add op function generator for dygraph

zhiqiu:
This PR tries to generate `pybind` functions for all operators automatically in dygraph mode, so In python side, each `layers.xx` function can call the corresponding generated `pybind` functions directly, other than `append_op` like static graph mode.

The reason for doing that  is to execute a `fast` path in dygraph mode, so the time cost in Python side could be reduced.

### DIFF
--- a/paddle/fluid/imperative/tracer.h
+++ b/paddle/fluid/imperative/tracer.h
@@ -58,6 +58,9 @@ class Tracer {
                const NameVarBaseMap& outs, framework::AttributeMap attrs,
                const platform::Place& place, bool trace_bacward);
 
+  void TraceOp(const std::string& type, const NameVarBaseMap& ins,
+               const NameVarBaseMap& outs, framework::AttributeMap attrs);
+
   bool ComputeRequiredGrad(const NameVarBaseMap& ins,
                            const NameVarBaseMap& outs, bool trace_backward);
 
--- a/paddle/fluid/pybind/op_function_generator.cc
+++ b/paddle/fluid/pybind/op_function_generator.cc
@@ -23,19 +23,8 @@
 #include "paddle/fluid/pybind/pybind.h"
 #include "paddle/fluid/string/string_helper.h"
 
-const int BUFFER_SIZE = 4096;
-
-const char* INPUT_TYPE = R"(imperative::NameVarBaseMap)";
-const char* ARGS_TYPE = R"(framework::AttributeMap)";
-
-const char* OUT_VAR_NUM_TYPE = R"(std::map<std::string, size_t>)";
-const char* RETURN_TYPE = R"(imperative::NameVarBaseMap)";
-
-const char* OUTS_INITIALIZER =
-    R"({{"out": VarBase("mul_out")}, "XX": VarBase("mul_xx")}})";
-
 const char* OUT_INITIALIZER_TEMPLATE =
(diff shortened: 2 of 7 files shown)


paddle/fluid/pybind/op_function_generator.cc
```diff
@@ -0,0 +1,163 @@
...
+      outs_initializer += ",";
+    }
+    if (outs_initializer.back() == ',') {
+      outs_initializer.pop_back();
+    }
+    outs_initializer += "}";
+
+    // generate op funtcion body
+    char op_function_buf[BUFFER_SIZE];
+    snprintf(op_function_buf, BUFFER_SIZE, OP_FUNCTION_TEMPLATE,
```
sneaxiy: How about use `paddle::string::Sprintf` instead of fixed max length `BUFFER_SIZE`?

paddle/fluid/pybind/op_function_generator.cc
```diff
@@ -0,0 +1,163 @@
...
+             outs_initializer.c_str(), op_type.c_str());
+
+    // generate pybind line
+    char pybind_buf[BUFFER_SIZE];
+    snprintf(pybind_buf, BUFFER_SIZE, PYBIND_TEMPLATE, module_name.c_str(),
```
sneaxiy: Same above.

paddle/fluid/pybind/op_function_generator.cc
```diff
@@ -0,0 +1,163 @@
...
+      tracer->TraceOp("%s", std::move(ins), std::move(outs), std::move(attrs));
+      return outs;
+    }
+  }, py::arg("ins"), py::arg("attrs")=framework::AttributeMap(), 
+  py::arg("outs")=imperative::NameVarBaseMap(), 
+  py::arg("out_nums")=std::map<std::string, size_t>())";
+
+const char* PYBIND_TEMPLATE = R"(%s.def("%s", %s);)";
+
+static std::string RefineName(std::string name) {
```
sneaxiy: Remove unused function.

paddle/fluid/imperative/tracer.h
```diff
@@ -58,6 +58,9 @@ class Tracer {
                const NameVarBaseMap& outs, framework::AttributeMap attrs,
                const platform::Place& place, bool trace_bacward);
 
+  void TraceOp(const std::string& type, const NameVarBaseMap& ins,
```
sneaxiy, line 61: Think about how to remove the first version `TraceOp` in the future.

zhiqiu: The first version `TraceOp` can be removed when all the legacy layers migrated to new generated op function.

paddle/fluid/pybind/op_function_generator.cc
zhiqiu: Done, thanks.

paddle/fluid/pybind/op_function_generator.cc
zhiqiu: Done, thanks.

paddle/fluid/pybind/op_function_generator.cc
zhiqiu: Done, thanks.

sneaxiy (approved):
Excellent!

PR #21582, opened 2019-12-05
Enhance checking on sub branch of backward

Aurelius84:
**What's new?**
+ enhance checking `params_grad` 
+ enhance checking `stop_gradient_grad_vars`
+ enhance checking `_find_op_path_`

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_backward.py
+++ b/python/paddle/fluid/tests/unittests/test_backward.py
@@ -16,21 +16,116 @@ from __future__ import print_function
 
 import unittest
 import paddle.fluid as fluid
-import collections
-from simple_nets import init_data
+import numpy as np
 
 
-class Simple_Net():
+class BackwardNet(object):
     def __init__(self):
-        self.stop_gradient_vars = []
-        self.no_grad_set = []
+        self.stop_gradient_grad_vars = []
+        self.no_grad_vars = []
+        self.params_names = []
+        self.op_path = []
+
+    def build_model(self):
+        raise NotImplementedError
+
+    def init_data(self):
+        raise NotImplementedError
+
+    def check_backward(self, loss, main_program):
+        params_grads = self.check_params_grad(loss)
+        # 1.1 get_stop_gradients
+        no_grad_dict = self.check_stop_gradient(main_program)
+        # 1.2 find_op_path
+        op_path, block_no_grad_set = self.check_op_path(
+            main_program.block(0), [loss], [], no_grad_dict)
+        # 1.3 _find_no_grad_vars
+        no_grad_vars = self.check_find_no_grad_vars(
+            main_program.block(0), op_path, [loss], block_no_grad_set)
+        # update no_grad_dict
+        block_no_grad_set.update(no_grad_vars)
+        no_grad_dict[0].update(
+            list(map(fluid.backward._append_grad_suffix_, block_no_grad_set)))
+
+    def check_params_grad(self, loss, parameter_list=None, no_grad_set=None):
+        params_grads = fluid.backward.append_backward(loss, parameter_list,
+                                                      no_grad_set)
+        params_names = set([var[0].name for var in params_grads])
+        assert len(params_names - set(self.params_names)) == 0
+
+        return params_grads
+
+    def check_stop_gradient(self, program):
+        no_grad_dict = fluid.backward._get_stop_gradients_(program)
+        if no_grad_dict is not None and isinstance(no_grad_dict, dict):
+            assert len(no_grad_dict[0] - set(self.stop_gradient_grad_vars)) == 0
+
+        return no_grad_dict
+
+    def check_op_path(self, root_block, outputs, inputs=[], no_grad_dict=None):
+        if no_grad_dict is None or not isinstance(no_grad_dict, dict):
+            block_no_grad_set = None
+        else:
+            block_no_grad_set = set(
+                map(fluid.backward._strip_grad_suffix_, no_grad_dict[0]))
+        op_path = fluid.backward._find_op_path_(root_block, outputs, inputs,
+                                                block_no_grad_set)
+        op_types = [op.type for op in op_path]
+        assert op_types == self.op_path
+
+        return op_path, block_no_grad_set
+
+    def check_find_no_grad_vars(self, root_block, op_path, targets,
+                                block_no_grad_set):
+        no_grad_vars = fluid.backward._find_no_grad_vars(
+            root_block, op_path, targets, block_no_grad_set)
+        assert len(no_grad_vars) == len(self.no_grad_vars)
+
+        return no_grad_vars
+
+
+class SimpleNet(BackwardNet):
+    def __init__(self):
+        super(BackwardNet, self).__init__()
+        self.stop_gradient_grad_vars = [
+            u'x_no_grad@GRAD', u'x2_no_grad@GRAD', u'x3_no_grad@GRAD',
+            u'label_no_grad@GRAD'
+        ]
+        self.no_grad_vars = []
+        self.params_names = [u'w2v', u'fc_predict.b_0', u'fc_w']
+        self.op_path = [
+            u'lookup_table_v2',
+            u'lookup_table_v2',  # embedding
+            u'elementwise_add',  # merge
+            u'mul',
+            u'elementwise_add',
+            u'softmax',  # fc
+            u'elementwise_sub',
+            u'square',
+            u'mean'
+        ]  # loss
+        self.shape = [16, 50]
+
+    def init_data(self):
+        assert len(self.shape) == 2
+        x = np.random.randint(0, 90, self.shape).astype('int64')
+        x2 = np.random.randint(0, 90, self.shape).astype('int64')
+        x3 = np.random.randint(0, 90, self.shape).astype('int64')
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
-    return loss
-
-
-def case4_with_no_grad_op_maker():
-    out = fluid.layers.gaussian_random(shape=[20, 30])
-    loss = fluid.layers.mean(out)
-    return loss
+import numpy as np
+
+
```
zhhsplendid, line 21: You can add python doc for this class to illustrate that it is a base class.

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
+                                    param_attr=fluid.ParamAttr(name='fc_w'),
+                                    name='fc_no_use')
+        # loss
+        cost = fluid.layers.square_error_cost(input=predict, label=label)
+        loss = fluid.layers.mean(cost, name='mean_loss')
+
+        return loss
+
+
+# NotImplemented
```
zhhsplendid: If it is a TODO thing, add TODO comment, like

TODO(Liujie Zhang): add conditional network test

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
+            main_program.block(0), [loss], [], no_grad_dict)
+        # 1.3 _find_no_grad_vars
+        no_grad_vars = self.check_find_no_grad_vars(
+            main_program.block(0), op_path, [loss], block_no_grad_set)
+        # update no_grad_dict
+        block_no_grad_set.update(no_grad_vars)
+        no_grad_dict[0].update(
+            list(map(fluid.backward._append_grad_suffix_, block_no_grad_set)))
+
+    def check_params_grad(self, loss, parameter_list=None, no_grad_set=None):
```
zhhsplendid: The inherited classes only calls check_backward, and check_params_grad is called by check_backward. You should rename check_params_grad to _check_params_grad to indicate it is used only in this class and inherited classes doesn't have to care about it. Same for other "private check" methods

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
+# NotImplemented
+class ConditionalNet(BackwardNet):
+    def __init__(self):
+        super(BackwardNet, self).__init__()
 
 
 class TestBackward(unittest.TestCase):
-    def check_backward(self, model, feed_dict):
+    def check_backward(self, net):
         place = fluid.CPUPlace()
```
zhhsplendid: Also put CUDAPlace(0), such as:

 place = fluid.CUDAPlace(0) if core.is_compiled_with_cuda(
        ) else fluid.CPUPlace()

I found that CUDA place could cause errors sometimes. If you already know there is a error in CUDA place, put TODO

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
+        # update no_grad_dict
+        block_no_grad_set.update(no_grad_vars)
+        no_grad_dict[0].update(
+            list(map(fluid.backward._append_grad_suffix_, block_no_grad_set)))
+
+    def check_params_grad(self, loss, parameter_list=None, no_grad_set=None):
+        params_grads = fluid.backward.append_backward(loss, parameter_list,
+                                                      no_grad_set)
+        params_names = set([var[0].name for var in params_grads])
+        assert len(params_names - set(self.params_names)) == 0
```
zhhsplendid: You convert params_names from a list to a set, can we initialize the params_names as a set at the beginning?
Same for self.stop_gradient_grad_vars.

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
+            main_program.block(0), op_path, [loss], block_no_grad_set)
+        # update no_grad_dict
+        block_no_grad_set.update(no_grad_vars)
+        no_grad_dict[0].update(
+            list(map(fluid.backward._append_grad_suffix_, block_no_grad_set)))
+
+    def check_params_grad(self, loss, parameter_list=None, no_grad_set=None):
+        params_grads = fluid.backward.append_backward(loss, parameter_list,
+                                                      no_grad_set)
+        params_names = set([var[0].name for var in params_grads])
```
zhhsplendid: "for var in params_grads"

"var" may not be a good name here. Because it is "a pair of param and its grad", so "var[0]" is the actual var.

Rename "var" to "param_grad_var_pair", "param_grad_pair", or something like that. Then param_grad_var_pair[0] will be easier to read.

python/paddle/fluid/tests/unittests/test_backward.py
zhhsplendid: Two problems here:

1. Assert is a running code, not a test code. You had better use assertEqual. Because "assert" statements aren't run when Python is run in "-O" (optimize) mode.

2. In python you can write set1 == set2, or assertEqual(set1, set2)

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
+        assert len(params_names - set(self.params_names)) == 0
+
+        return params_grads
+
+    def check_stop_gradient(self, program):
+        no_grad_dict = fluid.backward._get_stop_gradients_(program)
+        if no_grad_dict is not None and isinstance(no_grad_dict, dict):
+            assert len(no_grad_dict[0] - set(self.stop_gradient_grad_vars)) == 0
```
zhhsplendid: Readability:

People may not know no_grad_dict[0] is the set at global_block (0 is the idx of global_block)
You can change:

global_block_idx = 0
no_grad_dict[global_block_idx] ...

Or

global_block_no_grad_set = no_grad_dict[0]
assertEqual(global_block_no_grad_set, self.stop_gradient_grad_vars)

Same for the following code

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,173 @@
...
+        op_types = [op.type for op in op_path]
+        assert op_types == self.op_path
+
+        return op_path, block_no_grad_set
+
+    def check_find_no_grad_vars(self, root_block, op_path, targets,
+                                block_no_grad_set):
+        no_grad_vars = fluid.backward._find_no_grad_vars(
+            root_block, op_path, targets, block_no_grad_set)
+        assert len(no_grad_vars) == len(self.no_grad_vars)
```
zhhsplendid: Would you mind also assert the no_grad_vars are same (not only the length)?

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Thanks，done.

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Already added TODO

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Done.

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Current code is ok while running on GPU. Fixed.

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Done~~

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Done.

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Yes, it's a problem. I modified TestBackward class to inherit unittest.TestCase.

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Done

python/paddle/fluid/tests/unittests/test_backward.py
Aurelius84: Fixed.

python/paddle/fluid/tests/unittests/test_backward.py
```diff
@@ -16,72 +16,205 @@
...
-    return loss
+import numpy as np
+
+
+class BackwardNet(object):
+    """
+    Abstract Base Class.
+    All Net inherited this Class should implement two functions:
+        build_model: build net to test the logic of backward
+        init_data: fake input data to test all program.
```
zhhsplendid: NIT(not important): "all program" -> "all programs"

Aurelius84: Thanks, already fixed.

Aurelius84: Thanks, already fixed.

PR #21587, opened 2019-12-05
Refine VarBase init function

zhiqiu:
1. As the title, make it more robust and readable.
2. Add more unittests for VarBase.
3. Temporarily, revert `_var_base_to_np`, which should be deprecated.


sneaxiy (approved):
LGTM.

PR #21553, opened 2019-12-04
[MKL-DNN] Batch norm mkl-dnn NHWC support

jczaja:
This PR implements NHWC Paddle layout as discussed in #20964 .

This PR is still under testing.

PR #21588, opened 2019-12-05
[MKL-DNN] Temporary fix to regression in performance of ResNet-50 training

jczaja:
Our Validation detected that **training** performance is significantly decreased after introducing:
#21466 . Problem will be properly fix When Grad NHWC is supported. But since FWD of NHWC is ported to 1.6.2 but not Grad NHWC  then it is advisible to restore training performance without waiting for proper fix.  Changes are restoring implementation of Grad of conv MKLDNN kernel like it was before #21466.


grygielski (approved):
LGTM

luotao1 (approved):
LGTM

luotao1:
@bingyanghuang Please help cherry-pick #21588 and #21553 into release/1.6, thanks very much!

PR #21596, opened 2019-12-06
Using copy method to modify repo of external dependency that can't be used directly, test=develop

zhwesky2010:
- Modify external dependency repositories that cannot be used directly, using copy method, The cp & xcopy command for patch-command is safe and fix the problem that can't patch command second time.

- It Keep the same code style with CARES_PATCH_COMMAND and GRPC_PATCH_COMMAND.

- Code style Cpplint check in Paddle/patch/ directory is exempted.


luotao1 (approved):
LGTM

raindrops2sea (approved):

PR #21604, opened 2019-12-06
Set lod_level of Out in compile time of sequence_pool_op

Aurelius84:
+ Fix compile error  if input `lod_level > 1`
```cpp
ctx->SetLoDLevel("Out", in_lod_level - 1);
```


Xreki (approved):
LGTM

PR #21609, opened 2019-12-06
Refine English Doc of cond API

zhhsplendid:
Screenshot:

image

### DIFF
--- a/python/paddle/fluid/layers/control_flow.py
+++ b/python/paddle/fluid/layers/control_flow.py
@@ -1760,24 +1760,30 @@ def cond(pred, true_fn=None, false_fn=None, name=None):
     list of tensors.
     
     Note: 
-        The tuples or lists in ``true_fn`` and ``false_fn`` must have same
-        shape because of dataflow model of PaddlePaddle while the tensors in the
-        tuples or the lists can have different shapes.
+        The tuples or lists returned by ``true_fn`` and ``false_fn`` must have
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/layers/control_flow.py
```diff
@@ -1760,24 +1760,30 @@ def cond(pred, true_fn=None, false_fn=None, name=None):
     list of tensors.
     
     Note: 
-        The tuples or lists in ``true_fn`` and ``false_fn`` must have same
-        shape because of dataflow model of PaddlePaddle while the tensors in the
-        tuples or the lists can have different shapes.
+        The tuples or lists returned by ``true_fn`` and ``false_fn`` must have
+        same shape because of dataflow model of PaddlePaddle while the tensors
```
liym27: have same shape 建议改成 have the same shape

liym27 (dismissed):

python/paddle/fluid/layers/control_flow.py
zhhsplendid: Done

zhhsplendid:
Xiang Lan commented, we should also add notice that all operations out of cond will be executed, it is not lazier semantics. I will modify it in this PR

Done

saxon-zh (approved):

lanxianghit (approved):
LGTM

PR #21629, opened 2019-12-09
Refine some grad op makers

sneaxiy:
Refine some grad maker ops to remove unnecessary inputs, to save memory consumption of models.


luotao1 (approved):
LGTM

PR #21633, opened 2019-12-09
Make OperatorWithKernel::InferShape abstract

sneaxiy:
Make `OperatorWithKernel::InferShape` method to be an abstract method, so that all `OperatorWithKernel` must overwrite this method to avoid unexpected errors (for example, so operators does not registered `infer_shape_`).


paddle/fluid/imperative/tests/test_prepare_op.cc
```diff
@@ -215,4 +225,4 @@ TEST(test_prepare_op, test_prepare_data_same_place) {
 }  // namespace paddle
 
 USE_OP(split);
```
luotao1, line 227: Why `test_prepare_op` change usage from `assign_op` to `relu_op`?

luotao1 (approved):
LGTM

paddle/fluid/imperative/tests/test_prepare_op.cc
sneaxiy: Because the previous implementation of the unittest is wrong. Previously, it use `OperatorWithKernel` to create `assign` op, so that the virtual method (for example `GetKernelTypeForVar`) called is from `OperatorWithKernel` but not from `AssignOp`. Because the `GetKernelTypeForVar` method of assign op has been overrided, this unittests would fail if we still use assign op here.

原来单测用OperatorWithKernel来构造assign op，导致调用的一些虚函数GetKernelTypeForVar是OperatorWithKernel的方法，而不是AssignOp的方法。如果改对的话（用AssignOp这个类来构造的话），单测会错，因为assign的GetKernelTypeForVar方法重写了

PR #21655, opened 2019-12-10
fix bug: add attr_error only when default value is None. 

liym27:
- Before:
The error message will be added if default value is `{}` ,`[]` or `None`. But only `None`  should be added in error message.

- This PR:
fix this bug.


luotao1:
Please use `test=document_fix` to decrease the CI time

luotao1 (approved):
LGTM

liym27:
> Please use `test=document_fix` to decrease the CI time

Thanks. `test=document_fix` will be used next commit.

PR #21661, opened 2019-12-10
Fix ci bug for deleting data files when other test is running

juncaipeng:
Fix ci bug for deleting data files when other test is running


wzzju (dismissed):
LGTM.

wzzju (approved):
LGTM.

PR #21669, opened 2019-12-10
add ssl to requirements.txt

sandyhouse:
The [pr#20942](https://github.com/PaddlePaddle/Paddle/pull/20942) fixed the [issue#20749](https://github.com/PaddlePaddle/Paddle/issues/20749) but introduced the dependency on ssl. So, we have to add ssl to requirements.txt.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -17,3 +17,4 @@ pyyaml
 decorator
 prettytable
 objgraph
+ssl
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -17,4 +17,4 @@ pyyaml
 decorator
 prettytable
 objgraph
-ssl
+ssl ; python_version>"2.3.4" and python_version<"2.6"


python/requirements.txt
```diff
@@ -17,3 +17,4 @@ pyyaml
 decorator
 prettytable
 objgraph
+ssl ; python_version>"2.3.4" and python_version<"2.6"
```
luotao1, line 20: 这个ssl能否去掉呢？
```
pip install ssl --index-url=https://pypi.python.org/simple/ --user
Looking in indexes: https://pypi.python.org/simple/
Collecting ssl
  Using cached https://files.pythonhosted.org/packages/83/21/f469c9923235f8c36d5fd5334ed11e2681abad7e0032c5aba964dcaf9bbb/ssl-1.16.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-dr8rh5/ssl/setup.py", line 12, in <module>
        + "or earlier.")
    ValueError: This extension should not be used with Python 2.6 or later (already built in), and has not been tested with Python 2.3.4 or earlier.

    ----------------------------------------
```
本地安装也会提醒必须是老版本的python才能支持

PR #21675, opened 2019-12-10
Fix elementwise_pow bug on CUDA place with integer

zhiqiu:
On CUDA place, `elementwise_pow` fails when input integers, for example,
image

image

This PR fixes that.

### DIFF
--- a/paddle/fluid/operators/elementwise/elementwise_pow_op.h
+++ b/paddle/fluid/operators/elementwise/elementwise_pow_op.h
@@ -12,16 +12,58 @@ limitations under the License. */
 #pragma once
 
 #include <cmath>
+#include <type_traits>
 #include "paddle/fluid/operators/elementwise/elementwise_op.h"
 #include "paddle/fluid/operators/elementwise/elementwise_op_function.h"
 
 namespace paddle {
 namespace operators {
 
-template <typename T>
-struct PowFunctor {
+namespace details {
+
+// Only valid if T is an integral type.
+template <class T,
+          class = typename std::enable_if<std::is_integral<T>::value>::type>
+static inline HOSTDEVICE T ipow(T base, T exp) {
+  if (exp < 0) {
+    // Note(zhiqiu): In some library, like numpy and pytorch,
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/elementwise/elementwise_pow_op.h
```diff
@@ -12,17 +12,59 @@ limitations under the License. */
...
 namespace operators {
 
-template <typename T>
-struct PowFunctor {
+namespace details {
+
+// Only valid if T is an integral type.
+template <class T,
+          class = typename std::enable_if<std::is_integral<T>::value>::type>
+static inline HOSTDEVICE T ipow(T base, T exp) {
```
sneaxiy: `static_assert(std::is_integral<T>::type, "T must be integral");`, and remove `std::enable_if`?

paddle/fluid/operators/elementwise/elementwise_pow_op.h
```diff
@@ -12,17 +12,59 @@ limitations under the License. */
...
+  if (exp < 0) {
+    // Note(zhiqiu): In some library, like numpy and pytorch,
+    // integers to negative integer powers are not allowed.
+    // Since a^b = 1/(a^(-b)) when b < 0, it will truncate to 0.
+    // Return 0, for specialization to avoiding infinite loop.
+    return 0;
+  }
+
+  T result = 1;
```
sneaxiy: Add a link where the following codes from?

paddle/fluid/operators/elementwise/elementwise_pow_op.h
zhiqiu: Thanks, implementation changed.

paddle/fluid/operators/elementwise/elementwise_pow_op.h
zhiqiu: Thanks, implementation changed.

sneaxiy (approved):
LGTM.

PR #21687, opened 2019-12-11
fix bug that tuple(Variable) is converted to list(Variable) uncorrectly

zhwesky2010:
fix bug that tuple(Variable) is converted to list(Variable) uncorrectly.

It will result this problem when input tuple(Variable)  in py_func.
image


luotao1 (approved):
LGTM

PR #21713, opened 2019-12-12
Re-anble vgg16, vgg19 and resnet101 models download

Sand3r-:
The download of these models has been temporarily disabled by @grygielski during the introduction of MKL-DNN 1.0, but our validation team needs the models to be downloaded so that the performance and accuracy of these models can be assessed. 

For now the unit tests will remain disabled, but with enabled download, the models can be ran manually.

### DIFF
--- a/paddle/fluid/inference/tests/api/CMakeLists.txt
+++ b/paddle/fluid/inference/tests/api/CMakeLists.txt
@@ -249,20 +249,20 @@ if(WITH_MKLDNN)
  
   # resnet101 int8
   # TODO(grygielski) Enable after MKL-DNN 1.0 merge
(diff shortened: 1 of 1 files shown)


pmajchrzak (approved):

wojtuss (approved):
LGTM

luotao1:
But it will cause our ci timeout due to https://github.com/PaddlePaddle/Paddle/pull/20162#issuecomment-554262103

Sand3r-:
> For now the unit tests will remain disabled, but with enabled download, the models can be ran manually.

grygielski (approved):
LGTM

PR #21722, opened 2019-12-12
Simplify dygraph DataLoader code and init LoDTensor on CPU

chenwhql:
This PR do two things:
1. The dygraph DataLoader use ListTensorProvider to init LoDTensor by Place in `set_sample_list_generator`, which split the Tensor init position of `set_sample(_list)_generator` and `set_batch_generator`, I think that is not good design.
2. The tensor init on CPU is ok, then it will copy to gpu pinned memory, then copy to gpu. The time cost has no change.


sneaxiy (approved):
LGTM.

PR #21740, opened 2019-12-13
Remove self-set accuracy parameters of op tests: max_relative_error

GaoWei8:
Remove self-set accuracy parameters of op tests:
dropout_op
pad_constant_like
sequence_softmax_op


luotao1 (approved):
LGTM

PR #21745, opened 2019-12-13
Fix the bug of memory constrained in post training quantization and support the input be different for the same model

juncaipeng:
For post training quantization:
* fix the bug of memory constrained
* support the input be different for the same model

离线量化：
* 修复离线量化中校准数据量过大的内存限制问题，将采样数据保存到磁盘
* 支持同一个模型设置任何尺寸的输出

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/post_training_quantization.py
+++ b/python/paddle/fluid/contrib/slim/quantization/post_training_quantization.py
@@ -86,7 +86,7 @@ class PostTrainingQuantization(object):
                 is_full_quantized as False, only apply quantization to the op type 
                 according to the input quantizable_op_type.
             is_memory_constrained(bool, optional): If set is_memory_constrained as False,
-                all temp data will save in memory. If set is_save_sample_data as True,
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/contrib/slim/quantization/post_training_quantization.py
```diff
@@ -150,8 +165,8 @@ def __init__(self,
 
         self._op_real_in_out_name = _op_real_in_out_name
         self._bit_length = 8
-        self._quantized_weight_var_name = []
-        self._quantized_act_var_name = []
+        self._quantized_weight_var_name = set()
```
wzzju, line 168: What is the difference between list and set here?

wzzju (approved):
LGTM.

python/paddle/fluid/contrib/slim/quantization/post_training_quantization.py
juncaipeng: > What is the difference between list and set here?

之前会重复记录并计算scale，比如一个op的输出会是下一个op的输入，使用set可以避免。

PR #21747, opened 2019-12-13
[MKL-DNN] Pool & LRN Grad Ops NHWC support

jczaja:
This PR is introducing NHWC support for LRN and pool Grad ops of MKL-DNN as discussed at #20964 

**Note:**
There are changes in mean_op.cc and unused_var_check.cc that I would like @dzhwinter or @zhangting2020 or @sneaxiy to look at. Explanation is here:
https://github.com/PaddlePaddle/Paddle/issues/20964#issuecomment-565508163

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1122,46 +1122,6 @@ Scope* OperatorWithKernel::PrepareData(
         continue;
       }
 
-      // In the inference scenerio, the scopes will be reused across the
-      // batches, so the `new_scope` here will result in GPU memroy explosion
-      // over the  running of operators.
-      // We use a thread_local cache to fix that issue, the key in the cache is
-      // the combination of the `scope` argument, from_kernel_type,
-      // target_kernel_type.
-      // Have a discussion with @Superjomn or the inference developers if some
-      // changes on this logic for this macro might not tested on the other
-      // scenerios.
-      // If this op is not called by an Executor or ParallelExecutor, it should
-      // called by a NaiveExecutor, the NaiveExecutor will cache the scopes and
-      // variables, that behavior a lot different.
-      //
-      // To solve issue #15032, have a discussion with @Luotao for cpu
-      // inference, for all cpu kernels cases without GPU participation, here
-      // not do transfer scope caching, and cpu inference performance is not
-      // impacted by test.
-      enable_cache_transfer_scope_ = false;
-      if (!run_by_executor_ &&
-          (platform::is_gpu_place(kernel_type_for_var.place_) ||
-           platform::is_gpu_place(expected_kernel_key.place_))) {
-        new_scope = TryCreateTransferScope(kernel_type_for_var,
-                                           expected_kernel_key, &scope);
-        enable_cache_transfer_scope_ = true;
-      }
-      if (!new_scope) {
-        new_scope = &scope.NewScope();
-      }
-      // For inference, if a gpu model has an op which could only run on CPU,
-      // each result of different input will be the same with the first one.
-      // The reason is that if a gpu tensor is the input of a cpu kernel,
-      // we will create a new cpu tensor in new scope.
-      // However, if enable_cache_runtime_context_, we get the cpu tensor each
-      // time, not the gpu tensor.
-      // Thus, we set pre_scope_ = nullptr to trigger `new RuntimeContext()` in
-      // RunImpl().
-      if (enable_cache_runtime_context_) {
-        pre_scope_ = nullptr;
-      }
-
       auto* tensor_in = GetLoDTensorOrSelectedRowsValueFromVar(*var);
 
       // When no_buffer_ins then checking of Tensor::holder_ is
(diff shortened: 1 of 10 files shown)


sneaxiy (changes requested):
Should not remove `DECLARE_NO_NEED_BUFFER_VARS_INFERENCE`. In this PR, you remove them to satisfy some requirements, but in the future, all `DECLARE_NO_NEED_BUFFER_VARS_INFERENCE` of all ops would be removed if similar situation is required. Try to use a common way to solve this.

paddle/fluid/framework/operator.cc
```diff
@@ -1180,12 +1178,38 @@ Scope* OperatorWithKernel::PrepareData(
         pre_scope_ = nullptr;
       }
 
-      auto* trans_var = new_scope->Var(var_name);
-      input_vars[i] = trans_var;
-
-      Tensor out;
-      TransformData(expected_kernel_key, kernel_type_for_var, *tensor_in, &out);
-      SetTensorToVariable(*var, out, trans_var);
+      if (no_buffer_ins && no_buffer_ins->count(var_name_item.first) > 0) {
```
sneaxiy: You cannot move `if (no_buffer_ins && no_buffer_ins->count(var_name_item.first) > 0)` here. 

In the design of `no_buffer_ins`, the tensor buffer of vars (i.e. the `holder_` field of Tensor) inside `no_buffer_ins` may be written (clear or written with other value) by other threads at the same time. Therefore, it is not thread safe if you read or write the tensor buffer of vars inside `no_buffer_vars`. Because the data transform does not handle these vars, the current implementation of Paddle is right when running ops in multi-threads. The only methods you can use are `Tensor::dims()`, `Tensor::lod()`, `Tensor::layout()`, `Tensor::offset()`, etc. Any Tensor methods which read or write `Tensor::holder_` are not allowed, for example `Tensor::IsInitialized()`, `Tensor::type()`, `Tensor::place()`, `Tensor::data()`, etc.

 In the current implementation of this PR, you check whether the tensor inside `no_buffer_ins ` is initialized , which may cause error in multi-threads.
https://github.com/PaddlePaddle/Paddle/blob/a400a2f37fc4d4c636171be66088641b3889d906/paddle/fluid/framework/operator.cc#L1121

jczaja: Ok, So If I understand correctly when there is no_buffer_ins we should not call Tensor methods you listed. I have updated code (only operator.cc) Please tell me if this is fine approach.

jczaja:
@grygielski , @LeoZhao-Intel Could you please take a look at this PR ?

paddle/fluid/framework/operator.cc
```diff
@@ -1120,72 +1123,112 @@ Scope* OperatorWithKernel::PrepareData(
...
+                                       DataLayout::kNHWC);
+          VLOG(7) << "Created reshaped dummy input based on MKL-DNN Tensor , "
+                     "but kNHWC layout"
+                  << var_name_item.first << " in Operator " << type_;
+        } else {
+          VLOG(7) << "Skip scanning input " << var_name_item.first
+                  << " in Operator " << type_;
+        }
+#endif
+      } else {
```
LeoZhao-BJ: can you avoid adding else branch to make change more clear to reviewer? e.g.
if (should_skip_input == true)
{
  xxxx
  continue;
}

if (!tensor_in->IsInitialized()) {
          continue;
        }

LeoZhao-BJ (commented):
all troubles are introduced by format remapping for "standard->mkldnn" and "mkldnn->standard", maybe we can think later if we have a better way to do inside of MKLDNN op instead of framework?

paddle/fluid/framework/operator.cc
```diff
@@ -1102,10 +1106,9 @@ Scope* OperatorWithKernel::PrepareData(
   }
 
   for (auto& var_name_item : Inputs()) {
+    bool should_skip_input = false;
     if (no_buffer_ins && no_buffer_ins->count(var_name_item.first) > 0) {
-      VLOG(7) << "Skip scanning input " << var_name_item.first
-              << " in Operator " << type_;
-      continue;
+      should_skip_input = true;
```
grygielski: Wouldn't be here more clear to just initialize should_skip_input with value of predicate inside if()?

grygielski (dismissed):
LGTM

paddle/fluid/framework/operator.cc
jczaja: Good catch! Will change that.

paddle/fluid/framework/operator.cc
jczaja: Ok, will do that

jczaja:
@LeoZhao-Intel , @grygielski I have updated PR with your suggestions and also added fix to newly introduced  MatchShapeToLayout(). Please continue your review

grygielski (dismissed):
LGTM

LeoZhao-BJ:
LGTM

sneaxiy (dismissed):
LGTM for revision of `no_need_buffer_ins`.

paddle/fluid/framework/operator.cc
```diff
@@ -1143,14 +1182,17 @@ Scope* OperatorWithKernel::PrepareData(
       // In the inference scenerio, the scopes will be reused across the
       // batches, so the `new_scope` here will result in GPU memroy explosion
       // over the  running of operators.
-      // We use a thread_local cache to fix that issue, the key in the cache is
+      // We use a thread_local cache to fix that issue, the key in the cache
+      // is
```
luotao1, line 1186: line1186, line1193, line1220 have some formatting problem, please fix in next PR.

luotao1 (dismissed):
LGTM

paddle/fluid/framework/operator.cc
jczaja: ok

grygielski (approved):
LGTM

luotao1:
Please revert https://github.com/PaddlePaddle/Paddle/pull/21747/commits/a48ae6949047b38a660f2ddd45b66611d30cfaea back in next PR.

luotao1 (approved):

PR #21749, opened 2019-12-14
disable op test of kldiv_loss

juncaipeng:
disable op test of kldiv_loss
https://github.com/PaddlePaddle/Paddle/issues/21750


luotao1 (approved):
LGTM

PR #21754, opened 2019-12-15
Update MKL-DNN to 1.1

grygielski:
Update MKL-DNN version to Tag 1.1.1


grygielski:
@luotao1 Here comes problem with libraries linked in other programs once again. Since v1.1 MKL-DNN is renamed to DNNL so library files have names changed. For Linux I just modified it as in 1.0 PR but now Windows is also a problem. Do I have to prepare similar tricks? It would probably require creating copy of 2 libraries (.lib and .dll) but I'm not really familiar with Windows builds. Or maybe Windows builds aren't used in production and I can just modify demo_ci files?

luotao1:
> Or maybe Windows builds aren't used in production and I can just modify demo_ci files

@zhouwei25 Could you help see it? i.e. Could the mkldnn library name change in windows inference tar.gz?
在windows预测库中，如果mkldnn.so的名字发生了变化，对应用户写的CMakeLists.txt就要发生变化。请问windows预测库已经投入实际使用了么？其中库的名字可以发生变化么？

luotao1:
> I can just modify demo_ci files

demo_ci supports both Linux and windows. But it only has Linux CI now. Do we have a windows CI on demo_ci? @zhouwei25

zhwesky2010:
> > Or maybe Windows builds aren't used in production and I can just modify demo_ci files
> 
> @zhouwei25 Could you help see it? i.e. Could the mkldnn library name change in windows inference tar.gz?
> 在windows预测库中，如果mkldnn.so的名字发生了变化，对应用户写的CMakeLists.txt就要发生变化。请问windows预测库已经投入实际使用了么？其中库的名字可以发生变化么？

windows预测库已经投入使用了，这个PR合入后，库的名字发生变化，用户的CMakeLists.txt都需要改成dnnl.lib（Windows）

zhwesky2010:
> > I can just modify demo_ci files
> 
> demo_ci supports both Linux and windows. But it only has Linux CI now. Do we have a windows CI on demo_ci? @zhouwei25

demo_ci里的run.sh，在Windows现在还没有跑，可以部署一个windows_inference。这个PR合入后，demo_ci的CMakelists.txt也需要改下，最好作者能同步修改Paddle/paddle/fluid/inference/api/demo_ci/CMakeLists.txt

luotao1:
@grygielski Discussed with @zhouwei25, Windows builds and its inference library are already used in production. If you change the name from `mkldnn.lib` to `dnnl.lib`, then all the users should change the inference CMakeLists.txt
 https://github.com/PaddlePaddle/Paddle/blob/c9882bda1c64fd3470ff2d5f1b53ac7507b45716/paddle/fluid/inference/api/demo_ci/CMakeLists.txt#L110-L113

@zhouwei25 Could you add a windows inference CI to ensure above problem?

grygielski:
@luotao1 I can try to prepare solution similar to Linux and just copy these libraries to have both mkldnn.lib and dnnl.lib. Is it ok? As I'm not that familiar with WIndows build so having some CI on which I can test my changes would be really helpful.

luotao1:
@grygielski Yes, it is OK. You can pull the PR at first, and @zhouwei25 will help to confirm the correctness of windows inference locally.

grygielski:
@luotao1 Great, I'll do it by tomorrow

zhwesky2010:
> @luotao1 I can try to prepare solution similar to Linux and just copy these libraries to have both mkldnn.lib and dnnl.lib. Is it ok? As I'm not that familiar with WIndows build so having some CI on which I can test my changes would be really helpful.

I will check whether your changes is helpful, and now there isn't  windows_inference_PR_CI, which will be established. You should copy duplicate of dnnl.dll and dnnl.lib, and set MKLDNN_SHARED_LIB=mkldnn.dll and MKLDNN_LIB=mkldnn.lib

grygielski:
@zhouwei25 I've added copying libraries for Windows build. However, I'm not an expert here so please check it and ask questions if any.

cmake/external/mkldnn.cmake
```diff
@@ -71,13 +71,16 @@ ExternalProject_Add(
...
                         -DCMAKE_C_FLAGS=${MKLDNN_CFLAG}
                         -DCMAKE_CXX_FLAGS=${MKLDNN_CXXFLAG}
-                        -DMKLDNN_BUILD_TESTS=OFF -DMKLDNN_BUILD_EXAMPLES=OFF
+                        -DDNNL_BUILD_TESTS=OFF -DDNNL_BUILD_EXAMPLES=OFF
     CMAKE_CACHE_ARGS    -DCMAKE_INSTALL_PREFIX:PATH=${MKLDNN_INSTALL_DIR}
 )
 if(WIN32)
-    SET(MKLDNN_LIB "${MKLDNN_INSTALL_DIR}/${LIBDIR}/mkldnn.lib" CACHE FILEPATH "mkldnn library." FORCE)
+    SET(MKLDNN_LIB "${MKLDNN_INSTALL_DIR}/${LIBDIR}/dnnl.lib" CACHE FILEPATH "mkldnn library." FORCE)
+    SET(MKLDNN_LIB_OLD "${MKLDNN_INSTALL_DIR}/${LIBDIR}/mkldnn.lib" CACHE FILEPATH "mkldnn library." FORCE)
```
zhwesky2010: dnnl.lib is the import library of dnnl.dll, copy dnnl.lib to mkldnn.lib, mkldnn.lib will still import dnnl.dll, so mkldnn.dll will not work. So can you generate mkldnn.dll directly? or, use custom command in cmake to generate mkldnn.lib from mkldnn.dll,rather than copy mkldnn.lib from dnnl.lib.

zhwesky2010:
> @zhouwei25 I've added copying libraries for Windows build. However, I'm not an expert here so please check it and ask questions if any.

copy method in windows will not work, can you Generate the dll library in https://github.com/intel/mkl-dnn.git directly?

cmake/external/mkldnn.cmake
```diff
@@ -19,7 +19,7 @@ SET(MKLDNN_PREFIX_DIR    ${THIRD_PARTY_PATH}/mkldnn)
 SET(MKLDNN_INSTALL_DIR    ${THIRD_PARTY_PATH}/install/mkldnn)
 SET(MKLDNN_INC_DIR        "${MKLDNN_INSTALL_DIR}/include" CACHE PATH "mkldnn include directory." FORCE)
 SET(MKLDNN_REPOSITORY     https://github.com/intel/mkl-dnn.git)
```
zhwesky2010, line 21: can you have two target in MKLDNN repository, one can generate mkldnn.lib&mkldnn.dll, another can generate dnnl.lib&dnnl.dll. Then windows will use mkldnn.lib&mkldnn.dll

cmake/external/mkldnn.cmake
grygielski: I don't understand why mkldnn.lib importing dnnl.dll is a problem if we will have all 4 libraries in a folder?

cmake/external/mkldnn.cmake
grygielski: I don't choose library name inside our CMake, it's done in mkl-dnn build files so it's not possible.

grygielski:
> > @zhouwei25 I've added copying libraries for Windows build. However, I'm not an expert here so please check it and ask questions if any.
> 
> copy method in windows will not work, can you Generate the dll library in https://github.com/intel/mkl-dnn.git directly?

Is this mkldnn.lib -> dnnl.dll link an only problem or there are other problems that makes `copy method` not working for windows?

zhwesky2010:
> > > @zhouwei25 I've added copying libraries for Windows build. However, I'm not an expert here so please check it and ask questions if any.
> > 
> > 
> > copy method in windows will not work, can you Generate the dll library in https://github.com/intel/mkl-dnn.git directly?
> 
> Is this mkldnn.lib -> dnnl.dll link an only problem or there are other problems that makes `copy method` not working for windows?

Yes, thers is only mkldnn.lib -> dnnl.dll problem, if mkldnn.lib -> mkldnn.dll, there will not problem.

grygielski:
@zhouwei25 but when we have all 4 libraries: mkldnn.lib, dnnl.lib, mkldnn.dll and dnnl.dll why linking mkldnn.lib -> dnnl.dll is a problem?

zhouwei25:
> @zhouwei25 but when we have all 4 libraries: mkldnn.lib, dnnl.lib, mkldnn.dll and dnnl.dll why linking mkldnn.lib -> dnnl.dll is a problem?

Users use mkldnn.lib&mkldnn.dll on windows, libmkldnn.so.0 on Linux. So, they only copy mkldnn.dll to their Release directory, windows will find dll in Release directory defaultly. There will error: DLL not found, because the real useful dnn.dll have not been copied to Release directory. On linux, it doesn't need to copy libmkldnn.so.0, just link it and it will work.

zhouwei25:
> @zhouwei25 but when we have all 4 libraries: mkldnn.lib, dnnl.lib, mkldnn.dll and dnnl.dll why linking mkldnn.lib -> dnnl.dll is a problem?

Users use mkldnn.lib&mkldnn.dll on windows, libmkldnn.so.0 on Linux. So, they only copy mkldnn.dll to their Release directory, windows will find dll in Release directory defaultly. There will error: DLL not found, because the real useful dnn.dll have not been copied to Release directory. On linux, it doesn't need to copy libmkldnn.so.0, just link it and it will work.

zhwesky2010:
@grygielski  Users use mkldnn.lib&mkldnn.dll on windows, libmkldnn.so.0 on Linux. So, they only copy mkldnn.dll to their Release directory, windows will find dll in Release directory defaultly. There will error: DLL not found, because the real useful dnn.dll have not been copied to Release directory. On linux, it doesn't need to copy libmkldnn.so.0, just link it and it will work.

zhwesky2010:
@grygielski  This problem can be solved by:
1. Users fix their CMakeLists.txt of their project to copy dnnl.dll to release directory.
2. Use custom-command in mkldnn.cmake to generate mkldnn.lib from mkldnn.dll rather than copy dnnl.lib ->mkldnn.lib, this will trouble and I have not tried.

grygielski:
@zhouwei25 I'm afraid it's not possible to just simply generate .dll file from .lib on our side since .lib does not contain any code or something, just links to proper places in .dll file. I can see the issue but don't know better way of fixing it. Do you think 1st solution is achievable?

zhwesky2010:
@grygielski https://www.ojit.com/article/551175 This web site may help how to generate `lib` from `dll` on windows, it need `.def` to `lib /def:mkldnn.def /out:mkldnn.lib /machine:x86`, in this way, mkldnn.lib can point to the mkldnn.dll.

[PR_Windows_CI_Inference](http://ci.paddlepaddle.org/viewType.html?buildTypeId=PaddleWindows_PrWindowsCiInference_2&tab=pendingChangesDiv) can help to check it. Just pull requesets and run this PR manually.

luotao1:
@grygielski You can update MKL-DNN to 1.1 for Linux at first.

grygielski:
@zhouwei25 @luotao1 
Done, I've tested it with CI_Windows_Inference (http://ci.paddlepaddle.org/viewLog.html?tab=buildLog&buildTypeId=PaddleWindows_PrWindowsCiInference_2&buildId=261259) and it passed.
Now I'm starting working on generating .lib. I also asked MKL-DNN team for some suggestions on that problem.

luotao1 (approved):

luotao1 (approved):

PR #21766, opened 2019-12-16
Add some debug flags to auto growth allocator

sneaxiy:
Add some debug flags for `AutoGrowthBestFitAllocator`:

- `FLAGS_free_idle_chunk`: whether to call `cudaFree` when each allocation is destructed. This is used for measuring the actual GPU memory consumption of models (without any caching memory). 
- `FLAGS_free_when_no_cache_hit`: whether to call `cudaFree` when no memory cache is hit. If true, `cudaFree` would be called before calling `cudaMalloc`; if else, `cudaFree` would be called only when out of memory occurs, which is the same as the allocator strategy of PyTorch. This flag is used to debug the advantages of two free strategies.

All of these flags are only used to debug for developer, and it is not public to users.


paddle/fluid/memory/allocation/auto_growth_best_fit_allocator.cc
```diff
@@ -21,6 +21,16 @@
 #include <unordered_map>
 #include "paddle/fluid/memory/allocation/aligned_allocator.h"
 
+DEFINE_bool(free_idle_chunk, false,
+            "Whether to free idle chunk when each allocation is freed. "
+            "If false, all freed allocation would be cached to speed up next "
+            "allocation request. If true, no allocation would be cached.");
```
lanxianghit: 这两个flag应该是针对auto growth策略的吧？但是注释说明里没有提到与auto growth的关系，下同

sneaxiy: Done.

lanxianghit (approved):

PR #21777, opened 2019-12-17
Optimize adam speed

Aurelius84:
Before, `_finish_update()` is called to scale `beta1` and `beta2` every iteration when applying `adam.minimize()`. 

Two `scale_op` kernel  are launched here that may slow down the training. Because the launched process may exhaust more time than calculation indeed.

In this PR, we remove the two `scale_op` in `_finish_update()` and fuse it into `adam_op` to decrease the numbers of launched kernel.

> Note: regularizer is removed while testing.

#### 1. Dynamic Graph

The cost time of `minimizer` in `dyraph_mode` by `Dialogue-PLATO` model with single P40 GPU.

image

| minimize时间(s) | 优化前    | 优化后   |  提升 | 
|:-----------:|:------:|:------:|:------:|
| 1           | 0.070 | 0.038 | **45.7%** |

**we  trained 100 batches iteratively and sum up the total time. See details as follows:**
`total time = forward + optimize`

| 总时间(s) | 优化前    | 优化后   |  提升 | 
|:-----------:|:------:|:------:|:------:|
| 1           | 10.877 | 9.825 | **9.67%** |

**Conclusion:** This PR is with higher performance improvement mainly in dynamic graph mode.

#### 2. Static Graph
Also, we compared the running time on the static graph mode by `resNet50` model with 4 P40 GPU. 
> Note: Not all train dataset of ImageNet-2012 was used in above experiment，actually only 256000 pictures was used.

| 总时间(s/epoch) | 优化前    | 优化后   |  提升 |
|:-----------:|:------:|:------:|:------:|
| 1           | 850.96 | 848.94 ||
| 2           | 852.79 | 848.41 ||
| 3           | 853.53 | 850.20 ||
| 4           | 852.96 | 851.89 ||
| 5           | 852.92 | 850.98 ||
|平均值   | **852.63**|  **850.08** |0.3%|

We also compare the performance while turning on the `fuse_all_optimizer_ops = True` by `resNet50` model with 4 P40 GPU. 
> Note: Here only 25600 pictures was used to speed up the testing process.

| 总时间(s/epoch) | 优化前    | 优化后   |  提升 |
|:-----------:|:------:|:------:|:------:|
| 1           | 86.70 | 86.30 ||
| 2           | 86.60 | 86.46 ||
| 3           | 86.18 | 85.85 ||
| 4           | 86.44 | 85.97 ||
| 5           | 86.57 | 85.49 ||
|平均值   | **86.498**|  **86.014** |0.56%|

**Conclusion:** This PR will not affect the performance of static graphs too much.

### DIFF
--- a/paddle/fluid/framework/ir/fuse_optimizer_ops_pass/fuse_adam_op_pass.cc
+++ b/paddle/fluid/framework/ir/fuse_optimizer_ops_pass/fuse_adam_op_pass.cc
@@ -39,6 +39,7 @@ class FuseAdamOpPass : public FuseOptimizerOpPass {
       const std::vector<ir::Node *> &adam_ops, ir::Graph *graph) const {
     auto fused_adam_node =
         FuseAdamOps(aux_var_set, fused_vars_name, adam_ops, graph);
+    /*
     auto fused_scale1 =
         FuseScaleOps(aux_var_set.at("Beta1Pow"), fused_vars_name.at("Beta1Pow"),
                      adam_ops, graph);
@@ -46,6 +47,7 @@ class FuseAdamOpPass : public FuseOptimizerOpPass {
         FuseScaleOps(aux_var_set.at("Beta2Pow"), fused_vars_name.at("Beta2Pow"),
                      adam_ops, graph);
     RemoveCycleDepsBetweenOpNodes(graph, fused_scale1, fused_scale2);
+     */
     return fused_adam_node;
   }
 
@@ -139,6 +141,8 @@ class FuseAdamOpPass : public FuseOptimizerOpPass {
     adam_desc.SetOutput("ParamOut", {fused_vars_name.at(kParam)});
     adam_desc.SetOutput("Moment1Out", {fused_vars_name.at("Moment1")});
     adam_desc.SetOutput("Moment2Out", {fused_vars_name.at("Moment2")});
+    adam_desc.SetOutput("Beta1PowOut", {fused_vars_name.at("Beta1Pow")});
+    adam_desc.SetOutput("Beta2PowOut", {fused_vars_name.at("Beta2Pow")});
     adam_desc.SetAttr("beta1", beta1);
     adam_desc.SetAttr("beta2", beta2);
     adam_desc.SetAttr("epsilon", epsilon);
--- a/paddle/fluid/framework/ir/fuse_optimizer_ops_pass/fuse_adam_op_pass.cc
+++ b/paddle/fluid/framework/ir/fuse_optimizer_ops_pass/fuse_adam_op_pass.cc
@@ -39,15 +39,6 @@ class FuseAdamOpPass : public FuseOptimizerOpPass {
       const std::vector<ir::Node *> &adam_ops, ir::Graph *graph) const {
     auto fused_adam_node =
         FuseAdamOps(aux_var_set, fused_vars_name, adam_ops, graph);
-    /*
-    auto fused_scale1 =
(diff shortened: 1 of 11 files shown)


paddle/fluid/framework/ir/fuse_optimizer_ops_pass/fuse_adam_op_pass.cc
```diff
@@ -39,13 +39,15 @@ class FuseAdamOpPass : public FuseOptimizerOpPass {
       const std::vector<ir::Node *> &adam_ops, ir::Graph *graph) const {
     auto fused_adam_node =
         FuseAdamOps(aux_var_set, fused_vars_name, adam_ops, graph);
+    /*
```
phlrain: remove useless code

python/paddle/fluid/optimizer.py
```diff
@@ -1697,6 +1699,8 @@ def _append_optimize_op(self, block, param_and_grad):
     def _finish_update(self, block, param_and_grads):
         """Update Beta1 and Beta2 Power accumulators
         """
+        pass
+        '''
```
phlrain: remove useless code

paddle/fluid/framework/ir/fuse_optimizer_ops_pass/fuse_adam_op_pass.cc
Aurelius84: Thanks. Fixed.

python/paddle/fluid/optimizer.py
Aurelius84: Thanks. Fixed.

paddle/fluid/operators/optimizers/adam_op.h
```diff
@@ -51,9 +51,47 @@ class AdamOp : public framework::OperatorWithKernel {
...
 struct CPUAdam;
 
+template <typename T>
+struct BetaPowFunctor;
+
 template <typename T, typename Flavour>
 struct AdamFunctor;
 
+template <typename T>
+struct BetaPowFunctor {
```
zhhsplendid: Use "class" instead of "struct" for BetaPowFunctor, same for AdamFunctor

Use a struct only for passive objects that carry data; everything else is a class. You are implementing Functor here.

Reference:
https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes

paddle/fluid/operators/optimizers/adam_op.h
```diff
@@ -51,9 +51,47 @@ class AdamOp : public framework::OperatorWithKernel {
 struct GPUAdam;
 struct CPUAdam;
 
+template <typename T>
+struct BetaPowFunctor;
```
zhhsplendid: Is declaration necessary here?

zhhsplendid (approved):
LGTM

zhhsplendid (changes requested):
Undo LGTM. I found some backward backward compatibility issues and discussed with Liujie offline. We will try to make a decision on it.

paddle/fluid/operators/optimizers/adam_op.h
Aurelius84: Thanks, fixed.

paddle/fluid/operators/optimizers/adam_op.h
Aurelius84: Thanks, fixed.

zhhsplendid (approved):
We have done some offline experiment to make sure save/load and re-train doesn't have any problem. LGTM.

phlrain (approved):
LGTM

PR #21780, opened 2019-12-17
Fix lr assert

sneaxiy:
The former implementation assert a tuple with 2 items because there is a `()`, which always returns True.
```
assert (end_lr > start_lr, 
       "end_lr {} MUST GREATER than start_lr {}".format(end_lr, start_lr))
```

This PR fixes this bug by removing `()`.

### DIFF
--- a/python/paddle/fluid/dygraph/learning_rate_scheduler.py
+++ b/python/paddle/fluid/dygraph/learning_rate_scheduler.py
@@ -626,9 +626,8 @@ class LinearLrWarmup(LearningRateDecay):
                 format(learning_rate))
         self.learning_rate = learning_rate
         self.warmup_steps = warmup_steps
-        assert (end_lr > start_lr,
-                "end_lr {} MUST GREATER than start_lr {}".format(end_lr,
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/dygraph/learning_rate_scheduler.py
```diff
@@ -626,9 +626,8 @@ def __init__(self,
                 format(learning_rate))
         self.learning_rate = learning_rate
         self.warmup_steps = warmup_steps
-        assert (end_lr > start_lr,
-                "end_lr {} MUST GREATER than start_lr {}".format(end_lr,
-                                                                 start_lr))
+        assert end_lr > start_lr, "end_lr {} MUST GREATER than start_lr {}".format(
```
zhhsplendid: Tiny grammar issue: "MUST be GREATER than"

sneaxiy: Done.

zhhsplendid (approved):
LGTM

PR #21783, opened 2019-12-17
Update test precision from fp32 to fp64, test=develop

juncaipeng:
Update test precision from fp32 to fp64, test=develop


luotao1 (approved):

luotao1 (commented):
@juncaipeng Should you update the `/python/paddle/fluid/tests/unittests/white_list/op_accuracy_white_list.py`?

PR #21787, opened 2019-12-17
Add unary operator __neg__

zhiqiu:
As the title, this PR adds unary operator `__neg__` for VarBase in dygraph mode, and for Variable in static graph mode. 

Example,
```
import paddle.fluid as fluid
import numpy as np

x_np = np.ones([1, 2], np.float32)
with fluid.dygraph.guard():
    x = fluid.dygraph.to_variable(x_np)
    print(x.numpy())  # [[1, 1]]
    print(-x.numpy())  # [[-1, -1]]
```


sneaxiy (approved):
LGTM.

PR #21798, opened 2019-12-17
add conditional grad op when optimizer used in control flow.

liym27:
small pr from https://github.com/PaddlePaddle/Paddle/pull/21676

### DIFF
--- a/python/paddle/fluid/layers/control_flow.py
+++ b/python/paddle/fluid/layers/control_flow.py
@@ -1887,29 +1887,28 @@ class ConditionalBlock(object):
         new_op_desc.copy_from(grad_op_desc[0])
         new_op_desc._set_attr(op_role_attr_name, backward)
 
+        # set input and output manually
+        new_op_desc.set_input('Input', param_list)
+        new_op_desc.set_output('Input@GRAD',
+                               [param + "@GRAD" for param in param_list])
         new_vars = set()
         for grad_var_name in new_op_desc.output_arg_names():
-            if grad_sub_block.desc.has_var_recursive(
+            if parent_block.desc.has_var_recursive(
                     cpt.to_bytes(grad_var_name)
             ) or grad_var_name == core.empty_var_name():
                 continue
-            grad_sub_block.desc.var(cpt.to_bytes(grad_var_name))
+            parent_block.desc.var(cpt.to_bytes(grad_var_name))
             new_vars.add(grad_var_name)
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/tests/unittests/test_case.py
```diff
@@ -223,5 +224,52 @@ def type_error_default():
...
+        EPOCH_NUM = 2
+
+        x = fluid.data(
+            name='x', shape=[BATCH_SIZE, INPUT_SIZE], dtype='float32')
+        y = fluid.data(
+            name='y', shape=[BATCH_SIZE, INPUT_SIZE], dtype='float32')
+
+        switch_id = fluid.data(name='switch_id', shape=[1], dtype='int32')
+
+        one = layers.fill_constant(shape=[1], dtype='int32', value=0)
```
zhhsplendid: Variable name is one but value is 0

python/paddle/fluid/layers/control_flow.py
```diff
@@ -1823,6 +1825,75 @@ def complete(self):
...
+        for inner_out_name in params:
+            inner_var = parent_block._find_var_recursive(inner_out_name)
+            if inner_var:
+                param_list.append(cpt.to_text(inner_var.name))
+
+        grad_op_desc, op_grad_to_var = core.get_grad_op_desc(
+            conditional_block_op.desc,
+            cpt.to_text(set()), [grad_sub_block.desc])
+
+        # append op_desc in grad_op_descs to target_block
```
zhhsplendid, line 1883: To my understanding, here you are adding conditional_block_grad_op to parent_block and we have to do it manually because that optimizer.minimize/append_backward is called before we have a conditional_block_op.

So this English comment doesn't fully explain it, could you refine the English comment here?

python/paddle/fluid/tests/unittests/test_case.py
zhhsplendid: Variable name is "one" but the value is 0. Please rename the variable or set other value

liym27: Thanks. Done

python/paddle/fluid/layers/control_flow.py
liym27: Thanks.  Refine the comment.
image

zhhsplendid (approved):
LGTM

PR #21806, opened 2019-12-18
add package for centerloss

ForFishes:
Because the `center loss`  moves to `loss.py`, the `core` and `Constant` are discarded. Thus, we should add them to `loss.py` to avoid the bug.


gavin1332 (approved):
LGTM

PR #21809, opened 2019-12-18
Fix Backward Bugs in Conditional Block

zhhsplendid:
The fixed bugs:

1. The condition sub-graph is not pruned
When a condition block has operations, such as `out = layers.cond(a - b < -1.0, lambda: a, lambda: b)`, here `a - b < -1.0` will contain an "elementwise_sub_grad" op at backward stage but "less than" doesn't return any gradient since it is using an EmptyGradOpMaker, which can crash the program. We fixed it by pruning the backward graph:
image

2. When backward graph is extremely simple, the whole backward ops in a block are pruned. The very simple network is like `out = layers.cond(a < b, lambda: a, lambda: b)` or `y = x + 1`. We fixed it by:
image

3. Also fix some tiny bugs, typos, and ugly codes. Such as English grammar "some op" -> "some ops", "return boolean_variable ? true : false;" -> "return boolean_variable", add C++ template classes when determine whether an op is using default grad op desc maker, _find_op_path_ should accept vars as inputs, not var_names. etc.

### DIFF
--- a/paddle/fluid/framework/details/op_registry.h
+++ b/paddle/fluid/framework/details/op_registry.h
@@ -221,7 +221,15 @@ struct OpInfoFiller<T, kGradOpDescMaker> {
 
     info->use_default_grad_op_desc_maker_ =
         std::is_base_of<DefaultGradOpMaker<OpDesc, true>, T>::value ||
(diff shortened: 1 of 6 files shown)


sneaxiy (approved):
LGTM.

PR #21825, opened 2019-12-18
Copy stop_gradient property of variable in clone

zhiqiu:
As the title, original `clone` method will lose variable's `stop_gradient` property since `stop_gradient` is not in Variable's `VarDesc`, this PR fixes that.


sneaxiy (commented):
Cool.

sneaxiy (approved):
Cool.

PR #21826, opened 2019-12-18
Disable memory opt pass when DNNL is on

Sand3r-:
Temoparily disables memory optimization pass when DNNL is on. This is a temporary fix for until the memory optim pass will be fixed.


paddle/fluid/inference/api/analysis_config.cc
```diff
@@ -325,7 +325,7 @@ void AnalysisConfig::Update() {
 
 #ifdef PADDLE_WITH_MKLDNN
   // Do not optimize before quantization
```
wojtuss: Please, update the comment as well.

luotao1 (approved):
LGTM

PR #21846, opened 2019-12-19
fix stat shape

hutuxian:
Restore stat shape in global scenario for compatible reason: from [num_bucket] to [1, num_bucket].

### DIFF
--- a/paddle/fluid/operators/metrics/auc_op.cc
+++ b/paddle/fluid/operators/metrics/auc_op.cc
@@ -49,12 +49,15 @@ class AucOp : public framework::OperatorWithKernel {
 
     ctx->SetOutputDim("AUC", {1});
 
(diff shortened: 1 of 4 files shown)


paddle/fluid/operators/metrics/auc_op.cc
```diff
@@ -49,12 +49,15 @@ class AucOp : public framework::OperatorWithKernel {
 
     ctx->SetOutputDim("AUC", {1});
 
-    // slide_steps = slide_steps == 0 ? 1 : slide_steps;
```
gavin1332: what‘s this comment use for?

gavin1332 (commented):
LGTM, require other related partner checking.

yaoxuefeng6 (approved):
LGTM

PR #21851, opened 2019-12-19
control flow: support optimizer called

liym27:
## Support that `optimizer` is called and loss is created in control flow `switch`.

Example 1:  `loss` and `optimizer` are both in control flow
```
def fn_1():
    loss = layers.mean(cross_entropy_loss)
    adam.minimize(loss)

def fn_2():
    loss = layers.mean(softmax_loss)
    adagrad.minimize(loss)

layers.case([(switch_id==id, fn_2)], fn_1)
```
Example 2:   `optimizer`is in control flow, but `loss` is in parent block
```
loss = layers.mean(cross_entropy_loss)
def fn_1():
    adam.minimize(loss)

def fn_2():
    adagrad.minimize(loss)

layers.case([(switch_id==id, fn_2)], fn_1)
```
### (1) backward network
**Before:**   if loss is in control flow,  the grad ops corresponding to foreward ops in parent block are not appended.
**This PR:**  append all needed grad ops and build a complete backward network.

### (2) optimize ops
**Before:**   if `optimizer` is in control flow,  the optimize ops appended in **global block**
**This PR:**  The optimize ops appended in **sub grad block**
small PR https://github.com/PaddlePaddle/Paddle/pull/21791

### (3) conditional_block_grad
Grad op `conditional_block_grad` can't be appended when calling `optimizer.minimize()`,  manually append `conditional_block_grad` in control flow.
small PR https://github.com/PaddlePaddle/Paddle/pull/21798
...

---

image

### DIFF
--- a/python/paddle/fluid/backward.py
+++ b/python/paddle/fluid/backward.py
@@ -1182,6 +1182,7 @@ def append_backward(loss,
             if program._appending_grad_times > 1:
                 input_grad_names_set = set([_append_grad_suffix_(loss.name)])
 
+        # Todo: support _append_backward_ops_with_checkpoints_ in control flow
         if checkpoints != None and \
                 isinstance(checkpoints, list) and \
                 len(checkpoints) > 0:
@@ -1197,7 +1198,7 @@ def append_backward(loss,
                     checkpoints)
         else:
             _append_backward_ops_(
-                block,  # the block where forward ops are#
+                block,  # the block where forward ops are in
                 op_path_dict[block_idx],
                 target_grad_block,
                 no_grad_dict,
@@ -1205,14 +1206,15 @@ def append_backward(loss,
                 callbacks,
                 input_grad_names_set=input_grad_names_set)
 
-        # Because calc_gradient may be called multiple times,
-        # we need rename the internal gradient variables so that they have
-        # different names.
-        _rename_grad_(block, block_fwd_op_num_dict[block_idx], grad_to_var, {})
-
     grad_info_map = dict()
     fwd_op_num = block_fwd_op_num_dict[
         current_block_idx] if not is_in_control_flow else 0
+
+    # Because append_backward may be called multiple times,
+    # we need rename the internal gradient variables so that they have
+    # different names.
+    _rename_grad_(target_grad_block, fwd_op_num, grad_to_var, {})
+
     _append_backward_vars_(target_grad_block, fwd_op_num, grad_to_var,
                            grad_info_map)
 
--- a/python/paddle/fluid/backward.py
+++ b/python/paddle/fluid/backward.py
@@ -958,7 +958,7 @@ def _append_backward_vars_(block, start_op_idx, grad_to_var, grad_info_map):
         for grad_var_name in op_desc.output_arg_names():
             if block.desc.has_var_recursive(cpt.to_bytes(
                     grad_var_name)) or grad_var_name == core.empty_var_name():
-                print("1. grad_var_name : ", grad_var_name)
+                # print("1. grad_var_name : ", grad_var_name)
                 continue
             block.desc.var(cpt.to_bytes(grad_var_name))
             new_vars.add(grad_var_name)
@@ -1343,7 +1343,7 @@ def _find_op_path_(block, outputs, inputs, no_grad_set):
     no_grad_set will also be changed
     """
     input_names = set([inp.name for inp in inputs])
-    output_names = set([out.name for out in outputs])
+    output_names = _get_output_names(block, outputs)
 
     relevant_op_flags = [True] * len(block.ops)
 
--- a/python/paddle/fluid/backward.py
+++ b/python/paddle/fluid/backward.py
@@ -958,6 +958,7 @@ def _append_backward_vars_(block, start_op_idx, grad_to_var, grad_info_map):
         for grad_var_name in op_desc.output_arg_names():
             if block.desc.has_var_recursive(cpt.to_bytes(
                     grad_var_name)) or grad_var_name == core.empty_var_name():
+                print("1. grad_var_name : ", grad_var_name)
                 continue
             block.desc.var(cpt.to_bytes(grad_var_name))
             new_vars.add(grad_var_name)
--- a/python/paddle/fluid/backward.py
+++ b/python/paddle/fluid/backward.py
@@ -958,7 +958,6 @@ def _append_backward_vars_(block, start_op_idx, grad_to_var, grad_info_map):
         for grad_var_name in op_desc.output_arg_names():
             if block.desc.has_var_recursive(cpt.to_bytes(
                     grad_var_name)) or grad_var_name == core.empty_var_name():
-                # print("1. grad_var_name : ", grad_var_name)
                 continue
             block.desc.var(cpt.to_bytes(grad_var_name))
             new_vars.add(grad_var_name)
@@ -1008,6 +1007,17 @@ def _get_stop_gradients_(program):
                 block_no_grad_set.add(_append_grad_suffix_(var.name))
         no_grad_dict[block.idx] = block_no_grad_set
     return no_grad_dict
+
+
+def _get_son_parent_block_idx_dict(program, current_block_idx):
+
+    son_parent_block_idx_dict = collections.OrderedDict()
+    while current_block_idx >= 0:
+        parent_block_idx = program.block(current_block_idx).parent_idx
+        son_parent_block_idx_dict[current_block_idx] = parent_block_idx
+        current_block_idx = parent_block_idx
+
+    return son_parent_block_idx_dict
 
 
 def append_backward(loss,
@@ -1102,27 +1112,31 @@ def append_backward(loss,
         isinstance(callbacks, list)
 
     program = loss.block.program
+    root_block = program.block(0)
+    current_block_idx = program.current_block_idx
+    current_block = program.block(current_block_idx)
 
-    # Todo(liym27): when to add 1 in control flow
-    program._appending_grad_times += 1
+    is_in_control_flow = current_block_idx != 0
+
+    # Double grad is not supported in sub-block (control flow)
+    if not is_in_control_flow:
+        # _appending_grad_times used for double grad
+        program._appending_grad_times += 1
 
     if no_grad_set is None:
         no_grad_set = set()
     no_grad_set = copy.copy(no_grad_set)
     no_grad_dict = _get_stop_gradients_(program)
-    # no_grad_set only contain vars in block 0
+    # no_grad_set only contains vars in block 0
     # Todo(liym27): support vars in sub block
     no_grad_dict[0].update(list(map(_append_grad_suffix_, no_grad_set)))
 
-    root_block = program.block(0)
-
-    current_block_idx = program.current_block_idx
-
-    current_block = program.block(current_block_idx)
-
-    # create grad block if in control flow
-    is_in_control_flow = current_block_idx != 0
+    # Currently it is only to support the optimizer.minimize
+    # in a switch branch, which can append_backward in a sub_block.
+    # Note: while_loop is in control flow, but it makes no sense to call optimizer in while.
+    # Todo: report error when it is in while_loop
     if is_in_control_flow:
+        # create grad block if in switch control flow.
         target_grad_block = program._create_block(
             parent_idx=current_block.parent_idx)
         target_grad_block._set_forward_block_idx(current_block_idx)
@@ -1130,19 +1144,10 @@ def append_backward(loss,
     else:
         target_grad_block = root_block
 
-    def _get_son_parent_block_idx_dict(program, current_block_idx):
-        son_parent_block_idx_dict = collections.OrderedDict()
-        while current_block_idx >= 0:
-            parent_block_idx = program.block(current_block_idx).parent_idx
-            son_parent_block_idx_dict[current_block_idx] = parent_block_idx
-            current_block_idx = parent_block_idx
-        return son_parent_block_idx_dict
-
     son_parent_block_idx_dict = _get_son_parent_block_idx_dict(
         program, current_block_idx)
 
-    # block_id: fwd_op_num
-    block_fwd_op_num_dict = {}
+    block_fwd_op_num_dict = {}  # block_id: fwd_op_num
     for idx in son_parent_block_idx_dict:
         block_fwd_op_num_dict[idx] = program.block(idx).desc.op_size()
 
@@ -1150,9 +1155,6 @@ def append_backward(loss,
 
     op_desc = _create_loss_op_desc_(loss)
     target_grad_block.desc.append_op().copy_from(op_desc)
-
-    current_block_no_grad_set = set(
-        map(_strip_grad_suffix_, no_grad_dict[current_block_idx]))
 
     op_path_dict = {}  # block idx : op_path
     no_grad_vars_dict = {}
@@ -1173,16 +1175,17 @@ def append_backward(loss,
             list(map(_append_grad_suffix_, block_no_grad_set)))
 
         input_grad_names_set = None
-        # For double backward, input_grad_names is used for filter
-        # some non-used gradients op.
+        # For double backward, input_grad_names is used for filtering
+        # some non-used gradients op(s).
 
         # Todo(liym27): need a better design.
-        # not support double grad in control flow now
+        # not support double grad in control flow sub-block now.
         if not is_in_control_flow:
             if program._appending_grad_times > 1:
                 input_grad_names_set = set([_append_grad_suffix_(loss.name)])
 
-        # Todo: support _append_backward_ops_with_checkpoints_ in control flow
+        # Todo: support _append_backward_ops_with_checkpoints_ in
+        #  sub-block (control flow)
         if checkpoints != None and \
                 isinstance(checkpoints, list) and \
                 len(checkpoints) > 0:
@@ -1230,6 +1233,7 @@ def append_backward(loss,
     params_and_grads = []
     op_role_var_attr_name = core.op_proto_and_checker_maker.kOpRoleVarAttrName()
     if is_in_control_flow:
+        # control flow: grad_var not in current block but in corresponding grad block
         for param in parameters:
             if cpt.to_text(param) not in grad_info_map:
                 continue
@@ -1254,7 +1258,6 @@ def append_backward(loss,
                 attr_val.extend(grad_var.op.attr(op_role_var_attr_name))
             grad_var.op._set_attr(op_role_var_attr_name, attr_val)
 
-            # control flow: grad_var not in current block but in corresponding grad block
             params_and_grads.append((param_var, grad_var))
 
         return params_and_grads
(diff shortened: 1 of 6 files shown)


python/paddle/fluid/backward.py
```diff
@@ -958,6 +958,7 @@ def _append_backward_vars_(block, start_op_idx, grad_to_var, grad_info_map):
         for grad_var_name in op_desc.output_arg_names():
             if block.desc.has_var_recursive(cpt.to_bytes(
                     grad_var_name)) or grad_var_name == core.empty_var_name():
+                # print("1. grad_var_name : ", grad_var_name)
```
zhhsplendid: If it is useless debug print, remove it when you make a pull request.

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def fn_2():
+        softmax_loss = layers.softmax_with_cross_entropy(
+            logits=prediction, label=label)
+        mean_softmax_loss = layers.mean(softmax_loss, name='mean_softmax_loss')
+        adagrad.minimize(mean_softmax_loss)
+        return mean_softmax_loss
+
+    avg_loss = layers.case([(switch_id == id, fn_2)], fn_1)
+    main_program = fluid.default_main_program()
```
zhhsplendid: Try to use "fluid.Program" and "with fluid.program_guard" instead of fluid.default_main/startup_program(), then the different programs in different test cases won't interact with each other and it is easier to add tests for future developers.

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
```diff
@@ -0,0 +1,195 @@
...
+                      fetch_list=[
+                          hidden,
+                          prediction,
+                          avg_loss,
+                      ])
+        out_hidden, out_prediction, loss = out
+    return out_prediction, loss
+
+
+class MyLayer(fluid.dygraph.Layer):
```
zhhsplendid: MyLayer is not good name. Maybe we can call it "DygraphLayer" ?

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
```diff
@@ -0,0 +1,195 @@
...
+
+    image = fluid.data(
+        name='image', shape=[BATCH_SIZE, INPUT_SIZE], dtype='float32')
+    label = fluid.data(name='label', shape=[BATCH_SIZE, 1], dtype='int64')
+    switch_id = fluid.data(name='switch_id', shape=[1], dtype='int32')
+
+    id = layers.fill_constant(shape=[1], dtype='int32', value=Switch_ID)
+    hidden, prediction = simple_fc_net(image)
+
+    adam = optimizer.Adam(learning_rate=0.001)
```
zhhsplendid: Nothing to change, but just a question, can we put the "adam = optimizer.Adam(learning_rate=0.001)" in a branch such as "fn_1" ? Maybe we can add a test case to cover that.

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
```diff
@@ -0,0 +1,195 @@
...
+import paddle.fluid.layers as layers
+import paddle.fluid.optimizer as optimizer
+
+BATCH_SIZE = 1
+INPUT_SIZE = 784
+
+CLASS_NUM = 10
+FC_SIZE = 40
+EPOCH_NUM = 20
+Switch_ID = 3
```
zhhsplendid: Switch_ID -> SWITCH_ID

python/paddle/fluid/layers/control_flow.py
```diff
@@ -1798,6 +1800,9 @@ def complete(self):
                     intermediate.add(out_var_name)
         input_set = set([ipt.name for ipt in self.inputs])
 
+        # Todo(liym27) Here assume that all params are in recursive parent block
+        # but when minimize() called in control flow, some params may be in
```
zhhsplendid, line 1805: Question but nothing to change: in my understanding, params here are just forward input vars? So I think it won't be at conditional grad block?

python/paddle/fluid/backward.py
```diff
@@ -1215,12 +1298,35 @@ def _as_list(x):
     return list(x) if isinstance(x, collections.Sequence) else [x]
 
 
+def _get_output_names(block, outputs):
+
+    current_output_names = set([out.name for out in outputs])
+
+    if outputs:
+        current_block = outputs[0].block
+        while current_block.idx != block.idx:
```
zhhsplendid: I'm not quite understand these lines of codes. Please correct me if I am wrong about following comments:

1. link_output_names = set() at every beginning of while loop and current_output_names = link_output_names at every end of while loop. So the assignment to current_output_names actually takes effect at only the last iteration. Is there something wrong you write the while loop here?

2. The  while loop code can be wrong if caller inputs with "outputs[0].block" which is an ancestor of "block". Make an assert and throw some exception message for checking that case?

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1102,119 @@ def append_backward(loss,
...
     program = loss.block.program
+
+    # Todo(liym27): when to add 1 in control flow
     program._appending_grad_times += 1
 
     if no_grad_set is None:
         no_grad_set = set()
     no_grad_set = copy.copy(no_grad_set)
     no_grad_dict = _get_stop_gradients_(program)
+    # no_grad_set only contain vars in block 0
```
zhhsplendid: contain -> contains

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1102,119 @@ def append_backward(loss,
         isinstance(callbacks, list)
 
     program = loss.block.program
+
+    # Todo(liym27): when to add 1 in control flow
```
zhhsplendid: Could you fully explain the comment? Did you mean when to add 1 in sub-block (control flow)?

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1102,119 @@ def append_backward(loss,
...
+    is_in_control_flow = current_block_idx != 0
+    if is_in_control_flow:
+        target_grad_block = program._create_block(
+            parent_idx=current_block.parent_idx)
+        target_grad_block._set_forward_block_idx(current_block_idx)
+        # after _create_block, program.current_block changes
+    else:
+        target_grad_block = root_block
+
+    def _get_son_parent_block_idx_dict(program, current_block_idx):
```
zhhsplendid: Make this _get_son_parent_block_idx_dict function out of append_backward function

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1102,119 @@ def append_backward(loss,
...
     no_grad_dict[0].update(list(map(_append_grad_suffix_, no_grad_set)))
 
     root_block = program.block(0)
 
-    fwd_op_num = root_block.desc.op_size()
     current_block_idx = program.current_block_idx
+
+    current_block = program.block(current_block_idx)
+
+    # create grad block if in control flow
```
zhhsplendid: Please add comment about the use case: currently it is only to support the optimizer.minimize in a switch branch, which can append_backward in a sub_block.

We had better to explain it because while_loop is also in control flow, but adding optimizer in while makes no sense. We should report error when it is in while_loop. I will ask Fei to do it. You can just fully explain it or add a TODO about while_loop here.

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1102,119 @@ def append_backward(loss,
...
+
+        no_grad_vars_dict[block_idx] = _find_no_grad_vars(
+            block, op_path_dict[block_idx], [loss], block_no_grad_set)
+
+        block_no_grad_set.update(no_grad_vars_dict[block_idx])
+        no_grad_dict[block_idx].update(
+            list(map(_append_grad_suffix_, block_no_grad_set)))
+
+        input_grad_names_set = None
+        # For double backward, input_grad_names is used for filter
```
zhhsplendid: grammar issue:

"used for filter" -> "used for filtering"
"some non-used gradients op" -> "some non-used gradients op(s)"

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1102,119 @@ def append_backward(loss,
...
+        # some non-used gradients op.
+
+        # Todo(liym27): need a better design.
+        # not support double grad in control flow now
```
zhhsplendid: In my understanding, double grad is op level design. I can discuss it with you but that's not your fault.

python/paddle/fluid/backward.py
```diff
@@ -1173,6 +1226,37 @@ def append_backward(loss,
...
+                    grad_var.op = op
+                    break
+            if grad_var.op is None:
+                raise ValueError("Unexpected branch")
+            attr_val = [param_var.name, grad_var.name]
+            if grad_var.op.has_attr(op_role_var_attr_name):
+                attr_val.extend(grad_var.op.attr(op_role_var_attr_name))
+            grad_var.op._set_attr(op_role_var_attr_name, attr_val)
+
+            # control flow: grad_var not in current block but in corresponding grad block
```
zhhsplendid: Put comment at line 1255 under line 1230

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
zhhsplendid: Forgot to say, since you use random in unit test, it is better to set program random seed, too.
The code example is:

main_program.random_seed = 123

python/paddle/fluid/backward.py
liym27: Thanks. Done

python/paddle/fluid/backward.py
liym27: Thanks, done.

python/paddle/fluid/backward.py
liym27: Done

python/paddle/fluid/backward.py
liym27: Thanks. Done

python/paddle/fluid/backward.py
liym27: done

python/paddle/fluid/layers/control_flow.py
liym27: > Question but nothing to change: in my understanding, params here are just forward input vars? So I think it won't be at conditional grad block?

`params` are just forward input vars ususlly. But `params` will be at conditional grad block if `minimize()` is called in conditional block and return grad vars. 
example code:
```
def true_fn():
    
    ...
    optimizer_ops, para_grads = adam.minimize(loss)
    return para_grads
```

However, `return para_grads` is not supported in this PR.

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
liym27: Thanks. Done.

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
liym27: Thanks. Done

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
liym27: Thanks. Done

python/paddle/fluid/backward.py
liym27: Yes. `program._appending_grad_times` is used for double grad which is not supported if `minimize/append_backward` is in sub-block (control flow)

python/paddle/fluid/backward.py
liym27: Thanks. Done

python/paddle/fluid/backward.py
liym27: Thanks a lot

python/paddle/fluid/backward.py
liym27: > 1. link_output_names = set() at every beginning of while loop and current_output_names = link_output_names at every end of while loop. So the assignment to current_output_names actually takes effect at only the last iteration. Is there something wrong you write the while loop here?

Only the result of  the last iteration is needed because it is in `cur_block` . 
I modified variable names to make it easier to understand the while loop:
image



> 2. The while loop code can be wrong if caller inputs with "outputs[0].block" which is an ancestor of "block". Make an assert and throw some exception message for checking that case?

done.

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
liym27: yes. I will add a test case

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1112,114 @@ def append_backward(loss,
...
+        # Todo: support _append_backward_ops_with_checkpoints_ in
+        #  sub-block (control flow)
+        if checkpoints != None and \
+                isinstance(checkpoints, list) and \
+                len(checkpoints) > 0:
+            program_stat, checkpoint_names, \
+            vars_should_be_hold, \
+            recompute_segments = \
+                _append_backward_ops_with_checkpoints_(
+                    root_block,
```
sneaxiy, line 1192: Why `root_block` instead of `block`?

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1112,114 @@ def append_backward(loss,
...
+                    op_path_dict[block_idx],
+                    root_block,
```
sneaxiy, line 1194: Same above.

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1112,114 @@ def append_backward(loss,
...
+                block,  # the block where forward ops are in
+                op_path_dict[block_idx],
+                target_grad_block,
+                no_grad_dict,
+                grad_to_var,
+                callbacks,
+                input_grad_names_set=input_grad_names_set)
+
+    grad_info_map = dict()
+    fwd_op_num = block_fwd_op_num_dict[
```
sneaxiy, line 1212: Add a comment to why set `fwd_op_num` to zero. I guess it is because the grad ops of a forward control flow op are inside a different sub-block, so the forward op number in this grad sub-block is 0, is it right? Some developers may not know why there is this setting.

python/paddle/fluid/backward.py
```diff
@@ -1101,67 +1112,114 @@ def append_backward(loss,
...
+    target_grad_block.desc.append_op().copy_from(op_desc)
 
-    block_no_grad_set = set(map(_strip_grad_suffix_, no_grad_dict[0]))
-    op_path = _find_op_path_(root_block, [loss], [], block_no_grad_set)
-    no_grad_vars = _find_no_grad_vars(root_block, op_path, [loss],
-                                      block_no_grad_set)
-    block_no_grad_set.update(no_grad_vars)
-    no_grad_dict[0].update(list(map(_append_grad_suffix_, block_no_grad_set)))
+    op_path_dict = {}  # block idx : op_path
+    no_grad_vars_dict = {}
```
sneaxiy: Do we need these 2 dicts instead of using immediate variable in the following for loop codes?

python/paddle/fluid/backward.py
```diff
@@ -1173,6 +1231,37 @@ def append_backward(loss,
         parameters = [param.name for param in params if param.trainable]
 
     params_and_grads = []
+    op_role_var_attr_name = core.op_proto_and_checker_maker.kOpRoleVarAttrName()
+    if is_in_control_flow:
```
sneaxiy: Seems lots of duplicate codes with what is when `is_in_control_flow` is false. Can we combine these two parts of codes for better understanding?

python/paddle/fluid/layers/control_flow.py
```diff
@@ -1824,6 +1829,90 @@ def complete(self):
...
+                        params.add(in_var_name)
+
+            for oname in each_op.output_names:
+                for out_var_name in each_op.output(oname):
+                    intermediate.add(out_var_name)
+
+        param_list = []
+        for inner_out_name in params:
+            inner_var = parent_block._find_var_recursive(inner_out_name)
+            if inner_var:
```
sneaxiy, line 1877: In which case `inner_out_name` cannot be found in `parent_block`?

sneaxiy (commented):
Please check whether the strategy works in multi-gpu training.

python/paddle/fluid/backward.py
liym27: `_append_backward_ops_with_checkpoints_ ` does not handle programs with subblock, so it is not called in control flow now.
image

python/paddle/fluid/backward.py
liym27: `_append_backward_ops_with_checkpoints_` does not handle programs with subblock, so it is not called in control flow now.

python/paddle/fluid/backward.py
liym27: > I guess it is because the grad ops of a forward control flow op are inside a different sub-block, so the forward op number in this grad sub-block is 0, is it right? 

Yes.

> Add a comment to why set fwd_op_num to zero. 

Thanks. Done.

python/paddle/fluid/layers/control_flow.py
liym27: For better understanding, `inner_out_name` is renamed `inner_input_name` which means the input name in grad block.
`inner_input_name` can be from `forward sub block` or `parent block`. If `inner_input_name` is created in `forward sub block`, it can't be found in `parent block`.

**For example:**
`cross_entropy_loss` is created in block 1. it  is also the input of mean_grad op in block 2 and is can't be found in parent block.
```
op1
op2
...

def fn_1():
    # this is block 1 and the grad ops is in block 2
    cross_entropy_loss = layers.cross_entropy(input=prediction, label=label)
    loss1 = layers.mean(cross_entropy_loss)
    adam.minimize(loss1)

def fn_2():
    # this is block 3 and the grad ops is in block 4
    softmax_loss = layers.softmax_with_cross_entropy(logits=prediction, label=label)
    loss2 = layers.mean(softmax_loss)
    adagrad.minimize(loss2)

layers.case([(switch_id==id, fn_2)], fn_1)
```

python/paddle/fluid/backward.py
liym27: Yes. Done.

liym27:
> Please check whether the strategy works in multi-gpu training.

Thanks.  multi-gpu training will be supported later.

python/paddle/fluid/backward.py
liym27: No, use immediate variable instead.

python/paddle/fluid/layers/control_flow.py
zhhsplendid: Got you. LGTM for this point

zhhsplendid (approved):
LGTM

sneaxiy (approved):
LGTM.

PR #21853, opened 2019-12-19
 fix the unittest of test_check_import_scipy on V100

silingtong123:
fix the unittest of test_check_import_scipy on V100 and  remove useless comments


liupluswei (approved):
LGTM

PR #21864, opened 2019-12-20
Set dtype of compare op's output to bool

zhiqiu:
As the title.
For example, `c = a>b`, `c.dtype` should be bool, however it is set to `a.dtype` before.


liym27 (approved):

liym27:
LGTM

PR #21898, opened 2019-12-23
fix elementwise broadcast bug

danleifeng:
1. fix bug when x and y need do broadcast but do same-dimension function during x.numel() == y.numel() in elementwise_mul.
2. fig bug in elementwise_add/sub/mul/div default_elementwise_x function:change numel into size.


gavin1332 (approved):
LGTM

PR #21899, opened 2019-12-23
Improve readability of function: cache_third_party, fix the code to better

zhwesky2010:
Fix function of cache_third_party to better, set **${target}_SOURCE_DIR**  out of function.
And pass it to the function. These can improve readability of code.


liupluswei (approved):
LGTM

PR #21901, opened 2019-12-23
fix cp directory bug of warpctc

zhwesky2010:
fix cp directory bug of warpctc.

**It result:** 

1bb4f369326306aba7d9dcd73ea4cee0

**After this PR:**
image


kuke (approved):
LGTM

PR #21907, opened 2019-12-24
Use skip_check_grad_ci to skip grad check

GaoWei8:
Use skip_check_grad_ci to pass grad check.

In TestConcatOp3, check_grad on large input is too slow ( more than one hour ).
PR#17015 explains the reason for skipping grad check in TestConcatOp4.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_concat_op.py
+++ b/python/paddle/fluid/tests/unittests/test_concat_op.py
@@ -16,7 +16,7 @@ from __future__ import print_function
 
 import unittest
 import numpy as np
-from op_test import OpTest
+from op_test import OpTest, skip_check_grad_ci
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/tests/unittests/test_concat_op.py
```diff
@@ -76,6 +77,9 @@ def test_check_grad(self):
         pass
 
 
+@skip_check_grad_ci(
+    reason="This test will meet fetch error when there is a nul grad. The detailed information is in PR#17015."
```
zhangting2020: when there is a nul(typo) grad

GaoWei8: Done.

zhangting2020 (approved):

PR #21910, opened 2019-12-24
add decorator skip_check_grad_ci for test_group_norm 

lijianshe02:
In test_group_norm, the TestGroupNormOpLargeData and TestGroupNormOpLargeData_With_NHWC two test cases just to compare grads between CPU and GPU, the check_grad is not required.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_group_norm_op.py
+++ b/python/paddle/fluid/tests/unittests/test_group_norm_op.py
@@ -152,7 +152,8 @@ class TestGroupNormOpBigEps3(TestGroupNormOp):
 
 
 @skip_check_grad_ci(
-    reason="Compare grads between CPU and GPU, check_grad is not required.")
+    reason="This test case is to compare whether the grads computation results between CPU and GPU are consistent in the same inputs, check_grad is not required."
+)
 class TestGroupNormOpLargeData(TestGroupNormOp):
     def init_test_case(self):
         self.shape = (2, 32, 64, 64)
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/tests/unittests/test_group_norm_op.py
```diff
@@ -151,6 +151,8 @@ def init_test_case(self):
         self.attrs['epsilon'] = 0.5
 
 
+@skip_check_grad_ci(
+    reason="Compare grads between CPU and GPU, check_grad is not required.")
```
zhangting2020: 这个单测应该是比较相同输入的情况下，反向Op的CPU和GPU实现计算结果是否一致？注释可以稍详细点。

lijianshe02: 好的，谢谢

zhangting2020 (dismissed):
LGTM.

python/paddle/fluid/tests/unittests/test_group_norm_op.py
```diff
@@ -151,6 +151,10 @@ def init_test_case(self):
         self.attrs['epsilon'] = 0.5
 
 
+@skip_check_grad_ci(
+    reason='''This test case is to compare whether the grads computation results between CPU and GPU  
+            are consistent when using the same inputs, check_grad is not required.'''
```
luotao1: This test case is used to ensure whether the gradient checking results between CPU and GPU are consistent when using the same inputs, thus, it doesn't need to call check_grad.
英文可以再增强下。

lijianshe02: 好的，谢谢～

luotao1 (approved):
LGTM

PR #21934, opened 2019-12-25
[Bug Fix] Fix trt calib engine key bug

cryoco:
**Bug:** Graph pattern detector uses `unordered_map` and `unordered_set` to store pdnodes-to-nodes mappings, where the sorting is unstable, causing TRT engine key to change even if the model stays the same. This bug causes TRT int8 calibration to generate table data while it already exists.
**Fix:** Changed the `unordered_map` and `unordered_set` of pdnodes-to-nodes mapping to `map` and `set`. Now TRT int8 calibration works properly.


NHZlX (approved):
LGTM

PR #21942, opened 2019-12-25
Remove double registered dataType in Pad2d

Aurelius84:
+  Remove `double` in `pad2d` op .(Added by PR https://github.com/PaddlePaddle/Paddle/pull/21718)
+ Kernel with `double` will trigger a compile bug in CUDA10. The error info as follows:
http://ce.paddlepaddle.org:8080/viewLog.html?buildId=74793&buildTypeId=Benchmark_FrameworkBenchmark_BenchmarkBuild&tab=buildLog

## Reason
I found that `atomicAdd()` for double-precision floating-point numbers is not available on devices with compute capability lower than 6.0 but it can be implemented as follows:
```
#if __CUDA_ARCH__ < 600
__device__ double atomicAdd(double* address, double val)
{
    unsigned long long int* address_as_ull =
                              (unsigned long long int*)address;
    unsigned long long int old = *address_as_ull, assumed;

    do {
        assumed = old;
        old = atomicCAS(address_as_ull, assumed,
                        __double_as_longlong(val +
                               __longlong_as_double(assumed)));

    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
    } while (assumed != old);

    return __longlong_as_double(old);
}
#endif
```

**See details in** [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions)
[Same issue. ](https://devtalk.nvidia.com/default/topic/1027876/why-does-atomicadd-not-work-with-doubles-as-input-/)

### DIFF
--- a/paddle/fluid/operators/pad2d_op.cc
+++ b/paddle/fluid/operators/pad2d_op.cc
@@ -661,7 +661,5 @@ REGISTER_OPERATOR(pad2d, ops::Pad2dOp, ops::Pad2dOpMaker,
                   ops::Pad2dOpGradMaker<paddle::imperative::OpBase>);
 REGISTER_OPERATOR(pad2d_grad, ops::Pad2dOpGrad,
                   ops::Pad2dOpGradNoNeedBufferVarsInference);
-REGISTER_OP_CPU_KERNEL(pad2d, ops::Pad2dCPUKernel<float>,
-                       ops::Pad2dCPUKernel<double>);
-REGISTER_OP_CPU_KERNEL(pad2d_grad, ops::Pad2dGradCPUKernel<float>,
-                       ops::Pad2dGradCPUKernel<double>);
+REGISTER_OP_CPU_KERNEL(pad2d, ops::Pad2dCPUKernel<float>);
+REGISTER_OP_CPU_KERNEL(pad2d_grad, ops::Pad2dGradCPUKernel<float>);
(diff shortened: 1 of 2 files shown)


liym27 (approved):

liym27:
LGTM for  dataType  registration.

luotao1:
> will trigger a compile bug in CUDA10

But the CE is `[22:41:44]-- CUDA detected: 9.0`
http://ce.paddlepaddle.org:8080/viewLog.html?buildId=74793&tab=buildLog&buildTypeId=Benchmark_FrameworkBenchmark_BenchmarkBuild&logTab=tree&filter=all&_focus=477

> I found that atomicAdd() for double-precision floating-point numbers is not available on devices with compute capability lower than 6.0

`group_norm_op.cu` use `atomicAdd()` for double-precision, but it doesn't fail before.
https://github.com/PaddlePaddle/Paddle/blob/9c9645abd0497ffe4c332229c2f4629169dab6a7/paddle/fluid/operators/group_norm_op.cu#L365-L368

https://github.com/PaddlePaddle/Paddle/blob/9c9645abd0497ffe4c332229c2f4629169dab6a7/paddle/fluid/operators/group_norm_op.cu#L45

Anyway, you can revert double registered dataType in Pad2d at first.

Aurelius84:
> Anyway, you can revert double registered dataType in Pad2d at first.

Thanks suggestions very much! I will inform the relevant developer in charge of this op to have a look at it together.

PR #21947, opened 2019-12-25
update layers used in mnist dygraph model, test=develop

zhiqiu:
As the title, this PR updates several layers used in `mnist` model.
Before this PR, in dygraph mode, the layers reuse the same logic with static graph, which slow down the execution efficiency.
After this PR, in dygraph mode, these layers use core.ops.xx directly in order to execute more efficiently.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_imperative_resnet.py
+++ b/python/paddle/fluid/tests/unittests/test_imperative_resnet.py
@@ -275,7 +275,7 @@ class TestDygraphResnet(unittest.TestCase):
                 if traced_layer is not None:
                     resnet.eval()
                     traced_layer._switch(is_test=True)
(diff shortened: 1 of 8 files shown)


python/paddle/fluid/tests/unittests/test_imperative_resnet.py
```diff
@@ -275,7 +275,7 @@ def test_resnet_float32(self):
                 if traced_layer is not None:
                     resnet.eval()
                     traced_layer._switch(is_test=True)
-                    out_dygraph = resnet([img])
```
sneaxiy, line 278: Why change this?

sneaxiy (dismissed):
LGTM.

sneaxiy (approved):
LGTM.

PR #21961, opened 2019-12-26
Refine multihead kernel, align block to 32

zhaoyuchen2018:
Align blocksize to 32 as cuda run with 32 threads together, also no need to set lane mask.
Refine some log code.

test=develop


wangchaochaohu (approved):
LGTM

PR #21970, opened 2019-12-26
update cmake from version 3.5.1 to 3.16.0

zhwesky2010:
- Update cmake from version 3.5.1 to 3.16.0.

- It will support more cmake grammar, such as SUBSOURCE_DIR. And faster compilation slightly.


luotao1 (approved):
LGTM！！

PR #22015, opened 2019-12-30
Fix the doc of scatter_nd

ForFishes:
Because the `scatter_nd_add` and `scatter_nd` do not support the `int32` and 'int64',  we should remove these types from the reference of two ops.  
api_cn :https://github.com/PaddlePaddle/FluidDoc/pull/1695


Boyan-Liu (approved):
LGTM

gavin1332 (approved):
LGTM

PR #22016, opened 2019-12-30
add register op_data_type of pow/square

SunAhong1993:
In this PR, int/int64_t/float/double of above ops are all registered.


Aurelius84:
> In this PR, int/int64_t/float/double of above ops are all registered.

commit后面需要添加test=develop触发其他CI

SunAhong1993:
> > In this PR, int/int64_t/float/double of above ops are all registered.
> 
> commit后面需要添加test=develop触发其他CI

我重新提交了一个PR：https://github.com/PaddlePaddle/Paddle/pull/22023 ，稍后把这个PR close掉。

PR #22029, opened 2019-12-31
Add ParallelExecutor Test for Cond API and Fix PE Checks Shape Bug

zhhsplendid:
Add ParallelExecutor test to prove Cond API Can Work in ParallelExecutor.

Also found a bug in ParallelExecutor and fix it. The bug is that ParallelExecutor splits data evenly on devices, but we didn't divide the number of devices when checking feed data shape in ParallelExecutor


sneaxiy (approved):
LGTM.

sneaxiy (changes requested):
Think about lod cases?

zhhsplendid:
> Think about lod cases?

Added

sneaxiy (approved):
LGTM.

PR #22036, opened 2019-12-31
fix integer overflow in match_matrix

Aurelius84:
+ 修复了op在输入lod值过大时integer overflow问题。

### BUG复现
match_matrix_tensor_op.cc中MatchMatrixTensorOP::InferShape在处理生成`out_dim_0`时存在integer overflow，精心构造out张量的tensor shape

在Compute时memset out_data的size比较小，可导致Compute调用`call_gemm_with_lda`时由于使用到out_data数据而造成崩溃出错

例子：精心设置X的dims为{1789569713, 1}，lods为{1, 1073741823, 1789569713}，Y的dims为{6, 3}，lods为{1, 3, 6}，dim_t属性为1，这样在计算out_dim_0时由于整型溢出最终获得值为18

### 测试用例
将`test_match_matrix_tensor_op.py`代码改为如下：
```
class TestMatchMatrixTensorOpCase4(TestMatchMatrixTensorOp):
    def set_data(self):
        ix, iy, h, dim_t = [5, 9, 32, 1]
        x_lod = [[1, 2, 2]]
        y_lod = [[3, 2, 4]]
        self.init_data(ix, x_lod, iy, y_lod, h, dim_t)
    def test_api(self):
        x_lod_tensor = fluid.layers.data(name='x', shape=[1], lod_level=1)
        y_lod_tensor = fluid.layers.data(name='y', shape=[1], lod_level=1)
        out, out_tmp = fluid.contrib.match_matrix_tensor(
            x=x_lod_tensor, y=y_lod_tensor, channel_num=1)

        place = fluid.CPUPlace()
        x_data = np.random.rand(1789569713, 1).astype('float32')
        y_data = np.random.rand(6, 1).astype('float32')
        x = fluid.create_lod_tensor(x_data, [[1073741823, 715827890]], place)
        y = fluid.create_lod_tensor(y_data, [[2, 4]], place)

        exe = fluid.Executor(place=place)
        ret = exe.run(feed={'x': x,
                            'y': y},
                      fetch_list=[out],
                      return_numpy=False)
```

### DIFF
--- a/paddle/fluid/operators/match_matrix_tensor_op.cc
+++ b/paddle/fluid/operators/match_matrix_tensor_op.cc
@@ -206,7 +206,7 @@ class CPUMatchMatrixTensorOPKernel : public framework::OpKernel<T> {
     for (size_t b = 0; b < x->lod()[0].size() - 1; b++) {
       for (int t = 0; t < dim_t; t++) {
         size_t len_l = offset_l[b + 1] - offset_l[b];
-        sizt_t len_r = offset_r[b + 1] - offset_r[b];
+        size_t len_r = offset_r[b + 1] - offset_r[b];
         auto* top_data = out_data + top_offset[b] + t * len_l * len_r;
         const auto* l_t_data =
             bottom_l_trans_data + offset_l[b] * dim_t * dim_in + t * dim_in;
--- a/paddle/fluid/operators/match_matrix_tensor_op.cc
+++ b/paddle/fluid/operators/match_matrix_tensor_op.cc
@@ -174,7 +174,7 @@ class CPUMatchMatrixTensorOPKernel : public framework::OpKernel<T> {
     auto* tmp = ctx.Output<LoDTensor>("Tmp");
 
     int dim_t = ctx.Attr<int>("dim_t");
-    auto dim_in = x->dims()[1];
+    int64_t dim_in = x->dims()[1];
 
     const auto& offset_l = x->lod()[0];
     const auto& offset_r = y->lod()[0];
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/match_matrix_tensor_op.cc
```diff
@@ -174,17 +174,17 @@ class CPUMatchMatrixTensorOPKernel : public framework::OpKernel<T> {
     auto* tmp = ctx.Output<LoDTensor>("Tmp");
 
     int dim_t = ctx.Attr<int>("dim_t");
-    int dim_in = x->dims()[1];
+    auto dim_in = x->dims()[1];
```
zhhsplendid: The fundamental rule of using "auto" in C++:

1. When it is easy to understand the type from the context. In Paddle, an example is

auto x = ctx.Input < LoDTensor > ("X");
People with prior knowledge will know x is a LoDTensor.

2. Since C++ is type-sensitive, DO NOT use "auto" when the type is not very clear from the context. That's difficult for new Paddle engineers or people from other team.

Here, I don't think "x->dims()[1]" is very clear. You can change auto to int64_t.

But Paddle has used "auto" for convenience of writing everywhere, so **I would approve this PR**, but I would like you to know the skill of using "auto". **It's up to you whether you would like to change it, or you may like a fast development** .

Reference:
https://google.github.io/styleguide/cppguide.html#Type_deduction

zhhsplendid (dismissed):

paddle/fluid/operators/match_matrix_tensor_op.cc
Aurelius84: Thanks very much! Fixed.

zhhsplendid (approved):
LGTM

PR #22039, opened 2019-12-31
fix bug of checking op desc: not DUPLICABLE but DISPENSABLE.

liym27:
The added Input of OP must be dispensable. Fix bug in check_op_desc.py.

This bug is found in PR https://github.com/PaddlePaddle/Paddle/pull/21420
image

image


luotao1 (approved):
LGTM

PR #22042, opened 2020-01-01
fix comment of ConstantInitializer

littletomatodonkey:
fix comment of ConstantInitializer.
image

### DIFF
--- a/python/paddle/fluid/initializer.py
+++ b/python/paddle/fluid/initializer.py
@@ -132,8 +132,11 @@ class ConstantInitializer(Initializer):
     """Implements the constant initializer
 
     Args:
-        value (float32): constant value to initialize the variable 
-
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/initializer.py
```diff
@@ -132,8 +132,11 @@ class ConstantInitializer(Initializer):
     """Implements the constant initializer
 
     Args:
-        value (float32): constant value to initialize the variable 
-
+        value (float32): constant value to initialize the variable
+        force_cpu (bool, default False): whether to force initialization on cpu.
```
Boyan-Liu: cpu->CPU，与下面格式保持一致

littletomatodonkey: 好的，已修改~

FDInSky:
LGTM

PR #22075, opened 2020-01-03
Add register op_data_type of pad2d

FrostML:
Add register op_data_type of pad2d. In this PR, `float/double/int/int64_t` are all support for pad2d op. 
According to #21942 , `atomicAdd()` for double-precision floating-point numbers is not available on devices with compute capability lower than 6.0. 
This PR fix this problem by using `platform::CudaAtomicAdd()` rather than `atomicAdd()`.


Aurelius84 (approved):
LGTM for dataType registration.

FrostML:
image
According to QA's tests.

PR #22079, opened 2020-01-03
Fix segmentation fault bug in reduce ops

qjing666:
- Fix the segmentation fault bug when dim = [] in reduce ops

- If dim = [], execute the same as dim = None


guru4elephant (approved):
LGTM, please add ut in another pr

PR #22081, opened 2020-01-04
Remove the unittests that are no longer maintained

silingtong123:
Remove the unittests that are no longer maintained


seiriosPlus (approved):
LGTM

PR #22088, opened 2020-01-05
[MKL-DNN] Conv grad and Batch Norm grad NHWC support

jczaja:
This PR is introducing support to NHWC Paddle layout for conv and batch norm grad ops as discussed #20964.

### DIFF
--- a/paddle/fluid/platform/mkldnn_helper.h
+++ b/paddle/fluid/platform/mkldnn_helper.h
@@ -74,6 +74,11 @@ tf_pd<Type> MKLDNNBwdPrimitiveDesc(const Engine& e, const Primitive& p,
 inline void MatchShapeToLayout(framework::Tensor* tensor_in,
                                framework::DataLayout from,
                                framework::DataLayout to) {
+  // Shape changing makes sense for 3+ dims Tensors
+  if (tensor_in->dims().size() < 3) {
+    return;
+  }
+
   switch (from) {
     case framework::DataLayout::kMKLDNN:
       if (to == framework::DataLayout::kNHWC) {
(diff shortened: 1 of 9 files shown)


grygielski (approved):
LGTM

bingyanghuang:
@luotao1 Please start a review.

paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -74,6 +74,11 @@ tf_pd<Type> MKLDNNBwdPrimitiveDesc(const Engine& e, const Primitive& p,
 inline void MatchShapeToLayout(framework::Tensor* tensor_in,
                                framework::DataLayout from,
                                framework::DataLayout to) {
+  // Shape changing makes sense for 3+ dims Tensors
+  if (tensor_in->dims().size() < 3) {
+    return;
+  }
+
```
LeoZhao-BJ, line 81: what's expected behavior if dims <3 ? nothing to do?

LeoZhao-BJ (commented):
LGTM

paddle/fluid/platform/mkldnn_helper.h
grygielski: In these data layouts, channel dimension is either on 2nd position: nChw or at last nhwC, so for dim==2 these layouts are the same and nothing should be done. Similarly for dim==1 when you have just one possible combination.

luotao1: @grygielski Could you add the above comments in the codes in the next PR?

luotao1 (approved):
LGTM

PR #22090, opened 2020-01-05
Fix the global_step & continuous applying error in EMA

kuke:
Fix two problems:
1) The global step should count independently, instead of dependent on learning rate scheduler;
2) The result is not correct when applying EMA continuously.


LielinJiang:
tried in image classification task, everything looks good.

LielinJiang (dismissed):

Boyan-Liu (dismissed):
LGTM

wanghaoshuang (dismissed):

wanghaoshuang (approved):
LGTM

Boyan-Liu (approved):

swtkiwi (approved):
LGTM

PR #22105, opened 2020-01-06
Fix windows build no kernel issue, test=develop

zhaoyuchen2018:
windows conv_fusion failed as no kernel, maybe some bugs  in compiler


wangchaochaohu (approved):
LGTM

PR #22112, opened 2020-01-06
[Cherry-pick] Fix optimizer op infershape failed in dygraph multi-cards mode

chenwhql:
* add param & grad shape check for sgd op
* add _reshape_inplece interface for dygraph parallel
* refine unittest based paddle/models scripts, test=develop
* add unittest for parallel grad fuse, test=develop


sneaxiy (approved):
LGTM.

PR #22113, opened 2020-01-06
fix diff_unittest.py of ci rule to avoid Misjudgment

zhwesky2010:
fix diff_unittest.py of ci rule to avoid Misjudgment.

It have problem:
- If change order of two unittest `a` and `b`, it will be regaded as deleted `a` and deleted `b`.


luotao1 (approved):
LGTM

PR #22124, opened 2020-01-07
Fix Xception precision problem during inference

FrostML:
According to [models #4156](https://github.com/PaddlePaddle/models/issues/4156) Before this fix, Xception models may face a precision problem when using python inference api. 

Reason:
The diff is caused by conv_elementwise_add_fuse_pass. 
conv_elementwise_add_fuse_pass only supports elementwise_add op as conv's bias to trigger the fuse pattern. 
But in Xception models, conv_elementwise_add_fuse_pass was also triggered by doing elementwise_add between conv's result and an input variable(tensor).

Before this fix: 
image
After this fix:
image


NHZlX (approved):
LGTM

PR #22145, opened 2020-01-07
Fix init_loss_scaling value bug.

gongweibao:
The mixed precision interface will adapt to scaling value when it is greater than 1.0. So the initialization value can't be 1.0. It should be a greater value.


gavin1332 (approved):
lgtm

lanxianghit (approved):

PR #22152, opened 2020-01-07
[Bug Fix] Fix trt instance_norm serialize bug

cryoco:
Add plugin type size to instance_norm serialized size to fix double-free in serializing


NHZlX (approved):
LGTM

PR #22157, opened 2020-01-08
[Cherry-pick 1.6] fix batch_norm_grad shape=0 & allreduce shape enforce & sync_batch_norm hang in fleet

wangxicoding:
1. When scale_grad is null, bias_grad is not null, batch_norm_grad will not set bias_grad, shape will be 0. Now, enforce scale & bias must be trainable same time or no trainable same time.
2. enforce allreduce size must > 0.
3. Temp fix sync_batch_norm hang in fleet

more detail, see 
https://github.com/PaddlePaddle/Paddle/pull/21801
https://github.com/PaddlePaddle/Paddle/pull/21838


gongweibao (approved):
LGTM

PR #22165, opened 2020-01-08
Fix auto growth allocator bug when using non zero gpu card

sneaxiy:
The following codes would hang in dygraph mode when auto growth allocator is enabled:

```python
import paddle.fluid as fluid

with fluid.dygraph_guard(fluid.CUDAPlace(1)):
    emb = fluid.dygraph.Embedding(size=[10, 10])
    var = fluid.dygraph.to_variable(numpy.random.random([11, 13]).astype('float32'))
```

The hang happens when calling `cudaMemcpy` in `to_variable` method (found by gdb). It seems that `cudaMemcpy` would hang if the device id of the current cuda context is not right (for example, the device id is 0 but the pointer to be copied is located on device 1). 

It happens in auto growth allocator but does not happen in the previous allocator. It is because the previous `naive_best_fit` allocator would call `cudaSetDevice` when the first allocation request and call `platform::CUDADeviceGuard` before calling each `cudaMalloc`. But auto growth allocator would not call `cudaSetDevice` when the first allocation request and would call  `platform::CUDADeviceGuard` before calling each `cudaMalloc`. Supposing that the first allocation request is from device 1, `platform::CUDADeviceGuard` would make the current device id to be 0 (default device id of cuda) when destructed, which makes the wrong device id when calling the following `cudaMalloc`.

### DIFF
--- a/paddle/fluid/memory/allocation/cuda_allocator.cc
+++ b/paddle/fluid/memory/allocation/cuda_allocator.cc
@@ -33,6 +33,8 @@ void CUDAAllocator::FreeImpl(Allocation* allocation) {
 }
 
 Allocation* CUDAAllocator::AllocateImpl(size_t size) {
+  std::call_once(once_flag_, [this] { platform::SetDeviceId(place_.device); });
+
   platform::CUDADeviceGuard guard(place_.device);
   void* ptr;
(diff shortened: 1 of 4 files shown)


phlrain (approved):
LGTM

PR #22181, opened 2020-01-08
 add NotImplementedError for multi optimizers

liym27:
#### This PR adds `NotImplementedError`.

---

 - **For example**: 
Optimizer adam or sgd is used in control flow：
```
def fn_1(opt, avg_loss):
    opt.minimize(avg_loss)

def fn_2(opt, avg_loss):
    opt.minimize(avg_loss)

x = fluid.layers.data("X", [10], 'float32')
hidden = layers.fc(x, 5)
avg_loss = layers.mean(hidden)

adam = optimizer.Adam(learning_rate=LR)
sgd = optimizer.SGD(learning_rate=LR)

cond = layers.fill_constant([1], 'bool', True)

layers.case([(cond, lambda: fn_1(adam, avg_loss))],
            lambda: fn_2(sgd, avg_loss))
```

- #### if it is run on multi cards with parallel executor, **error message** is as follows:
```
NotImplementedError: If optimizer is used in control flow, training on multi-places is not supported now.

```

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
+++ b/python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
@@ -252,23 +252,32 @@ class TestMultiOptimizersMultiCardsError(unittest.TestCase):
             layers.case([(cond, lambda: fn_1(adam, avg_loss))],
                         lambda: fn_2(sgd, avg_loss))
 
-        place = fluid.CUDAPlace(0) if use_cuda else fluid.CPUPlace()
-        exe = fluid.Executor(place)
-        exe.run(startup_program)
+        cpu_place = fluid.CPUPlace()
+        cuda_place = fluid.CUDAPlace(0) if use_cuda else fluid.CPUPlace()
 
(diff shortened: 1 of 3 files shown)


sneaxiy (dismissed):

Aurelius84 (dismissed):

python/paddle/fluid/tests/unittests/test_optimizer_in_control_flow.py
```diff
@@ -225,5 +227,58 @@ def test_optimzier_in_switch(self):
                                                                    loss_2))
 
 
+class TestMultiOptimizersMultiCardsError(unittest.TestCase):
```
sneaxiy, line 230: You should add this unittest here https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/CMakeLists.txt#L327 so that it can run in 2 gpu cards in CI.

liym27: Thanks. Done

sneaxiy (approved):

PR #22195, opened 2020-01-09
add backward gradient computation for op argsort

FlyingQianMM:
The function argsort does not implement backward gradient computation, so we add backward gradient computation.
test=develop


paddle/fluid/operators/argsort_op.cu
```diff
@@ -46,18 +46,31 @@ struct SegmentOffsetIter {
   int num_cols_;
 };
 
-template <typename T>
-static __global__ void FillIndex(T* indices, T num_rows, T num_cols) {
+template <typename t>
```
wawltor, line 49: change the template t to template T

python/paddle/fluid/tests/unittests/test_argsort_op.py
```diff
@@ -48,14 +48,17 @@ def init_axis(self):
...
 
     def init_direction(self):
         self.descending = False
 
     def test_check_output(self):
         self.check_output()
 
+    def test_check_grad(self):
+        self.check_grad(['X'], 'Out')
+
```
wawltor, line 61: Add the float32 test case.

PR #21378, opened 2019-11-26
Polish the codes of fc when needs padding

GaoWei8:
optimize fc padding code
fix https://github.com/PaddlePaddle/Paddle/pull/20972#pullrequestreview-322709753

PR #20895, opened 2019-10-30
Optimize the kernel implementation of layernorm with openmp

yihuaxu:
Based on [investigation of ERNIE](https://github.com/PaddlePaddle/benchmark/issues/180#issuecomment-546296907), we found that layernorm is lack of multi-threads JIT implemention. This PR is to add the layernorm multi-threads optimizing by using OpenMP.
But based on initial benchmark with ERNIE, single thread layernorm performance seems to be worse than before.

**CPU Model:** Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz

Thread Number | Baseline | Optimization
-- | -- | --
1 thread | 2341.13ms | 2502.5ms
20 threads | 4009.16ms | 847.298ms

PR #22198, opened 2020-01-09
[cherry-pick] Add FC padding, ernie test unit and layernorm parallel

GaoWei8:
Optimize the kernel implementation of layernorm with openmp (#20895)
Add ernie c++ inference test (#21015)
fix cmake fails on inference_download_and_uncompress (#21185)
Add fc padding to improve mkl GEMM's performance when N and K are multiple of 128. (#20972)
Polish the codes of fc when needs padding (#21378)
Add ernie large c++ inference test (#21365)
Modify padding strategy: remove weight copy in fc padding (#21650)
optimize fc jit (#21878)


Xreki (approved):
LGTM

PR #22217, opened 2020-01-10
No fp64 check for mkldnn op test for now, test=develop

juncaipeng:
No fp64 check for mkldnn op test for now, test=develop

### DIFF
--- a/python/paddle/fluid/tests/unittests/op_test.py
+++ b/python/paddle/fluid/tests/unittests/op_test.py
@@ -209,8 +209,8 @@ class OpTest(unittest.TestCase):
 
         if not hasattr(cls, "op_type"):
             raise AssertionError(
-                "This test do not have op_type in class attrs,"
-                " please set self.__class__.op_type=the_real_op_type manually.")
+                "This test do not have op_type in class attrs, "
+                "please set self.__class__.op_type=the_real_op_type manually.")
 
         # case in NO_FP64_CHECK_GRAD_CASES and op in NO_FP64_CHECK_GRAD_OP_LIST should be fixed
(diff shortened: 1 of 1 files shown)


zhangting2020 (approved):
LGTM

luotao1 (approved):
LGTM! @bingyanghuang @jczaja @LeoZhao-Intel Do you have a plan to support double type for all mkldnn kernel?

jczaja:
@bingyanghuang Currently DNNL (MKL-DNN) Does not support double data type. Supported data types are:
* int8
* fp32
* bfloat16

So we are not planning update to float64.

luotao1:
Got it. Thus, does Ngraph not support double data type, and have no planning update to float64 in Ngraph as well? @LeoZhao-Intel @baojun-nervana ?

baojun-nervana:
NGraph supports fp64. But NGraph cpu performant kernel relies on MKLDNN, so there is a dependency on mkldnn support.

luotao1:
@baojun-nervana @LeoZhao-Intel Since MKLDNN don't have a plan to update to fp64, and we skip all mkldnn op test in #22363, do you have some idea to skip all ngraph op test in `op_test.py` or using some other ways?

We use `use_mkldnn` attribute in `op_test.py` to judge whether an op is MKLDNN kernel, but since ngraph is integrated like sub-graph, we could not use the same way what we skip all mkldnn op test in #22363.

baojun-nervana:
Is it feasible to use "use_ngraph" flag to disable FP64 tests?

luotao1:
> Is it feasible to use "use_ngraph" flag to disable FP64 tests?

@baojun-nervana Do you have any ideas of this?

PR #22222, opened 2020-01-10
fix bug for test_dygraph_mnist_fp16.py

JepsonWong:
#22163 open the test_dygraph_mnist_fp16 optest, but there is a bug in generation of input data y. In mnist model, the output size of the last layer(Linear) is 10, so the range of input data y is 0-9.
But if program use np.random.randn(1, 1).astype("int64") function to generate y, the number generated may not be in this range.

image

error message in http://ci.paddlepaddle.org/viewLog.html?buildId=267450&buildTypeId=Paddle_PrCiCoverage&tab=buildLog&branch_Paddle=pull%2F22163 and http://ci.paddlepaddle.org/viewLog.html?tab=buildLog&buildTypeId=Paddle_PrCiCoverage&buildId=267450 :

image


JepsonWong:
@chenwhql please help review it, thx.

chenwhql (approved):
LGTM

PR #22223, opened 2020-01-10
Add caching mechanism to requantize_mkldnn_op

grygielski:
This PR adds caching mechanism to requantize_mkldnn_op in order to store DNNL reorder primitives in blobs instead of creating them in every call.


bingyanghuang:
This PR is to fix the performance regression of int8 model on Paddle caused by upgrading to MKL-DNN 1.1.

luotao1:
> This PR is to fix the performance regression of int8 model on Paddle caused by upgrading to MKL-DNN 1.1.

Please add some detail benchmark of performance regression.

wojtuss (approved):
LGTM

PR #22240, opened 2020-01-13
fix use cache file in post quantization, test=develop

juncaipeng:
* fix use cache file, test=develop
* remove some ops for quantization in resnet50 test


wzzju (approved):
LGTM.

PR #22246, opened 2020-01-13
Polish backward.py to prune more ops

sneaxiy:
This PR tries to fix some issues that users have to set `var.stop_gradient = True` when appending backward. The idea is simple: if the all gradient inputs or outputs of the backward op does not exist, prune this op directly. But there are 2 corner cases that should be concerned:

- the while grad op may take a var that does not exist in forward/parent block as its input. This input can not be found when calling `_append_backward_vars_`. These kinds of vars are the recursive outputs of while grad op, so that it is also the output of while grad op. This PR adds the restriction on this case.
- **`rnn_memory_helper_grad` op takes a var that does not exist in the program as input**. This should not happen in normal ops! But the design of this op is a bit ugly. So I skip this case in this PR, otherwise this op would be pruned, and the calculation result would be wrong.

CE has passed: http://ce.paddlepaddle.org:8080/viewLog.html?buildId=78889&tab=buildResultsDiv&buildTypeId=PaddleModesl_ModelDebug_Model11Build

### DIFF
--- a/python/paddle/fluid/backward.py
+++ b/python/paddle/fluid/backward.py
@@ -937,6 +937,27 @@ def _is_grad_var_(var):
     return core.grad_var_suffix() in var
 
 
+# Find the op whose holds the sub_block as its "sub_block" attr
+def _find_parent_op_(sub_block):
+    sub_block_id = sub_block.idx
+
+    if sub_block_id == 0:
+        return None
+
+    program = sub_block.program
+    for block_id in six.moves.range(program.num_blocks):
+        block_desc = program.block(block_id).desc
+        for op_idx in six.moves.range(block_desc.op_size()):
+            op = block_desc.op(op_idx)
+            if op.has_attr("sub_block") and op._block_attr_id(
+                    "sub_block") == sub_block_id:
+                return op
+
+    # NOTE(paddle-dev): When optimizer is added in conditional block, 
+    # sub_block may not be found.
+    return None
+
+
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/backward.py
```diff
@@ -933,6 +933,31 @@ def _append_backward_ops_(block,
                 cb(block=target_block, context=grad_to_var)
 
 
+def _is_grad_var_(var):
```
zhhsplendid: rename "var" as "var_name"?

python/paddle/fluid/backward.py
```diff
@@ -933,6 +933,31 @@ def _append_backward_ops_(block,
...
+    return core.grad_var_suffix() in var
+
+
+# Find the op whose holds the sub_block as its "sub_block" attr
```
zhhsplendid: NIT: "Find the op whose holds" -> "Find the op who holds"

zhhsplendid (commented):
Other things LGTM

python/paddle/fluid/backward.py
sneaxiy: Done.

python/paddle/fluid/backward.py
sneaxiy: Done.

zhhsplendid (approved):
LGTM now

Aurelius84 (approved):

luotao1 (approved):
LGTM for backward.py

PR #22255, opened 2020-01-13
Make PaddleBuf's malloc consistent with paddle_api.h

FrostML:
In api.cc, `malloc/free` are used for initializing `PaddleBuf` while in paddle_api.h, `new/delete` are involved. 
Since the valgrind will regard this inconsistency as memory leaking, this PR tries to fix this.

### DIFF
--- a/paddle/fluid/inference/api/api.cc
+++ b/paddle/fluid/inference/api/api.cc
@@ -85,7 +85,7 @@ void PaddleBuf::Resize(size_t length) {
   if (length_ >= length) return;
   if (memory_owned_) {
     Free();
-    data_ = malloc(length);
+    data_ = new void *[length];
     length_ = length;
     memory_owned_ = true;
   } else {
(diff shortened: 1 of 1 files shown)


paddle/fluid/inference/api/api.cc
```diff
@@ -103,7 +103,7 @@ void PaddleBuf::Reset(void *data, size_t length) {
 void PaddleBuf::Free() {
   if (memory_owned_ && data_) {
     PADDLE_ENFORCE_GT(length_, 0UL);
-    free(static_cast<char *>(data_));
+    delete static_cast<char *>(data_);
```
fc500110: delete[]

fc500110 (approved):
LGTM

paddle/fluid/inference/api/api.cc
FrostML: Done. Thanks.

PR #22262, opened 2020-01-14
Fix the bug of assert_is_op_output.

wzzju:
Fix the bug of assert_is_op_output in the graph_pattern_detector.
* If the op `A` does not have an output named `B`, `assert_is_op_output(A, B)` will throw an exception.


sneaxiy (approved):

luotao1:
The PR description is too simple, please write how to trigger this bug, and the detailed error of this bug.

wzzju:
> The PR description is too simple, please write how to trigger this bug, and the detailed error of this bug.

Done.

PR #22275, opened 2020-01-14
[UT Coverage]Improve sum_mkldnn_op line coverage

lidanqing-vv:
Improve sum_mkldnn_op.cc (in_place) coverage

### DIFF
--- a/paddle/fluid/operators/sum_op.cc
+++ b/paddle/fluid/operators/sum_op.cc
@@ -113,14 +113,6 @@ class SumOp : public framework::OperatorWithKernel {
     framework::LibraryType library{framework::LibraryType::kPlain};
     framework::DataLayout layout{framework::DataLayout::kAnyLayout};
 
-#ifdef PADDLE_WITH_MKLDNN
-    if (library == framework::LibraryType::kPlain &&
-        platform::CanMKLDNNBeUsed(ctx)) {
-      library = framework::LibraryType::kMKLDNN;
-      layout = framework::DataLayout::kMKLDNN;
-    }
-#endif
-
     if (x_vars[0]->IsType<framework::LoDTensor>()) {
       int dtype = -1;
       for (size_t idx = 0; idx < x_vars.size(); ++idx) {
@@ -140,6 +132,18 @@ class SumOp : public framework::OperatorWithKernel {
       }
       PADDLE_ENFORCE_NE(dtype, -1,
                         "Sum operator should have at least one tensor");
+
+#ifdef PADDLE_WITH_MKLDNN
+      if (library == framework::LibraryType::kPlain &&
+          platform::CanMKLDNNBeUsed(ctx)) {
+        if (static_cast<framework::proto::VarType::Type>(dtype) ==
+            framework::proto::VarType::FP32) {
+          return framework::OpKernelType(
+              framework::proto::VarType::FP32, ctx.GetPlace(),
+              framework::DataLayout::kMKLDNN, framework::LibraryType::kMKLDNN);
+        }
+      }
+#endif
 
       return framework::OpKernelType(
           static_cast<framework::proto::VarType::Type>(dtype), ctx.GetPlace(),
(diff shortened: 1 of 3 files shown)


bingyanghuang:
@lidanqing-intel Please add `test=develop` to trigger CI.

lidanqing-vv:
Hi, @wojtuss please help review. Thanks

lidanqing-intel:
This test passed on my local machine, but failed on CI, I am investigating.

lidanqing-vv:
@wojtuss @Sand3r- Hi, Please take a review. I removed the "fallback to naive version", locally sum_mkldnn_op coverage is 98%. Click "hide whitespaces changes" in settings so that it could be clear to review


image

paddle/fluid/operators/sum_op.cc
```diff
@@ -141,6 +133,18 @@ class SumOp : public framework::OperatorWithKernel {
       PADDLE_ENFORCE_NE(dtype, -1,
                         "Sum operator should have at least one tensor");
 
+#ifdef PADDLE_WITH_MKLDNN
+      if (library == framework::LibraryType::kPlain &&
+          platform::CanMKLDNNBeUsed(ctx)) {
+        if (static_cast<framework::proto::VarType::Type>(dtype) ==
+            framework::proto::VarType::FP32) {
```
wojtuss: This condition should match the one from the original `paddle/fluid/operators/mkldnn/sum_mkldnn_op.cc:62`.

wojtuss: Please, join the three `if`s into a single one.

lidanqing-vv: > Please, join the three `if`s into a single one.

Thank you very much. I have fixed these two reviews. please take a look.

paddle/fluid/operators/sum_op.cc
```diff
@@ -141,6 +133,26 @@ class SumOp : public framework::OperatorWithKernel {
...
+          static_cast<framework::proto::VarType::Type>(dtype) ==
+              framework::proto::VarType::FP32 &&
+          ctx.OutputVar("Out")->IsType<framework::LoDTensor>()) {
+        bool can_use_mkldnn_sum = true;
+        for (size_t idx = 0; idx < x_vars.size(); ++idx) {
+          if (!x_vars[idx]->IsType<LoDTensor>()) {
+            can_use_mkldnn_sum = false;
+          }
+        }
+        if (can_use_mkldnn_sum) {
```
wojtuss: The `can_use_mkldnn_sum` value obtained using the above `for` loop could be calculated in a little bit simpler way, using a condition like this:
```
if (std::all_of(x_vars.begin(), x_vars.end(), [](const Variable* v){return v->IsType<LoDTensor>();})) {
```
Otherwise the `break;` instruction should be used after `can_us_mkldnn_sum = false;` instruction, to exit the loop when the result is already known.

lidanqing-vv: Thank you, you are right. I changed to all_of version now.

paddle/fluid/operators/sum_op.cc
```diff
@@ -141,6 +134,21 @@ class SumOp : public framework::OperatorWithKernel {
...
                         "Sum operator should have at least one tensor");
 
+#ifdef PADDLE_WITH_MKLDNN
+      if (library == framework::LibraryType::kPlain &&
+          platform::CanMKLDNNBeUsed(ctx) &&
+          static_cast<framework::proto::VarType::Type>(dtype) ==
+              framework::proto::VarType::FP32 &&
+          ctx.OutputVar("Out")->IsType<framework::LoDTensor>()) {
+        if (std::all_of(x_vars.begin(), x_vars.end(),
+                        [](Variable* v) { return v->IsType<LoDTensor>(); })) {
```
wojtuss: please, do add `const` in front of the lambda's argument declaration

wojtuss (commented):
LGTM

wojtuss (approved):
LGTM

PR #22292, opened 2020-01-15
add partial_sum op in contrib

ForFishes:
Add partial_sum op implementation and the unit test codes of it in contrib. Different from sum op, this op can sum the vars by specifying the initial position(start_index) and length(length). This op is used to add 'vars' up in part.This OP exists in contrib, which means that it is not shown to the public.


lanxianghit (dismissed):

lanxianghit (approved):

Boyan-Liu (approved):

paddle/fluid/operators/partial_sum_op.cc
```diff
@@ -0,0 +1,207 @@
...
+        "(bool, default false) Indicates if MKL-DNN kernel will be used")
+        .SetDefault(false);
+    AddAttr<int>("start_index", "The start index of tensor wanted to be added.")
+        .SetDefault(0);
+    AddAttr<int>("length", "The length of tensor wanted to be added.")
+        .SetDefault(-1);
+    AddComment(R"DOC(
+PartialSum Operator.
+This Op can sum the vars by specifying the initial position(start_index) and length(length). 
+This OP exists in contrib, which means that it is not shown to the public.
```
Boyan-Liu, line 154: 记得后面改书写错误

ForFishes: OK

gavin1332 (approved):
LGTM

PR #22308, opened 2020-01-15
delete generate_op_use_grad_op_desc_maker_spec.py

sneaxiy:
`generate_op_use_grad_op_desc_maker_spec.py` is not needed any more, since PR #22300 has change the CI checking method.


zhiqiu (approved):
LGTM

PR #22309, opened 2020-01-15
Make assign op support LoDTensorArray and modify while_loop API

gfwm2013:
assign能够支持LoDTensorArray，并且在while_loop中loop_vars支持嵌套tuple或list


paddle/fluid/framework/operator.cc
```diff
@@ -1245,6 +1245,9 @@ void OperatorWithKernel::ParseInputDataType(
         t = &var->Get<LoDTensor>();
       } else if (var->IsType<SelectedRows>()) {
         t = &(var->Get<SelectedRows>().value());
+      } else if (var->IsType<LoDTensorArray>()) {
+        auto t_arr = var->Get<LoDTensorArray>();
+        t = &(t_arr[0]);
```
zhhsplendid: What if the size of LoDTensorArray is empty? Please handle that exception here.

zhhsplendid (commented):
Add the test covering the case

python/paddle/fluid/tests/unittests/test_while_loop_op.py
```diff
@@ -189,63 +175,84 @@ def body(i, x):
...
-            out = layers.while_loop(external_cond, external_body, [i, x, y])
-
-            mean = layers.mean(out[2])
-            append_backward(mean)
-
-        place = fluid.CUDAPlace(0) if core.is_compiled_with_cuda(
-        ) else fluid.CPUPlace()
-        exe = fluid.Executor(place)
+            i = layers.zeros(shape=[1], dtype='int64')
+            i.stop_gradient = True
```
zhhsplendid, line 215: You don't have to put unnecessary stop_gradient, since these are test code we would like this PR in release 1.7. I will approve this PR and please change them in a follow up PR.

zhhsplendid (approved):

PR #22348, opened 2020-01-17
use function instead of lambda, test=develop

zhiqiu:
Polish #21569.

For each operator, we define a specific function for pybind in dygraph mode, called `imperative_xx`, which reduces the size of `BindOpFunctions`.

The generated file is [here](https://gist.github.com/zhiqiu/e054da44f2731bd3c02bd88e2b5e4153).

### DIFF
--- a/paddle/fluid/pybind/op_function_generator.cc
+++ b/paddle/fluid/pybind/op_function_generator.cc
@@ -29,7 +29,7 @@ const char* OUT_INITIALIZER_TEMPLATE =
 
 const char* OP_FUNCTION_TEMPLATE =
 R"(
-imperative::NameVarBaseMap %s(const imperative::NameVarBaseMap& ins, const framework::AttributeMap& attrs, 
+inline imperative::NameVarBaseMap %s(const imperative::NameVarBaseMap& ins, const framework::AttributeMap& attrs, 
   imperative::NameVarBaseMap outs, const std::map<std::string, size_t>& out_nums)
 {
   auto tracer = imperative::GetCurrentTracer();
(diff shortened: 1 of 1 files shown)


paddle/fluid/pybind/op_function_generator.cc
```diff
@@ -23,45 +23,47 @@
 #include "paddle/fluid/pybind/pybind.h"
 #include "paddle/fluid/string/string_helper.h"
 
+// clang-format off
```
sneaxiy, line 26: Why disable clang-format?

paddle/fluid/pybind/op_function_generator.cc
```diff
@@ -23,45 +23,47 @@
...
-        outs = std::move(outs_);
-      } else {
-        for (auto &pair : out_nums) {
-          for (size_t i = 0; i < pair.second; i ++) {
-            auto var_base_name = tracer->GenerateUniqueName();
-            auto out = new imperative::VarBase(var_base_name);
-            outs[pair.first].emplace_back(std::shared_ptr<imperative::VarBase>(out));
-          }
+R"(
+imperative::NameVarBaseMap %s(const imperative::NameVarBaseMap& ins, const framework::AttributeMap& attrs,
```
sneaxiy: Add `inline` to avoid function redefinition when this header file is included directly or indirectly twice.

paddle/fluid/pybind/op_function_generator.cc
zhiqiu: clang-format changes the format of `template string` and the format of generated functions, which is not wanted.

paddle/fluid/pybind/op_function_generator.cc
zhiqiu: Done, thanks.

sneaxiy (approved):

PR #22365, opened 2020-01-19
Fix static graph dataloader early reset bug

sneaxiy:
In static graph mode, non-iterable `DataLoader` cannot work well when `reset` is called when the thread does not ends. Because the `reset` method would close the queue first and reopen it. If the queue is not closed when calling `reset`, the thread which pushes data into the queue may hang.

This PR fixes this bug.


chenwhql (approved):
LGTM

PR #22386, opened 2020-01-21
refine convert type for numpy type

wangchaochaohu:
https://github.com/PaddlePaddle/Paddle/issues/22383 in order to solve convert type error when type is numpy type, add this.


heavengate (approved):
LGTM

PR #20712, opened 2019-10-18
fix_depthwise_conv_cudnn

ceci3:
when groups > 1, SearchAlgorithm find algo is CUDNN_CONVOLUTION_FWD_ALGO_WINOGRAD_NONFUSED, but this kind of algorithm is unstable in groups convolution computation, so change the algorithm to CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM manually.

fix https://github.com/PaddlePaddle/Paddle/issues/18939

test environment Paddle 1.6, P40, cuda9.0 cudnn 7.5
test code:

```python
import paddle.fluid as fluid
import time

def test(place, data, use_cudnn):
    input = fluid.layers.create_global_var(dtype='float32', shape=[5, 384, 28, 28], name='data', persistable=True, value=1.0)
    result = fluid.layers.conv2d(input, bias_attr=False, param_attr="depthwise_kernel",
        num_filters=384, filter_size=[5, 5], stride=[1, 1], groups=384, use_cudnn=use_cudnn)

    exe = fluid.Executor(place)
    exe.run(fluid.default_startup_program())

    np_input = fluid.global_scope().find_var('data').get_tensor()
    np_input.set(data, place)

    fluid.io.load_params(executor=exe, dirname='test_model', main_program=fluid.default_main_program())

    for i in range(10):
        exe.run(fluid.default_main_program())

    s_time = time.time()
    for i in range(10):
        exe.run(fluid.default_main_program())
    print("time: ", (time.time() - s_time) / float(10))

import numpy
numpy.random.seed(13)
data = numpy.random.rand(5, 384, 28, 28).astype('float32')
test(fluid.CUDAPlace(0), data, True)
#test(fluid.CUDAPlace(0), data, False)
```

                     |  after(GEMM)   |before(WINOGRAD)
    use_cudnn=True   |  0.00147512    |  0.0144635
    use_cudnn=False  |  0.00173819    |  0.00183482

PR #22389, opened 2020-01-21
Add benchmark flag for conv_transpose

ceci3:
1. the reason about the change of  depthwise conv please reference: https://github.com/PaddlePaddle/Paddle/pull/20712
   after the change of depthwise conv:
 
        MobileNetV1(with depthwise conv)  |  use_cudnn=False |   use_cudnn=True
        accuracy                          |   0.7099/08968   |    0.71150/0.89904
 
2. fixed algorithm to 1(algorithm 1 is stable when conv forward, data backward and filter backward)if set FLAGS_cudnn_deterministic=True  in conv_transpose when use_cudnn=True


qingqing01 (approved):
LGTM

PR #22391, opened 2020-01-22
Add api to clear intermediate tensors in AnalysisPredictor

cryoco:
On some devices, memory/graphic-memory is limited. Instead of running paddle inference parallelly, we have to load, analyze, and run them one after another(with each one's resources released after running) on these devices to avoid OOM. In such situations, loading and analysis phases become bottleneck of latency. 
So we add an API to clear intermediate tensors, enabling AnalysisPredictor to load and analyze all models first, and run inference one by one(with each one's intermediate tensors released to graphic memory pool by calling ClearIntermediateTensor()) to cut the overhead mentioned above.

### DIFF
--- a/paddle/fluid/inference/api/analysis_predictor.cc
+++ b/paddle/fluid/inference/api/analysis_predictor.cc
@@ -749,6 +749,24 @@ bool AnalysisPredictor::LoadParameters() {
   return true;
 }
 
+void AnalysisPredictor::ClearIntermediateTensor() {
+  PADDLE_ENFORCE_NOT_NULL(inference_program_.get(),
+                          platform::errors::PreconditionNotMet(
+                              "The inference program should be loaded first."));
+  const auto &global_block = inference_program_->MutableBlock(0);
+  for (auto *var : global_block->AllVars()) {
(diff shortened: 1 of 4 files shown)


paddle/fluid/inference/api/analysis_predictor.cc
```diff
@@ -749,6 +749,24 @@ bool AnalysisPredictor::LoadParameters() {
...
+                          platform::errors::PreconditionNotMet(
+                              "The inference program should be loaded first."));
+  const auto &global_block = inference_program_->MutableBlock(0);
+  for (auto *var : global_block->AllVars()) {
+    if (!IsPersistable(var)) {
+      const std::string name = var->Name();
+      auto *variable = executor_->scope()->FindVar(name);
+      if (variable != nullptr && name != "feed" && name != "fetch") {
+        VLOG(3) << "Clear Intermediate Tensor: " << name;
+        auto *t = variable->GetMutable<framework::LoDTensor>();
```
NHZlX, line 764: Should add an another check: The variable can be deleted under the condition`variable.IsType<framework::LoDTensor>()`.

PR #22394, opened 2020-01-22
Update the precision of pad, pad2d, pad_constant_like's unit tests from fp32 to fp64

FrostML:
According to [单测规范](https://github.com/PaddlePaddle/Paddle/wiki/Upgrade-OP-Precision-to-Float64), it's necessary to update the precision of pad, pad2d, pad_constant_like's unit tests from fp32 to fp64. 
What's more, fix the data type error caused by pad_value, which is float, but got by double.


juncaipeng (approved):
LGTM

Please update the pr in

Issue #21219, opened 2019-11-18
Feature Request: Add Mish Activation


digantamisra98:
Mish is a new novel activation function proposed in this [paper](https://arxiv.org/abs/1908.08681).
It has shown promising results so far and has been adopted in several packages including:

- [TensorFlow-Addons](https://github.com/tensorflow/addons/tree/master/tensorflow_addons/activations)
- [SpaCy (Tok2Vec Layer)](https://github.com/explosion/spaCy)
- [Thinc - SpaCy's official NLP based ML library](https://github.com/explosion/thinc/releases/tag/v7.3.0)
- [Echo AI](https://github.com/digantamisra98/Echo)
- [CNTKX - Extension of Microsoft's CNTK](https://github.com/delzac/cntkx)
- [FastAI-Dev](https://github.com/fastai/fastai_dev/blob/0f613ba3205990c83de9dba0c8798a9eec5452ce/dev/local/layers.py#L441)
- [Darknet](https://github.com/AlexeyAB/darknet/commit/bf8ea4183dc265ac17f7c9d939dc815269f0a213)
- [Yolov3](https://github.com/ultralytics/yolov3/commit/444a9f7099d4ff1aef12783704e3df9a8c3aa4b3)
- [BeeDNN - Library in C++](https://github.com/edeforas/BeeDNN)
- [Gen-EfficientNet-PyTorch](https://github.com/rwightman/gen-efficientnet-pytorch)
- [dnet](https://github.com/umangjpatel/dnet/blob/master/nn/activations.py)

All benchmarks, analysis and links to official package implementations can be found in this [repository](https://github.com/digantamisra98/Mish) 

It would be nice to have Mish as an option within the activation function group.

This is the comparison of Mish with other conventional activation functions in a SEResNet-50 for CIFAR-10:  (Better accuracy and faster than GELU)
![se50_1](https://user-images.githubusercontent.com/34192716/69002745-0de37980-091b-11ea-87da-ac8d17c79e07.png)

parap1uie-s:
Try this:

```
    def mish(self, x):
        return x * FL.tanh(FL.softplus(x))
```

digantamisra98:
@parap1uie-s Thanks. That's really helpful. Hoping that it can be added to Paddle.

zhhsplendid:
[...]

PR #22401, opened 2020-01-26
Adding Mish Activation

thebhatman:
This PR implements the Mish Activation and resolves https://github.com/PaddlePaddle/Paddle/issues/21219


thebhatman:
I am not sure why the CI builds are still yellow and loading... Also do I have to do something in python folder to allow for python imports of Mish?

luotao1:
https://www.paddlepaddle.org.cn/documentation/docs/en/advanced_usage/development/contribute_to_paddle/local_dev_guide_en.html#commit

Sorry for the late reply, please add `test=develop` in commit information to trigger the CI.

paddle/fluid/inference/anakin/convert/activation.cc
```diff
@@ -61,4 +61,5 @@ void ActivationOpConverter<TargetT, PrecisionT>::operator()(
 REGISTER_ANAKIN_OP_CONVERTER(sigmoid, SigmoidOpConverter);
 REGISTER_ANAKIN_OP_CONVERTER(tanh, TanhOpConverter);
 REGISTER_ANAKIN_OP_CONVERTER(swish, SwishOpConverter);
+REGISTER_ANAKIN_OP_CONVERTER(mish, MishOpConverter);
```
luotao1: You don't need to add mish in Anakin, since it doesn't support it.

thebhatman: So shall I remove Mish from Anakin completely in all files that come under anakin?

thebhatman: Done.

luotao1:
> Also do I have to do something in python folder to allow for python imports of Mish

Yes, you should add python wrapper in `python/paddle/fluid/layers/nn.py`, like some activations_op such as `relu6`, `elu` etc.

Besides, you should add unit-tests in `python/paddle/fluid/tests/unittests/test_activation_op.py`

luotao1:
What's the python wrapper of mish？If it is `paddle.fluid.layers.mish(x, name=None)`, you could follow the way `paddle.fluid.layers.abs(x, name=None)` does, and don't need to add python wrapper in `python/paddle/fluid/layers/nn.py`

thebhatman:
I have added the python wrapper for Mish now.

thebhatman:
Can I add Mish in tensorrt/convert and operators, and the python wrappers together in one PR? That would basically be adding the whole Mish module in one PR.

luotao1:
Yes, you can, but the review process will be longer.

PR #22404, opened 2020-01-28
[DNNL] Fix accuracy in INT8 FC

Sand3r-:
PR introduces changes that enable int8 FC to correctly perform computations in ernie topology.

### Terminology used in description:
`nhwc`-like or `nchw`-like means any format that is descendant of these formats, such as ncw or nwc for 3-dim data blobs, nhwc, nchw for 4-dims and so on.

#### Background:
The `nhwc` data layout was formerly used in all int8 computations so far, and quantize op had it hardcoded to reorder to nhwc-like format during quantizaiton.

#### Problem:
Unfortunately, due to how this format is laid out in memory, it was not possible to correctly use this format when fully-connected op used `in_num_col_dims` parameter to merge some of the first dimensions, since it assumes an nchw-like format.

#### Solution:
To counter this, the PR makes it possible to set to which format (or in fact layout) should quantize reorder to during quantizaiton. Possible options being either `NCHW` or `NCHW` by setting its `output_format` parameter. 

To further utilise this new feature, `cpu_quantize_pass` now accepts an optional attribute `data_layout` which allows to set desired data format for all quantizes that will be added to the graph.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
@@ -63,7 +63,7 @@ void FCMKLDNNPass::ApplyImpl(ir::Graph* graph) const {
         dim_num == 4 ? (dims[width_axis] == 1 && dims[height_axis] == 1) : true;
     if (!are_dims_supported || !is_size_supported) {
       VLOG(3) << "Do not enable FC MKL-DNN for dimensions different than"
-                 "2, 3 & 4, and when both width and height are equal to one";
+                 "2, 3 & 4, or when width or height is different than one.";
       return;
     }
     desc->SetAttr("use_mkldnn", true);
--- a/paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
@@ -56,13 +56,14 @@ void FCMKLDNNPass::ApplyImpl(ir::Graph* graph) const {
     OpDesc* desc = fc->Op();
     auto dims = fc->inputs[0]->Var()->GetShape();
     auto dim_num = dims.size();
-    bool are_dims_supported = dim_num == 2 || dim_num == 4;
+    bool are_dims_supported = dim_num >= 2 && dim_num <= 4;
     constexpr size_t height_axis = 2;
     constexpr size_t width_axis = 3;
     bool is_size_supported =
         dim_num == 4 ? (dims[width_axis] == 1 && dims[height_axis] == 1) : true;
     if (!are_dims_supported || !is_size_supported) {
-      VLOG(3) << "Do not enable FC MKL-DNN for dimensions different than 2 & 4";
+      VLOG(3)
+          << "Do not enable FC MKL-DNN for dimensions different than 2, 3 & 4";
       VLOG(3) << "Or when width and height are different than one";
       return;
     }
--- a/paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
@@ -62,9 +62,8 @@ void FCMKLDNNPass::ApplyImpl(ir::Graph* graph) const {
     bool is_size_supported =
         dim_num == 4 ? (dims[width_axis] == 1 && dims[height_axis] == 1) : true;
     if (!are_dims_supported || !is_size_supported) {
-      VLOG(3)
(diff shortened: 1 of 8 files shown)


paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
```diff
@@ -56,13 +56,14 @@ void FCMKLDNNPass::ApplyImpl(ir::Graph* graph) const {
...
+    bool are_dims_supported = dim_num >= 2 && dim_num <= 4;
     constexpr size_t height_axis = 2;
     constexpr size_t width_axis = 3;
     bool is_size_supported =
         dim_num == 4 ? (dims[width_axis] == 1 && dims[height_axis] == 1) : true;
     if (!are_dims_supported || !is_size_supported) {
-      VLOG(3) << "Do not enable FC MKL-DNN for dimensions different than 2 & 4";
+      VLOG(3)
+          << "Do not enable FC MKL-DNN for dimensions different than 2, 3 & 4";
       VLOG(3) << "Or when width and height are different than one";
```
wojtuss: What about joining the two messages into one?
Also:
`width and height are` -> `width or height is`.

paddle/fluid/operators/fc_op.cc
```diff
@@ -69,11 +69,13 @@ class FCOp : public framework::OperatorWithKernel {
                         activation_type.c_str());
     }
     if (ctx->Attrs().Get<bool>("use_mkldnn")) {
-      PADDLE_ENFORCE_EQ(in_dims.size() == 2 || in_dims.size() == 4, true,
-                        "Fully Connected input should be 2-D or 4-D tensor.");
+      PADDLE_ENFORCE_EQ(
+          in_dims.size() >= 2 || in_dims.size() <= 4, true,
```
wojtuss: `||` -> `&&`

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -54,6 +54,19 @@ class FCPrimitiveFactory {
...
+      PADDLE_ENFORCE_EQ(
+          input->dims().size(), 3,
+          platform::errors::Unimplemented(
+              "DNNL FC only supports in_num_col_dims equal to 2 when "
+              "3 dim input is provided."));
+    } else if (in_col_dims > 2) {
+      PADDLE_THROW(platform::errors::Unimplemented(
+          "DNNL FC doesn't support in_num_col_dims paramter to be higher than "
+          "2."));
+    }
```
wojtuss: What about structuring this part like this:
```
PADDLE_ENFORCE_LE(in_col_dims, 2, ...);
if (in_col_dims == 2) {
  PADDLE_ENFORCE_EQ(input->dims().size(), 3, ...);
}
```
?

paddle/fluid/inference/analysis/argument.h
```diff
@@ -169,6 +169,9 @@ struct Argument {
 
   // Scales for variables to be quantized
   DECL_ARGUMENT_FIELD(quant_var_scales, QuantVarScales, VarQuantScale);
+
+  // Destination data layout of quantization
+  DECL_ARGUMENT_FIELD(quantization_data_layout, DataLayout, std::string);
```
wojtuss: You do not have to declare any field here for an optional pass attribute.

paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
Sand3r-: I just wanted to separate them into  two lines as I thought It'd be cleaner to do that, but can sure merge them. Btw - should've written `when both width and height are equal to one`. Will do in an upcoming commit.

paddle/fluid/inference/analysis/argument.h
Sand3r-: Alrighty, didn't know that. Will remove the attribute.

paddle/fluid/operators/fc_op.cc
Sand3r-: Nice catch! Will fix in a moment.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
Sand3r-: Sure, that would definitely make it look much clearer.

paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
Sand3r-: Is it fine, to leave it the way I did it right now?

wojtuss (approved):
LGTM

luotao1:
Please cherry-pick into release/1.7

PR #22424, opened 2020-02-03
Optimize again for Ernie GPU Inference 

NHZlX:
1. optim the multihead fuse pass( fuse the three fc to multihead) and also update the multihead op.
2. Add embedding eltwise layernorm pass and embedding eltwise layernorm op.


NHZlX:
Split this pr to ：

https://github.com/PaddlePaddle/Paddle/pull/22486 

https://github.com/PaddlePaddle/Paddle/pull/22494

PR #22426, opened 2020-02-03
Add MKL-DNN GELU activation

grygielski:
This PR adds MKL-DNN support of GELU activation, however formula used in MKL-DNN kernel is a bit different from used in native Paddle implementation. 

**Paddle uses following formula:**
`out(x) = 0.5 * x * (1 + erf(x / sqrt(2)))`
This exact equation for GELU uses ERF function which can't be represented using elementary functions. Implementations of ERF are mostly based on Maclaurin series approximation.

**MKL-DNN uses following formula:**
`out(x) = 0.5 * x * (1 + tanh(sqrt(2/pi) * (x + 0.044715 * x^3)))`
This approximation has been proposed in original GELU paper (https://arxiv.org/pdf/1606.08415.pdf) and consists of only elementary functions.

**Plot comparison:**
image
As it's not possible to distinguish these two implementations from each other on general graph, I plotted absolute difference between them:
image

**Maximum absolute error between 2 implementations:** 
`0.00047326088`

### DIFF
--- a/paddle/fluid/operators/gelu_op.cc
+++ b/paddle/fluid/operators/gelu_op.cc
@@ -44,8 +44,19 @@ class GeluOp : public framework::OperatorWithKernel {
  protected:
   framework::OpKernelType GetExpectedKernelType(
       const framework::ExecutionContext &ctx) const override {
+    framework::LibraryType library{framework::LibraryType::kPlain};
+    framework::DataLayout layout = framework::DataLayout::kAnyLayout;
+#ifdef PADDLE_WITH_MKLDNN
+    auto it = this->Attrs().find("use_mkldnn");
+    if (library == framework::LibraryType::kPlain &&
+        it != this->Attrs().end() && platform::CanMKLDNNBeUsed(ctx)) {
+      library = framework::LibraryType::kMKLDNN;
+      layout = framework::DataLayout::kMKLDNN;
+    }
+#endif
     return framework::OpKernelType(
-        OperatorWithKernel::IndicateVarDataType(ctx, "X"), ctx.GetPlace());
+        OperatorWithKernel::IndicateVarDataType(ctx, "X"), ctx.GetPlace(),
+        layout, library);
   }
 };
 
@@ -73,8 +84,19 @@ class GeluGradOp : public framework::OperatorWithKernel {
  protected:
   framework::OpKernelType GetExpectedKernelType(
       const framework::ExecutionContext &ctx) const override {
+    framework::LibraryType library{framework::LibraryType::kPlain};
+    framework::DataLayout layout = framework::DataLayout::kAnyLayout;
+#ifdef PADDLE_WITH_MKLDNN
+    auto it = this->Attrs().find("use_mkldnn");
+    if (library == framework::LibraryType::kPlain &&
+        it != this->Attrs().end() && platform::CanMKLDNNBeUsed(ctx)) {
+      library = framework::LibraryType::kMKLDNN;
+      layout = framework::DataLayout::kMKLDNN;
+    }
+#endif
     return framework::OpKernelType(
-        OperatorWithKernel::IndicateVarDataType(ctx, "X"), ctx.GetPlace());
+        OperatorWithKernel::IndicateVarDataType(ctx, "X"), ctx.GetPlace(),
+        layout, library);
   }
 };
 
--- a/python/paddle/fluid/tests/unittests/mkldnn/test_activation_mkldnn_op.py
+++ b/python/paddle/fluid/tests/unittests/mkldnn/test_activation_mkldnn_op.py
@@ -19,7 +19,8 @@ import numpy as np
 from scipy.special import expit
 import paddle.fluid.core as core
 from paddle.fluid.tests.unittests.op_test import OpTest
-from paddle.fluid.tests.unittests.test_activation_op import TestRelu, TestTanh, TestSqrt, TestAbs, TestLeakyRelu, TestSwish
+from paddle.fluid.tests.unittests.test_activation_op import TestActivation, TestRelu, TestTanh, TestSqrt, TestAbs, TestLeakyRelu, TestSwish
+from paddle.fluid.tests.unittests.test_gelu_op import gelu
 from mkldnn_op_test import check_if_mkldnn_primitives_exist_in_bwd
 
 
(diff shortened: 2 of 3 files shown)


grygielski:
@luotao1 Could you advice on failing CI? This is because of dygraph=false part of UTs but I've just copied it from other activations following wangzhongpu comment `# TODO(wangzhongpu): support mkldnn op in dygraph mode`

luotao1:
@grygielski Don't be care about it. @JepsonWong When will you support mkldnn op in dygraph mode? Or could just skip mkldnn op in the monitor?

jczaja (dismissed):
LGTM

luotao1:
@grygielski What does the first graph mean? It has two lines (gelu_erf and gelu_tanh) in the left-top of the graph, but there is only one line in the graph.

jczaja:
@luotao1 blue and orange lines look the same at that scale of chart. That is why @grygielski  presented another plot with plotted actual differences.

luotao1:
Oh, I got it, the blue and orange lines are the same, and it becomes a brown line in the first graph

grygielski:
Exactly, sorry for inconvenience, I've added some alpha to make it kind of visible that these are 2 separate lines but it didn't work well. However, second graph is much more explanatory.

luotao1:
@grygielski @jczaja Thanks for the contribution of GELU activation. To make the formula unified, @lanxianghit  decide to add an attribute in GELU API to distinguish these two formulae like TF:
image

@iclementine will implement the approximation formula by Eigen and update the GELU API ASAP. After @iclementine merges his PR, this MKLDNN PR could be merged.

python/paddle/fluid/tests/unittests/mkldnn/test_activation_mkldnn_op.py
```diff
@@ -58,6 +59,29 @@ def test_check_grad(self):
             ['X'], 'Out', max_relative_error=0.007, check_dygraph=False)
 
 
+class TestMKLDNNGeluDim2(TestGelu):
+    def setUp(self):
+        super(TestMKLDNNGeluDim2, self).setUp()
+
+        x = np.random.uniform(-1, 1, [11, 17]).astype(np.float32)
+        out = 0.5 * x * (1.0 + np.tanh(
```
wojtuss: What about defining the formula as a function and reusing it below?

grygielski: Great idea, but it would be even better to add it in general `test_activation_op.py` and include in UTs when `approximate=True` will be implemented as a part of an API.

GaoWei8:
The PR  (#22961) of updating the GELU API is processing.

GaoWei8:
The PR (#22961 ) of updating the GELU API is merged.

grygielski:
@GaoWei8 Do I have to change anything in this PR after GELU API change?

iclementine:
> @GaoWei8 Do I have to change anything in this PR after GELU API change?

Now gelu_op has 2 different behaviors depending on the boolean attr `approximate`. So it is better to support these 2 behaviors in a `Kernel`. (I've found that latest version of dnnl has two functors, namely `gelu_tanh` and `gelu_erf`. May be updating dnnl to use these two fucntors should be considered.)

One more thing to be noticed is that since `MKLDNNActivationKernel` is a template class with a template parameter `Functor`, so may be define a Kernel for gelu may be a better solution for it to use two different functors.

图片

https://github.com/PaddlePaddle/Paddle/blob/5257530492b639ceaffd4cacf4dccb862c9b1f47/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc#L31

grygielski:
@iclementine Thanks for the reply. The thing is that MKL-DNN doesn't have kernel for "exact" GELU as MKL-DNN GELU kernel uses formula with tanh anyway so only approximate solution is possible.

iclementine:
> @iclementine Thanks for the reply. The thing is that MKL-DNN doesn't have kernel for "exact" GELU as MKL-DNN GELU kernel uses formula with tanh anyway so only approximate solution is possible.

Thank you for your replay.

But [dev_guide_eltwise](https://intel.github.io/mkl-dnn/dev_guide_eltwise.html) states that `gelu_erf` is available. Does it also use the formula with tanh-approximation?

grygielski:
You are right, they've added it very recently, it wasn't available when I created this PR. I'll take a look at it and change my PR in following days, thanks for pointing that out!

paddle/fluid/operators/gelu_op.cc
```diff
@@ -73,8 +84,19 @@ class GeluGradOp : public framework::OperatorWithKernel {
...
+    auto it = this->Attrs().find("use_mkldnn");
+    if (library == framework::LibraryType::kPlain &&
+        it != this->Attrs().end() && platform::CanMKLDNNBeUsed(ctx)) {
+      library = framework::LibraryType::kMKLDNN;
+      layout = framework::DataLayout::kMKLDNN;
+    }
+#endif
     return framework::OpKernelType(
-        OperatorWithKernel::IndicateVarDataType(ctx, "X"), ctx.GetPlace());
+        OperatorWithKernel::IndicateVarDataType(ctx, "X"), ctx.GetPlace(),
```
grygielski, line 98: @iclementine Just a quick question I came up with during developing this PR. Shouldn't it be `IndicateVarDataType(ctx, "Out")` since Grad operator depends on `Out` variable?

jczaja (approved):
LGTM

grygielski:
For current develop branch (Eigen version of GELU) on my machine (i9-7940X):
|BERT model | Develop | DNNL GELU  | Increase |
| ------------- | ------------- | ------------  | -- |
| FPS  | 266.104  | 316.219  | 18.8%|
| GELU % of the model  | 20.9  | 5.92  | -71.6%|

grygielski:
@luotao1 Here `PR_Windows_CI` also randomly failed, please restart if you can.

grygielski:
Unfortunately, there is another random fail during building on `PR_CI_Coverage`

luotao1:
I have rerun it.

paddle/fluid/operators/gelu_op.cc
iclementine: In this case, GeluGradOpMaker does not have `Out` as input, only `X` and `DOut`. So we use `X` to determine the data type of `DX`. It's also okay to use `Dout`.

https://github.com/PaddlePaddle/Paddle/blob/dc4bc2624a2161887d941e32e7fb5fe1ca499e4f/paddle/fluid/operators/gelu_op.cc#L145

danqing-li:
@luotao1  Could you please review this PR? We would like to merge it in 2.0. Thanks~

lidanqing-vv:
@luotao1  This PR could be merged now, the failed PR_CI_Coverage is on Teamcity, and it was rerun on efficiency cloud and passed already. Thanks!

luotao1 (approved):
LGTM

PR #22403, opened 2020-01-27
Sigmoid bug fix

thebhatman:
Sigmoid activation was calling the wrong CUDNN functor. This PR fixes the bug and ensures it calls the correct Sigmoid Functor and Sigmoid Grad Functor.

PR #22439, opened 2020-02-04
fix sigmoid cudnn bug

luotao1:
Sigmoid activation was calling the wrong CUDNN functor. This PR fixes the bug and ensures it calls the correct Sigmoid Functor and Sigmoid Grad Functor.

Same with #22403


lanxianghit (approved):
LGTM

PR #22447, opened 2020-02-04
cherry pick 1.7 , fix copy table, add hdfs ls retry, fix save inference

xjqbest:
* fix copy table bug of lost some feasign
* add hdfs ls retry time and sleep time, fix save inference
* test=develop


123malin (approved):
LGTM

PR #22455, opened 2020-02-05
update readme of imdb training demo

mapingshuo:
The `IMDB training demo` is an example of training with fluid inference lib. The script for build fluid lib is added to the REDAME.txt to complete the document.


guru4elephant (approved):
LGTM

PR #22470, opened 2020-02-06
Remove TensorRT lib from Paddle Inference libs

cryoco:
At the require of NVIDIA that TensorRT lib should only be downloaded from [NVIDIA official website](https://developer.nvidia.com/nvidia-tensorrt-download), Paddle Inference will not bundle TRT lib, starting from version 1.7.0. So we remove the copying of TRT lib during building Paddle Inference libs.


NHZlX (approved):
LGTM

PR #22474, opened 2020-02-06
[feature] prune program by feed and fetch_list

zhiqiu:
This PR support pruning the program by feed and fetch_list.
Developers can enable this feature by setting `use_prune=True` in `Executor.run()`.
Related #19589


python/paddle/fluid/executor.py
```diff
@@ -595,6 +630,156 @@ def _fetch_data(self, fetch_list, fetch_var_name, scope):
...
+                raise TypeError(
+                    "The item in fetch_list should be str, variable or optimize_op, but recieved %s.",
+                    type(item))
+
+        for item in fetch_list:
+            if isinstance(item, list):
+                for i in item:
+                    _get_targets(_optimize_ops, _fetch_list, i)
+            elif isinstance(item, tuple):
+                for i in item[0]:
```
sneaxiy, line 674: I know this `tuple` if-else branch is to handle `(optimizer_ops, param_and_grads)` cases, but other SWE may not know why tuple and list branches are not the same. Try to add some comments on it.

BTW it is still a bit strange. What if the users parse tuple of optimizer ops? Is there other better cases to distinguish these 2 cases?

sneaxiy (dismissed):
TODO: find a better way to distinguish tuple of optimizer ops and tuple of (optimizer ops, param_grads).

zhiqiu:
document update:

1. `Optimizer.mimimize()` API

image

2. `Executor.run() `API
image

zhiqiu:
- Optimizer.mimimize() API

image

- Executor.run() API
image

image

python/paddle/fluid/executor.py
zhiqiu: I add some comments on it, and add a TODO for better implementation in future.

sneaxiy (approved):

swtkiwi (approved):
LGTM

lanxianghit (approved):

PR #22478, opened 2020-02-06
Fix the integer overflow problem of sequence2batch

ZHUI:
In the /paddle/fluid/operators/math/sequence2batch.h#L122.

The type of `max_seqlen` is int. If we carefully set `max_seqlen= 2147483647`, then `max_seqlen + 1` will be `-2147483648 `. Thus may cause  integer overflow problem.

image

Please reference more deatails in http://newicafe.baidu.com/issue/77812981/show?cid=5&spaceId=23205


ZHUI:
Wrong account for CLA.

PR #22492, opened 2020-02-07
Post_training_quantization support set quant 8/16 bits

juncaipeng:
* Post_training_quantization support set quant bits, 8bits or 16bits
* update the inputs of Post_training_quantization


wzzju (approved):
LGTM.

Issue #21493, opened 2019-12-02
[MKL-DNN] Analyze if it is possible to keep one copy of params not two


jczaja:
MKL-DNN is often keeping params (weights, bias) in its own blocked format. Essentaily it means that  each operators that is having params is keeping two sets of weights (one paddle Tensors and then converted(reordered) weights into MKL-DNN format). This may lead to increased memory cosnumption. Goal of this is task is to analyze if it is possible to keep only one set of weights&biases not two. For example if it is possible to release paddle weights in inference and keep only mkl-dnn weights.

willthefrog:
please take a look @luotao1

jczaja:
I'm working on that one. Very early results are promising e.g. It should be possible to reduce memory usage of MKL-DNN integration quite close to level of Paddle CPU operators.

jczaja:
@luotao1 , @willthefrog 
It is possible to reduce peak memory consumption of DNNL(mkl-dnn) workloads, by overwritting original
filter's params with DNNL ones. 

Before(develop):
![dnnl_memory](https://user-images.githubusercontent.com/15085062/74055432-51acc300-49e0-11ea-8d1e-81a3b096666c.png)

After(#22505):
![dnnl_memory_optimizer](https://user-images.githubusercontent.com/15085062/74055533-79039000-49e0-11ea-93f5-e451b378621b.png)

Notes:
1) Inference only
2) Currently works for convolutions in FP32 DNNL execution . Should be extended for Fully connected
    params to get very close to Paddle CPU memory consumption
3) Problem is that although I overwrite original params , I do not know how to change outer scope's filter's Tensor *layout* and *data_format* as each of Tensors (one from outer/original scope and other from
execution scope are having they own copy of *layout* and *data_format*. See Question 1
4) Int8 support require more work (need to adjust quantization NCHW16C) I need to get an answer to Question 1
5) Current implementation #22505 is only modifying filter data when *memory_size* of DNNL params of specific op is matching original filter's memory_size
6) As for savings, currently only C-API MobiletNetv1 was tested (more tests are running). Peak Memory Consumption:
* Paddle CPU : ~550 MB
* Before MKL-DNN: ~640MB
* After MKL-DNN: ~580 MB

Questions:
1) Data (allocation) is shared among execution scope and original scope. So it can be overwritten. But How can I change original scope's Tensor's atribute eg. *layout* and *data_format* ?

luotao1:
[...]

PR #22505, opened 2020-02-07
[DNNL] Draft of DNNL Memory optimizer - Part 1

jczaja:
This is first PR trying to reduce MKL-DNN memory consumption to the level quite close to native CPU Paddle implementation. Relevant discussion takes place at #21493. 

 Changes introduced are very limited e.g.
* only inference of fp32 is supported (no int8 support yet)
* Memory savings happen only when original model params size is the same as DNNL weights size.

Both of limitations can be overcomed with more changes in core of PaddlePaddle.

Model | Peak memory cosumption : develop | Peak memory consumption: this PR
------------ | -------------|--------
googlenetv1 | 600 MB| 580 MB
demark | 622 MB | 600 MB
mobilenetv1 | 640 MB | 580 MB

I would appreciate if @LeoZhao-Intel  and @zhangting2020  would express their opinion on those changes.  Please tell me if this kind of changes are conceptually fine to you. In short : original model's param  is replaced  with DNNL's params.

CI can only pass with approval.

### DIFF
--- a/paddle/fluid/inference/api/analysis_predictor.cc
+++ b/paddle/fluid/inference/api/analysis_predictor.cc
@@ -524,6 +524,10 @@ std::unique_ptr<PaddlePredictor> CreatePaddlePredictor<
     return nullptr;
   }
 
+#ifdef PADDLE_WITH_MKLDNN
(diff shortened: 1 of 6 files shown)


LeoZhao-BJ:
LGTM

PR #22513, opened 2020-02-10
deprecated for distribute transpiler api

seiriosPlus:
API is deprecated since 1.8.0 Please use FleetAPI instead.
WIKI: https://github.com/PaddlePaddle/Fleet/blob/develop/markdown_doc/transpiler


Boyan-Liu (dismissed):

MrChengmo (approved):
LGTM

PR #22514, opened 2020-02-10
Debug for Fei Guo's PR 22309

zhhsplendid:
This PR is from Fei's PR #22309.
It makes assign op support LoDTensorArray and enable the loop_vars in
while_loop to support tuple or list.


zhhsplendid:
Close because we have debugged it

PR #22515, opened 2020-02-10
Add wrong info when use DGC in cpu

wangxicoding:
DGC doesn't support CPU for now, so assert with wrong information when using CPU paddle.


gongweibao (approved):
LGTM

PR #22540, opened 2020-02-11
Support if/else in dygraph_to_static

Aurelius84:
## 1.背景
目前Paddle框架支持动态图组网训练，为了更易于部署，可以将动态图转为静态图，更方便基于图做进一步的优化和上线。动态图下的`if/else`语法为python原生语法，转静态图时，需要解析判断、并适应转为paddle的控制流API `fluid.layers.cond`。

### 2.1 支持简单if/else语句转写
**动态图代码：**
```python
@dygraph_to_static_output
def dyfunc_with_ifElse(x_v):
    if (x_v).numpy()[0] > 5:
        x_v = x_v - 1
    else:
        x_v = x_v + 1
    return x_v
```
**转换后的静态图代码：**
```python
def true_fn(x_v):
    x_v = x_v - 1
    return ()

def false_fn(x_v):
    x_v = x_v + 1
    return ()

def dyfunc_with_ifElse2(x_v):
    x_v = fluid.layers.cond((x_v[0]) > 5), lambda : true_fn(x_v), lambda : false_fn(x_v))
    return x_v
```

### 2.2 支持嵌套的if/else
**动态图代码：**
```python
@dygraph_to_static_output
def dyfunc_with_ifElse(x_v):
    batch_size = 10
    shape = [batch_size, 16]
    bias = fluid.layers.fill_constant([shape[-1]], dtype='float32', value=1)
    if fluid.layers.mean(x_v).numpy()[0] < 0:
        x_v = x_v + bias
        w = fluid.layers.fill_constant([shape[-1]], dtype='float32', value=10)
        new_var = True
        # 嵌套的if，且缺省了else
        if x_v.numpy()[0] < 10:
            tmp = x_v * w
            x_v = fluid.layers.relu(tmp)
    else:
        new_var = False
        x_v = x_v - bias
    return x_v
```
**转换后的静态图代码：**
```python
# 对应最内层的if
def true_fn_0(w, x_v):
    tmp = x_v * w
    x_v = fluid.layers.relu(tmp)
    return x_v

# 对应最内层缺省else，函数内部没有实际处理逻辑
# 但由于layers.cond对返回值嵌套结构需一致，故需要返回与true_fn一致的变量
def false_fn_0(x_v):
    return x_v

# 对应最外层的if
def true_fn_1(True, bias, shape, x_v):
    x_v = x_v + bias
    w = fluid.layers.fill_constant([shape[-1]], dtype='float32', value=10)
    new_var = True
    x_v = fluid.layers.cond((x_v[0] < 10), lambda : true_fn_0(w, x_v), lambda : false_fn_0(x_v))
    return (new_var, x_v)

# 对应最外层的else
def false_fn_1(False, bias, x_v):
    new_var = False
    x_v = x_v - bias
    return (new_var, x_v)

def dyfunc_with_ifElse(x_v):
    batch_size = 10
    shape = [batch_size, 16]
    bias = fluid.layers.fill_constant([shape[-1]], dtype='float32', value=1)
    (new_var, x_v) = fluid.layers.cond((fluid.layers.mean(x_v)[0] < 0), lambda : true_fn_1(True, bias, shape, x_v), lambda : false_fn_1(False, bias, x_v))
    return x_v
```

### 2.3 支持控制流分支函数输入和输出参数自动解析
见2.2中构建`true_fn`和`false_fn`时对输入参数和输出参数的解析

## 3. What's Next?
- [x] 分支函数名的友好管理方案
- [x] AST转回可调用python函数的功能实现
- [x] 各子逻辑函数的单元测试


python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
```diff
@@ -105,12 +112,58 @@ class DygraphToStaticAst(gast.NodeTransformer):
...
+    def visit_Call(self, node):
+        # Remove `numpy()` statement
+        # like `Tensor.numpy()[i]` -> `Tensor[i]`
+        if isinstance(node.func, gast.Attribute):
+            attribute = node.func
+            if attribute.attr == 'numpy':
+                node = attribute.value
+        return node
+
+    def visit_Compare(self, node):
```
zhhsplendid: I saw you do nothing in visit_Compare, do we must have this method?

python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
```diff
@@ -105,12 +112,58 @@ class DygraphToStaticAst(gast.NodeTransformer):
...
-        return self.static_analysis_root
+        return self.static_analysis_root, self.decorate_func_name
 
     def transfer_from_node_type(self, node):
-        print("Not implemented")
+        self.visit(node)
+        # add new ast.funcDef of `if/else`
+        node.body = self.new_func_nodes + node.body
+
+    def visit_If(self, node):
```
zhhsplendid, line 52: Maybe I didn't talk to you clearly about the thinking in my mind.

I thought it is better to have something like this in `transfer_from_node_type`:

if "current_node is if node": # the case you are handling
   then do a visit at current_node

But in this PR, you are implementing visit_XXX on all cases. For example, no matter whether the AST has Call node, you replace `YYY.numpy` to `YYY`, this may be right in your `if` case, but may not be true in all cases (for example, a user-defined class which has a field named `numpy`).

There will be two advantage if you can visit only the `if` node:
1. The numpy case will be limited into your `if condition`
2. Other RD may also change visit_Call and you won't have conflict. 

Based on the reasons, my suggestion on current PR will be:
at visit_If, make a you-defined visitor/transformer, then do your visit in that `if` node only, not the whole AST.

python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
```diff
@@ -0,0 +1,323 @@
...
+        posonlyargs=[],
+        vararg=None,
+        kwonlyargs=[],
+        kw_defaults=None,
+        kwarg=None,
+        defaults=[])
+    return arguments
+
+
+def parse_return(parent_vars_dict, if_vars_dict, else_vars_dict):
```
zhhsplendid: Is this used for parsing return for cond? If yes, parse_cond_return may be better name.

python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
```diff
@@ -0,0 +1,323 @@
...
+                    node = copy.copy(node)
+                    node._fields = ('value', 'targets')
+                for field, value in gast.iter_fields(node):
+                    value = value if isinstance(value, list) else [value]
+                    update(name_ids,
+                           get_name_ids(value, not_name_set, node_black_list))
+    return name_ids
+
+
+def parse_args(var_ids_dict, return_ids=None, ctx=gast.Load):
```
zhhsplendid: Is this used for parsing args for cond? If yes, parse_cond_args may be better name because parse_args is very common in python

zhhsplendid (commented):
Great PR :-)

python/paddle/fluid/dygraph/jit.py
```diff
@@ -54,14 +56,12 @@ def _dygraph_to_static_output_(dygraph_func):
     def __impl__(*args, **kwargs):
         # Get AST from dygraph function
         dygraph_code = inspect.getsource(dygraph_func)
+        dygraph_code = textwrap.dedent(dygraph_code)
         root = gast.parse(dygraph_code)
 
-        root = DygraphToStaticAst().get_static_ast(root)
+        root, func_name = DygraphToStaticAst().get_static_ast(root)
+        static_func, file_name = ast_to_func(root, func_name)
```
liym27: Is it better that the return value of `get_static_ast` only include ast.node because "get_static_ast" means get static ast ?

Aurelius84: Thanks, fixed

python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
Aurelius84: Thanks, it was renamed  into `parse_cond_args`

python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
Aurelius84: Thanks, it was renamed  into `parse_cond_return`.

python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
Aurelius84: Great advice! 
I extract the code related to `if/else` transformation into `class IfElseTransformation`. 

There are two things to note：
1. Only `ast.If` will  be parsed and modified in `visit_if`， even though `root` be required.(We need root to parse parameters of nested `true_fn` and `false_fn`). The Transformer   should be decoupled with `root` ideally. This will be optimized in later PR.

2. `tensor.numpy()[]` is an interface of Dygraph and may be considered as `Basic API`, I will remove it after yamei's PR is merged. It's needed for unittest temporarily.

zhhsplendid: OK, since tensor.numpy() will be a Basic API, and my PR to analysis type is merged. Can we only remove the .numpy when the variable is tensor? But this will be Yamei's work :-)

zhhsplendid: Sorry, considering the parameters and returns, the `if` node requires `root`, thank you! But let's split in a different transformer for conflict reason

zhhsplendid (approved):

luotao1 (approved):
LGTM for python/requirements.txt

python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
Aurelius84: > OK, since tensor.numpy() will be a Basic API, and my PR to analysis type is merged. Can we only remove the .numpy when the variable is tensor? But this will be Yamei's work :-)

I will  merge this function into BasicApiTransformer in next PR.

liym27 (approved):

PR #21634, opened 2019-12-09
Refine dygraph DataLoader implementation

chenwhql:
Now the dygraph DataLoader will decrease training speed, because of some implementation issues.
This PR refine the dygraph DataLoader implementation to solve following problem.

image

### Optimization effect (single card)

**transformer（s）**
batch_size = 256, epoch = 20

| original DataLoader | New DataLoader | Time decrease | 
|---|---|---|
| 1418.013 | 1320.847 | -6.9% |

**resnet（s）**
batch_size = 256, epoch = 120

| original DataLoader | New DataLoader | Time decrease | 
|---|---|---|
| 12221.662 | 11468.035 | -6.2% |

PR #22541, opened 2020-02-11
Speed up dygraph DataLoader based on shared memory and LoDTensor serialization

chenwhql:
This PR is to speed up dygraph DataLoader, adding two mechanisms:
1. LoDTensor shared memory
- The data of LoDTensor created by DataLoader child process will be moved to shared memory.
- Then it  can be reconstructed in another process based on the shared memory identifier.
2. LoDTensor serialization
- We can put the LoDTensor into multiprocessing.Queue or Pipe in one process, and then take out the tensor in another process
- Tensor data copy does not occur between processes

##### Compared to the original multi-process DataLoader

- Singal Card 1 epoch (s)

| model | DataLoader 1.7 | DataLoader New | UpRate |
|---|---|---|---|
| resnet | 62.04 | 53.88 | **+15.1%** | 
| mobilenet v1 | 6614.68 | 3303.86 | **+100.2%** |

> 1. For mnist, se_resnet, ptb and transformer, the time to read data on the DataLoader1.7 is already close to 0, no obvious improvement
> 2. Due to the heavy burden of child threads, DataLoader1.7 multi-process has no acceleration effect on mobilenet, so the improvement is obvious

##### Current performance

- mobilenet v1 1 epoch train time(s)

| mode | DataLoader SingPro  | DataLoader MultiPro | SppedUpRate |
|---|---|---|---|
| 1 card | 5038.19 | 3303.86 | **52.5%** |
| 4 cards | 1250.87 | 933.70 | **34.0%** |

- mobilenet v1 1 epoch read next approximate time(s)

| mode | DataLoader SingPro  | DataLoader MultiPro |
|---|---|---|
| 1 card | 0.39 | 0.02 ~ 0.13 |
| 4 cards | 0.05 ~ 0.07 | 0.0001 |

- Other simple models train time(s) - 1 card 1 epoch

| model |  Reader | DataLoader SingPro | DataLoader MultiPro  |  Fake Data | Batch Read Time |
|---|---|---|---|---|---|
| mnist | 10.60 | 10.28 | 6.60 **(+60.6%)** | 6.16 |  0.0001  |
| resnet | 94.69 | 83.98 | 53.88 **(+75.7%)** |  44.08 | 0.020 |
| se_resnext | 178.92 | 132.29 | 125.51 **(+42.6%)** | 123.83 | 0.0001 |
| ptb_lm | 120.33 | 108.56 | 108.27 **(+11.1%)**  | 107.43 | 0.0001 |

#### Related PR:
- https://github.com/PaddlePaddle/Paddle/pull/21762
- https://github.com/PaddlePaddle/Paddle/pull/21634

#### Note:
- Docker `/dev/shm` size: This PR use mmap(Memory Map) to implement shared memory. You need to confirm if your shared memory is big enough, otherwise a bus error will occur. The docker default `/dev/shm` size is 64M, you can set it by docker `--shm-size` parameter, for example, `docker run --name [Name of container] -it --shm-size=2g -v $PWD:/paddle <imagename> /bin/bash`, you can see docker's shared memory size by `df -h`
- You need to make sure your `/dev/shm` size is larger than the following calculation results, otherwise, it will report an `Bus error` because of insufficient shared memory
```
DataLoader Num * queue capacity * 1 batch data size
```

----

#### 1. 分3种情况：
- DataLoader使用全部数据，读取完，正常结束：无需清理
- DataLoader仅使用部分数据，到一定step break，正常结束：需要清理
- DataLoader运行到一半，用户Ctrl+C，异常结束：需要清理

#### 2. 需要清理的内容：
- 主进程未来得及被析构的内存映射文件需要清理（大部分在BlockingQueue中）
- 进程间Queue中的内存映射文件需要清理
- 子进程中已申请但未来得及被放入进程间Queue的内存映射文件需要清理

#### 3. 清理方法：
- 对于主进程包括BlockingQueue残留文件的清理
  - 创建记录主进程文件描述符的全局Set单例，在程序退出时Clear
  - 为什么不采用BlockingQueue的全局Set单例，在程序退出时Clear
    - 实现其实写完了，但这种方式总是会剩下几个清不掉
    - 这几个没清掉的文件不在BlockingQueue中，也不在进程间Queue中
    - 应该是线程读出来的实例，尚未放进BlockingQueue中，但由于突然退出也未被析构
    - 所以采取了直接记录文件描述符的方法
- 对于进程间Queue残留文件的清理
  - 创建全局的multiprocessing.Queue set，记录DataLoader中创建的进程间Queue
  - 实现register_exit_func，注册python退出时执行的_cleanup函数，对multiprocessing.Queue set中所有非空的Queue进行清理
  - 这里涉及到对原DataLoader实现的改动
    - _cleanup函数执行时，multiprocessing.Queue可能已经被close了，而Queue close的时候不会对内存映射文件进行清理
    - 另外，Queue.cancel_join_thread(allow_exit_without_flush)也是为了不等待队列中数据清除，直接退出进程，避免hang的函数，而现在既然会对所有的Queue在最后进行清除，理论上也不再需要调用这个方法了
    - 因此去掉了几处主动close和cancel_join_thread multiprocessing.Queue的实现，改为由程序自动调用
      - https://docs.python.org/3.5/library/multiprocessing.html#multiprocessing.Queue.close
      - https://docs.python.org/3.5/library/multiprocessing.html#multiprocessing.Queue.cancel_join_thread
- 对于子进程中残留文件的清理（非常少，不超过一个batch）
  - 在子进程中也维护文件描述符全局Set集合
  - 在创建内存映射文件的时候，记录到Set集合中，在tensor list插入到进程间Queue之后，从Set集合中移除
  - 最后用注册的exit函数清理子进程文件描述符Set
  - 这样创建了但未被放入进程间Queue的文件也能够被清理

#### 4. 经初步测试，以上机制，能够在训练中途break以及ctrl+c时完成对内存映射文件的清理
- 执行单测test_imperative_data_loader_fds_clear.py 20次，/dev/shm下无残留文件


paddle/fluid/memory/allocation/mmap_allocator.h
```diff
@@ -0,0 +1,95 @@
...
+
+namespace paddle {
+namespace memory {
+namespace allocation {
+
+class MemoryMapWriterAllocation : public Allocation {
+ public:
+  explicit MemoryMapWriterAllocation(void *ptr, size_t size,
+                                     std::string ipc_name)
+      : Allocation(ptr, size, platform::CPUPlace()), ipc_name_(ipc_name) {}
```
sneaxiy: Use `ipc_name_(std::move(ipc_name))` instead.

paddle/fluid/memory/allocation/mmap_allocator.h
```diff
@@ -0,0 +1,95 @@
...
+
+  inline std::string ipc_name() const { return ipc_name_; }
```
sneaxiy: Return const reference instead.

paddle/fluid/memory/allocation/mmap_allocator.h
```diff
@@ -0,0 +1,95 @@
...
+
+ private:
+  std::string ipc_name_;
+};
+
+class MemoryMapReaderAllocation : public Allocation {
+ public:
+  explicit MemoryMapReaderAllocation(void *ptr, size_t size,
+                                     std::string ipc_name)
+      : Allocation(ptr, size, platform::CPUPlace()), ipc_name_(ipc_name) {}
```
sneaxiy: `ipc_name_(std::move(ipc_name))`

paddle/fluid/memory/allocation/mmap_allocator.h
```diff
@@ -0,0 +1,95 @@
...
+
+  inline std::string ipc_name() const { return ipc_name_; }
```
sneaxiy: Return const reference.

paddle/fluid/memory/allocation/mmap_allocator.h
```diff
@@ -0,0 +1,95 @@
...
+};
+
+class MemoryMapAllocator {
+ public:
+  MemoryMapAllocator() {}
+  ~MemoryMapAllocator() {}
+
+  std::shared_ptr<MemoryMapWriterAllocation> Allocate(size_t size);
+
+  std::shared_ptr<MemoryMapReaderAllocation> Rebuild(std::string ipc_name,
```
sneaxiy: Use `const std::string &`.

paddle/fluid/memory/allocation/mmap_allocator.h
```diff
@@ -0,0 +1,95 @@
...
+
+  inline std::string ipc_name() const { return ipc_name_; }
+
+  ~MemoryMapReaderAllocation() override;
+
+ private:
+  std::string ipc_name_;
+};
+
+class MemoryMapAllocator {
```
sneaxiy: Seems that this allocator is not necessary. We can just use 2 global functions `AllocateMemoryMapWriterAllocation` and `RebuildMemoryMapReaderAllocation`.

paddle/fluid/pybind/imperative.cc
```diff
@@ -220,6 +221,80 @@ void BindImperative(py::module *m_ptr) {
...
+                    "lists with different lengths.\n  * Check the reader "
+                    "function passed to 'set_(sample/sample_list/batch)"
+                    "_generator' to locate the data causes this issue."));
+            // 2. construcct LoDTensor
+            framework::LoDTensor t;
+            SetTensorFromPyArray<platform::CPUPlace>(
+                &t, array, platform::CPUPlace(), true);
+            // 3. allocate shared memory
+            memory::allocation::MemoryMapAllocator allocator;
+            void *data_ptr = reinterpret_cast<void *>(
```
sneaxiy: Why not invoke `void *data_ptr = t.data<void>()` instead?

paddle/fluid/pybind/imperative.cc
```diff
@@ -220,6 +221,80 @@ void BindImperative(py::module *m_ptr) {
...
+                reinterpret_cast<uintptr_t>(t.Holder()->ptr()) + t.offset());
+            size_t data_size = t.memory_size();
```
sneaxiy: `memory_size()` may be larger than `t.numel() * SizeOfType(t.type())`. How about use `t.numel() * SizeOfType(t.type())` instead?

paddle/fluid/pybind/imperative.cc
```diff
@@ -220,6 +221,80 @@ void BindImperative(py::module *m_ptr) {
...
+            auto shared_writer_holder = allocator.Allocate(data_size);
+            // 4. maintain mmap fd set & backup ipc_name
+            std::string ipc_name = shared_writer_holder->ipc_name();
```
sneaxiy: Use const reference.

paddle/fluid/pybind/imperative.cc
```diff
@@ -220,6 +221,80 @@ void BindImperative(py::module *m_ptr) {
...
+            tensors.append(t);
+          }
+          return tensors;
+        },
+        py::return_value_policy::take_ownership);
+
+  m.def("_remove_tensor_list_mmap_fds", [](py::list &tensor_list) {
+    for (size_t i = 0; i < tensor_list.size(); ++i) {
+      auto t = tensor_list[i].cast<framework::LoDTensor>();
+      auto *mmap_writer_allocation =
```
sneaxiy, line 286: Check whether `mmap_writer_allocation` is nullptr?

paddle/fluid/pybind/pybind.cc
```diff
@@ -859,7 +860,57 @@ PYBIND11_MODULE(core_noavx, m) {
...
+          [](py::tuple t) {  // __setstate__
+            if (t.size() != 5)
+              throw std::runtime_error("Invalid LoDTensor state!");
+
+            // 1. Create a new C++ instance
+            LoDTensor tensor;
+
+            // 2. Rebuild Allocation
+            memory::allocation::MemoryMapAllocator alloctor;
+            std::string ipc_name = t[0].cast<std::string>();
```
sneaxiy: Const reference.

python/paddle/fluid/reader.py
```diff
@@ -27,16 +27,90 @@
...
+            _registered_exit_funcs.add(func)
+
+    register_func(func, signals)
+
+
+def _clear_multiprocess_queue_set():
+    global multiprocess_queue_set
+    for data_queue in multiprocess_queue_set:
+        if data_queue is None:
+            return
```
sneaxiy: Why return but not continue and why it would be None?

python/paddle/fluid/reader.py
```diff
@@ -414,8 +501,12 @@ def _init_iterable(self):
 
     def _start(self):
         if self._use_multiprocess:
+            self._thread_done_event = threading.Event()
             # Set data_queue and process
             self._data_queue = multiprocessing.Queue(self._capacity)
+            # add _data_queue into global queue set
+            global multiprocess_queue_set
+            multiprocess_queue_set.add(self._data_queue)
```
sneaxiy, line 528: Seems that `multiprocess_queue_set` would increase as epoch increase. Try to remove the original `self._data_queue` and pop all data outs, and then create a new `multiprocess.Queue`?

python/paddle/fluid/reader.py
```diff
@@ -27,16 +27,90 @@
...
+# NOTE: [ avoid hanging & failed quickly ] These value is used in getting data from another process
+QUEUE_GET_TIMEOUT = 60
+
+# NOTE: [ mmap files clear ] If there is still data in the multiprocess queue when the main process finishes reading,
+# the data in the queue needs to be popped. Then the LoDTensor read by the main process
+# from the child process will automatically clear the memory-mapped file.
+multiprocess_queue_set = set()
+
+_registered_exit_funcs = set()
+_executed_exit_funcs = set()
```
sneaxiy: What is `_executed_exit_funcs` used for?

paddle/fluid/memory/allocation/mmap_allocator.h
chenwhql: done

paddle/fluid/memory/allocation/mmap_allocator.h
chenwhql: done

paddle/fluid/memory/allocation/mmap_allocator.h
chenwhql: done

paddle/fluid/memory/allocation/mmap_allocator.h
chenwhql: done

paddle/fluid/memory/allocation/mmap_allocator.h
chenwhql: done

paddle/fluid/memory/allocation/mmap_allocator.h
chenwhql: done

paddle/fluid/pybind/imperative.cc
chenwhql: done

paddle/fluid/pybind/imperative.cc
chenwhql: done

paddle/fluid/pybind/imperative.cc
chenwhql: done

paddle/fluid/pybind/imperative.cc
chenwhql: done

paddle/fluid/pybind/pybind.cc
chenwhql: done

python/paddle/fluid/reader.py
chenwhql: remove it, it would not be None

python/paddle/fluid/reader.py
chenwhql: yes, done

python/paddle/fluid/reader.py
chenwhql: Avoid the same function being executed repeatedly

sneaxiy (approved):

luotao1 (approved):

PR #22553, opened 2020-02-12
Use available ports instead of static ports.

gongweibao:
If the program uses static ports, there will be a port conflict when many jobs started and so use the available ports will resolve this.


python/paddle/distributed/launch.py
```diff
@@ -63,6 +65,32 @@ def _print_arguments(args):
...
+def find_free_ports(num):
+    def __free_port():
+        with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+            s.bind(('', 0))
+            return s.getsockname()[1]
+
+    port_set = set()
+    step = 0
+    while True:
+        port = __free_port()
```
wangxicoding, line 77: if this iter get port1, will next iter get port1 again?

python/paddle/distributed/launch.py
```diff
@@ -212,12 +240,25 @@ def start_procs(args):
                 logger.warning("Use Cloud specified port:{}.".format(
                     cloud_paddle_port))
 
+    free_ports = None
+    if not args.use_paddlecloud and num_nodes <= 1:
```
wangxicoding: only use in single-node?

python/paddle/distributed/launch.py
gongweibao: No, the probability is small and the port will be filtered by the port_set.

python/paddle/distributed/launch.py
gongweibao: Yes. This PR will only resolve the single node and when they use multi-nodes, the cluster-system will assign the property port.

wangxicoding (dismissed):
LGTM

python/paddle/distributed/launch.py
```diff
@@ -212,12 +240,25 @@ def start_procs(args):
...
+        free_ports = find_free_ports(selected_gpus_num)
+        if free_ports is not None:
+            free_ports = list(free_ports)
+
+    if free_ports is None:
+        free_ports = [
+            x
+            for x in range(args.started_port, args.started_port +
```
gavin1332, line 256: What's the priority of the started_port config and the free_ports? It seems that if the free_ports are available, the launcher will use free_ports regardless of started_port set by user.

gavin1332 (commented):
need discussion

python/paddle/distributed/launch.py
gongweibao: Done.

gavin1332 (approved):
LGTM

PR #22506, opened 2020-02-08
Add support for Ernie NLP model to the Slim QAT

wojtuss:
This PR adds support for the Ernie NLP model in QAT2.0.
A documentation update explaining how to run accuracy and performance benchmarks on this model is submitted in a separate PR: https://github.com/PaddlePaddle/Paddle/pull/22519.
For the performance benchmark to work, the PR https://github.com/PaddlePaddle/benchmark/pull/311 needs to be merged.

PR #22575, opened 2020-02-12
[Cherry-Pick][1.7] Enable MKL-DNN Quantization of Ernie in Slim

Sand3r-:
Cherry-pick and fix of #22506


bingyanghuang:
The small fix added in this PR is because #22209 introduce the following lines used in our pass not cherry-picked to release/1.7:
```c++
      .def("set", [](ir::Pass &self, const std::string &name,
                     bool val) { self.Set<bool>(name, new bool(val)); })
```

luotao1 (approved):

PR #22588, opened 2020-02-13
Fix mismatch with plus sign in the line

chenwhql:
When there is a plus sign in the line, the deleted line will also be mismatched for error specification check. This PR fix this problem.

- mismatch examples: these two lines are deleted content
http://ci.paddlepaddle.org/viewLog.html?buildId=282587&tab=buildLog&buildTypeId=YYTest_PrCiCpuPy2&logTab=tail
 
replace `grep "+"` with `grep "^+"`, resolve this problem.

- success examples: the code diff keep the same as mismatch examples
http://ci.paddlepaddle.org/viewLog.html?buildId=282627&buildTypeId=YYTest_PrCiCpuPy2&tab=buildLog


luotao1 (approved):
LGTM

PR #22594, opened 2020-02-13
optimizer var exlcude fluid.data var

heavengate:
**need to filter out form optimizer var if fluid.data Variable is set as persistable**


phlrain (approved):
LGTM

PR #22595, opened 2020-02-13
fix py func bug when out is type[list] and add unitest case

zhwesky2010:
fix issue #22581 
fix py func when out is `type[list]`, out can be `Variable/tuple(Variable)/list(Variable)` type .
And add unittest case when `out=tuple(Variable)` or `out=list(Variable)`, and improve coverage.


liupluswei (approved):
LGTM

PR #22596, opened 2020-02-13
[cherry-pick1.7]fix py func bug when out is type[list] and add unitest case

zhwesky2010:
fix issue #22581 
fix py func bug when out is `type[list]`, out can be `Variable/tuple/list` type .


liupluswei (approved):
LGTM

PR #22617, opened 2020-02-15
deprecated for distribute transpiler api (#22513)

seiriosPlus:
* add deprecated for distribute transpiler, will delete it after 2.0.0, test=develop


MrChengmo (approved):
LGTM

PR #22632, opened 2020-02-17
Fix instag

yaoxuefeng6:
Add "out_val_ifempty" attribute in fileter_by_instag op.
By setting this attribute, user can determine the output value if all instances in this batch are filtered. 
Additional, we filter negative label in auc op. With default all zero out value in filter_by_instag op, there might be some invalid data fed to auc op. we can set negative out value in filter_by_instag op to avoid this.


hutuxian:
LGTM for AUC part.

saxon-zh (dismissed):

bjjwwang (dismissed):
LGTM

paddle/fluid/operators/filter_by_instag_op.cc
```diff
@@ -60,6 +60,9 @@ class FilterByInstagOpMaker : public framework::OpProtoAndCheckerMaker {
     AddInput("Ins_tag", "(LoDTensor) ins tag list");
     AddInput("Filter_tag", "(1D Tensor) filter tag list");
     AddAttr<bool>("is_lod", "is Ins with LoD info or not, default True");
+    AddAttr<int64_t>("out_val_ifempty",
```
XiaoguangHu01: 建议out_val_ifempty -> out_val_if_empty

saxon-zh (approved):

XiaoguangHu01 (approved):

PR #22634, opened 2020-02-17
support num_flatten_dims=-1 of API fc.

liym27:
**This PR** : support `num_flatten_dims`=-1 of static graph API [fluid.layers.fc](https://www.paddlepaddle.org.cn/documentation/docs/en/develop/api/layers/fc.html)

The corresponding dygraph API [Linear](https://www.paddlepaddle.org.cn/documentation/docs/en/develop/api/dygraph/Linear.html) doesn't have parameter `num_flatten_dims`. And the default output of `Linear` is not equal to `fc`.

- fc
```
paddle.fluid.layers.fc(input, size, num_flatten_dims=1, param_attr=None, bias_attr=None, act=None, name=None)
```
- Linear
```
class paddle.fluid.dygraph.Linear(input_dim, output_dim, param_attr=None, bias_attr=None, act=None, dtype='float32')
```

support `num_flatten_dims`=-1 so that the output of `Linear` and `fc` is the same.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_fc_op.py
+++ b/python/paddle/fluid/tests/unittests/test_fc_op.py
@@ -18,6 +18,8 @@ from op_test import OpTest
 import paddle.fluid as fluid
 from paddle.fluid import Program, program_guard
 
+SEED = 2020
+
 
 def fc_refer(matrix, with_bias, with_relu=False):
     in_n, in_c, in_h, in_w = matrix.input.shape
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/tests/unittests/test_fc_op.py
```diff
@@ -131,6 +133,34 @@ def config(self):
         self.matrix = MatrixGenerate(1, 4, 3, 128, 128, 2)
 
 
+class TestFcAPI(unittest.TestCase):
+    # test for parameter num_flatten_dims=-1
```
zhhsplendid: Rename class "TestFcAPI" to "TestFcOp_NumFlattenDims_NegOne" or  "TestFcOpWithNegOneNumFlattenDims"then delete comment at line 137.

Reason:
1. TestFcAPI doesn't contain lots of information considering that there is a class TestFcOp in the same file
2. After renaming, the class speaks the meaning itself.

python/paddle/fluid/tests/unittests/test_fc_op.py
```diff
@@ -131,6 +133,34 @@ def config(self):
...
+            x = fluid.layers.data(
+                name="x",
+                shape=[2, 2, 25],
+                append_batch_size=False,
+                dtype="float32")
+
+            out_1 = fluid.layers.fc(input=x, size=1, num_flatten_dims=-1)
+            out_2 = fluid.layers.fc(input=x, size=1, num_flatten_dims=2)
+
+        exe = fluid.Executor(place=fluid.CPUPlace())
```
zhhsplendid: Could you write something like:

exe = fluid.Executor(fluid.CPUPlace(
            ) if not core.is_compiled_with_cuda() else fluid.CUDAPlace(0))

It tests GPU case with minor modification.

python/paddle/fluid/tests/unittests/test_fc_op.py
liym27: Thanks. Done

python/paddle/fluid/tests/unittests/test_fc_op.py
liym27: Thanks. Done

zhhsplendid (approved):
LGTM

PR #22657, opened 2020-02-18
support cond in clone, test=develop

zhiqiu:
When `clone(for_test=True)` called after minimize, the appended backward part should be pruned correctly. 
The origin implementation cannot handle program with conditional block. For example, it raises error like this,
image

The main problem is that, the loss may not occur in global block and the number of blocks in cloned program will be less than the original program. 
 There are two cases,

1. Without minimize in conditional block

```
def gen_loss1: ...
def gen_loss2: ...
sgd = fluid.optimizer.SGD()
# 在运行时根据pred选择执行gen_loss1还是gen_loss2
loss = fluid.layers.cond(pred, gen_loss1, gen_loss2)
sgd.minimize(loss) 

for i in range(10):
	res = exe.run(feed, fetch_list=[loss.name])
```
image
In this case, the id of a block in cloned program and its corresponding origin one are identical, though the number of blocks is decreased.

2. With minimize in conditional block

```
def gen_loss1: 
	loss1 = Net1()
	sgd,m.minimize(loss1)
return loss1def gen_loss2: 
	loss2 = Net1()
	sgd.minimize(loss2)
	return loss2
# 在运行时根据pred选择执行gen_loss1还是gen_loss2
loss = fluid.layers.cond(pred, gen_loss1, gen_loss2)

for i in range(10):
	res = exe.run(feed, fetch_list=[loss.name])
```
image
In this case, the id of a block in cloned program and its corresponding origin one are not identical.

This PR handles both cases.

### DIFF
--- a/paddle/fluid/framework/prune.cc
+++ b/paddle/fluid/framework/prune.cc
@@ -83,20 +83,13 @@ bool HasFalseTarget(const proto::OpDesc& op_desc) {
 }
 
 int GetSubBlockIndex(const proto::OpDesc& op_desc) {
+  // the block index >= 0, so -1 is used to indicate "NotFound".
   for (auto& attr : op_desc.attrs()) {
     if (attr.type() == proto::AttrType::BLOCK) {
-      PADDLE_ENFORCE(attr.has_block_idx());
+      PADDLE_ENFORCE_EQ(
+          attr.has_block_idx(), true,
+          platform::errors::NotFound("Attribute sub_block is empty"));
       return attr.block_idx();
-    }
-  }
-  return -1;
-}
-
-int GetOpRole(const proto::OpDesc& op_desc) {
-  for (auto& attr : op_desc.attrs()) {
-    if (attr.name() == OpProtoAndCheckerMaker::OpRoleAttrName()) {
-      PADDLE_ENFORCE(attr.has_i());
-      return attr.i();
     }
   }
   return -1;
@@ -105,14 +98,30 @@ int GetOpRole(const proto::OpDesc& op_desc) {
 void SetSubBlockIndex(proto::OpDesc* op_desc, int sub_idx) {
   for (auto& attr : *op_desc->mutable_attrs()) {
     if (attr.type() == proto::AttrType::BLOCK) {
-      PADDLE_ENFORCE(attr.has_block_idx());
-      return attr.set_block_idx(sub_idx);
+      PADDLE_ENFORCE_EQ(
+          attr.has_block_idx(), true,
+          platform::errors::NotFound("Attribute sub_block is empty"));
+      attr.set_block_idx(sub_idx);
     }
   }
 }
 
 bool HasSubBlock(const proto::OpDesc& op_desc) {
   return GetSubBlockIndex(op_desc) > 0;
+}
+
+int GetOpRole(const proto::OpDesc& op_desc) {
+  // the normal op role >= 0, so -1 is used to indicate "NotFound".
+  for (auto& attr : op_desc.attrs()) {
+    if (attr.name() == OpProtoAndCheckerMaker::OpRoleAttrName()) {
+      PADDLE_ENFORCE_EQ(
+          attr.has_i(), true,
+          platform::errors::NotFound("Attribute %s is empty",
+                                     OpProtoAndCheckerMaker::OpRoleAttrName()));
+      return attr.i();
+    }
+  }
+  return -1;
 }
 
 void AppendOpInputVarNames(const proto::OpDesc& op_desc,
@@ -280,11 +289,6 @@ void Prune(const proto::ProgramDesc& input,
   prune_impl(input, output, 0, -1, &dependent_vars, feed_var_names);
 }
 
-void UpdateBlockIdx(proto::ProgramDesc* prog,
-                    std::map<int, int> pruned_origin_block_map) {
-  return;
-}
-
 void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
                      int block_id, int parent_block_id) {
   auto* block_field = output->mutable_blocks();
@@ -296,6 +300,7 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
 }
 
 int FindMapByValue(std::map<int, int> m, int val) {
+  // the content in map should be >= 0, so -1 is used to indicate "NotFound".
   std::map<int, int>::iterator it;
   for (it = m.begin(); it != m.end(); ++it) {
     if (it->second == val) {
@@ -306,14 +311,13 @@ int FindMapByValue(std::map<int, int> m, int val) {
 }
 
 void PruneBackwardImpl(proto::BlockDesc* origin, proto::BlockDesc* pruned) {
-  // Step 3. Mark backward & optimize ops on main branch
   std::unordered_set<std::string> op_input_vars;
   std::unordered_set<std::string> op_output_vars;
 
+  // Step 1. Mark backward, optimize and lrsched ops in the block
   auto* ops = origin->mutable_ops();
   for (auto op_iter = ops->begin(); op_iter != ops->end(); ++op_iter) {
     auto& op_desc = *op_iter;
-
     auto op_role = GetOpRole(op_desc);
     if (op_role & static_cast<int>(OpRole::kOptimize) ||
         op_role & static_cast<int>(OpRole::kBackward) ||
@@ -322,9 +326,10 @@ void PruneBackwardImpl(proto::BlockDesc* origin, proto::BlockDesc* pruned) {
     }
   }
 
-  // Step 5. Copy the forward ops to new ProgramDesc
-  //   Note: The proto::ProgramDesc doesn't have interface
-  //         to remove op and var
+  // Step 2. Copy the forward ops which have not been set false target to new
+  // ProgramDesc
+  // Note: The proto::ProgramDesc doesn't have interface
+  //       to remove op and var
   auto* op_field = pruned->mutable_ops();
   op_field->Clear();
   for (auto op_iter = ops->begin(); op_iter != ops->end(); ++op_iter) {
@@ -336,7 +341,7 @@ void PruneBackwardImpl(proto::BlockDesc* origin, proto::BlockDesc* pruned) {
     }
   }
 
-  // Step 6. Copy the forward vars to new ProgramDesc
+  // Step 3. Copy the forward vars to new ProgramDesc,
   // construct all var's map before clear
   auto* origin_vars = origin->mutable_vars();
   auto* pruned_vars = pruned->mutable_vars();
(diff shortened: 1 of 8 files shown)


paddle/fluid/framework/prune.cc
```diff
@@ -269,124 +299,154 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
   output_block->set_parent_idx(parent_block_id);
 }
 
-void PruneBackwardImpl(proto::ProgramDesc* input, proto::ProgramDesc* output,
-                       int block_id, int parent_block_id) {
-  // Step 1. Copy the current input block to output
-  CloneWholeBlock(input, output, block_id, parent_block_id);
-  int output_block_id = output->blocks_size() - 1;
-  auto* output_block = output->mutable_blocks(output_block_id);
+int FindMapByValue(std::map<int, int> m, int val) {
```
sneaxiy: Use `const std::map<int, int>&` instead.

paddle/fluid/framework/prune.cc
```diff
@@ -269,124 +299,154 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
...
+  // the content in map should be >= 0, so -1 is used to indicate "NotFound".
+  std::map<int, int>::iterator it;
+  for (it = m.begin(); it != m.end(); ++it) {
```
sneaxiy: You can just use for-range iteration like this:
```
for (auto &pair : m) {
  if (pair.second == val) {
    return pair.first;
  }
}
```

python/paddle/fluid/framework.py
```diff
@@ -4479,6 +4478,10 @@ def _copy_data_info_from(self, other):
 
         Args:
             other(Program): Other program
+            pruned_origin_block_id_map(map{int:int}): A map from block id in self
```
liym27: `map{int:int}` means `dict{int:int}` ?  The type of `pruned_origin_block_id_map` is `dict` in Python.

paddle/fluid/framework/prune.cc
```diff
@@ -81,19 +83,47 @@ bool HasFalseTarget(const proto::OpDesc& op_desc) {
 }
 
 int GetSubBlockIndex(const proto::OpDesc& op_desc) {
+  // the block index >= 0, so -1 is used to indicate "NotFound".
```
liym27: Better to capitalize the first word since it is a new sentence.

paddle/fluid/framework/prune.cc
```diff
@@ -81,19 +83,47 @@ bool HasFalseTarget(const proto::OpDesc& op_desc) {
...
+    }
+  }
+}
+
 bool HasSubBlock(const proto::OpDesc& op_desc) {
   return GetSubBlockIndex(op_desc) > 0;
 }
 
+int GetOpRole(const proto::OpDesc& op_desc) {
+  // the normal op role >= 0, so -1 is used to indicate "NotFound".
```
liym27: the -> The.  
" the normal op role >= 0" may mean "The op_role >= 0 normally" ?

liym27 (commented):
Please check the syntax and  unify the style of code comments.

paddle/fluid/framework/prune.cc
```diff
@@ -269,124 +299,154 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
...
-                    static_cast<int>(OpRole::kLoss))) {
-      op->SetAttr(OpProtoAndCheckerMaker::OpRoleAttrName(),
-                  static_cast<int>(OpRole::kForward));
-      op->SetIsTarget(true);
-      has_loss_op = true;
-    } else if (op_role == (static_cast<int>(OpRole::kBackward) |
-                           static_cast<int>(OpRole::kLoss))) {
-      op->SetIsTarget(false);
-      break;
+  // Step 1: check if the program contains grad loss operator.
```
liym27: Step 1:  -> Step 1.

paddle/fluid/framework/prune.cc
```diff
@@ -269,124 +299,154 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
...
+      int parent_idx = -1;
+      if (origin->parent_idx() == -1) {
+        parent_idx = -1;
+      } else {
+        parent_idx =
+            FindMapByValue(pruned_progin_block_id_map, origin->parent_idx());
+        PADDLE_ENFORCE_NE(parent_idx, -1,
+                          "The parent block id should be found in "
+                          "pruned_progin_block_id_map");
+      }
```
liym27, line 420:
```
int parent_idx = -1;
if (origin->parent_idx() == -1) {
     parent_idx = -1;
} else {...}
```
to
```
int parent_idx = -1;
if (origin->parent_idx() != -1){
...
}
```

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
```diff
@@ -198,14 +251,56 @@ def optimizer():
...
+
+    def test_optimization_in_cond(self):
+        x_in = np.random.random(size=(10, 4)).astype('float32')
+        label_in = np.random.randint(1, size=(10, 1)).astype('int64')
+        feed_dict = {'x': x_in, 'label': label_in}
+        with self.program_scope_guard():
+            loss = optimization_in_cond_net(False)
+            main_program = fluid.default_main_program()
+            test_prog_orig = main_program.clone(for_test=True)
+            place = core.CPUPlace()
```
liym27, line 278: Could you write something like 
`place = core.CPUPlace() if not core.is_compiled_with_cuda() else core.CUDAPlace(0) ` which tests GPU case ?

python/paddle/fluid/backward.py
```diff
@@ -1209,9 +1209,8 @@ def append_backward(loss,
     is_in_control_flow = current_block_idx != 0
 
     # Double grad is not supported in sub-block (control flow)
-    if not is_in_control_flow:
```
liym27: Why `if not is_in_control_flow` was removed?

paddle/fluid/framework/prune.cc
```diff
@@ -81,19 +83,47 @@ bool HasFalseTarget(const proto::OpDesc& op_desc) {
...
   }
   return -1;
 }
 
+void SetSubBlockIndex(proto::OpDesc* op_desc, int sub_idx) {
+  for (auto& attr : *op_desc->mutable_attrs()) {
+    if (attr.type() == proto::AttrType::BLOCK) {
+      PADDLE_ENFORCE_EQ(
+          attr.has_block_idx(), true,
+          platform::errors::NotFound("Attribute sub_block is empty"));
```
liym27: Better to print the Op's name in the error message.

paddle/fluid/framework/prune.cc
```diff
@@ -269,124 +299,154 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
...
   }
+  pruned_vars->Clear();
+
   std::unordered_set<std::string> var_names;
   var_names.insert(op_input_vars.begin(), op_input_vars.end());
   var_names.insert(op_output_vars.begin(), op_output_vars.end());
-  var_field->Clear();
   for (const auto& name : var_names) {
-    *var_field->Add() = var_map[name];
+    if (var_map.count(name)) {
```
sneaxiy, line 350: Why `name` does not exist in `var_map`? It seems that `var_map` contains all vars from original block and `name` is a var from prune block ops. So, is it better to use `PADDLE_ENFORCE_xxx` to avoid bug?

paddle/fluid/framework/prune.cc
```diff
@@ -269,124 +299,154 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
...
+        op->SetIsTarget(false);
+        has_loss_grad_op = true;
+        break;
+      }
     }
   }
-  PADDLE_ENFORCE_EQ(has_loss_op, true,
-                    "The Program need to be pruned its backward part"
-                    "should have loss operator.");
+  PADDLE_ENFORCE_EQ(
```
sneaxiy: How about make that this function can be invoked with a forward network. In `framework.py`, the original implementation use `self._appending_grad_times >= 1` to distinguish whether it is a forward network. But it is a bit hard coding...

paddle/fluid/framework/prune.cc
```diff
@@ -269,124 +299,154 @@ void CloneWholeBlock(proto::ProgramDesc* input, proto::ProgramDesc* output,
...
-  // Step 3. Contruct new framework::ProgramDesc
-  return std::unique_ptr<framework::ProgramDesc>(
-      new framework::ProgramDesc(pruned_desc));
-}
+  // Step 3: Update subblock attribute for conditional operator.
+  // This should be performed after all blocks pruned.
+  for (int i = 0; i < pruned_desc.blocks_size(); i++) {
+    auto* pruned = pruned_desc.mutable_blocks(i);
+    auto* ops = pruned->mutable_ops();
+    for (auto op_iter = ops->rbegin(); op_iter != ops->rend(); ++op_iter) {
```
sneaxiy: Why `rbegin`? It seems that `begin` is ok here.

python/paddle/fluid/backward.py
```diff
@@ -1209,9 +1209,8 @@ def append_backward(loss,
...
-        # _appending_grad_times used for double grad
-        program._appending_grad_times += 1
+    # Note(zhiqiu): This should be refined. 
+    program._appending_grad_times += 1
```
sneaxiy: If we always invoke `prune_backward` even with a forward network, I think that this changing can be removed. In fact, we need to remove `_appending_grad_times`  from `Program` (not in this PR) because: (1) it is  a bit hard coding; (2) this attribute cannot be serialized and deserialized.

paddle/fluid/framework/prune.cc
sneaxiy: Got it. The var may be from the the parent block or forward block. The code is right here.

python/paddle/fluid/backward.py
zhiqiu: Agreed, `_appending_grad_times` is not a good design.

paddle/fluid/framework/prune.cc
zhiqiu: Done, thanks.

paddle/fluid/framework/prune.cc
zhiqiu: Done，thanks.

paddle/fluid/framework/prune.cc
zhiqiu: Done, thanks.

paddle/fluid/framework/prune.cc
zhiqiu: Done, thanks.

paddle/fluid/framework/prune.cc
zhiqiu: Done, thanks.

python/paddle/fluid/framework.py
zhiqiu: Yes, I will change it.

paddle/fluid/framework/prune.cc
zhiqiu: Done, thanks.

paddle/fluid/framework/prune.cc
zhiqiu: Done, thanks.

paddle/fluid/framework/prune.cc
zhiqiu: Done, thanks.

python/paddle/fluid/backward.py
zhiqiu: I will revert this change, and I think  `_appending_grad_times` should be refined as we discussed offline.

python/paddle/fluid/tests/unittests/test_program_prune_backward.py
zhiqiu: Done, thanks.

paddle/fluid/framework/prune.cc
zhiqiu: I refined the logic here.

paddle/fluid/framework/prune.cc
zhiqiu: I add some comments for that.

sneaxiy (approved):
Cool.

PR #22661, opened 2020-02-18
unexpose tracer, test=develop

zhiqiu:
`Tracer` is an internal API, users do not need to use it directly when programing in dygraph mode.
Further more, it may be deprecated in future.
So, this PR make it unexposed.


phlrain (approved):

lanxianghit (approved):

swtkiwi (approved):
LGTM

saxon-zh (approved):

PR #22663, opened 2020-02-18
Fix pointer and c-api encapsulation

FrostML:
Before this fix, the `PD_Tensor`'s `PD_GetPaddleTensorShape` need a pointer to a pointer to get the shape's size. This PR correct this. 
Moreover, directly return shape.data() may not work properly every time and this problem is also fixed. 
Moreover, fix predictor cannot print profile message bug.


paddle/fluid/inference/capi/pd_tensor.cc
```diff
@@ -73,12 +73,14 @@ PD_PaddleBuf* PD_GetPaddleTensorData(const PD_Tensor* tensor) {
   return ret;
 }
 
-int* PD_GetPaddleTensorShape(const PD_Tensor* tensor, int** size) {
+int* PD_GetPaddleTensorShape(const PD_Tensor* tensor, int* size) {
   PADDLE_ENFORCE_NOT_NULL(tensor);
   std::vector<int> shape = tensor->tensor.shape;
```
liupluswei: 这里会有一次拷贝对吧，是不是可以换成引用的方式？

paddle/fluid/inference/capi/pd_tensor.cc
```diff
@@ -73,12 +73,14 @@ PD_PaddleBuf* PD_GetPaddleTensorData(const PD_Tensor* tensor) {
...
   int s = shape.size();
-  *size = &s;
-  return shape.data();
+  *size = s;
+  int* ret_shape = new int[s];
```
liupluswei: 建议上面换成引用之后(如果可以)直接return shape.data();
这里如果使用new int[]你得保证外面会主动去掉用delete来释放这块内存

paddle/fluid/inference/capi/pd_tensor.cc
FrostML: Done. Thanks.

paddle/fluid/inference/capi/pd_tensor.cc
FrostML: Done.Thanks.

liupluswei (dismissed):
我这块没什么问题了，因为也涉及预测库接口的事，也让预测组的相关同事review一下吧

fc500110 (approved):
LGTM

PR #22667, opened 2020-02-18
Change ernie accuracy results in README.md document

lidanqing-vv:
test=release/1.7
There is a error in the document of ernie. Here is the fix.
@bingyanghuang @luotao1


bingyanghuang:
Wrong branch to commit.

PR #22682, opened 2020-02-19
Fix batch_num default value

wojtuss:
With this change, the default behavior of the accuracy benchmarking scripts for QAT image classification and QAT NLP will be to use the whole dataset instead of a single iteration.

test=develop


Sand3r- (approved):

Sand3r-:
Won't this affect any UTs?

wojtuss:
> Won't this affect any UTs?

No. In UTs the `--batch_num` option is set to 2.

juncaipeng (approved):
LGTM

PR #22692, opened 2020-02-20
fix concat_mkldnn op. test=develop

jiweibo:
当启用mkldnn预测时，如果concat的多个输入中，有一个size为0，则实际concat_mkldnn kernel在Compute计算过程中其输入multi_input的实际size就会减一。如果预测多个图片，且该shape的key_prim已经设置过，则当遇到有输入size为0的情况，则再取输入的数据指针即multi_input[i]->data<T>()会core掉。此时应当初始化新的key_prim。

When mkldnn prediction is enabled, if the size of one of the inputs of concat is 0, the actual size of the input `multi_input` of the concat_mkldnn kernel during Compute calculation will be reduced by one. As shown in the figure below, although there are 4 inputs, only 3 are valid in the end.
If multiple sets of data are predicted continuously, and the `key_prim` of the shape has been set, when one of the multiple inputs has an input size of 0, then taking the input data pointer, which is `multi_input [i]-> data()` will failure. A new `key_prim` should be initialized at this point.

image

问题来源：卡片4155


wojtuss (approved):
LGTM

PR #22705, opened 2020-02-21
fix detection_map OP for class gt_num > 0 and true_pos_num = 0

heavengate:
**fix detection_map_op for class gt_num > 0 and true_pos_num = 0**
In detection_map OP, if gt number > 0 and true_positive prediction number = 0, this class will be skipped, but it should be calculated as a valid class, this bug may make mAP higher.
Normally, like in Pascal VOC dataset, predictions for a class would not be 0, this bug has no influence on released models currently


qingqing01 (dismissed):

jerrywgz (approved):
LGTM

PR #22709, opened 2020-02-21
Add IsControlFlowIfVisitor in ast_transformer

Aurelius84:
In this PR, `IsControlFlowIfVisitor` is added to distinguish `if/else` is plain python statement or control_flow statement.

There are some case should be considered as control flow `if/else`:
+ **case 1:** `if Tensor.numpy()[i] > 2` or `if fluid.layers.XXX(x).numpy()[i] > 2 `
+ **case 2:** `if Tensor.shape[i] > 2` where `Tensor.shape[i]` is `-1` or `None` in compile time.

Currently, we firstly support control flow statement   **case 1**:
+ involve `Tensor.numpy()[i]`

In **case 2**，Tensor.shape  is unknown in gast and should be inferred by other method. Moreover,  pred in ConditionalBlock require variable, which means all vars should be Tensor or transformed into Tensor, like `fill_constant(shape=[1], dtype='int32', value=Tensor.shape[i]).`

** Examples:**
```python
# node is not ast.Compare
node = gast.parse("a + b")
is_control_flow_if(node) == False

# node is not ast.Compare
node = gast.parse("a + x.numpy()[1]")
is_control_flow_if(node) == False

# node is not dependent on Tensor
node = gast.parse("x is None")
is_control_flow_if(node) == False

# `is None` is not dependent on Tensor.data
node = gast.parse("fluid.layers.sum(x) is None")
is_control_flow_if(node) == False

node = gast.parse("fluid.layers.sum(x).numpy() != None")
is_control_flow_if(node) == False

# .numpy() is Dygraph api and return shape=[1] boolean
node = gast.parse("x.numpy()[1] > 1")
is_control_flow_if(node) == True

# complex statement, but include control flow `if` in sub_expr
node = gast.parse("x is not None and 1 < x.numpy()[1]")
is_control_flow_if(node) == True

node = gast.parse("1 < fluid.layers.sum(x).numpy()[2] or x+y < 0")
is_control_flow_if(node) == True
```

Advanced features will be implemented while we test in real models.


python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
```diff
@@ -31,12 +31,60 @@
...
+        3. involve Tensor.shape[i] and the shape[i] is unknown in compile time.
+    The following examples should not be considered as control_flow_if:
+        1. `if Tensor_var` or `if Tensor_var is None`
+        2. if Tensor.shape[i] is determined with fixed value (not -1 or None)
+
+    Note: pred in ConditionalBlock require variable, which means all vars should be Tensor
+          or transformed into Tensor, like fill_constant(shape=[1], dtype='int32', value=Tensor.shape[i]).
+
+    TODO: 1. need to deal with `tensor.shape[i]` which need eval the data of shape[i],
+             because reshape_op maybe called before this statement.
```
liym27: typo:
need eval -> need to eval
 maybe -> may be.

liym27:
LGTM

liym27 (dismissed):

python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
Aurelius84: Fixed, thanks.

liym27 (approved):

python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
```diff
@@ -51,8 +51,9 @@ def ast_visit(self):
 
     def visit_If(self, node):
         assert isinstance(node, gast.If)
+        need_transform = is_control_flow_if(node.test)
         self.generic_visit(node)
-        if is_control_flow_if(node.test):
+        if need_transform:
```
zhhsplendid, line 56: Why split here？Is that due to now the generic_visit may change the node?

python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
```diff
@@ -27,16 +27,64 @@
...
-FALSE_FUNC_PRFIX = 'false_fn'
+TRUE_FUNC_PREFIX = 'true_fn'
+FALSE_FUNC_PREFIX = 'false_fn'
+
+
+class IsControlFlowIfVisitor(gast.NodeTransformer):
+    """
+    Judge whether the node.test from Dygraph code dependent on paddle Tensor.
+    If does, it should satisfy:
+        1. must involve at least one var whose type is Tensor.
```
zhhsplendid, line 38: Tiny English issues:

If you write period after number as list item such as ("1." , "2."), you should capitalize the first word since it is new sentence. If you don't like to capitalize, you can change "1." -> "1, "

So here you may change to:
    1. Must involve ... (this one is more common)
Or
    1, must involve ...

zhhsplendid (approved):
LGTM, you can change the English issue in this PR or in a future PR

python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
Aurelius84: Yes, generic_visit() will remove `numpy()` attribute in some case.

More Comments are needed here, and I will  add  them in next PR.

python/paddle/fluid/dygraph/dygraph_to_static/ast_utils.py
Aurelius84: Great advice! Got it. Thanks very much.

python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
zhhsplendid: That's OK to not add comment. I just thought whether we can inline the variable `need_transform`. You have explained that :-)

Issue #22636, opened 2020-02-17
【安装错误】Paddle源码编译时设置-DWITH_XBYAK=OFF，安装后import出错


zhangting2020:
## 问题描述
docker编译：镜像 hub.baidubce.com/paddlepaddle/paddle:latest-gpu-cuda9.0-cudnn7-dev 

- 使用最新的develop分支代码，源码编译，设置-DWITH_XBYAK=OFF，例如：
```
cmake .. -DPY_VERSION=2.7 -DWITH_GPU=ON -DWITH_TESTING=ON -DCMAKE_BUILD_TYPE=Release -DWITH_XBYAK=OFF
```
- 编译完成后安装，import 出错，如下：
```
λ paddle-gpu /workspace/Paddle/build {develop} python
Python 2.7.12 (default, Nov 12 2018, 14:36:49)
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import paddle.fluid as fluid
grep: warning: GREP_OPTIONS is deprecated; please use an alias or script
F0217 05:33:03.666159 22983 init.cc:195] This version is compiled on higher instruction(AVX) system, you may encounter illegal instruction error running on your local CPU machine. Please reinstall the NonAVX version or compile from source code.
*** Check failure stack trace: ***
    @     0x7fe444bf525d  google::LogMessage::Fail()
    @     0x7fe444bf75a8  google::LogMessage::SendToLog()
    @     0x7fe444bf4d6b  google::LogMessage::Flush()
    @     0x7fe444bf847e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7fe444d17a6d  paddle::framework::InitDevices()
    @     0x7fe444d17b7c  paddle::framework::InitDevices()
    @     0x7fe4448a69f2  _ZZN8pybind1112cpp_function10initializeIZN6paddle6pybindL22pybind11_init_core_avxERNS_6moduleEEUlbE107_vJbEJNS_4nameENS_5scopeENS_7siblingEEEEvOT_PFT0_DpT1_EDpRKT2_ENUlRNS_6detail13function_callEE1_4_FUNESN_
    @     0x7fe4449322f1  pybind11::cpp_function::dispatcher()
    @           0x4bc4aa  PyEval_EvalFrameEx
    @           0x4b9b66  PyEval_EvalCodeEx
    @           0x4c1f56  PyEval_EvalFrameEx
[...]

PR #22716, opened 2020-02-21
Add cpu_info without XBYAK

grygielski:
Add functionality inside `MayIUse()` function allowing to check for AVX flags when Paddle is built without XBYAK.

### DIFF
--- a/paddle/fluid/platform/cpu_info.cc
+++ b/paddle/fluid/platform/cpu_info.cc
@@ -139,6 +139,31 @@ bool MayIUse(const cpu_isa_t cpu_isa) {
   if (cpu_isa == isa_any) {
     return true;
   } else {
+    int reg[4];
+    cpuid(reg, 0);
+    int nIds = reg[0];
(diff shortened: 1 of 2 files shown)


wojtuss:
This PR fixes https://github.com/PaddlePaddle/Paddle/issues/22636

grygielski:
@luotao1 Could you advice on failing CI_GPU_Py2_GCC8 or just rerun if this is random fail?

luotao1:
I rerun CI_GPU_Py2_GCC8 since it fails on the network.

wojtuss (approved):
LGTM

jczaja (commented):
It is great, but There is no detection of VNNI . Don't we need that ?

grygielski:
@jczaja There is no need for checking for VNNI currently in Paddle, hence I didn't add it. But it's great idea of improvement.

jczaja:
@grygielski Ok, thanks for answer. LGTM

jczaja (approved):

PR #22747, opened 2020-02-24
Add Basic Function Return Type Analysis

zhhsplendid:
1. Considering functions, I have to change the node type from single value to a set. Because python function is allowed to return different types. The set represent all possible types

2. I added scope_name and scope_type for AstVarScope, because in python functions, variable may have different scope. For example:

```
a = 3
def foo(b):
    a = 9
    return a + b
```

the `a` in `foo` is different to the `a` out of `foo`. Similar to class field. The scope_name will help me to know the function name when static analysis finds a `return` sentence.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
@@ -34,11 +34,12 @@ def _is_api_in_module_helper(obj, module_prefix):
 # function code together when Yamei finish her PR.
 def is_api_in_module(node, module_prefix):
     assert isinstance(node, gast.Call), "Input non-Call node for is_dygraph_api"
-    func_src = astor.to_source(node.func)
+    func_str = astor.to_source(node.func)
     try:
         import paddle.fluid as fluid
-        return eval("_is_api_in_module({}, '{}')".format(func_src,
-                                                         module_prefix))
+        import paddle
+        return eval("_is_api_in_module_helper({}, '{}')".format(func_str,
+                                                                module_prefix))
     except NameError:
         return False
 
@@ -57,8 +58,8 @@ def is_numpy_api(node):
     func_str = astor.to_source(node.func)
     try:
         import numpy as np
-        module_result = eval("_is_api_in_module({}, '{}')".format(func_src,
-                                                                  "numpy"))
+        module_result = eval("_is_api_in_module_helper({}, '{}')".format(
+            func_str, "numpy"))
         # BUG: np.random.uniform doesn't have module and cannot be analyzed
         # TODO: find a better way
         if not module_result:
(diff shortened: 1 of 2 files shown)


zhhsplendid:
Close and re-open to trigger PR_CI_Inference

python/paddle/fluid/tests/unittests/test_ast_transformer_static_analysis.py
```diff
@@ -81,15 +81,48 @@ def func_to_test4():
...
+
+    a = inner_int_func()
+    b = inner_bool_float_func(3)
+    c = inner_unknown_func(None)
+    d = paddle.fluid.data('x', [1, 2])
+
+
+result_var_type5 = {
+    'a': {NodeVarType.INT},
+    'b': {NodeVarType.FLOAT, NodeVarType.BOOLEAN},
```
liym27, line 112: Why the var type of `b` is not `{NodeVarType.FLOAT}` because `b` is 1.0

python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
```diff
@@ -25,34 +25,41 @@
...
+    return m is not None and m.__name__.startswith(module_prefix)
 
 
 # TODO: is_dygraph_api is duplicated in Yamei's utils.py. Merge the two
 # function code together when Yamei finish her PR.
-def is_dygraph_api(node):
+def is_api_in_module(node, module_prefix):
     assert isinstance(node, gast.Call), "Input non-Call node for is_dygraph_api"
-    func_src = astor.to_source(node.func)
+    func_str = astor.to_source(node.func)
```
liym27: Better to use `astor.to_source(gas.gast_to_ast(node.func))`.
Because `astor.to_source`  converts a ast.node but `node.func` is gast.node. Sometimes gast.node is different from ast.node and it will fail to convert a gast.node to source.

python/paddle/fluid/tests/unittests/test_ast_transformer_static_analysis.py
zhhsplendid: Static analysis didn't run the python program. So the b's type is possible FLOAT or BOOLEAN before running.

zhhsplendid: Or in another word, it is hard work to get the type by analyzing the control flow here :-) we can do more optimization work if we find it is required in the real model

python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
zhhsplendid: Done

liym27 (approved):

liym27:
LGTM

zhhsplendid:
Close and reopen to trigger inference CI

PR #22760, opened 2020-02-25
Add @skip_check_grad_ci to UT conv_mkldnn and int8_mul_mkldnn

lidanqing-intel:
Use `@skip_check_grad_ci` decorator to skip `check_grad` tests for `test_conv2d_mkldnn_op.py` and `test_int8_mul_mkldnn_op.py`

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_conv2d_op.py
+++ b/python/paddle/fluid/tests/unittests/test_conv2d_op.py
@@ -355,7 +355,8 @@ class TestConv2dOp(OpTest):
             place, atol=1e-5, check_dygraph=(self.use_mkldnn == False))
 
     def test_check_grad(self):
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/tests/unittests/test_conv2d_op.py
```diff
@@ -355,7 +355,7 @@ def test_check_output(self):
             place, atol=1e-5, check_dygraph=(self.use_mkldnn == False))
 
     def test_check_grad(self):
-        if self.dtype == np.float16:
+        if self.dtype == np.float16 or hasattr(self, "no_need_check_grad"):
```
wojtuss: please also make sure the value of the `no_need_check_grad` attribute ist `True`
here, and below

lidanqing-vv: Done.

PR #22766, opened 2020-02-26
Add program_cache in dygrapht_to_static

Aurelius84:
### Adds ProgramCache

In this PR, `FunctionCache` and `ProgramCache` are added to avoid transforming code of dygrah function by AST analyze and  inserting ops into main_program, respectively.

That means program will be built once, even users put `myNet(input)`  into `for` loop.
 
**code example:**
```python
def train_mnist():
    with fluid.program_guard(fluid.Program()):
        mnist = MNIST()

        for epoch in range(self.epoch_num):
            for batch_id, data in enumerate(self.train_reader()):
                dy_x_data = np.array([
                    x[0].reshape(1, 28, 28) for x in data
                ]).astype('float32')
                y_data = np.array([x[1] for x in data
                                    ]).astype('int64').reshape(-1, 1)
                # fetch outputs dynamically as dygraph model.
                # Program will be built only once.
                prediction, acc, avg_loss = mnist(dy_x_data, label=y_data)
```

Moreover, code related to `exe.run` is moved into `ProgramCache` to allow users specifying `optimizer` which used to minimize loss.


python/paddle/fluid/dygraph/dygraph_to_static/cache_program.py
```diff
@@ -0,0 +1,292 @@
...
+
+    @synchronized
+    def __new__(cls, *args, **kwargs):
+        if cls._instance is None:
+            cls._instance = object.__new__(cls, *args, **kwargs)
+            cls._instance.__initialized = False
+        return cls._instance
+
+    @classmethod
+    def getInstance(cls):
```
zhhsplendid: Rename it to get_instance to match our code style?

python/paddle/fluid/dygraph/dygraph_to_static/cache_program.py
```diff
@@ -0,0 +1,292 @@
...
+__all__ = ['ProgramCache']
+
+
+class FunctionCache(object):
+    """
+    Caches the transformed functions to avoid redundant conversions of the same function.
+    """
+
+    def __init__(self):
+        self._cache_funcs = collections.OrderedDict()
```
zhhsplendid: Why do we use OrderedDict here? (usually OrderedDict is implemented as balanced tree which means query and update can use O(log N) while a Dict which is usually implemented by hash table usually takes average O(1)) Can we just use dict or DefaultDict?

python/paddle/fluid/dygraph/dygraph_to_static/cache_program.py
```diff
@@ -0,0 +1,292 @@
...
+                    args[idx] = feed_layer
+                self.is_build_process = False
+                fetch_list = func(*args, **kwargs)
+                self._add_optimizer(fetch_list)
+            else:
+                fetch_list = func(*args, **kwargs)
+        self.outputs = fetch_list
+        return fetch_list
+
+    def _run(self, args, kwargs):
```
zhhsplendid: As we have discussed, the run logic had better to be extract out. Let's make the cache only does cache thing, check valid, query and get, reset. Move the logic of adding/modifying programs to somewhere else. That's Object Oriented Programming (OOP) :-)

Then it will increase readability and people with cache background knowledge can read it easier.

python/paddle/fluid/dygraph/dygraph_to_static/cache_program.py
```diff
@@ -0,0 +1,292 @@
...
+        input_names = [in_var.name for in_var in self.inputs]
+        exe = executor.Executor(core.CPUPlace())
+        if self.need_startup:
+            self.need_startup = False
+            exe.run(framework.default_startup_program())
+        feed_dict = dict(zip(input_names, self.batch_data))
+        res = exe.run(self.program, feed=feed_dict, fetch_list=self.outputs)
+        return res
+
+    def set_optimizer(self, optimizer, force_update=True):
```
zhhsplendid: As we discussed in our group chat with Yamei, this part `set_optimizer` and `add_optimizer` is not natural. For Paddle user, they may just know program, and when calling your program cache to set_optimizer, 1. they may feel strange. 2. Here you still didn't set optimizer to real program, it's more like a "pre-set".

Let's wait you to move run logic out or cache and we will see you how to handle this case.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_cache_program.py
```diff
@@ -0,0 +1,115 @@
...
+class TestCacheProgramWithOptimizer(unittest.TestCase):
+    def setUp(self):
+        self.dygraph_class = Linear
+        self.data = np.random.random((4, 10)).astype('float32')
+        self.batch_num = 5
+
+    def train_static(self):
+        main_program = fluid.Program()
+        loss_data = []
+        with fluid.program_guard(main_program):
```
zhhsplendid, line 75: I can understand that you have to pre-set optimizer and append it to real program only at run time. (because you can only add data feed layer then optimizer) you are doing right thing but API interface name is not natural as we have discussed.

Suggestion:

Make ProgramCache just do cache thing. Use another class to do those whole control thing out of cache (your run logic), like "DygraphToStaticController" or other name if you have a better one, where  your can put running logic (but I am thinking whether now we should give our dygraph to static module a name like AutoGraph for TF, Scripting for PyTorch)

For a Controller class, it can now have method "set_optimizer", much natural.

python/paddle/fluid/dygraph/dygraph_to_static/cache_program.py
Aurelius84: Thanks, fixed

python/paddle/fluid/dygraph/dygraph_to_static/cache_program.py
Aurelius84: Great advice. Fixed.

python/paddle/fluid/dygraph/dygraph_to_static/cache_program.py
Aurelius84: That's right. Here we need a better way to expose the interface that allows users to set optimizer easily. I think maybe we can bind the `set_optimizier` to params `self` of decorator. I will continue to explore new approaches.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_cache_program.py
Aurelius84: Thanks. I add a new class `AutoTracer` to handle this problem.

liym27:
Great job! LGTM

liym27 (approved):

zhhsplendid (approved):
LGTM, we will discuss some interfaces after project design discussion

Boyan-Liu (approved):

PR #22768, opened 2020-02-26
Fix inference c api PD_GetZeroCopyOutput lod

FrostML:
Fix inference c api `PD_GetZeroCopyOutput` obtaining lod.
Before this fix, if the lod of the output is empty, `lod.front()` will core when get output of the predictor. 
This PR fix this this problem. 
Moreover, add unit tests for new c api tests.


fc500110 (approved):
LGTM

PR #22786, opened 2020-02-27
Update the precision of the tanh op test from float32 to float64

wawltor:
Update the precision of the tanh op test, remove from the white list


wawltor:
**Tanh在inplace的单测下会出现前向精度问题，环境windows CI** 
出现diff情况如下
image

PR #22787, opened 2020-02-27
update the precision and big shape of matmul op test

wawltor:
update the precision and big shape of matmul op test, update the precision from float32 to float64, and update the big shape

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_matmul_op.py
+++ b/python/paddle/fluid/tests/unittests/test_matmul_op.py
@@ -16,16 +16,13 @@ from __future__ import print_function
 
 import unittest
 import numpy as np
-from op_test import OpTest
+from op_test import OpTest, skip_check_grad_ci
 import paddle.fluid as fluid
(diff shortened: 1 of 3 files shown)


wawltor:
**matmul在大shape的单测精度问题**
目前发现matmul在计算输入x和y的shape为1 * 100， 100 * 1的时候，或者是x和y的shape为1， 100 *  1的时候,  数值梯度和op计算梯度的diff会出现超过1e-7的情况 
下面是出现梯度diff的情况
image

lidanqing-vv:
@luotao1  Hi, this is not mkldnn op right? Should intel also look at it? We are going to add matmul_mlkdnn_op in 2.0

GaoWei8:
> @luotao1 Hi, this is not mkldnn op right? Should intel also look at it? We are going to add matmul_mlkdnn_op in 2.0

@lidanqing-intel  We have constructed one test case in #22944 when dim=1, K>=100. Could you help see it?

PR #22486, opened 2020-02-07
[Ernie GPU Optim]: Fuse three fc to multihtead matmul

NHZlX:
Fuse three Fc to mutlithead matmul op and combined three Fc to one Fc.
Under cuda10.1 p4 card, the normal ernie model performance is improved from 10ms to 9.5ms.

Original:

image

Now:

image

PR #22805, opened 2020-02-28
[Paddle-TRT ERNIE][WIP]Optimize multihead matmul

cryoco:
Optimize multihead matmul of ERNIE according to [PR#22486](https://github.com/PaddlePaddle/Paddle/pull/22486)

TODO:
1. use cublas api to compute matmul;
2. detect precision in trt plugin to use fp16 mode;
3. move matmul functions in multihead_matmul_op.cu to operators/math for reusability.


jeng1220:
@NHZlX ,
looks like this patch can improve ERNIE performance, doesn't? Why did no one review this PR?

cryoco:
> @NHZlX ,
> looks like this patch can improve ERNIE performance, doesn't? Why did no one review this PR?

The work in this PR is already done by @NHZlX . It should be closed. Thanks.

PR #22814, opened 2020-03-02
Cleanup mmap files in special unittest

chenwhql:
`test_imperative_data_loader_process.py` was added to solve the problem that the coverage CI cannot cover the sub-process code, so it is not the normal use method of DataLoader, and no multi-process is used. 

Therefore, after adding the shared memory mechanism, the mmap files in this unittest need to be manually cleaned.


sneaxiy (approved):
LGTM.

PR #22826, opened 2020-03-03
[Cherry-pick] Fix inference c api PD_GetZeroCopyOutput lod and c api encapsulation

FrostML:
Fix inference c api PD_GetZeroCopyOutput lod and c api encapsulation. 

Before this fix, the PD_Tensor's PD_GetPaddleTensorShape need a pointer to a pointer to get the shape's size. This PR correct this.
Moreover, directly return shape.data() may not work properly every time and this problem is also fixed.
Moreover, fix predictor cannot print profile message bug.
Fix inference c api PD_GetZeroCopyOutput obtaining lod.

Before this fix, if the lod of the output is empty, lod.front() will core when get output of the predictor.
This PR fix this this problem.
Moreover, add unit tests for new c api tests.

Cherry-pick #22663 and #22768 .


fc500110 (approved):
LGTM

PR #22835, opened 2020-03-03
support get pslib version

xjqbest:
*  support get pslib version
*  usage:
```
>>> import paddle.fluid.incubate.fleet.parameter_server.pslib.version as pv
>>> pv.show()
```

image


yaoxuefeng6 (approved):
LGTM

PR #22837, opened 2020-03-03
Modify the way of inserting newly defined func_nodes

Aurelius84:
In this PR, we changed the way of inserting newly defined `functionDef` nodes into AST root.

Instead of inserting all the func_nodes into outer body of AST root, we used nested structure by adding func_nodes in front of the current `if/else` statement. 

**Pros:**
+ More clearly to understand the relationship between `cond` layer and `true_func/false_func`
+ In some cases, it's easier to deal with the parameter `self` scope problem in Class. Because private variable with `self` can't be visited friendly if insert func_nodes into AST root.

**Raw code:**:
```python
def foo():
    if  control_flow_if:
        out= do_something
    else:
        out = do_nothing
    return out
```

**Before this PR:**
```python
def true_func():
    return do_something

def false_func():
    return do_nothing

def foo():
    out = layers.cond(control_flow_if, true_func, false_func)
    return out
```

**In this PR:**
```python
def foo():
    def true_func():
        return do_something

    def false_func():
        return do_nothing

    out = layers.cond(control_flow_if, true_func, false_func)
    return out
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
@@ -15,7 +15,6 @@
 from __future__ import print_function
 from .utils import *
 import gast
-import six
 # gast is a generic AST to represent Python2 and Python3's Abstract Syntax Tree(AST).
 # It provides a compatibility layer between the AST of various Python versions,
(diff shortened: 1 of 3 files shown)


liym27:
LGTM.

liym27 (dismissed):

liym27 (approved):

PR #22847, opened 2020-03-04
Fix NameVisitor bugs

zhhsplendid:
1. copy.deepcopy in NameVisitor should be changed to copy.copy to make hash or set work
2. read_context should be type of gast.Load()/gast.AugLoad(), not gast.Load/gast.AugLoad


liym27 (approved):

liym27:
LGTM

PR #22848, opened 2020-03-04
[BUG]: Multihead matmul op's ouput size should be BxSx(N*H)

NHZlX:
We set the same dims of  Multihead matmul op 's input to output original.

eg:
```
 auto dim_input = context->GetInputDim("Input");
 context->SetOutputDim("Out", dim_input);
 context->ShareLoD("Input", /*->*/ "Out");
```
https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/fused/multihead_matmul_op.cc#L80

The setting above is wrong. 

Let's first introduce some noun variables in Multihead matmul op:
```
1) Batch : the batch size.
2) SeqLen: the sequence length.
3) HiddenSize: the second dims size of the Embedding table.
4) HeadNumber: the head number of the encoder.
5) HeadSize: the size of each Head.
```
The Multihead matmul op's Input Dims is [Batch * SeqLen * HiddenSize], the Output Dim should be [Batch * SeqLen * (HeadNumber * HeadSize)].

### DIFF
--- a/paddle/fluid/operators/fused/multihead_matmul_op.cu
+++ b/paddle/fluid/operators/fused/multihead_matmul_op.cu
@@ -440,13 +440,11 @@ class MultiHeadMatMulV2Kernel : public framework::OpKernel<T> {
 
     auto &bias_qk = detail::Ref(context.Input<framework::Tensor>("BiasQK"),
                                 "Cannot find QK");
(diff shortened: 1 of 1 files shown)


zhaoyuchen2018 (approved):
LGTM

PR #22854, opened 2020-03-04
Add unittest for dict in dygraph_to_static

Aurelius84:
In this PR, unittest with `subNet` called by `MainNet` is added for testing usage of python `dict`, which maybe hold some `Tensor` or `Variable` and can be modified everywhere.


liym27 (approved):

liym27:
LGTM

PR #22857, opened 2020-03-04
Prevent loading of warmup data in analyzer_int8 if enable_int8 is set to false

Sand3r-:
* Prevent loading of warmup data in analyzer_int8 if `enable_int8` is set to `false`
* Improve message about `warmup_batch_size` being incorrectly set. Since its default value is set to 100 users who didn't set it manually, and set their batch size to be lower than 100 might end up being confused.


wojtuss (approved):
LGTM

PR #22858, opened 2020-03-04
Added an option to use external FP32 model in QAT comparison test

wojtuss:
With this patch, when running the Ernie QAT INT8 comparison test, a user can use the `--fp32_model` option to provide an FP32 model for accuracy measurement and comparison with an INT8 model obtained from the QAT model (passed by `--qat_model` option). If the `--fp32_model` option is not used (the default in QAT INT8 tests), the QAT model accuracy will be measured and compared.

It also adds printing average accuracy result after each (FP32 and INT8) inference run.

test=develop

### DIFF
--- a/python/paddle/fluid/contrib/slim/tests/qat_int8_nlp_comparison.py
+++ b/python/paddle/fluid/contrib/slim/tests/qat_int8_nlp_comparison.py
@@ -48,6 +48,12 @@ def parse_args():
     parser.add_argument(
         '--qat_model', type=str, default='', help='A path to a QAT model.')
     parser.add_argument(
+        '--fp32_model',
+        type=str,
+        default='',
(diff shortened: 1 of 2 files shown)


lidanqing-vv:
LGTM

python/paddle/fluid/contrib/slim/tests/qat_int8_nlp_comparison.py
```diff
@@ -251,6 +256,7 @@ def test_graph_transformation(self):
 
         _logger.info('QAT FP32 & INT8 prediction run.')
```
lidanqing-vv, line 257: Here it is not necessarily QAT FP32, it can also be FP32 original. Maybe we can set it a variable and pass into this log.

juncaipeng (approved):
LGTM

Maybe it is necessary to use FP32 model instead of QAT model in unittest   (TODO)

PR #22866, opened 2020-03-05
Tensor.shape support control flow if/for/while and bugfix

liym27:
1. Add support for Tensor.shape in control flow conditions of if/while/for;
2. fix bug of if/while transfoemer;
3. Separate TensorShapeTransformer from BasicApiTransformer.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
@@ -20,7 +20,7 @@ import gast
 from collections import defaultdict
 from paddle.fluid import unique_name
 from paddle.fluid.dygraph.dygraph_to_static.utils import generate_name_node
-from paddle.fluid.dygraph.dygraph_to_static.static_analysis import AstNodeWrapper
+from paddle.fluid.dygraph.dygraph_to_static.static_analysis import AstNodeWrapper, StaticAnalysisVisitor
 from paddle.fluid.dygraph.dygraph_to_static.utils import get_constant_variable_node
 from paddle.fluid.dygraph.dygraph_to_static.utils import get_attribute_full_name
 from paddle.fluid.dygraph.dygraph_to_static.variable_trans_func import create_static_variable_gast_node
@@ -79,6 +79,10 @@ class NameVisitor(gast.NodeVisitor):
         # Mapping from gast.While/gast.For to variable nodes
         self.before_loop_body_vars = defaultdict(set)
         self.in_loop_vars = defaultdict(set)
+
+        self.static_analysis_visitor = StaticAnalysisVisitor(root_node)
+        self.node_to_wrapper_map = self.static_analysis_visitor.get_node_to_wrapper_map(
+        )
 
         self.visit(root_node)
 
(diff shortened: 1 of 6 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
```diff
@@ -160,8 +164,18 @@ def _var_nodes_to_names(self, node_set, ctx_filter_set=None):
         for node in node_set:
             if ctx_filter_set is None or type(node.ctx) in ctx_filter_set:
                 if isinstance(node, gast.Name):
+                    parent_node = self.node_to_wrapper_map[node].parent.node
```
zhhsplendid: This condition is very similar to my `_is_call_func_name_node`, and I found my bug may be line 184:

parent_node = self.ancestor_nodes[-1] -> parent_node = self.ancestor_nodes[-2] 

The self.ancestor_nodes and _is_call_func_name_node are doing same thing as your condition. So I would recommend to remove duplicate codes. I suggest two options for you:

Option 1: fix the bug I mentioned at line 184, remove the code you added to see whether the bug is fixed. In this way we don't make static analysis in this Transformer (**Program may be slightly faster**).

Option 2: keep the code you added and help me to remove self.ancestor_nodes, _is_call_func_name_node, in this way the code is **safer** because the parent_node is analyzed by static_analysis.

Both are OK to me, you can pick one. Personally I slightly prefer Option2

python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
```diff
@@ -313,6 +313,10 @@ def _get_node_var_type(self, cur_wrapper):
             return self.var_env.get_var_type(node.id)
 
         if isinstance(node, gast.Return):
+            # If return nothing:
+            if node.value is None:
+                return {NodeVarType.NONE}
```
zhhsplendid, line 318: Great code here! Thank you for considering this case!

python/paddle/fluid/tests/unittests/dygraph_to_static/test_tensor_shape.py
```diff
@@ -58,17 +58,116 @@ def dyfunc_tensor_shape_5(x):
...
+    while x_shape_0 > i:
+        res += 1
+        i = i + 2
+    return res
 
 
-class TestTensorShape(unittest.TestCase):
+def dyfunc_with_while_3(x):
+    # TODO(liym27):
+    #  It will fail to run because `while len(x_shape) < i` will be transformed into Paddle while_loop and `x_shape` will be loop_vars.
```
zhhsplendid: Is here still error? I think x_shape should be one of loop vars, because it is in the condition

python/paddle/fluid/tests/unittests/dygraph_to_static/test_tensor_shape.py
```diff
@@ -58,17 +58,116 @@ def dyfunc_tensor_shape_5(x):
...
+    x_shape_0 = x.shape[0]
+    res = fluid.layers.fill_constant(value=0, shape=[1], dtype="int32")
+
+    # `x_shape_0` is transformed into `fluid.layers.shape(x)[0]`
+    for i in range(x_shape_0):
+        res += 1
+    return res
+
+
+def dyfunc_with_for_3(x):
```
zhhsplendid, line 107: Please also explain why this will fail when `for i in range` is transformer into Paddle while_loop

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
liym27: OK. I also prefer Option2. I have removed  self.ancestor_nodes.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_tensor_shape.py
liym27: Here the python for-loop in `for i in range(len(x.shape))` will be transformed into Paddle `while_loop`, and the python list `x.shape` will be added to `loop_vars`. However, `loop_vars` doesn't support python list.
And the condition of `for i in range(len(x.shape))`  only uses the length of x.shape, so it doesn't have to be transformed into Paddle `while_loop`.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_tensor_shape.py
liym27: Same as above. x_shape is a python list. But here the while-loop shouldn't be transformed into `while_loop`.

PR #22867, opened 2020-03-05
Support Simple For Range Loop in Dygraph to Static

zhhsplendid:
1. Add basic support for `for in range` loop
2. Move `test_dygraph_to_static_*` to `dygraph_to_static` dir and rename them
3. Add test case for dict in while_loop


zhwesky2010 (dismissed):
No unittest deletions, just rename and changes path

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
```diff
@@ -215,6 +215,23 @@ def create_api_shape_node(tensor_shape_node):
...
+        node,
+        gast.Attribute), "Input non-Attribute node to get attribute full name"
+
+    if isinstance(node.value, gast.Name):
+        return node.value.id + "." + node.attr
+    elif isinstance(node.value, gast.Constant):
+        return str(node.value.value) + "." + node.attr
+    return get_attribute_full_name(node.value) + "." + node.attr
+
+
```
liym27, line 229: One question:
Can this function `get_attribute_full_name` be replaced by it:
```
def get_attribute_full_name(node):
    assert isinstance()
    return astor.to_source(gast.gast_to_ast(node)).strip()
```

liym27:
LGTM

liym27 (dismissed):

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
```diff
@@ -111,15 +125,29 @@ def visit_Name(self, node):
             self.in_loop_vars[loop_node].add(node)
         self.generic_visit(node)
 
+    def visit_Attribute(self, node):
+        attr_full_name = get_attribute_full_name(node)
+        if callable(attr_full_name):
```
Aurelius84: Type of `attr_full_name ` is `str` and `callable(str)` will always return `False`. Is this as expected？

Aurelius84 (commented):
LGTM.

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
zhhsplendid: My fault, should be eval(attr_full_name)

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
zhhsplendid: Sure

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
zhhsplendid: Found that eval(attr_full_name) may also have problem such as attr_full_name = "self.XXX", please check my new solution, which checks whether the gast.Name/gast.Attribute node is at `func` field of gast.Call

liym27 (approved):

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
```diff
@@ -106,23 +113,65 @@ def get_loop_var_names(self, node):
         return loop_var_names, create_var_names
 
     def visit_Name(self, node):
+        if self._is_call_func_name_node(node):
+            self.generic_visit(node)
```
Aurelius84, line 117: 👍🏻👍🏻
Whether `gast.Name` is a leaf node in AST.  If does, it's no need to call `self.generic_visit`. I'm not sure for that and it's just a small question.😁

Aurelius84 (approved):

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
zhhsplendid: Considering gast.Name has annotation as gast.Expr, let's keep visiting.

image

PR #22878, opened 2020-03-05
Polish detail implement of dygraph data loader

chenwhql:
as title:
- enhance error message
- polish function name
- add except branch
- polish unittests


zhwesky2010 (approved):
only modify unitest name, not delete

sneaxiy (approved):
LGTM.

PR #22880, opened 2020-03-05
Fix the elementwise_min doc api bug, correct the api used 

wawltor:
In elementwise_min api doc has the error, all the api just use the  elementwise_max op, we just correct the doc.


Boyan-Liu (approved):

FlyingQianMM (approved):
LGTM

PR #22884, opened 2020-03-05
Support AucRunner in PaddleBox

hutuxian:
"Auc Runner" is a method to evaluate feature importance.
The feature value of slots that are to be evaluated will be replaced by history values, which are selected by reservoir sampling.
For performance reason, we use multi-thread technology to accelerate the evaluating process, and maintain multiple pools for each thread.
Furthermore, we update some variable names that are not proper before.

### DIFF
--- a/paddle/fluid/framework/fleet/box_wrapper.h
+++ b/paddle/fluid/framework/fleet/box_wrapper.h
@@ -471,16 +471,16 @@ class BoxWrapper {
    public:
     MetricMsg() {}
     MetricMsg(const std::string& label_varname, const std::string& pred_varname,
-              int is_join, int bucket_size = 1000000)
+              int metric_phase, int bucket_size = 1000000)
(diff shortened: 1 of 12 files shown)


paddle/fluid/framework/fleet/box_wrapper.h
```diff
@@ -753,7 +778,8 @@ class BoxWrapper {
   std::unordered_set<std::string> slot_name_omited_in_feedpass_;
 
   // Metric Related
-  int pass_flag_ = 1;  // join: 1, update: 0
+  int phase_ = 1;
+  int phase_num_ = 2;
```
gavin1332, line 782: auc runner runs on one pass (one dataset) which is split into multiple phase.

gavin1332 (dismissed):
LGTM

gavin1332 (approved):
LGTM

PR #22888, opened 2020-03-05
Fix dgc param regularizer

wangxicoding:
Every parameter may have its own regularizer, which has a higher priority than the default. Fixes the problem of the parameter's regularizer not processed by DGC.


gongweibao:
LGTM

gongweibao (approved):
LGTM

PR #22891, opened 2020-03-06
Fix model int8 quant fail, test=develop

zhaoyuchen2018:
As model fails when enable int8 quant, so disable allocate memory in cpu
for small variable.


juncaipeng (approved):
LGTM

对常见分类模型进行量化训练，测试该问题修复。

PR #22914, opened 2020-03-09
Close fuse when using dgc & move DGC strategy from PE to compiler

wangxicoding:
1. Build strategy *fuse_all_reduce_ops* require that gradients should not be sparse types, close fuse when using DGC. 
2. Move program._enable_dgc strategy from parallel_executor.py to the compiler.py for the compiler.py is the final entry.


gongweibao (approved):
LGTM

sneaxiy (approved):

PR #22915, opened 2020-03-09
Skip type/dtype check in dygraph mode

chenwhql:
Why skip dynamic graph check:
1. If the input type / dtype of a layer is wrong, it will be reported directly on that line. User can easily print the relevant information on which line. It is easier to debug, so there is no need to check in dynamic graph mode.
2. Performance considerations. Because these checks are executed at each step in dynamic graph mode, it will bring a heavy performance burden.


zhiqiu (approved):
LGTM

PR #22916, opened 2020-03-09
support Tensor.shape in control_flow_if

Aurelius84:
### support Tensor.shape in control_flow_if

**Example code:**
```python
def foo(x):
    batch_size = fluid.layers.shape(x)
    if batch_size[0] > 16:
        x = x + 1
    return x
```

```python
def foo(x):
    if fluid.layers.shape(x)[0] > 16:
        x = x + 1
    return x
```

Currently, we can identify more cases that need to transform into `layers.cond`. But now the case with `and` / `or` is not supported and will be added in next PR.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
@@ -46,11 +46,11 @@ class IfElseTransformer(gast.NodeTransformer):
             wrapper_root, AstNodeWrapper
         ), "Type of input node should be AstNodeWrapper, but received %s ." % type(
             wrapper_root)
-        self.wrapper_root = wrapper_root
         self.root = wrapper_root.node
+        self.static_analysis_visitor = StaticAnalysisVisitor(self.root)
         self.new_func_nodes = {}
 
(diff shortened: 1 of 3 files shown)


liym27:
This PR is more like improving `is_control_flow_if `. More tests are needed in file `test_dygraph_to_static_basic.py` to test the transformation of control flow if.

Aurelius84:
> This PR is more like improving `is_control_flow_if `. More tests are needed in file `test_dygraph_to_static_basic.py` to test the transformation of control flow if.

The transformation with complicate `and`、`or` in node.test which includes Tensor and non-Tensor will be added in next PR.

zhhsplendid (commented):
LGTM，let's wait for Yamei's comment

liym27:
> > This PR is more like improving `is_control_flow_if `. More tests are needed in file `test_dygraph_to_static_basic.py` to test the transformation of control flow if.
> 
> The transformation with complicate `and`、`or` in node.test which includes Tensor and non-Tensor will be added in next PR.

OK 🙆‍♂️

liym27:
LGTM

liym27 (approved):

PR #22919, opened 2020-03-09
fix communicator when break under pyreder mode (#22911)

seiriosPlus:
* fix communicator when breaking under PyReader mode, test=develop
* revert some vlog level to 0, test=develop


123malin (approved):
LGTM

PR #22924, opened 2020-03-09
fix add grad bug test=develop

yaoxuefeng6:
fix add grad before append_bacward bug, which can result in no sparse grads in fleet desc file.


xjqbest (approved):
LGTB now，and it should later support multi loss share botttom scenario

PR #22932, opened 2020-03-10
Add Unittest for backward of fusion group

wangchaochaohu:
1.We add backward for the fusion group fuction in  unittest 
2.polish code of codegen in fusion group


zhaoyuchen2018 (approved):
LGTM

PR #22968, opened 2020-03-11
refine the profiler print 

wangchaochaohu:
The AllOpDetail Option print call time it not correct, refine it and the result is like This:

image


zhangting2020 (approved):
LGTM

PR #22976, opened 2020-03-12
Add transpose_flatten_concat_fuse_pass tests for gpu and trt

FrostML:
Add inference transpose_flatten_concat_fuse_pass tests for GPU and TRT to check the number of output and precision on GPU and enable TRT. 
CPU do not run this pass. 
* GPU:
image
* TRT:
image


fc500110 (approved):
LGTM

Issue #22964, opened 2020-03-11
Small performance drop on Resnets


grygielski:
Merging https://github.com/PaddlePaddle/Paddle/pull/22695 introduced small overhead in convolutions with residual connection. It resulted in performance drop on all Resnet models:

**Performance change [FPS]**

*CLX platform:*
```
resnet50_int8     47.97 -> 46.60
resnet50_fp32     13.28 -> 12.01
resnet101_fp32    7.29 -> 6.65
resnet101_int8    26.78 -> 25.01
```

*SKX platform:*
```
resnet50_int8     22.87 -> 22.46
resnet50_fp32     13.65 -> 12.91
resnet101_fp32    7.37 -> 6.99
resnet101_int8    12.20 -> 12.07
```
It's inevitable result of change from moving Tensor's pointer to copying it's data. I just wanted to let you know and ask what do you think about it?

GaoWei8:
@grygielski Since performance drop on all Resnet models, PR #22695 can be revert. After reverting, the reasons for not following Paddle Operators Standards and this issue link can be added in code comments with one new PR.

lidanqing-vv:
#22985  is merged, so I close this PR @grygielski

PR #22695, opened 2020-02-20
Change ShareDataWith() to TensorCopy() in conv_mkldnn

grygielski:
Change ShareDataWith() to TensorCopy() in conv_mkldnn to meet Paddle Operators Standards.

PR #22985, opened 2020-03-12
Revert "Change ShareDataWith() to TensorCopy() in conv_mkldnn (#22695)"

grygielski:
This reverts commit https://github.com/PaddlePaddle/Paddle/pull/22695 based on https://github.com/PaddlePaddle/Paddle/issues/22964#issuecomment-597617886 comment.

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -316,6 +316,9 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
         dst_memory_p = handler.AcquireDstMemoryFromResidualDataMemory(
             user_residual_memory_p, to_void_cast<T>(output_data), pipeline);
       } else {
+        // Changing ShareDataWith to TensorCopy results in performance drop
+        // on ResNet architectures
+        // (https://github.com/PaddlePaddle/Paddle/issues/22964)
         output->ShareDataWith(*residual_param);
         auto output_data = output->mutable_data<T>(ctx.GetPlace());
         dst_memory_p =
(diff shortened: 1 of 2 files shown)


grygielski:
@GaoWei8 Could you advice on that?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -316,9 +316,8 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
         dst_memory_p = handler.AcquireDstMemoryFromResidualDataMemory(
             user_residual_memory_p, to_void_cast<T>(output_data), pipeline);
       } else {
-        auto output_data = output->mutable_data<T>(
-            ctx.GetPlace(), residual_param->memory_size());
-        framework::TensorCopy(*residual_param, residual_param->place(), output);
+        output->ShareDataWith(*residual_param);
```
GaoWei8, line 322: The reasons for using the ShareDataWith function and the issue link (https://github.com/PaddlePaddle/Paddle/issues/22964) can be added here.

grygielski: Ok, I'll do that in a moment

zhhsplendid:
Hi @grygielski , I can approve the PR but I would like to make sure the correctness and performance of this PR. May I ask the reason why we have to write the ShareDataWith here?

1. Do we guarantee the correctness of using ShareDataWith here? Could you explain a bit more or show some test cases. 
2. What's the reason of the performance drop when using TensorCopy?

grygielski:
Hi @zhhsplendid of course, let me explain it to you.

1. MKL-DNN convolution primitive has an attribute that allows to perform convolution together with addition with residual data. It works in the following way: 
Standard approach is that conv primitive acts like this `dst = conv(src)` and then addition with residual tensor is performed `dst += residual`
But MKL-DNN convolution with residual parameter is given by `dst += conv(src)`. In this scenario we have to store residual tensor data in dst pointer (`dst = residual`) before executing convolution primitive.
2. Performance drop here is quite obvious. In ResNets tensors are pretty large and copying them (`TensorCopy()`) instead of just performing pointer reassignment (`ShareDataWith()`) takes time.

If you still have any questions please don't hesitate asking.

zhhsplendid (approved):
LGTM

zhhsplendid:
Hi @grygielski ，thank you for explaining. Approved.

PR #22997, opened 2020-03-13
Delete Ref & VectorRef and add GetDataSafely

chenwhql:
This PR revome two invalid error check interface, and add a valid error check interface:

**not change the code logic, only enhances the error message**

#### Remove`Ref & VectorRef` check (remove safe_ref.h)

`Ref & VectorRef` are used to safely obtain the data pointed to by a pointer. Before `* ptr`, they will check whether the pointer is `nullptr`. but...

- The code used these two interface get data from ptr, almost no error messages
- use `Ref & VectorRef` cannot tell users the actual error pointer
- use `Ref & VectorRef` cannot tell users the actual error file & line number
- More than **100 places** used

example:

```
int* a = nullptr;
paddle::operators::detail::Ref(a);
```

```
----------------------
Error Message Summary:
----------------------
Error:
[Hint: ptr should not be null.] at (/work/paddle/paddle/fluid/operators/detail/safe_ref.h:28)
```

- no error message!
- error ptr value name is `a`, not `ptr`!
- error file & line are `enforce_test.cc:369`, not `safe_ref.h:28`!

#### Add `GET_DATA_SAFELY `

- A new error check interface in `enforce.h` to replace `Ref & VectorRef`, resolve all above problems!
- You only need use `GET_DATA_SAFELY ` macro, and pass in 3 key word, you can get a complete and accurate error message
- The error message using the new interface is like follows:

```
int* a = nullptr;
GET_DATA_SAFELY(a, "Input", "X", "dummy");
```

```
----------------------
Error Message Summary:
----------------------
InvalidArgumentError: Unable to get int data of Input X in operator dummy. Possible reasons are:
  1. The X is not the Input of operator dummy;
  2. The X has no corresponding variable passed in;
  3. The X corresponding variable is not initialized.
  [Hint: pointer a should not be null.] at (/work/paddle/paddle/fluid/platform/enforce_test.cc:383)
```


luotao1 (approved):

sneaxiy (approved):

raindrops2sea (approved):
LGTM

PR #23010, opened 2020-03-13
Fix problem use recompute and dgc same time

wangxicoding:
Fix the problem which when Recompute and DGC use at the same time will be wrong.


python/paddle/fluid/tests/unittests/test_dgc_optimizer.py
```diff
@@ -112,8 +120,9 @@ def check_dgc_momentum_optimizer(self,
                 self.assertAlmostEqual(op.attr('regular_coeff'), coeff)
                 print("dgc regular_coeff=" + str(coeff))
 
-        with open("test_dgc_optimizer_" + name + ".log", "w") as f:
-            program_to_code(program, fout=f)
+        # for local test debug
+        #with open("test_dgc_optimizer_" + name + str(use_recompute) + ".log", "w") as f:
```
gongweibao, line 124: It's useful?

gongweibao (approved):
LGTM

PR #23022, opened 2020-03-16
[Cherry-pick 1.7] Fix problem use recompute and dgc same time

wangxicoding:
Cherry-pick from #23010
Fix the problem which when Recompute and DGC use at the same time will be wrong.


gongweibao (approved):
LGTM

PR #23024, opened 2020-03-16
fix add grad bug test=develop (#22924)

yaoxuefeng6:
* fix add grad bug test=develop

* update style test=develop


xjqbest (approved):
LGTM

PR #23035, opened 2020-03-16
BugFix for parsing Arguments and inserting funcs in IfElseTransormer

Aurelius84:
+ Fixed the bug that the generated `false_fn` and `true_fn`  functions could not be inserted successfully.

+ Fixed the bug of parsing arguments in `nested_if`.

**Example Case:**
```python
def nested_if_else_2(x):
    y = fluid.layers.reshape(x, [-1, 1])
    b = 2
    if b < 1:
        # var `z` is not visible for outer scope
        z = y

    if fluid.layers.shape(x)[0] < 1:
        if fluid.layers.shape(y)[0] > 1:
            res = fluid.layers.fill_constant(
                value=2, shape=x.shape, dtype="int32")
            # `z` is a new var here.
            z = y + 1
        else:
            res = fluid.layers.fill_constant(
                value=3, shape=x.shape, dtype="int32")
    else:
        res = x
    return res
```

```python
def nested_if_else_3(x):
    y = fluid.layers.reshape(x, [-1, 1])
    b = 2
    # var `z` is visible for func.body
    if b < 1:
        z = y
    else:
        z = x

    if b < 1:
        res = x
        # var `out` is only visible for current `if`
        if b > 1:
            out = x + 1
        else:
            out = x - 1
    else:
        if fluid.layers.shape(y)[0] > 1:
            res = fluid.layers.fill_constant(
                value=2, shape=x.shape, dtype="int32")
            # `z` is created in above code block.
            z = y + 1
        else:
            res = fluid.layers.fill_constant(
                value=3, shape=x.shape, dtype="int32")
            # `out` is a new var.
            out = x + 1
    return 
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -191,8 +191,6 @@ class IsControlFlowVisitor(gast.NodeVisitor):
 
     def visit_Compare(self, node):
         # Ignores child node with `if x` or `if x is None`
-        # TODO(Aurelius84): `if tensor` will be supported in dygraph
-        # and should be considered as is_control_flow.
         pre_control_flow_num = self.is_control_flow_num
         if not compare_with_none(node):
             self.generic_visit(node)
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -372,11 +372,13 @@ class IfConditionVisitor(object):
 
 
 class NameVisitor(gast.NodeVisitor):
-    def __init__(self, not_name_set=None, node_black_set=None):
-        self.not_name_set = not_name_set or set()
+    def __init__(self, node_black_set=None):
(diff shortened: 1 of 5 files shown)


zhwesky2010 (dismissed):

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -366,51 +371,125 @@ def transform(self):
...
 
 
-def get_name_ids(nodes, not_name_set=None, node_black_list=None):
+class NameVisitor(gast.NodeVisitor):
+    def __init__(self, not_name_set=None, node_black_set=None):
+        self.not_name_set = not_name_set or set()
+        self.node_black_set = node_black_set or set()
+        self.name_ids = defaultdict(list)
+        self.name_nodes = []
+        # Available only when node_black_set is set.
```
Aurelius84, line 382: `self.name_nodes` is not used. I will  fix it.

liym27 (dismissed):

liym27:
LGTM

zhwesky2010 (approved):

liym27 (approved):

zhhsplendid (approved):
LGTM

PR #23036, opened 2020-03-16
Fix trt int8 calib

cryoco:
fixed bug:sending num_ele instead of data_size in trt int8 calib causes precision drop.


NHZlX (approved):
LGTM

PR #23037, opened 2020-03-16
initialize global nccl context in dygraph

gavin1332:
As dygraph use nccl comm and stream in `CUDADeviceContext` from `DeviceContextPool`, we bind nccl comm initialized by `NCCLCommContext` of ring id 0 to the DeviceContext mentioned before. So that dygraph and static program execution could have the same entry for nccl comm.

sample code: https://github.com/gavin1332/samples/tree/master/collective

TODO:
- Initialize static program execution while the nccl of PE is initialized
- Set nccl comm in CUDADeviceContext while the nccl ring 0 of static program execution is initialized


LielinJiang (approved):

phlrain (approved):
LGTM

PR #23042, opened 2020-03-16
Add default pass attributes

wozna:
This PR adds an option that allows adding a default value for pass attributes. Solves a python problem where an empty set is always a set of strings.

The default value is set when we register pass. Then PassRegistrary takes the ownership of the default value that will be used in the pass. Thanks to that, we don't have to set this attribute at all when we want to use the default setting. If we set a new value for this attribute it will override the default one. 

An example where we add the default value is `cpu_quantize_palcement_pass`:
```
REGISTER_PASS(cpu_quantize_placement_pass,
              paddle::framework::ir::CPUQuantizePlacementPass)
     // a vector of operator ids that are to be excluded from quantization
    // the second param is the default value for this vector
    .DefaultPassAttr("quantize_excluded_op_ids",
                      new std::unordered_set<int>());
```

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
@@ -111,6 +111,30 @@ void MainTest(std::initializer_list<std::string> quantize_enabled_op_types,
   EXPECT_EQ(use_quantizer_true_count, expected_use_quantizer_true_count);
 }
 
+void DefaultAttrTest(unsigned expected_use_quantizer_true_count) {
+  auto prog = BuildProgramDesc();
+
+  std::unique_ptr<ir::Graph> graph(new ir::Graph(prog));
+
+  auto pass = PassRegistry::Instance().Get("cpu_quantize_placement_pass");
+
+  graph.reset(pass->Apply(graph.release()));
+
+  unsigned use_quantizer_true_count = 0;
+
+  for (auto* node : graph->Nodes()) {
+    if (node->IsOp()) {
+      auto* op = node->Op();
+      if (op->HasAttr("use_quantizer") &&
+          boost::get<bool>(op->GetAttr("use_quantizer"))) {
+        ++use_quantizer_true_count;
+      }
+    }
+  }
+
+  EXPECT_EQ(use_quantizer_true_count, expected_use_quantizer_true_count);
+}
+
 TEST(QuantizerPlacementPass, enabled_pool) { MainTest({"pool2d"}, {}, 2); }
 
 TEST(QuantizerPlacementPass, enabled_conv_excluded_one) {
@@ -120,6 +144,11 @@ TEST(QuantizerPlacementPass, enabled_conv_excluded_one) {
 TEST(QuantizerPlacementPass, excluded_none) {
   // 2 conv + 2 pool
   MainTest({}, {}, 4);
+}
+
+TEST(QuantizerPlacementPass, default_attr_value) {
+  // 2 conv + 2 pool
+  DefaultAttrTest(4);
 }
 
 }  // namespace ir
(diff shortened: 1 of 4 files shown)


paddle/fluid/framework/ir/pass.h
```diff
@@ -100,6 +100,10 @@ class Pass {
   // Set a pointer to the attribute. Pass takes ownership of the attribute.
   template <typename AttrType>
   void Set(const std::string &attr_name, AttrType *attr) {
+    if (default_pass_attrs_.count(attr_name) > 0 &&
+        attrs_.count(attr_name) > 0) {
+      attrs_.erase(attr_name);
```
wojtuss: I would suggest simply overwriting the attribute value (without erasing it first) and changing the `PADDLE_ENFORCE` below into a log message about updating the attribute value (and putting the log first).

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
```diff
@@ -111,6 +111,30 @@ void MainTest(std::initializer_list<std::string> quantize_enabled_op_types,
   EXPECT_EQ(use_quantizer_true_count, expected_use_quantizer_true_count);
 }
 
+void DefaultAttrTest(unsigned expected_use_quantizer_true_count) {
+  auto prog = BuildProgramDesc();
+
```
wojtuss: Please, remove empty lines.

paddle/fluid/framework/ir/pass.h
```diff
@@ -100,8 +100,13 @@ class Pass {
...
   template <typename AttrType>
   void Set(const std::string &attr_name, AttrType *attr) {
-    PADDLE_ENFORCE(attrs_.count(attr_name) == 0, "%s already set in the pass",
-                   attr_name);
+    if (default_pass_attrs_.count(attr_name) == 0) {
+      PADDLE_ENFORCE_EQ(attrs_.count(attr_name), 0,
+                        platform::errors::InvalidArgument(
+                            "Attribute %s already set in the pass", attr_name));
+    } else {
+      VLOG(3) << "Default pass attr " << attr_name << "is updated";
```
wojtuss: I'd suggest a message: `"Setting the attribute " << attr_name << " for the pass " << type_`.

paddle/fluid/framework/ir/pass.h
```diff
@@ -135,16 +140,26 @@ class Pass {
...
+  void CopyDefaultAttrs(std::map<std::string, boost::any> default_attrs_) {
+    attrs_.insert(default_attrs_.begin(), default_attrs_.end());
+  }
+
   void RegisterType(const std::string &type) { type_ = type; }
 
   mutable bool applied_{false};
   std::string type_;
   std::unordered_set<std::string> required_pass_attrs_;
+  std::unordered_set<std::string> default_pass_attrs_;
```
wojtuss, line 163: IMO the name `default_attrs_` would be sufficient as it is already in the `Pass` class.

wozna: Are you thinking about changing only default_pass_attrs to default_attrs_? Because just below you have required_graph_attrs and I wanted to clarify that the default attrs relates to pass attr, not graph attr.

wozna:
@luotao1  Could you please send me the logs from PR_CI_Coverage?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
wozna: Done.

paddle/fluid/framework/ir/pass.h
wozna: Done

paddle/fluid/framework/ir/pass.h
wozna: Done

GaoWei8:
I have rerun PR_CI_Coverage CI and it still failed.
Please check the code coverage report and repair it.
[Paddle _ PR_CI_Coverage _ #53980 (25 Mar 20 01_34) _ Diff Coverage — TeamCity.pdf](https://github.com/PaddlePaddle/Paddle/files/4378912/Paddle._.PR_CI_Coverage._.53980.25.Mar.20.01_34._.Diff.Coverage.TeamCity.pdf)

wozna:
@GaoWei8 could you repeat the coverage test, I can see that it failed on random test.

GaoWei8:
> @GaoWei8 could you repeat the coverage test, I can see that it failed on random test.

Done. I will keep watching it until it passed.

PR #23043, opened 2020-03-16
fix datsset test=develop

yaoxuefeng6:
fix slots shuffle bug which can results in no instance be trained after slots shuffle


xjqbest (approved):
remember to cherry-pick to 1.7，or this pr will be lost.

PR #23057, opened 2020-03-17
Support to use external function

Aurelius84:
In this PR, we support to use external functions in decorated dygraph function. 

Because of import mechanism in python, if you  use `from A import a_func` in `file_A`, `a_func` is visible everywhere in `file_A`. But the decorated function will be applied AST transformation and then will be saved into `tmpfile`. This will lead to the `a_func` is not visible in `tmpfile` which doesn't include statement of `from A import a_func`.

Here, We insert the import statements defined in source file into the tmpfile. `sys.modules` is used to cache all modules and packages that avoids to import same modules twice.

**Example:**
```python
# in file ifelse_simple_func.py

def add_fn(x):
    x = x + 1
    return x

def loss_fn(x, lable):
    loss = fluid.layers.cross_entropy(x, lable)
    return loss


# test_basic.py
def call_external_func(x, label=None):
    if fluid.layers.mean(x).numpy()[0] > 5:
        x_v = x - 1
    else:
        x_v = add_fn(x)  # now add_fn is visible here

    if label is not None:
        loss = loss_fn(x_v, label)  # loss_fn is also visible here
        return loss
    return x_v
```

#### TODO
In future PR, we should consider only use decorate `dygraph_to_static_graph` in entry function instead of decorating all the sub funtion.

#### Reference
+ [The import system of Python](https://docs.python.org/3/reference/import.html)


python/paddle/fluid/dygraph/dygraph_to_static/ast_transformer.py
```diff
@@ -240,5 +240,8 @@ def convert_to_static(dyfunc):
 
     # Get static_func from AST
     func_name = dygraph_to_static.get_module_name()
-    static_func, file_name = ast_to_func(root_wrapper.node, func_name)
+    # Transform into callable function
+    source_file = inspect.getfile(dyfunc)
+    static_func, file_name = ast_to_func(root_wrapper.node, func_name,
+                                         source_file)
```
liym27: Can the code `source_file = inspect.getfile(dyfunc)` be moved to the function `ast_to_func`, because it is one  step of ast_to_func and  `source_file` is only used in `ast_to_func`.
```
static_func, file_name = ast_to_func(root_wrapper.node, func_name,dyfunc)
```

```
def ast_to_func(...):
   source_file = inspect.getfile(dyfunc)
```

Aurelius84: Thanks. I refined the params of `ast_to_func` and only `dyfunc` was required because `func_name` could be got by `dyfunc.__name__`

liym27: 👍
`dygraph_to_static.get_module_name()` won't be used, we can remove it next PR.

liym27 (dismissed):

liym27:
LGTM

liym27 (approved):

zhhsplendid (approved):
LGTM

PR #23067, opened 2020-03-17
Add Support for Break and Continue in Dygraph to Static

zhhsplendid:
1. Add support for Break and Continue in Dygraph to Static
2. Also add support for gast.Not in NodeTestTransformer
3. Also add support for logical op transformation in LoopTransformer


python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
```diff
@@ -62,6 +64,54 @@ def create_while_node(condition_name, body_name, loop_var_names):
...
+
+    def _create_bool_op_node(self, nodes, api_type):
+        assert len(
+            nodes
+        ) > 1, "The length of BoolOp should be at least 2, but received {}.".format(
+            len(nodes))
+        if len(nodes) > 2:
+            # Creates logic_and/logic_or node recursively.
+            pre_assign_node = _create_node(nodes[:2], api_type)
+            nodes = [pre_assign_node] + nodes[2:]
```
Aurelius84, line 107: It seems that`_create_node` is not defined or imported here.

In `NodeTestTransformer`, I added  `_create_node` function to be called recursively because if `BoolOp` was `A and B and C`，the `node.values` will be `values=[A, B, C]` with `length=3`. However, `logical_and` only receives two params (`x` and `y`). So I transformed them recursively.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_break_continue.py
```diff
@@ -0,0 +1,163 @@
...
+
+def test_continue_in_while(x):
+    x = fluid.dygraph.to_variable(x)
+    i = fluid.layers.fill_constant(shape=[1], dtype='int32', value=0)
+    while i < 10:
+        i += 1
+        if i > 5:
+            continue
+            x += 10086
+        x += i
```
liym27, line 54: What if `x += 10086` is before `continue` ? Maybe a test that modifying `x` before `continue`/`break` can be added.

Aurelius84 (commented):
I saw you had implemented `LogicalOpTransformer ` to apply transformation with `while.test`.  It seems that `NodeTestTransformer` is not fully functional. I will keep improving it based on your feedback😄

python/paddle/fluid/dygraph/dygraph_to_static/variable_trans_func.py
```diff
@@ -34,6 +34,28 @@ def create_static_variable_gast_node(name):
...
+            func_code += "dtype='int32', value={})".format(value)
+            return gast.parse(func_code)
+        if isinstance(value, long):
+            func_code += "dtype='int64', value={})".format(value)
+            return gast.parse(func_code)
+    else:
+        if isinstance(value, int):
+            func_code += "dtype='int64', value={})".format(value)
+            return gast.parse(func_code)
+
```
liym27, line 57: Is the expected return value of `create_fill_constant_node ` is `gast.Assigin` ?  
The type of `gast.parse(func_code)` is `gast.Module` and the type of `gast.parse(func_code).body[0]` is `gast.Assigin`

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
Aurelius84: Here, if you just want to transform `A and B`, you can assert `len(nodes) == 2`, and remove other unused code.

python/paddle/fluid/dygraph/dygraph_to_static/break_continue_transformer.py
```diff
@@ -0,0 +1,353 @@
...
+                op=gast.And(), values=[loop_node.test, cond_var_node])
+        elif isinstance(loop_node, gast.For):
+            parent_node = self.ancestor_nodes[loop_node_index - 1]
+            for_to_while = ForToWhileTransformer(parent_node, loop_node,
+                                                 cond_var_node)
+            for_to_while.transform()
+
+    def visit_Continue(self, node):
+        loop_node_index = self._find_ancestor_loop_index(node)
+        assert loop_node_index != -1, "SyntaxError: 'break' outside loop"
```
liym27: The message may be "SyntaxError: 'continue' outside loop"

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
zhhsplendid: Yes, some codes were copied from your code and I didn't find the bug here completely before. Now I found it when adding test cases and I have done the change. PTAL (please take another look)

python/paddle/fluid/dygraph/dygraph_to_static/variable_trans_func.py
zhhsplendid: Thank you for clarifying. Although these code passes tests and still workable as gast.Module, I think changing them as you suggested is better

python/paddle/fluid/dygraph/dygraph_to_static/break_continue_transformer.py
zhhsplendid: Done. These lines were copied so I included the copied error :-)

python/paddle/fluid/tests/unittests/dygraph_to_static/test_break_continue.py
zhhsplendid: Sure, I added a test case before continue/break. Please see test_break_continue_in_for

liym27:
LGTM

liym27 (dismissed):

Aurelius84 (dismissed):
LGTM

liym27 (approved):

PR #23097, opened 2020-03-19
add need_save_delta config to solve OOM

hutuxian:
Solve OOM problem.
Previously, BoxPS don't know whether they should release the memory so the feasign will be accumulated, so we add this parameter to notify BoxPS.


gavin1332 (approved):
LGTM

PR #23104, opened 2020-03-19
Update QAT INT8 related code

lidanqing-intel:
With this patch we do the following:
* separate QAT1.0 and QAT2.0 code - the QAT documentation update will come in a separate PR,
* make QAT2.0 use plain FP32 model to compare its accuracy with an INT8 model derived from the QAT model,
* keep the `--quantized_ops` option in the QAT2.0 tests - a detailed explanation on how to use it will come in the QAT documentation update.

This is Part 1 of the code changes.
Part 2 comes with the PR https://github.com/PaddlePaddle/Paddle/pull/23284.

### DIFF
--- a/python/paddle/fluid/contrib/slim/tests/qat2_int8_image_classification_comparison.py
+++ b/python/paddle/fluid/contrib/slim/tests/qat2_int8_image_classification_comparison.py
@@ -24,7 +24,7 @@ import time
 import paddle
 import paddle.fluid as fluid
 from paddle.fluid.framework import IrGraph
-from paddle.fluid.contrib.slim.quantization import QatInt8MkldnnPass
+from paddle.fluid.contrib.slim.quantization import Qat2Int8MkldnnPass
 from paddle.fluid import core
 
 logging.basicConfig(format='%(asctime)s-%(levelname)s: %(message)s')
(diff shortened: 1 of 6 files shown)


GaoWei8:
The failed log is:

wojtuss:
@luotao1 , this PR requires Dianhai's approval for +1000 lines change. The number is mainly from splitting and renaming files.

luotao1:
> this PR requires Dianhai's approval for +1000 lines change

How about separate this PR?

wojtuss:
@luotao1 OK, I will split it into 2 parts

python/paddle/fluid/contrib/slim/tests/qat2_int8_image_classification_comparison.py
```diff
@@ -0,0 +1,348 @@
+#   copyright (c) 2019 paddlepaddle authors. all rights reserved.
```
sfraczek, line 1: 2020

luotao1: @sfraczek Since windows ci is not stable now, how about merge this PR at first, and update the above typo in the next PR?

PR #23142, opened 2020-03-22
Update rnn.py

Cli212:
在rnn.py添加引用core，否则使用lstm模块时报错 NameError: name 'core' is not defined


luotao1:
Thanks for your contribution, we solve it in #23289 already.

PR #23151, opened 2020-03-23
reorganize the paddle api test=develop

XiaoguangHu01:
add template to make sure that the new api will be in the right file.


raindrops2sea (approved):
LGTM

PR #23152, opened 2020-03-23
Add cuda resource pool for BufferedReader

sneaxiy:
Add CudaResourcePool for BufferedReader to avoid to call `cudaStreamCreate` every time an epoch starts.


chenwhql (approved):
LGTM

PR #23153, opened 2020-03-23
fix bug of ListTransformer in dygraph_to_static

liym27:
Before this PR,  list `a` won't be transformed to LoDTensorArray because `a` is reassigned with a Tensor `fluid.layers.concat(a, axis=0)`. However, `a=[]` should be transformed into `a=fluid.layers.create_array(dtype="float32")`.
```Python
def test_list_in_for_loop_with_concat(x, iter_num):
    a = []  # Before this PR, `a=[]` is not transformed
    for i in range(iter_num):
        a.append(x)
    a = fluid.layers.concat(a, axis=0)
    return a
```


zhhsplendid (approved):
LGTM

PR #23155, opened 2020-03-23
Fix dygraph double grad unittests

sneaxiy:
The `relu` op would output 0, which causes `np.allclose` fails randomly (two numbers which are all nearly 0s may fail in `np.allclose`). This PR fixes the double grad unittest by replacing `relu` with `leaky_relu`.


chenwhql (approved):
LGTM

PR #23169, opened 2020-03-23
Model converter to dot file

wozna:
This is a script that converts models to dot file (optional pdf file). 
Needed to analyze the graph structure of various models.

Example of use:
```
python ../python/paddle/fluid/contrib/slim/tests/convert_model_to_dot.py \
--model_path=${PATH_TO_MODEL}  \
--model_save_path=${PATH_TO_SAVE_DIR}  \
--file_name=${NAME_OF_DOT_PDF_FILES}  \
--pdf=True 
```

### DIFF
--- a/python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
+++ b/python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
@@ -100,6 +100,14 @@ function(save_qat_model_test target qat_model_dir fp32_model_save_path int8_mode
 		 --quantized_ops ${quantized_ops})
 endfunction()
 
+function(convert_model_test target model_path model_save_path file_name pdf)
+    py_test(${target} SRCS ${CMAKE_CURRENT_SOURCE_DIR}/convert_model.py
+            ARGS --model_path ${model_path}
+	         --model_save_path ${model_save_path}
+	         --file_name ${file_name}
+			 --pdf ${pdf})
+endfunction()
+
 if(WIN32)
     list(REMOVE_ITEM TEST_OPS test_light_nas)
     list(REMOVE_ITEM TEST_OPS test_post_training_quantization_mobilenetv1)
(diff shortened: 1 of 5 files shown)


wozna:
This script is needed for Intel DO Team and this is just an example of its location.

luotao1:
https://github.com/rakhimov/cppdep/wiki/How-to-view-or-work-with-Graphviz-Dot-files

We can use Graphviz to convert Dot to PDF locally, Is this PR needed?

wozna:
We wanted to have a tool to convert model to a dot file. We don't really need pdf file that is why I added option in this PR to disable pdf file in `framework.py`.

lidanqing-vv:
@wojtuss  Hi, what do you think removing the changes in framework.py? And also change the name of convert_model.py to test_convert_model2dot.py. Also change copyright from 2019 to 2020
and you can add a link about locally converting pdf file in some comments of the code.

python/paddle/fluid/framework.py
```diff
@@ -3503,7 +3503,12 @@ def build_adjacency_list(self):
             wrapped_adj_list[IrNode(k)] = {IrNode(n) for n in v}
         return wrapped_adj_list
 
-    def draw(self, save_path, name, marked_nodes=None, remove_ctr_var=True):
+    def draw(self,
```
wojtuss: IMO the previous approach could be preserved here. The .pdf files could be generated automatically, without the `--pdf` option.

python/paddle/fluid/contrib/slim/tests/convert_model.py
```diff
@@ -0,0 +1,85 @@
...
+        '--pdf',
+        type=bool,
+        default=False,
+        help='Convert dot file to pdf. Default False.')
+
+    test_args, args = parser.parse_known_args(namespace=unittest)
+    return test_args, sys.argv[:1] + args
+
+
+def convert_model(original_path, save_path, file_name, to_pdf):
```
wojtuss: The name of the method, the script file, and the test seems a bit misleading to me, as it suggests transformation from one form of a model to another one. What about calling it `generate_dot_for_model` or `generate_graph_for_model`? Regardless of whether a .pdf file is generated or not.

python/paddle/fluid/contrib/slim/tests/CMakeLists.txt
```diff
@@ -100,6 +100,14 @@ function(save_qat_model_test target qat_model_dir fp32_model_save_path int8_mode
 		 --quantized_ops ${quantized_ops})
 endfunction()
 
+function(convert_model_test target model_path model_save_path file_name pdf)
+    py_test(${target} SRCS ${CMAKE_CURRENT_SOURCE_DIR}/convert_model.py
+            ARGS --model_path ${model_path}
+	         --model_save_path ${model_save_path}
+	         --file_name ${file_name}
```
wojtuss: IMO the `model_save_path` and `file_name` do not convey the purpose of the options. Maybe `save_graph_dir` and `save_graph_name`?

lidanqing-vv:
ARGS --model_path ${model_path}
	         --save_graph_dir ${save_graph_dir}
	         --save_graph_name ${save_graph_name})

PR #23175, opened 2020-03-24
Add Tensor::IsSharedBufferWith method

sneaxiy:
Code polish for better distinguishing whether two tensors share same memory.


luotao1 (approved):
LGTM

PR #23179, opened 2020-03-24
Support int64 for op assign_value

liym27:
Support that value type of op `assign_value`  can be `int64`.

### DIFF
--- a/paddle/fluid/operators/assign_value_op.cc
+++ b/paddle/fluid/operators/assign_value_op.cc
@@ -52,10 +52,13 @@ class AssignValueOpMaker : public framework::OpProtoAndCheckerMaker {
                               "Shape of values.");
     AddAttr<int>("dtype", "data type of values")
         .InEnum({framework::proto::VarType::INT32,
-                 framework::proto::VarType::FP32});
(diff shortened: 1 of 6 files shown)


zhhsplendid (approved):
Should it also support float64? But considering it is not blocking our dy2stat, I will approve this PR. You can change in this PR or TODO it in the future.

Aurelius84 (approved):
LGTM. But this PR lacks `double` kernel registration, please fill in the reason [here](http://agroup.baidu.com/paddlepaddle/view/office/2348282)

liym27:
> Should it also support float64? But considering it is not blocking our dy2stat, I will approve this PR. You can change in this PR or TODO it in the future.

Yes, float64/bool should be supported too. There are still some bugs that need to be fixed to support float64/bool, so this PR supports `int64` first to avoid blocking dy2stat. `float64` will be supported next PR.

liym27:
> LGTM. But this PR lacks `double` kernel registration, please fill in the reason [here](http://agroup.baidu.com/paddlepaddle/view/office/2348282)

Thanks. Done

wawltor (approved):
LGTM

Issue #22827, opened 2020-03-03
[dev] mkl多线程下存在内存泄露问题


jiweibo:
mkl多线程下存在内存泄露问题。

 测试模型（只包括一个mul）
![image](https://user-images.githubusercontent.com/26377421/76203929-688a4380-6232-11ea-814e-0ed35c9e2691.png)

多线程下重复运行1小时，内存曲线变化
![image](https://user-images.githubusercontent.com/26377421/76203970-7f309a80-6232-11ea-88b2-968221102fc1.png)

尝试将mklml换成 2019.3版本，依旧存在泄露情况，切换成openblas则无问题。

需要intel跟进。

复现文件整理：

复现方法：
unzip mul.zip

cd mul
编译，参考https://www.paddlepaddle.org.cn/documentation/docs/zh/advanced_guide/inference_deployment/inference/native_infer.html#a-name-c-c-a
sh run_impl.sh fluid_inference_install_dir mul_demo
nohup ./build/mul_demo --model_dir=mul_model --thread_num=3 --num=-1 &
监视内存
nohup sh mem_use.sh &
利用top实时查看
top

运行1小时后，kill掉进程
查看memlog.txt 内存变化

cpu型号：2620、2650

[mul.zip](https://github.com/PaddlePaddle/Paddle/files/4306911/mul.zip)

GaoWei8:
@lidanqing-intel Could you help see it? We verified this problem happens in E5-2620 and E5-2650 on develop branch.

lidanqing-vv:
> @lidanqing-intel Could you help see it? We verified this problem happens in E5-2620 and E5-2650 on develop branch.

We are looking at it.

wojtuss:
@jiweibo , @GaoWei8 
I have followed the instructions but could not reproduce the issue on CLX 6248.
The `mul_demo` app did not use MKL-DNN and the memory usage did not change during more than 1 hour long test. If you tested it with MKL-DNN, than it would suggest MKL-DNN primitives caching to be the culprit.

GaoWei8:
> @jiweibo , @GaoWei8
> I have followed the instructions but could not reproduce the issue on CLX 6248.
> The `mul_demo` app did not use MKL-DNN and the memory usage did not change during more than 1 hour long test. If you tested it with MKL-DNN, than it would suggest MKL-DNN primitives caching to be the culprit.

[...]

PR #23187, opened 2020-03-24
[WIP] Solve mklml memory leakage [Issue #22827]

GaoWei8:
Solve memory leakage by removing one intel libiomp5 library and keeping gnu libgomp5. #22827


GaoWei8:
Memory leakage is still existing after changing cmake file. @yinghu5

PR #23191, opened 2020-03-24
Fix graph attr copy issues

sneaxiy:
`fuse_all_optimizer` pass would generate some attributes inside Graph. They should be copied into the separated multiple single device graphs. This PR fixes it.


chenwhql (approved):
LGTM

PR #23120, opened 2020-03-20
Improve SGD jit code to work with large data

grygielski:
This PR resolves https://github.com/PaddlePaddle/Paddle/issues/22757

PR #23195, opened 2020-03-24
Add UT for SGD operator with large inputs

grygielski:
Added UT mentioned here: https://github.com/PaddlePaddle/Paddle/pull/23120


luotao1 (approved):
LGTM

PR #23205, opened 2020-03-25
Remove crf auto growth ut

sneaxiy:
AutoGrowth has been already enabled by default since 1.7. `test_parallel_executor_crf_auto_growth` is the same as `test_parallel_executor_crf`, but it fails randomly in Py35 CI due to out of gpu memory. This PR removes one of the uts.


zhiqiu (approved):
LGTM

PR #23209, opened 2020-03-25
Add elementwise test for PR23001, test=develop

zhaoyuchen2018:
As elementwise has cpu precision issue, add a single pr for debug.

CPU precision has problems as following:
image


luotao1:
It seems a similar problem with #22787, @lidanqing-intel @yinghu5 Could you help see it?

PR #23212, opened 2020-03-25
refine the error message

wangchaochaohu:
对部分报错信息进行优化 优化准则参考https://github.com/PaddlePaddle/Paddle/wiki/PADDLE_ENFORCE-Rewriting-Specification
https://github.com/PaddlePaddle/Paddle/wiki/Paddle-Error-Message-Writing-Specification

### DIFF
--- a/paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
+++ b/paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
@@ -85,7 +85,7 @@ class WriteToArrayInferShape : public framework::InferShapeBase {
     PADDLE_ENFORCE_EQ(
         context->HasInput("I"), true,
         platform::errors::InvalidArgument(
-            "Write to array operation must set the subscript index."));
+            "Read/Write array operation must set the subscript index."));
 
     // TODO(wangchaochaohu) control flow Op do not support runtime infer shape
     // Later we add [ontext->GetInputDim("I")) == 1] check when it's supported
--- a/paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
+++ b/paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
@@ -82,16 +82,26 @@ $$A[i] = T$$
 class WriteToArrayInferShape : public framework::InferShapeBase {
  public:
   void operator()(framework::InferShapeContext *context) const override {
-    PADDLE_ENFORCE(context->HasInput("I"), "Must set the subscript index");
+    PADDLE_ENFORCE_EQ(
+        context->HasInput("I"), true,
+        platform::errors::InvalidArgument(
+            "Write to array operation must set the subscript index."));
     if (context->IsRuntime()) {
-      PADDLE_ENFORCE_EQ(framework::product(context->GetInputDim("I")), 1,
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
```diff
@@ -82,16 +82,23 @@ equation is
...
+
     if (!context->HasInput("X")) {
       return;
     }
 
-    PADDLE_ENFORCE(context->HasOutput("Out"), NotHasOutError());
+    PADDLE_ENFORCE_EQ(
+        context->HasOutput("Out"), true,
+        platform::errors::InvalidArgument(
+            "Do array operation (write/read)must set the output Tensor "
```
zhhsplendid: Tiny English issue: 

1. What's "Do array operation"? Did you try to say "Write to array operation"? 
2. ")must" -> ") must".    That is adding a space.

paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
```diff
@@ -140,10 +140,16 @@ class ReadFromArrayOp : public ArrayOp {
   void RunImpl(const framework::Scope &scope,
                const platform::Place &place) const override {
     auto *x = scope.FindVar(Input("X"));
-    PADDLE_ENFORCE(x != nullptr, "X must be set");
+    PADDLE_ENFORCE_NOT_NULL(
+        x,
+        platform::errors::InvalidArgument(
+            "X(Input Variable) must be set when we do read array operation"));
```
zhhsplendid: I understand the meaning of "do read array operation" but a better expression can be "when we call/use read operation"

paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
```diff
@@ -140,10 +140,16 @@ class ReadFromArrayOp : public ArrayOp {
...
     auto &x_array = x->Get<framework::LoDTensorArray>();
     auto *out = scope.FindVar(Output("Out"));
-    PADDLE_ENFORCE(out != nullptr, "Out must be set");
+    PADDLE_ENFORCE_NOT_NULL(
+        out,
+        platform::errors::InvalidArgument(
+            "Out(Output Varibale) must be set when we do read read operation"));
```
zhhsplendid: I understand the meaning of "do read array operation" but a better expression can be "when we call read operation" or "when we use read operation"

wangchaochaohu: Done Thanks

paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
wangchaochaohu: Done Thanks

paddle/fluid/operators/controlflow/tensor_array_read_write_op.cc
wangchaochaohu: Because read and write infer shape is same, so we comment it like this: Read/Write

zhhsplendid (approved):
LGTM

PR #23228, opened 2020-03-25
Add new tensor in API2.0: max,min,t,eye,log1p

danleifeng:
Add new tensor in API2.0: max,min,t,eye,log1p
max: call `reduce_max` op
min : call `reduce_min` op
eye: call `eye` op, different from `layers.eye` (no batch_shape)
t:  call `transpose2` op, only support <2D tensor.
log1p: new log1p op.


guru4elephant (dismissed):
LGTM

XiaoguangHu01 (approved):
LGTM

Boyan-Liu (approved):

PR #23251, opened 2020-03-26
fix is_controw_flow bug with `if Tensor.numpy()`

Aurelius84:
+ Fix bug with `if tensor.numpy() > 0` that cannot be identified as `isControlFlowIF`
+ Add transformation with `ast.IfExpr`, see example codes as followed.

**Example code:**
```python
# It will be converted into `layers.cond` as followed.
# map_func(lambda x: fluid.layers.cond(i==0, lambda: x, lambda: add_fn(x), y)

y = map_func(lambda x: x if i == 0 else add_fn(x), y)
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -148,7 +148,7 @@ def is_candidate_node(node):
     is_compare_node = isinstance(node,
                                  (gast.Compare, gast.BoolOp, gast.UnaryOp))
     # TODO(Aurelius84): `.numpy()` may be an customized function,
-    # and should consider a more elegant way to solve this problem。
+    # and should consider a more elegant way to solve this problem.
     has_numpy_attr = ".numpy()" in ast_to_source_code(node)
     return is_compare_node or has_numpy_attr
 
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -62,8 +62,6 @@ class IfElseTransformer(gast.NodeTransformer):
         self.after_visit(self.root)
 
     def visit_If(self, node):
-        if 'numpy' not in ast_to_source_code(node):
-            return node
         if_condition_visitor = IfConditionVisitor(node.test,
                                                   self.static_analysis_visitor)
         need_transform = if_condition_visitor.is_control_flow()
@@ -97,15 +95,6 @@ class IfElseTransformer(gast.NodeTransformer):
             return new_node
         else:
             return node
-
-    # def visit_Call(self, node):
-    #     # self.generic_visit(node)
-    #     # Remove `numpy()` statement, like `Tensor.numpy()[i]` -> `Tensor[i]`
-    #     if isinstance(node.func, gast.Attribute):
-    #         attribute = node.func
-    #         if attribute.attr == 'numpy':
-    #             node = attribute.value
-    #     return node
 
     def after_visit(self, node):
         """
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -62,7 +62,8 @@ class IfElseTransformer(gast.NodeTransformer):
         self.after_visit(self.root)
 
     def visit_If(self, node):
-        assert isinstance(node, gast.If)
+        if 'numpy' not in ast_to_source_code(node):
+            return node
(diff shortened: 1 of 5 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -62,7 +62,6 @@ def transform(self):
         self.after_visit(self.root)
 
     def visit_If(self, node):
-        assert isinstance(node, gast.If)
```
zhhsplendid, line 65: Why did you delete this assertion? Is gast.IfExp also goes through visit_If?

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -130,7 +145,10 @@ def is_candidate_node(node):
     """
     Nodes with specified type will be dependent on tensor.
     """
-    return isinstance(node, (gast.Compare, gast.BoolOp, gast.UnaryOp))
+    is_compare_node = isinstance(node,
+                                 (gast.Compare, gast.BoolOp, gast.UnaryOp))
+    has_numpy_attr = ".numpy()" in ast_to_source_code(node)
```
zhhsplendid, line 152: This can be a TODO thing, nothing is required now. 

We used some strings in ProgramTranslator to classify usage, but users may create a class with same string name. We had better to find a better way in the future.

python/paddle/fluid/tests/unittests/dygraph_to_static/ifelse_simple_func.py
```diff
@@ -28,6 +28,28 @@ def loss_fn(x, lable):
...
+    def body(i, ten, y):
+        # It will be converted into `layers.cond` as followed.
+        # map_func(lambda x: fluid.layers.cond(i==0, lambda: x, lambda: add_fn(x), y)
+        y = map_func(lambda x: x if i == 0 else add_fn(x), y)
+        i += 1
+        return [i, ten, y]
+
+    i = fluid.layers.fill_constant(shape=[1], dtype='int64', value=0)
+    ten = fluid.layers.fill_constant(shape=[1], dtype='int64', value=10)
+    i, ten, y = fluid.layers.while_loop(cond, body, [i, ten, y])
```
zhhsplendid: It is dyfunc but with static `while_loop`, is that due to your test only run `IfElseTransformer` so you won't like to use python `while`? If so I suggest you to move this function to the unit test where you use the function and comment about it.

In my understanding, you organized `ifelse_simple_func.py` as dyfunc which can be transformed, not "partial transformed" function like this func

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -656,46 +685,43 @@ def transform_if_else(node, root):
...
+                     pred,
+                     true_func,
+                     false_func,
+                     is_if_expr=False):
     """
     Create `fluid.layers.cond(pred, true_fn, false_fn)` to replace
     original `python if/else` statement.
     """
+
+    def create_lambda_node(func_node, is_if_expr=False):
```
zhhsplendid: NIT: func_node is not a function name when `is_if_expr=True`. So I suggest to rename `func_node` to `func_or_expr_node`

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
Aurelius84: gast.IfExp will  go through visit_IfExp. I delete it because once enter the `visit_If` function, the node is gast.If exactly. If node is not `gast.If`, this function will not be called.

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
Aurelius84: You are right, this is hard code and we should deal with this case more elegent.

python/paddle/fluid/tests/unittests/dygraph_to_static/ifelse_simple_func.py
Aurelius84: Thanks, fixed.

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
Aurelius84: Thanks, fixed.

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
zhhsplendid: Sounds good to me

zhhsplendid (approved):
LGTM

liym27 (approved):
LGTM

PR #23269, opened 2020-03-27
Distinguish public/private global vars

sneaxiy:
This PR adds `is_public` field to each global var. It is ready to develop `set_flags` and `get_flags` interfaces in the future PR.


zhhsplendid (approved):
LGTM

PR #23275, opened 2020-03-27
Initialize global nccl_comm in PE

gavin1332:
As a sequential work of #23037, we assign nccl_comm initialized by ParallelExecutor to NCCLCommContext to support Executor running collective operators.

The procedure of nccl context initialization of PE is creating nccl id in startup program, and create nccl comm in its inner code, which does effect on neither dynamic graph (dygraph) nor Executor execution. And the procedure of nccl context initialization of dygraph is directly creating nccl id and nccl comm in its inner code as well. Since we have a global nccl context, `NCCLCommContext`, we could initialize it in both procedures of PE and dygraph.

The implementation detail is that we separate the creation (`CreateNCCLComm` and `CreateAllNCCLComm`) and preparation (`AssignNCCLComm`) of nccl context in the global environment, so we have a entry (`AssignNCCLComm`) for global env initialization. We do creation in PE, dygraph, global env or any other situation, and assign the created nccl context to global env to active collective ops which are depending on the global env.

example: https://github.com/gavin1332/samples/tree/master/collective

### DIFF
--- a/paddle/fluid/platform/collective_helper.h
+++ b/paddle/fluid/platform/collective_helper.h
@@ -71,6 +71,11 @@ class NCCLCommContext {
 
   void CreateAllNCCLComms(const std::vector<int>& dev_ids, int ring_id = 0);
 
+  // a latter comm with the same dev_id and the same ring_id
+  // will override the former
+  NCCLComm* AssignNCCLComm(ncclComm_t comm, int nranks, int rank, int dev_id,
+                           int ring_id = 0);
+
   // retrieve a communicator by the ring id in multiprocessing mode
   NCCLComm* Get(int ring_id) const {
     PADDLE_ENFORCE_GT(comm_map_.count(ring_id), 0,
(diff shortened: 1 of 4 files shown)


paddle/fluid/platform/collective_helper.h
```diff
@@ -71,6 +71,11 @@ class NCCLCommContext {
 
   void CreateAllNCCLComms(const std::vector<int>& dev_ids, int ring_id = 0);
 
+  // a latter comm with the same dev_id and the same ring_id
+  // will override the former
+  NCCLComm* AssignNCCLComm(ncclComm_t comm, int nranks, int rank, int dev_id,
```
guru4elephant, line 76: could you document this interface?

guru4elephant (commented):
please add some document for the interface.

gongweibao:
https://github.com/gavin1332/samples/tree/master/collective
这个里边的例子我觉得可以拿过来当单测和样例。

gongweibao (approved):
LGTM

gavin1332:
I will append the document and move the example to unittest in the following PR

PR #23287, opened 2020-03-28
fix loop_var bug in loop_transformer

Aurelius84:
+ Fixed the bug that var.name of argument in defined function  was incorrectly included in the `loop_vars`
```python
def while_loop_dyfun_with_conflict_var(x):
    i = fluid.dygraph.to_variable(x)

    def relu(y):
        # 'y' is not visible outside the scope.
        return fluid.layers.relu(y)

    while x < 10:
        # If a tmp variable is created which has same name
        # with a argument in function, it should not be
        # included in the loop_vars.
        
        # Before this PR, y will appear in loop_vars, because `self.current_seen_vars` 
        # add function arguments incorrectly.
        add_fn = lambda x, y: x + y
        i = add_fn(i, x)
        x = x + 1
    return i
```

+ Fixed the bug that `self.XX` was incorrectly included in the `loop_vars`

```python
def class_function(self, *args):
    # Class variables are not allowed to appear in the arguments list
    # of defined function under class methods in Python.
    def while_loop_body(self.x, y):  # `self.x` is illegal here in python.
        ......
        return something

    return something
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
@@ -125,6 +125,11 @@ class NameVisitor(gast.NodeVisitor):
         # List of gast.While/gast.For nodes
         self.current_loop = []
 
+        # List of nodes that have scope of variables.
+        self.nodes_with_scope = []
+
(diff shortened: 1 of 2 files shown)


zhhsplendid (approved):
Great and LGTM

PR #23299, opened 2020-03-30
Remove Read-Only Basic Type Variable in loop_vars

zhhsplendid:
Remove Read-Only Basic Type Variable in loop_vars

I didn't add test because these is an existing test case:

```
def while_loop_dyfunc_with_none(x):
    i = fluid.dygraph.to_variable(x)\
        if x is not None \
        else fluid.dygraph.to_variable(x+1)
    flag = 1
    while x < 10:
        i = i + x if flag is not None else x + i
        x = x + 1
    return i
```

As you can see, `flag` is removed from loop_vars thus not converted into boolean tensor and the numerical test is still correct.

BTW, notice this PR doesn't add attribute into condition / write consideration because we found we are not handling class well now. I will put it as TODO thing.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
@@ -254,9 +254,6 @@ class NameVisitor(gast.NodeVisitor):
             return
         self.current_seen_vars.add(node)
 
-        write_context = {
-            type(gast.Store()), type(gast.AugStore()), type(gast.Del())
-        }
         for loop_node in self.current_loop:
             self.in_loop_vars[loop_node].add(node)
 
@@ -267,6 +264,7 @@ class NameVisitor(gast.NodeVisitor):
         self.current_loop.append(node)
         self.in_condition = True
         self.visit(node.target)
+        self.visit(node.iter)
         self.in_condition = False
         self.before_loop_body_vars[node] = copy.copy(self.current_seen_vars)
         self.generic_visit(node)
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
```diff
@@ -217,21 +253,30 @@ def while_loop_body(self.x, y) # `self.x` is illegal.
         if attr_full_name.startswith("self."):
             return
         self.current_seen_vars.add(node)
+
+        write_context = {
+            type(gast.Store()), type(gast.AugStore()), type(gast.Del())
+        }
```
liym27: `write_context` doesn't seem to be used in `visit_FunctionDef `.

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
```diff
@@ -217,21 +253,30 @@ def while_loop_body(self.x, y) # `self.x` is illegal.
...
         for loop_node in self.current_loop:
             self.in_loop_vars[loop_node].add(node)
+
         # sub-nodes are visited during get_attribute_full_name and we shouldn't
         # visit again
 
     def visit_For(self, node):
         self.current_loop.append(node)
+        self.in_condition = True
         self.visit(node.target)
```
liym27, line 266: Doesn't `node.iter` also need to visit ? 

Because in the following code, `iter_num` may be not appended in `loop_vars`, but it should be.
```Python
def fn(x):
  iter_num = 3
  for i in range(iter_num):
     x = x+1
  return x
```

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
zhhsplendid: Thank you for reviewing carefully. I will delete it.

python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
zhhsplendid: You are right, I will add it.

Aurelius84 (approved):
LGTM

liym27 (approved):
LGTM

PR #23301, opened 2020-03-30
Add reader dependency pass

sneaxiy:
Add a reader dependency pass to make sure all ops must run after `read` op runs. This is designed to avoid the following cases:

```
input_data = read_op()
x = any_inplace_op_that_does_not_rely_on_read_op(x)
```

If `read` op does not run before `any_inplace_op_that_does_not_rely_on_read_op`, `x` would be changed even though there is no data in the current batch. It is extremely serious in learning rate decay cases.


chenwhql (approved):
LGTM

PR #23302, opened 2020-03-30
Support Parameter type determination in StaticAnalysis

Aurelius84:
+ Support parameter type determination in StaticAnalysis

In this PR, the `NodeVarType` of parameter will also be analyzed.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
@@ -333,8 +333,8 @@ class StaticAnalysisVisitor(object):
                             var_type = self._get_constant_node_type(
                                 defaults_node)
 
-                # Add node with identified type into cur_env.
-                self.var_env.set_var_type(node.id, var_type)
+                            # Add node with identified type into cur_env.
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/static_analysis.py
```diff
@@ -260,20 +260,25 @@ def get_node_to_wrapper_map(self):
...
+        if node.value is None:
+            return {NodeVarType.NONE}
+        if isinstance(node.value, bool):
+            return {NodeVarType.BOOLEAN}
+        if isinstance(node.value, int):
+            return {NodeVarType.INT}
+        if isinstance(node.value, float):
+            return {NodeVarType.FLOAT}
+        if isinstance(node.value, str):
+            return {NodeVarType.STRING}
```
zhhsplendid, line 276: Could you return {NodeVarType.UNKNOWN} at last of _get_constant_node_type in case that these is a case we didn't consider? Other things LGTM

Aurelius84: You're right. I already add commit to fix it.

PR #23304, opened 2020-03-30
Fix test_launch random fail.

gongweibao:
1. Some time python interprets `python -m` to `python -c`.
2. And some time python interprets `python -m` to the path of the module.
And the the user may get `Unknown command line flag `-c`(or -m)` error.
This is temperily fix of them.


gavin1332 (approved):
LGTM

Issue #22904, opened 2020-03-06
[DNNL] Add inplace support to DNNL ops


jczaja:
DNNL primitives works faster if computation is in-place e.g. Input and output are the same memory
so input is overwritten by outcome of computation. We checked (using external program) that
DNNL's layer norm takes 30% less time to execute when in-place. We hope that other operators
will gain as well.

Following DNNL operators are supporting in place:
- Layer Norm
- Batch Norm
- softmax
- binary

Goal of this issue is to:
Add IR pass to scan a graph and  and make In = Out tensors, if input is not used by more than one operator (We cannot overwrite input of layer norm if this input is to be used in other operator as well)

jczaja:
Here are  early results(latency in ms, lower is better) of experiments with softmax inplace mkl-dnn vs develop):
| Branch        | FP32, Threads: 1         | FP32, Threads: 20| Int8, Threads: 1 | Int8, Threads: 20|
| ------------- |:-------------:| -----:|--------:|-----:|
| develop      |  171.736 | 29.8331 | 64.6135  | 17.1636 |
| softmax dnnl inplace     | 170.047 | 29.6391 | 64.1744 | 16.7482 |

Platform: Intel Xeon 6248 , Turbo ON
Instructions how to was run: https://github.com/PaddlePaddle/benchmark/tree/master/Inference/c%2B%2B/ernie/mkldnn

jczaja:
done

PR #23315, opened 2020-03-30
[DNNL] Added MKL-DNN inplace pass for C-API inference

jczaja:
This PR is introducing C-API MKL-DNN pass to take adventage of in-place ops of MKL-DNN for C-API inference. Currently only softmax in-place is available. 

There should be little performance improvement visible on BERT fp32 and ERNIE int8(
https://github.com/PaddlePaddle/Paddle/issues/22904#issuecomment-600227776)

Apart from performance improvement, memory consumption should also be smaller, but due to fact that PaddlePaddle is preallocating memory blocks , I did not observed lower memory usage in tested model (BERT).

After this PR, elementwise_add and activation in-place support will follow


grygielski (approved):
LGTM

jczaja:
@luotao1 , @sneaxiy Could you please review this PR?

luotao1 (approved):
LGTM

PR #23322, opened 2020-03-30
Dev/add l1 loss

zhiqiu:
Add a new public nn API, `paddle.nn.loss.L1Loss(reduction='mean')`.

1. As the form of class.
2. Under module `paddle`directly, instead of `fluid`.
3. The functionality of `L1Loss` is to calculate the L1 loss.

usage:
```
import paddle.fluid as fluid
import paddle
import numpy as np
l1_loss = paddle.nn.loss.L1Loss()
input = fluid.data(name="input", shape=[1])
label = fluid.data(name="label", shape=[1])
output = l1_loss(input, label)
place = fluid.CPUPlace()
exe = fluid.Executor(place)

input_data = np.array([1.5]).astype("float32")
label_data = np.array([1.7]).astype("float32")
output_data = exe.run(fluid.default_main_program(),
        feed={"input":input_data, "label":label_data},
        fetch_list=[output],
        return_numpy=True)

print(output_data)  # [array([0.2], dtype=float32)]
```


python/paddle/nn/layer/loss.py
```diff
@@ -13,9 +13,93 @@
...
+        super(L1Loss, self).__init__()
+        self.reduction = reduction
+
+    def forward(self, input, label):
+        fluid.data_feeder.check_variable_and_dtype(
+            input, 'input', ['float32', 'float64', 'int32', 'int64'], 'l1_loss')
+        fluid.data_feeder.check_variable_and_dtype(
+            label, 'label', ['float32', 'float64', 'int32', 'int64'], 'l1_loss')
+
+        if self.reduction not in ['sum', 'mean', 'none']:
```
sneaxiy: Move this check in constructor?

python/paddle/fluid/tests/unittests/test_l1_loss.py
```diff
@@ -0,0 +1,60 @@
...
+                fetch_list=[ret])
+
+        with fluid.dygraph.guard():
+            l1_loss = paddle.nn.loss.L1Loss()
+            dy_ret = l1_loss(
+                fluid.dygraph.to_variable(input_np),
+                fluid.dygraph.to_variable(label_np))
+            dy_result = dy_ret.numpy()
+
+        expected = np.mean(np.abs(input_np - label_np))
```
sneaxiy, line 53: Add more checks? `mean`, `sum` and `none` ?

python/paddle/nn/layer/loss.py
```diff
@@ -13,9 +13,93 @@
...
+    'L1Loss',
+    #    'NLLLoss',
+    #   'BCELoss'
+]
+
+
+class L1Loss(fluid.dygraph.Layer):
+    """
+    This interface is used to construct a callable object of the ``L1Loss`` class.
+    The L1Loss layer calculates the mean absolute error of input predictions and target label.
```
sneaxiy: Not only `mean absolute error`? `reduction` may be `sum`, `mean` or `none`.

python/paddle/nn/layer/loss.py
```diff
@@ -13,9 +13,93 @@
...
+    .. math::
+        Out = |input - label|
+    If :attr:`reduction` set to ``'mean'``, the reduced mean loss is:
+    .. math::
+        Out = MEAN(|input - label|)
+    If :attr:`reduction` set to ``'sum'``, the reduced sum loss is:
+    .. math::
+        Out = SUM(|input - label|)
+    Parameters:
+        reduction (str, optional): Indicate how to average the loss by batch_size,
```
sneaxiy: By `batch_size` ? Please make sure that whether this API returns tensor with shape [1] or shape [1, other_dim].

python/paddle/nn/layer/loss.py
zhiqiu: Done, thanks.

python/paddle/fluid/tests/unittests/test_l1_loss.py
zhiqiu: Done, thanks.

python/paddle/nn/layer/loss.py
zhiqiu: Done, thanks.

python/paddle/nn/layer/loss.py
zhiqiu: Thanks, it should return loss with shape [1]  when reduction is 'sum' or 'mean', and return loss with the same shape of input when reduction is 'none'.

sneaxiy (approved):
LGTM.

PR #23334, opened 2020-03-31
Add sigmoid op in paddle.nn and paddle.nn.functional

hong19860320:
新增如下API:
paddle.nn.Sigmoid
paddle.nn.functional.sigmoid


Xreki (approved):
LGTM

PR #23335, opened 2020-03-31
Add allclose_op

wzzju:
Add `allclose` Op, and its function is analogous to `numpy.allclose`.  It returns True if two tensors are element-wise equal within a tolerance.

`def allclose(input, other, rtol=1e-05, atol=1e-08, equal_nan=False, name=None)`
`numpy.allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)`

API 文档预览：
image
image
image


sneaxiy (dismissed):
LGTM.

sneaxiy (dismissed):

sneaxiy (dismissed):

sneaxiy (approved):

PR #23338, opened 2020-03-31
Change allreduce 2 c allreduce

phlrain:
allreduce接口 替换为 _allreduce 接口，减少底层依赖，修复double free的错误


python/paddle/fluid/dygraph/parallel.py
```diff
@@ -448,8 +448,8 @@ def apply_collective_grads(self):
         coalesced_grads_and_vars = self._coalesce_tensors(grad_var_groups)
 
         for coalesced_grad, g_vars, g_shapes in coalesced_grads_and_vars:
-            collective._allreduce(
-                coalesced_grad, coalesced_grad, sync_mode=False)
```
gavin1332, line 452: broadcast op needs to be replaced as well if dygraph uses

PR #23344, opened 2020-03-31
fix nccl comm double free bug

gavin1332:
As nccl comm is not created by CUDADeviceContext, it should be destroyed by the creator as the best practice of RAII.


gongweibao (approved):
LGTM

PR #23345, opened 2020-03-31
Add the sum of API2.0, support input, output dtype set, test=develop

wawltor:
Add new sum op for API 2.0 in this pr, and support  the in dtype and out dtype for sum op


wawltor:
image

wawltor:
图片

PR #23351, opened 2020-03-31
[Dy2stat] Add data_layer_not_check

zhhsplendid:
1. Add data_layer_not_check because it is needed in dygraph_to_static where input can be variable size
2. Remove warnings in static analysis because python cannot do exact static analysis


liym27 (approved):
LGTM

Aurelius84 (approved):
LGTM

PR #23364, opened 2020-04-01
Add full_like op. 

cryoco:
Add full like op: full_like(input, fill_value=0.0, out=None)
This function creates a tensor filled with `fill_value` which has identical shape and dtype with `input`.

```
import paddle
import paddle.fluid as fluid
import numpy as np

input = fluid.data(name='input', dtype='float32', shape=[2, 3])
output = paddle.tensor.full_like(input, 2.0)
exe = fluid.Executor(fluid.CPUPlace())
img=np.array([[1, 2, 3], [4, 5, 6]]).astype(np.float32)
res = exe.run(fluid.default_main_program(), feed={'input':img}, fetch_list=[output])
print(res) # [array([[2., 2., 2.], [2., 2., 2.]], dtype=float32)]
```


NHZlX (dismissed):
LGTM

NHZlX (approved):
lgtm

PR #23380, opened 2020-04-01
add paralell_executor dependancy to collective_helper

gavin1332:
add paralell_executor dependancy to collective_helper to fix the build problem


wawltor (approved):
LGTM

PR #23381, opened 2020-04-01
Fix elementwise compile error, test=develop

zhaoyuchen2018:
elementwise function used before definition then failed in cuda 8, move it ahead.


wangchaochaohu (approved):
LGTM

PR #23387, opened 2020-04-01
move sin, sqrt, tanh, atan to paddle.tensor.math and add a new parameter "out"

kinghuin:
sin, sqrt, tanh, atan add parameter "out" and moved to paddle.tensor.math

When both name and out have been specify, only out takes effect, warning:

image

image

image

image

image


Boyan-Liu (dismissed):

Boyan-Liu (approved):

wawltor (approved):
LGTM

PR #23396, opened 2020-04-01
[Paddle-TRT]: add eltwise,pool2d, prelu, scale, concat, gelu dynamic shape support  

NHZlX:
Different with trt static shape，The batch info needs to be in the shape description in trt dynamic shape mode. Therefore, it is necessary to modify some converters of TRT to adapt to dynamic shape which contains:

elementwise_add, elementwise_mul, pool2d, prelu, scale, concat, shufflechannel.


luotao1 (approved):
LGTM for remove const_cast

Shixiaowei02 (approved):

PR #23398, opened 2020-04-01
fix dump slot in strategy

xjqbest:
* fix dump slot when accesor is DownpourCtrDoubleAccessor, it may cause nan during training,
* test=develop


yaoxuefeng6 (approved):
LGTM

PR #23400, opened 2020-04-02
Fix bug of inference partial feed when FLAGS_enable_parallel_graph=1

sneaxiy:
When `FLAGS_enable_parallel_graph=1`, ParallelExecutor should support feed partial data when running inference network. This PR adds the support.


chenwhql (approved):
LGTM

PR #23404, opened 2020-04-02
[Not Merge]Supprot LoDTensorArray in fetch

gfwm2013:
使fetch能够支持LoDTensorArray

### DIFF
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -276,7 +276,6 @@ static bool has_fetch_operators(
     PADDLE_ENFORCE_EQ(
         fetch_count, fetch_targets.size(),
         "The number of fetch operators should match 'fetch_targets'");
-
     if (!fetch_holder_name.empty()) {
       // When fetch operator are present, so should be fetch_holder.
       auto var = block.FindVar(fetch_holder_name);
@@ -284,6 +283,44 @@ static bool has_fetch_operators(
                               fetch_holder_name);
       PADDLE_ENFORCE_EQ(var->GetType(), proto::VarType::FETCH_LIST,
                         "'%s' variable should be 'FETCH_LIST' type",
+                        fetch_holder_name);
+    }
+  }
+
+  return fetch_count > 0;
+}
+
+static bool has_fetch_operators(
+    const BlockDesc& block,
+    const std::map<std::string, FetchVar*>& fetch_targets,
+    const std::string& fetch_holder_name) {
+  size_t fetch_count = 0;
+  for (auto* op : block.AllOps()) {
+    if (op->Type() == kFetchOpType) {
+      fetch_count++;
+      // The output variable's name of fetch_op should be fetch_holder_name.
+      PADDLE_ENFORCE_EQ(op->Output("Out")[0], fetch_holder_name,
+                        "Output of fetch op should be '%s'", fetch_holder_name);
+      std::string fetch_target_name = op->Input("X")[0];
+      PADDLE_ENFORCE(
+          fetch_targets.find(fetch_target_name) != fetch_targets.end(),
+          "Fetch operator input name '%s' cannot be found in 'fetch_targets'",
+          fetch_target_name);
+    }
+  }
+
+  if (fetch_count > 0) {
+    PADDLE_ENFORCE_EQ(
+        fetch_count, fetch_targets.size(),
+        "The number of fetch operators should match 'fetch_targets'");
+
+    if (!fetch_holder_name.empty()) {
+      // When fetch operator are present, so should be fetch_holder.
+      auto var = block.FindVar(fetch_holder_name);
+      PADDLE_ENFORCE_NOT_NULL(var, "Block should already have a '%s' variable",
+                              fetch_holder_name);
+      PADDLE_ENFORCE_EQ(var->GetType(), proto::VarType::FETCH_VAR_LIST,
+                        "'%s' variable should be 'FETCH_VAR_LIST' type",
                         fetch_holder_name);
     }
   }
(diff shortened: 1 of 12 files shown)


paddle/fluid/framework/executor.cc
```diff
@@ -291,6 +290,44 @@ static bool has_fetch_operators(
   return fetch_count > 0;
 }
 
+static bool has_fetch_operators(
```
zhhsplendid, line 293: Is the `has_fetch_operators` above still being used? If not, you can delete above function

paddle/fluid/framework/executor.cc
```diff
@@ -364,6 +401,79 @@ void Executor::Run(const ProgramDesc& program, Scope* scope,
                      fetch_holder_name);
 }
 
+void Executor::Run(const ProgramDesc& program, Scope* scope,
```
zhhsplendid, line 404: Did you just copy above `run` code? If yes, you should make old `Run` interface to call the new one, you remove duplicated code in this way.

paddle/fluid/framework/executor.cc
```diff
@@ -509,6 +619,43 @@ void Executor::RunPreparedContext(
   }
 }
 
+void Executor::RunPreparedContext(
```
zhhsplendid, line 622: Did you just copy above `RunPreparedContext` code? If yes, you should make old `RunPreparedContext` interface to call the new one, you remove duplicated code in this way.

paddle/fluid/framework/feed_fetch_type.h
```diff
@@ -24,6 +25,38 @@ using FeedFetchList = std::vector<FeedFetchType>;
 using FetchUnmergedList = std::vector<std::vector<FeedFetchType>>;
```
zhhsplendid, line 25: Here I saw FeedFetchType = LoDTensor, can we using FeedFetchType = boost::variant<LoDTensor, LoDTensorArray>, is the change big?

paddle/fluid/framework/feed_fetch_type.h
```diff
@@ -24,6 +25,38 @@ using FeedFetchList = std::vector<FeedFetchType>;
...
+  FetchVar() = default;
+  FetchVar(const LoDTensor &lod_tensor) : FetchVarBase(lod_tensor) {}  // NOLINT
+  FetchVar(const LoDTensorArray &lod_tensor_array)                     // NOLINT
+      : FetchVarBase(lod_tensor_array) {}                              // NOLINT
+};
+
+// using FetchVar = boost::variant<LoDTensor, LoDTensorArray>;
+using FetchVarList = std::vector<FetchVar>;
+
+struct DataIsLoDTensor : public boost::static_visitor<bool> {
```
zhhsplendid, line 42: Even you use a functor, you should declare it as a `class`

paddle/fluid/framework/framework.proto
```diff
@@ -126,6 +126,7 @@ message VarType {
     LOD_TENSOR_ARRAY = 13;
     PLACE_LIST = 14;
     READER = 15;
+    FETCH_VAR_LIST = 16;
```
zhhsplendid, line 129: I feel FETCH_VAR_LIST is very similar to FETCH_LIST by meaning. Could you find whether we can reuse the FETCH_LIST type?

python/paddle/fluid/executor.py
```diff
@@ -901,6 +911,11 @@ def _run_impl(self, program, feed, fetch_list, feed_var_name,
                 use_program_cache=use_program_cache)
 
         program._compile(scope, self.place)
+
+        if _has_lod_tensor_array(program._program, fetch_list):
+            raise ValueError("Currently, The type of item in fetch_list should be" \
```
zhhsplendid, line 916: “Currently, The type of item in fetch_list should be LoDTensor the program is compiled”

Did you miss "when" ?

It is better to be “Currently, the type of item in fetch_list should be LoDTensor when using CompiledProgram”

paddle/fluid/operators/controlflow/fetch_op.cc
```diff
@@ -20,6 +20,32 @@ limitations under the License. */
 
 namespace paddle {
 namespace operators {
+void TransDataLayout(const framework::LoDTensor &src_var,
+                     std::string fetch_var_name,
```
zhhsplendid, line 24: Do not pass value of object in C++, it will be slow. Please use "const std::string&"

paddle/fluid/operators/controlflow/fetch_op.cc
```diff
@@ -36,53 +62,43 @@ class FetchOp : public framework::OperatorBase {
...
+        TransDataLayout(src_item, fetch_var_name, dst_item);
+      } else if (fetch_var->IsType<framework::LoDTensorArray>()) {
+        auto &src_item = fetch_var->Get<framework::LoDTensorArray>();
+        auto &item = fetch_list->at(col);
+        std::vector<framework::LoDTensor> temp;
+        temp.resize(src_item.size());
+        item = temp;
+        framework::LoDTensorArray *dst_item =
+            &(boost::get<framework::LoDTensorArray>(item));
+        for (size_t i = 0; i < src_item.size(); i++) {
```
zhhsplendid, line 97: Not Important (NIT): write ++i in the future. It isn't here but i++ is much slower than ++i at many other cases.

paddle/fluid/operators/controlflow/fetch_op.cc
```diff
@@ -36,53 +62,43 @@ class FetchOp : public framework::OperatorBase {
...
-        TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
+    if (out_var->IsType<framework::FeedFetchList>()) {
+      auto *fetch_list = out_var->GetMutable<framework::FeedFetchList>();
+      if (col >= fetch_list->size()) {
+        fetch_list->resize(col + 1);
       }
-#else
-      TensorCopySync(src_item, platform::CPUPlace(), &dst_item);
-#endif
+      auto *dst_item = &(fetch_list->at(col));
```
zhhsplendid, line 76: Using auto when the type is obvious, otherwise you should specify the type in C++

For example, `auto &src_item = fetch_var->Get<framework::FeedFetchType>()` is very clear that `src_item` is a reference of `framework::FeedFetchType`, so that's good.

But not `auto *dst_item = &(fetch_list->at(col));`

paddle/fluid/operators/controlflow/fetch_op.cc
```diff
@@ -36,53 +62,43 @@ class FetchOp : public framework::OperatorBase {
...
+        fetch_list->resize(col + 1);
+      }
+      if (fetch_var->IsType<framework::LoDTensor>()) {
+        auto &src_item = fetch_var->Get<framework::LoDTensor>();
+        auto *dst_item =
+            &(boost::get<framework::LoDTensor>(fetch_list->at(col)));
+        TransDataLayout(src_item, fetch_var_name, dst_item);
+      } else if (fetch_var->IsType<framework::LoDTensorArray>()) {
+        auto &src_item = fetch_var->Get<framework::LoDTensorArray>();
+        auto &item = fetch_list->at(col);
```
zhhsplendid, line 91: Don't use auto at line 91

paddle/fluid/operators/controlflow/fetch_op.cc
```diff
@@ -36,53 +62,43 @@ class FetchOp : public framework::OperatorBase {
...
+        std::vector<framework::LoDTensor> temp;
+        temp.resize(src_item.size());
+        item = temp;
```
zhhsplendid, line 94: Is that possible you directly make change to `item` as a vector instead of creating a vector and copy it into `item`? Later one will be slower and the `temp` decreases readability.

PR #23405, opened 2020-04-02
improve elementwise performance.

zhaoyuchen2018:
- improve data pattern 10X10X1X128, 1X32X128 performance

|elementwise grad|before opt(ms)| after opt(ms)|pytorch(ms)|
|--|--|--|--|
|x=2048X100X1X32,y=1X128X32| 46 |25 |30|
- support contiguous data pattern like 2048x1x1x128.


wawltor (approved):
LGTM

PR #23408, opened 2020-04-02
Add functional convolutions in paddle.nn.functional

iclementine:
1. add `conv2d/conv2d_transpose/conv3d/conv3d_transpose` in `paddle.nn.functional`;
2. minor fix for ConvTransposeOp (InferShape's behavior in compile time and run time. Now it only infers shape at run time or input's shape is not -1.).
3. update error message for COnvTransposeOp's InferShape.
4. fix the bug that `fluid.layers.conv3d_transopse` does not pass `output_size` to OP. 
5. fix `fluid.layers.conv3d_transpose`'s `output_size`'s size, (3 instead of 2).
6. support tuple of int as `output_size` (in addition to int and list[int]).
7. unify the way to handle all kinds of padding( in all the 4 apis.)
8. use a different order of keyword arguments in conv2d/conv3d, `padding` now precedes `stride`. This is the same as `conv2d_transpose` and `conv3d_transpose`, but different than that in `fluid.layers.conv2d` and `fluid.layers.conv3d`.


kolinwei (approved):

kuke (approved):
LGTM

raindrops2sea (approved):

PR #23413, opened 2020-04-02
[Fix bug] Set fuse_all_reduce_ops=False for quantization test

juncaipeng:
Set fuse_all_reduce_ops=False for quantization test


wzzju (approved):
LGTM.

PR #23442, opened 2020-04-03
add init value to  Member variables in paddle_analysis_config.h

NHZlX:
Detected by valgrind, If not initialized, there will be a memory leak


luotao1 (approved):
LGTM! Thanks very much!

PR #23443, opened 2020-04-03
Fix Tracer::NoGrad bug

sneaxiy:
The concept `Tracer::no_grad_` is wrong. When `Tracer::no_grad_` is true, it means backward is traced now, which is wrong. This PR fixes it.


zhiqiu (approved):
LGTM

PR #23444, opened 2020-04-03
Add the argmax op to API 2.0

wawltor:
Add the argmax op to api 2.0， add the dtype, keepdims, out parameter to the op


FlyingQianMM (approved):
LGTM

PR #23449, opened 2020-04-03
add save with prefix

xjqbest:
* add save with prefix
* test=develop

usage:
```
fleet.save_model("./model0",)
fleet.save_model("./model0", mode=0)

fleet.save_one_table(0, "./model1")
fleet.save_one_table(0, "./model1", mode=0)

# if you set prefix, the saved part has prefix:  "part-hahaha-"
# the save path should contain table id like 000, 001.etc.
fleet.save_one_table(0, "./model1/000", prefix="hahaha")
fleet.save_one_table(1, "./model1/000", mode=0, prefix="hehehe")

fleet.load_model("./model0")
fleet.load_model("./model0", mode=0)

fleet.load_one_table("./model0")
fleet.load_one_table("./model0", mode=0)
```

note:  load model or load one table does not support prefix, which means it can not load models with prefix.

### DIFF
--- a/paddle/fluid/framework/fleet/fleet_wrapper.cc
+++ b/paddle/fluid/framework/fleet/fleet_wrapper.cc
@@ -917,6 +917,37 @@ void FleetWrapper::SaveModel(const std::string& path, const int mode) {
 #endif
 }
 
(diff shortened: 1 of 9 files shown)


123malin (approved):
LGTM

PR #23456, opened 2020-04-03
Post_training_quantizaion sets batch_generator to support lod tensor

juncaipeng:
Post_training_quantizaion sets batch_generator to support lod tensor


wzzju (approved):
LGTM.

PR #23464, opened 2020-04-03
Fix inplace_abn compile error on Windows

heavengate:
**Fix inplace_abn compile error on Windows**

SyncBatchNorm not compiled on Windows currently, which is used by inplace_abn OP, fix as follow:
- compile sync_batch_norm on Windows, for the NCCL functions is limited by `#ifndef WIN32`, compiling is ok
- If on WIN32, do not use `sync_batch_norm` in `sync_batch_norm_pass.cc`

Tested on windows OK


qingqing01 (dismissed):
LGTM.

wangchaochaohu (approved):
LGTM

PR #23465, opened 2020-04-03
Dev/fix init flags

zhiqiu:
This PR fixes a bug that existed for a long time.

The `__init__.py` in `paddle.fluid ` pass `sys.argv[0]` to `core.init_flags`.
Commonly, if we run `python xx.py`, then `sys.argv[0]`  is `xx.py`, and it makes no error, since it does not start with `-`.
But, in the following 2 cases, it may cause `ERROR`

1. run `python -m paddle.fluid`,
2. run `python -c "import paddle.fluid"`,

Examples in paddle-1.6,
image

The reason is that, in these 2 cases, `sys.argv[0]` is `-c` and `-m`.
And, starting with `-` is treated as `key` of args in `gflags`, see more details in [gflags](https://gflags.github.io/gflags/). So it cannot be properly interpreted as `gflags` defined in paddle. 

Plus, I found that `sys.argv[0]` is useless even if it is `xx.py`, since not start with `-` or `--`, so I remove them.

Related #23265

### DIFF
--- a/paddle/fluid/platform/init.cc
+++ b/paddle/fluid/platform/init.cc
@@ -53,7 +53,6 @@ std::once_flag glog_warning_once_flag;
 void InitGflags(std::vector<std::string> argv) {
   std::call_once(gflags_init_flag, [&]() {
     FLAGS_logtostderr = true;
-    argv.insert(argv.begin(), "dummy");
     int argc = argv.size();
     char **arr = new char *[argv.size()];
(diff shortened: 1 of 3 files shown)


sneaxiy (dismissed):
LGTM.

lanxianghit (approved):
LGTM

sneaxiy (approved):

PR #22892, opened 2020-03-06
support if logic for Variable in dygraph

JepsonWong:
support if logic for Variable in dygraph.

```
with fluid.dygraph.guard():
    var1 = fluid.dygraph.to_variable(np.array([[[0]]]))
    var1_bool = False
    if var1:
        var1_bool = True
```

related pr: https://github.com/PaddlePaddle/models/pull/4447

PR #23478, opened 2020-04-04
[Dy2stat] Make loop_transformer supports class variable

zhhsplendid:
This CR makes two changes:

1. In old loop_transformer, if a class variable, such as "self.a, foo.bar" is a loop var, the Dy2stat will fail because `def func(self.foo)` is not legal syntax. We support class variable by renaming.

2. After https://github.com/PaddlePaddle/Paddle/pull/22892 is merged, we can support `while x < 10` in dygraph. I enable those tests in corresponding Dy2stat


Aurelius84 (approved):
LGTM

PR #23500, opened 2020-04-05
Api/Op (shrink_rnn_memory) error message enhancement

gfwm2013:
As the title.

### DIFF
--- a/python/paddle/fluid/layers/control_flow.py
+++ b/python/paddle/fluid/layers/control_flow.py
@@ -1724,9 +1724,9 @@ def shrink_memory(x, i, table):
         usage.
     """
     helper = LayerHelper('shrink_memory', **locals())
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/layers/control_flow.py
```diff
@@ -1724,6 +1724,9 @@ def shrink_memory(x, i, table):
         usage.
     """
     helper = LayerHelper('shrink_memory', **locals())
+    check_type(x, 'x', (Variable), 'shrink_memory')
```
zhhsplendid: Why do you put "()" around Variable? Suggest to use `Variable` or `(Variable, )` instead

gfwm2013: Done.

zhhsplendid (approved):
LGTM

PR #23529, opened 2020-04-07
[dy2static] Revert to usage of 'fill_constant' in test_transformer. 

liym27:
`fill_constant` has support `value` is `Tensor` after #23491, so `fill_constant` can be used in test_transformer.


Aurelius84 (approved):
LGTM

PR #23536, opened 2020-04-07
[Paddle-TRT] Add hard_sigmoid and hard_swish support(support MobilenetV3)

cryoco:
- added hard_sigmoid TRT op converter;
- added hard_swish TRT op converter and plugin.


NHZlX (approved):
LGTM

PR #23573, opened 2020-04-07
Solve the conflict of ops with the same name, test for CI.

wzzju:
The problem is described in ISSUE #23198.
image


sneaxiy (approved):
LGTM.

PR #23576, opened 2020-04-07
fix adaptive_pool2d/pool3d error message

heavengate:
**fix adaptive_pool2d/pool3d error message**
- C++ `PADDLE_ENFORCE_XXX` is ok
- `check_type`... in python API


heavengate:

PR #23580, opened 2020-04-07
fix grid_sampler error message

heavengate:
**fix grid_sampler error message**
- fix C++ `PADDLE_ENFORCE_XXX`
- fix `check_xtype` in python API


heavengate:

PR #23586, opened 2020-04-07
fix temporal_shift error message

heavengate:
**fix temporal_shift error message**
- `PADDLE_ENFORCE_XXX` in C++
- `check_type`... in python API


heavengate:

PR #23597, opened 2020-04-08
Pimpl to polish PartialGradEngine code

sneaxiy:
Use [PImpl](https://en.cppreference.com/w/cpp/language/pimpl) method to polish `PartialGradEngine` code.


chenwhql (approved):
LGTM

PR #23613, opened 2020-04-08
 lod_rank_table error message enhance

yongqiangma:
1.check input x data type, it should be Variable.
2.check input level  data type, it should be int32 or int64.


hong19860320 (approved):
LGTM

PR #23619, opened 2020-04-08
Update print signatures to support paddle APIs

sneaxiy:
Update `print_signatures.py` to support 2.0 APIs in `paddle` module.

image


luotao1 (approved):
LGTM

PR #23627, opened 2020-04-08
Op(leaky_relu) error message enhancement

zhupengyang:
- Op(leaky_relu) error message enhancement
    - check x type and dtype
    - no PADDLE_ENFORCE
    - no need to check shape


hong19860320 (approved):
LGTM

PR #23644, opened 2020-04-08
API (save/load  series) error message enhancement 

chenwhql:
## io.save/load series Python API报错信息增强
- 完善了save load系列python api中的assert和raise报错信息


luotao1 (approved):
LGTM

PR #23652, opened 2020-04-08
Support transform two independent functions in dygraph_to_static_output 

Aurelius84:
+ Fix the bug that `NameVisitor` parsed `True/False` as arguments.
+ Support transform two independent function in dygraph_to_static_output.
+ Polish the comment of function.

Here，if decorate many function by `dygraph_to_static_output`, the variables and ops will be added into the same main_program, which is consistent with static graph.

+ The variables and ops in foo and bar will belong to the same main_program.
```python
@dygraph_to_static_output
def foo(x, y):
      ....

@dygraph_to_static_output
def bar(x):
    ....

# run program
z = foo(x, y)
out = bar(z)
```

If want to split the two functions into different main_program, `with program_guard` is required, which is also consistent with static graph.

+ The variables and ops in foo and bar will belong to the different main_program.
```python
@dygraph_to_static_output
def foo(x, y):
      ....

@dygraph_to_static_output
def bar(x):
    ....

# run program
with program_guard(fluid.Program()):
    out = foo(x, y)

with program_guard(fluid.Program()):
    out = bar(x)
```


liym27 (dismissed):
LGTM

zhhsplendid (approved):
LGTM

PR #23674, opened 2020-04-09
Fix the test case bug for matmul, test=develop

wawltor:
Fix the test case bug，change the matmul dtype from float32 to float64


FlyingQianMM (approved):
LGTM

PR #23678, opened 2020-04-09
enhance attention_lstm and param_attr error message

MyPandaShaoxiang:
fix param_attr python check_type
fix attention_lstm c++ error message


NHZlX (approved):
LGTM

PR #23679, opened 2020-04-09
Fix GET_DATA_SAFELY ptr

sneaxiy:
Replace `ptr` with `__ptr` in `GET_DATA_SAFELY` macro, because `ptr` would be a common name which developers would like to use.


chenwhql (approved):
LGTM

PR #23687, opened 2020-04-09
polish tensor.where codes and english document

GaoWei8:
polish tensor.where codes and en document
192 168 43 53_8088_documentation_docs_en_api_search_where html

### DIFF
--- a/python/paddle/tensor/search.py
+++ b/python/paddle/tensor/search.py
@@ -13,17 +13,8 @@
 # limitations under the License.
 from __future__ import print_function
 import numpy as np
-import warnings
-import six
-import os
-import inspect
 from ..fluid.layer_helper import LayerHelper
 from ..fluid.data_feeder import check_variable_and_dtype, check_type, check_dtype
-from ..fluid.initializer import Normal, Constant, NumpyArrayInitializer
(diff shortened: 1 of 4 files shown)


python/paddle/tensor/search.py
```diff
@@ -227,27 +218,41 @@ def sort(input, axis=-1, descending=False, out=None, name=None):
 def where(Condition, X, Y):
     """
```
Xreki, line 219: Python接口里面，参数都用小写的吧，`where(condition, x, y)`

GaoWei8: Done.

python/paddle/tensor/search.py
```diff
@@ -224,60 +215,78 @@ def sort(input, axis=-1, descending=False, out=None, name=None):
...
-        out : The tensor. 
+        Variable: A Tensor with the same data dype as x. 
+
     Examples:
         .. code-block:: python
 
           import numpy as np
-          import paddle as paddle
           import paddle.fluid as fluid
+          import paddle.tensor as paddle
```
Xreki, line 248: import paddle.tensor as tensor

PR #23688, opened 2020-04-09
Refine mechanism of calling outer function in dy2static

Aurelius84:
In dy2static, the transformed static function is saved in tmp file and is loaded as `module` by `imp` . But, in the tmp file, the static function lost global variables from import statements and defined in source file.

Before this PR, we used `ImportVisitor` to parse the statements related `import` and extracted them, then added them into tmp file where the static function was. However, the functions and variables defined in the source file are still invisible because they didn't belong to the case of `import`.

In this PR, we recover the global variables or functions seen in dyfunc source file by `__globals__` attribute. We only apply transformation on `dyfunc` before, so that the variables seen by dyfunc should be also visible by transformed static function.

### Example code:
+ call function defined in other place
```python
# defined in different file with `net`, should be imported.
def bar(x):
    ......
    return out

# defined in same file
def foo_ahead(x):
    ......
    return out

@dygraph_to_static_func
def net(x):
    y = foo_ahead(x)
    y = foo_behind(y)
    out = bar(y)
    return out

# defined in same file but behind `net`
def foo_behind(x):
    ......
    return out
```

+ call function from outside class
```python
# defined in different file with `net`, and will be imported.
def bar(x):
    ......
    return out

# defined in same file
def foo_ahead(x):
    ......
    return out

Class Net(Layer):
    def __init__(self, ......):
        ......

    @dygraph_to_static_func
    def forward(x):
        y = foo_ahead(x)
        y = foo_behind(y)
        out = bar(y)
       return out

# defined in same file but behind `net`
def foo_behind(x):
    ......
    return out
```

+ nested function

```python
@dygraph_to_static_func
def net_func(x):
    
    def bar(x):
        ......
        return out

    def foo(x):
        y = bar(x)
        return y

    return foo(x)
```

#### TODO
If only decorate one of inner function instead of decorating the main function, the other inner functions are invisible for the decorated function.

**For example:**

```python
def net_func(x):
    
    def bar(x):
        ......
        return out

    @dygraph_to_static_func
    def foo(x):
        # Here function `bar` is not visible for foo.
        # Recommend user to add decorator on `net_func`
        y = bar(x)
        return y
    
    return foo(x)
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/utils.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/utils.py
@@ -356,6 +356,8 @@ class RenameTransformer(gast.NodeTransformer):
 def ast_to_func(ast_root, dyfunc, delete_on_exit=True):
     """
     Transform modified AST of decorated function into python callable object.
+    TODO: If only decorate one of inner function instead of decorating the main
+    function, the other inner functions are invisible for the decorated function.
     """
     source = ast_to_source_code(ast_root)
     if six.PY2:
--- a/python/paddle/fluid/dygraph/dygraph_to_static/utils.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/utils.py
@@ -308,42 +308,6 @@ def create_funcDef_node(nodes, name, input_args, return_name_ids):
     return func_def_node
 
 
-class ImportVisitor(gast.NodeVisitor):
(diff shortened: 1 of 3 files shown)


liym27 (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #23692, opened 2020-04-09
Add user local data preprocess support

lidanqing-intel:
To provide demo to users, we need to explain how to preprocess and convert data to binary. There should be an option to choose users' local data.
Added unit test for it.

@wojtuss Please review, thank you

### DIFF
--- a/paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
+++ b/paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
@@ -24,6 +24,7 @@ import math
 from paddle.dataset.common import download
 import tarfile
 import StringIO
+import argparse
 
 random.seed(0)
 np.random.seed(0)
@@ -131,7 +132,7 @@ def check_integrity(filename, target_hash):
         return False
 
 
-def convert(tar_file, output_file):
+def convert_Imagenet_tar2bin(tar_file, output_file):
     print('Converting 50000 images to binary file ...\n')
     tar = tarfile.open(name=tar_file, mode='r:gz')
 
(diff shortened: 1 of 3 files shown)


paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
```diff
@@ -205,9 +206,86 @@ def run_convert():
...
+                of.write(np_img.astype('float32').tobytes())
+
+                #save label(int64_t) to file
+                label_int = (int)(label)
+                np_label = np.array(label_int)
+                of.seek(SIZE_INT64 + SIZE_FLOAT32 * DATA_DIM * DATA_DIM * 3 *
+                        num_images + idx * SIZE_INT64)
+                of.write(np_label.astype('int64').tobytes())
+
+        target_size = 8 + num_images * 3 * args.data_dim * args.data_dim * 4 + num_images * 8
```
wojtuss: Could you please add an explanation of what particular factors mean?

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
```diff
@@ -205,9 +206,86 @@ def run_convert():
...
+        else:
+            print("Conversion failed!")
+
+
+def main_preprocess_Imagenet(args):
+    parser = argparse.ArgumentParser(
+        description="Convert the full Imagenet val set or local data to binary file.",
+        usage=None,
+        add_help=True)
+    parser.add_argument('--local', action="store_true")
```
wojtuss: Please, add a description of this option.

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
```diff
@@ -205,9 +206,86 @@ def run_convert():
...
+    parser.add_argument(
+        "--data_dir",
+        default="./third_party/inference_demo/int8v2/imagenet_small",
```
wojtuss: I think I would not define the default values for this path and the next filename/path options. These values would be set either in the test definition (in a cmake file) or deliberately by a user. In both cases setting the default values here can be misleading.

paddle/fluid/inference/tests/api/full_pascalvoc_test_preprocess.py
```diff
@@ -268,8 +268,7 @@ def main_pascalvoc_preprocess(args):
         description="Convert the full pascalvoc val set or local data to binary file.",
         usage=None,
         add_help=True)
-    parser.add_argument(
-        '--choice', choices=['local', 'VOC_test_2007'], required=True)
+    parser.add_argument('--local', action="store_true")
```
wojtuss: Same as above.

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
lidanqing-vv: You are right. I will change this.

paddle/fluid/inference/tests/api/full_pascalvoc_test_preprocess.py
lidanqing-vv: Done

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
lidanqing-vv: Done

paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py
lidanqing-vv: Done

PR #23713, opened 2020-04-10
Fix the matmul test case, avoid double and float precision diff

wawltor:
just avoid the test precision for the matmul


FlyingQianMM (approved):
LGTM

PR #23723, opened 2020-04-10
[Eernie TRT]: add slice op and add emb eltwise layernorm fp16 support

NHZlX:
1. add slice op converter
2. add emb eltwise layernorm fp16 support
test=develop


paddle/fluid/inference/tensorrt/convert/slice_op.cc
```diff
@@ -0,0 +1,67 @@
...
+
+namespace paddle {
+namespace inference {
+namespace tensorrt {
+
+class SliceOpConverter : public OpConverter {
+ public:
+  void operator()(const framework::proto::OpDesc& op,
+                  const framework::Scope& scope, bool test_mode) override {
+#if IS_TRT_VERSION_GE(6000)
```
Superjomn, line 28: add some comments here to explain why TRT should >= 6000

or replace this line with something like

```
#define TRT_SUPPORT_FP16  IS_TRT_VERSION_GE(6000)


#if TRT_SUPPORT_FP16
...
#endif  // TRT_SUPPORT_FP16
```

paddle/fluid/inference/tensorrt/plugin/emb_eltwise_layernorm_plugin.cu
```diff
@@ -31,11 +31,34 @@ namespace plugin {
...
+  int nb_emb = embs_.size();
+  std::vector<void *> ptr_vector(nb_emb);
+  std::vector<std::vector<half>> emb_fp16(nb_emb);
+
+  if (sizeof(T) == sizeof(float)) {
+    // FP32
+    for (int i = 0; i < nb_emb; ++i) {
+      ptr_vector[i] = embs_[i];
+    }
+  } else {
```
Superjomn, line 45:
```c++
else if (sizeof(T) == sizeof(fp16)) { }
else { NOT_IMPLEMENTED }
```

paddle/fluid/inference/tensorrt/plugin/emb_eltwise_layernorm_plugin.cu
```diff
@@ -160,18 +193,32 @@ int EmbEltwiseLayernormPluginDynamic::enqueue(
...
-      out_type == nvinfer1::DataType::kFLOAT, true,
-      platform::errors::InvalidArgument(
-          "The EmbEltwiseLayernorm Plugin only only support fp32 input."));
+  if (sizeof(T) == sizeof(float)) {
+    PADDLE_ENFORCE_EQ(
+        out_type == nvinfer1::DataType::kFLOAT, true,
+        platform::errors::InvalidArgument(
+            "The EmbEltwiseLayernorm Plugin only support fp32 input."));
+  } else {
+    PADDLE_ENFORCE_EQ(
```
Superjomn, line 202: Is `PADDLE_ENFORCE` works here?

NHZlX: Using PADDLE_ENFORCE will result in CI failure

paddle/fluid/inference/tensorrt/convert/slice_op.cc
NHZlX: done

paddle/fluid/inference/tensorrt/plugin/emb_eltwise_layernorm_plugin.cu
NHZlX: done

paddle/fluid/inference/tensorrt/plugin/emb_eltwise_layernorm_plugin.cu
NHZlX: done

paddle/fluid/inference/tensorrt/plugin/emb_eltwise_layernorm_plugin.cu
```diff
@@ -31,11 +31,34 @@ namespace plugin {
...
+      auto &tmp = emb_fp16[i];
+      tmp.resize(emb_size);
+
+      for (int j = 0; j < emb_size; ++j) {
+        tmp[j] = static_cast<half>(embs_[i][j]);
+      }
+      ptr_vector[i] = tmp.data();
+    }
+  }
   embs_gpu_.reserve(embs_.size());
```
jeng1220: @NHZlX,
Should replace `embs_gpu_.reserve` with `embs_gpu_.resize`.

It is not safe to access the pointers of element of `embs_gpu_` before push anything into the vector.

This is a simple example. Try it and you will see that `bar` is reserved to 10, but size of `bar` still is 0.
```c++
#include <iostream>
#include <vector>

int main ()
{
  std::vector<int> bar;
  int N = 10;
  bar.reserve(N);

  for (int i=0; i<N; ++i) {
    int* ptr = &bar[i];
    std::cout << "iteration: " << i << std::endl;
    std::cout << "get &bar[i]: " << ptr << std::endl;
    std::cout << "bar.size: " << bar.size() << std::endl;
    std::cout << "bar.capacity: " << bar.capacity() << std::endl;
  }
  return 0;
}
```

Superjomn (approved):
LGTM

PR #23749, opened 2020-04-11
add cuda kernel for seed, test=develop

mapingshuo:
add Cuda kernel for seed, so that no conversion is needed when there is a GPU op behind.


guru4elephant (approved):
LGTM

kolinwei (approved):

PR #23774, opened 2020-04-12
Add the support of bool list for assign_value op

wawltor:
1. add the support of bool_values for assign_value op 
2. add the test case and api test case for this function


liym27 (dismissed):
LGTM

liym27 (approved):

PR #23777, opened 2020-04-12
[dy2static] Revert to usage of 'assign' in test_transformer

liym27:
`assign` has supported numpy.bool  after #23774, so assign(bool) can be used in test_transformer.


Aurelius84 (approved):
LGTM for transformer.

PR #23779, opened 2020-04-12
Fix some sample codes bug in api2.0, format the sample code

wawltor:
Those ops that be fixed in this pr, include ones, ones_like, zeors_like, matmul, pow, mul


Heeenrrry (approved):

Boyan-Liu (approved):

FlyingQianMM (approved):
LGTM

PR #23795, opened 2020-04-13
fix bug in reduce op

qjing666:
- Bug description: When the length of dim equals to the length of input shape, reduce_ops that are executed on cpu will return [0.], no matter what the input is.

- Fix the bug by assigning reduce_all to 'True' when the mentioned condition is met.


guru4elephant (approved):
LGTM

PR #23802, opened 2020-04-13
Correct reader device index

sneaxiy:
`AsyncSSAGraphExecutor` is not compatible with ordered `DataLoader`, because:

- `AsyncSSAGraphExecutor` would call `BuildStrategy::Apply` methods for each sub-graph, so that we cannot get device index using `ComputationOpHandle::GetScopeIdx()`. 
- The ordered non-iterable `DataLoader` would put `create_py_reader` op inside main program, so that the global scope would have the persistable var of blocking queue. Inside `AsyncSSAGraphExecutor`, we have to init variables in local scopes in the reverse order, so that  each `create_py_reader` op of each device would get the right blocking queue. Otherwise, all `create_py_reader` ops in all devicecs would get the same blocking queue, i.e., the 1st one inside `OrderedMultiDeviceLoDTensorBlockingQueue`.


luotao1 (approved):
LGTM for const_cast

seiriosPlus (approved):
LGTM

chenwhql (approved):
LGTM

PR #23805, opened 2020-04-13
Remove conv_bn_fuse_pass and fc_fuse_pass in trt int8 calibration

cryoco:
Remove conv_bn_fuse_pass and fc_fuse_pass in trt int8 calibration mode to avoid possible inconsistence between local and server calibration tables caused by the unstableness in these 2 passes.


NHZlX (approved):
LGTM

PR #23821, opened 2020-04-14
optimize compare and logical ops error info, add test case for this ops

kinghuin:
* optimize compare and logical ops error info
* add out and cond dtype test


wawltor (dismissed):
LGMT

kinghuin:
DO NOT MERGE NOW

PR #23831, opened 2020-04-14
fix compile problem on windows and some invalid argument check.

ZHUI:
fix compile problem on windows and some invalid argument check. test=develop


wawltor (approved):
LGTM

PR #23852, opened 2020-04-14
Fix trt fc fuse test

FrostML:
Fix `test_trt_fc_fuse_pass` and `test_trt_transpose_flatten_concat_fuse_pass` caused by wrong setting of max_batch_size.


NHZlX (approved):
LGTM

PR #23856, opened 2020-04-14
improve efficiency of dot op in dygraph mode

liupluswei:
Type check consumes too much time in dygraph mode: about 8~10 times to other framework.
Change the code to call API directly in dygraph mode.

### DIFF
--- a/python/paddle/tensor/linalg.py
+++ b/python/paddle/tensor/linalg.py
@@ -440,6 +440,10 @@ def dot(x, y, name=None):
 
     """
     op_type = 'dot'
+    if in_dygraph_mode():
+        op = getattr(core.ops, op_type)
+        return op(x, y)
+
     assert x is not None, 'x cannot be None in {}'.format(op_type)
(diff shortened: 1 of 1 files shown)


python/paddle/tensor/linalg.py
```diff
@@ -440,6 +440,10 @@ def dot(x, y, name=None):
 
     """
     op_type = 'dot'
+    if in_dygraph_mode():
```
lanxianghit, line 444: 加个注释说明吧，以便于其他人理解这里的逻辑

lanxianghit: 由于目前ci排队时间比较长，可以先merge，后续补充注释

lanxianghit (dismissed):

python/paddle/tensor/linalg.py
liupluswei: done

phlrain (approved):

PR #23754, opened 2020-04-11
added reshape transpose matmul fuse pass

sfraczek:
Introduce MKLDNN fuse pass (designed especially for Ernie) named `reshape_transpose_matmul_mkldnn_fuse_pass`. It fusses reshape and transpose operations in the inputs of matmul into the matmul operator itself.

To acheive this it adds 4 new attributes to matmul_op.
* fused_transpose_X, 
* fused_transpose_Y,
* fused_reshape_X, 
* fused_reshape_Y.

PR #23866, opened 2020-04-14
added fusing matmul-transpose-reshape pass

arlesniak:
This PR fuses the following sequence of ops for Ernie:
prev_op-(4dim)->mkldnn_matmul-(4dim)->transpose-(4dim)->reshape-(3dim)->next_op
------------------------------------------->  axis={0,2,1,3}---->shape={0,0,[-2]*[-1]}
into:
prev_op-(4dim)->mkldnn_matmul-(3dim)->next_op 

Also the fuse adds two new attributes to matmul_op:  fused_transpose_Out (axis from transpose) and fused_reshape_Out (shape from reshape) to be used by mkldnn MM kernel


lidanqing-vv:
Hi @arlesniak  Please refer to [PADDLE_ENFORCE-Rewriting-Specification](https://github.com/PaddlePaddle/Paddle/wiki/PADDLE_ENFORCE-Rewriting-Specification) and change all the `PADDLE_ENFORCE` in your modified files, otherwise `PR_CI_CPU_Py2 (Compile)` will always fail. 

You can also refer to https://github.com/PaddlePaddle/Paddle/pull/23754#pullrequestreview-393223492

python/paddle/fluid/tests/unittests/ir/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
```
lidanqing-vv: Are these two commented files needed?

python/paddle/fluid/tests/unittests/test_matmul_op_output_fused.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.inputs = {'X': self.x, 'Y': self.y}
+        self.attrs = {'use_mkldnn': self.use_mkldnn, }
+        if len(self.shape_out) > 0:
+            self.attrs['reshape_Out'] = self.shape_out
+            self.attrs['axis_Out'] = self.axis_out
+
+        self.outputs = {'Out': self.out}
+
+    def test_check_output(self):
```
lidanqing-vv: Do you consider using `check_output_with_option` here too, I saw you use it in later functions

python/paddle/fluid/tests/unittests/test_matmul_op_output_fused.py
```diff
@@ -0,0 +1,145 @@
...
+        self.feeds = {
+            "x": np.random.random([bs] + shape_x).astype(d_type),
+            "y": np.random.random([bs] + shape_y).astype(d_type)
+        }
+        self.fetch_list = [out]
+        self.enable_mkldnn = True
+
+    def test_check_output(self):
+        use_gpu = False
+        self.check_output_with_option(use_gpu)
```
lidanqing-vv: I think the op with `reshape_out` and `transpose_out` , should not mix up with op without these two attributes set. But in this PR, I did not see key changes, so it means, when reuse, you may use wrong cached op, because you did not add attribute value to the key.

Could you also add a cache mechanism check for the fused op. Simply speaking, you just need to run 2 iterations. You can refer Michal's matmul PR, I think he did this work.

python/paddle/fluid/tests/unittests/ir/__init__.py
arlesniak: Yes they allow to do the following import:  
> from ir.inference.inference_pass_test import InferencePassTest
in the test_matmul_op_output_fused.py

python/paddle/fluid/tests/unittests/test_matmul_op_output_fused.py
arlesniak: The `TestMatMulOpSpecial` class inherits from `OpTest` class which does not have the `check_output_with_option` method. The latter tests i.e. `TestMKLDNNMatmulFuseOp` inherit from `InferencePassTest` class there the method exists.

python/paddle/fluid/tests/unittests/test_matmul_op_output_fused.py
arlesniak: Concerning mixing op without attributes.
The test you mentioned is doing two runs of the little topology from `setUp`. First is raw inference pass, the second is with MKLDNN on and than they are compared. There is no need to put the attributes `reshape_out` and `transpose_out` to MKLDNN MM here because they are modified implicitly during the fuse pass.
Concerning caching:
Thank you for the comment I'll check this

arlesniak: @lidanqing-intel  OK I changed caching key and PADDLE_ENFORCE could you please see new version.

arlesniak:
PR_CI_CPU_Py2 complains about only @skip_check_grad

> 10:22:32] :	 [Step 1/1] 0. It is an Op accuracy problem, please take care of it. You must have one RD (zhangting2020 (Recommend), luotao1 or phlrain) approval for the usage (either add or delete) of @skip_check_grad_ci. For more information, please refer to: https://github.com/PaddlePaddle/Paddle/wiki/Gradient-Check-Is-Required-for-Op-Test. The corresponding lines are as follows:
[10:22:32] :	 [Step 1/1] 
[10:22:32] :	 [Step 1/1] python/paddle/fluid/tests/unittests/test_matmul_op_output_fused.py
[10:22:32] :	 [Step 1/1] 
[10:22:32] :	 [Step 1/1] +	@skip_check_grad_ci(reason="Tests inference only optimization.")
[10:22:32] :	 [Step 1/1] 
[10:22:32] :	 [Step 1/1] 
[10:22:32] :	 [Step 1/1] There are 1 approved errors.

arlesniak:
Please check new version

lidanqing-vv:
LGTM

python/paddle/fluid/tests/unittests/test_matmul_op_output_fused.py
lidanqing-vv: Thanks for changing the key. LGTM.

paddle/fluid/framework/ir/mkldnn/matmul_transpose_reshape_fuse_pass_tester.cc
```diff
@@ -0,0 +1,93 @@
...
+  graph.reset(pass->Apply(graph.release()));
+
+  int current_nodes_num = graph->Nodes().size();
+  EXPECT_EQ(original_nodes_num - 6, current_nodes_num);
+
+  for (auto *node : graph->Nodes()) {
+    if (node->IsOp()) {
+      auto *op = node->Op();
+      if (op->Type() == "matmul") {
+        ASSERT_TRUE(op->HasAttr("reshape_Out"));
```
wojtuss: Please, verify also the attributes' values here.

paddle/fluid/framework/ir/mkldnn/matmul_transpose_reshape_fuse_pass_tester.cc
```diff
@@ -0,0 +1,93 @@
...
+  }
+  if (type == "reshape2") {
+    op->SetAttr("shape", std::vector<int>({0, 2, 1}));
+    op->SetOutput("XShape", {outputs[1]});
+  }
+
+  if (type == "matmul") {
+    op->SetInput("Y", {inputs[1]});
+    op->SetAttr("use_mkldnn", true);
+    op->SetAttr("reshape_Out", std::vector<int64_t>({1, 2, 3}));
```
wojtuss: Should the `reshape_Out` and `transpose_Out` attributes be set before applying the pass?

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_mkldnn_op.py
```diff
@@ -161,5 +161,59 @@ def set_attributes(self):
...
+    def generate_data(self):
+        bs = 8
+        dt = "float32"
+        self.x = np.random.random([bs, 12, 128, 128]).astype(dt)
+        self.y = np.random.random([bs, 12, 128, 64]).astype(dt)
+        self.transpose_out = [0, 2, 1, 3]
+        self.reshape_out = [0, 0, 768]
+        self.out = np.matmul(self.x, self.y).transpose([0, 2, 1, 3]).reshape(
+            [bs, -1, 768])
+
```
wojtuss, line 291: A test with rank 3 `reshape_Out` is missing.

sfraczek:
I would prefer not using transpose_Out name for attribute because there is transpose_X and transpose_Y already which have different meaning

paddle/fluid/framework/ir/mkldnn/matmul_transpose_reshape_fuse_pass_tester.cc
arlesniak: OK

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_mkldnn_op.py
```diff
@@ -161,5 +161,59 @@ def set_attributes(self):
...
```
wojtuss, line 291: A test with `int8` data type is also needed imo. This could be achieved by defining a class method
```
def init_data_type(self):
   self.data_type_ = 'float32'
```
then calling it in the setUp() method, and using self.data_type_ in the body of the `generate_data()` method instead of using the "float32" string. In the next class only override the init_data_type() method for INT8 test.

wojtuss:
@luotao1, some builds failed due to exhausted space on a disc:
```
fatal error: error in backend: IO failure on output stream: No space left on device
```

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -100,34 +114,45 @@ class MatMulFactory {
     const memory::dim M = mat_dim_x.height_;
     const memory::dim N = mat_dim_y.width_;
     const memory::dim K = mat_dim_x.width_;
-    return {BS, M, N, K};
+
+    batch_size_ = 1;
+    auto b = BS;
+    if (BS > 1 && IsOutputFused(ctx)) {
+      batch_size_ = ctx.Input<Tensor>("X")->dims()[0];
+      b = BS / batch_size_;
```
Sand3r-, line 127: Batch size doesn't necessarily have to be defined in "X" tensor. This is the reason why it was inferred in such a non trivial way few lines above (`const memory::dim BS = x_bs || y_bs ? std::max(x_bs, y_bs) : 1;`)

In my opinion This should be rewritten as follows:
```C++
  if (BS > 1 && IsOutputFused(ctx)) {
      batch_size_ = BS;
      BS = 1;
  }
```
Alternatively
```
  if (BS > 1 && IsOutputFused(ctx)) {
      std::swap(batch_size_, BS);
  }
```
Variable `b` can therefore be omitted.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -100,34 +114,45 @@ class MatMulFactory {
...
+    // Translate transA and transB
+    memory::dims strides_x = !ctx.Attr<bool>("transpose_X")
+                                 ? memory::dims{M * K, K, 1}
+                                 : memory::dims{M * K, 1, M};
+    memory::dims strides_y = !ctx.Attr<bool>("transpose_Y")
+                                 ? memory::dims{N * K, N, 1}
+                                 : memory::dims{N * K, 1, K};
+    memory::dims out_strides = memory::dims{M * N, N, 1};
+
+    correctStridesWhenOutputFused(ctx, N, b, &out_strides);
```
Sand3r-: Please consider renaming `correctStridesWhenOutputFused` to `CorrectStridesWhenOutputFused`

GaoWei8:
> @luotao1, some builds failed due to exhausted space on a disc:
> 
> ```
> fatal error: error in backend: IO failure on output stream: No space left on device
> ```

The CIs have already rerun.   It will take more than 10 hours because of the long CI queue.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -156,11 +181,25 @@ class MatMulFactory {
...
+      out_mem_.set_data_handle(out_ptr);
+      matmul_prim_.execute(stream, {
+                                       {MKLDNN_ARG_SRC, x_mem_},
+                                       {MKLDNN_ARG_WEIGHTS, y_mem_},
+                                       {MKLDNN_ARG_DST, out_mem_},
+                                   });
+      x_ptr = static_cast<char*>(x_ptr) + offsets.x_offset;
+      y_ptr = static_cast<char*>(y_ptr) + offsets.y_offset;
+      out_ptr = static_cast<char*>(out_ptr) + offsets.out_offset;
+    }
```
Sand3r-, line 207: Since this for loop isn't parallelised, wouldn't it be more efficient to calculate each offset on the go?
This way we avoid a cache miss on fetching offsets_ table which could potentially cost us more cycles than just increasing `x_ptr`, `y_ptr ` and `out_ptr ` by the respective products of each memory dimensions (Please see the [documentation](https://oneapi-src.github.io/oneDNN/structdnnl_1_1memory.html#afdd20764d58c0b517d5a31276672aeb8)).
An exemplary implementation could look like this:
```C++
    unsigned bs = batch_size_;
    void* x_ptr = x_mem_.get_data_handle();
    void* y_ptr = y_mem_.get_data_handle();
    void* out_ptr = out_mem_.get_data_handle();
    char* x_size = x_mem_.dims[1] * x_mem_.dims[2];
    char* y_size = y_mem_.dims[1] * y_mem_.dims[2];
    char* out_size = out_mem_.dims[1] * out_mem_.dims[2];
    for (unsigned i = 0; i < bs; i++) {
      x_mem_.set_data_handle(x_ptr);
      y_mem_.set_data_handle(y_ptr);
      out_mem_.set_data_handle(out_ptr);
      matmul_prim_.execute(stream, {
                                       {MKLDNN_ARG_SRC, x_mem_},
                                       {MKLDNN_ARG_WEIGHTS, y_mem_},
                                       {MKLDNN_ARG_DST, out_mem_},
                                   });
      x_ptr = static_cast<char*>(x_ptr) + x_size;
      y_ptr = static_cast<char*>(y_ptr) + y_size;
      out_ptr = static_cast<char*>(out_ptr) + out_size;
    }
```
This would also reduce cache trashing as we reduce the size of the class by removing the offsets_ field.

paddle/fluid/framework/ddim.h
```diff
@@ -126,6 +126,67 @@ class DDim {
 
   std::string to_str() const;
 
+  DDim reshape(const std::vector<int>& shape) const {
```
Sand3r-: Since the header file only attempts to implement template and inline methods, perhaps it would be more suited to move the implementations of `reshape` and `transpose` to .cpp file

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -80,6 +81,19 @@ class MatMulFactory {
...
+    auto& fused_reshape_Out = ctx.Attr<std::vector<int>>("fused_reshape_Out");
+    auto& fused_transpose_Out =
+        ctx.Attr<std::vector<int>>("fused_transpose_Out");
+    return !fused_reshape_Out.empty() && !fused_transpose_Out.empty();
+  }
+
+  void correctStridesWhenOutputFused(const ExecutionContext& ctx,
+                                     const memory::dim N, memory::dim b,
+                                     memory::dims* out_strides) const {
+    if (IsOutputFused(ctx)) *out_strides = {N, b * N, 1};
```
Sand3r-: Please, correct me if I'm wrong, but I have an impression that the axis stored in `fused_transpose_Out` is currently completely disregarded in the code of dnnl's matmul, while as far I understand it, it influences how the strides are laid out. Perhaps it'd be advised to ensure with paddle enforce mechanism what transpose axis is currently supported.

paddle/fluid/framework/ddim.h
```diff
@@ -126,6 +126,67 @@ class DDim {
...
+    const int64_t copy_dim_val = 0;
+    const DDim& in_dims = *this;
+    DDim out_dims;
+    out_dims.rank_ = shape.size();
+    for (size_t i = 0; i < shape.size(); ++i) {
+      if (shape[i] == copy_dim_val) {
+        PADDLE_ENFORCE_LT(static_cast<int>(i), in_dims.size(),
+                          platform::errors::InvalidArgument(
+                              "The index of 0 in `shape` must be less than "
+                              "the dimensions. But received shape[%d] "
```
Sand3r-: `The index of 0 in `shape` must be less than the dimensions.`

Could you please clarify the message? @sfraczek Could you help here?
My best guess is that "Index %d of shape under which the value of 0 is stored, must be lower than the **number** of old dimensions." was meant.

Sand3r- (commented):
@arlesniak thank you so much for all the hard work you've spent on the PR. Good job in general, I really appreciate the UTs you've prepared, as well as keeping the solution clean in general. Please refer to my comments.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
sfraczek: I see it used for shape_out but its a different file

https://github.com/PaddlePaddle/Paddle/pull/23866/files/d93bb88437fbbb690d11338e80f36707cec3750a#diff-126803a76a1b0cbf6198f30e23058f8dR419-R436

arlesniak: It's used in the Op not the kernel

Sand3r-: That's right, thank you for clarification.

paddle/fluid/framework/ddim.h
sfraczek: Sounds good to me. We can take your suggestion.

sfraczek: @Sand3r-  Should we keep the part "But received shape[...."?

Sand3r-: Oh yes, totally :)

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
sfraczek: it doesn't compile. 
` error: invalid use of ‘dnnl::memory::dims’`

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
sfraczek: Why should we change batch size to 1 then?

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
sfraczek: Anyway, matmul didn't work when we multiplied batch size with 3rd dim so I guess batch 1 will also not work

sfraczek: Checking this modification would add +1 / +2 weeks to development and testing

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
arlesniak: done

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_mkldnn_op.py
arlesniak: done

paddle/fluid/framework/ddim.h
arlesniak: done

paddle/fluid/framework/ddim.h
arlesniak: done

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Sand3r-: To set 1 as the first entity in the shape of memory descriptors. This is what's done here anyway isn't it?:
```
b = BS / batch_size_;
```
After all BS is the same thing as batch_size_.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Sand3r-: > it doesn't compile.
> ` error: invalid use of ‘dnnl::memory::dims’`

It might be, I haven't compiled this but attempted to give a brief idea of what I mean, hence "An exemplary implementation could look like this:" disclaimer. 

> Anyway, matmul didn't work when we multiplied batch size with 3rd dim so I guess batch 1 will also not work

Hmm I don't quite understand how does this refer to the proposal? Nothing regarding batch size is changed here.

lidanqing-vv:
@luotao1  Please approve this PR_CI_CPU_Py2, it fail because of check_grad_ci. We want to cherry-pick it to 2.0

paddle/fluid/operators/matmul_op.cc
```diff
@@ -409,7 +409,45 @@ class MatMulOp : public framework::OperatorWithKernel {
...
+    if (!reshape_out.empty() && !transpose_out.empty()) {
+      auto reshape_out_size = reshape_out.size();
+      auto transpose_out_size = transpose_out.size();
+      PADDLE_ENFORCE_EQ(transpose_out_size, 4,
+                        platform::errors::InvalidArgument(
+                            "transpose_out_size supported rank is 4, "
+                            "received %d",
+                            transpose_out_size));
+      bool supported_transpose_axis =
+          (transpose_out.at(0) == 0 && transpose_out.at(1) == 2 &&
```
wojtuss: IMO if we check with enforce for supported transpose axis and reshape out size here, we should check these conditions in the fuse pass before executing it (now we apply the fuse unconditionally). With the current version, it should work fine with the Ernie model, but can lead to failure with some other model.

luotao1:
Please update your PR description.

paddle/fluid/operators/matmul_op.cc
```diff
@@ -448,6 +486,14 @@ class MatMulOpMaker : public framework::OpProtoAndCheckerMaker {
...
         "(bool, default false) Indicates if MKL-DNN kernel will be used")
         .SetDefault(false);
+    AddAttr<std::vector<int>>(
+        "fused_reshape_Out",
+        R"DOC(Shape of fused reshape of `Out` output.)DOC")
+        .SetDefault({});
+    AddAttr<std::vector<int>>(
+        "fused_transpose_Out",
+        R"DOC(Axis of fused transpose of `Out` output.)DOC")
+        .SetDefault({});
```
luotao1, line 498: Is this necessary to add two attributes in matmul_op? And the comments are not enough, maybe you can improve the comments in the next PR?

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
sfraczek: let's discuss this change after its merged. I can do proper testing then to verify if that change would still work. At least regarding my PR. Artur's PR is up to him but he probably feels the same on this.

paddle/fluid/operators/matmul_op.cc
sfraczek: It would be nice to see a pair of vectors for example because we require both attributes present anyway. Hower, we didn't want to extend the Attribute definition or hack it. Should we?
https://github.com/PaddlePaddle/Paddle/blob/a31d7328b70fde9c860dfad918222d7b2d842d71/paddle/fluid/framework/type_defs.h#L41-L45

arlesniak: I can improve comments in the next PR

luotao1 (dismissed):
LGTM

luotao1:
> Please approve this PR_CI_CPU_Py2

@lidanqing-intel Please have any other LGTM at first.

arlesniak:
@luotao1 I have improved comments and review fixes ready. Should I push them now or they can wait till next PR ?

paddle/fluid/operators/matmul_op.cc
arlesniak: Done

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Sand3r-: @sfraczek and I have figured out that it's actually not possible to access the dimensions directly from memory primitive. 
Instead we'd need to fetch the memory descriptor and then from it the actual dimensions which are represented by a vector returned by value. 
Such copying of vectors and multiple levels of indirection would actually result in slower execution.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Sand3r-: After a discussion with @sfraczek we've came to an understanding on how should this work. 
We've figured out that the batch_size_ should actually be assigned that way:
```C++
auto& x_dims = ctx.Input<Tensor>("X")->dims();
auto& y_dims = ctx.Input<Tensor>("Y")->dims();
batch_size_ = x_bs > y_bs ? x_dims[0] : y_dims [0];
```

The reset can stay the same.

arlesniak: @Sand3r- Doesn't the prior ENFORCE_EQ (line 112) rules the conditions earlier, so we don;t have to be worried about eqaulity of the x, y batches ?
 PADDLE_ENFORCE_EQ(x_bs > 0 && y_bs > 0 && x_bs != y_bs, false,
                          "If batch sizes of X and Y are positive,"
                          "they have to be equal."));

wojtuss (approved):
LGTM

luotao1 (approved):

wojtuss:
@luotao1 , is there still time to cherry-pick this PR to 2.0 release?

luotao1:
Yes, you can cherry-pick to 2.0 release

PR #23877, opened 2020-04-15
Remove redundant visit_FunctionDef

chenwhql:
The `visit_FunctionDef` to remove dygraph_to_static decorator has been implemented in `DygraphToStaticAst`, I think the same code in `BasicApiTransformer` is redundant.


liym27 (approved):
LGTM

PR #23893, opened 2020-04-15
fix example bug for tensor eye,t,min 

danleifeng:
1. add blank line for min tensor api
2. correct explanation for eye tensor api
3. fix type example for t tensor api


Boyan-Liu (approved):

guru4elephant (approved):
LGTM

PR #23900, opened 2020-04-15
[dy2static]Support recursive call

liym27:
Support recursive call.
1. Add a api `convert_call` to call ast transformer recursively.

For tests:
1. Remove redundant decorator in the existing test files
2. The new test file `test_recursive_call` tests other situations.


zhhsplendid:
test_recursive_call failed at python35 CIs, please check whether there is error under python3

liym27:
> test_recursive_call failed at python35 CIs, please check whether there is error under python3

The test failed because difference exists in some third party functions between python3 and python2, I'm trying to fix the bug.

PR #23905, opened 2020-04-16
support set_test_mode

hutuxian:
To solve no device space problem, add a new interface "set_test_mode" to notify boxps to do some optimization.
What's more, we set low log level(4) for comlog in AFSAPI to release the disk pressure.


gavin1332 (approved):
LGTM

PR #23751, opened 2020-04-11
OP Normal, Uniform, Xavier Initializer, smooth_l1, mean_iou error message enhancement

slf12:
OP Normal, Uniform, Xavier Initializer, smooth_l1, mean_iou error message enhancement

PR #23912, opened 2020-04-16
update enhance error message for Initializer, smooth_l1

slf12:
print number in error message when InvalidArgument

update for https://github.com/PaddlePaddle/Paddle/pull/23751


chenwhql (approved):
LGTM

PR #23938, opened 2020-04-17
fix AMP and recompute, test=develop (#23551)

mapingshuo:
* allow amp and recompute working together


sneaxiy (approved):
LGTM for backward.py.

PR #23742, opened 2020-04-11
OP clip, merge_lod_tensor, convert/elementwise error message enhancement

haozech:
test=develop

PR #23944, opened 2020-04-17
OP clip, merge_lod_tensor, convert/elementwise error message enhancem…

haozech:
…ent (#23742)

* OP clip, merge_lod_tensor, convert/elementwise error message enhancement. test=develop


fc500110 (approved):
LGTM

PR #23948, opened 2020-04-17
[Cherry-pick Release/2.0] error message enhancement for l2_normalize,  etc. #23783

ZHUI:
OP error message enhancement for l2_normalize, matmul, mean, etc. (#23783) 
test=release/2.0

* fix error message of l2_normalize, matmul, mean, etc.
* add the test case for those ops


wawltor (approved):
LGTM

PR #23955, opened 2020-04-17
Fix OOM of paddleboxdatafeed

ForFishes:
Because the `std::vector<T> input_records_` holds the real data memory, but we don't use `postprocess_instance` function, it will not release the vecotor memory and cause OOM. Therefore,r we should  release the memory in `ReleaseMemory()` function.


gavin1332 (approved):
LGTM

PR #23961, opened 2020-04-18
fix logger propagate

heavengate:
**fix logger propagate**
`logger.propagate` is True by default, log message will propagate from `logger.handler` to default root handler and print twice, set `logger.propagate` as False to prevent this


wanghaoshuang (approved):
LGTM

PR #23978, opened 2020-04-20
Add the support dygraph out attribute for the op of mm in api2.0

wawltor:
Add the support of out parameter  for the op of mm in dygraph mode


zhiqiu (approved):
LGTM

ZeyuChen (approved):
LGTM

PR #23984, opened 2020-04-20
TRT fc fuse pass test

FrostML:
Diff occurred between GPU and TRT. 
image

After having a discussion with NHZIX, in order to provide TRT CI ASAP, this test for trt part is disabled temporarily. 
CPU and GPU tests still work fine.


NHZlX (approved):
LGTM

PR #24017, opened 2020-04-20
Add batch_fc op in contrib

ForFishes:
This Op can calculate batch fc. Notice: It currently supports GPU device. This Op exists in contrib, which means that it is not shown to the public. 
Part dim of batch_fc_op:
X.dim = slot_pairs_num * ins_num * in_dim
W.dim = slot_pairs_num * in_dim * out_dim
b.dim = slot_pairs_num * out_dim
output.dim = slot_pairs_num * ins_num * out_dim

### DIFF
--- a/python/paddle/fluid/contrib/layers/nn.py
+++ b/python/paddle/fluid/contrib/layers/nn.py
@@ -1355,7 +1355,7 @@ def batch_fc(input, param_size, param_attr, bias_size, bias_attr, act=None):
     pre_act = helper.create_variable_for_type_inference(dtype)
     helper.append_op(
         type="batch_fc",
-        inputs={"Input": input_var,
+        inputs={"Input": input,
                 "W": w,
                 "Bias": b},
         outputs={"Out": pre_act})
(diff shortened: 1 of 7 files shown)


gavin1332 (commented):
Does bias add and activation calculated separately?

gavin1332 (commented):
LGTM

lanxianghit (dismissed):

python/paddle/fluid/contrib/layers/nn.py
```diff
@@ -1298,3 +1298,65 @@ def rank_attention(input,
         attrs={"MaxRank": max_rank,
                "MaxSize": max_size})
     return output
+
+
+def batch_fc(input, param_size, param_attr, bias_size, bias_attr, act=None):
+    """
+    **Batch FC layer**
+    This Op can calculate BatcheFC.
```
Boyan-Liu: BatcheFC是拼写错误？

ForFishes: OK， i will fix it.

Boyan-Liu: 也请解释下具体释义，目前不能直观理解op功能

ForFishes: OK, i will fix it,thanks~

Boyan-Liu (approved):

lanxianghit (approved):

PR #24070, opened 2020-04-22
Fix isolated var fetch bug

sneaxiy:
Fix the bug that ParallelExecutor cannot fetch variables that are not in/out of any op.


chenwhql (approved):
LGTM

PR #24079, opened 2020-04-22
Monitor Framework

hutuxian:
### PR types

New features

### PR changes

Others

### Describe

Monitor Framework
This PR is mainly for developer to monitor some runtime states in Paddle.  Developer can use this feature in the following scenario :
* running with a shared library, and want to know how many resources used belong to paddle.
* need to observe the real-time changes of system or application states.

Design
* Add a StatValue class in the backend to represent a stat.
* Add a singleton `StatRegistry` to maintain the collection of stats.
* For the sake of code neatness, we only support type of `int` and `float`, which can cover most of the scenarios.

Usage
* Registration your stat in `monitor.cpp`, using `DEFINE_INT_STATUS` or `DEFINE_FLOAT_STATUS`
* Call `USE_INT_STAT` or `USE_INT_STAT` in the source file where you want to update the stat.
* Call `STAT_ADD` or `STAT_SUB` to update your state value.
* Through the PyBind interface, the value of each state is exposed to the python side.
* After get the real-time state value in python, you can do whatever you want, such as print it to text file or display it by monitoring service.

Others
* We will give a complete example later, please stay tuned.


paddle/fluid/platform/gpu_info.cc
```diff
@@ -32,7 +33,14 @@ DECLARE_string(selected_gpus);
...
 constexpr static float fraction_reserve_gpu_memory = 0.05f;
-
+USE_STAT(STAT_gpu0_mem_size);
+USE_STAT(STAT_gpu1_mem_size);
+USE_STAT(STAT_gpu2_mem_size);
+USE_STAT(STAT_gpu3_mem_size);
+USE_STAT(STAT_gpu4_mem_size);
+USE_STAT(STAT_gpu5_mem_size);
+USE_STAT(STAT_gpu6_mem_size);
+USE_STAT(STAT_gpu7_mem_size);
```
gavin1332: Too strong assumption about the number of GPUs

paddle/fluid/platform/monitor.h
```diff
@@ -0,0 +1,182 @@
...
+
+namespace paddle {
+namespace platform {
+
+template <typename T>
+class StatRegistry;
+
+class MonitorRegistrar {
+ public:
+  // The design is followed by OperatorRegistrar: To avoid the removal of global
```
gavin1332, line 35: OperatorRegistrar

paddle/fluid/platform/monitor.h
```diff
@@ -0,0 +1,182 @@
...
+  int TouchStatRegistrar_##item() {                    \
+    _##item.Touch();                                   \
+    return 0;                                          \
+  }
+
+#define USE_STAT(item)                    \
+  extern int TouchStatRegistrar_##item(); \
+  UNUSED static int use_stat_##item = TouchStatRegistrar_##item()
+
+REGISTER_INT_STATUS(STAT_total_feasign_num_in_mem)
```
gavin1332: why not move these registration to where they need?

paddle/fluid/platform/monitor.cc
```diff
@@ -0,0 +1,30 @@
...
+
+DEFINE_INT_STATUS(STAT_total_feasign_num_in_mem)
+DEFINE_INT_STATUS(STAT_gpu0_mem_size)
+DEFINE_INT_STATUS(STAT_gpu1_mem_size)
+DEFINE_INT_STATUS(STAT_gpu2_mem_size)
+DEFINE_INT_STATUS(STAT_gpu3_mem_size)
+DEFINE_INT_STATUS(STAT_gpu4_mem_size)
+DEFINE_INT_STATUS(STAT_gpu5_mem_size)
+DEFINE_INT_STATUS(STAT_gpu6_mem_size)
+DEFINE_INT_STATUS(STAT_gpu7_mem_size)
```
gavin1332, line 30: Can we directly define these stats without declaration?

paddle/fluid/platform/gpu_info.cc
gavin1332, line 43: I think it is better move these monitor

paddle/fluid/platform/monitor.h
hutuxian: followed by OperatorRegistrar

paddle/fluid/platform/monitor.cc
hutuxian: The declaration may be necessary, cause it must be recognized by the file using it.
An alternative way is to declare it in USE_STAT

paddle/fluid/platform/gpu_info.cc
```diff
@@ -476,5 +472,13 @@ bool IsCudaMallocRecorded(int dev_id) {
   return RecordedCudaMallocHelper::Instance(dev_id)->NeedRecord();
 }
 
+void UpdateGpuMemStat(int mode, int dev_id, size_t size) {
+  if (mode == 0) {
+    STAT_INT_ADD("STAT_gpu" + std::to_string(dev_id) + "_mem_size", size);
```
gavin1332: where is the definition of STAT_INT_ADD and STAT_INT_SUB?

paddle/fluid/platform/gpu_info.cc
```diff
@@ -476,5 +472,13 @@ bool IsCudaMallocRecorded(int dev_id) {
   return RecordedCudaMallocHelper::Instance(dev_id)->NeedRecord();
 }
 
+void UpdateGpuMemStat(int mode, int dev_id, size_t size) {
```
gavin1332: Bad code smell of the use of `mode` for arithmetic method selection. I prefer the original `STAT_INT_ADD` statement to the use of `mode` parameter if no better way could be found.

hutuxian: done

paddle/fluid/platform/monitor.h
hutuxian: Maybe for monitor stat, putting them together is more clear?

paddle/fluid/platform/gpu_info.cc
hutuxian: Agree, and wrap them in another Macro

gavin1332 (approved):
LGTM

PR #24083, opened 2020-04-22
Dev/add fake_interface_only decorator for some function of Variable

zhiqiu:
In PR #21359, we split VarBase from Variable. Currently, Variable (for static graph) and VarBase (for dygraph) are logically same type for developers. However, some of the internal APIs has small differences. This PR handles those APIs.

1.  Some functions of Variable is not supported, such as `detach()`. Historically, these APIs are decorated by `dygraph_only` when VarBase have not been split from Variable. Basically, we should remove them.  While, actually, now we keep these functions under class Variable for generating API docs only, since VarBase is not exposed and not mentioned in our official website. 
So, I add a `fake_interface_only` decorator for those APIs.

2. Some functions of Variable is not supported for dygraph Variable when VarBase have not been split from Variable, like `lod_level`. Historically, these APIs are decorated by `dygraph_not_support`. Since we have split VarBase, users cannot access these APIs by  VarBase.
So, I remove the `dygraph_not_support` decorators for those APIs.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -3879,7 +3879,6 @@ class Program(object):
     def _version(self):
         return self.desc._version()
 
-    @dygraph_not_support
     def clone(self, for_test=False):
         """
         **Notes**:
@@ -4555,7 +4554,6 @@ class Program(object):
                 if other_var.stop_gradient:
                     var.stop_gradient = True
 
-    @dygraph_not_support
     def list_vars(self):
         """
         Get all :ref:`api_guide_Variable_en` from this Program. A iterable object is returned.
@@ -4578,7 +4576,6 @@ class Program(object):
             for each_var in list(each_block.vars.values()):
                 yield each_var
 
-    @dygraph_not_support
     def all_parameters(self):
         """
         Get all :ref:`api_guide_parameter_en` from this Program. A list object is returned.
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/framework.py
```diff
@@ -217,8 +217,25 @@ def __impl__(*args, **kwargs):
...
+# used to make Variable and VarBase has same interfaces, like numpy. Since VarBase is not exposed in our
+# official docments, logically, we want to keep VarBase and logically consistent. While, actually,
+# in our implementation, there some APIs not supported, like numpy, because Variable contains the desc.
+# So, those APIs are listed under class Variable to generate docs only.
+# TODO(zhiqiu): We should make VarBase consistent with Variable in future, for example, by inheritting
+# same base class. 
+def _fake_interface_only_(func):
+    def __impl__(*args, **kwargs):
+        raise AssertionError(
+            "'%s' should be called by Dygraph Varible in Dygraph mode, please use fluid.dygraph.guard() as context to run it in Dygraph Mode"
```
phlrain: 这个Dygraph 改成Imperative，我们正式发布会使用Imperative 这个名称

zhiqiu: 好的

sneaxiy (approved):
LGTM.

PR #24103, opened 2020-04-23
fix error when out is not None in std api

slf12:
当out是Variable, if out 判断的不是out是否为None, 触发了动态图中的bool判断，https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/dygraph/varbase_patch_methods.py#L209， 所以改为 if out is not None


baiyfbupt (approved):
LGTM

PR #24106, opened 2020-04-23
[Paddle-TRT] fix memory leak and unsafe emb_norm plugin

jeng1220:
`IPlugin*::initialize` and `IPlugin*::terminate` should be always implemented as a pair. Also, resources acquired in `IPlugin*::initialize` should be released in `IPlugin*::terminate`.  Otherwise, it causes memory leak.

The all data members of cloned plugin must be set in `PluginV2::clone`. The plugin only calls `IPlugin*::initialize` once when creating ICudaEngine. The cloned plugin won't call `IPlugin*::initialize` after IExecutionContext is created, but TensorRT may clone the plugin during the creation of IExecutionContext and use the cloned plugin to infer the model.

Read https://docs.nvidia.com/deeplearning/sdk/tensorrt-developer-guide/index.html#ipluginext to know more.


jeng1220:
@NHZlX please take a look

PR #24123, opened 2020-04-23
[DNNL] activations Inplace support

jczaja:
This PR is extending DNNL inplace mechanism to support activation op. It is follow up to PR#23477.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
+++ b/paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
@@ -23,9 +23,10 @@ USE_OP(softmax);
 USE_OP_DEVICE_KERNEL(softmax, MKLDNN);
 USE_OP(elementwise_add);
 USE_OP_DEVICE_KERNEL(elementwise_add, MKLDNN);
+USE_OP(gelu);
+USE_OP_DEVICE_KERNEL(gelu, MKLDNN);
 USE_OP(relu);
 USE_OP(tanh);
-USE_OP(gelu);
 USE_OP_DEVICE_KERNEL(tanh, MKLDNN);
 
 namespace paddle {
--- a/paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
+++ b/paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
@@ -24,6 +24,8 @@ USE_OP_DEVICE_KERNEL(softmax, MKLDNN);
 USE_OP(elementwise_add);
 USE_OP_DEVICE_KERNEL(elementwise_add, MKLDNN);
 USE_OP(relu);
+USE_OP(tanh);
+USE_OP_DEVICE_KERNEL(tanh, MKLDNN);
 
 namespace paddle {
 namespace framework {
@@ -48,6 +50,8 @@ class MKLDNNInplacePassTest {
       op->SetInput("Filter", {inputs[1]});
       op->SetInput("Bias", {inputs[2]});
     } else if (type == "relu") {
+      op->SetInput("X", inputs);
+    } else if (type == "tanh") {
       op->SetInput("X", inputs);
     } else if (type == "softmax") {
       op->SetAttr("axis", -1);
@@ -90,6 +94,12 @@ class MKLDNNInplacePassTest {
     SetOp(&prog, "relu", "relu2", std::vector<std::string>({"j"}),
           std::vector<std::string>({"k"}),
           mkldnn_enabled_op.compare("softmax") == 0);
+    SetOp(&prog, "tanh", "tanh1", std::vector<std::string>({"k"}),
+          std::vector<std::string>({"l"}),
+          mkldnn_enabled_op.compare("tanh") == 0);
+    SetOp(&prog, "relu", "relu2", std::vector<std::string>({"l"}),
+          std::vector<std::string>({"m"}),
+          mkldnn_enabled_op.compare("relu") == 0);
     if (branched == true) {
       SetOp(&prog, "softmax", "softmax2", std::vector<std::string>({"g"}),
             std::vector<std::string>({"z"}),
@@ -113,11 +123,6 @@ class MKLDNNInplacePassTest {
     std::unordered_map<std::string, std::string> input_names;
     std::unordered_map<std::string, std::string> output_names;
 
-    input_names["softmax"] = "X";
-    output_names["softmax"] = "Out";
-    input_names["elementwise_add"] = "X";
-    output_names["elementwise_add"] = "Out";
-
     VLOG(3) << DebugString(graph);
 
     for (auto* node : graph->Nodes()) {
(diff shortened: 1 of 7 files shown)


grygielski (approved):
LGTM

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
```diff
@@ -47,8 +52,14 @@ class MKLDNNInplacePassTest {
       op->SetInput("Input", {inputs[0]});
       op->SetInput("Filter", {inputs[1]});
       op->SetInput("Bias", {inputs[2]});
+    } else if (type == "gelu") {
```
wojtuss, line 55: What about grouping:
```c++
} else if (std::set<std::string>{"gelu", "leaky_relu", "relu", "tanh"}.count(type)) {
   op->SetInput("X", inputs);
} ...

File: paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc [hunk 2]
```diff
@@ -90,6 +101,18 @@ class MKLDNNInplacePassTest {
...
+          mkldnn_enabled_op.compare("tanh") == 0);
+    SetOp(&prog, "relu", "relu2", std::vector<std::string>({"l"}),
+          std::vector<std::string>({"m"}),
+          mkldnn_enabled_op.compare("relu") == 0);
+    SetOp(&prog, "leaky_relu", "leaky_relu1", std::vector<std::string>({"m"}),
+          std::vector<std::string>({"n"}),
+          mkldnn_enabled_op.compare("leaky_relu") == 0);
+    SetOp(&prog, "gelu", "gelu1", std::vector<std::string>({"n"}),
+          std::vector<std::string>({"m"}),
+          mkldnn_enabled_op.compare("relu") == 0);
```
wojtuss, line 115: `relu` -> `gelu`

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
```diff
@@ -90,6 +101,18 @@ class MKLDNNInplacePassTest {
     SetOp(&prog, "relu", "relu2", std::vector<std::string>({"j"}),
           std::vector<std::string>({"k"}),
           mkldnn_enabled_op.compare("softmax") == 0);
```
wojtuss, line 103: `softmax`->`relu`

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
```diff
@@ -90,6 +101,18 @@ class MKLDNNInplacePassTest {
...
+          std::vector<std::string>({"l"}),
+          mkldnn_enabled_op.compare("tanh") == 0);
+    SetOp(&prog, "relu", "relu2", std::vector<std::string>({"l"}),
+          std::vector<std::string>({"m"}),
+          mkldnn_enabled_op.compare("relu") == 0);
+    SetOp(&prog, "leaky_relu", "leaky_relu1", std::vector<std::string>({"m"}),
+          std::vector<std::string>({"n"}),
+          mkldnn_enabled_op.compare("leaky_relu") == 0);
+    SetOp(&prog, "gelu", "gelu1", std::vector<std::string>({"n"}),
+          std::vector<std::string>({"m"}),
```
wojtuss, line 114: `m`->`o`?

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
```diff
@@ -90,6 +101,18 @@ class MKLDNNInplacePassTest {
     SetOp(&prog, "relu", "relu2", std::vector<std::string>({"j"}),
           std::vector<std::string>({"k"}),
           mkldnn_enabled_op.compare("softmax") == 0);
+    SetOp(&prog, "tanh", "tanh1", std::vector<std::string>({"k"}),
+          std::vector<std::string>({"l"}),
+          mkldnn_enabled_op.compare("tanh") == 0);
+    SetOp(&prog, "relu", "relu2", std::vector<std::string>({"l"}),
```
wojtuss, line 107: repeated `relu`?

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
```diff
@@ -153,6 +172,15 @@ TEST(MKLDNNInplacePass, inplace_elementwise_add) {
...
 }
+TEST(MKLDNNInplacePass, inplace_tanh) {
+  MKLDNNInplacePassTest().MainTest("tanh", false, 1);
+}
+
+TEST(MKLDNNInplacePass, inplace_leaky_relu) {
+  // Input of leaky_relu is used as output of subsequent gelu, so no inplace
+  // cannot be done
+  MKLDNNInplacePassTest().MainTest("leaky_relu", false, 0);
+}
```
wojtuss, line 183: What about adding tests for `relu` and `gelu`?

luotao1 (approved):

luotao1:
Sorry, I miss the comment from @wojtuss, please fix in next PR

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
jczaja: Gelu technically is capable of in-place , but PaddlePaddle does not provide InplaceInferer for Gelu (gelu is implemented separatly from other activations) . In other words , currently inplace pass will not work for GELU as inplaceinferer is not present , once in general code there is inplaceinferer for gelu it all starts to work. As for relu/leaky_relu itp. In our code it is the same implementation. So does not matter which activation I choose for UT. Those tests are more for testing certain graphs confguration.

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
jczaja: goot suggestion , will do that

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
jczaja: good catch!

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
jczaja: Yes, I should have give name relu3. good catch

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
jczaja: No:) This is intentional. Some aggressive optimization may force paddle to reuse buffers. If one op is having input A that , and that op may run  in-place execution then there may be a situation when we do not do in-place which is when next op is using tensor A as output and this next op cannot do inplace.
So Initial graph is like this:   A -> inplace_to_be_op -> B -> next_op -> A   , then we cannot do:
A -> inplace_to_be_op -> A -> next_op -> A , becouse we would enforce that next_op is working inplace , while it cannot

paddle/fluid/framework/ir/mkldnn/mkldnn_inplace_pass_tester.cc
jczaja: good catch

PR #24136, opened 2020-04-24
fix the set dtype bug of uniform_random op

ZHUI:
Fix the set dtype bug of uniform_random op. 
The attrs of  uniform_random is missing, and the dtype of out variable will always be the default FP32.


wawltor (approved):
LGTM

PR #24155, opened 2020-04-26
Opt the postprocess in dataset

ForFishes:
Because the `postprocess_instance` function has the `vector<instance> input_records_` and `localshuffle` need move the record from channel to vector, we can save the process in `localshuffle` and shuffle in `postprocess_instance`  when it is in `merge_pv` mode. In fact, we find it can save about 5 min to train one day.


gavin1332 (approved):
LGTM

PR #24178, opened 2020-04-26
add the graph batch reader for pslib mode

wawltor:
Add the graph batch reader for pslib mode, support the PGL training in pslib mode.


xjqbest (approved):
LGTM

PR #24229, opened 2020-04-28
Check nan and inf in dygraph

wanghaoshuang:
Check nan of inf in dygraph.


python/paddle/fluid/tests/unittests/check_dygraph_nan_inf_base.py
```diff
@@ -0,0 +1,113 @@
...
+        try:
+            check(use_cuda=True)
+            assert False
+        except Exception as e:
+            print(e)
+            print(type(e))
+            # Note. Enforce in cuda kernel may not catch in paddle, and
+            # Exception type will be RuntimeError
+            assert type(e) == core.EnforceNotMet or type(e) == RuntimeError
+#    try:
```
wangxicoding, line 107: need cpu unittest?

PR #23091, opened 2020-03-18
[Dy2Stat]Support LoDTensorArray for slice op

liym27:
1. Support the input of slice is LoDTensorArray;
2. Support read elements of list in dygraph_to_static. For example:
```Python
arr = fluid.layers.array_write(x, i)
res1 = arr[i]
res2 = arr[0:i]
```

Note: https://github.com/PaddlePaddle/Paddle/pull/24250  (Support list pop in dy2stat) depends on this PR.

PR #24250, opened 2020-04-28
[Dy2Stat]Support list pop

liym27:
#### 1. Support list `pop` in dygraph to static.

- For python list, support `pop(i)` which pops the i-th item and `pop()` which pops the last item.

- For TensorArray, support `pop()`  and `pop(i)`.

- Before
```Python
@declarative
def test_list_pop_without_control_flow_1(x):
    a = []
    if 2 > 1:
        a.append(x)
    a.pop()
    return a
```

- After transformer
```Python
def test_list_pop_without_control_flow_1(x):
    x = fluid.layers.assign(x)
    a = []
    if 2 > 1:
        fluid.dygraph.dygraph_to_static.convert_call(a.append)(x)
    fluid.dygraph.dygraph_to_static.convert_list_pop(a, None)
    return a
```

2.Support an empty tensor array as Input, and set the kernel type is float. 
Because in control flow,  an empty tensor array needs to be assigned out to the parent block. And when the last item is poped, the LoDTensprArray need to be assigned  with an empty LoDTensorArray.

3.Fix bug in assign op: support to infer shape for LOD_TENSOR_ARRAY 
https://github.com/PaddlePaddle/Paddle/pull/24268

4.Fix bug in `ProgramTranslator.get_output`, convert all items into `VarBase` in nested list
https://github.com/PaddlePaddle/Paddle/pull/24267

And fetch rebase https://github.com/PaddlePaddle/Paddle/pull/23091, support LoDTensorArray for slice op.

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1274,6 +1274,11 @@ void OperatorWithKernel::ParseInputDataType(
         t = &(var->Get<SelectedRows>().value());
       } else if (var->IsType<LoDTensorArray>()) {
         auto t_arr = var->Get<LoDTensorArray>();
+        // NOTE(liym27): Support an empty tensor array as Input.
+        // And set the kernel type is float.
+        if (t_arr.size() == 0) {
+          *data_type = proto::VarType::FP32;
+        }
         for (size_t j = 0; j < t_arr.size(); j++) {
           if (t_arr[j].IsInitialized()) {
             t = &(t_arr[j]);
@@ -1296,6 +1301,7 @@ void OperatorWithKernel::ParseInputDataType(
                 "previous variable type is (%s).",
                 Type(), name, DataTypeToString(tmp),
                 DataTypeToString(*data_type)));
+
         *data_type = tmp;
       }
     }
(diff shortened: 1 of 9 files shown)


paddle/fluid/framework/operator.cc
```diff
@@ -1274,6 +1274,11 @@ void OperatorWithKernel::ParseInputDataType(
         t = &(var->Get<SelectedRows>().value());
       } else if (var->IsType<LoDTensorArray>()) {
         auto t_arr = var->Get<LoDTensorArray>();
+        // NOTE(liym27): Support an empty tensor array as Input.
+        // And set the kernel type is float.
+        if (t_arr.size() == 0) {
+          *data_type = proto::VarType::FP32;
```
zhhsplendid: Is this necessary? I understand that the program may be wrong if we don't set it, but it is unclear whether set default kernel type to FP32 will cause problem.

I'm fine with the change here but just asked is there other way to do so?

zhhsplendid (dismissed):
Other things LGTM

paddle/fluid/framework/operator.cc
liym27: > Is this necessary? I understand that the program may be wrong if we don't set it, but it is unclear whether set default kernel type to FP32 will cause problem.

There seems to be no problem at present, because only a few OPs without kernel support that inputs are LoDTensorArray and `assign` op with kernel support LoDTensorArray.  But as you said, there may be problems.

> I'm fine with the change here but just asked is there other way to do so?

Because empty tensor array is used in assign op and assign op has kernel, I will try to  set the kenel type in assign op when tensor array is empty .

Aurelius84 (dismissed):
LGTM

paddle/fluid/framework/operator.cc
zhhsplendid: OK, LGTM

zhhsplendid (approved):
LGTM

Boyan-Liu (approved):

PR #24068, opened 2020-04-22
[dy2static] Add print transformer and unify print format

chenwhql:
## Print Transformer

`print`在PY2和PY3中存在诸多不同之处，详见代码中单测

转换前代码示例：
```
def dyfunc_print_variable(x):
    x_v = fluid.dygraph.to_variable(x)
    # NOTE: transform to static code, var name will be changed
    print(x_v)
```

转换后代码示例：
```
def dyfunc_print_variable(x):
    x_v = fluid.layers.assign(x)
    fluid.layers.Print(x_v)
```

### 修改前
- 原格式，无论哪个都无法令人满意
- 最重要的是，两种打印格式无法统一

#### print(VarBase)
```
name generated_var_0, dtype: VarType.INT32 shape: [5]   lod: {}
        dim: 5
        layout: NCHW
        dtype: int
        data: [1 1 1 1 1]
```
问题包括但不限于：
- lod位置突兀
- 两个dtype，重复
- shape和dim重复
- 缺少place信息
- 整体信息混乱

#### layers.Print(Variable)示例
```
1587219215              The place is:CUDAPlace(0)
Tensor[assign_0.tmp_0]
        shape: [5,]
        dtype: i
        data: 1,1,1,1,1,
```
问题包括但不限于：
- time打出来有多大意义
- message参数没有会导致第一行中间两个\t\t，不美观
- dtype的name没有demangle转换，导致int显示为i，当然其他类型也会显示有问题
- 没有layout
- 细节处理糟糕，最后的逗号对强迫症患者不友好

### 修改后

修改要点：
- print(VarBase)和layers.Print(Variable)主要都是打印类型为LoDTensor的Variable，在LoDTensor上结果格式应该统一
- `\t`统一转换成`  -`，`\t`有时候是4个空格，有时候是8个，不美观
- 如果lod为空，就不打印了
- 统一成如下格式：
```
Variable: (Name)
  - lod: ***
  - place: ***
  - shape: ***
  - layout: ***
  - dtype: ***
  - data: ***
```
结果示例：
```
Variable: assign_0.tmp_0
  - place: CUDAPlace(0)
  - shape: [5]
  - layout: NCHW
  - dtype: int
  - data: [1 1 1 1 1]
```

## 文档变动

image

## 补充问题

1. 为什么还要给Print op新增参数？为什么不干脆在Print op中直接print(Tensor)?

[...]

PR #24251, opened 2020-04-28
Tensor Reorder Memory Dump v3.0

pawelpiotrowicz:
* export TENSOR_DUMP_OPERATORS="fc_mkldnn_fwd,elemwise_fwd{tanh},elemwise_fwd{relu},transpose_mkldnn_fwd=NHWC"

* export TENSOR_DUMP_FILE="/dev/stdout"

* export TENSOR_DUMP_LIMIT_SIZEOF_1="128"

* export TENSOR_DUMP_LIMIT_SIZEOF_4="128"

* export TENSOR_DUMP_SYNCHRONIZE="1" /* in case of parallel executor */

test=develop

### DIFF
--- a/paddle/fluid/framework/tensor.cc
+++ b/paddle/fluid/framework/tensor.cc
@@ -120,5 +120,25 @@ void Tensor::ResetHolderWithType(std::shared_ptr<memory::Allocation> holder,
   type_ = type;
 }
 
+int Tensor::dump(const char* label, const char* name) {
+  if (TensorDumpConfig<>::get().is_disabled()) {
+    return -1;
+  }
+
+  auto it = TensorDumpConfig<>::get().fetchOperator(label);
+
+  if (it == TensorDumpConfig<>::get().fetchOperatorEnd()) {
+    return -2;
+  }
+
+  DumpComposit<uint8_t, int8_t, float>::execute(label, name, *this, it->layout);
+  return 0;
+}
+
+void* Tensor::raw_data() const {
+  return reinterpret_cast<void*>(reinterpret_cast<uintptr_t>(holder_->ptr()) +
+                                 offset_);
+}
+
 }  // namespace framework
 }  // namespace paddle
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -13,16 +13,19 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 #pragma once
-
 #include <cstdint>
 #include <cstring>
 #include <memory>
+#include <string>
+#include <tuple>
 #include <typeindex>
 #include <utility>
 #include <vector>
 #include "paddle/fluid/framework/data_layout.h"
+#include "paddle/fluid/framework/data_type.h"
 #include "paddle/fluid/framework/ddim.h"
 #include "paddle/fluid/framework/framework.pb.h"
+#include "paddle/fluid/framework/tensor_extension.h"
 #include "paddle/fluid/memory/memory.h"
 #include "paddle/fluid/platform/device_context.h"
 #include "paddle/fluid/platform/enforce.h"
@@ -79,6 +82,16 @@ class Tensor {
   /*! Return a pointer to constant memory block. */
   template <typename T>
   const T* data() const;
+
+  template <class T>
+  bool hasType() const {
+    return type_ == ::paddle::framework::DataTypeTrait<T>::DataType();
+  }
+
+  void* raw_data() const;
+
+  /*! Serialize tensor to file with label name */
+  int dump(const char* label, const char* name = nullptr);
 
   inline bool IsInitialized() const;
 
--- a/paddle/fluid/framework/tensor_extension.h
+++ b/paddle/fluid/framework/tensor_extension.h
@@ -1,0 +1,496 @@
+// Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+#include <memory>
+#include <mutex>
+#include <string>
+#include <tuple>
+#include <typeindex>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include <cstdint>
+#include <cstring>
+#include <dnnl.hpp>
+#include "boost/crc.hpp"
+
+namespace paddle {
+namespace framework {
+
+// ########################
+
+namespace colors {
+
+constexpr const char* RESET() { return "\033[0m"; }
+
+constexpr const char* RED() { return "\033[0;31m"; }
+
+constexpr const char* GREEN() { return "\033[1;32m"; }
+
+constexpr const char* CYAN() { return "\033[0;36m"; }
+
+#define msg_color(c, x) \
+  std::cout << c << "[tensor-dump]: " << x << colors::RESET() << std::endl;
+#define msg_green(x) msg_color(colors::GREEN(), x)
+}  // namespace colors
+
+template <DataLayout LayoutType>
+struct ConvPaddleLayout2Mkldnn;
+
+template <>
+struct ConvPaddleLayout2Mkldnn<DataLayout::kNCHW> {
+  static const dnnl::memory::format_tag value = dnnl::memory::format_tag::nchw;
+};
+
+template <>
+struct ConvPaddleLayout2Mkldnn<DataLayout::kNHWC> {
+  static const dnnl::memory::format_tag value = dnnl::memory::format_tag::nhwc;
+};
+
+template <class L>
+dnnl::memory::format_tag tensor_layout2_mkldnn(L layout) {
+  using tag = dnnl::memory::format_tag;
+
+  if (layout == DataLayout::kNHWC) return tag::nhwc;
+
+  return tag::nchw;
+}
+
+template <class T>
+struct type2mkldnntype;
+
+template <>
+struct type2mkldnntype<float> {
+  static const dnnl::memory::data_type value = dnnl::memory::data_type::f32;
+};
+
+template <>
+struct type2mkldnntype<uint8_t> {
+  static const dnnl::memory::data_type value = dnnl::memory::data_type::u8;
+};
+
+template <>
+struct type2mkldnntype<int8_t> {
+  static const dnnl::memory::data_type value = dnnl::memory::data_type::s8;
+};
+
+template <class T, class TensorType>
+void reorder_via_mkldnn(void* out, TensorType* tensor_obj, DataLayout layout) {
+  dnnl::engine::kind engine_kind = dnnl::engine::kind::cpu;
+  dnnl::engine engine(engine_kind, 0);
+  // Create dnnl::stream.
+  dnnl::stream engine_stream(engine);
+  dnnl::memory::dims src_dims;
+  auto& paddle_dims = tensor_obj->dims();
+
+  for (decltype(paddle_dims.size()) i = 0; i < paddle_dims.size(); ++i)
+    src_dims.push_back(paddle_dims.at(i));
+  auto current_layout = tensor_obj->layout();
+  /*
+  std::cout << "Reorder from=" << DataLayoutToString(current_layout)
+            << " to=" << DataLayoutToString(layout) << std::endl;
+  */
+  auto src_md = dnnl::memory::desc(src_dims, type2mkldnntype<T>::value,
+                                   tensor_layout2_mkldnn(current_layout));
+  auto dst_md = dnnl::memory::desc(src_dims, type2mkldnntype<T>::value,
+                                   tensor_layout2_mkldnn(layout));
+
+  auto src_mem = dnnl::memory(src_md, engine, tensor_obj->raw_data());
+  auto dst_mem = dnnl::memory(dst_md, engine, out);
+
+  // Create primitive descriptor.
+  auto reorder_pd =
+      dnnl::reorder::primitive_desc(engine, src_md, engine, dst_md);
+  // Create the primitive.
+  auto reorder_prim = dnnl::reorder(reorder_pd);
+  // Primitive arguments.
+  std::unordered_map<int, dnnl::memory> reorder_args;
+  reorder_args.insert({DNNL_ARG_SRC, src_mem});
+  reorder_args.insert({DNNL_ARG_DST, dst_mem});
+  // Primitive execution: reorder with scaled sum.
+  reorder_prim.execute(engine_stream, reorder_args);
+  // Wait for the computation to finalize.
+  engine_stream.wait();
+}
+
+template <class U, class TensorType>
+struct TensorOutStreamer {
+  typedef TensorOutStreamer<U, TensorType> self;
+  const char* label;
+  const char* name;
+  const TensorType& tensor;
+  size_t limit;
+  DataLayout layout;
+  TensorOutStreamer(const char* label, const char* name,
+                    const TensorType& _tensor, DataLayout layout);
+  self& setLimit(size_t _limit);
+  const U* begin() const;
+  const U* end() const;
+  int checksum() const;
+};
+
+template <int version = 0>
+class TensorDumpConfig {
+ public:
+  static DataLayout getDefaultDataLayout() { return DataLayout::kAnyLayout; }
+
+ protected:
+  std::vector<std::string> split(const std::string& s, char delimiter) {
+    std::vector<std::string> tokens;
+    std::string token;
+    std::istringstream tokenStream(s);
+    while (std::getline(tokenStream, token, delimiter)) {
+      tokens.push_back(token);
+    }
+    return tokens;
+  }
+
+  template <class F>
+  void env_exe(const char* name, F f) {
+    auto* env = std::getenv(name);
+    if (env) {
+      f(env);
+    }
+  }
+
+  size_t limit_1;
+  size_t limit_4;
+  std::string filename;
+
+  struct OperatorDetails {
+    const std::string name;
+    DataLayout layout;
+    OperatorDetails(const std::string& _name,
+                    DataLayout _layout = getDefaultDataLayout())
+        : name{_name}, layout{_layout} {}
+    const std::string getLayoutString() { return DataLayoutToString(layout); }
+  };
+
+  // std::vector<std::string> ops;
+  std::vector<OperatorDetails> ops;
+
+  bool synchronized;
+  TensorDumpConfig()
+      : limit_1(128),
+        limit_4(128),
+        filename("/dev/stdout"),
+        synchronized(false) {
+    env_exe("TENSOR_DUMP_OPERATORS", [this](const char* value) {
+      auto tmp_ops = split(value, ',');
+      auto it = std::back_inserter(ops);
+      for (auto& _op : tmp_ops) {
+        auto key_value = split(_op, '=');
+        auto size = key_value.size();
+        if (!size || size > 2) {
+          throw std::runtime_error("incorrect key=value");
+        }
+        if (size == 1) {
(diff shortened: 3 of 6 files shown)


lidanqing-vv:
@jczaja  Please review.

lidanqing-vv:
Hi @pawelpiotrowicz  Please give MORE explanation of background, what is it for, the results and where to find the values.
If it is a tool, please give Readme, otherwise users don't know

paddle/fluid/framework/tensor.h
```diff
@@ -80,6 +83,16 @@ class Tensor {
...
 
+  template <class T>
+  bool hasType() const {
+    return type_ == ::paddle::framework::DataTypeTrait<T>::DataType();
+  }
+
+  void* raw_data() const;
+
+  /*! Serialize tensor to file with label name */
+  int dump(const char* label, const char* name = nullptr);
```
arogowie-intel, line 94: This doc should be expanded. I don't know which one of the arguments relates to the file name and which to the tensor. There should also be explanation regarding this default value and default behavior.

paddle/fluid/framework/tensor.h
```diff
@@ -13,16 +13,19 @@ See the License for the specific language governing permissions and
...
+#include <string>
+#include <tuple>
 #include <typeindex>
 #include <utility>
 #include <vector>
 #include "paddle/fluid/framework/data_layout.h"
+#include "paddle/fluid/framework/data_type.h"
 #include "paddle/fluid/framework/ddim.h"
 #include "paddle/fluid/framework/framework.pb.h"
+#include "paddle/fluid/framework/tensor_extension.h"
```
arogowie-intel, line 28: This header is not used here. Please move it to source file (`tensor.cc`)

paddle/fluid/framework/tensor.h
```diff
@@ -13,16 +13,19 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 #pragma once
-
 #include <cstdint>
 #include <cstring>
 #include <memory>
+#include <string>
+#include <tuple>
```
arogowie-intel, line 20: Those two headers are unused. Please remove.

paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
```diff
@@ -101,6 +101,14 @@ void eltwise_forward(const framework::ExecutionContext &ctx,
 
   y->set_layout(DataLayout::kMKLDNN);
   y->set_format(GetMKLDNNFormat(*dst_memory_p));
+
+  if (algorithm == mkldnn::algorithm::eltwise_tanh) {
```
arogowie-intel, line 105: Maybe we should introduce some map of `mkldnn::algorithm`s to its string name representation? This would let to get rid of all this ladder of `if ... else if`'s .

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+
+#include <cstdint>
+#include <cstring>
+#include <dnnl.hpp>
+#include "boost/crc.hpp"
+
+namespace paddle {
+namespace framework {
+
+// ########################
```
arogowie-intel, line 33: Please remove.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+constexpr const char* RED() { return "\033[0;31m"; }
+
+constexpr const char* GREEN() { return "\033[1;32m"; }
+
+constexpr const char* CYAN() { return "\033[0;36m"; }
+
+#define msg_color(c, x) \
+  std::cout << c << "[tensor-dump]: " << x << colors::RESET() << std::endl;
+#define msg_green(x) msg_color(colors::GREEN(), x)
+}  // namespace colors
```
arogowie-intel, line 48: Please split this file into header and source files. Move this implementation detail namespace into the source file.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+
+template <>
+struct ConvPaddleLayout2Mkldnn<DataLayout::kNCHW> {
+  static const dnnl::memory::format_tag value = dnnl::memory::format_tag::nchw;
+};
+
+template <>
+struct ConvPaddleLayout2Mkldnn<DataLayout::kNHWC> {
+  static const dnnl::memory::format_tag value = dnnl::memory::format_tag::nhwc;
+};
```
arogowie-intel, line 61: Those structures seem to be unused ( I can't find any of ). Please verify and if so remove.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+
+template <class L>
```
arogowie-intel, line 63: I guess this doesn't have to be a template function. Looks like the parameter `layout` could safely be of `DataLayout` type.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+dnnl::memory::format_tag tensor_layout2_mkldnn(L layout) {
+  using tag = dnnl::memory::format_tag;
+
+  if (layout == DataLayout::kNHWC) return tag::nhwc;
+
+  return tag::nchw;
```
arogowie-intel, line 69: Do we always want to return this data type? What if `layout` is different? Just looks suspicious to me.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+struct type2mkldnntype<uint8_t> {
+  static const dnnl::memory::data_type value = dnnl::memory::data_type::u8;
+};
+
+template <>
+struct type2mkldnntype<int8_t> {
+  static const dnnl::memory::data_type value = dnnl::memory::data_type::s8;
+};
+
+template <class T, class TensorType>
```
arogowie-intel, line 90: If the `TensorType` template parameter should derive from specific class please enforce it using [`std::enable_if`](https://en.cppreference.com/w/cpp/types/enable_if) and [`std::is_base_of`](https://en.cppreference.com/w/cpp/types/is_base_of)

paddle/fluid/framework/tensor.cc
```diff
@@ -120,5 +120,25 @@ void Tensor::ResetHolderWithType(std::shared_ptr<memory::Allocation> holder,
   type_ = type;
 }
 
+int Tensor::dump(const char* label, const char* name) {
```
arogowie-intel, line 123: The return type should be `void`. Instead of using old C-style of handling errors we should leverage C++ exceptions mechanism.

paddle/fluid/framework/tensor.cc
```diff
@@ -120,5 +120,25 @@ void Tensor::ResetHolderWithType(std::shared_ptr<memory::Allocation> holder,
...
+  if (TensorDumpConfig<>::get().is_disabled()) {
+    return -1;
```
arogowie-intel, line 125:
```suggestion
    return;
```

paddle/fluid/framework/tensor.cc
```diff
@@ -120,5 +120,25 @@ void Tensor::ResetHolderWithType(std::shared_ptr<memory::Allocation> holder,
...
+  }
+
+  auto it = TensorDumpConfig<>::get().fetchOperator(label);
+
+  if (it == TensorDumpConfig<>::get().fetchOperatorEnd()) {
+    return -2;
```
arogowie-intel, line 131: Better to create ie. `empty_operator_list` error class which derives from general Paddle error class, if there is any and throw it here. However I'm not sure whether this situation is an error.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+  DataLayout layout;
+  TensorOutStreamer(const char* label, const char* name,
+                    const TensorType& _tensor, DataLayout layout);
+  self& setLimit(size_t _limit);
+  const U* begin() const;
+  const U* end() const;
+  int checksum() const;
+};
+
+template <int version = 0>
```
arogowie-intel, line 145: This template is unnecessary.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+class TensorDumpConfig {
```
arogowie-intel, line 146: Please split this class into declarations and definition in separate source file. Also leave public only those methods and members which are used as a public API {`get`, `fetchOperator`, `fetchOperatorEnd`, `is_disabled`}. Please also use uniform function name style. I don't know yet which convention is used in Paddle whether the camel case or Hungarian notation.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+    std::string token;
+    std::istringstream tokenStream(s);
+    while (std::getline(tokenStream, token, delimiter)) {
+      tokens.push_back(token);
+    }
+    return tokens;
+  }
+
+  template <class F>
+  void env_exe(const char* name, F f) {
```
arogowie-intel, line 162: This function should be changed in one of the following two ways:

1.  Inline it everywhere changing like 
```
if (auto* env = std::getenv(name))
{
  // do lambda stuff here safely using env value.
}
```

2. Rename it like `call_on_env_var` and update its documentation to make clear its purpose, parameters and functionality.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+  int checksum() const;
+};
+
+template <int version = 0>
+class TensorDumpConfig {
+ public:
+  static DataLayout getDefaultDataLayout() { return DataLayout::kAnyLayout; }
+
+ protected:
+  std::vector<std::string> split(const std::string& s, char delimiter) {
```
arogowie-intel, line 151: This function should be private. It's too specific for this class. Please move its impl to separate source file.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+#include <tuple>
+#include <typeindex>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include <cstdint>
+#include <cstring>
+#include <dnnl.hpp>
+#include "boost/crc.hpp"
```
arogowie-intel, line 28: While separating declarations from definitions please review these headers and leave here only those necessary and move rest to the source file.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+            [this](const char* value) { synchronized = true; });
+  }
+
+ public:
+  bool is_disabled() const { return !ops.size(); }
+  bool hasOperator(const char* name) {
+    return std::find_if(ops.begin(), ops.end(),
+                        [name](const OperatorDetails& item) {
+                          return item.name == name;
+                        }) != ops.end();
```
arogowie-intel, line 240:
```suggestion
    return fetchOperator(name) != ops.end();
```

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+        ops.begin(), ops.end(),
+        [name](const OperatorDetails& item) { return item.name == name; });
+  }
+
+  auto fetchOperatorEnd() -> decltype(ops.end()) { return ops.end(); }
+
+  DataLayout getLayoutForOperator(const char* name) {
+    auto it = std::find_if(
+        ops.begin(), ops.end(),
+        [name](const OperatorDetails& item) { return item.name == name; });
```
arogowie-intel, line 254:
```suggestion
    auto it = fetchOperator(name);
```

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+    }
+    return getDefaultDataLayout();
+  }
+
+  bool is_synchronized() const { return synchronized; }
+  const std::string& getFilename() { return filename; }
+  std::ofstream& getOutputStream() {
+    static std::unique_ptr<std::ofstream> ptr(
+        new std::ofstream(filename.c_str()));
+    return *ptr;
```
arogowie-intel, line 267: IMHO this stream shouldn't be static. It shouldn't be shared across `TensorDumpConfig` instances. Let's make a class member for stream and return reference to it here.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+  }
+  size_t getLimit_1() { return limit_1; }
+  size_t getLimit_4() { return limit_4; }
+  size_t getLimitViaSize(int size) { return (size == 1) ? limit_1 : limit_4; }
```
arogowie-intel, line 271: Please add documentation to those functions. Please explain what are those limits used for? What is their meaning?

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+    return seq++;
+  }
+
+  static std::mutex& getMutex() {
+    static std::mutex mx;
+    return mx;
+  }
+};
+
+template <int version>
```
arogowie-intel, line 287: Unnecessary template.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+std::ostream& operator<<(
+    std::ostream& out,
+    const typename TensorDumpConfig<version>::OperatorDetails& c) {
+  out << "OperatorDetails{ " << c.name << "," << c.getLayoutString() << " }"
+      << std::endl;
+  return out;
+}
+
+template <class I, class F>
+void for_each_no_more(I b, I e, std::size_t count, F f) {
```
arogowie-intel, line 297: This function is implementation detail. Please move it to the separate source file and make it either static or move to unnamed namespace to prevent polluting global symbol namespace.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+    reorder_via_mkldnn<U>(vout.data(), &tensor, ts.layout);
+
+    for_each_no_more(vout.begin(), vout.end(), conf.getLimitViaSize(sizeof(U)),
+                     [&out, &br](U unit) {
+                       if (type_desc<U>::break_line == br++) {
+                         br = 0;
+                         out << std::endl;
+                       }
+                       type_desc<U>::format(out, unit) << std::dec;
+                     });
```
arogowie-intel, line 411: This lambda function is duplicated in the `else` branch below. Let's make a variable for it and reuse it.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+                         const TensorOutStreamer<U, TensorType>& ts) {
+  auto& tensor = ts.tensor;
+  auto& dims = tensor.dims();
+  auto& conf = TensorDumpConfig<>::get();
+
+  out << std::setw(8) << std::setfill(' ') << (TensorDumpConfig<>::NextRecord())
+      << ") type=[" << type_desc<U>::name() << "]  => label=" << ts.label
+      << " name=" << ((ts.name) ? ts.name : "undef") << " crc32=" << std::hex
+      << ts.checksum() << std::dec << " elem=" << tensor.numel()
+      << " dims=" << dims.size() << "=>";
```
arogowie-intel, line 388: Let's make it more readable:
```suggestion
  out << std::setw(8) << std::setfill(' ') 
      << (TensorDumpConfig<>::NextRecord())
      << ") type=[" << type_desc<U>::name() 
      << "]  => label=" << ts.label
      << " name=" << ((ts.name) ? ts.name : "undef") 
      << " crc32=" << std::hex << ts.checksum() << std::dec 
      << " elem=" << tensor.numel()
      << " dims=" << dims.size() << "=>";
```

paddle/fluid/framework/tensor_extension.h
arogowie-intel, line 297: Please also document that when passing `count = 0` this function call `F` object on each element in the range b -> e.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+template <class U, class TensorType>
+TensorOutStreamer<U, TensorType>& TensorOutStreamer<U, TensorType>::setLimit(
+    size_t _limit) {
+  limit = _limit;
+  return *this;
+}
+
+template <class U, class TensorType>
+const U* TensorOutStreamer<U, TensorType>::begin() const {
+  // return tensor.data<U>();
```
arogowie-intel, line 325: Please remove commented code.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+  return begin() + tensor.memory_size() / sizeof(U);
+}
+
+template <class U, class TensorType>
+int TensorOutStreamer<U, TensorType>::checksum() const {
+  boost::crc_32_type result;
+  result.process_bytes(reinterpret_cast<const unsigned char*>(begin()),
+                       tensor.memory_size());
+  return result.checksum();
+}
```
arogowie-intel, line 340: Impl detail, please move to separate source file.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+
+template <>
+struct type_desc<float> {
+  enum { pad = 8, break_line = 16 };
+  static const char* name() { return "float"; }
+  static std::ostream& format(std::ostream& out, float v) {
+    out << std::setw(pad) << std::setfill(' ') << v << " ";
+    return out;
+  }
+};
```
arogowie-intel, line 375: Same here: Impl detail, please move to separate source file.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+
+template <class U, class TensorType>
+std::ostream& operator<<(std::ostream& out,
```
arogowie-intel, line 378: Impl detail, please move to separate source file.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+struct TupleExtractor<std::tuple<First, Tail...>> {
+  typedef First head;
+  typedef std::tuple<Tail...> tail;
+};
+
+template <>
+struct TupleExtractor<std::tuple<>> {};
+
+template <int v>
+struct Int2Type {};
```
arogowie-intel, line 441: Unused, please remove.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+struct TupleExtractor;
+
+template <class First, class... Tail>
+struct TupleExtractor<std::tuple<First, Tail...>> {
+  typedef First head;
+  typedef std::tuple<Tail...> tail;
+};
+
+template <>
+struct TupleExtractor<std::tuple<>> {};
```
arogowie-intel, line 438: Unused. Please remove.

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+template <class First, class... Tail>
+struct DumpComposit {
+  typedef std::tuple<First, Tail...> args_tuple;
+  enum { size = std::tuple_size<args_tuple>::value };
+  typedef typename std::tuple_element<size - 1, args_tuple>::type last;
+  static_assert(std::is_same<last, float>::value,
+                "Last one must be float type!");
+  static_assert(std::is_arithmetic<First>::value,
+                "Incorrect type - is not an arithmetic type!");
+  template <class TensorType>
```
arogowie-intel, line 451:
```suggestion

  template <class TensorType>
```

paddle/fluid/framework/tensor_extension.h
```diff
@@ -0,0 +1,496 @@
...
+  typedef std::tuple<Tail...> tail;
+};
+
+template <>
+struct TupleExtractor<std::tuple<>> {};
+
+template <int v>
+struct Int2Type {};
+template <class First, class... Tail>
+struct DumpComposit {
```
arogowie-intel, line 443: @lidanqing-intel Let's talk about this tomorrow.

arogowie-intel (commented):
Some of my thoughts:
* In general let's separate as much implementation details into separate source `*.cc` file. 
* Maybe more generic approach would be to create some kind of pass which goes through all of ops within the model graph and dumps ops of specific type's outputs? It could be enabled by some env variable.

lidanqing-vv:
Related overload operator, but not fully implemented #24068

PR #24277, opened 2020-04-29
[Refactoring] Unify op-dequant squashes

wozna:
This PR unifies `op-dequant` squash passes, that squash dequantize operator if the previous operator has `force_fp32_output` attribute. Operators with such attribute are: `matmul`, `fc`, `conv2d`.


wozna:
@Sand3r- and @wojtuss can I ask you for a review?

PR #24281, opened 2020-04-29
Enabled quantize all and skip missing in QAT

wojtuss:
With this patch in QAT:
* the `--quantized_ops` option was renamed into `--ops_to_quantize` and made optional; if the option is not used, an attempt to quantize all quantizable oprators is made,
* if quantization scale is missing for any quantized operator, the quantization of the op is skipped with a message in the log, without error,
* QAT INT8 test for Ernie does not use the `--ops_to_quantize` option, i.e. all quantizable ops are quantized,
* the algorithm is refactored,
* QAT doc is updated.

test=develop

### DIFF
--- a/python/paddle/fluid/contrib/slim/tests/QAT_mkldnn_int8_readme.md
+++ b/python/paddle/fluid/contrib/slim/tests/QAT_mkldnn_int8_readme.md
@@ -273,7 +273,7 @@ You can use the `qat2_int8_image_classification_comparison.py` script to reprodu
 * `--infer_data` - a path to the validation dataset.
 
 The following option is also accepted:
-* `--quantized_ops` - a comma-separated list of operator types to quantize. If the option is not used, an attempt to quantize all quantizable operators will be made, and in that case only quantizable operators which have quantization scales provided in the QAT model will be quantized. When deciding which operators to put on the list, the following have to be considered:
+* `--ops_to_quantize` - a comma-separated list of operator types to quantize. If the option is not used, an attempt to quantize all quantizable operators will be made, and in that case only quantizable operators which have quantization scales provided in the QAT model will be quantized. When deciding which operators to put on the list, the following have to be considered:
   * Only operators which support quantization will be taken into account.
   * All the quantizable operators from the list, which are present in the model, must have quantization scales provided in the model. Otherwise, quantization of the operator will be skipped with a message saying which variable is missing a quantization scale.
   * Sometimes it may be suboptimal to quantize all quantizable operators in the model (cf. *Notes* in the **Gathering scales** section above). To find the optimal configuration for this option, user can run benchmark a few times with different lists of quantized operators present in the model and compare the results. For Image Classification models mentioned above the list usually comprises of `conv2d` and `pool2d` operators.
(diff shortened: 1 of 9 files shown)


python/paddle/fluid/contrib/slim/tests/QAT_mkldnn_int8_readme.md
```diff
@@ -270,15 +270,17 @@ You can use the `qat2_int8_image_classification_comparison.py` script to reprodu
...
-  * All the quantizable operators from the list, which are present in the model, must have quantization scales provided in the model. Otherwise, the quantization procedure will fail with a message saying which variable is missing a quantization scale.
-  * Sometimes it may be suboptimal to quantize all quantizable operators in the model (cf. *Notes* in the **Gathering scales** section above). To find the optimal configuration for this option, user can run benchmark a few times with different lists of quantized operators present in the model and compare the results. For Image Classification models mentioned above the list comprises of `conv2d` and `pool2d` operators.
 * `--infer_data` - a path to the validation dataset.
 
+The following option is also accepted:
+* `--quantized_ops` - a comma-separated list of operator types to quantize. If the option is not used, an attempt to quantize all quantizable operators will be made, and in that case only quantizable operators which have quantization scales provided in the QAT model will be quantized. When deciding which operators to put on the list, the following have to be considered:
```
lidanqing-vv: here need to change too

wojtuss: True. Done.

python/paddle/fluid/contrib/slim/tests/save_qat_model.py
```diff
@@ -43,10 +43,11 @@ def parse_args():
         default='',
         help='Saved optimized and quantized INT8 model')
     parser.add_argument(
-        '--quantized_ops',
+        '--ops_to_quantize',
         type=str,
         default='',
-        help='A comma separated list of quantized operators.')
```
Sand3r-, line 49: Not exactly related to the PR itself, but it seems to me, that save_qat_model, qat2_int8_nlp_comparison.py, qat2_int8_image_classification_comparison.py all share similar argument options. Perhaps it would be worth to consider moving shared options to some common `argparse.ArgumentParser()` that would be reused acrossed these 3 files.

Sand3r- (approved):
@wojtuss Thank you so much for submitted PR. Well done, I really like the code changes you've made.

python/paddle/fluid/contrib/slim/tests/save_qat_model.py
wojtuss: It is a good idea! We will add it when refactoring the QAT code.

lidanqing-vv: Could you not move the parameters in save_qat_model.py? Because it will be copied by users in PaddleSlim, it need to run independently. Thanks

lidanqing-vv:
@juncaipeng  Please review this PR, we want to merge this PR in 2.0 too

juncaipeng (approved):
LGTM

PR #24286, opened 2020-04-30
Fix double_grad bug in statig-graph (#24190)

qingqing01:
* Rename internal gradient variables in multiple backward
* so that they have different names with previous backward
* For example:
*  y = x * x, grad = fluid.gradients(fluid.gradients(y, x) + y * y, x)
* In second-time backward, gradient variable names of partial
* forward network (y * y) may be have same names with first-time
* fluid.gradients(y, x).

test=develop


XiaoguangHu01 (approved):
LGTM

Xreki (commented):
LGTM

Xreki (approved):

PR #24326, opened 2020-05-07
Fix slice bug in while loop

chenwhql:
This PR fix a bug, when while_loop body contains tensor slice operation, the slice op won't be add into sub block, which will cause calculate result error.

The reason for this error is that when `x[i]` adds slice-related ops, add ops according to the block where the var is located. Since x, i, etc. are variables of the global block, all ops are not correctly added to the sub-block, but the op should be added into the block it is executed, so this PR modified the block added for slice related ops.

### error case
- except result is `15`, but print result is `5`
```
def static_func(x):
    z = fluid.layers.fill_constant([1], 'int32', 0)
    x = fluid.layers.assign(x)
    x_shape = fluid.layers.shape(x)
    i = fluid.layers.fill_constant([1], 'int32', 0)

    def for_loop_condition_0(z, i):
        return i + 1 <= x_shape[0]

    def for_loop_body_0(z, i):
        z = z + x[i]
        i += 1
        return z, i

    z, i = fluid.layers.while_loop(for_loop_condition_0, for_loop_body_0, [z, i])
    return z

def execute():
    startup_program = fluid.Program()
    main_program = fluid.Program()
    with fluid.program_guard(main_program, startup_program):
        # prepare input
        np_x = np.array([1, 2, 3, 4, 5], dtype='int32')
        
        # add feed var
        x = data_layer_not_check(name='x', shape=list(np_x.shape), dtype=str(np_x.dtype))

        # build net
        result = static_func(x)

        # prepare exe
        place = fluid.CPUPlace()
        exe = fluid.Executor(place)

        # run
        # exe.run(startup_program)
        out, = exe.run(main_program,
            feed={'x': np_x},
            fetch_list=[result])
        print(out[0])

if __name__== '__main__':
    execute()
```

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -647,6 +647,7 @@ def _getitem_impl_(var, item):
     slice_step = []
     use_strided_slice = False
     reverse_axis = []
+    target_block = default_main_program().current_block()
 
     def fill_constant(shape, value, force_cpu=False, out=None):
         var.block.append_op(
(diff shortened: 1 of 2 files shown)


zhhsplendid (approved):
LGTM

liym27 (approved):
LGTM

luotao1 (approved):

PR #24337, opened 2020-05-07
OPT ccache support CUDA

T8T9:
**Motivation:**
The building stage of ci is too slow because
 1. Paddle use `Find(CUDA)` to introduce CUDA currently, which doesn't support ccache. Because  old cmake doesn't support CUDA, it can only add ccache prefix to `gcc` and `g++` command. `Find(Cuda)` includes `FindCUDA` module underlying. It will generate a `.cu.o.Release.cmake` script for every `.cu` file，then call `cmake -E <cmake script>` to compile `.cu` files. cmake wouldn't add `ccache` prefix to cmake command. To confirm this, you can check `paddle/fluid/operators/math/CMakeFiles/fc.dir/build.make` and `paddle/fluid/operators/math/CMakeFiles/fc.dir/fc_generated_fc.cu.o.Release.cmake` in your build directory.

    [cmake 3.10](https://cmake.org/cmake/help/v3.10/module/FindCUDA.html) support CUDA, it can compile `.cu` files by calling `nvcc` command directly, just like `gcc` and `g++`. So, to ccache CUDA object files, we need to use cmake built-in way. 

2. cmake will use `-x cu` to compile `.cu` files using `nvcc` compiler explicitly. but ccache less than 3.7.9 can not recognize this option, this bug has been fixed in [ccache 3.7.9](https://ccache.dev/releasenotes.html#_ccache_3_7_9). 

    Paddle Ci use ccache 3.6 currently, so we need to upgrade ccache to 3.7.9 to ccache CUDA object files.

3. Paddle pass compiler options(-Wno-unused-function, -Werror etc.) to gcc/g++ by adding `-Xcompiler -Wno-unused-function -Xcompiler -Werror` flags to nvcc. We should notice there is a ' ' between flag  `-Xcompiler` and  `-Werror`, ccache might treat it as two separate flags. The problem is both `g++` and `nvcc` have a built-in option `-Werror`, and `-Werror` is a compiler option, so ccache will remove this option when preprocess source files. In this case, ccache will change`-Xcompiler -Werror` to `-Xcompiler`, and preprocessor can not recognize `-Xcompiler`, this will make preprocess fail, and then ccache.

    To fix this, we should use `-Xcompiler=-Werror` to tell ccache these two flags are  binded, and they should be processed together. 

**Changes:**
1. use `enable_language(CUDA)` which is the cmake built-in way to support ccache.
2. upgrade ccache to 3.7.9.
3. set ccache max_size to 20GB. ccache will consume about 7GB disk space to cache object files after supporting CUDA, but the default max_size is 5GB, this will cause auto clean frequently, and cache will be deleted circularly, which can increase cache miss significantly. So we need to increase max_size bigger than 7GB. However, 7GB is space consumed by a clean build, paddle ci will build many times for different PRs, so we should make max_size bigger than 7GB to cache more object files to guarantee a high cache hit. I think 20GB is enough.

**Expected:** 
Time consumption of building stage should be reduced to about 11 minutes from about 55 minutes.


luotao1:
1. 为了便于开发者更容易读懂这个PR和带来的影响，请提供更多的信息
> Paddle use Find(CUDA) to introduce CUDA currently, which doesn't support ccache. 

为什么`Find(CUDA)`不支持ccache，请拿出实验结果。

> CUDA support was built in cmake 3.0, which will add "ccache" prefix to nvcc command.

这句描述没太理解，因为Paddle内部的cmake是3.16版本，是支持`"ccache" prefix to nvcc command`

> Paddle Ci use ccache 3.6 currently, which doesn't support "-x cu" option when compiling ".cu" files. ccache 3.7.9 fixed this bug.

- 请给出不支持`"-x cu" option`的理由，ccache官方链接即可。
- `"-x cu`这个编译选项是起什么作用的，为什么要用这个选项。

> The way we pass compiler options(-Wno-unused-function, -Werror etc.) from nvcc to gcc/g++ might cause confusion to ccache because both g++ and nvcc have a built-in option -Werror.

`might cause confusion`可能引起冲突？需要有确切结论。以及原来和现在的方式各是什么

2. 请给出修改的一些论据

> use enable_language(CUDA) to support ccache

这个语法的好处是什么？

> set ccache max_size to 20GB

为什么设成20G？

3. 数据效果

> Time consumption of building stage should be reduced to about 11 minutes from about 55 minutes.

从编译55分钟到11分钟，11分钟的效果如何体现，从PR_CI_Coverage中看还是55分钟。

4. 此PR改动较多，建议拆成小PR来做
- 升级所有镜像的ccache到3.7.9，和设置ccache_size
- `enable_language(CUDA)`相关工作。

T8T9:
> 1. 为什么`Find(CUDA)`不支持ccache，请拿出实验结果。
Because  old cmake doesn't support CUDA, it can only add ccache prefix to `gcc` and `g++` command. `Find(Cuda)` includes `FindCUDA` module underlying. It will generate a `.cu.o.Release.cmake` script for every `.cu` file，then call `cmake -E <cmake script>` to compile `.cu` files. cmake wouldn't add `ccache` prefix to cmake command. To confirm this, you can check `paddle/fluid/operators/math/CMakeFiles/fc.dir/build.make` and `paddle/fluid/operators/math/CMakeFiles/fc.dir/fc_generated_fc.cu.o.Release.cmake` in your build directory.

[cmake 3.10](https://cmake.org/cmake/help/v3.10/module/FindCUDA.html) support CUDA, it can compile `.cu` files by calling `nvcc` command directly, just like `gcc` and `g++`. **So, to ccache CUDA object files, we need to use cmake built-in way.** 

> "CUDA support was built in cmake 3.0, which will add "ccache" prefix to nvcc command." 这句描述没太理解，因为Paddle内部的cmake是3.16版本，是支持`"ccache" prefix to nvcc command`
Yes, paddle use cmake 3.16(>3.10) in ci, you can just ignore this.

> 请给出不支持`"-x cu" option`的理由，ccache官方链接即可。
cmake will use `-x cu` to compile `.cu` files using `nvcc` compiler explicitly. but ccache less than 3.7.9 can not recognize this option, this bug has been fixed in [ccache 3.7.9](https://ccache.dev/releasenotes.html#_ccache_3_7_9)

> * `"-x cu`这个编译选项是起什么作用的，为什么要用这个选项。
cmake explicitly add `-x cu` option to nvcc to make sure nvcc treat source files as CUDA code, even though file extensions are `.cpp` or `.c`. You can check this in `paddle/fluid/operators/math/CMakeFiles/fc.dir/build.make`.

> `might cause confusion`可能引起冲突？需要有确切结论。以及原来和现在的方式各是什么
Paddle pass compiler options(-Wno-unused-function, -Werror etc.) to gcc/g++ by adding `-Xcompiler -Wno-unused-function -Xcompiler -Werror` flags to nvcc. Notice that there is a ' ' between flag  `-Xcompiler` and  `-Werror`, ccache might treat it as two separate flags in specific circumstance. The problem is both `g++` and `nvcc` have a built-in option `-Werror`, and `-Werror` is a compiler option, so ccache will remove this option when preprocessing source files. In this case, ccache will change`-Xcompiler -Werror` to `-Xcompiler`, but preprocessor can not recognize `-Xcompiler`, this will make preprocess fail, and then ccache.

To fix this, we should use `-Xcompiler=-Werror` to tell ccache these two flags are  binded, and they should be processed together. 

> use enable_language(CUDA) to support ccache 这个语法的好处是什么？
This is the built-in way how cmake support CUDA language. Only in this way, cmake can add `ccache` prefix to `nvcc` command.

> set ccache max_size to 20GB. 为什么设成20G？
ccache will consume about 7GB disk space to cache object files after supporting CUDA, but the default max_size is 5GB, this will cause auto-clean frequently, and old cached object files will be deleted circularly, this will increase cache miss significantly. So we need to increase max_size bigger than 7GB. However, 7GB is space consumed by a clean build, paddle ci will build many times on different PRs, so we should make max_size bigger than 7GB to cache more object files to guarantee a high cache hit. I think 20GB is enough.
 
> 从编译55分钟到11分钟，11分钟的效果如何体现，从PR_CI_Coverage中看还是55分钟。
**只有这个PR merge之后，从其他的PR的编译效果上可以体现出来**，有两个原因：
1. 执行这个PR的机器是第一次启用CUDA缓存，相当于首次编译，当然需要50多分钟。
2. 即使两次执行PR_CI_Coverage 都调度到同一台主机上，因为现在ci上的max_size 还是5GB，所以旧的缓存会被循环删除。效果跟没有cache 差不多。


> 此PR改动较多，请拆成小PR来做
这些改动都是同一个功能必须的，拆成多个PR的话，反而看不出来到底是哪一个地方有问题。一个PR的话，如果有问题，回滚的话也方便一些，所以我觉得还是放在一个PR里吧

luotao1:
> you can check paddle/fluid/operators/math/CMakeFiles/fc.dir/build.make and paddle/fluid/operators/math/CMakeFiles/fc.dir/fc_generated_fc.cu.o.Release.cmake in your build directory.

如何check，ccache过了和没过的结果应该各自是如何的？

> 只有这个PR merge之后，从其他的PR的编译效果上可以体现出来，有两个原因：

- 执行这个PR的机器是第一次启用CUDA缓存，相当于首次编译，当然需要50多分钟。
- 即使两次执行PR_CI_Coverage 都调度到同一台主机上，因为现在ci上的max_size 还是5GB，所以旧的缓存会被循环删除。效果跟没有cache 差不多。

如果是必须merge后从其他PR的编译效果上看出来，请问11分钟的数据是怎么得到的？

> so we should make max_size bigger than 7GB to cache more object files to guarantee a high cache hit. I think 20GB is enough.

- 必须超过7GB，那为什么不设10GB，一下子设20GB这么大呢？
- 是否有机制，能查出来当前编译目录有多少，并报警呢？后续如果10GB不够了，还需要手动设置？
- 目前设置ccache_max_size是在dockerfile镜像里，这样会导致如果增大max_size需要频繁改镜像，可以放在paddle_build.sh脚本里。

> 这些改动都是同一个功能必须的，拆成多个PR的话，反而看不出来到底是哪一个地方有问题。

- 由于该PR必须merge后才能体现效果，所以可以拆成多个PR，通过最后一个PR的merge来体现效果。
- 该PR内容较多，而且内容上是可以做拆分的。（PR1和PR2可以并行提，PR3依赖PR1， PR4依赖前面三者）
  - PR1: 升级ccache到3.7.9，所有镜像需要统一升级。
  - PR2: 去掉`Find(CUDA)`使用`enable_language(CUDA)`，虽然目的是为了支持ccache，但这个功能可以单独升级。因为如果开发者裸机不装ccache的话，`enable_language(CUDA)`也要能跑。
  - PR3: CCACHE_MAXSIZE=20G不建议一下子设这么大，会导致开发者本地磁盘空间小时，无法编译，建议710G，并有相应的报警或自动扩缩机制。
  - PR4: 再解决`might cause confusion可能引起冲突`的问题。

> 一个PR的话，如果有问题，回滚的话也方便一些

该PR对nvcc还有编译选项都做了一定修改，内容较多，如果有问题，而且其他PR在上面做了其他修改的话，无法自动revert，需要手动revert，比较麻烦。
image

cmake/init.cmake
```diff
@@ -0,0 +1,26 @@
+# default: "-g"
```
luotao1, line 1: 这个init.cmake文件是什么用？没有看到在其他地方用过。

cmake/configure.cmake
```diff
@@ -76,7 +76,7 @@ if(WITH_GPU)
 
     FIND_PACKAGE(CUDA REQUIRED)
 
-    if(${CUDA_VERSION_MAJOR} VERSION_LESS 7)
+    if(${CUDA_VERSION} VERSION_LESS 7)
```
luotao1, line 79: 为什么把`CUDA_VERSION_MAJOR`改成`CUDA_VERSION`？我们有对应的`CUDA_VERSION_MINOR`。

cmake/configure.cmake
```diff
@@ -89,19 +89,19 @@ if(WITH_GPU)
     else()
         message(STATUS "Cannot find CUPTI, GPU Profiling is incorrect.")
     endif()
-    set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS} "-Xcompiler ${SIMD_FLAG}")
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Xcompiler=\"${SIMD_FLAG}\"")
```
luotao1, line 92: 为什么要改名`CUDA_NVCC_FLAGS`? 如果要改，可以在另一个PR中统一改。一个PR尽量保持少的修改。

cmake/cuda.cmake
```diff
@@ -142,19 +142,19 @@ function(select_nvcc_arch_flags out_variable)
...
-      list(APPEND nvcc_flags -gencode arch=compute_${CMAKE_MATCH_2},code=sm_${CMAKE_MATCH_1})
-      list(APPEND nvcc_archs_readable sm_${CMAKE_MATCH_1})
+      string(APPEND nvcc_flags " -gencode arch=compute_${CMAKE_MATCH_2},code=sm_${CMAKE_MATCH_1}")
+      string(APPEND nvcc_archs_readable " sm_${CMAKE_MATCH_1}")
     else()
       # User didn't explicitly specify PTX for the concrete BIN, we assume PTX=BIN
-      list(APPEND nvcc_flags -gencode arch=compute_${arch},code=sm_${arch})
-      list(APPEND nvcc_archs_readable sm_${arch})
+      string(APPEND nvcc_flags " -gencode arch=compute_${arch},code=sm_${arch}")
+      string(APPEND nvcc_archs_readable " sm_${arch}")
```
luotao1, line 150: 这里为什么要从list改成string，list的功能有什么问题么？

cmake/cuda.cmake
```diff
@@ -196,37 +197,24 @@ endif(NOT WITH_DSO)
...
 
 # Set C++11 support
 set(CUDA_PROPAGATE_HOST_FLAGS OFF)
-
 # Release/Debug flags set by cmake. Such as -O3 -g -DNDEBUG etc.
 # So, don't set these flags here.
-if (NOT WIN32) # windows msvc2015 support c++11 natively. 
-# -std=c++11 -fPIC not recoginize by msvc, -Xcompiler will be added by cmake.
-list(APPEND CUDA_NVCC_FLAGS "-std=c++11")
-list(APPEND CUDA_NVCC_FLAGS "-Xcompiler -fPIC")
```
luotao1, line 210: 这里去掉`-Xcompiler -fPIC`是什么考虑？

cmake/cuda.cmake
```diff
@@ -196,37 +197,24 @@ endif(NOT WITH_DSO)
...
-  elseif(CMAKE_BUILD_TYPE  STREQUAL "Release")
-      list(APPEND CUDA_NVCC_FLAGS  ${CMAKE_CXX_FLAGS_RELEASE})
-  elseif(CMAKE_BUILD_TYPE  STREQUAL "RelWithDebInfo")
-      list(APPEND CUDA_NVCC_FLAGS  ${CMAKE_CXX_FLAGS_RELWITHDEBINFO})
-  elseif(CMAKE_BUILD_TYPE  STREQUAL "MinSizeRel")
-      # nvcc 9 does not support -Os. Use Release flags instead
-      list(APPEND CUDA_NVCC_FLAGS  ${CMAKE_CXX_FLAGS_RELEASE})
-  endif()
-else(NOT WIN32)
+if (WIN32)
```
luotao1, line 217: 这段逻辑从原来的是(NOT win32)改成了win32？而且去掉了`if(CMAKE_BUILD_TYPE  STREQUAL "Debug")`的逻辑，为什么？

cmake/flags.cmake
```diff
@@ -137,59 +137,59 @@ endif()
...
+        -Wno-unused-parameter
+        -Wno-unused-function
+        -Wno-error=literal-suffix
+        -Wno-error=unused-local-typedefs
+        -Wno-error=unused-function  # Warnings in Numpy Header.
+        -Wno-error=array-bounds # Warnings in Eigen::array
+    )
+    if (NOT WITH_NV_JETSON)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -m64")
     endif()
```
luotao1, line 192: 140-190行的修改作用是什么？

cmake/generic.cmake
```diff
@@ -413,9 +413,9 @@ function(nv_library TARGET_NAME)
     cmake_parse_arguments(nv_library "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
     if(nv_library_SRCS)
       if (nv_library_SHARED OR nv_library_shared) # build *.so
-        cuda_add_library(${TARGET_NAME} SHARED ${nv_library_SRCS})
+        add_library(${TARGET_NAME} SHARED ${nv_library_SRCS})
```
luotao1, line 416: `cuda_add_library`换成`add_library`的作用是什么？

cmake/configure.cmake
T8T9: CUDA_VERSION_MAJOR 是 find(CUDA) 引入的变量，因为有多个地方使用了 find(CUDA) 引入的其他变量，所以这次没有完全删除 find(CUDA) ，目前 find(CUDA)  的作用就是导入变量，与编译已经没有关系了。
先修改这一个地方，看看有没有副作用，后面再完全删除对 find(CUDA) 的依赖。

cmake/configure.cmake
T8T9: CUDA_NVCC_FLAGS 是 find(CUDA)中的变量。后面要完全摆脱对find(CUDA)的依赖。
CMAKE_CUDA_FLAGS 是 cmake 中的内置变量。使用cmake内置的cuda支持功能，必须这么改。
不能单独提PR。

cmake/cuda.cmake
T8T9: 跟 CMAKE_CUDA_FLAGS 有关，需要一个string 进行 append 操作，list 还需要遍历。

cmake/cuda.cmake
T8T9: 跟 cmake/flags.cmake 中的设置重复了，重复的参数，会导致 ccache 解析出错。导致无法缓存

cmake/cuda.cmake
T8T9: 这就是加入init.cmake 的目的了。不应该在这里显式地追加 CMAKE_CXX_FLAGS_DEBUG 之类的参数，cmake会自动追加，否则的话就会出现重复参数的问题。
所以 if(NOT WIN32)不需要了，把原来的else(NOT WIN32) 改成 if (WIN32) 即可.

cmake/flags.cmake
T8T9: 这段主要是格式化

cmake/generic.cmake
T8T9: add_library 和 add_executable 就是cmake内置的支持cuda的方式。
cuda_add_library是find(CUDA)的方式。不支持ccache

cmake/init.cmake
T8T9: 全局 flag 在这里设置，由 cmake 自动追加，防止出现重复参数。

T8T9:
> 如何check，ccache过了和没过的结果应该各自是如何的？

想check的话，把pr代码co到本地，看下原来的和修改之后的，这两个文件中的命令就可以了。

> 如果是必须merge后从其他PR的编译效果上看出来，请问11分钟的数据是怎么得到的？

本地模拟ci环境测试

> * 必须超过7GB，那为什么不设10GB，一下子设20GB这么大呢？

ccache 清理的阈值是 max_size * limit_multiple(0.8), 10G的话，实际上8G就开始清理了。

> * 是否有机制，能查出来当前编译目录有多少，并报警呢？后续如果10GB不够了，还需要手动设置？

没有。10G不够了，就清理旧缓存。

> * 目前设置ccache_max_size是在dockerfile镜像里，这样会导致如果增大max_size需要频繁改镜像，可以放在paddle_build.sh脚本里。

可以

> * 该PR内容较多，而且内容上是可以做拆分的。（PR1和PR2可以并行提，PR3依赖PR1， PR4依赖前面三者）

可以。拆分成：
PR1: https://github.com/PaddlePaddle/Paddle/pull/24390
PR2: https://github.com/PaddlePaddle/Paddle/pull/24395
PR3: TBD

luotao1:
> 想check的话，把pr代码co到本地，看下原来的和修改之后的，这两个文件中的命令就可以了。

能否在PR里提出这两个文件的命令

> 本地模拟ci环境测试

本地如何模拟，有具体的命令和实验流程么？也就是如何复现出11分钟的效果

> ccache 清理的阈值是 max_size * limit_multiple(0.8), 10G的话，实际上8G就开始清理了。

可以加到代码注释里

此外，7GB的实验数据是如何得到的？

CMakeLists.txt
```diff
@@ -20,6 +20,19 @@ set(PADDLE_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
...
 project(paddle CXX C)
+
+include(init)
+
+# enable language CUDA
+find_package(CUDA QUIET)
+option(WITH_GPU         "Compile PaddlePaddle with NVIDIA GPU"          ${CUDA_FOUND})
+if(WITH_GPU AND NOT APPLE)
+    enable_language(CUDA)
+    set(CUDA_VERSION ${CMAKE_CUDA_COMPILER_VERSION})
```
luotao1, line 31: 可以直接用`CMAKE_CUDA_COMPILER_VERSION`，不需要转成`CUDA_VERSION`，便于理解，这里加注释说下这是类似`CUDA_VERSION_MAJOR`的自带变量。

cmake/cuda.cmake
```diff
@@ -167,27 +167,28 @@ if (${CUDA_VERSION} LESS 7.0)
...
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -D__STRICT_ANSI__")
 elseif (${CUDA_VERSION} LESS 11.0) # CUDA 10.x
   set(paddle_known_gpu_archs ${paddle_known_gpu_archs10})
-  list(APPEND CUDA_NVCC_FLAGS "-D_MWAITXINTRIN_H_INCLUDED")
-  list(APPEND CUDA_NVCC_FLAGS "-D__STRICT_ANSI__")
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -D_MWAITXINTRIN_H_INCLUDED")
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -D__STRICT_ANSI__")
 endif()
+
+message(STATUS "PADDLE_CUDA_BINVER=${CUDA_VERSION_MAJOR}${CUDA_VERSION_MINOR}")
```
luotao1, line 189: 189行为什么还要打印呢？

cmake/flags.cmake
luotao1: 格式化是指什么？是调整下代码格式，更加优雅么？

cmake/generic.cmake
luotao1: 请加代码注释

cmake/generic.cmake
```diff
@@ -464,7 +464,7 @@ function(nv_test TARGET_NAME)
     set(oneValueArgs "")
     set(multiValueArgs SRCS DEPS)
     cmake_parse_arguments(nv_test "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
-    cuda_add_executable(${TARGET_NAME} ${nv_test_SRCS})
+    add_executable(${TARGET_NAME} ${nv_test_SRCS})
```
luotao1, line 467: 请加代码注释

cmake/configure.cmake
luotao1: 请加一个TODO

CMakeLists.txt
T8T9: done

cmake/cuda.cmake
T8T9: 删了

cmake/generic.cmake
T8T9: 加什么注释？哪里看不懂呢？

cmake/configure.cmake
T8T9: done

cmake/flags.cmake
T8T9: 是的，原来if没缩进

cmake/generic.cmake
T8T9: done

cmake/generic.cmake
T8T9: done

T8T9:
> PR_CI_GPU_Py2_GCC8 是GPU编译，目前30分钟。由于TeamCity可以指定机器跑，因此可以指定一个机器跑两遍本PR，看第二遍的编译速度是否有提升

先把 https://github.com/PaddlePaddle/Paddle/pull/24390 这个PR merge 了吧

T8T9:
> > 想check的话，把pr代码co到本地，看下原来的和修改之后的，这两个文件中的命令就可以了。
> 能否在PR里贴出这两个文件的命令

旧命令:
```bash
cd /workspace/Paddle/build/paddle/fluid/operators/math/CMakeFiles/fc.dir && /home/cmake-3.16.0-Linux-x86_64/bin/cmake -D verbose:BOOL=$(VERBOSE) -D build_configuration:STRING=Release -D generated_file:STRING=/workspace/Paddle/build/paddle/fluid/operators/math/CMakeFiles/fc.dir//./fc_generated_fc.cu.o -D generated_cubin_file:STRING=/workspace/Paddle/build/paddle/fluid/operators/math/CMakeFiles/fc.dir//./fc_generated_fc.cu.o.cubin.txt -P /workspace/Paddle/build/paddle/fluid/operators/math/CMakeFiles/fc.dir//fc_generated_fc.cu.o.Release.cmake
```

新命令:
```bash
cd /workspace/Paddle/build/paddle/fluid/operators/math && /usr/local/bin/ccache /usr/local/cuda/bin/nvcc  $(CUDA_DEFINES) $(CUDA_INCLUDES) $(CUDA_FLAGS) -x cu -c /workspace/Paddle/paddle/fluid/operators/math/fc.cu -o CMakeFiles/fc.dir/fc.cu.o
```

> > 本地模拟ci环境测试
> 本地如何模拟，有具体的命令和实验流程么？也就是如何复现出11分钟的效果

执行的命令是：
```bash
cmake .. -DCMAKE_BUILD_TYPE=Release -DWITH_COVERAGE=ON -DWITH_DSO=ON -DWITH_GPU=ON -DWITH_AMD_GPU=OFF -DWITH_DISTRIBUTE=ON -DWITH_MKL=ON -DWITH_AVX=ON -DNOAVX_CORE_FILE= -DWITH_GOLANG=OFF -DCUDA_ARCH_NAME=Auto -DCUDA_ARCH_BIN= -DWITH_PYTHON=ON -DCUDNN_ROOT=/usr/ -DWITH_TESTING=ON -DCMAKE_MODULE_PATH=/opt/rocm/hip/cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DWITH_CONTRIB=ON -DWITH_INFERENCE_API_TEST=ON -DINFERENCE_DEMO_INSTALL_DIR=/root/.cache/inference_demo -DPY_VERSION=2.7 -DCMAKE_INSTALL_PREFIX=/paddle/build -DWITH_GRPC=ON -DWITH_LITE=OFF && make install -j 16
```

> > ccache 清理的阈值是 max_size * limit_multiple(0.8), 10G的话，实际上8G就开始清理了。
> 此外，7GB的实验数据是如何得到的？

清空缓存-删除build目录-执行build-查看缓存大小。

cmake/flags.cmake
luotao1: 调整代码格式，可以单独一个PR（此PR只调整格式，优化代码）做，不要和新功能混在一起。这里还调整了178-192行的位置

PR #24259, opened 2020-04-29
[dy2static] Support for static graph training with @declarative decorator

Aurelius84:
### 1. 背景
框架目前通过`@declarative`对动态图`forward`函数添加装饰，可以将网络转为静态图执行。但需要对`train`函数代码进行部分改写，放到`fluid.program_guard`语法下进行训练。

在动转静训练时，为了尽可能避免代码的修改，提升`@declarative`的易用性，将`@declarative`下的所有layers涉及的op转换为静态图模式执行。

image

### 2. 基本原理
 - 复用动态图的Trace执行流程，将被转为静态图的所有op，放到单独的子Program里
 - 解析子Program中Ops，组网并放到`run_program_op`进行trace，进行前向计算，并添加反向

如上图所示，`sub_layers`中的所有的ops计算都会放进`run_program_op`里。在执行sub_layers时，通过Trace `run_program_op`动态执行所有子Program前向ops。

在反向计算时，借助静态图的`fluid.gradients`接口，在子Program中添加反向grad_op。当执行`loss.backward()`时，会调用`run_program_op`的GradKernel，依次调用子Program中的`grad_op Kernel`计算grad。

**Example:**
```python
class Linear(fluid.dygraph.Layer):
    def __init__(self, input_dim=10, output_dim=5):
        super(Linear, self).__init__()
        self.fc = fluid.dygraph.Linear(input_dim, output_dim, act='relu')

    @declarative  # 添加装饰器
    def forward(self, x):
        pre = self.fc(x)
        loss = fluid.layers.mean(pre)
        return pre, loss

# 训练代码不需要做任何改动
def train(self, to_static=False):
    batch_num = 10
    with fluid.dygraph.guard(fluid.CPUPlace()):
        net = Linear()
        adam = fluid.optimizer.AdamOptimizer(learning_rate=0.1, parameter_list=net.parameters())
        
        for batch_id in range(batch_num):
            input = fluid.dygraph.to_variable(batch_data)
            pred, avg_loss = net(input)

            avg_loss.backward()
            adam.minimize(avg_loss)
            dygraph_net.clear_gradients()

    # 保存训练好的模型参数
    fluid.save_dygraph(net.state_dict(), "./test_dy2stat_save_load")

    infer_net = Linear()
    # 加载训练好的参数，进行预测
    model_dict, _ = fluid.load_dygraph("./test_dy2stat_save_load")
    infer_net.set_dict(model_dict)
    pred, _ = infer_net(x)
```

#### 新增
[...]

PR #24353, opened 2020-05-08
[dy2stat]Support save_inference_model in program_translator

Aurelius84:
### 1. 背景
在https://github.com/PaddlePaddle/Paddle/pull/24259 中支持将`@declarative`装饰的函数转为静态子图的方式运行，本质上复用的动态图Trace框架。但`save_inference_model`接口仅支持静态图下调用。

因此在此PR中，支持用户在最外层`forward`添加`@declarative`，得到完整的main_program，用于保存可预测部署的静态图模型。

**Example:**
```python
import numpy as np
import paddle.fluid as fluid
from paddle.fluid.dygraph import Linear
from paddle.fluid.dygraph import ProgramTranslator

class SimpleNet(fluid.dygraph.Layer):
        super(SimpleNet, self).__init__()
        self._linear = Linear(in_size, out_size)

    @declarative
    def forward(self, x):
        y = self._linear(x)
        z = self._linear(y)
        loss = fluid.layers.mean(z)
        return z, loss

with fluid.dygraph.guard(fluid.CPUPlace()):
    net = SimpleNet(8, 8)
    adam = fluid.optimizer.AdamOptimizer(learning_rate=0.1, parameter_list=net.parameters())
    x = fluid.dygraph.to_variable(np.random.random((4, 8)).astype('float32'))
    for i in range(10):
        loss, out = net(x)
        loss.backward()
        adam.minimize(loss)
        net.clear_gradients()
# Save inference model.
# Note that fetch=[0] means we set 'y' as the inference output.
prog_trans = ProgramTranslator()
prog_trans.save_inference_model("./dy2stat_infer_model", fetch=[0])
```

#### 新增
- 在`program_translator`暴露`save_inference_model`接口，支持用户在动转静训练后，可以保存静态图可预测部署的模型
- 在`test_mnist`和`test_save_inference_model`中，添加`train/save/load`完整流程的单测，验证正确性

#### 优化
- 统一`functionSpec`返回的参数类型为`OrderedDict`
- 同步用户在前端设置的program的random_seed到cache_program
- 在动转静训练中添加param检查和报错模块，明确告知用户避免在`@declarative`装饰的函数中定义带参数的layers
- 优化了相关单测代码，抽离可复用的变量声明到`SetUp`函数中
- 规范化了部分变量和参数的命名，如`start_up`改为`startup_program`

#### 移除
- 移除了`list_transformer`中的`__all__`，内部接口不应该暴露给用户
- 移除了`partial_program`中无用的函数`append_grad_suffix`


python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
```diff
@@ -522,6 +552,93 @@ def func(x):
         source_code = ast_to_source_code(root_wrapper.node)
         return source_code
 
+    def save_inference_model(self, dirname, feed=None, fetch=None):
+        """
+        Saves current model as the inference model. The saved
+        inference model can be loaded by C++ inference APIs.
```
chenwhql: 这里的文档，应该不止能被C++ API load吧，是否容易误解

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
```diff
@@ -522,6 +552,93 @@ def func(x):
...
+        Args:
+            dirname (str): the directory to save the inference model.
+            feed (list[int], optional): the input variable indices of the saved
+                inference model. If None, all input variables of the
+                ProgramTranslator would be the inputs of the saved inference
+                model. Default None.
+            fetch (list[int], optional): the output variable indices of the
```
chenwhql, line 565: 这里feed和fetch一定要用int索引吗？能否和`io.save_infernece_model`的输入形式统一？

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
Aurelius84: Thanks，也可以被python端load，已经参考框架的api，进行了完善。

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
```diff
@@ -218,6 +229,15 @@ def __getitem__(self, item):
             self._caches[item] = self._build_once(item)
         return self._caches[item]
 
+    def __len__(self):
+        return len(self._caches)
```
liym27: It seems that `len(ProgramCache)` is not called.

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
```diff
@@ -174,13 +178,20 @@ def from_func_spec(func_spec):
         dygaph_function = func_spec.dyfunc
         static_func = convert_function_with_cache(dygaph_function)
 
-        main_program, start_up = framework.Program(), framework.Program()
-        with framework.program_guard(main_program, start_up):
+        main_program, startup_program = framework.Program(), framework.Program()
+        # Note: The random seed should be synchronized into cached program
```
zhhsplendid, line 182: Good code!

python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, main_program, inputs, outputs, parameters=None):
         self.outputs = outputs
         self._params = parameters
         self._infer_program = main_program
+        check_params_init_outside(main_program, parameters)
```
zhhsplendid: Not very good ... you have background that ``parameters`` must be ``param_from_init``, but future software engineers may not know it ... then it disobeys OOP: caller just knows how to use interface but doesn't care about background and  implementation details. In ``__init__``, the input ``parameters`` can even be ``None`` which can cause problem.

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
Aurelius84: Thanks，I will remove these codes.

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
Aurelius84: 这里可能还得用list[int]，原因如下：
- 动态图里用户通过`to_variable(x_data)`创建VarBase，若未指定name，则会创建`generate_var_0`且每个batch都会重新创建（id递增）。转静态图program时，只会创建一次variable作为feed var，避免每次都要遍历program以更新var_name带来的开销
- 另外，这里接口和`TraceLayer.save_inference_model()`保持了一致

python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
Aurelius84: Thanks, I refine these code. Actually, `partial_program` only require two conditions about `parameters` and `main_program`:
 - parameter in `self._params` should be type `framework.ParamBase` which are created in dygraph.
- parameter from transformed main_program  shall be found in `self._params`. Because they share same data with `ParamBase` of original dygraph.

`parameters` can be None，which means there is no trainable parameter in `transformed main_program`. I add related checks for this case to make sure no `Parameter` in program.

So, the `check_params_init_outside ` means `check params in program is initialized and can be found in self._params`, I modified the function name.

python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
```diff
@@ -155,6 +164,36 @@ def _set_grad_type(self, params):
...
+                    'Type of self._params[{}] shoule be framework.ParamBase, but received {}.'.
+                    format(i, type(param)))
+            params_name_set.add(param.name)
+
+        for block in main_program.blocks:
+            for name, var in block.vars.items():
+                if isinstance(var, framework.Parameter):
+                    if name not in params_name_set:
+                        raise ValueError(
+                            "We don't support to define layer with parameters in the function "
```
zhhsplendid: Here you explain that "we don't do so" but not the reason "why we don't do so". Could you add the reason to users why they have to do it? Just like our chat in this morning:

"We don't support to define layer with parameters in the function decorated by `@declarative` because that will re-define parameters every time when you run the function. We found parameter(%s) created in the decorated function. Please define the parameter layer in in `__init__` function."

This is good for both SWEs and users because they know background why we check the parameter is in __init__ here.

Aurelius84: Thanks, the error message will be output as followed:
```
ValueError: 
	We don't support to define layer with parameters in the function decorated by `@declarative`.
	Because that will re-defined parameters every time when you run the function.
	But we found parameter(linear_0.w_0) was created in the decorated function.
	Please define the layer with parameters in `__init__` function.
```

zhhsplendid (approved):
LGTM

liym27 (approved):
LGTM

chenwhql (approved):
LGTM

lanxianghit (approved):

Boyan-Liu (approved):

PR #24357, opened 2020-05-08
Fix the error of recurrnet op in multithreading in eval process

gfwm2013:
Recurrent Op中的函数CreateStepScopes在多线程预测阶段会导致程序segmentation fault错误，目前先使用mutex确保recurrent Op的计算过程的正常运行，该问题将会在下个PR中修复。

### DIFF
--- a/paddle/fluid/operators/recurrent_op.cc
+++ b/paddle/fluid/operators/recurrent_op.cc
@@ -197,7 +197,6 @@ void RecurrentOp::RunImpl(const framework::Scope &scope,
   auto &dev_ctx = *pool.Get(place);
 
   VLOG(3) << "Static RNN input sequence length = " << seq_len;
-  StepScopes scopes = CreateStepScopes(dev_ctx, scope, seq_len);
   auto reverse = Attr<bool>(kReverse);
 
   framework::Executor executor(place);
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/recurrent_op.cc
```diff
@@ -208,6 +207,13 @@ void RecurrentOp::RunImpl(const framework::Scope &scope,
       *program, block->ID(), Attr<std::vector<std::string>>(
                                  kSkipEagerDeletionVars) /*skip_ref_cnt_vars*/);
 
+  static std::mutex mutex;
+  std::lock_guard<std::mutex> lock(mutex);
+  StepScopes scopes = CreateStepScopes(dev_ctx, scope, seq_len);
+  // TODO(gfwm2013) Function CreateStepScopes would make segmentation fault in
```
zhhsplendid, line 213: You just wrote the problem but not the cause. We have found the reason. So please write down the reason of the segmentation fault. Then if you are going to do other things and forget it or some other software engineer takes place of you to fix here, we can still know the reason and debug easily.

zhhsplendid (approved):
Discussed offline, I will approve this PR and we will add the reason of segmentation fault in the next PR.

PR #24390, opened 2020-05-11
upgrade ccache to 3.7.9 to support CUDA

T8T9:
upgrade ccache to 3.7.9 to support CUDA.

cmake will use `-x cu` to compile .cu files using nvcc compiler explicitly. but ccache less than 3.7.9 can not recognize this option, this bug has been fixed in [ccache 3.7.9](https://ccache.dev/releasenotes.html#_ccache_3_7_9).

Paddle Ci use ccache 3.6 currently, so we need to upgrade ccache to 3.7.9 to ccache CUDA object files.

related #24337

### DIFF
--- a/tools/manylinux1/Dockerfile.GCC8
+++ b/tools/manylinux1/Dockerfile.GCC8
@@ -75,7 +75,7 @@ RUN apt-get update && \
     python-matplotlib \
     automake locales clang-format swig  \
     liblapack-dev liblapacke-dev \
-    net-tools libtool ccache module-init-tools && \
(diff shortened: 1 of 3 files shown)


luotao1:
https://github.com/PaddlePaddle/Paddle/blob/develop/Dockerfile 是latest-dev镜像（开发镜像），请同步升级

luotao1:
@T8T9 为了让本地开发者不使用docker镜像的时候也不会出现
image
因此，代码里需要将相关最小cmake version设成3.10
```
cmake_minimum_required(VERSION 3.0)
```

PR #24395, opened 2020-05-11
support CUDA using cmake built-in way

T8T9:
**Motivation:**
The building stage of ci is too slow because
 1. Paddle use `Find(CUDA)` to introduce CUDA currently, which doesn't support ccache. Because  old cmake doesn't support CUDA, it can only add ccache prefix to `gcc` and `g++` command. `Find(Cuda)` includes `FindCUDA` module underlying. It will generate a `.cu.o.Release.cmake` script for every `.cu` file，then call `cmake -E <cmake script>` to compile `.cu` files. cmake wouldn't add `ccache` prefix to cmake command. To confirm this, you can check `paddle/fluid/operators/math/CMakeFiles/fc.dir/build.make` and `paddle/fluid/operators/math/CMakeFiles/fc.dir/fc_generated_fc.cu.o.Release.cmake` in your build directory.

    [cmake 3.10](https://cmake.org/cmake/help/v3.10/module/FindCUDA.html) support CUDA, it can compile `.cu` files by calling `nvcc` command directly, just like `gcc` and `g++`. So, to ccache CUDA object files, we need to use cmake built-in way. 

3. Paddle pass compiler options(-Wno-unused-function, -Werror etc.) to gcc/g++ by adding `-Xcompiler -Wno-unused-function -Xcompiler -Werror` flags to nvcc. We should notice there is a ' ' between flag  `-Xcompiler` and  `-Werror`, ccache might treat it as two separate flags. The problem is both `g++` and `nvcc` have a built-in option `-Werror`, and `-Werror` is a compiler option, so ccache will remove this option when preprocess source files. In this case, ccache will change`-Xcompiler -Werror` to `-Xcompiler`, and preprocessor can not recognize `-Xcompiler`, this will make preprocess fail, and then ccache.

    To fix this, we should use `-Xcompiler=-Werror` to tell ccache these two flags are  binded, and they should be processed together. 

related #24337


CMakeLists.txt
```diff
@@ -20,6 +20,19 @@ set(PADDLE_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 include(system)
 
 project(paddle CXX C)
+
+include(init)
+
+# enable language CUDA
+# TODO: replace find_package(CUDA) completely.
```
luotao1: TODO(taoshibo)

cmake/generic.cmake
```diff
@@ -412,10 +412,12 @@ function(nv_library TARGET_NAME)
     set(multiValueArgs SRCS DEPS)
     cmake_parse_arguments(nv_library "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
     if(nv_library_SRCS)
+      # Attention: cuda_add_library is deprecated, use add_library for CUDA please.
+      # https://cmake.org/cmake/help/v3.10/module/FindCUDA.html
```
luotao1: cuda_add_library is deprecated after cmake v3.10? 补充下从什么版本开始废弃的。此外补充下，cuda_add_library不能用于ccache

cmake/generic.cmake
```diff
@@ -464,7 +466,9 @@ function(nv_test TARGET_NAME)
     set(oneValueArgs "")
     set(multiValueArgs SRCS DEPS)
     cmake_parse_arguments(nv_test "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
-    cuda_add_executable(${TARGET_NAME} ${nv_test_SRCS})
+    # Attention: cuda_add_executable is deprecated, use add_executable for CUDA please.
+    # https://cmake.org/cmake/help/v3.10/module/FindCUDA.html
```
luotao1: 同上

cmake/flags.cmake
```diff
@@ -137,59 +137,59 @@ endif()
...
+        -Wno-unused-parameter
+        -Wno-unused-function
+        -Wno-error=literal-suffix
+        -Wno-error=unused-local-typedefs
+        -Wno-error=unused-function  # Warnings in Numpy Header.
+        -Wno-error=array-bounds # Warnings in Eigen::array
+    )
+    if (NOT WITH_NV_JETSON)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -m64")
     endif()
```
luotao1, line 175: 格式修改请换一个PR做，https://github.com/PaddlePaddle/Paddle/pull/24337/files#r423459781

luotao1:
PR_CI_GPU_Py2_GCC8 是GPU编译，目前30分钟，而且ccache量在5G内应该够。由于TeamCity可以指定机器跑，因此可以指定一个机器跑两遍本PR，看第二遍的编译速度是否有提升。

CMakeLists.txt
T8T9: done

cmake/generic.cmake
T8T9: done

cmake/generic.cmake
T8T9: done

cmake/flags.cmake
T8T9: 已复原

T8T9:
> PR_CI_GPU_Py2_GCC8 是GPU编译，目前30分钟，而且ccache量在5G内应该够。由于TeamCity可以指定机器跑，因此可以指定一个机器跑两遍本PR，看第二遍的编译速度是否有提升。

[优化前](http://10.87.145.41:8111/viewLog.html?buildId=342452&buildTypeId=YYTest_PrCiCpuPy2Test&tab=buildLog&state=&expand=all&_focus=6351#_state=)：06:10:43-06:43:38，大约33分钟。

[优化后](http://10.87.145.41:8111/viewLog.html?buildId=342679&buildTypeId=YYTest_PrCiCpuPy2Test&tab=buildLog&branch_YYTest=pull%2F24395&_focus=6293)： 09:12:35-09:19:20，大约7分钟。

Issue #24394, opened 2020-05-11
Ernie INT8 test crash


lidanqing-vv:
Still updating...
Ernie INT8 Qat test crashed on develop on 6271.
The failure range:
Crashed commit id: newest develop
Crashed commit id: 9eedf05d2f30671da5332f2f8b50e2061d5a0758
Crashed commit id: 1fc6cc502ac7303c879460b050cb24d6fa3a6023
Crashed commit id: 50330c6c6d323446744ac65613ea45130e7714b7
Success commit id: 

Command:
```
OMP_NUM_THREADS=28 FLAGS_use_mkldnn=true python python/paddle/fluid/contrib/slim/tests/qat2_int8_nlp_comparison.py --qat_model=${qat_model_dir} --fp32_model=${fp32_model_dir} --infer_data=${dataset_dir}/1.8w.bs1 --labels=${dataset_dir}/label.xnli.dev --batch_size=50 --batch_num=0 --quantized_ops="fc,reshape2,transpose2,matmul" --acc_diff_threshold=0.01
```
Note if it is latest commit, change param `quantized_ops` to `ops_to_quantize`

lidanqing-vv:
```
ERROR: test_graph_transformation (__main__.QatInt8NLPComparisonTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "python/paddle/fluid/contrib/slim/tests/qat2_int8_nlp_comparison.py", line 287, in test_graph
_transformation
    transform_to_int8=True)
  File "python/paddle/fluid/contrib/slim/tests/qat2_int8_nlp_comparison.py", line 189, in _predict
    fetch_list=fetch_targets)
  File "/home/danqing/repo/Paddle/build/python/paddle/fluid/executor.py", line 1073, in run
    six.reraise(*sys.exc_info())
  File "/home/danqing/repo/Paddle/build/python/paddle/fluid/executor.py", line 1068, in run
    return_merged=return_merged)
  File "/home/danqing/repo/Paddle/build/python/paddle/fluid/executor.py", line 1156, in _run_impl
    use_program_cache=use_program_cache)
  File "/home/danqing/repo/Paddle/build/python/paddle/fluid/executor.py", line 1231, in _run_program
    fetch_var_name)
EnforceNotMet:

--------------------------------------------
C++ Call Stacks (More useful to developers):
--------------------------------------------
0   std::string paddle::platform::GetTraceBackString<std::string const&>(std::string const&, char co
[...]

PR #24408, opened 2020-05-11
add batch size to the mkldnn matmul cache key

wojtuss:
This patch fixes the Ernie crash reported in the issue https://github.com/PaddlePaddle/Paddle/issues/24394.

test=develop


juncaipeng (approved):
LGTM

PR #24433, opened 2020-05-12
Fix bug with `if Tensor` in is_control_flow

Aurelius84:
In dygraph, `if tensor:` is supported  as followed when numel of `tensor` is 1:
```python
if tensor:
    .....

# it is  equivalent to:
if tensor != 0:
    .....
```
In this PR, we fix the bug when `node.if`  contains statement as above in `IfConditionVisitor` and support to transform `if tensor` into  `layers.cond`.
```python
if tensor:
    .....

# it will be transformed into:
bool_tensor_0 = fluid.layers.cast(x=tensor, dtype='bool')
res =  fluid.layers.cond(bool_tensor_0, lambda :true_fn_0(x), lambda : false_fn_0(x))
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -197,7 +197,6 @@ class NodeTestTransformer(gast.NodeTransformer):
                     node.values[i] = self._create_cast_node(child)
                 else:
                     node.values[i] = self._create_bool_node(child)
-                continue
         self.generic_visit(node)
         new_node = self._create_logic_node(node)
         return new_node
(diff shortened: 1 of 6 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -182,7 +192,11 @@ def visit_UnaryOp(self, node):
     def visit_BoolOp(self, node):
         for i, child in enumerate(node.values):
             if not is_candidate_node(child):
-                node.values[i] = self._create_bool_node(child)
+                node_wrapper = self.node_to_wrapper_map.get(child, None)
+                if node_wrapper and node_wrapper.node_var_type & NodeVarType.TENSOR_TYPES:
+                    node.values[i] = self._create_cast_node(child)
+                else:
+                    node.values[i] = self._create_bool_node(child)
                 continue
```
zhhsplendid: Do we need this continue?

Aurelius84: Thanks, fixed~~

liym27 (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #24453, opened 2020-05-12
[WIP] Set fp32_model optional in QAT test 

lidanqing-intel:
Modify `apply_fp32` function so that it could optimize fp32 and qat model by DNNL passes

There was a discussion that in qat test we should only test fp32 and int8 model. But now it turns out that it is not considerable enough. Because the qat model generated by user may have accuracy difference with original fp32 model (Not trained properly). So comparing original fp32 model and int8 model is not enough.


lidanqing-vv:
Need to disable fc in optimize_fp32_pass but not in INT8

lidanqing-vv:
@wojtuss  new PR had this feature. So I close this PR

PR #24456, opened 2020-05-12
Optimize error message, include dgc, nccl, size op

wangxicoding:
1. Optimize error message, include dgc_clip_by_norm_op, dgc_op, nccl_op, size_op.
2. Repair nccl op test. Move NcclReduce, NcclAllReduce, NcclBcast test to the same test function, for the reason serial execution is required for the same nccl comm.


gongweibao (approved):
LGTM

PR #24492, opened 2020-05-13
[Dy2Stat] Add assert for ProgramTranslator

zhhsplendid:
- PR type（PR 类型） is (A):
A. New features（新功能）--------------- D. Performance optimization（性能优化）
C. Function optimization（功能优化）-- F.  Others（其它）

- PR changes（改动点）is (B):
A. OPs（operators）------------------------ C. Docs（文档）
B. APIs（接口）------------------------------D. Others（其它）

- Use one sentence to describe what this PR does.（简述本次PR的目的和改动）

Add `assert` grammar for ProgramTranslator.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/assert_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/assert_transformer.py
@@ -1,0 +1,54 @@
+#   Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+import gast
+
+from paddle.fluid.dygraph.dygraph_to_static.static_analysis import AstNodeWrapper
+from paddle.fluid.dygraph.dygraph_to_static.static_analysis import NodeVarType
+from paddle.fluid.dygraph.dygraph_to_static.static_analysis import StaticAnalysisVisitor
+
+
+class AssertTransformer(gast.NodeTransformer):
+    """
+    A class transforms python assert to fluid.layers.Assert.
+    """
+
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/tests/unittests/dygraph_to_static/test_assert.py
```diff
@@ -0,0 +1,71 @@
...
+@declarative
+def dyfunc_assert_variable(x):
+    x_v = fluid.dygraph.to_variable(x)
+    assert x_v
+
+
+@declarative
+def dyfunc_assert_non_variable(x=True):
+    assert x
+
```
Aurelius84, line 34: LGTM.

I have a question: do we support `assert x, "some error msg"`?

python/paddle/fluid/dygraph/dygraph_to_static/assert_transformer.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def transform(self):
+        self.visit(self.root)
+
+    def visit_Assert(self, node):
+        if not self.static_analysis_visitor.is_tensor_node(node.test):
+            return node
+        cast_node = gast.Call(
+            func=gast.parse("fluid.layers.cast").body[0].value,
+            args=[node.test, gast.Constant(
```
chenwhql, line 46: two questions:
1. if `node .test` is a gast.Call or others, such as `assert func(x)`, whether we need tell users that this situation is temporarily not supported
2. if `node.msg` is not None, our transform will cause user's writed message loss, whether we need to tell users that

liym27 (commented):
LGTM for the rest

python/paddle/fluid/dygraph/dygraph_to_static/assert_transformer.py
zhhsplendid: 1. node.test is supported when it is a gast.Call. For example, if it is paddle API, it can be referred as PADDLE_RETURN_TYPE. Static analysis can also analyzes the return type of visible function (see some test example at test_static_analysis.py). Although not tested in unit test yet, I would say gast.Call or other node is supported.

2. I think one way is adding warning information. In my opinion, it's both OK to add or not to add warning. How about your suggestion? I can do as you suggest.

chenwhql: for 2, maybe we can add warning in later PR

chenwhql (approved):
LGTM

python/paddle/fluid/tests/unittests/dygraph_to_static/test_assert.py
zhhsplendid: Yes, but as Weihang commented, "some error msg" will loss

PR #24551, opened 2020-05-14
Add more version info to version.txt

cryoco:
When using Paddle Inference libs, it's vital to keep the local versions of dependency libs(CUDA, CUDNN,  GCC, etc.) the same as the inference lib. For users to do this check more easily, we added more version information to version.txt:
- CUDNN minor version(if using GPU)
- CXX compiler version
- TensorRT version (if using TensorRT)

because inconsistency of the above versions might cause serious compiling or runtime problems.
Now the version.txt looks like:
image


NHZlX (approved):
LGTM

PR #24560, opened 2020-05-14
add the support for pipeline

sandyhouse:
### PR types

New features

### PR changes

Others

### Describe

This pr implements the pipeline_trainer and the device worker (i.e., section_worker) to support pipeline.
With pipeline, we mean a program is split into multiple sub-programs (sections) each of which is run on a device.
The main purpose of pipeline is to train large-scale models that cannot fit on a single device or take advantage of different features of heterogeneous devices to make training more efficiently. 

Currently, you have to use device_guard to assign the device on which a sub-program (section) runs; the whole dataset is used for each iteration as the train_from_dataset interface is used. 

Todo:
1. use Executor.run instead of Executor.train_from_dataset.
2. use auto pipeline with fleet instead of device_guard.


paddle/fluid/framework/trainer_desc.proto
```diff
@@ -83,6 +83,7 @@ message SectionWorkerParameter {
   optional int64 sync_steps = 3 [ default = 1 ];
   optional int32 start_cpu_core_id = 4 [ default = 1 ];
   repeated string param_need_sync = 5;
+  optional int32 pipeline_micro_batch = 6;
```
gavin1332: be consistence with other code: microbatch, or micro_batch

python/paddle/fluid/tests/unittests/test_pipeline.py
```diff
@@ -140,38 +140,66 @@ def build_network(input, layers=50, class_dim=1000):
...
+    def _run(self):
+        main_prog = fluid.Program()
+        startup_prog = fluid.Program()
+        with fluid.program_guard(main_prog, startup_prog):
+            with fluid.device_guard("cpu"):
+                image = fluid.layers.data(
+                    name="image", shape=[3, 224, 224], dtype="float32")
+                label = fluid.layers.data(
+                    name="label", shape=[1], dtype="int64")
+                data_loader = fluid.io.DataLoader.from_generator(
```
gavin1332, line 152: why declare dataloader while building program?

python/paddle/fluid/tests/unittests/test_pipeline.py
```diff
@@ -140,38 +140,66 @@ def build_network(input, layers=50, class_dim=1000):
...
+
+        def train_reader():
+            for _ in range(4):
+                img = np.random.random(size=[3, 224, 224]).astype('float32')
+                label = np.random.random(size=[1]).astype('int64')
+                yield img, label
+
+        data_loader.set_sample_generator(train_reader, batch_size=1)
+        place = fluid.CPUPlace()
+        dataset = fluid.DatasetFactory().create_dataset('FileInstantDataset')
```
gavin1332, line 192: dummy dataset?

python/paddle/fluid/tests/unittests/test_pipeline.py
```diff
@@ -140,38 +140,66 @@ def build_network(input, layers=50, class_dim=1000):
...
+        dataset.set_thread(1)
+        dataset.set_filelist(['/tmp/tmp_2.txt'])
+        dataset.set_use_var([image, label])
+        exe = fluid.Executor(place)
+        exe.run(startup_prog)
+        data_loader.start()
+        exe.train_from_dataset(main_prog, dataset, debug=False)
+
+        data_loader.start()
+        exe.train_from_dataset(main_prog, dataset, debug=True)
```
gavin1332: why not checking the executing result?

python/paddle/fluid/tests/unittests/test_pipeline.py
sandyhouse: We have to put dataloader under the scope of 'with fluid.program_guard' as we use it as non-iterable.

paddle/fluid/framework/trainer_desc.proto
sandyhouse: Done.

python/paddle/fluid/tests/unittests/test_pipeline.py
sandyhouse: The dataset is only use for the interface 'train_from_dataset' and it has no actual meaning in our case. And I have added a comment to clarify this.

python/paddle/fluid/tests/unittests/test_pipeline.py
sandyhouse: The main aim of this test is to verify the correction function and satisfy the requirement of coverage. Moreover, for train_from_dataset, there is no good ways to get the results to check.

paddle/fluid/framework/pipeline_trainer.cc
```diff
@@ -141,10 +114,8 @@ std::string PipelineTrainer::GetDumpPath(int tid) {
...
   // Only set dump channel on the last section
-  for (int j = 0; j < pipeline_num_; ++j) {
-    for (size_t k = 0; k < workers_[section_num_ - 1][j].size(); ++k) {
-      workers_[section_num_ - 1][j][k]->SetChannelWriter(queue_.get());
-    }
+  for (int j = 0; j < section_num_; ++j) {
+    workers_[j]->SetChannelWriter(queue_.get());
   }
   // TODO(hutuxian): should make it as a config
   dump_thread_num_ = 1;
```
guru4elephant, line 115: is this necessary in pipeline trainer?

paddle/fluid/framework/pipeline_trainer.cc
```diff
@@ -154,150 +125,105 @@ void PipelineTrainer::InitDumpEnv() {
...
+                                     const platform::Place& place) {
+  auto& global_block = program.Block(0);
+  for (auto& var : global_block.AllVars()) {
+    int is_feed_var =
+        std::count(feed_var_names_.begin(), feed_var_names_.end(), var->Name());
+    if ((var->Persistable() || is_feed_var) && microbatch_id == 0) {
+      if (is_feed_var) {
+        auto* new_ptr = minibatch_scopes_[section_id]->Var(var->Name());
+        VLOG(3) << "data name: " << var->Name() << ", ptr: " << new_ptr;
+        InitializeVariable(new_ptr, var->GetType());
```
guru4elephant, line 133: can you just insert feed_op before calling the function?

python/paddle/fluid/device_worker.py
```diff
@@ -404,17 +404,16 @@ def _gen_worker_desc(self, trainer_desc):
         pipeline_opt = self._program._pipeline_opt
         section_param = trainer_desc.section_param
         section_param.queue_size = pipeline_opt["queue_size"]
-        section_param.sync_steps = pipeline_opt["sync_steps"]
+        section_param.microbatch = pipeline_opt["queue_size"]
```
guru4elephant: maybe just microbatch is enough, we don't need a new concept for configuration, namely queue_size.

guru4elephant: even though this is generated through inner API:)

python/paddle/fluid/tests/unittests/test_pipeline.py
```diff
@@ -140,38 +140,68 @@ def build_network(input, layers=50, class_dim=1000):
...
+        # And it has no actual meaning.
+        dataset = fluid.DatasetFactory().create_dataset('FileInstantDataset')
+        dataset.set_batch_size(1)
+        dataset.set_thread(1)
+        dataset.set_filelist(['/tmp/tmp_2.txt'])
+        dataset.set_use_var([image, label])
+        exe = fluid.Executor(place)
+        exe.run(startup_prog)
+        data_loader.start()
+        exe.train_from_dataset(main_prog, dataset, debug=debug)
```
guru4elephant, line 200: can you show a new example that users can run with Executor.run interface.

guru4elephant:
Please add more description in PR information. 
Two very important things to do: 
- make pipeline training compatible with Executor.run.
- add auto strategy for pipeline training so that users do not need to define device_guard which is very hard for users to understand.

paddle/fluid/framework/pipeline_trainer.cc
sandyhouse: Done. Deleted the unnecessary statements.

paddle/fluid/framework/pipeline_trainer.cc
sandyhouse: No feed_op to insert here. This function copies all vars from root scope to mini-batch scope or micro-batch scopes corresponding to var's type. Specifically, data vars and persistable vars are copied from root scope to mini-batch scopes and other vars are copied from root scope to micro-batch scopes.

python/paddle/fluid/device_worker.py
sandyhouse: Done.

python/paddle/fluid/tests/unittests/test_pipeline.py
sandyhouse: I am implementing the unified Executor.run interface. So the new demo to use Executor.run interface will be added in a later pr.

PR #24563, opened 2020-05-14
Add isCached() mechanism to elementwise_add DNNL

grygielski:
This PR introduced check if primitive is already cached and omit unnecessary actions in execution hot path. I've also added const identifiers to variables that don't change during execution. This is a concept and can be extended to other operator leading to improved performance.


jczaja (approved):
LGTM

luotao1 (approved):

PR #24609, opened 2020-05-18
Support create LoDTensorArray and return LoDTensorArray in control flow

liym27:
#### 1. Support to create LoDTensorArray in control flow (cond and while_loop)
For example:
```Python
# Create a null array in control flow sub-block
def true_fn():
    null_array = create_array("float32")
    # do something else
    return

def false_fn():
   # do something else
   return

res = cond(A==B, true_fn, false_fn)
```

Before this PR, it throws an error:
```
ValueError: Var array_0.out is not found recursively
```
Because `null_array(array_0.out )` is considered as input not output of OP.

#### 2. Fix bug: return LoDTensorArray in while_loop
Before this PR, `while_loop` doesn't assign LoDTensorArray to parent block. It causes that the returned LoDTensorArray can't be created in while block, which should assign to parent block.
For example:
```Python
def cond(i, array_parent): 
    return i < ten

def body(i, array_parent): 
    i = i + 1
    x = fluid.layers.ones(shape=[3], dtype="int32")
    array_sub = fluid.layers.create_array("int32")
    for j in range(5):
        x = func(array_parent[j])
        tensor_j = fluid.layers.fill_constant(shape=[1], dtype="int64", value=j)
        fluid.layers.array_write(x,tensor_j, array_sub)
    array_parent = array_sub
    return i, array_parent

# do something

i = layers.fill_constant(shape=[1], dtype='int64', value=0)
ten = layers.fill_constant(shape=[1], dtype='int64', value=10)
i, array_parent = layers.while_loop(cond, body, [i, array_parent])
```

#### 3. Delete unnecessary code in `list_transformer.py`

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/list_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/list_transformer.py
@@ -19,35 +19,17 @@ import gast
 
 from paddle.fluid.dygraph.dygraph_to_static.static_analysis import AstNodeWrapper, NodeVarType, StaticAnalysisVisitor
 from paddle.fluid.dygraph.dygraph_to_static.utils import ast_to_source_code, is_control_flow_to_transform
-from paddle.fluid.framework import core, default_main_program, Variable
+from paddle.fluid.framework import core, Variable
 from paddle.fluid.layers import array_length, array_read, array_write, create_array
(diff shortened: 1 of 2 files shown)


zhhsplendid (approved):
LGTM

gfwm2013 (commented):
LGTM

PR #24620, opened 2020-05-18
Use vector instead of pointer to init gflags

zhiqiu:
As the title.
The args of `InitGflags` passed from python to c++ is a vector of string, like `['--tryfromenv=check_nan_inf,fast_check_nan_inf,xxx']`. 

In function `InitGflags`, since `ParseCommandLineFlags` takes `char***` as the second input arg, the original implementation `new` a `char**` by:
    `char **arr = new char *[argv.size()];` 

It did not delete `arr`, and may result in a smalle size `memory leak` detected by some code analysis tools, such as valgrind.  For example, the result of `valgrind`:
image

It should be ponited out that the lambda function inside `InitGflags` uses `call_once`, so it is will not cause much memory leak indeed. The newed memory will be re-collected by os when the process exit. 

But, it is still not a good code practice, since `delete` is not paired with `new` explictly.

When I try to add delete in the lamdda function, I got `Error in python3: free(): invalid pointer:xx`. Debugging found that `arr` is assigned to a new memory address instread of the original newed one.

To avoid errors, I use vector instead of newed pointer to do it safely.

### DIFF
--- a/paddle/fluid/platform/init.cc
+++ b/paddle/fluid/platform/init.cc
@@ -50,7 +50,7 @@ std::once_flag glog_init_flag;
 std::once_flag p2p_init_flag;
 std::once_flag glog_warning_once_flag;
 
-bool InitGflags(std::vector<std::string> argv) {
+bool InitGflags(std::vector<std::string> args) {
   bool successed = false;
   std::call_once(gflags_init_flag, [&]() {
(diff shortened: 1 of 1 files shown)


phlrain (approved):

luotao1 (approved):
LGTM

PR #24633, opened 2020-05-19
Fix test_cond flaky test under Windows

zhhsplendid:
In the past, the test_cond will fail with 2% probability and easy to re-produce.

Now I re-run 300 times and no failure occurs. The probability of still has the failure is (1 - 2%) ^ 300 ~= 0.00004. We can say the random failure disappears. Maybe someone fixed some bugs in PE.


liym27 (approved):
LGTM

Aurelius84 (approved):

PR #24648, opened 2020-05-20
fix compile error about cub

ccmeteorljh:
+ fix cub error while compiling
image
The reason is: https://github.com/NVlabs/cub/releases update yesterday (May 19th), and we don't know what's happen.


luotao1 (approved):
LGTM

PR #24651, opened 2020-05-20
Decrease the Batch Size of SE-Resnext in Unittest.

zhhsplendid:
As discussed with QA, we will use p4 machine for unit test and the GPU on those machine may not have enough GPU, which can cause "test_parallel_executor_seresnext_base_gpu" failed. So I decrease the batch size.


liym27 (approved):
LGTM

PR #24655, opened 2020-05-20
Fix _forget_bias.stop_gradient workround in hapi.text.BasicLSTMCell

guoshengCS:
Fix _forget_bias.stop_gradient workround in hapi.text.BasicLSTMCell


kuke (approved):
LGTM

PR #24692, opened 2020-05-21
Fix test_bert on GPU

liym27:
- **PR type（PR 类型） is (F ):**
A. New features（新功能）---------------- D. Performance optimization（性能优化）
C. Function optimization（功能优化）------**F.  Others（其它）**

- **PR changes（改动点）is ( D ):**
A. OPs（operators）---------------------- C. Docs（文档）
B. APIs（接口）--------------------------- **D. Others（其它）**

- **Use one sentence to describe what this PR does.（简述本次PR的目的和改动）**
Disable test_bert on GPU. 
-----------------------
#### Optional（选填, If None, please delete it）

- **Describe what this PR does in detail. If this PR fixes an issue, please give the issue id.**
  
DataLoader makes the data diff even if the data of reader is the same.  This PR doesn't use DataLoader to pass the test.

- **If you modified docs, please make sure that both Chinese and English docs were modified and provide a preview screenshot. （文档必填）**
   

- **Please write down other information you want to tell reviewers.**

### DIFF
--- a/python/paddle/fluid/tests/unittests/dygraph_to_static/bert_utils.py
+++ b/python/paddle/fluid/tests/unittests/dygraph_to_static/bert_utils.py
@@ -260,51 +260,48 @@ class DataReader(object):
             yield token_ids, sent_ids, pos_ids, label
 
     def data_generator(self):
-        def wrapper():
-            def reader():
-                for epoch in range(self.epoch):
-                    self.current_epoch = epoch + 1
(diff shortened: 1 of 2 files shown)


zhhsplendid (approved):
I will approve this PR, but we should consider to use DataLoader back after we fix it.

liym27:
> I will approve this PR, but we should consider to use DataLoader back after we fix it.

OK, I will use DataLoader back.

PR #24727, opened 2020-05-25
[Dy2static] Fix random bug in bert unittest

chenwhql:
- **PR type（PR 类型） is ( B ):**
A. New features（新功能）---------------- D. Performance optimization（性能优化）
C. Function optimization（功能优化）------F.  Others（其它）

- **PR changes（改动点）is ( D ):**
A. OPs（operators）---------------------- C. Docs（文档）
B. APIs（接口）--------------------------- D. Others（其它）

- **Use one sentence to describe what this PR does.（简述本次PR的目的和改动）**

This PR fix random using problem in dy2static unittest `test_bert`, fix the problem in https://github.com/PaddlePaddle/Paddle/pull/24692.

-----------------------
#### Optional（选填, If None, please delete it）

- **Describe what this PR does in detail. If this PR fixes an issue, please give the issue id.**
  

`test_bert` data reader shows different generate result with the same seed, this is very strange, so we suspect that the DataLoader made changes to the input data.

But after testing, I found that these are the problems of `numpy.random` and `python random`, not the problem of `DataLoader`

1. `numpy.random.seed` is not thread-safe, but `fluid.DataLoader` use child thread to speed up data reading, the seed may be different in different thread even if you set the seed by `np.random.seed`. related doc:
- [Numpy: random seed and multithreading causes differing results](https://stackoverflow.com/questions/58875076/numpy-random-seed-and-multithreading-causes-differing-results)
- [Differences between numpy.random and random.random in Python](https://stackoverflow.com/questions/7029993/differences-between-numpy-random-and-random-random-in-python)

2. python `random` has bug in python2, we should avoid using it in unittest. related doc:
- [random.randint shows different output in Python 2.x and Python 3.x with same seed](https://stackoverflow.com/questions/55647936/random-randint-shows-different-output-in-python-2-x-and-python-3-x-with-same-see)

3. for this case, we can use [`numpy.random.RandomState`](https://het.as.utexas.edu/HET/Software/Numpy/reference/generated/numpy.random.RandomState.html)

- **If you modified docs, please make sure that both Chinese and English docs were modified and provide a preview screenshot. （文档必填）**
   

- **Please write down other information you want to tell reviewers.**

### DIFF
--- a/python/paddle/fluid/tests/unittests/dygraph_to_static/bert_utils.py
+++ b/python/paddle/fluid/tests/unittests/dygraph_to_static/bert_utils.py
@@ -49,10 +49,13 @@ def mask(batch_tokens, total_token_num, vocab_size, CLS=1, SEP=2, MASK=3):
     max_len = max([len(sent) for sent in batch_tokens])
     mask_label = []
     mask_pos = []
-    np.random.seed(SEED)
-    prob_mask = np.random.rand(total_token_num)
+    # NOTE: numpy random is not thread-safe, for async DataLoader,
+    # using np.random.seed() directly is risky, using RandomState
+    # class is a better way
+    self_random = np.random.RandomState(SEED)
+    prob_mask = self_random.rand(total_token_num)
     # Note: the first token is [CLS], so [low=1]
-    replace_ids = np.random.randint(1, high=vocab_size, size=total_token_num)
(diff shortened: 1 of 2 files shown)


zhhsplendid (approved):
LGTM

liym27 (approved):
LGTM

phlrain (approved):

PR #24742, opened 2020-05-26
Collecting concat output threshold for quantization

juncaipeng:
PR types:Function optimization

PR changes: Collecting concat output threshold for quantization

Describe: Add concat to _out_scale_op_list


cryoco (approved):
LGTM

Issue #24585, opened 2020-05-15
1.8.0版本，conv_transpose_eltwiseadd_bn_fuse_pass导致PaddleOCR检测模型CPU预测结果错误


cryoco:
**System information**
-PaddlePaddle version: `v1.8.0`, `0231f58e592ad9f673ac1832d8c495c8ed65d24f`
-CPU: `Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz`
-GPU: V100
-OS Platform (eg.Mac OS 10.14): Ubuntu1404 
-Docker image: `hub.baidubce.com/paddlepaddle/paddle:latest-dev`
-Python version: Python3.7
-Cmake orders
-C++version.txt: used python api
-API information: inference configuration
```python
config.switch_use_feed_fetch_ops(False)
config.disable_gpu()
```

**To Reproduce** 

1. run test with `conv_transpose_eltwiseadd_bn_fuse_pass`

``` shell
python3.7 test_ocr_det.py --model_file=det/model --params_file=det/params --float32
```

The head part of output：
![image](https://user-images.githubusercontent.com/12407750/82036396-709fad00-96d3-11ea-8793-0194dae32297.png)

2. run test without `conv_transpose_eltwiseadd_bn_fuse_pass`
by adding 
`config.delete_pass("conv_transpose_eltwiseadd_bn_fuse_pass")` in AnalysisConfig.

``` shell
python3.7 test_ocr_det.py --model_file=det/model --params_file=det/params --float32
```

The head part of output：
![image](https://user-images.githubusercontent.com/12407750/82036641-b8becf80-96d3-11ea-9b2d-0f37c0b32fa3.png)

**Describe your current behavior**
It has been verified that when using `conv_transpose_eltwiseadd_bn_fuse_pass`, the cpu inference output is wrong. It's easy to check the right output by using GPU to perform inference:

```python
#config.disable_gpu()
config.enable_use_gpu(100, 0)
```

This issue can be reproduced with and without MKLDNN.

**Code to reproduce the issue**
[test_ocr_det.txt](https://github.com/PaddlePaddle/Paddle/files/4633635/test_ocr_det.txt)

Need to be renamed to test_ocr_det.py

**Model to reproduce the issue**

```shell
wget https://paddleocr.bj.bcebos.com/inference.tar
tar -xf inference.tar
```
The `det` model is the one which reproduces this issue.

**Other info / logs**

luotao1:
[...]

PR #24756, opened 2020-05-26
[oneDNN] Clearing mkldnn cache in naiveexecutor destructor

jczaja:
PR types:
 Bug fixes
PR changes:
Add clearing of oneDNN cache in NaiveExecutor destructor and when IR pass is called. Also there is a unit tests testing fuse conv {eltwise_add} batch_norm that compare ir optimized execution aginast one without IR fuse pass applied
Describe:


paddle/fluid/framework/ir/mkldnn/mkldnn_conv_bn_fuse_pass_tester.cc
```diff
@@ -0,0 +1,255 @@
...
+    // Get Program from graph
+    ProgramDesc optimized_prog;
+    auto graph2program_pass =
+        paddle::framework::ir::PassRegistry::Instance().Get(
+            "graph_to_program_pass");
+    graph2program_pass->SetNotOwned<paddle::framework::ProgramDesc>(
+        "program", &optimized_prog);
+    graph2program_pass->Apply(graph.release());
+
+    exe.Prepare(&scope, optimized_prog, 0, false);
```
lidanqing-vv, line 277: I have one question. Why do you need to execute two graphs in one exe? Is it because the one QAT test which failed run like this? 
Does this PR also solve the customer model issue? Thanks

jczaja: I used one Executor as it was not needed to have two. Two executors can be used as well.

This PR is fixing some issues with cache clearing that were discovered when working on customer issue. UT introduced here is also used to reproduce an issue, but there is no fix to customer issue yet.

jczaja: I have updated the PR with candidate fix to issue #24585 . Currently testing

paddle/fluid/framework/ir/mkldnn/mkldnn_conv_bn_fuse_pass_tester.cc
```diff
@@ -0,0 +1,303 @@
...
+      FillTensorWithRandomData(mean2_tensor, 1.0f, 2.0f, place);
+      FillTensorWithRandomData(variance2_tensor, 1.0f, 2.0f, place);
+    }
+
+    exe.Run();
+
+    // Get result without IR passes applied
+    // Need to copy result over as the same scope is used in both executors
+    // so first result will be overwritten by second
+    auto* j_tensor = exe.FindTensor("m");
```
sfraczek: "j_tensor" causes a little confusion. Consider changing to "m_tensor".

sfraczek (commented):
LGTM. good job :)

paddle/fluid/framework/ir/mkldnn/mkldnn_conv_bn_fuse_pass_tester.cc
jczaja: Good catch! thanks.

sfraczek (approved):
LGTM

luotao1 (approved):

grygielski (approved):
LGTM

PR #24774, opened 2020-05-27
Fix out of range error for outs map.

wzzju:
PR types: Bug fixes
PR changes: Others
Describe:
image

As shown above, when `OutState` is `nullptr`, `outs["OutState"][0]` will cause the out of range error.


zhiqiu (approved):
LGTM

PR #24783, opened 2020-05-27
[Paddle-TRT] slice kernel optimization

zlsh80826:
### PR types

Function optimization
### PR changes

OPs
### Describe

In inference::tensorrt::plugin::SliceKernel, we should parallelize the shared memory movement to each thread. There is 15% performance improvement by parallelizing the shared memory movement.


jeng1220:
@NHZlX 
Plz take a look

PR #24786, opened 2020-05-28
add mock barrier all

xjqbest:
* add mock barrier all, when worker restart, we need fake barrier all to avoid hang
* test=develop


yaoxuefeng6 (approved):
LGTM

PR #24799, opened 2020-05-28
[Dy2Static]Convert while stmt and convert logical_XX

liym27:
PR types: New features

PR changes: dygraph to static graph

Describe:
#### 2. Add functions `convert_XX` to run the transformed code dynamically. Which means, 
- All while stmts and logical stmts are transformed whether it depends on `Tensor` or not;
- But in run-time, the code will be run in different ways (1) run paddle op like `while_loop`; (2) run python while. How it runs depends on whether it depends on `Tensor`. 
> This method will not use static variable types while transforming ast node, because the variable types got from static analysis are often inaccurate.

---

`convert_while_loop `: A function representation of a Python ``while`` statement.
`convert_logical_and `:  A function representation of a Python ``and`` statement.
`convert_logical_or `: A function representation of a Python ``or`` statement.
`convert_logical_not `: A function representation of a Python ``not`` statement.

---
Related PR :
convert_ifelse: https://github.com/PaddlePaddle/Paddle/pull/24866

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
@@ -22,11 +22,9 @@ def convert_while_loop(cond, body, loop_vars):
     A function representation of a Python ``while`` statement.
 
     Args:
-        cond(Callable): A callable returning a boolean variable controlling whether to continue looping. And ``cond`` takes
-        as many arguments as ``loop_vars`` .
-        body(Callable): A callable returning a tuple or list of variables of the same arity
-            (length and structure) and types as ``loops_vars`` . And ``body`` takes as many arguments as ``loop_vars`` .
-        loop_vars(list|tuple): A list or tuple of variables that is passed to both ``cond`` and ``body`` .
+        cond(Callable): A callable object that returns a boolean variable to control whether to  execute the loop body.  It takes  ``loop_vars`` as arguments.
+        body(Callable): A callable object that returns a tuple or list of variables with the same arguments ``loops_vars`` as ``cond`` .
+        loop_vars(list|tuple): A list or tuple of variables passed to ``cond`` and ``body`` .
 
     Returns:
         A list or tuple of variables which returned by ``body`` .
@@ -79,6 +77,8 @@ def _run_paddle_logical_and(x, y):
 
 
 def _run_py_logical_and(x, y):
+    assert not isinstance(x, Variable)
+    # NOTE: Returns y if x is True
     return x and y
 
 
@@ -108,6 +108,8 @@ def _run_paddle_logical_or(x, y):
 
 
 def _run_py_logical_or(x, y):
+    assert not isinstance(x, Variable)
+    # NOTE: Returns y if x is False
     return x or y
 
 
--- a/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from paddle.fluid.framework import Variable
-from paddle.fluid.layers import fill_constant, control_flow, logical_and, logical_or, logical_not
+from paddle.fluid.layers import control_flow, logical_and, logical_or, logical_not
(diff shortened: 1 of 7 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    A function representation of a Python ``while`` statement.
+
+    Args:
+        cond(Callable): A callable returning a boolean variable controlling whether to continue looping. And ``cond`` takes
+        as many arguments as ``loop_vars`` .
+        body(Callable): A callable returning a tuple or list of variables of the same arity
+            (length and structure) and types as ``loops_vars`` . And ``body`` takes as many arguments as ``loop_vars`` .
+        loop_vars(list|tuple): A list or tuple of variables that is passed to both ``cond`` and ``body`` .
+
```
Aurelius84, line 28: + cond(Callable): A callable object that returns a boolean variable to control whether to  execute the loop body.  It takes  ``loop_vars`` as arguments.

+ body(Callable): A callable object that returns a tuple or list of variables with the same arguments ``loops_vars`` as ``cond`` . 

+ loop_vars(list|tuple): A list or tuple of variables passed to ``cond`` and ``body`` .

Aurelius84 (commented):
Great!

chenwhql (commented):
Do we need to use the same logic to deal with `for stmt`?

liym27:
> Do we need to use the same logic to deal with `for stmt`?

@chenwhql 
Yes, I am dealing with `for stmt` too.

python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
liym27: > * cond(Callable): A callable object that returns a boolean variable to control whether to  execute the loop body.  It takes  `loop_vars` as arguments.
> * body(Callable): A callable object that returns a tuple or list of variables with the same arguments `loops_vars` as `cond` .
> * loop_vars(list|tuple): A list or tuple of variables passed to `cond` and `body` .

Thanks, done

Aurelius84 (approved):
LGTM

chenwhql (approved):
LGTM

python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
```diff
@@ -0,0 +1,138 @@
...
+    Args:
+        cond(Callable): A callable object that returns a boolean variable to control whether to  execute the loop body.  It takes  ``loop_vars`` as arguments.
+        body(Callable): A callable object that returns a tuple or list of variables with the same arguments ``loops_vars`` as ``cond`` .
+        loop_vars(list|tuple): A list or tuple of variables passed to ``cond`` and ``body`` .
+
+    Returns:
+        A list or tuple of variables which returned by ``body`` .
+    """
+
+    pred = cond(*loop_vars)
```
zhhsplendid, line 33: Great! This convert_call writing is clearer than I thought.

But there is a potential very very tiny issue:

At line 33, we calculate cond(*loop_vars) once and in  _run_py_while or _run_paddle_while_loop we calculate it again.
1. It may be slower if cond is very expensive.
2. If loop_vars is changed during cond callable, then it causes bug.

But both above issues won't happen now because usually cond is just O(1) and current logical_and/logical_not/... doesn't change the loop_vars.

Since CI has passed, and I notice that you commented some test cases for `if` case, you can change in the future PR. I will approve this PR now.

zhhsplendid (approved):
LGTM

python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
liym27: > 1. It may be slower if cond is very expensive.
> 2. If loop_vars is changed during cond callable, then it causes bug.
> But both above issues won't happen now because usually cond is just O(1) and current logical_and/logical_not/... doesn't change the loop_vars.
> 

Thanks!  Considering that there is no error thrown, I didn't deal it in this PR.  I will change it in the future PR.

PR #24810, opened 2020-05-29
[do not merge] add fc parallel fuse pass

sfraczek:
### PR types
Performance optimization

### PR changes
Others

### Describe
For Ernie model, adds a fuse that replaces 3 parallel FC ops that accept the same input with a single FC op and a split.

The fuse is applied only when the following attributes have the same values across 3 ops:
* in_num_col_dims,
* Scale_in, 
* Scale_out,
* activation_type,
* use_mkldnn,
* padding_weights.

It works by concatenating weights and biases of FCs along dim 1 and adding a Split on FC's output along axis 2 which splits it back into 3 paths. 
#### before
image
#### after
image
### padding
If padding_weights is enabled it handles that case also.
MicrosoftTeams-image


sfraczek:
I have here the same problem as there: https://github.com/PaddlePaddle/Paddle/pull/23754#issuecomment-615273545
It seems unrelated to my functionality. @luotao1 do you know anything about this error? Has anyone else experienced this?

luotao1:
> do you know anything about this error

I don't know, how do you solve it in #23754 before?

sfraczek:
> > do you know anything about this error
> 
> I don't know, how do you solve it in #23754 before?

It was only showing up on COVERAGE build so I did this:
https://github.com/PaddlePaddle/Paddle/pull/23754/files#diff-dc2f06a75c6fa6e441469e2ef1365b5eR149-R151
Now it happens on more builds so I don't want to resort to this. I want to look into the cause deeper.

sfraczek:
I have found a solution! Removing include of google mock has fixed it! I will try to also apply this to the previous PR. https://github.com/PaddlePaddle/Paddle/pull/24810/commits/c1881f8b6e631b0bc2e97185ed2f4ce55a43cbd5

sfraczek:
Latency degradation, closing PR.

test | develop | parallel fc  
-- | -- | --
TA_B_bert_performance_b1 | 373.529 | 372.504 | 99.73% | -0.27%
ERNIE_int8_performance_b1 | 82.5472 | 86.1672 | 104.39% | 4.39%
ERNIE_fp32_performance_b1 | 239.376 | 241.409 | 100.85% | 0.85%
ERNIE_int8_20_threads_performance_b1 | 12.281 | 15.7741 | 128.44% | 28.44%
ERNIE_fp32_20_threads_performance_b1 | 23.6172 | 25.2045 | 106.72% | 6.72%

PR #24823, opened 2020-06-01
Support gradient accumulation of fp16 in imperative mode

zhiqiu:
PR types: New features

PR changes: Others

Describe: Support gradient accumulation of fp16 in imperative mode

Note, original gradient accumulation of `float` and `double` uses `AXPY` of blas library, which does not support `float16`. So `float16` calculation uses `eigen` library.

### DIFF
--- a/paddle/fluid/imperative/gradient_accumulator.cc
+++ b/paddle/fluid/imperative/gradient_accumulator.cc
@@ -113,43 +113,8 @@ void TensorAddImpl(const framework::Tensor& src, framework::Tensor* dst,
   platform::DeviceContextPool& pool = platform::DeviceContextPool::Instance();
   paddle::platform::DeviceContext* ctx = pool.Get(place);
   auto dev_ctx = dynamic_cast<DeviceContext*>(ctx);
-  switch (src.dims().size()) {
-    case 1: {
-      operators::math::AddTensor<DeviceContext, T, 1> func;
-      func(dev_ctx, src, dst);
-      break;
-    }
-    case 2: {
-      operators::math::AddTensor<DeviceContext, T, 2> func;
-      func(dev_ctx, src, dst);
-      break;
-    }
-    case 3: {
-      operators::math::AddTensor<DeviceContext, T, 3> func;
-      func(dev_ctx, src, dst);
-      break;
-    }
-    case 4: {
-      operators::math::AddTensor<DeviceContext, T, 4> func;
-      func(dev_ctx, src, dst);
-      break;
-    }
-    case 5: {
-      operators::math::AddTensor<DeviceContext, T, 5> func;
-      func(dev_ctx, src, dst);
-      break;
-    }
-    case 6: {
-      operators::math::AddTensor<DeviceContext, T, 6> func;
-      func(dev_ctx, src, dst);
-      break;
-    }
-    default:
-      PADDLE_THROW(platform::errors::InvalidArgument(
-          "TensorAddImpl doesn't supports tensors whose ranks are greater "
-          "than 6."));
-      break;
-  }
+  operators::math::ElementwiseAddTo<DeviceContext, T> func;
+  func(dev_ctx, src, dst);
 }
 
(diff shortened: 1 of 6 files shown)


chalsliu (dismissed):

paddle/fluid/imperative/gradient_accumulator.cc
```diff
@@ -100,6 +107,51 @@ class TensorAddFunctor : public boost::static_visitor<> {
...
+    }
+    case 6: {
+      operators::math::AddTensor<DeviceContext, T, 6> func;
+      func(dev_ctx, src, dst);
+      break;
+    }
+    default:
+      PADDLE_THROW(platform::errors::InvalidArgument(
+          "TensorAddImpl doesn't supports tensors whose ranks are greater "
+          "than 6."));
```
chenwhql: why doesn't support?

paddle/fluid/imperative/gradient_accumulator.cc
```diff
@@ -133,8 +185,34 @@ void TensorAdd(const framework::Variable& src, framework::Variable* dst) {
...
+      TensorAddImpl<platform::CUDADeviceContext, platform::float16>(
+          src_tensor, dst_tensor, place);
+#else
+      PADDLE_THROW(platform::errors::PermissionDenied(
+          "Gradient accumulation of data type (%s) on place (%s) is not "
+          "supported in imperative mode",
+          framework::DataTypeToString(data_type), place));
+#endif
+    }
+    if (platform::is_cpu_place(place)) {
```
chenwhql: why not use `else if`?

paddle/fluid/imperative/gradient_accumulator.cc
```diff
@@ -133,8 +185,34 @@ void TensorAdd(const framework::Variable& src, framework::Variable* dst) {
...
 
-  PADDLE_THROW("Not supported data type %s for AddTo",
-               framework::DataTypeToString(data_type));
+  if (data_type == framework::proto::VarType::FP16) {
+    if (platform::is_gpu_place(place)) {
+#ifdef PADDLE_WITH_CUDA
+      TensorAddImpl<platform::CUDADeviceContext, platform::float16>(
+          src_tensor, dst_tensor, place);
+#else
+      PADDLE_THROW(platform::errors::PermissionDenied(
```
chenwhql: may be `Unimplemented` is better than `PermissionDenied` here?

paddle/fluid/imperative/gradient_accumulator.cc
zhiqiu: Not sure, there are several operators which check `rank <= 6`, for example, `transpose` op. I will remove `switch...case...` and simplify the code.

paddle/fluid/imperative/gradient_accumulator.cc
zhiqiu: I will change that.

zhiqiu: I will change that, thx.

paddle/fluid/imperative/gradient_accumulator.cc
zhiqiu: I will change that, thx.

paddle/fluid/imperative/gradient_accumulator.cc
```diff
@@ -133,8 +150,26 @@ void TensorAdd(const framework::Variable& src, framework::Variable* dst) {
...
+#else
+      PADDLE_THROW(platform::errors::Unimplemented(
+          "Gradient accumulation of data type (%s) on place (%s) is not "
+          "supported in imperative mode",
+          framework::DataTypeToString(data_type), place));
+#endif
+    } else if (platform::is_cpu_place(place)) {
+      return TensorAddImpl<platform::CPUDeviceContext, platform::float16>(
+          src_tensor, dst_tensor, place);
+    }
```
chenwhql, line 167: why remove `cuda_pinned_place` check? do we need an `else` branch here?

chenwhql (approved):
LGTM

PR #21093, opened 2019-11-08
Further simplify the C++ error info stack

chenwhql:
### Chinese

经过之前对C++报错信息栈的简化，去掉了提示效用比较差的函数地址，代码偏移等信息，但是由于boost::variant的函数参数过多，C++信息栈中还是会出现很长的函数名，占据大片的内容，但是这些冗余的参数事实上是没有有效信息的，例如

3   paddle::framework::OperatorWithKernel::RunImpl(paddle::framework::Scope const&, boost::variant<paddle::platform::CUDAPlace, paddle::platform::CPUPlace, paddle::platform::CUDAPinnedPlace, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_, boost::detail::variant::void_> const&, paddle::framework::RuntimeContext*) const

这行报错信息中，有效的只有以下内容，剩余的`boost::detail::variant::void_,`均不必显示

3   paddle::framework::OperatorWithKernel::RunImpl(paddle::framework::Scope const&, boost::variant<paddle::platform::CUDAPlace, paddle::platform::CPUPlace, paddle::platform::CUDAPinnedPlace> const&, paddle::framework::RuntimeContext*) const

因此该PR通过将Place设计为继承`boost::variant<CUDAPlace, CPUPlace, CUDAPinnedPlace>`的类，简化由于Place参数过长导致的C++信息栈信息密度低，阅读体验差的问题

> 正则表达式替换字符串在多平台上可能有风险，这种解决方法最安全
> 主要就是Place参数导致的信息冗长，没必要删除其他参数的前缀了

### English

The PR make `Place` class inherit `boost::variant<CUDAPlace, CPUPlace, CUDAPinnedPlace>`, simplifying the low information density of the C++ message stack due to excessive Place parameters and poor reading experience.

#### Original

[...]

PR #19895, opened 2019-09-19
Paddle error message stack shaping and optimization

chenwhql:
What this PR does：
- Change the arrangement of the error message stack, from top to bottom is `C++ Call Stack`, `Python Call Stack`, `Error message summary`
- Add the separation and prompt between stack information
- Simplify C++ stack information
- Discard error type that confuses users, like `Enforce failed`

Original error message:
```
paddle.fluid.core_avx.EnforceNotMet: Invoke operator mul error.
Python Callstacks:
  File "/usr/local/lib/python3.5/dist-packages/paddle/fluid/framework.py", line 1771, in append_op
    attrs=kwargs.get("attrs", None))
  File "/usr/local/lib/python3.5/dist-packages/paddle/fluid/layer_helper.py", line 43, in append_op
    return self.main_program.current_block().append_op(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/paddle/fluid/layers/nn.py", line 334, in fc
    "y_num_col_dims": 1})
  File "startup.py", line 15, in <module>
    y_ = fluid.layers.fc(input=x, size=1, act=None)
C++ Callstacks:
Enforce failed. Expected x_mat_dims[1] == y_mat_dims[0], but received x_mat_dims[1]:13 != y_mat_dims[0]:12.
First matrix's width must be equal with second matrix's height. 13, 12 at [/work/Paddle/paddle/fluid/operators/mul_op.cc:63]
PaddlePaddle Call Stacks:
0       0x7fb7b6d511ddp void paddle::platform::EnforceNotMet::Init<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, char const*, int) + 509
1       0x7fb7b6d51ecdp paddle::platform::EnforceNotMet::EnforceNotMet(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, char const*, int) + 125
2       0x7fb7b70a5d4bp paddle::operators::MulOp::InferShape(paddle::framework::InferShapeContext*) const + 3435
[...]

PR #24831, opened 2020-06-01
The third time to simplify the C ++ error stack

chenwhql:
PR types: Function optimization

PR changes: Others

Describe:

The paddle C ++ error stack has always had the problem that the error info is too long and difficult to read. I have tried to solve this problem twice before:
- The first time, I made the C ++ error stack clearer by deleting almost useless information and useless entries, https://github.com/PaddlePaddle/Paddle/pull/19895.
- And the second time, I rewritten `Place` to reduce too much useless information due to the too long `boost` variable name introduced by `Place`, https://github.com/PaddlePaddle/Paddle/pull/21093.

However, the above methods cannot completely solve this problem.

Now in dygraph mode,  the C ++ error stack is still very long and almost unreadable!

### For example:

- code:
```python
import numpy
import paddle.fluid as fluid

with fluid.dygraph.guard():
    x = numpy.random.random(size=(10, 2)).astype('float32')
    linear = fluid.dygraph.Linear(1, 10)
    data = fluid.dygraph.to_variable(x)
    res = linear(data)
```

- original C++ error stack example:
image

**This is terrible!**

**So I simplified the C ++ error stack once again. Example as follow:**

- new C++ error stack example:
image

This solution has three advantages:
1. the error information expression efficiency has been significantly improved! 
2. no lost usefull information
3. if more complex variable types added again, we can easily add processing to it.

备注：这里没有使用`std::string*`直接在一个字符串上操作，报错过程中copy 字符串不是性能瓶颈，在这种情况下最好避免`std::string*`的使用，这里用指针会在CI并行跑单测的时候导致一部分单测意外挂掉

### DIFF
--- a/paddle/fluid/platform/enforce.h
+++ b/paddle/fluid/platform/enforce.h
@@ -196,30 +196,32 @@ struct BinaryCompareMessageConverter<false> {
 }  // namespace details
 
 template <typename T>
-inline void ReplaceComplexTypeStr(std::string* str, const char* type_name) {
-  auto type_str = demangle(typeid(T).name());
-  std::size_t found = str->find(type_str);
-  if (found != std::string::npos) {
-    str->replace(found, type_str.length(), type_name);
-    ReplaceComplexTypeStr<T>(str, type_name);
+inline std::string ReplaceComplexTypeStr(std::string str,
+                                         const std::string& type_name) {
+  auto demangle_type_str = demangle(typeid(T).name());
+  size_t start_pos = 0;
+  while ((start_pos = str.find(demangle_type_str, start_pos)) !=
+         std::string::npos) {
+    str.replace(start_pos, demangle_type_str.length(), type_name);
+    start_pos += type_name.length();
   }
+  return str;
 }
 
-#define __REPLACE_COMPLEX_TYPE_STR__(__TYPENAME, __STR)                        \
-  do {                                                                         \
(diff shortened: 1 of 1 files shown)


luotao1 (approved):

paddle/fluid/platform/enforce.h
```diff
@@ -191,6 +195,35 @@ struct BinaryCompareMessageConverter<false> {
...
+}
+
+#define __REPLACE_COMPLEX_TYPE_STR__(__TYPENAME, __STR)                       \
+  do {                                                                        \
+    __STR = paddle::platform::ReplaceComplexTypeStr<__TYPENAME>(__STR,        \
+                                                                #__TYPENAME); \
+  } while (0)
+
+inline std::string SimplifyDemangleStr(std::string str) {
+  // the older is important, you have to put complex types in front
```
zhiqiu, line 218: Not important.
Why should put complex types in front?

zhiqiu (approved):
LGTM

paddle/fluid/platform/enforce.h
chenwhql: for example, if we replace `Attribute` firstly, then we can't found `AttributeMap`, because the part of `AttributeMap` had been replaced, but actually, we can replace whole `AttributeMap` firstly.

lanxianghit (approved):

PR #24834, opened 2020-06-01
[Paddle-TRT] Fixes #24731, opt for SoftmaxKernelWithEltadd kernel, test=develop

### LINKED ISSUE #24731: [Paddle-TRT] Different behaviors between blockReduce and warpReduce
1) Paddle Version: develop
2) GPU/cuda-10.2/cudnn7.6.5
3) Ubuntu16.04

Hello! There are different behaviors between [blockReduceSum](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/math_cuda_utils.h#L138) and [warpReduceSum](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/math_cuda_utils.h#L126). 
[warpReduceSum](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/math_cuda_utils.h#L126) makes each thread in the warp has a copy of summation. However, [blockReduceSum](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/math_cuda_utils.h#L138) only ensures the first warp has the summation, which results in one more block synchronization like [here](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/bert_encoder_functor.cu#L157).

I suggest two solutions, 
1) Change the behavior of blockReduceSum to [AllReduce](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/operations.html#allreduce), which computes the summation and ensures each thread in the block have the summation copy.
We only need to change one line to do this. Change the [code]((https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/math_cuda_utils.h#L151)) ``val = (threadIdx.x < block_span) ? shared[lane] : static_cast<T>(0.0f);`` to ``val = (lane < block_span) ? shared[lane] : static_cast<T>(0.0f);``. By doing this, each warp has a copy of first round result. Thus, all threads in the block have the reduced values after the second warpReduce.
2) In blockReduceSum, add a condition which lets only the first warp in the block does the second reduction. Because we only need to ensure ``threadIdx.x`` has the correct reduced value ([SoftmaxKernelWithEltadd](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/bert_encoder_functor.cu#L161) is the only caller), so only the first warp needs to do the second reduction.
i.e. 
```
if (threadIdx.x < warpSize) {  
    val = (threadIdx.x < block_span) ? shared[lane] : static_cast<T>(0.0f);
    val = warpReduceSum<T>(val, mask);
}
```

In my experiments, the first one has higher performance, because we can delete the unnecessary [shared memory copy](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/bert_encoder_functor.cu#L164) and [__syncthreads](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/bert_encoder_functor.cu#L166) after calling the blockReduceXXX function.

BTW, the same issue exists in the [blockReduceMax](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/operators/math/math_cuda_utils.h#L170) too.

zlsh80826:
### PR types

Function optimization
### PR changes

OPs
### Describe

Few changes to give the kernel ``SoftmaxKernelWithEltadd`` 1.25x speedup
1. Change the blockReduce behavior as  issue #24731 describe. Every thread calls the blockReduceXXX can directly obtain the reduced value. Thus, shared memory broadcast isn't needed.
2. Number of launch threads for ``SoftmaxKernelWithEltadd`` align to 32 instead of power of 2.
3. vectorization when available


jeng1220:
@NHZlX 
Try this simple demo and you will see that each of thread in the block gets the summation. This PR should be safe to be merged.

```cuda
#include <stdio.h>
#define WARP_SIZE 32
#define HALF_WARP 16
#define NB_WARP 2

template <typename T>
__inline__ __device__ T warpReduceSum(T val, unsigned lane_mask) {
  for (int mask = HALF_WARP; mask > 0; mask >>= 1)
    val += __shfl_xor_sync(lane_mask, val, mask, warpSize);
  return val;
}

template <typename T>
__inline__ __device__ T blockReduceSum(T val, unsigned mask) {
  static __shared__ T shared[WARP_SIZE];
  int lane = threadIdx.x & 0x1f;
  int wid = threadIdx.x >> 5;
  val = warpReduceSum<T>(val, mask);
  if (lane == 0) shared[wid] = val;
  __syncthreads();
  // align block_span to warpSize
  int block_span = (blockDim.x + warpSize - 1) >> 5;
  val = (lane < block_span) ? shared[lane] : static_cast<T>(0.0f);
  val = warpReduceSum<T>(val, mask);
  return val;
}

__global__ void foo(const int* src) {
    int tid = blockIdx.x * blockDim.x + threadIdx.x;
    int val = src[tid];
    int sum = 0;
    sum = blockReduceSum(val, 0xFFFFFFFF);
    printf("flatten thread Idx:%d, sum:%d\n", tid, sum);
}

int main() {
  int* ptr;
  cudaMallocManaged(&ptr, WARP_SIZE*NB_WARP*sizeof(int));
  for (int i = 0; i <  WARP_SIZE * NB_WARP; ++i) {
    ptr[i] = 1;
  }
  foo<<<1,  WARP_SIZE * NB_WARP>>>(ptr);
  cudaDeviceSynchronize();
  return 0;
}
```

build:
```bash
$ nvcc foo.cu -o reduce_demo
```

run:
```bash
$ ./reduce_demo
```

PR #24841, opened 2020-06-02
 [Dy2Static]Support return variable created in only one of If.body or If.orelse

liym27:
- PR types: New features

- PR changes: dygraph to static graph

- Describe:
#### 2. Support return variable in only one of If.body or If.orelse node.
E.g.
```Python
x, y = 5, 10
if x > 4:
    x = x+1
    z = x*x
    q = 10
else:
    y = y - 1
    z = y*y
    m = 20
    n = 20

print(q)
n = 30
print(n)
```
The return_ids are (x, y, z, q) for `If.body` and `If.orelse`node, because
1. `x` is modified in If.body node,
2. `y` is modified in If.body node,
3. `z` is both created in If.body and If.orelse node,
4. `q` is created only in If.body, and it is used by `print(q)` as gast.Load.
Note:
After transformed, q is created in parent scope. For example,
```Python
x, y = 5, 10
q = fluid.dygraph.dygraph_to_static.variable_trans_func.data_layer_not_check(name='q', shape=[-1], dtype='float32')
z = fluid.dygraph.dygraph_to_static.variable_trans_func.data_layer_not_check(name='z', shape=[-1], dtype='float32')
def true_func(x, y, q):
    x = x+1
    z = x*x
    q = 10
    return x,y,z,q

def false_func(x, y, q):
    y = y - 1
    z = y*y
    m = 20
    n = 20
    return x,y,z,q
x,y,z,q = fluid.layers.cond(x>4, lambda: true_func(x, y), lambda: false_func(x, y, q))
```
m and n are not in return_ids, because
5. `m` is created only in If.orelse, but it is not used after gast.If node.
6. `n` is created only in If.orelse, and it is used by `n = 30` and `print(n)`, but it is not used as gast.Load firstly but gast.Store .

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -542,10 +542,12 @@ def parse_cond_return(parent_vars_dict, if_vars_dict, else_vars_dict,
     3. z is both created in If.body and If.orelse node,
     4. q is created only in If.body, and it is used by `print(q)` as gast.Load.
     Note:
-        After transformed, q is created in parent scope. For example,
+        After transformed, q and z are created in parent scope. For example,
 
         x, y = 5, 10
         q = fluid.dygraph.dygraph_to_static.variable_trans_func.data_layer_not_check(name='q', shape=[-1], dtype='float32')
+        z = fluid.dygraph.dygraph_to_static.variable_trans_func.data_layer_not_check(name='z', shape=[-1], dtype='float32')
+
         def true_func(x, y, q):
             x = x+1
             z = x*x
@@ -585,7 +587,7 @@ def parse_cond_return(parent_vars_dict, if_vars_dict, else_vars_dict,
             var for var in _vars_with_store(child_dict) if var in parent_dict
         ])
 
-    def _vars_before_store(ids_dict):
+    def _vars_loaded_before_store(ids_dict):
         new_dict = defaultdict(list)
         for k, ctxs in ids_dict.items():
             for ctx in ctxs:
@@ -609,22 +611,22 @@ def parse_cond_return(parent_vars_dict, if_vars_dict, else_vars_dict,
         var for var in _vars_with_store(else_vars_dict)
         if var not in parent_vars_dict
     ])
-    new_vars_in_if_or_else = body_new_vars | orelse_new_vars
-    new_vars_in_one_of_if_or_else = body_new_vars ^ orelse_new_vars
+    new_vars_in_body_or_orelse = body_new_vars | orelse_new_vars
+    new_vars_in_one_of_body_or_orelse = body_new_vars ^ orelse_new_vars
 
     # 1. the var in parent scope is modified in If.body or If.orelse node.
-    modified_vars_from_parent = modified_vars - new_vars_in_if_or_else
+    modified_vars_from_parent = modified_vars - new_vars_in_body_or_orelse
 
     # 2. new var is both created in If.body and If.orelse node.
     new_vars_in_body_and_orelse = body_new_vars & orelse_new_vars
 
     # 3. new var is created only in one of If.body or If.orelse node, and it used as gast.Load firstly after gast.If node.
     used_vars_after_ifelse = set(
-        [var for var in _vars_before_store(after_ifelse_vars_dict)])
-    new_vars_to_create = new_vars_in_one_of_if_or_else & used_vars_after_ifelse
+        [var for var in _vars_loaded_before_store(after_ifelse_vars_dict)])
+    new_vars_to_create = new_vars_in_one_of_body_or_orelse & used_vars_after_ifelse | new_vars_in_body_and_orelse
 
     # 4. generate return_ids of if/else node.
-    return_ids = list(modified_vars | new_vars_in_body_and_orelse |
+    return_ids = list(modified_vars_from_parent | new_vars_in_body_and_orelse |
                       new_vars_to_create)
     return_ids.sort()
 
@@ -635,6 +637,7 @@ def transform_if_else(node, root):
     """
     Transform ast.If into control flow statement of Paddle static graph.
     """
+    # TODO(liym27): Consider variable like `self.a` modified in if/else node.
     parent_name_ids = get_name_ids([root], end_node=node)
     body_name_ids = get_name_ids(node.body)
     orelse_name_ids = get_name_ids(node.orelse)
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -71,14 +69,14 @@ def visit_If(self, node):
...
-                                        true_func_node, false_func_node)
-            self.new_func_nodes[new_node] = [true_func_node, false_func_node
-                                             ] + new_assign_nodes
-            return new_node
+            cond_node = create_cond_node(return_name_ids, pred_node,
+                                         true_func_node, false_func_node)
+
+            return new_vars_stmts + [true_func_node, false_func_node
+                                     ] + new_assign_nodes + [cond_node]
         else:
```
Aurelius84, line 80: Great! Thanks for polishing these codes.

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -587,57 +580,108 @@ def _vars_with_store(ids_dict):
...
+    after_ifelse_name_ids = defaultdict(list)
+    all_name_ids = get_name_ids([root])
+    for name in all_name_ids:
+        before_var_nodes = parent_name_ids.get(name, []) + \
+                           body_name_ids.get(name, []) + orelse_name_ids.get(name, [])
+        after_var_nodes = [
+            ctx for ctx in all_name_ids[name] if ctx not in before_var_nodes
+        ]
+        if after_var_nodes:
+            after_ifelse_name_ids[name] = after_var_nodes
```
Aurelius84: `before_var_nodes ` holds `ctx.Load()` or `ctx.Store()` info, which need a better name to avoid confusing other developers.
Could you add a note to explain that `context of node.Name like ctx.Load() is a concrete object which has unique id with other ctx.Load().`

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -587,57 +580,108 @@ def _vars_with_store(ids_dict):
...
+    # E.g.
+    #
+    # if x > 5:
+    #   a = 10
+    # print(a)
+    #
+    # Create static variable for those variables
+    create_new_vars_in_parent_stmts = []
+    for name in new_vars_to_create:
+        if "." not in name:
```
Aurelius84, line 673: Why we need add `if . not in name` here?

Aurelius84 (commented):
Great work! LGTM for the rest.

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -553,25 +510,61 @@ def parse_cond_args(var_ids_dict, return_ids=None, ctx=gast.Load):
...
+def parse_cond_return(parent_vars_dict, if_vars_dict, else_vars_dict,
+                      after_ifelse_vars_dict):
     """
     Find out the ast.Name list of output by analyzing node's AST information.
-    Following conditions should be satisfied while determining whether a variable is a return value:
-    1. the var in parent scope is modified in if/else node.
-    2. new var is both created in if and else node.
+    One of the following conditions should be satisfied while determining whether a variable is a return value:
+    1. the var in parent scope is modified in If.body or If.orelse node.
+    2. new var is both created in If.body and If.orelse node.
```
chenwhql, line 519: if new var is both created in If.body and If.orelse node, but it is not used after gast.If node, we also need it?

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -587,57 +580,108 @@ def _vars_with_store(ids_dict):
...
         var for var in _vars_with_store(if_vars_dict)
         if var not in parent_vars_dict
     ])
-    else_new_vars = set([
+    orelse_new_vars = set([
         var for var in _vars_with_store(else_vars_dict)
         if var not in parent_vars_dict
     ])
-    new_vars = if_new_vars & else_new_vars
+    new_vars_in_if_or_else = body_new_vars | orelse_new_vars
```
chenwhql: I understand the meaning here, but this variable names may be confusing, this `or` is not its literal meaning, maybe `new_vars_in_ifbody_and_orelse` is better? oh...this is a little long

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
```diff
@@ -587,57 +580,108 @@ def _vars_with_store(ids_dict):
...
             var for var in _vars_with_store(child_dict) if var in parent_dict
         ])
 
-    # 1. the var in parent_ids is modified in if/else node.
-    if_candidate_vars = _candidate_vars(if_vars_dict, parent_vars_dict)
-    else_candidate_vars = _candidate_vars(else_vars_dict, parent_vars_dict)
-
-    # 2. new var is both created in if and else node.
-    if_new_vars = set([
+    def _vars_before_store(ids_dict):
```
chenwhql: maybe `_vars_loaded_before_store` is better?

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
liym27: Here we doesn't need it. But we need to add `if . not in name` if we consider variable like `self.a` modified in if/else node.
I add a note here.

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
liym27: Thanks. Done

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
liym27: Thanks, done

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
liym27: Yes. A var is both created in If.body and If.orelse node should be visible.

python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
liym27: Replace it with `new_vars_in_body_or_orelse `

Aurelius84 (approved):
LGTM

chenwhql (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #24853, opened 2020-06-02
Fix bug in ShareBufferWith causing non-fp32 inplace uts to fail

Sand3r-:
PR types: Bug fixes

PR changes: APIs

Describe: During call to ShareBufferWith, the type wasn't assigned when holders were assigned. This causes inplace UTs that use non-fp32 tensors (for example int8, uint8) to fail on FetchOp.


jczaja (approved):
LGTM

luotao1 (approved):

PR #24854, opened 2020-06-02
Fix OpTester from incorrectly converting uint8 data to utf-8 string

Sand3r-:
PR types: Bug fixes

PR changes: APIs

Describe: When uint8 datatype is to be outputted in the message by `_compare_expect_and_actual_outputs`, direct cast using str(), causes to throw an error related to encoding (data is incorrectly treated as array of characters instead of array of numbers). To mitigate this, the data is cast to np.array which then can be safely cast to string.


luotao1:
Please trigger the CI

Sand3r-:
@luotao1 All CIs have passed

lidanqing-vv:
@luotao1  Could this PR be merged ?

jczaja (approved):
LGTM

luotao1 (approved):

PR #24858, opened 2020-06-02
add slot_padding_zero, test=develop

ForFishes:
PR types: Function optimization  

PR changes: Others

Describe: Add options for set_slot_padding_zero, its default is true. When it set false, it will rm 0 slot.

### DIFF
--- a/paddle/fluid/framework/data_feed.cc
+++ b/paddle/fluid/framework/data_feed.cc
@@ -236,6 +236,7 @@ InMemoryDataFeed<T>::InMemoryDataFeed() {
   this->parse_logkey_ = false;
   this->enable_pv_merge_ = false;
   this->current_phase_ = 1;  // 1:join ;0:update
+  this->slot_padding_zero_ = true;
   this->input_channel_ = nullptr;
   this->output_channel_ = nullptr;
   this->consume_channel_ = nullptr;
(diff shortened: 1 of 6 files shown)


paddle/fluid/framework/data_feed.cc
```diff
@@ -357,6 +358,11 @@ void InMemoryDataFeed<T>::SetCurrentPhase(int current_phase) {
   current_phase_ = current_phase;
 }
 
+template <typename T>
+void InMemoryDataFeed<T>::SetSlotPaddingZero(bool slot_padding_zero) {
```
hutuxian: where to call this function?

paddle/fluid/operators/sequence_ops/sequence_pool_op.h
```diff
@@ -43,10 +43,6 @@ class SequencePoolKernel : public framework::OpKernel<T> {
                                         "does not contain LoD information."));
     PADDLE_ENFORCE_LE(lod_level, 2UL,
                       "The lod level of input shall be no more than 2.");
-    PADDLE_ENFORCE_GE(
```
hutuxian, line 46: need confirmation of this op's owner

paddle/fluid/framework/data_feed.cc
ForFishes: i will delete it.

paddle/fluid/operators/sequence_ops/sequence_pool_op.h
ForFishes: Ok. Done.

PR #24875, opened 2020-06-03
Feature/add amp_checkout_finite_and_scale op

zhiqiu:
PR types: New features

PR changes: OPs

Describe: Add amp_checkout_finite_and_scale op

When doing mixed-precision training / inferring in imperative mode, we will first scale the loss to `scale * loss` and perform `backward`. After that, we need to check if the gradients of each parameter contains NAN or INF. If not, the gradients are unscaled to `loss * (1 / scale)` and the parameters are updated by gradients. If yes, the gradients are discarded and the parameters will not updated in this step.

In order to achieve high performance, reduce CUDA synchronization and Python/C++ interaction, Operator `amp_checkout_finite_and_scale` is added to check if a list of gradients (tensors) contain NAN or INF, and scale them if not contain.

### DIFF
--- a/paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
+++ b/paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
@@ -36,7 +36,9 @@ class AmpCheckFiniteAndScaleOp : public framework::OperatorWithKernel {
         ctx->Inputs("X").size(), ctx->Outputs("Out").size(),
         platform::errors::InvalidArgument(
             "The input(X) and output(Out) should have same size in "
-            "Operator(amp_check_finite_and_unscale)."));
+            "Operator(amp_check_finite_and_unscale), size of input(X) is %d "
+            "and size of output(Out) is %d.",
+            ctx->Inputs("X").size(), ctx->Outputs("Out").size()));
     auto x_dims = ctx->GetInputsDim("X");
     ctx->SetOutputsDim("Out", x_dims);
     ctx->SetOutputDim("FoundInfinite", {1});
@@ -55,13 +57,13 @@ class AmpCheckFiniteAndScaleOpMaker : public framework::OpProtoAndCheckerMaker {
   void Make() override {
     AddInput(
         "X",
-        "(Tensor) The input tensors of amp_check_finite_and_scale operator.")
+        "(Tensors) The input tensors of amp_check_finite_and_scale operator.")
         .AsDuplicable();
     AddInput("Scale",
              "(Tensor) 1-dim tensor, the scale of amp_check_finite_and_scale "
              "operator.");
     AddOutput("Out",
-              "(Tensor) The scaled output tensor of "
+              "(Tensors) The scaled output tensor of "
               "amp_check_finite_and_unscale operator.")
         .AsDuplicable();
     AddOutput("FoundInfinite",
@@ -78,7 +80,6 @@ FoundInfinite will be 1 (True), and Out will not be scaled. In this case, the da
 Out should not be used, and its data may not be deterministic. 
 Otherwise, FoundInfinite will be 0 (False).
 
-%s
 )DOC");
   }
 };
--- a/paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
+++ b/paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
@@ -1,0 +1,100 @@
+/* Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */
+
+#include "paddle/fluid/operators/amp/amp_check_finite_and_scale_op.h"
+#include <string>
+#include <vector>
+
+namespace paddle {
+namespace operators {
+
+class AmpCheckFiniteAndScaleOp : public framework::OperatorWithKernel {
+ public:
+  AmpCheckFiniteAndScaleOp(const std::string &type,
+                           const framework::VariableNameMap &inputs,
+                           const framework::VariableNameMap &outputs,
+                           const framework::AttributeMap &attrs)
+      : OperatorWithKernel(type, inputs, outputs, attrs) {}
+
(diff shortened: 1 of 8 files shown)


python/paddle/fluid/tests/unittests/test_amp_check_finite_and_scale_op.py
```diff
@@ -0,0 +1,97 @@
...
+    def init_dtype(self):
+        self.dtype = np.float32
+
+    def test_check_output(self):
+        self.check_output()
+
+
+@skip_check_grad_ci(
+    reason="Operator amp_check_finite_and_scale is used for dygraph auto-mixed_precision training only. It has no grad op."
+)
```
zhangting2020: First it will be checked if there is a grad op. If no grad op, the unittest will not be required checking grad. So there is no need to use skip_check_grad_ci.

cryoco (dismissed):
LGTM with no_check_set.

python/paddle/fluid/tests/unittests/test_amp_check_finite_and_scale_op.py
zhiqiu: Done, thanks.

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
```diff
@@ -0,0 +1,102 @@
...
+
+  void InferShape(framework::InferShapeContext *ctx) const override {
+    OP_INOUT_CHECK(ctx->HasInputs("X"), "Input", "X",
+                   "amp_check_finite_and_unscale");
+    OP_INOUT_CHECK(ctx->HasOutputs("Out"), "Output", "Out",
+                   "amp_check_finite_and_unscale");
+    PADDLE_ENFORCE_EQ(
+        ctx->Inputs("X").size(), ctx->Outputs("Out").size(),
+        platform::errors::InvalidArgument(
+            "The input(X) and output(Out) should have same size in "
```
chenwhql, line 38: tell users received input and output size is better

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
```diff
@@ -0,0 +1,102 @@
...
+  }
+};
+
+class AmpCheckFiniteAndScaleOpMaker : public framework::OpProtoAndCheckerMaker {
+ public:
+  void Make() override {
+    AddInput(
+        "X",
+        "(Tensor) The input tensors of amp_check_finite_and_scale operator.")
+        .AsDuplicable();
```
chenwhql, line 61: why `AsDuplicable`, but type is Tensor? is it vector<Tensor>?

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
```diff
@@ -0,0 +1,102 @@
...
+Check if input X contains all finite data, if yes, scale it by input Scale.
+
+$$Out = X * scale$$
+
+If any tensor in X contains Inf or Nan, the Out will generate a indicator.
+FoundInfinite will be 1 (True), and Out will not be scaled. In this case, the data of 
+Out should not be used, and its data may not be deterministic. 
+Otherwise, FoundInfinite will be 0 (False).
+
+%s
```
chenwhql: why add `%s` here?

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.h
```diff
@@ -0,0 +1,66 @@
...
+template <typename DeviceContext, typename T>
+class AmpCheckFiniteAndScaleKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& ctx) const {
+    auto& dev_ctx = ctx.template device_context<DeviceContext>();
+    const auto xs = ctx.MultiInput<framework::Tensor>("X");
+    const auto* scale = ctx.Input<framework::Tensor>("Scale");
+    auto outs = ctx.MultiOutput<framework::Tensor>("Out");
+    auto* found_inf = ctx.Output<framework::Tensor>("FoundInfinite");
+
```
chenwhql, line 35: Is it possible that the tensor accessed here is uninitialized and whether some security checks are required?

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
```diff
@@ -0,0 +1,102 @@
...
+limitations under the License. */
+
+#include "paddle/fluid/operators/amp/amp_check_finite_and_scale_op.h"
+#include <string>
+#include <vector>
+
+namespace paddle {
+namespace operators {
+
+class AmpCheckFiniteAndScaleOp : public framework::OperatorWithKernel {
```
chenwhql, line 22: `Amp` is not a word, do we need change it to `AMP`?

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.h
zhiqiu: To my knowledge, inputs and outputs of an operator are prepared by framework, for example, in imperative mode, `PrepareData` is called which checks variable is initialized and `tensor.data<T>` checks the tensor and raise exception if tensor is not initialized. 
If the an input or output is not `dispensable`, it is not necessary to check each input or output variable or tensor, otherwise, the op kernel will be bloated.

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
zhiqiu: Removed, thx.

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
zhiqiu: According to Google coding style, `Amp` is recommended instead of `AMP`, even though `AMP` is short for `Auto-Mxied-Precision`. 
Ref: https://zh-google-styleguide.readthedocs.io/en/latest/google-cpp-styleguide/naming/

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
zhiqiu: Changed to `Tensors`, thx.

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
zhiqiu: Done, thanks.

paddle/fluid/operators/amp/amp_check_finite_and_scale_op.cc
chenwhql: Great

chenwhql (approved):
LGTM

cryoco (approved):

phlrain (approved):

PR #24901, opened 2020-06-04
[Dy2Static]convert for stmt and support variable loaded and created in loop

liym27:
### PR types: 
New features

### PR changes: 
dygraph to static graph

#### 1. New feature 1
Use function `convert_while_loop` to run the transformed `for loop` code dynamically:
- `for` stmts  are transformed whether it depends on `Tensor` or not
        1). for x in range(var[*]|var.numpy()[*])
        2). for x in var|var.numpy()
        3). for i, x enumerate(var|var.numpy())
    >Note:  Here var is VarBase(Tensor) or python variable

- But in run-time, the code will be run in different ways (1) run paddle op like `while_loop`; (2) run python while. How it runs depends on whether it depends on `Tensor`. 

#### 2. New feature:  
support that case : variable is used and created in loop, but used before created.
For example, `var_a` should be in loop_var and we should create it out loop.

```Python
res = 0
for i, x in enumerate(x_array):
      if i > 2:
          x = func1(var_a)
      var_a = func2(x)
```

#### 3. Fix bug: 
Fix bug: raise None -> return None. 

#### 4. Code optimization: 
Move `convert_len` to file `convert_operators.py`


python/paddle/fluid/dygraph/dygraph_to_static/utils.py
```diff
@@ -700,12 +700,11 @@ def __init__(self, for_node):
...
         self.iter_idx_name = unique_name.generate(FOR_ITER_INDEX_PREFIX)
 
-        # - created shape var to build loop condition: __for_loop_var_shape_0
+        # - created shape var to build loop condition: __for_loop_var_len_0
         #   - for x in var|var.numpy()
         #   - for i, x enumerate(var|var.numpy())
         #   - for x in var
-        self.iter_var_shape_name = unique_name.generate(
-            FOR_ITER_VAR_SHAPE_PREFIX)
+        self.iter_var_shape_name = unique_name.generate(FOR_ITER_VAR_LEN_PREFIX)
```
chenwhql: `iter_var_shape_name` -> `iter_var_len_name`?

chenwhql (commented):
Great!

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
liym27: Thanks, done

Aurelius84 (approved):
LTGM

zhhsplendid (approved):

PR #24911, opened 2020-06-04
[Dy2Static]Convert var.shape stmt and Convert the return variables of Tensor-dependent 'if' staments to Tensor if it not 

liym27:
### PR types

New features

### PR changes

Others

### Describe

- ### Describe:
#### 1. New feature: Use function `convert_var_shape` to run the transformed `var.shape` code dynamically:
- `var.shape` stmts  are transformed whether it depends on `Tensor` or not
- But in run-time, the code will be run in different ways (1) run paddle op `layers.shape(var)`; (2) run python var.shape. How it runs depends on whether `var` is `Tensor`. 

For example:
- Before:
```
res = fluid.layers.reshape(x, shape=x.shape)
```
- After
```
res = fluid.layers.reshape(x, shape=fluid.dygraph.dygraph_to_static.
        convert_operators.convert_var_shape(x))
```
#### 2. New feature : Convert the return variables of Tensor-dependent 'if' staments to Tensor if it not.
For example:

- Before transformed:
```Python
def func1(x_v):
    batch_size = 16
    if x_v.shape[0] != batch_size:
        batch_size = fluid.layers.shape(x_v)[0]
     # do something else
```

- After transformed:
In `false_fn_0`, batch_size is integer value 16, but the return variables of `false_fn` of `layers.cond` must be Tensor. This PR convert `batch_size` to Tensor before run `layers.cond`
```Python
def nested_if_else(x_v):
   batch_size = 16

   def true_fn_0(batch_size, x_v):
        batch_size = fluid.layers.shape(x_v)[0]
        return batch_size

   def false_fn_0(batch_size):
        return batch_size  # NOTE here
   
    batch_size = fluid.dygraph.dygraph_to_static.convert_operators.convert_ifelse(
    fluid.dygraph.dygraph_to_static.convert_operators.convert_var_shape(x_v)[0] != batch_size, 
    true_fn_0, 
    false_fn_0, 
    (batch_size, x_v), 
    (batch_size,), 
    (batch_size,))

```

#### 3. Fix bug: 
- Support int and long: int or long -> six.integer_types.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
@@ -43,6 +43,7 @@ def convert_while_loop(cond, body, loop_vars):
 
 
 def _run_paddle_while_loop(cond, body, loop_vars):
+    # NOTE: loop_vars of Paddle op `control_flow.while_loop` must be Paddle Variable.
     loop_vars = [to_static_variable(var) for var in loop_vars]
     loop_vars = control_flow.while_loop(cond, body, loop_vars)
     return loop_vars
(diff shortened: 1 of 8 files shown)


Aurelius84 (approved):
Great! LGTM

zhhsplendid (approved):
LGTM

PR #24913, opened 2020-06-04
fix LinearLrWarmup bug

phlrain:
### PR types
Bug fixes

### PR changes
APIs

### Describe
fix liear lr warmup bug. 

change return value from
self.lr_ratio_before_warmup * self.step_num
to
self.lr_ratio_before_warmup * self.step_num + self.start_lr


zhiqiu (approved):
LGTM

PR #24923, opened 2020-06-05
Add support the 5d, 6d tensor support for the reduce ops

wawltor:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Add the support the 5d,6d tensor support for the reduce ops;
Add the same time,  the compile time,  it was 22 minutes, it was 21 minutes after fixed.


ZeyuChen (approved):
LGTM

PR #24924, opened 2020-06-05
add default ctor for AnalysisConfig python api

cryoco:
### PR types

Function optimization
### PR changes

APIs
### Describe

Added default constructor for `AnalysisConfig` in python api for the use of `set_model_buffer`


fc500110 (approved):
LGTM

PR #24929, opened 2020-06-05
fix WARNING: ThreadSanitizer: heap-use-after-free

sfraczek:
### PR types
Bug fixes

### PR changes
Others

### Describe
This fix removes risk of reading deleted memory in theFC Elementwise Layernorm Fuse Pass.
If `layer_norm_mean` and variance can be deleted in GraphSafeRemoveNodes, then we shouldn't access its members afterwards. I saw warning from ThreadSanitizer so I just fixed it.


wojtuss (approved):
LGTM

PR #24944, opened 2020-06-05
[Dy2stat] Add word2vec as unittest

zhhsplendid:
### PR types
Others

### PR changes
Others

### Describe
Add word2vec as ProgramTranslator unit test。
The word2vec model is referred from https://github.com/PaddlePaddle/models/blob/856c428fe0f1f451f8a22acee50c2e3af33c834e/dygraph/word2vec/word2vec.py#L238

### DIFF
--- a/python/paddle/fluid/tests/unittests/dygraph_to_static/test_word2vec.py
+++ b/python/paddle/fluid/tests/unittests/dygraph_to_static/test_word2vec.py
@@ -1,0 +1,325 @@
+# Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+import os
+import sys
+import requests
+from collections import OrderedDict
+import math
+import random
+import numpy as np
+import paddle
+import paddle.fluid as fluid
+import unittest
+
+from paddle.fluid.dygraph.nn import Embedding
+from paddle.fluid.dygraph import ProgramTranslator
+from paddle.fluid.dygraph import declarative
+
+
+def fake_text():
+    corpus = []
+    for i in range(100):
+        line = "i love paddlepaddle"
+        corpus.append(line)
+    return corpus
+
+
+corpus = fake_text()
+
+
+def data_preprocess(corpus):
+    new_corpus = []
+    for line in corpus:
+        line = line.strip().lower()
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/tests/unittests/dygraph_to_static/test_word2vec.py
```diff
@@ -0,0 +1,325 @@
...
+def get_similar_tokens(query_token, k, embed):
+    W = embed.numpy()
+    x = W[word2id_dict[query_token]]
+    cos = np.dot(W, x) / np.sqrt(np.sum(W * W, axis=1) * np.sum(x * x) + 1e-9)
+    flat = cos.flatten()
+    indices = np.argpartition(flat, -k)[-k:]
+    indices = indices[np.argsort(-flat[indices])]
+    for i in indices:  # Remove the input words
+        print('for word %s, the similar word is %s' %
+              (query_token, str(id2word_dict[i])))
```
Aurelius84: It seems that `get_similar_tokens ` is not used in the unittest.

Aurelius84 (commented):
LGTM for the rest.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_word2vec.py
```diff
@@ -0,0 +1,325 @@
...
+                eval_word_batch = []
+
+    if len(center_word_batch) > 0:
+        yield np.array(center_word_batch).astype("int64"), np.array(
+            target_word_batch).astype("int64"), np.array(label_batch).astype(
+                "float32"), np.array(eval_word_batch).astype("int64")
+
+
+#for _, batch in zip(range(10), build_batch(dataset, 128, 3)):
+#    print(batch)
```
liym27: The same as above

python/paddle/fluid/tests/unittests/dygraph_to_static/test_word2vec.py
```diff
@@ -0,0 +1,325 @@
...
+def build_batch(dataset, batch_size, epoch_num):
+
+    center_word_batch = []
+    target_word_batch = []
+    label_batch = []
+    eval_word_batch = []
+
+    for epoch in range(epoch_num):
+
+        #random.shuffle(dataset)
```
liym27: Is this commented code necessary?

zhhsplendid: No, the commented code is in original code. I commented because the random.shuffle can cause the input to dygraph/static be different thus our check can fail.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_word2vec.py
zhhsplendid: This is copied code so I didn't care much about this. The get_similar_tokens is used in the original code but it now cannot run because of API compatibility issue. I will remove it in this unittest.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_word2vec.py
zhhsplendid: The code is at the original code. I commented it so that we don't print much info. I will remove line 208-209.

Aurelius84 (approved):

PR #24946, opened 2020-06-05
Fix test layers unitest error

phlrain:
### PR types
Bug fixs

### PR changes
Others

### Describe
修复 test_layers 在cpu下的错误
test_layers 在cpu下，动态图和静态图结果不一致，分析发现是动态图中用到了 zero_copy的问题，设置zero_copy 等于 false


zhiqiu (approved):
LGTM for fixing `test_layers`.
We should find out why it fails when `zero_copy=True`.

chalsliu (approved):

PR #24965, opened 2020-06-08
Fix the elementwise_div op broadcast failed in some shape

wawltor:
### PR types
 Bug fixes

### PR changes
 OPs

### Describe
Fix the bug for elementwise_div op, when the first var is scalar; Use the shape 1 replace the -1 in shape.


jerrywgz (approved):
LGTM

PR #24970, opened 2020-06-08
Reshape transpose matmul coverage

sfraczek:
### PR types
Bug fixes

### PR changes
Others

### Describe
Enable unit test of reshape+transpose+matmul fuse pass with COVERAGE build flag by removing dependency on gmock. I found that gmock dependency in fuse UTs introduces weird mutex errors and this is the fix.


jczaja (approved):
LGTM

luotao1 (approved):

PR #24975, opened 2020-06-09
Fix the bug that Input(Offsets) and attr(offsets) cannot be set at the same time.

sandyhouse:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix the bug that Input(Offsets) and attr(offsets) cannot be set at the same time. 
The python side set them both (as the following figure shows), and Input(Offsets) has a high priority， thus we can remove the ENFORCE_XX statement in the source codes.
image


zhangting2020 (approved):
LGTM. 因为参数要支持Tensor，但是同时需要在compile time推断出shape，因此就必须同时用Offsets(Input)和attr offsets.  attr offsets用于shape推断。

PR #25000, opened 2020-06-10
fix datanorm op add attr bug test=develop

yaoxuefeng6:
### PR types
Bug fixes

### PR changes
OPs

### Describe
fix datanorm op add attr if the program does not have the attr, add default attr in grad op


fuyinno4 (dismissed):

zhiqiu (approved):
LGTM, data_norm is used in Static graph

PR #25017, opened 2020-06-11
fix cos_sim nan

ceci3:
### PR types
Bug fixes

### PR changes
OPs

### Describe
add eps = 1e-8 in cos_sim op to prevent nan


qingqing01 (approved):
LGTM.

PR #25069, opened 2020-06-16
Add api to clear intermediate tensors in AnalysisPredictor

cryoco:
### PR types

Function optimization
### PR changes

APIs
### Describe

On some devices, memory/graphic-memory is limited. Instead of running paddle inference parallelly, we have to load, analyze, and run them one after another(with each one's resources released after running) on these devices to avoid OOM. In such situations, loading and analysis phases become bottleneck of latency.
So we add an API to clear intermediate tensors, enabling AnalysisPredictor to load and analyze all models first, and run inference one by one(with each one's intermediate tensors released to graphic memory pool by calling ClearIntermediateTensor()) to cut the overhead mentioned above.

### DIFF
--- a/paddle/fluid/inference/api/analysis_predictor.cc
+++ b/paddle/fluid/inference/api/analysis_predictor.cc
@@ -828,6 +828,25 @@ bool AnalysisPredictor::LoadParameters() {
   return true;
 }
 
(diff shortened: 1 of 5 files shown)


NHZlX:
python api 需要添加下

NHZlX (approved):
LGTM

PR #25072, opened 2020-06-16
[Dy2stat] Add CycleGAN model for unitttest

Aurelius84:
### PR types

Others

### PR changes

Others

### Describe

+ Add CycleGAN model for @declarative unitttest
+ Prune unnecessary `parameters`in main_program.
> When decorate the sub functions with `@declarative`, the new `main_program` in `partial_program` may not use all parameters created in `__init__`. The unused parameters should be pruned to avoid redundant operatorations in `run_program_op`.

#### TODO
+ Cycle GAN has diff on GPU but it doesn't come from Dy2stat. The model has diff if we just train it in dygraph mode on GPU. I'm working on it and will fix it in next PR.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
@@ -110,7 +110,7 @@ class PartialProgramLayer(layers.Layer):
         self._inputs = NestSequence(inputs)
         self._outputs = NestSequence(outputs, need_check=True)
         self._params = parameters if parameters is not None else []
-        
+
         self._prune_unused_params(main_program)
         # Check all params from main program can be found in self._params:
         # 1. parameter in self._params should be type `framework.ParamBase` which are created in dygraph.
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
```diff
@@ -138,6 +140,23 @@ def _append_backward_desc(self):
 
         return program
 
+    def _prune_unused_params(self, program):
+        """
+        Prune the parameters not used anywhere in the program.
+        The `@declarative` may only decorated a sub function which
```
zhhsplendid, line 146: Nit: "decorated" -> decorate

zhhsplendid (approved):

python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
Aurelius84: Thanks, I will fix it with GPU problem.

PR #25085, opened 2020-06-16
[Dy2stat]Refine caches of converting function

Aurelius84:
### PR types

Function optimization
### PR changes

Others
### Describe

+ Refine caches of converting function.
>  For the function with same detent source code, only once convert into ast nodes  once and cache them.
> For two function with same dedent code,  the second function will reuse the transformed ast node and save a new tmp file and reload a independent module because they can hold difference global variables in their located files.

+ clean useless code related with `feed_name_id`


python/paddle/fluid/dygraph/dygraph_to_static/utils.py
```diff
@@ -407,9 +408,24 @@ def recover_globals_attribute(src_obj, dst_obj):
...
+        raise TypeError(
+            "The type of 'function' should be a function or method, but received {}.".
+            format(type(function).__name__))
+    source_code = inspect.getsource(function)
+    if dedent:
+        source_code = textwrap.dedent(source_code)
+
+    return source_code
+
+
```
liym27, line 425: function `func_to_source_code` only called in `convert_to_static_with_cache` and too many functions in `utils.py`, how about moving it to file `program_translator.py`?

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
```diff
@@ -46,43 +47,60 @@ class FunctionCache(object):
...
+            root_wrapper = self._dygraph_to_static.get_static_ast(root)
+            self._code_to_ast_caches[source_code] = root_wrapper
 
-    def _get_dedent_code_string(self, func):
-        raw_code = inspect.getsource(func)
-        dedent_code = textwrap.dedent(raw_code)
-        return dedent_code
+        # Get static_func from AST
+        static_func, file_name = ast_to_func(root_wrapper.node, func)
+        return static_func
```
liym27, line 102: how about combining these two functions (`get_or_cache_func ` and `convert_to_static_with_cache ` ) into only one function to make logic clearer? And add comment as the follows:
> For the function with same detent source code, only once convert into ast nodes once and cache them.
For two function with same dedent code, the second function will reuse the transformed ast node and save a new tmp file and reload a independent module because they can hold difference global variables in their located files.

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
Aurelius84: `func_to_source_code` is a common function so I put it into `utils.py`. If put it into `program_translator.py` and used in the future, it may lead to cycle import problem.

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
Aurelius84: Add more comment to explain the function.

liym27 (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #25091, opened 2020-06-17
fix dtype error in retinanet_target_assgin example codes

FlyingQianMM:
### PR types
Others

### PR changes
APIs

### Describe
Example code of retinanet_target_assgin op raises dtype errors, input `gt_label` and `is_crowd` must be int32 variables but they are float32 variables before，so this bug need to be been fixed.

**Before**
image
image
image
image

**After**
image
image
image
image


Heeenrrry (approved):
LGTM

wawltor (approved):
LGTM

PR #25097, opened 2020-06-17
[Paddle-TRT] Better Paddle-TensorRT support for PaddleSlim quant models

cryoco:
### PR types

Function optimization
### PR changes

Others
### Remove restraints of quantizing extra ops for TRT inference 
While using Paddle-TRT int8 to preform inference with models trained by PaddleSlim quant-aware training, there are some restraints during the training phase: if there are ops like elementwise_add, pool2d, leaky_relu, etc. in the model, they must be quantized, or TRT will throw an error during inference. This damages the flexibility for the utilization of the int8 quantization. We added support to ensure that while training with PaddleSlim QAT, users doesn't have to add extra quant ops for inferencing with TensorRT, with little performance damaging.

### quant_conv2d_dequant_fuse_pass refactor
quant_conv2d_dequant_fuse_pass is used to extract input and output scales from quant/dequant op and set them into attributes of quantized op.
The former quant_conv2d_dequant_fuse_pass has problems as following:
1. Using a matrix-like structure to store IR nodes requires a threshold of the max number of quantized nodes preceded by the same quant node, which leads to serious problems when model structure is like:
image

2. Converting weights from int8 range to fp32 range happens in tensorrt_subgraph_pass, which means if some quantized op is outside trt subgraph, it's weight will not be converted. This might cause wrong results when trt subgraph can't cover all quantized nodes.

We refactored this pass by splitting the fusion into 2 phases, DeleteQuant Fuse and Dequant fuse, so that the threshold of branch nodes is not needed. Moreover, the range of quantized weights in conv/mul/fc is converted to fp32 in this pass instead of tensorrt_subgraph_pass, to produce the right result when trt subgraph can't cover all quantized nodes.

### DIFF
--- a/paddle/fluid/framework/ir/quant_conv2d_dequant_fuse_pass.cc
+++ b/paddle/fluid/framework/ir/quant_conv2d_dequant_fuse_pass.cc
@@ -24,6 +24,8 @@ namespace paddle {
 namespace framework {
 namespace ir {
 
+// Delete quant op before quantized ops, and set input scale in the attr of
+// quantized ops
 void DeleteQuant(ir::Graph* graph, Scope* scope,
                  const std::string& quant_type) {
   const std::string pattern_name = "delete_quant_fuse";
@@ -33,8 +35,12 @@ void DeleteQuant(ir::Graph* graph, Scope* scope,
                              ->assert_is_op_input(quant_type, "X")
                              ->AsInput();
 
+  // Create pattern
   patterns::DeleteQuantOpFuse pattern(gpd.mutable_pattern(), pattern_name);
   pattern(input_act_node, quant_type);
+
+  // extract input scale from quant op input to set it in attr of all quantized
+  // ops linked from it
   auto handler = [&](const GraphPatternDetector::subgraph_t& subgraph,
                      Graph* g) {
     PADDLE_ENFORCE_EQ(subgraph.count(input_act_node), true,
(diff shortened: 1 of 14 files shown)


paddle/fluid/inference/api/analysis_config.cc
```diff
@@ -281,11 +281,8 @@ void AnalysisConfig::Update() {
 
   if (use_tensorrt_) {
     pass_builder()->ClearPasses();
-    bool use_calib_int8 =
-        (tensorrt_precision_mode_ == AnalysisConfig::Precision::kInt8) &&
-        trt_use_calib_mode_;
     for (const auto &pass : kTRTSubgraphPasses) {
-      if (use_calib_int8 &&
+      if (tensorrt_precision_mode_ == AnalysisConfig::Precision::kInt8 &&
```
NHZlX, line 285: Delete the conv_bn_fuse_pass & fc_fuse_pass before trt_pass in pass_builder ?

paddle/fluid/framework/ir/quant_conv2d_dequant_fuse_pass.cc
```diff
@@ -24,159 +24,197 @@ namespace paddle {
...
-  const int kQuantizedOpOutOffset = 2;
-  const int kDequantOpOffset = 3;
-  const int kDequantOpOutOffset = 4;
-  const int kDequantOpWeightScaleOffset = 5;
-
-  if (dequant_type == "fake_channel_wise_dequantize_max_abs") {
-    kNumFields += 1;
-  }
-
+void DeleteQuant(ir::Graph* graph, Scope* scope,
```
NHZlX, line 29: Add comments for this pass.

cryoco: Done.

NHZlX (approved):
LGTM

luotao1 (approved):

PR #25117, opened 2020-06-18
[Paddle-TRT] SkipLayernorm vectorized memory optimization

zlsh80826:
### PR types
 Performance optimization

### PR changes
 OPs

### Describe

SkipLayernorm is an op spent time on moving data from global memory. So adding vector load/store when ``hidden % 2 == 0`` can improve the performance of this operation. The experiments show 25% improvement after adding this optimization.


jeng1220:
@NHZlX ,
Plz take a look

PR #25147, opened 2020-06-21
cat log to stdout when setting log_dir in launch

gavin1332:
### PR types
Others

### PR changes
Others

### Describe
Print the log of local rank 0 to `stdout` while using `paddle.distributed.launch` with `--log_dir` set.

`paddle.distributed.launch` creates multiple sub-processes and these sub-processes redirect their `stdout` to `log_dir` while it is set. The main process waits all sub-processes with no training log printed to the `stdout` which make a false appearance of hanging to users and is not friendly to monitor the progress of training if a user is not aware of `workerlog`.

So this PR prints the log of local rank 0 to `stdout` for easy inspecting of training status.

### DIFF
--- a/python/paddle/distributed/utils.py
+++ b/python/paddle/distributed/utils.py
@@ -21,6 +21,7 @@ import signal
 import copy
 import sys
 import subprocess
+import threading
 from contextlib import closing
 import socket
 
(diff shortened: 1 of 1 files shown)


python/paddle/distributed/utils.py
```diff
@@ -370,7 +371,28 @@ def start_local_trainers(cluster,
...
+                    stderr=subprocess.STDOUT)
+
+                class ShellTee(threading.Thread):
+                    def __init__(self, proc, fn):
+                        super(ShellTee, self).__init__()
+                        self.proc = proc
+                        self.fn = fn
+
+                    def run(self):
+                        for line in self.proc.stdout:
```
gongweibao: 这个地方不是很理解，为何要假设数据是一行一行的？直接read然后write到target是否可以？

gavin1332: 直接read我理解是需要待日志数据全部生产完毕以后才能write吧，这样不能实时打印日志会导致日志的时效性较差？

python/paddle/distributed/utils.py
```diff
@@ -370,7 +371,28 @@ def start_local_trainers(cluster,
...
+                            sys.stdout.write(line.decode(sys.stdout.encoding))
+                            fn.write(line.decode(sys.stdout.encoding))
```
gavin1332: As gongweibao suggested, I will update these code to read and write binary.

gongweibao (approved):
LGTM

PR #25176, opened 2020-06-23
[Dy2stat] Add Basic Support for Grammar 'return'

zhhsplendid:
### PR types

New features

### PR changes

Others

### Describe

This PR added basic support for 'return' grammar in dy2stat. It supports the control flow of 'return'.

The basics idea is using a return value variable to store the early return statements and boolean state variables with if-else to skip the statements after the return statements.

**This PR is very basic support. There are some corner cases I didn't develop/test**. For example, 'return None', 'return different length of variables', 'return non-tensor and tensor together', 'no return statement'. **These corner cases will be done in my next PRs**. Target date is this week.

**Note**: 
1. for the unit test, I changed test_program_translator.py because the StaticCode of `dyfunc_with_if_else` will change. To guarantee the correctness of `dyfunc_with_if_else`, I also run it in `TestRecursiveReturn` in test_return.py.

2. I commented the early return code in bert_dygraph_model.py because 'return different length of variables' is unsupported now. I also know that there are some other models used early return and we didn't enable it in the unit test. I will add support for it in next PRs and then re-enable those tests.


python/paddle/fluid/dygraph/dygraph_to_static/return_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+            self.count_return[cur_func] += 1
+        else:
+            self.count_return[cur_func] = 1
+        self.generic_visit(node)
+
+    def get_func_return_count(self, func_node):
+        return self.count_return[func_node]
+
+    def set_func_return_count(self, func_node, count):
+        self.count_return[func_node] = count
```
liym27, line 71: This function `set_func_return_count ` doesn't seem to be called.

liym27 (commented):
LGTM for the rest

python/paddle/fluid/dygraph/dygraph_to_static/return_transformer.py
zhhsplendid: Yes, I found it. I will remove in my next PR which supports various-length return. Indeed it is already deleted in my working station.

liym27 (approved):
LGTM

PR #25178, opened 2020-06-23
adding elementwiseadd quantization

sfraczek:
### PR types
 Performance optimization

### PR changes 
Others

### Describe
Adds Quantization of _elementwise_add_ operator.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
@@ -55,8 +55,8 @@ void LogScaleIsMissingForVar(Node* var) {
 void LogQuantizationDisabled(Node* op) {
   std::stringstream msg_ss;
   VLOG(4) << "Qantization skipped for operator " << op->Name()
-         << " (type: " << op->Op()->Type() << ", id: " << op->id()
-         << "). Attribute use_quantizer = false.";
+          << " (type: " << op->Op()->Type() << ", id: " << op->id()
+          << "). Attribute use_quantizer = false.";
 }
 
 }  // namespace
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
@@ -52,11 +52,11 @@ void LogScaleIsMissingForVar(Node* var) {
   PrettyLogDetail(msg_ss.str().c_str());
 }
 
-void LogQuantizationDisabled(Node* op) {   std::stringstream msg_ss;
-  msg_ss << "Qantization skipped for operator " << op->Name()
+void LogQuantizationDisabled(Node* op) {
+  std::stringstream msg_ss;
+  VLOG(4) << "Qantization skipped for operator " << op->Name()
          << " (type: " << op->Op()->Type() << ", id: " << op->id()
          << "). Attribute use_quantizer = false.";
-  PrettyLogDetail(msg_ss.str().c_str());
 }
 
 }  // namespace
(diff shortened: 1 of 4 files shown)


wojtuss (commented):
Please, add test cases:
* with different `is_unsigned` values, 
* with output scale missing.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -676,6 +677,79 @@ void CPUQuantizePass::QuantizeMatmul(Graph* graph) const {
...
+  auto handler = [&](const GraphPatternDetector::subgraph_t& subgraph,
+                     Graph* g) {
+    VLOG(4) << "Quantize elementwise_add op";
+    GET_IR_NODE_FROM_SUBGRAPH(elementwise_add_op, elementwise_add_op,
+                              elementwise_add_pattern);
+    auto* elementwise_add_op_desc = elementwise_add_op->Op();
+
+    // skip if should not be quantized
+    if (!elementwise_add_op_desc->GetAttrIfExists<bool>("use_quantizer")) {
+      return;
```
arlesniak, line 726: Please add a VLOG like info why it was skipped

sfraczek: I've seen other quantization passes in this file and all of them don't have a VLOG. I think we should either put VLOGs in all of them or not. Should I add it to all passes in cpu_quantize_pass @arlesniak ? @wojtuss what do you think?

arlesniak: I referred to line 692 with VLOG saying "Quantize ..." suggesting the quantization will be applied. But here (line 699) we skip the next steps silently by return. I could suggest that moving the VLOG message from 692 to 701 would be sufficient to give clear message to the user.

arlesniak: or adding LogCannotQuantizeOp with explanation. What do you think ?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -676,6 +677,79 @@ void CPUQuantizePass::QuantizeMatmul(Graph* graph) const {
...
+    GET_IR_NODE_FROM_SUBGRAPH(elementwise_add_x, elementwise_add_x,
+                              elementwise_add_pattern);
+    GET_IR_NODE_FROM_SUBGRAPH(elementwise_add_y, elementwise_add_y,
+                              elementwise_add_pattern);
+    GET_IR_NODE_FROM_SUBGRAPH(elementwise_add_out, elementwise_add_out,
+                              elementwise_add_pattern);
+
+    if (!AreScalesPresentForNodes(elementwise_add_op,
+                                  {elementwise_add_x, elementwise_add_y})) {
+      LogCannotQuantizeOp(elementwise_add_op);
```
arlesniak, line 738: I think here some details why we stop here (Scales are nor present for nodes) would also be beneficial for the user

arlesniak:
After looking at the file let's treat my comments just as a suggestion :)

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
sfraczek: I think it's a good idea

sfraczek: Actually, AreScalesPresentForNodes() method is logging the missing scales already here https://github.com/PaddlePaddle/Paddle/pull/25178/files/f34886c473e4b9de7ef16515ce10a023d721410e#diff-ea9ef40ecab0417d79504fb4c55d96a6L194.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
sfraczek: As far as I know, `use_quantizer` is is in `cpu_quantize_placement_pass` to all operators that have that attribute (in other words are quantizable), unless they are explicitly excluded/not-included from command-line. I think that this is sufficiently self-explanatory that we do not quantize such operators so personally I think we can skip it because there is not much added value.

wojtuss: I agree that adding the VLOG could be done here in a general manner, i.e. in other places as well.
To me the "Quantize ..." log says only that quantization happens, but at the end of the pass we report how many operators were actually quantized.

wojtuss (approved):
LGTM

PR #25184, opened 2020-06-24
Fix random fail of test_mobile_net

Aurelius84:
### PR typeshttps://github.com/PaddlePaddle/Paddle/pull/25184/files

Bug fixes
### PR changes

Others
### Describe

Fix random fail of test_mobile_net。

I ran the unittest on local machine but can not reproduce the problem. I supposed it's because the two `test_func` Interact with each other and attempted to fix this.


zhhsplendid (approved):
LGTM

PR #25195, opened 2020-06-24
[DygraphToStatic]Fix import fluid error in dygraph_to_static.

wzzju:
### PR types
Bug fixes

### PR changes
Others

### Describe
Fix the bug described as bellow:
image

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/utils.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/utils.py
@@ -369,6 +369,8 @@ def ast_to_func(ast_root, dyfunc, delete_on_exit=True):
     function, the other inner functions are invisible for the decorated function.
     """
     source = ast_to_source_code(ast_root)
+    import_fluid = "import paddle.fluid as fluid\n"
+    source = import_fluid + source
     if six.PY2:
         source = source.encode('utf-8')
         f = tempfile.NamedTemporaryFile(mode='w', suffix='.py', delete=False)


python/paddle/fluid/dygraph/dygraph_to_static/utils.py
```diff
@@ -369,6 +369,8 @@ def ast_to_func(ast_root, dyfunc, delete_on_exit=True):
     function, the other inner functions are invisible for the decorated function.
     """
     source = ast_to_source_code(ast_root)
+    import_fluid = "import paddle.fluid as fluid\n"
```
zhhsplendid, line 372: If you have time, can we put these at two different function?

ast_to_func means ast_to_func. When you add these two lines, you mean ast_to_func for paddle. The the function cannot be re-used if we re-use the function for non-paddle thing at the future.

One way is:
```
def new_func(...):
       source = ast_to_source_code(ast_root)
       source = import_fluid + source
       return source_to_func(source)
```
But your change is also `fine` to me (In English, fine is not as good as `good`) because currently we just use it in dy2stat... If you don't have time, tell me and I can also approve this PR.

zhhsplendid: After discussed offline, there are many places call ast_to_func and I’m fine with current change. I will approve

zhhsplendid (approved):

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
wzzju: OK. It is a good suggestion. Thanks.

PR #25197, opened 2020-06-24
fix Layer.__dir__  to show the parameters/sublayers/buffers/attr/method of the Layer

zhwesky2010:
### PR types

New features
### PR changes

APIs
### Describe

New API: 
```
Layer.__dir__
```
modify dir() to make it easy for users to view the attr/parameters/sublayers/function of the Layer.

For example:

```
import paddle.fluid as fluid
import numpy as np

fluid.dygraph.enable_dygraph()

class Mylayer(fluid.dygraph.Layer):
    def __init__(self):
        super(Mylayer, self).__init__()
        self.linear1 = fluid.dygraph.Linear(10, 10)
        self.linear2 = fluid.dygraph.Linear(5, 5)
        self.conv2d = fluid.dygraph.Conv2D(3, 2, 3)
        self.embedding = fluid.dygraph.Embedding(size=[128, 16])
        self.h_0 = fluid.dygraph.to_variable(np.zeros([10, 10]).astype('float32'))

mylayer = Mylayer()
print(dir(mylayer))
```

image

### DIFF
--- a/python/paddle/fluid/dygraph/layers.py
+++ b/python/paddle/fluid/dygraph/layers.py
@@ -762,7 +762,7 @@ class Layer(core.Layer):
 
             fluid.dygraph.enable_dygraph()
 
-            class Mylayer1(fluid.dygraph.Layer):
+            class Mylayer(fluid.dygraph.Layer):
                 def __init__(self):
                     super(Mylayer1, self).__init__()
                     self.linear1 = fluid.dygraph.Linear(10, 10)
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/layers.py
```diff
@@ -752,6 +752,40 @@ def __delattr__(self, name):
...
+                def __init__(self):
+                    super(Mylayer1, self).__init__()
+                    self.linear1 = fluid.dygraph.Linear(10, 10)
+                    self.linear2 = fluid.dygraph.Linear(5, 5)
+                    self.conv2d = fluid.dygraph.Conv2D(3, 2, 3)
+                    self.embedding = fluid.dygraph.Embedding(size=[128, 16])
+                    self.h_0 = fluid.dygraph.to_variable(np.zeros([10, 10]).astype('float32'))
+
+            mylayer = Mylayer()
+            print(dir(mylayer))
```
zhiqiu, line 775: Better add the output of the example code

zhwesky2010: Ok

zhwesky2010: Done

zhiqiu (approved):
LGTM

PR #25249, opened 2020-06-29
[Dy2stat] Support Various-Length Return Grammar in Dy2stat

zhhsplendid:
### PR types

New features

### PR changes

Others

### Describe

Support Various-Length Return Grammar in Dy2stat. This PR is a follow-up of https://github.com/PaddlePaddle/Paddle/pull/25176 .

The basic idea is putting no-value placeholder variables at `return` statement to make all `return` statement have same length, after that the static graph can have fixed fetch output (code at return_transformer.py). Then remove those no-value placeholder when we finally return dygraph result (code at partial_program.py).

However, various length return in Bert model is still not supported. The dy2stat can change the code as I wish but some ops which check shape at compile time (e.g. Reshape, MatMul) will throw error because of the no-value-placeholder may not have the required shape. Is this a matter? To me, those no-value placeholder will be replaced as really values meeting shape requirements at run time, so I think the solution should be some way to do the compile-time checking. By the way, every time when we have dynamic shape, it often causes problem in dy2stat. We should find a way to handle it in the future.

Fixing various return in Bert is my TODO thing and I will also find some other existing models for verification.


python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
```diff
@@ -239,11 +241,44 @@ def _restore_out(self, out_vars):
         for i, idx in enumerate(self._outputs.var_ids):
             flatten_outputs[idx] = out_vars[i]
         outs = self._outputs.restore(flatten_outputs)
-        if len(outs) == 1:
+        if outs is not None and len(outs) == 1:
             outs = outs[0]
```
Aurelius84, line 245: Thanks to fix this.

chenwhql (approved):
LGTM

Aurelius84 (approved):
LGTM.
Is the rest corn case related with independent on tensor?

zhhsplendid:
> LGTM.
> Is the rest corn case related with independent on tensor?

No, the rest corner case is related to dynamic shape and compile time check

PR #25280, opened 2020-06-30
Modify tmp var name prefix in dygraph

Aurelius84:
### PR types

Others
### PR changes

Others
### Describe

Modify tmp var name prefix in dygraph

The `tmp` is used as prefix key while naming a temporary intermediate var both in imperative and static mode.
But the `UniqueNameGenerator`  in C++ and `unique_name.py` in Python doesn't share the same auto-increment id.
It will create a variable repeatedly with same name like `tmp_0` in some cases when transform dygraph into static layers.
So we modify the default prefix key into `eager_tmp` to distinguish with static graph.


zhhsplendid (approved):
LGTM

phlrain (approved):

PR #25292, opened 2020-07-01
fix the formula of floor OP and ceil OP

WuHaobo:
### PR types
Bug fixes

### PR changes
Docs 

### Describe

1. fix the formula of the [floor OP](https://www.paddlepaddle.org.cn/documentation/docs/en/2.0-alpha/api/layers/floor.html)
image
Now is :
image

2. fix the formula of the [ceil OP](https://www.paddlepaddle.org.cn/documentation/docs/en/2.0-alpha/api/layers/ceil.html)
image
Now is:
image


saxon-zh (approved):
LGTM

PR #25304, opened 2020-07-01
fix the compatibility of PY2 and PY3 in paddle.distributed.launch

gavin1332:
### PR types
Bug fixes

### PR changes
Others

### Describe
As `sys.stdout` has an attribute `buffer` in python3 but no in python2. So in another way, we directly pull the worker log and write to the `stdout` of the main process


phlrain (dismissed):
LGTM

hutuxian (approved):

PR #25313, opened 2020-07-01
Quant2 updates and fixes

wojtuss:
### PR types
Function optimization

### PR changes
Others

### Describe
In this PR several improvements to the Quant2 -> INT8 transformation process and tests have been added:
- process:
> - the list of optimization passes is supplemented,
> - improved logging,
> - fixed conv2d quantization in case of ResidualData are present,
-  tests:
> - INT8 model: accuracy and performance results are reported,
> - Quant model: accuracy results are reported,
> - accuracy is compared between INT8 and Quant models,
> - if an FP32 model is provided using the `--fp32_model` option, it will be optimized and both accuracy and performance are reported,
> - Image Classification tests run without the `--ops_to_quantize` option, enabling quantization of all supported operators,
> - NLP test run with the `--ops_to_quantize` option for code coverage purposes,
> - removed option to save FP32 model obtained from the Quant model.


python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
```diff
@@ -333,20 +339,38 @@ def _remove_ctrl_vars(self, graph):
     def _optimize_fp32_graph(self, graph):
         graph = self._update_activations(graph)
         graph = self._remove_ctrl_vars(graph)
+        graph = self._apply_pass(graph, 'attention_lstm_fuse_pass')
+        graph = self._apply_pass(graph, 'seqconv_eltadd_relu_fuse_pass')
+        #  graph = self._apply_pass(graph, 'seqpool_concat_fuse_pass')
+        graph = self._apply_pass(graph, 'seqpool_cvm_concat_fuse_pass')
```
lidanqing-vv, line 345: Are these fueses tested to be faster than original unfused graph? If lstm did not use mkldnn, will the fuse also be fast?

sfraczek (approved):
LGTM

python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
wojtuss: The list of passes was supplemented to reflect the the list of optimizations applied by PaddlePredictor for FP32 inference.

arlesniak (approved):
LGTM

lidanqing-vv (approved):
LGTM

lidanqing-intel:
Hi, @wojtuss  Do you think  `graph = self._apply_pass(graph, 'simplify_with_basic_ops_pass')` should happen before `graph = self._quantize_fp32_graph(graph)` ?

wojtuss:
> Hi, @wojtuss Do you think `graph = self._apply_pass(graph, 'simplify_with_basic_ops_pass')` should happen before `graph = self._quantize_fp32_graph(graph)` ?

No, it is not necessary.

luotao1 (approved):

PR #25352, opened 2020-07-02
Fixlaunch

gongweibao:
### PR types
Bug fixes

### PR changes
APIs

### Describe
1. Kill descendants of the process which is killed.

### DIFF
--- a/python/paddle/distributed/utils.py
+++ b/python/paddle/distributed/utils.py
@@ -270,6 +270,9 @@ def terminate_local_procs(procs):
     # still alive?
     gone, alive = psutil.wait_procs(decents, timeout=1)
     if len(alive) != 0:
+        for p in alive:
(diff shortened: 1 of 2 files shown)


python/paddle/distributed/utils.py
```diff
@@ -250,29 +251,29 @@ def get_cluster(node_ips, node_ip, paddle_ports, selected_gpus):
...
+    # wait
+    gone, alive = psutil.wait_procs(decents, timeout=3)
+    for p in alive:
+        p.kill()
 
-        time.sleep(3)
+    # still alive?
+    gone, alive = psutil.wait_procs(decents, timeout=1)
+    if len(alive) != 0:
+        logger.fatal("can't kill all process and exit")
```
gavin1332, line 276: show the process id if it could not be killed and give user changes to kill these processes

PR #25359, opened 2020-07-03
[Dy2stat] Disable test_lac on windows

Aurelius84:
### PR types

Others
### PR changes

Others
### Describe

Fix test_lac timeout.

`test_lac` unittest cost only 5s on my dev machine and some windows machine in teamcity, such as `windows2`. But it costs `~350s` on `windows4-1` whose CPU device performance is a bit low. 

The unittest will hang sometimes under command `ctest`  on Windows platform. Because it failed to execute the `ctest` command that leads failing to start up this unittest.  So the unittest raised timeout.


zhhsplendid (approved):
LGTM

PR #25363, opened 2020-07-03
cherry-pick :fix emb eltwise layernorm (#24873)

NHZlX:
### PR types

Bug fixes
### PR changes

Others
### Describe

When using paddle-trt for ernie type model inference，there may be inputs‘ num other than four (standard model has 4 inputs)， There will be an bug.

We fixed this in this PR.


Superjomn (approved):
LGTM

PR #25375, opened 2020-07-04
Revert "add device attr for regularization-related ops"

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

Reverts PaddlePaddle/Paddle#24981, as it may cause failure for some model, for example the c_gan model.


zhangting2020 (approved):
LGTM

PR #25383, opened 2020-07-06
[Dy2stat] Fix Memory Optimization in run_program_op and Add SimNet as Unit Test

zhhsplendid:
### PR types

Others
### PR changes

Others
### Describe

Add Similarity Net as unit test. During the unit test, we found three problems:

1. The run_program_op has memory optimization error when running dy2stat net multiple times.
2. The support for SelectedRows can cause problem in dy2stat.
3. The return grammar has problem.

This PR fixes the 1. problem but modify codes for the 2. 3. problems to make PR smaller. I will fix those two problems in the next PR(s)


liym27 (dismissed):
LGTM

paddle/fluid/operators/run_program_op.h
```diff
@@ -149,14 +153,44 @@ static void ShareVarsFromScope(const std::vector<Variable *> &vars,
...
+    std::vector<std::string> *skip_vars) {
+  const framework::BlockDesc &block = program.Block(0);
+  const std::vector<framework::OpDesc *> &all_ops = block.AllOps();
+
+  std::unordered_set<std::string> grad_op_output;
+  std::unordered_set<std::string> grad_op_input;
+  for (const framework::OpDesc *op : all_ops) {
+    if (!boost::algorithm::ends_with(op->Type(), "_grad")) {
+      continue;
+    }
```
Aurelius84, line 174: I have  a question:  here we use  `_grad` to determine whether a op is `grad_op`, but some op's grad_op is not with `_grad` suffix, such `arrary_write`'s grad_op is `array_read`.

chenwhql (dismissed):
LGTM

Aurelius84 (dismissed):
LGTM for the rest

paddle/fluid/operators/run_program_op.h
chenwhql: maybe also can use `op_role`?

zhhsplendid: Thank you for great suggestion. Done

chenwhql (approved):

Aurelius84 (approved):

liym27 (approved):

PR #25385, opened 2020-07-06
Add If-Else-Return of Yolov3 as Unittest

zhhsplendid:
### PR types

Others
### PR changes

Others
### Describe

Add If-Else-Return of Yolov3 as Unittest


Aurelius84 (approved):
LGTM

liym27 (approved):
LGTM

PR #22001, opened 2019-12-28
fix TLS issue, test=develop

zhiqiu:
Fix #21760 

This PR forces `libmkldnn.so` to link `libiomp5.so` (provided by intel mkl) instead of `libgomp.so `(provided by gcc).

The reasons why we do this include,
-  According to CMAKE of fluid, `core_avx.so` links `libiomp5.so`,  `libmkldnn.so` and other dynamic libraries. If `libmkldnn.so` links `libgomp5.so`, then `core_avx.so` links both `libiomp5.so` and `libgomp5.so`, which is not recommended. (`libgomp5.so` and `libiomp5.so` has same symbols.)

- `libgomp5.so` is compiled with static TLS, which may leads to 

> ImportError：dlopen：cannot load any more object with static TLS

when using more than 14 DSOs with STATIC_TLS and glibc  < 2.22.

Reference: I also opened an issue at mkldnn repo [#625](https://github.com/intel/mkl-dnn/issues/625).

Issue #21760, opened 2019-12-16
import paddle出现ImportError：dlopen：cannot load any more object with static TLS


OliverLPH:
- 标题：import paddle.fluid 时报错dlopen：cannot load any more object with static TLS
- 版本、环境信息：
    1）PaddlePaddle版本：develop：a6e935f4243e52
    2）CPU：MKL/openblas
    3）GPU：CUDA9/CUDA10
    4）系统环境：paddlepaddle/paddle_manylinux_devel:cuda10.0_cudnn7，centos6.1，gcc482
- 安装方式信息：
1）pip安装/docker安装
2）本地编译：
3）docker编译：paddlepaddle/paddle_manylinux_devel:cuda10.0_cudnn7  
  特殊环境请注明：需升级镜像中的glibc版本
- 复现信息：

由于镜像中的glibc版本仅为2.13，而TensorRT的lib依赖的glibc需高于2.13，所以先升级镜像中的glibc至2.18，升级命令如下。
```shell
wget http://ftp.gnu.org/gnu/glibc/glibc-2.18.tar.gz
tar -xvf glibc-2.18.tar.gz
cd glibc-2.18
unset LD_LIBRARY_PATH # must unset LD_LIBRARY_PATH before compile glibc
mkdir build && cd build && ../configure --prefix=/usr && make -j4 && make install
export LD_LIBRARY_PATH=/opt/rh/devtoolset-2/root/usr/lib64:/opt/rh/devtoolset-2/root/usr/lib:/usr/local/lib64:/usr/local/lib:
```

然后安装ci上编译的develop的cuda9.0_cudnn7_mkl whl包

打开Python 后import paddle.fluid

其中mkl的包会报错。
```shell
Python 2.7.11 (default, Nov 15 2019, 12:44:37)
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import paddle
>>> import paddle.fluid as fluid
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/_internal/cpython-2.7.11-ucs4/lib/python2.7/site-packages/paddle/fluid/__init__.py", line 35, in <module>
    from . import framework
  File "/opt/_internal/cpython-2.7.11-ucs4/lib/python2.7/site-packages/paddle/fluid/framework.py", line 35, in <module>
    from . import core
[...]

PR #25390, opened 2020-07-06
Attempt to resolve the TLS problem

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Attempt to resolve the TLS problem

As described in issue #21760 and PR #22001, a error of TLS may happens when import paddle on system with glibc < 2.22.

> ImportError：dlopen：cannot load any more object with static TLS

In PR #22001, I attempt to resolve this problem by forcing libmkldnn.so to link libiomp5.so (provided by intel mkl) instead of libgomp.so, using cmake flags `CMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS `.

However, paddle upgrades the cmake version from 3.5 to 3.10, and `DCMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS` is removed in the newer cmake. So PR #22001 fails to resolve the TLS problem.

This PR tries another way to solve the problem by **preloading `libgomp` before `core_avx.so`.**

**It should be pointed out that, this a tricky way, and the best solution is to upgrade glibc since the TLS bug is fixed in glibc > 2.23.**

### DIFF
--- a/cmake/external/mkldnn.cmake
+++ b/cmake/external/mkldnn.cmake
@@ -36,28 +36,12 @@ SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}" "${MKLDNN_INSTALL_DIR}/${LIBDIR
 
 INCLUDE_DIRECTORIES(${MKLDNN_INC_DIR}) # For MKLDNN code to include internal headers.
 
-IF(${CBLAS_PROVIDER} STREQUAL "MKLML")
-    SET(MKLDNN_DEPENDS   ${MKLML_PROJECT})
-    MESSAGE(STATUS "Build MKLDNN with MKLML ${MKLML_ROOT}")
-ELSE()
-    MESSAGE(STATUS "Build MKLDNN without MKLML")
-ENDIF()
 
 IF(NOT WIN32)
     SET(MKLDNN_FLAG "-Wno-error=strict-overflow -Wno-error=unused-result -Wno-error=array-bounds")
     SET(MKLDNN_FLAG "${MKLDNN_FLAG} -Wno-unused-result -Wno-unused-value")
     SET(MKLDNN_CFLAG "${CMAKE_C_FLAGS} ${MKLDNN_FLAG}")
     SET(MKLDNN_CXXFLAG "${CMAKE_CXX_FLAGS} ${MKLDNN_FLAG}")
-
-    IF(${CBLAS_PROVIDER} STREQUAL "MKLML")
-    # Force libmkldnn.so to link libiomp5.so (provided by intel mkl) instead of libgomp.so (provided by gcc),
-    # since core_avx.so links libiomp5.so 
-        set(MKLDNN_SHARED_LINKER_FLAG "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-as-needed -L${MKLML_LIB_DIR} -liomp5")
-        set(FORBID "-fopenmp")
-    ELSE()
-        set(MKLDNN_SHARED_LINKER_FLAG "${CMAKE_SHARED_LINKER_FLAGS}")
-        set(FORBID "")        
-    ENDIF()
 ELSE()
     SET(MKLDNN_CXXFLAG "${CMAKE_CXX_FLAGS} /EHsc")
 ENDIF(NOT WIN32)
@@ -91,8 +75,6 @@ ExternalProject_Add(
                         -DCMAKE_C_FLAGS=${MKLDNN_CFLAG}
                         -DCMAKE_CXX_FLAGS=${MKLDNN_CXXFLAG}
                         -DDNNL_BUILD_TESTS=OFF -DDNNL_BUILD_EXAMPLES=OFF
-                        -DCMAKE_SHARED_LINKER_FLAGS=${MKLDNN_SHARED_LINKER_FLAG}
-                        -DCMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS=${FORBID}
     CMAKE_CACHE_ARGS    -DCMAKE_INSTALL_PREFIX:PATH=${MKLDNN_INSTALL_DIR}
 )
 if(WIN32)
--- a/python/paddle/fluid/core.py
+++ b/python/paddle/fluid/core.py
@@ -17,6 +17,8 @@ from __future__ import print_function
 import site
 import sys
 import os
+import warnings
+import platform
 
 core_suffix = 'so'
 if os.name == 'nt':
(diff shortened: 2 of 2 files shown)


cmake/external/mkldnn.cmake
```diff
@@ -36,28 +36,12 @@ SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH}" "${MKLDNN_INSTALL_DIR}/${LIBDIR
...
-
-    IF(${CBLAS_PROVIDER} STREQUAL "MKLML")
-    # Force libmkldnn.so to link libiomp5.so (provided by intel mkl) instead of libgomp.so (provided by gcc),
-    # since core_avx.so links libiomp5.so 
-        set(MKLDNN_SHARED_LINKER_FLAG "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-as-needed -L${MKLML_LIB_DIR} -liomp5")
-        set(FORBID "-fopenmp")
-    ELSE()
-        set(MKLDNN_SHARED_LINKER_FLAG "${CMAKE_SHARED_LINKER_FLAGS}")
-        set(FORBID "")        
-    ENDIF()
```
luotao1, line 60: Why do you remove it?
- when `WITH_MKL=ON`, libintel_mklml.so and libmkldnn.so should force use `libiomp5.so`
-  `libiomp5.so` is conflict with `libgomp.so`

zhiqiu: Since `CMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS` is removed in newer CMake, like CMake-3.10. We cannot force libmkldnn.so to use libiomp5.so.
This reverts PR #22001.

luotao1:
Before：
image
After：
image
gomp is conflict with iomp5

zhiqiu:
> Before：
> image
> After：
> image
> gomp is conflict with iomp5

Basically, the difference in the above figures `Before` and  `After` is not introduced by this PR.

It is because Paddle links libiomp, and mkldnn links libgomp when building with gcc. And, we used to  set `CMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS=-fopenmp`  to force libmkldnn.so to use libiomp5.so in PR #22001. But, this flags has been removed in CMake-3.10.

zhiqiu:
> Having both libgomp.so and libiomp5.so linked is not a good idea. In some corner situation there maybe some undefined behaviour.

Agreed that it is not a good idea. I wonder If there is any known specific situations or problems, so we can test it?

> I have a question : why do you prevent "-fopenmp" from passing to link options ?

In order to let libmkldnn.so not link libgomp.so, and link libiomp, so there is only one libiomp5.so

luotao1:
> Having both libgomp.so and libiomp5.so linked is not a good idea

@jczaja @grygielski Do you have some suggestions?

jczaja:
@zhiqiu How about just removing  forbidding "-fopenmp" flag e.g. Could you please show before after with removed only:

`-DCMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS=${FORBID}`

and intact other changes to mkldnn.cmake ?

grygielski:
I don't see any other possible solutions than mentioned here: https://stackoverflow.com/questions/25986091/telling-gcc-to-not-link-libgomp-so-it-links-libiomp5-instead/25988011#25988011 .

zhiqiu:
> @zhiqiu How about just removing forbidding "-fopenmp" flag e.g. Could you please show before after with removed only:
> 
> `-DCMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS=${FORBID}`
> 
> and intact other changes to mkldnn.cmake ?

`CMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS` has no effect with CMake > 3.8, so removing it or not is the same.

zhiqiu:
> I don't see any other possible solutions than mentioned here: https://stackoverflow.com/questions/25986091/telling-gcc-to-not-link-libgomp-so-it-links-libiomp5-instead/25988011#25988011 .

Yes, the solution in StackOverflow is what we did with `CMAKE_CXX_CREATE_SHARED_LIBRARY_FORBIDDEN_FLAGS` before. 

However, since the flags is removed, we cannot make it without` modifying the code of mkldnn library`. 
Maybe `mkldnn` could provide such option in future?

zhiqiu:
Hi, @luotao1 @grygielski @jczaja @lanxianghit 

Since currently many users are facing the problem `ImportError：dlopen：cannot load any more object with static TLS`, and we didn't find a possible solution to force paddle only link `libiomp.so` yet. 

I think we can solve this problem in two steps,

1. Temporarily, resolve the problem`ImportError：dlopen：cannot load any more object with static TLS` by pre-loading `libgomp.so` before other DSOs. 

2. TODO in future, find an effective solution to force paddle only link `libiomp.so`, which not only sovles this problem but also reduce the potential conflict of  having both `libgomp.so` and `libiomp5.so`.


因为现在用cmake-3.10编译paddle，不论是1.8还是develop都有libiomp和libgomp同时存在（两个一起用可能存在冲突隐患），且在一些机器上会在import时出错。暂时未找到完美的解决方案，我想是不是可以分两步，
（1）暂时解决import出错的问题，通过pre load的方式；
（2）后续再寻找更好的方案，只link一个libiomp，从源头解决这个问题。

jczaja:
@zhiqiu How can I reproduce this error with develop? It should be possible with quite recent version of cmake ? How can I do that?

zhiqiu:
> @zhiqiu How can I reproduce this error with develop? It should be possible with quite recent version of cmake ? How can I do that?

Build with CMake-3.10, and then install it on system with glibc < 2.23

lidanqing-vv:
Hi, @jczaja Luotao ask if you have some formal solutions, for now they could only think about temporary fix for release/1.8

luotao1:
Discussed with @yinghu5, we can remove iomp5 and only use gomp later.

luotao1:
@T8T9 After this PR is merged, we can support cmake3.12?

zhiqiu:
> Discussed with @yinghu5, we can remove iomp5 and only use gomp later.

OliverLPH:
@zhiqiu I have verified this PR has fixed TLS issue.

whufengchen:
when glibc <2.23 ,and load more than 14 dso  this happens。 There are three ways to solve this problem：
（1）upgrade glic to above 2.23
  (2)  do not use dso with static TLS
 (3) preload this dso before the 14th TLS DOS

e.g.
a. use LD_PRELOAD
     export LD_REPLOAD=libgomp.so
     python a.py

or 

b.  modify code
     from ctypes import cdll
     cdll.LoadLibrary(dso_absolute_path)
     import paddle.fluid as fluid

PR #25395, opened 2020-07-06
Speedup reader check_input_array when item is array

wangxicoding:
### PR types

Performance optimization

### PR changes

Others

### Describe

When the item is ndarry, _check_input_array will copy it as a new array, this copy is unnecessary. Now use `asarray` to optimize this scene.

image
As shown in the figure, when the amount of data is 100 * 300 * 768 * 4 = 87.89MB, it takes 77.7ms to copy. While `asarray` uses zerocopy in this scenario, which takes only 326ns.

### DIFF
--- a/python/paddle/fluid/reader.py
+++ b/python/paddle/fluid/reader.py
@@ -809,7 +809,7 @@ class DygraphGeneratorLoader(DataLoaderBase):
 
     @classmethod
     def _check_input_array(cls, item):
-        arr = np.array(item)
+        arr = np.asarray(item)
         if arr.dtype == np.object:
             raise TypeError(
                 "\n\tFaild to convert input data to a regular ndarray :\n\t* Usually "
@@ -817,6 +817,7 @@ class DygraphGeneratorLoader(DataLoaderBase):
                 "\n\t* Check the reader function passed to 'decorate_batch_generator'"
                 " to locate the data causes this issue.\n\t* Please consider using "
                 "'fluid.create_lod_tensor' to convert it to a LoD-Tensor.")
+        return arr
 
     def _exit_thread_expectedly(self):
         self._thread_done_event.set()
@@ -894,7 +895,7 @@ class DygraphGeneratorLoader(DataLoaderBase):
                 array = core.LoDTensorArray()
                 for item in sample:
                     if not isinstance(item, core.LoDTensor):
-                        self._check_input_array(item)
+                        item = self._check_input_array(item)
                         tmp = core.LoDTensor()
                         tmp.set(item, core.CPUPlace())
                         item = tmp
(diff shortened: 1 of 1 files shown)


wangxicoding:
> In ernie distill train, 4 * 87.89MB = 351MB copy time from 0.4s-> 0.08s

image
image
image

python/paddle/fluid/reader.py
```diff
@@ -1117,7 +1118,7 @@ def reset(self):
 
     @classmethod
     def _check_input_array(cls, item):
-        arr = np.array(item)
+        arr = np.asarray(item)
```
gongweibao: Duplicate `_check_input_array`?

python/paddle/fluid/reader.py
```diff
@@ -809,14 +809,15 @@ def __next__(self):
...
         if arr.dtype == np.object:
             raise TypeError(
                 "\n\tFaild to convert input data to a regular ndarray :\n\t* Usually "
                 "this means the input data contains nested lists with different lengths. "
                 "\n\t* Check the reader function passed to 'decorate_batch_generator'"
                 " to locate the data causes this issue.\n\t* Please consider using "
                 "'fluid.create_lod_tensor' to convert it to a LoD-Tensor.")
+        return arr
```
gavin1332: why returns converted array

python/paddle/fluid/reader.py
```diff
@@ -894,7 +895,7 @@ def _reader_thread_loop_for_singleprocess(self):
                 array = core.LoDTensorArray()
                 for item in sample:
                     if not isinstance(item, core.LoDTensor):
-                        self._check_input_array(item)
+                        item = self._check_input_array(item)
                         tmp = core.LoDTensor()
                         tmp.set(item, core.CPUPlace())
```
gavin1332, line 900: Is there any efficiency difference between setting list or ndarray to a LoDTensor ?

wangxicoding:
https://github.com/PaddlePaddle/Paddle/blob/b4702a242d95a30f3d1a8fa4685872d68abd5667/python/paddle/fluid/reader.py#L897-L901

wangxicoding:
https://github.com/PaddlePaddle/Paddle/blob/b4702a242d95a30f3d1a8fa4685872d68abd5667/python/paddle/fluid/reader.py#L897-L901
https://github.com/PaddlePaddle/Paddle/blob/eb3173e2b6f6ac65e47236eca9ed7134e7871b1c/paddle/fluid/pybind/tensor_py.h#L220-L223
In _check_input_array, it will convert input(list or ndarray) to ndarray. While in tensor set, it will convert input to ndarray again. 
_check_input_array return converted array can reduce one convert.
If input is list with 100 * 300 * 768 data, can save 1.04s。。
image
@gavin1332

gavin1332:
I see, the function setting LoDTensor will convert any input to ndarray, and then to LoDTensor.

python/paddle/fluid/reader.py
wangxicoding: Done. Move _check_input_array  to DataLoaderBase.

PR #25402, opened 2020-07-06
Add bfloat16 data type

wozna:
### PR types

Others 
### PR changes

Others 

### Describe

PR adds bfloat16 data type implementation with a test for this type. 
Formats of float32 and bfloat16 are presented in the picture below.
image
The implementation of bfloat16 focuses on the conversion from float to bfloat16, which involves copying the first 2 bytes from float and saving them as bfloat16. The rest of the conversion from other types is done using the default conversion to float and then to bfloat16.

This is the first step associated with model inference on bfloat16 using the OneDNN library on supporting devices such as Cooper Lake.

This PR does not need verification on Cooper Lake, because it is a universal implementation of the bfloat16 data type.

### DIFF
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -116,6 +116,8 @@ void* GetDataFromTensor(const Tensor& tensor, mkldnn::memory::data_type type) {
       return platform::to_void_cast(tensor.data<unsigned char>());
     case mkldnn::memory::data_type::s32:
       return platform::to_void_cast(tensor.data<int32_t>());
+    case mkldnn::memory::data_type::bf16:
+      return platform::to_void_cast(tensor.data<paddle::platform::bfloat16>());
     default:
       PADDLE_THROW(
           platform::errors::InvalidArgument("Wrong mkldnn type provided."));
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -106,7 +106,8 @@ using mkldnn::memory;
 using mkldnn::primitive;
 using mkldnn::reorder;
 
-void* GetDataFromTensor(const Tensor& tensor, mkldnn::memory::data_type type) {
+template <class T>
+void* GetDataFromTensor(const Tensor& tensor, T type) {
   switch (type) {
     case mkldnn::memory::data_type::f32:
       return platform::to_void_cast(tensor.data<float>());
@@ -180,7 +181,7 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
   out->Resize(in.dims());
 
   if (in_format != out_format) {
-    void* in_data = GetDataFromTensor(in, in_type);
+    void* in_data = GetDataFromTensor<mkldnn::memory::data_type>(in, in_type);
     const std::string key =
         platform::CreateKey(in_tz, in_format, out_format, in_type);
 
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -106,8 +106,7 @@ using mkldnn::memory;
 using mkldnn::primitive;
 using mkldnn::reorder;
 
-template <class T>
-void* GetDataFromTensor(const Tensor& tensor, T type) {
+void* GetDataFromTensor(const Tensor& tensor, mkldnn::memory::data_type type) {
   switch (type) {
     case mkldnn::memory::data_type::f32:
       return platform::to_void_cast(tensor.data<float>());
@@ -181,7 +180,7 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
   out->Resize(in.dims());
 
   if (in_format != out_format) {
-    void* in_data = GetDataFromTensor<mkldnn::memory::data_type>(in, in_type);
+    void* in_data = GetDataFromTensor(in, in_type);
     const std::string key =
         platform::CreateKey(in_tz, in_format, out_format, in_type);
 
--- a/paddle/fluid/framework/data_layout_transform.h
+++ b/paddle/fluid/framework/data_layout_transform.h
@@ -61,7 +61,8 @@ inline MKLDNNDataType ToMKLDNNDataType(proto::VarType::Type type) {
       {DataTypeTrait<float>::DataType(), MKLDNNDataType::f32},
       {DataTypeTrait<int8_t>::DataType(), MKLDNNDataType::s8},
       {DataTypeTrait<uint8_t>::DataType(), MKLDNNDataType::u8},
-      {DataTypeTrait<int32_t>::DataType(), MKLDNNDataType::s32}};
+      {DataTypeTrait<int32_t>::DataType(), MKLDNNDataType::s32},
+      {DataTypeTrait<platform::bfloat16>::DataType(), MKLDNNDataType::bf16}};
   auto iter = dict.find(static_cast<int>(type));
   if (iter != dict.end()) return iter->second;
   return MKLDNNDataType::undef;
--- a/paddle/fluid/framework/data_layout_transform.h
+++ b/paddle/fluid/framework/data_layout_transform.h
@@ -76,7 +76,9 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
                                const OpKernelType& expected_kernel_type,
                                const Tensor& in, Tensor* out);
 #endif
-void* GetDataFromTensor(const Tensor& tensor, ::dnnl::memory::data_type type);
+
+template <class T>
+void* GetDataFromTensor(const Tensor& tensor, T type);
 
 std::vector<int> GetAxis(const DataLayout& from, const DataLayout& to);
 
--- a/paddle/fluid/framework/data_layout_transform.h
+++ b/paddle/fluid/framework/data_layout_transform.h
@@ -76,6 +76,7 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
                                const OpKernelType& expected_kernel_type,
                                const Tensor& in, Tensor* out);
 #endif
+void* GetDataFromTensor(const Tensor& tensor, ::dnnl::memory::data_type type);
 
 std::vector<int> GetAxis(const DataLayout& from, const DataLayout& to);
 
--- a/paddle/fluid/framework/data_layout_transform.h
+++ b/paddle/fluid/framework/data_layout_transform.h
@@ -75,10 +75,10 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
 void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
                                const OpKernelType& expected_kernel_type,
                                const Tensor& in, Tensor* out);
-#endif
 
-template <class T>
-void* GetDataFromTensor(const Tensor& tensor, T type);
+void* GetDataFromTensor(const Tensor& tensor, MKLDNNDataType type);
+
+#endif
 
 std::vector<int> GetAxis(const DataLayout& from, const DataLayout& to);
 
--- a/paddle/fluid/framework/data_layout_transform_test.cc
+++ b/paddle/fluid/framework/data_layout_transform_test.cc
@@ -43,3 +43,18 @@ TEST(DataTransform, DataLayoutFunction) {
   EXPECT_TRUE(in.layout() == paddle::framework::DataLayout::kNHWC);
   EXPECT_TRUE(in.dims() == paddle::framework::make_ddim({2, 3, 1, 2}));
 }
+
+#ifdef PADDLE_WITH_MKLDNN
+TEST(DataTransform, GetDataFromTensorDNNL) {
+  auto place = paddle::platform::CPUPlace();
+  paddle::framework::Tensor in = paddle::framework::Tensor();
+  paddle::framework::Tensor out = paddle::framework::Tensor();
+  in.mutable_data<paddle::platform::bfloat16>(
+      paddle::framework::make_ddim({2, 3, 1, 2}), place);
+
+  void* in_data =
+      paddle::framework::GetDataFromTensor(in, dnnl::memory::data_type::bf16);
+  EXPECT_EQ(in_data, paddle::platform::to_void_cast(
+                         in.data<paddle::platform::bfloat16>()));
+}
+#endif
--- a/paddle/fluid/framework/data_layout_transform_test.cc
+++ b/paddle/fluid/framework/data_layout_transform_test.cc
@@ -48,7 +48,6 @@ TEST(DataTransform, DataLayoutFunction) {
 TEST(DataTransform, GetDataFromTensorDNNL) {
   auto place = paddle::platform::CPUPlace();
   paddle::framework::Tensor in = paddle::framework::Tensor();
-  paddle::framework::Tensor out = paddle::framework::Tensor();
   in.mutable_data<paddle::platform::bfloat16>(
       paddle::framework::make_ddim({2, 3, 1, 2}), place);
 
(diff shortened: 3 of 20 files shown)


paddle/fluid/pybind/tensor_py.h
```diff
@@ -241,8 +245,8 @@ void SetTensorFromPyArray(framework::Tensor *self, const py::object &obj,
   } else if (py::isinstance<py::array_t<uint16_t>>(array)) {
     // TODO(cql): temporary keeping uint16, which is used for casting float16
     // before. It should be depracated later.
-    SetTensorFromPyArrayT<paddle::platform::float16, P>(self, array, place,
-                                                        zero_copy);
+    SetTensorFromPyArrayT<paddle::platform::bfloat16, P>(self, array, place,
+                                                         zero_copy);
```
grygielski: Can we safely replace that? Isn't it used somewhere in GPU operators to convert `uint16_t` numpy arrays to `float16` ones?

wozna: I checked that since NumPy supports type float16, tests have been changed.
@zhiqiu  Do you know if conversion from uint16 to float16 is still needed?

zhiqiu: Since PR #19964, ` Tensor.set()` method supports  numpy array of `float16` directly, which means conversion from `uint16` to `float16` is not needed if user want to get a `float16` tensor.

paddle/fluid/framework/io/shell.h
```diff
@@ -17,6 +17,7 @@
 #include <fcntl.h>
 #include <sys/stat.h>
 #ifdef _WIN32
+#define NOMINMAX  // msvc max/min macro conflict with std::min/max
```
jczaja: Perhaps is would be good to add this define if it is not already defined:

#ifndef NOMINMAX
#define NOMINMAX
#endif

wozna: You are right that is a better way. I will add as a separate PR.

paddle/fluid/pybind/tensor_py.h
wozna: Thank you @zhiqiu. It's good news.

wozna:
@luotao1 @wzzju  Can ask you for review and approval for framework.proto because **PR-CI-CPU-Py2** informs about it.
In **PR-CI-Coverage** there is a build error that BF16 is not a proto VarType although it is defined in framework.proto. I wonder if this is also due to a lack of approval or if it is a bug in the code.

luotao1:
> In PR-CI-Coverage there is a build error that BF16 is not a proto VarType although it is defined in framework.proto. I wonder if this is also due to a lack of approval or if it is a bug in the code.

It is a bug in the code, not due to lack of approval

wozna:
@luotao1 You're right, but I still have a problem with that **PR-CI-Coverage** that BF16 is not visible as proto VarType although it is defined in framework.proto. The problem came when **PR-CI-Coverage** started to be built with -DWITH_LITE=ON. Could you ask someone related to Lite to help me with this problem?

luotao1:
> The problem came when PR-CI-Coverage started to be built with -DWITH_LITE=ON. Could you ask someone related to Lite to help me with this problem?

`WITH_LITE=ON` seems not to affect the BF16.  @lidanqing-intel Could you help @wozna internal?

lidanqing-vv:
Adam Osewski is looking at this WITH_LITE=ON issue.   @arogowie-intel

wozna:
@luotao1  We found out that problem with DWITH_LITE=ON is caused by https://github.com/PaddlePaddle/Paddle-Lite/blob/develop/lite/core/framework.proto which is the same file in Paddle and Paddle-Lite.  That is why BF16 wasn't visible in Paddle-Lite, because Lite was using its framework.proto without defined BF16.
The simples solution is to first add BF16 to framework.proto in Paddle-Lite and then in Paddle.  
What do you think about it? Is there any procedure for this?

luotao1:
@wozna The update of `framework.proto` should be cautious. Discussed with @Superjomn who is responsible for inference, to avoid any incompatible among different release versions, you should provide two experiment results before.
- You could create a separate PR only changed `framework.proto`, such as PR1
- Training a model with develop+PR1, and then do C++ Inference with 1.8.x library.
- Training a model with 1.8.x whl, and then do C++ inference with develop+PR1 library.

If  two experiment results are both OK, you can first add BF16 to framework.proto in Paddle-Lite and then in Paddle.
If not, it is an incompatible update, and we will make a big review first.

wozna:
Thank you @luotao1 . I will provide these experiments. 

> * Training a model with develop+PR1, and then do C++ Inference with 1.8.x library.

@luotao1 or @Superjomn Do you think about some particular model training?

luotao1:
> Do you think about some particular model training

@wozna You can choose any model.

arogowie-intel:
Hi @luotao1 
I'm going to continue work on this PR since @wozna is going on vacation next week. 

(Please correct me if I get sth fundamentally wrong - I'm a newbie here :) )

Let me note that Paddle-Lite contains already 3 copies of `framework.proto` file:

- [lite/core/framework.proto](https://github.com/PaddlePaddle/Paddle-Lite/blob/develop/lite/core/framework.proto)
- [mobile/src/framework/framework.proto](https://github.com/PaddlePaddle/Paddle-Lite/blob/develop/mobile/src/framework/framework.proto)
- [mobile/tools/python/modeltools/core/framework.proto](https://github.com/PaddlePaddle/Paddle-Lite/blob/develop/mobile/tools/python/modeltools/core/framework.proto)

I'm not sure that keeping all those files in sync with Paddle's `framework.proto` is a good solution. I wonder why Paddle-Lite is not using `framework.proto` protobuf message deffinitions available in Paddle already? As I can see Paddle builds it's `framework_proto` target [paddle/fluid/framework/CMakeLists.txt#L29](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/CMakeLists.txt#L29) and this target is used as a dependency ([paddle/fluid/inference/lite/CMakeLists.txt#L8 ](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/inference/lite/CMakeLists.txt#L8)) for file where compiler error is reported. Maybe this is the problem that should be fixed? I checked and both `framework.proto` files generated correct sources (`*pb.h`, `*.pb.cc`) at different paths so this is not the case of overwriting itself.  Apparently it seems like Paddle-Lite used only this proto file: `lite/core/framework.proto`.

Regarding the experiments with PaddleLite you asked I also wonder whether they're necessary since:

- We'd add only new value in protobuf message, no additional changes needed. Not using it. -PR1
- The trained model by either PaddleLite:develop+PR1 or 1.8.x whl I suppose would be FP32 not yet BF16, thus there should be no difference - there are no passes which could change anything with BF16 yet.

arogowie-intel:
@luotao1 
After my investigation the problem is solved by updating the Paddle-Lite commit hash used in Paddle to the latest develop. Can we update Paddle-Lite version?

luotao1:
Sorry for the late reply.
> I'm not sure that keeping all those files in sync with Paddle's framework.proto is a good solution. I wonder why Paddle-Lite is not using framework.proto protobuf message deffinitions available in Paddle already? 

keeping all those files in sync is a **monitor** to make the compatibility among different training whl and inference library, since update of `framework.proto` should be very careful.

In a trained model (you can obtail a model from @lidanqing-intel, and to watch it), there is `__model__` and `__param__`, where `__model__` is some information generated by `framework.proto`, and `__model__` always trained by old training whl. Thus, if you directly update `framework.proto`, some user old models will fail.

> We'd add only new value in protobuf message, no additional changes needed. Not using it. -PR1

Any update should be verified before

> After my investigation the problem is solved by updating the Paddle-Lite commit hash used in Paddle to the latest develop. Can we update Paddle-Lite version?

update Paddle-Lite version is very easy, but you should do two experiments before https://github.com/PaddlePaddle/Paddle/pull/25402#issuecomment-672643072

@lidanqing-intel Please add some details, thanks very much!

luotao1:
I see #26413, could you update #26413 with add BF16 in `framework.proto` together?

arogowie-intel:
> I see #26413, could you update #26413 with add BF16 in `framework.proto` together?

Done

arogowie-intel:
@luotao1 I've made the two experiments you asked on this PR #26413.

luotao1:
> Will bfloat16 compile and run well on non-CPX machines

Answer from  @lidanqing-intel  Yes. Accuracy on each machine will be all correct, because if it is not CPX machine bfloat16 simulation is running, or the reference version is running. Just like INT8 could also run on SKX.
@Wozna, Joanna has checked on Skylake, bfloat16 code compile well and run well with good accuracy.

@OliverLPH 和Intel沟通了BF16的测试问题。CI机器是6148的，BF16的精度是可以在6148机器上确认的。CPX 两台机器用于CE即可，类似之前两台6271机器用于INT8性能测试一样。

@OliverLPH Since BF16 run well with good accuracy on 6148 machine, the two CPX machines could be used for CE, like previous 6271 machines test for INT8.

luotao1:
Since #26413 is merged, please re-trigger the CI

paddle/fluid/framework/data_layout_transform.cc
```diff
@@ -106,7 +106,8 @@ using mkldnn::memory;
 using mkldnn::primitive;
 using mkldnn::reorder;
 
-void* GetDataFromTensor(const Tensor& tensor, mkldnn::memory::data_type type) {
+template <class T>
+void* GetDataFromTensor(const Tensor& tensor, T type) {
```
wojtuss: Templatization of the function is redundant here.

paddle/fluid/framework/data_type_test.cc
```diff
@@ -38,3 +38,25 @@ TEST(DataType, float16) {
   std::string type = "::paddle::platform::float16";
   EXPECT_STREQ(f::DataTypeToString(dtype).c_str(), type.c_str());
 }
+
+TEST(DataType, bfloat16) {
```
wojtuss, line 42: The first test could be parameterized and reused here, to avoid duplicating of code.
This can be done later as a refactoring.

paddle/fluid/framework/data_type_transform_test.cc
```diff
@@ -189,4 +194,120 @@ TEST(DataTypeTransform, CPUTransform) {
                 static_cast<paddle::platform::float16>(in_data_bool[i]).x);
     }
   }
+
+  // data type transform from/to bfloat16
+  {
```
wojtuss, line 199: The previous block could be parameterized as a function and reused here, to avoid duplication of code.
This can be done later as a refactoring.

wozna:
@luotao1  
Unfortunately, PR https://github.com/PaddlePaddle/Paddle/pull/25402 didn't resolve the problem with Paddle-Lite. 
`framework.proto` in Paddle is still somehow replaced by `framework.proto` from Paddle-Lite. 

For now, my last proposition that resolves this error for sure is to add **BF16** to  `framework.proto`  from Paddle-Lite. Then update Paddle-Lite hash in Paddle.
Right now I am going to do the same experiment that you asked @arogowie-intel  before. 

Do I need to do something more?

paddle/fluid/framework/data_layout_transform.h
```diff
@@ -76,6 +77,9 @@ void TransDataLayoutFromMKLDNN(const OpKernelType& kernel_type_for_var,
                                const Tensor& in, Tensor* out);
 #endif
 
+template <class T>
+void* GetDataFromTensor(const Tensor& tensor, T type);
```
wojtuss: Not needed if template is removed.

luotao1:
>  Unfortunately, PR #25402 didn't resolve the problem with Paddle-Lite.
framework.proto in Paddle is still somehow replaced by framework.proto from Paddle-Lite.

@arogowie-intel Does https://github.com/PaddlePaddle/Paddle/pull/25402#issuecomment-675494623 before not solve the problem? where is the omission in the before investigation?

paddle/fluid/framework/data_layout_transform_test.cc
```diff
@@ -43,3 +43,18 @@ TEST(DataTransform, DataLayoutFunction) {
   EXPECT_TRUE(in.layout() == paddle::framework::DataLayout::kNHWC);
   EXPECT_TRUE(in.dims() == paddle::framework::make_ddim({2, 3, 1, 2}));
 }
+
+#ifdef PADDLE_WITH_MKLDNN
+TEST(DataTransform, GetDataFromTensorDNNL) {
+  auto place = paddle::platform::CPUPlace();
+  paddle::framework::Tensor in = paddle::framework::Tensor();
+  paddle::framework::Tensor out = paddle::framework::Tensor();
```
wojtuss: What should be the role of `out` here?

arogowie-intel:
Hi @luotao1 
 
> @arogowie-intel Does [#25402 (comment)](https://github.com/PaddlePaddle/Paddle/pull/25402#issuecomment-675494623) before not solve the problem? where is the omission in the before investigation?

I'm truly sorry, but it looks like that was my fault. Just before the solution I've proposed in that comment I've been experimenting with Paddle-Lite and adding to its `framework.proto` file `BF16` data type definition. I was compiling Paddle-Lite locally with this change and instructed Paddle to link with it and not to download and compile it by itself. I suspect that after those experiments I might have my working environmet not cleaned entirely which caused that Paddle was still using my local version of Paddle-Lite with updated `framework.proto`. 

After all it seems that this is still somehow unexpected behavior, that the Paddle-Lite proto files have priority over the Paddle ones, or is it correct? I have been trying to find why this happen, however searching through cmake files in Paddle and Paddle-Lite I couldn't find anything suspicious. The update of `framework.proto` of external project used in Paddle is rather a workaround for this issue. Please help us to explain this.

luotao1:
> The update of framework.proto of external project used in Paddle is rather a workaround for this issue.

Got it. I will discuss with @Superjomn

luotao1:
@wozna @arogowie-intel You can create a PR into Paddle-Lite repo to update the framework.proto.

paddle/fluid/framework/data_layout_transform.cc
wozna: You are right. I removed it.

paddle/fluid/framework/data_layout_transform.h
wozna: Done

paddle/fluid/framework/data_layout_transform_test.cc
wozna: It is unnecessary, so I removed it.

paddle/fluid/framework/data_type_test.cc
wozna: You are right. I will refactor it later.

paddle/fluid/framework/data_type_transform_test.cc
wozna: You are right. I will refactor it later.

paddle/fluid/framework/io/shell.h
```diff
@@ -17,9 +17,6 @@
 #include <fcntl.h>
 #include <sys/stat.h>
 #ifdef _WIN32
-#ifndef NOMINMAX
-#define NOMINMAX  // msvc max/min macro conflict with std::min/max
-#endif
```
wozna: This change in the file should be removed. It was added accidentally.

wozna: After **PR-CI-Coverage** will be done, I will remove it, thanks to that there will be only 19 files changed.

jczaja:
@luotao1 , @Superjomn We looked into this problem with this PR not building with -DWITH_LITE=ON option added and actually
root cause is that some of test files: _/Paddle/paddle/fluid/inference/lite/test_engine.cc_
are indirectly including two framework.pb.h files when PaddlePaddle is built with Paddle-Lite (one framework.pb.h comes from Paddle framework.proto and the other comes from Paddle-Lite). Depending which one of those two headers will be processed first by a compiler , that one will be used. So if those framework.proto files are not in sync then there building error. In this PR we modified the order how headers are included so that Paddle framework.pb.h is included before Paddle-Lite framework.pb.h and that way this PR should be buildable. This is a workaround . Ultimately it would be good 
to use only one framework.proto not two or guarantee that both framework.proto are always in sync.

image

jczaja (approved):

grygielski (approved):
LGTM

luotao1 (approved):
LGTM

PR #25435, opened 2020-07-08
Fix index overflow bug of the CUDA kernel loop increment

chenwhql:
### PR types

Bug fixes
### PR changes

Others
### Describe

This PR's work:
1. add macro `CUDA_KERNEL_LOOP` and fix `softmax_with_cross_entropy_grad` cuda kernel for condition index overflow bug.
2. remove all duplicate old macro and replace some old for condition writing to avoid risk.

#### Problem

The original looping in CUDA kernel is:

`for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < (n); i += blockDim.x * gridDim.x)`

This for condition is risky. The value of `blockIdx.x * blockDim.x` may be large, such as over 1GB, the first iteration is no problem here, but when `i += blockDim.x * gridDim.x` is executed, the value of i will greater than INT_MAX and overflow becomes negative value, at this time, the cycle condition `i < (n)` is still satisfied, so it will cause illegal access to cuda memory.
 
Here is a real example in ERINE, it will trigger above error in `softmax_with_cross_entropy_grad` op. The related data are:

    - blockIdx.x = 2172938
    - blockDim.x = 512
    - blockIdx.x * blockDim.x = 1112543864
    - INT_MAX = 2147483647

So we polish the for condition as follow, the int64_t `__index__` will prevent overflow in the loop increment.

```
#define CUDA_KERNEL_LOOP(i, num)                             \
  int64_t __index__ = blockIdx.x * blockDim.x + threadIdx.x; \
  for (int i = __index__; __index__ < (num);                 \
       __index__ += blockDim.x * gridDim.x, i = __index__)
```

#### Using examples
```
   template <typename T>
   __global__ void Scale(T* logit_grad, const T* loss_grad, const int num,
                     const int d, const int remain) {
   CUDA_KERNEL_LOOP(index, num) {
     int idx_n = index / d;
     int idx_remain = index % remain;
     logit_grad[index] *= loss_grad[idx_n * remain + idx_remain];
     }
   }
```

### DIFF
--- a/paddle/fluid/platform/cuda_helper.h
+++ b/paddle/fluid/platform/cuda_helper.h
@@ -26,6 +26,54 @@ enum cublasMath_t { CUBLAS_DEFAULT_MATH = 0 };
 namespace paddle {
 namespace platform {
 
+/*
+ * Summary: Grid stride looping macro in CUDA kernel
+ *
+ *  [ Why need this macro? ]
+ *
+ *    The original looping in CUDA kernel is:
+ *
+ *    `for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < (n); \
+ *        i += blockDim.x * gridDim.x)`
+ *
+ *    This for condition is risky. The value of `blockIdx.x * blockDim.x`
+ *    may be large, such as over 1GB, the first iteration is no problem here,
+ *    but when `i += blockDim.x * gridDim.x` is executed, the value of i
+ *    will greater than INT_MAX and overflow becomes negative value, at
+ *    this time, the cycle condition `i < (n)` is still satisfied, so it
+ *    will cause illegal access to cuda memory.
+ *
+ *    Here is a real example in ERINE, it will trigger above error.
+ *    The related data are:
+ *      - blockIdx.x = 2172938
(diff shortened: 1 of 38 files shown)


paddle/fluid/platform/cuda_helper.h
```diff
@@ -26,6 +26,54 @@ enum cublasMath_t { CUBLAS_DEFAULT_MATH = 0 };
...
+ *                      const int d, const int remain) {
+ *    CUDA_KERNEL_LOOP(index, num) {
+ *      int idx_n = index / d;
+ *      int idx_remain = index % remain;
+ *      logit_grad[index] *= loss_grad[idx_n * remain + idx_remain];
+ *      }
+ *    }
+ *
+*/
+#define CUDA_KERNEL_LOOP(i, nthreads)                        \
```
Xreki: 第2个参数不是线程数，应该是元素个数。

chenwhql: done, thx!

phlrain (approved):

raindrops2sea (approved):
Nice work!

PR #25517, opened 2020-07-14
[Quantization] Skip collecting out threshold for output tensor of which the type is not fp32 or fp64

juncaipeng:
### PR types

Bug fixes
### PR changes

Others
### Describe

Skip collecting out threshold for output tensor of which the type is not fp32 or fp64

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/quantization_pass.py
+++ b/python/paddle/fluid/contrib/slim/quantization/quantization_pass.py
@@ -1442,6 +1442,10 @@ class OutScaleForTrainingPass(object):
         for op in target_ops:
             for output_var_name in _get_op_output_var_names(op):
                 in_node = graph._find_node_by_name(op.outputs, output_var_name)
+                if in_node.dtype() not in \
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/contrib/slim/quantization/quantization_pass.py
```diff
@@ -1548,15 +1552,17 @@ def apply(self, graph):
                 for var_name in var_names:
                     # For compatibility, we save output threshold by two methods.
                     scale_name = self._scale_name(var_name)
-                    scale_v = np.array(
-                        self._scope.find_var(scale_name).get_tensor())[0]
-                    op_node.op()._set_attr("out_threshold", float(scale_v))
+                    scale_var = self._scope.find_var(scale_name)
+                    if scale_var is None:
```
wzzju: Add some warnings before `continue`.

juncaipeng: Done

PR #25522, opened 2020-07-14
Fix hang  in seq_topk_avg_pooling op

Aurelius84:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Related issue： https://github.com/PaddlePaddle/Paddle/issues/24304

#### 1. Problem Description
While using `circle_loss` in user's defined model, the training thread will hang up and never exit. The log shows thread hang in function `get_topk_pos` from `sequence_topk_avg_pooing_op`.

**The indeed reason is as followed:**
```cpp
{
std::vector<T> v(data, data+length);

T min_val = std::numeric_limits<T>::lowest();
T max_val = min_val;
// if v[i] is NAN, `v[i] > max_val` will return false.
if (v[i] > max_val) {
        max_pos = i;
        max_val = v[i];
      }
// so here max_pos = -1 which leads to illegal visit for vector[-1]
v[max_pos] = min_val;   // illegal visit
}
```

In the end of function, C++ will deconstruct vector `v` but found it has illegal visit. However, it will not throw exception while deconstruction process.

So ,the outer `thread` will not exit and `threads.join()` will wait the child thread finish but no way. So the main program hangs.

#### 2. Solution
The direct reason is `NAN` comes out in training model. We refactor the implementation of the `get_top_pos` to avoid illegal visit in vector. If encounter `NAN`, this op will execute normally and will not hang.

### DIFF
--- a/paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
+++ b/paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
@@ -50,7 +50,7 @@ static void get_topk_pos(const T* data, int length, int k, int* pos) {
     }
     VLOG(3) << "top" << id_k + 1 << " index is : " << max_pos;
     PADDLE_ENFORCE_GE(max_pos, 0,
-                      framework::platform::errors::PreconditionNotMet(
+                      platform::errors::PreconditionNotMet(
                           "Expected max_pos >= 0, but received -1. Probably "
                           "because the input data contains `Nan`."));
 
--- a/paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
+++ b/paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
@@ -15,6 +15,7 @@ limitations under the License. */
 #pragma once
 #include <limits>
 #include <string>
+#include <unordered_set>
 #include <vector>
 #include "paddle/fluid/framework/eigen.h"
 #include "paddle/fluid/framework/op_registry.h"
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
```diff
@@ -70,20 +74,29 @@ class SequenceTopkAvgPoolingKernel : public framework::OpKernel<T> {
...
+    PADDLE_ENFORCE_EQ(
+        col->lod().empty(), false,
+        platform::errors::InvalidArgument(
+            "Input(COLUMN) Tensor of SequenceTopkAvgPoolingOp does "
+            "not contain LoD information."));
 
     auto channel_num = context.Attr<int>("channel_num");
     auto topks = context.Attr<std::vector<int>>("topks");
     auto k_num = topks.size();
     auto max_k = topks[topks.size() - 1];
```
zhiqiu, line 97: Better check `tops.size() > 0`, otherwise `topks.size() - 1` may be negative which is prone to error.

Aurelius84: Thanks, I check `topk.size>0` and `channel_num >0` in `Infershape`.

paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
```diff
@@ -13,52 +13,56 @@ See the License for the specific language governing permissions and
...
+  // reversely assign value
+  int real_k = topk_queue.size();
+  for (int i = real_k - 1; i >= 0; i--) {
+    pos[i] = topk_queue.top().second;
+    topk_queue.pop();
   }
-
-  for (size_t i = 0; i < topk_pos.size(); ++i) {
-    pos[i] = topk_pos[i];
+  // fill -1 if needed
```
zhhsplendid: 1. -1 has better to be a constant at the head of file
2. comment "if needed" should be more detailed. For example, "fill 'const name of -1' for the positions where don't have data"

paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
```diff
@@ -13,52 +13,56 @@ See the License for the specific language governing permissions and
...
-        max_pos = i;
-        max_val = v[i];
+static void get_topk_pos(const T* data, int length, int k, int* pos) {
+  VLOG(3) << "length: " << length << " , k : " << k;
+
+  std::priority_queue<std::pair<T, int>, std::vector<std::pair<T, int>>,
+                      std::greater<std::pair<T, int>>>
+      topk_queue;
+
+  for (int i = 0; i < length; i++) {
```
zhhsplendid: In the future, you can use ++i/--i instead of i++/i--.

The reason is: https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
Aurelius84: Thx, done~

paddle/fluid/operators/sequence_ops/sequence_topk_avg_pooling_op.h
Aurelius84: Thx, done~

PR #25527, opened 2020-07-14
make preloading ligbomp more robust, test=develop

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Related #25390
Make preloading ligbomp more robust, for example, handle the case when `libgomp.so` is not used.


luotao1 (approved):
LGTM

PR #25551, opened 2020-07-15
[Dy2Static] Convert assert stmt with new function `convert_assert`

liym27:
### PR types

Others
### PR changes

Others
### Describe

- Add function `convert_assert` to run the transformed code dynamically:
All assert stmts are transformed whether it depends on Tensor or not;
But in run-time, the code will be run in different ways (1) run paddle op like `layers.Assert`; (2) run python assert. How it runs depends on whether it depends on Tensor.


zhhsplendid (approved):
LGTM

PR #23181, opened 2020-03-24
Feature/expand params in auto-generated pybind functions for dygraph operators

zhiqiu:
Related #21569
This PR changes the auto-generated pybind functions for dygraph operators, remove maps in function args to improve performance.

Train ptb for one epoch on v100 GPU place,
before: 112.269 
after: 107.502 
time reduced: -4%

The generated pybind functions can be see [here](https://gist.github.com/zhiqiu/8ed09c348fa27228d409c7740e25c01e).

PR #25552, opened 2020-07-15
add tool to count all op-related api that without core.ops, test=develop

zhiqiu:
### PR types

Others
### PR changes

Others
### Describe

Add tool to count all op-related APIs that without `core.ops`

- Background
In order to get better performance, we use code-generation technique to generate a pybind function for each Operator in PR #21569 and PR #23181, named `core.ops.xxx`.
So, for op-related APIs, we need to upgrade them, to call these auto-generated pybind functions in imperative(dygraph) mode instead of `append_op`. 
For example,

```
def log(x, name=None):
    if in_dygraph_mode():
        return core.ops.log(x)  // call core.ops.log 

    check_variable_and_dtype(x, 'x', ['float32', 'float64'], "log")
    inputs = {'X': [x]}
    helper = LayerHelper('log', **locals())
    dtype = helper.input_dtype(input_param_name='x')
    out = helper.create_variable_for_type_inference(dtype)
    helper.append_op(type="log", inputs={"X": x}, outputs={"Out": out})
    return out
```

This PR adds a tool to count and list the op-related APIs that haven't upgrade to `core.ops`.

### DIFF
--- a/tools/count_api_without_ops.py
+++ b/tools/count_api_without_ops.py
@@ -1,0 +1,140 @@
+# Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


zhiqiu:
Current progess:

OP-related APIs with core.ops : 208
OP-related APIs without core.ops : 468

chenwhql (approved):
LGTM

PR #25579, opened 2020-07-16
[Dy2Stat] Fix scope in run_program_op

Aurelius84:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Fix scope in run_program_op

While training some model, such as Reinforcement learning，we will call multi-step forward and then apply backward all at once. When calling forward every time, we should keep all vars data in step scope, but only one shared scope is available in `run_program_op`. In this PR, I use sub scope to hold the vars instead of inner scope to handle this problem, and drops all sub scopes after backward or `is_test=True`


python/paddle/fluid/tests/unittests/dygraph_to_static/test_reinforcement_learning.py
```diff
@@ -0,0 +1,214 @@
...
+            del policy.rewards[:]
+            del policy.saved_log_probs[:]
+
+            return returns
+
+        loss_data = []
+        running_reward = 10
+        for i_episode in itertools.count(1):
+            state, ep_reward = env.reset(), 0
+            for t in range(1, 1000):  # default 1000
```
zhhsplendid: I saw the original model code is 10000? May be you can change "#default 1000" to "# The default loop number is 10000 in models, we changed it to 1000 for smaller test"

paddle/fluid/operators/run_program_op.h
```diff
@@ -232,10 +232,17 @@ class RunProgramOpKernel : public framework::OpKernel<T> {
 
     auto exe_ctx = exe.Prepare(*program, 0, skip_vars);
 
-    // get scope and clear old vars
-    framework::Scope &scope = *(out_scope_vec->front());
-    auto local_vars = scope.LocalVarNames();
-    scope.EraseVars(local_vars);
+    // NOTE(Aurelius84): While training some models, forward can be called many
+    // times and then apply
```
zhhsplendid: Why is the comment line break so strange? Can you fix it?

paddle/fluid/operators/run_program_op.h
```diff
@@ -251,6 +258,12 @@ class RunProgramOpKernel : public framework::OpKernel<T> {
 
     // Debug info: scope info when run end
     VLOG(3) << framework::GenScopeTreeDebugInfo(out_scope_vec->front());
+    // Step 5. Drop all children scopes while testing.
+    if (is_test) {
+      out_scope_vec->front()->DropKids();
+    }
+    VLOG(2) << "number of sub scope after forward: "
```
zhhsplendid: Tiny English issue:

"The number of sub scopes after forward:"

Notice the sub scope(s)

Same as the similar sentences below.

paddle/fluid/operators/run_program_op.h
Aurelius84: It seems modified by `pre-commit` tools. 

Thx, fixed.

python/paddle/fluid/tests/unittests/dygraph_to_static/test_reinforcement_learning.py
Aurelius84: Thx, done~

paddle/fluid/operators/run_program_op.h
Aurelius84: Thx, done~

PR #25581, opened 2020-07-16
just skip case of distribute on windows

zhwesky2010:
### PR types

Others
### PR changes

Others
### Describe

Add unittests on windows, and skip case of distribution on windows.


luotao1 (approved):
LGTM

PR #25584, opened 2020-07-17
Fix test pe seresnext unittests failed on windows

chenwhql:
### PR types

Others
### PR changes

Others
### Describe

`test_parallel_executor_seresnext_***_cpu` unittests will appear NaN steadily in windows CI. After analysis, it is found that windows CI will run all the training unittests with the ON_INFER option turned on, which will not appear in other CIs. The calculation behavior of some ops in inference mode is inconsistent with that in non-inference mode.


zhwesky2010 (approved):
LGTM

PR #25534, opened 2020-07-14
GRU model xnli dataset C++ tester

lidanqing-intel:
### PR types
Others

### PR changes
Others 

### Describe
Add C++ test for GRU model because python script has more overhead.

Issue #14837, opened 2018-12-10
The `fc_gru_fuse_pass` breaks accuracy in C-API CRNN-CTC inference.


wojtuss:
The `fc_gru_fuse_pass` breaks accuracy when running CRNN-CTC inference using the C-API.

The application
https://github.com/wojtuss/models/tree/wojtuss/capi-ocr-recognition/fluid/PaddleCV/ocr_recognition/capi
runs inference using the CRNN-CTC trained model
https://github.com/wojtuss/models/blob/wojtuss/capi-ocr-recognition/fluid/PaddleCV/ocr_recognition/models/CRNN-CTC_model.tar.gz
 (with weights from Baidu, with EditDistance evaluator added) on the standard dataset.

With the `fc_gru_fuse_pass` disabled in `infer_ocr_recognition.cc:233` the accuracy we get from the app is equal to the accuracy we get using python `eval.py` script: avg distance = 0.09206844, avg seq error = 0.2425.
With the `fc_gru_fuse_pass` enabled, the accuracy gets lost: avg seq error = 1.0000.

The script
https://github.com/wojtuss/models/blob/wojtuss/capi-ocr-recognition/fluid/PaddleCV/ocr_recognition/capi/infer_ocr_with_accuracy.sh
provides an example of how to run the application.

luotao1:
Yes, we already find that `fc_gru_fuse_pass` breaks accuracy in CRNN-CTC, but we don't have a priority to fix it. You can remove this pass in CRNN-CTC at first:
```
config.pass_builder()->DeletePass("fc_gru_fuse_pass");
```

wojtuss:
OK, thanks!

wojtuss:
@frankwhzhang , was this issue fixed?

luotao1:
@wojtuss It hasn't fixed now.

Sand3r-:
Is it known whether the problem lies within the kernel or within the pass? Or perhaps both?

luotao1:
Sorry, we don't confirm it so far. @Sand3r-

luotao1:
@bingyanghuang

PR #25594, opened 2020-07-17
Add oneDNN fusion_gru kernel

grygielski:
### PR types
New features

### PR changes
OPs

### Describe
This PR adds oneDNN fp32 GRU kernel. It also fixes the fc+gru pass.

### DIFF
--- a/cmake/operators.cmake
+++ b/cmake/operators.cmake
@@ -118,13 +118,13 @@ function(op_library TARGET)
 "tensor_array_read_write_op" "tensorrt_engine_op" "conv_fusion_op"
 "fusion_transpose_flatten_concat_op" "fusion_conv_inception_op"
 "sync_batch_norm_op" "dgc_op" "fused_fc_elementwise_layernorm_op"
-"multihead_matmul_op" "fusion_group_op" "fused_bn_activation_op" "fused_embedding_eltwise_layernorm_op")
+"multihead_matmul_op" "fusion_group_op" "fused_bn_activation_op" "fused_embedding_eltwise_layernorm_op" "fusion_gru_op")
         if ("${TARGET}" STREQUAL "${manual_pybind_op}")
             set(pybind_flag 1)
         endif()
     endforeach()
 
-    # The registration of USE_OP, please refer to paddle/fluid/framework/op_registry.h.
+# The registration of USE_OP, please refer to paddle/fluid/framework/op_registry.h.
     # Note that it's enough to just adding one operator to pybind in a *_op.cc file.
     # And for detail pybind information, please see generated paddle/pybind/pybind.h.
     file(READ ${TARGET}.cc TARGET_CONTENT)
--- a/cmake/operators.cmake
+++ b/cmake/operators.cmake
@@ -124,7 +124,7 @@ function(op_library TARGET)
         endif()
     endforeach()
 
-# The registration of USE_OP, please refer to paddle/fluid/framework/op_registry.h.
+    # The registration of USE_OP, please refer to paddle/fluid/framework/op_registry.h.
     # Note that it's enough to just adding one operator to pybind in a *_op.cc file.
     # And for detail pybind information, please see generated paddle/pybind/pybind.h.
     file(READ ${TARGET}.cc TARGET_CONTENT)
--- a/paddle/fluid/operators/fused/CMakeLists.txt
+++ b/paddle/fluid/operators/fused/CMakeLists.txt
@@ -7,7 +7,12 @@ register_operators(EXCLUDES
     fused_fc_elementwise_layernorm_op
     multihead_matmul_op
     fused_embedding_eltwise_layernorm_op
-    fusion_group_op)
+    fusion_group_op
+    fusion_gru_op)
+
+# fusion_gru_op does not have CUDA kernel
+op_library(fusion_gru_op)
+file(APPEND ${pybind_file} "USE_CPU_ONLY_OP(fusion_gru);\n")
 
 if (WITH_GPU)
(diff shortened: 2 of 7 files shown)


grygielski:
@luotao1 Could you advice me on failing PR-CI-Coverage? It seems to look for CUDA kernel of `fusion_gru` but I've only registered MKLDNN version of CPU kernel.

grygielski:
@luotao1 I've tested this PR on CAPI model prepared by @lidanqing-intel (https://github.com/PaddlePaddle/Paddle/pull/25534) and received significant performance boost. For native PP GRU kernel, model runs with 2660.65 FPS and for oneDNN fusion_gru, introduced by this PR, the performance increase to 4770.75 FPS. We have also tested bfloat16 performance of this model on CPX machine and received 5803.67 FPS. We have some ideas that could speed this model even further but we need this PR to be merged. I'm having some problems with CIs, could you look at logs and ask appropriate people for help (there are some CUDA-related errors)?

luotao1:
> It also fixes the fc+gru pass.

Could you fix the fc+gru pass in a separate PR, since you can cherry-pick into release/1.8, and we will release 1.8.4 later.

If you fix the fc+gru pass, #14837 could be closed? And could you verify it in a unit-test, i.e, remove line60 in `analyzer_vis_tester.cc`
https://github.com/PaddlePaddle/Paddle/blob/b4b169467ba27d04e6d4bd5c6bc3f7abbba04c65/paddle/fluid/inference/tests/api/analyzer_vis_tester.cc#L53-L61

grygielski:
@luotao1 I've done that in https://github.com/PaddlePaddle/Paddle/pull/25687

grygielski:
@luotao1 I'm still getting the same error about fusion_gru CUDA kernel not being registered. It seems to be a bug with pybind (and cmake/operators.cmake file), can someone help with that?

luotao1:
> I'm still getting the same error about fusion_gru CUDA kernel not being registered

@donproc Could you help see it?

donproc:
> @luotao1 I'm still getting the same error about fusion_gru CUDA kernel not being registered. It seems to be a bug with pybind (and cmake/operators.cmake file), can someone help with that?

@grygielski 

Would you please try to add `"fusion_gru_op"` to operators.cmake as following

```
    # Define operators that don't need pybind here.

    foreach(manual_pybind_op "compare_reduce_op" "compare_op" "logical_op" "nccl_op"
"tensor_array_read_write_op" "tensorrt_engine_op" "conv_fusion_op"
"fusion_transpose_flatten_concat_op" "fusion_conv_inception_op"
"sync_batch_norm_op" "dgc_op" "fused_fc_elementwise_layernorm_op"
"multihead_matmul_op" "fusion_group_op" "fused_bn_activation_op" "fused_embedding_eltwise_layernorm_op" "fusion_gru_op")
        if ("${TARGET}" STREQUAL "${manual_pybind_op}")
            set(pybind_flag 1)
        endif()
    endforeach()
```

We can pass the compiling but I'm not clear about the side effects now

grygielski:
@donproc It seems that it doesn't work well

donproc:
> @donproc It seems that it doesn't work well

@grygielski May you show more details? Can you pass the compiling? Side effects?

grygielski:
@donproc 
fusion_gru UTs fail with:
` ValueError: Operator "fusion_gru" has not been registered. `
model UTs with fusion_gru fails with:
` Error: Operator fusion_gru has not been registered [Hint: op_info_ptr should not be null.] at (/paddle/paddle/fluid/framework/op_info.h:140) `

But it passes compiling

donproc:
@grygielski 

This fused Op applys only in oneDNN, seems no cuda implementation; But the default bind requires cpu and cuda implementation, so we need bind this op manually.

Would you please try to add xxx to fused/CMakeLists.txt as following

```
register_operators(EXCLUDES
    fused_bn_activation_op
    conv_fusion_op
    fusion_transpose_flatten_concat_op
    fusion_conv_inception_op
    fused_fc_elementwise_layernorm_op
    multihead_matmul_op
    fused_embedding_eltwise_layernorm_op
    fusion_group_op
    fusion_gru_op) # exclude this op bind
```

and 

```
if (WITH_MKLDNN)
    # fusion_gru_op
    op_library(fusion_gru_op)
    file(APPEND ${pybind_file} "USE_CPU_ONLY_OP(fusion_gru);\n")
endif(WITH_MKLDNN)
```
Seems ugly, we may revise lately.

I can pass the compiling and can bind op, but the unittest failed.  Would you please check again?

grygielski:
@donproc  I'll try that but keep in mind that `fusion_gru` not only has oneDNN kernel but also plain CPU kernel so it should be included also if PP is being built without MKLDNN.

donproc:
> @donproc I'll try that but keep in mind that `fusion_gru` not only has oneDNN kernel but also plain CPU kernel so it should be included also if PP is being built without MKLDNN.

Yes,  we should keep the cpu plain kernel

1/2 Test #442: test_fusion_gru_op ...............   Passed    3.60 sec

donproc:
> @donproc I'll try that but keep in mind that `fusion_gru` not only has oneDNN kernel but also plain CPU kernel so it should be included also if PP is being built without MKLDNN.

I just understood what you said, yes , cpu should be enabled defalut.

```

# fusion_gru_op
op_library(fusion_gru_op)
file(APPEND ${pybind_file} "USE_CPU_ONLY_OP(fusion_gru);\n")

```

grygielski:
It seems everything works now, thanks a lot @donproc :)

wojtuss (approved):
LGTM

PR #25652, opened 2020-07-22
update init gflag to support update user-defined flags test=develop

qjing666:
### PR types
Function optimization

### PR changes
APIs

### Describe
Previous InitGflags api only call once when import fluid, while users cannot update the gflags during the runtime. 
In this PR, I update the api for optimization.

### DIFF
--- a/paddle/fluid/platform/init.cc
+++ b/paddle/fluid/platform/init.cc
@@ -55,38 +55,35 @@ namespace framework {
 #define strdup _strdup
 #endif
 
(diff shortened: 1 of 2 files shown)


zhiqiu:
Maybe we can use API `set_flags` to update flags.

qjing666:
> Maybe we can use API `set_flags` to update flags.

Thank you for the reply, I will close the PR if set_flag api meet my requirement.

PR #25676, opened 2020-07-23
[Paddle-TRT]Add trt int8 support for elementwise_mul and scale

cryoco:
### PR types

Performance optimization
### PR changes

Others
### Describe

Add trt int8 support for elementwise_mul and scale.


juncaipeng (approved):
LGTM

PR #25687, opened 2020-07-23
Fix FC + GRU fuse pass

### LINKED ISSUE #14837: The `fc_gru_fuse_pass` breaks accuracy in C-API CRNN-CTC inference.

grygielski:
### PR types
Bug fixes

### PR changes
Others

### Describe
This PR fixes FC + GRU fuse pass and enables it in OCR unit test. It should fix https://github.com/PaddlePaddle/Paddle/issues/14837 issue.

### DIFF
--- a/paddle/fluid/framework/ir/fc_gru_fuse_pass.cc
+++ b/paddle/fluid/framework/ir/fc_gru_fuse_pass.cc
@@ -26,15 +26,15 @@ static int BuildFusion(Graph* graph, const std::string& name_scope,
   GraphPatternDetector gpd;
   auto* pattern = gpd.mutable_pattern();
 
-  // Create pattern.
-  patterns::FC fc_pattern(pattern, name_scope);
-  patterns::GRU gru_pattern(pattern, name_scope);
-
   PDNode* x =
       pattern->NewNode(patterns::UniqueKey("x"))->assert_var_not_persistable();
 
+  // Create pattern.
+  patterns::FC fc_pattern(pattern, name_scope);
   auto* fc_out = fc_pattern(x, with_fc_bias, /* with_relu */ false);
   fc_out->AsIntermediate();  // fc_out is a tmp var, will be removed after fuse.
+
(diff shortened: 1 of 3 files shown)


luotao1:
Nice!! Could you explain where is the bug and how do you find it? We are so interested in it.

grygielski:
The main bug for OCR model was not preserving activation functions during fuse. GRU in OCR has following activation attributes:
image
And during pass, these attributes were not forwarded so `fusion_gru` kernel used default ones: sigmoid/tanh.

Another problem was with recalculating bias in a weird way. There was a new variable created during pass and its values were calculated but then, the old bias was linked to `fusion_gru`. I've changed that in a way that Bias variable from GRU is recalculated based on FC bias and then linked to new `fusion_gru` operator.

luotao1 (approved):
LGTM

luotao1:
Please cherry-pick into release/1.8

PR #25688, opened 2020-07-23
correct the LookaheadOptimizer programDesc

mapingshuo:
### PR types

Bug fixes

### PR changes

APIs

### Describe

correct the LookaheadOptimizer programDes


zhiqiu (approved):
LGTM.
LookaheadOptimizer does not support dygraph.

PR #25694, opened 2020-07-24
Add macro check for using TRT api dynamicRangeIsSet()

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

Add macro check for using TRT api dynamicRangeIsSet(),  because it is only available in TRT version equal or  higher than 5.1.2.2


NHZlX (approved):
LGTM

PR #25714, opened 2020-07-24
fix bug: cross OP behavior is not as expected when axis=0, test=develop

LutaoChu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix bug: Axis =0 does not actually work


wawltor (approved):
LGTM

PR #25725, opened 2020-07-26
records the offset of log when creating by paddle.distributed.launch

gavin1332:
### PR types
Function optimization

### PR changes
Others

### Describe
when a distributed training invoked by paddle.distributed.launch with `--log_dir` set, the log is appended to the log file instead of overridden, so we should record the offset of log file to distinguish the new log from the old one.
the preceding related PR: https://github.com/PaddlePaddle/Paddle/pull/25304/files


gongweibao (approved):
LGTM

PR #25730, opened 2020-07-27
Fix trt instance norm plugin

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix trt instance norm plugin by adding `return`, and re-enable trt instance norm unittest on GCC8.


chalsliu (approved):

NHZlX (approved):
LGTM

PR #25749, opened 2020-07-27
saving inference model when user define activation or weight preprocess function

yukavio:
### PR types

New features
### PR changes

APIs
### Describe

saving inference model when user define activation or weight preprocess function


baiyfbupt (approved):

baiyfbupt:
need add save_dict and load_dict in paddleslim.quant

PR #25768, opened 2020-07-28
add pipeline optimizer for fleet

sandyhouse:
### PR types

New features
### PR changes

APIs
### Describe

Add pipeline optimizer for fleet api.
With pipeline, we mean splitting a model into multiple sub-models, each of which is put on a computing device (CPU, or GPU).
A simple example to use it:
```python
loss = build_model()
strategy = paddle.fleet.DistributedStrategy()
strategy.pipeline = True
strategy.pipeline_configs = {'micro_batch': 2}

optimizer = paddle.optimizer.SGD(learning_rate=0.01)
optimizer.minimize(loss)
```

Two steps are needed to use pipeline:

- use fleet.distributed_optimizer to wrap the traditional optimizer;
- set strategy.pipeline to True

You also can use strategy.pipeline_configs to configure the parameter for pipeline. The only parameter is micro_batch, which is the number of microbatches for each mini-batch.


python/paddle/fluid/tests/unittests/test_fleet_pipeline_meta_optimizer.py
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+import unittest
+import paddle
+import os
+
+
+class TestFleetMetaOptimizer(unittest.TestCase):
+    def setUp(self):
+        os.environ["POD_IP"] = "127.0.0.1"
```
guru4elephant: You can just remove this setUp I think

sandyhouse: Some of these environment variables are needed for the current version of PaddleCloudRoleMaker, for example PADDLE_TRAINER_ENDPOINTS.

PR #25781, opened 2020-07-28
Fix dygraph grad bugs

phlrain:
### PR types
Bug fixes

### PR changes
APIs

### Describe

修复 #25703


paddle/fluid/imperative/partial_grad_engine.cc
```diff
@@ -36,6 +36,15 @@
 namespace paddle {
 namespace imperative {
 
+struct hash_pair {
```
zhhsplendid: NIT: suggest to name struct like "HashPair". However, since Paddle already has lot of places named like this, it also OK to me if we don't change.

Reference:
https://google.github.io/styleguide/cppguide.html#Type_Names

phlrain: DONE

paddle/fluid/imperative/partial_grad_engine.cc
```diff
@@ -36,6 +36,15 @@
 namespace paddle {
 namespace imperative {
 
+struct HashPair {
+  template <class T1, class T2>
+  size_t operator()(const std::pair<T1, T2> &p) const {
```
wzzju: Prefer to add the `noexcept` key word here for the consistency with stl hash, namely `size_t operator()(const std::pair<T1, T2> &p) const noexcept`

paddle/fluid/imperative/partial_grad_engine.cc
```diff
@@ -152,8 +161,10 @@ static void GetGraphInfoBetweenTargets(
   target_vars = *input_target_grads;
 
   std::queue<std::pair<OpBase * /*op*/, OpBase * /*pending op*/>> op_queue;
+  std::unordered_set<std::pair<OpBase *, OpBase *>, HashPair> op_base_visited;
   for (auto &endpoint_op : endpoint_ops) {
     op_queue.emplace(endpoint_op, nullptr);
+    op_base_visited.insert(std::make_pair(endpoint_op, nullptr));
```
wzzju: Prefer to use `op_base_visited.emplace(endpoint_op, nullptr);` for the conciseness and the performance. This is just a suggestion.

paddle/fluid/imperative/partial_grad_engine.cc
```diff
@@ -216,7 +228,10 @@ static void GetGraphInfoBetweenTargets(
     auto iter = preceding_ops.find(op);
     if (iter != preceding_ops.end()) {
       for (auto &preceding_op : iter->second) {
-        op_queue.emplace(preceding_op, op);
+        if (op_base_visited.count(std::make_pair(preceding_op, op)) == 0) {
+          op_queue.emplace(preceding_op, op);
+          op_base_visited.insert(std::make_pair(preceding_op, op));
```
wzzju: Prefer to use `op_base_visited.emplace(preceding_op, op);` for the conciseness and the performance. This is just a suggestion.

wzzju (commented):
LGTM.

paddle/fluid/imperative/partial_grad_engine.cc
phlrain: DONE

paddle/fluid/imperative/partial_grad_engine.cc
phlrain: DONE

paddle/fluid/imperative/partial_grad_engine.cc
phlrain: DONE

wzzju (approved):
LGTM.

zhangylch:
你好，请问下你们这个bug修复后，我们如何才能使用修复后的版本？是等版本更新还是有什么临时的版本使用？

PR #25809, opened 2020-07-29
【Dy2stat】Fix is_test switch incorrectly in PartialProgram

Aurelius84:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix is_test switch incorrectly in PartialProgram.

While training dygraph using `@declarative`，the `is_test` in some ops like BatchNorm` is not switched correctly. In this case, the mean and variance of BatchNorm will be changed after calling `programTranslator.get_output(mode, *args)`.

The reason is that `model.training` is not passed into PartialProgram correctly. We fix it in this PR.

The Pseudo code is:
```python
model.eval()

fake_data = np.random.random(shape)
inputs = to_variable(fake_data)

# The random data will change mean and variance of BN if is_test is not set True.
prog_trans.get_output(_convert, model, *input_vars)
prog_trans.save_inference_model(dirname, feed, fetch)
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
@@ -112,10 +112,10 @@ class PartialProgramLayer(layers.Layer):
         self._outputs = NestSequence(outputs, need_check=True)
         self._params = parameters if parameters is not None else []
 
-        self._infer_program = self._verify_program(main_program)
-        self._train_program = self._append_backward_desc()
-        # Switch infer or train by train() and eval()
-        self._trace_program = None
+        main_program = self._verify_program(main_program)
+        self._infer_program = self._change_is_test_status(main_program, True)
+        self._train_program = self._append_backward_desc(main_program)
+
         self._set_grad_type(self._params)
         self._inner_scope = core.Scope()
         # Set default mode to train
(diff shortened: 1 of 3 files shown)


zhhsplendid (dismissed):
LGTM

python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
```diff
@@ -112,14 +112,14 @@ def __init__(self, main_program, inputs, outputs, parameters=None):
         self._outputs = NestSequence(outputs, need_check=True)
         self._params = parameters if parameters is not None else []
 
-        self._infer_program = self._verify_program(main_program)
-        self._train_program = self._append_backward_desc()
-        # Switch infer or train by train() and eval()
-        self._trace_program = None
+        main_program = self._verify_program(main_program)
+        self._infer_program = self._clone_for_test(main_program)
```
chenwhql, line 116: why not use `Program.clone(for_test=True)`directly?

liym27 (dismissed):
LGTM

python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
Aurelius84: Thanks, done

chenwhql (approved):
LGTM

PR #25838, opened 2020-07-30
Change use_quantizer attribute name and data type

wozna:
### PR types

Others 
### PR changes

OPs 
### Describe

This PR change boolean attribute "use_qunatizer" to "mkldnn_data_type" with type std::string. As a result, the changed attribute will inform on what data type operator should be run. Thanks to this, we don't have to add any new bfloat16-related attribute.
Data types: 

- float32
- int8
- bfloat16


paddle/fluid/inference/api/mkldnn_quantizer.cc
```diff
@@ -50,8 +50,7 @@ bool AnalysisPredictor::MkldnnQuantizer::CalculateScales() {
   using VariableNameMap = std::map<std::string, std::vector<std::string>>;
   std::map<std::string, std::map<std::string, LoDTensor>> gathered_data;
   for (const auto* op : predictor_.inference_program_->Block(0).AllOps()) {
-    if (op->HasAttr("use_quantizer") &&
-        BOOST_GET_CONST(bool, op->GetAttr("use_quantizer"))) {
+    if (op->GetAttrIfExists<std::string>("mkldnn_data_type") == "int8") {
```
grygielski: As before

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
```diff
@@ -96,38 +94,36 @@ void MainTest(std::initializer_list<std::string> quantize_enabled_op_types,
...
 
-  unsigned use_quantizer_true_count = 0;
+  unsigned int8_data_type_count = 0;
   for (auto* node : graph->Nodes()) {
     if (node->IsOp()) {
       auto* op = node->Op();
-      if (op->HasAttr("use_quantizer") &&
-          BOOST_GET_CONST(bool, op->GetAttr("use_quantizer"))) {
-        ++use_quantizer_true_count;
+      if (op->GetAttrIfExists<std::string>("mkldnn_data_type") == "int8") {
```
grygielski: As before

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
```diff
@@ -96,38 +94,36 @@ void MainTest(std::initializer_list<std::string> quantize_enabled_op_types,
...
-  unsigned use_quantizer_true_count = 0;
+  unsigned int8_data_type_count = 0;
 
   for (auto* node : graph->Nodes()) {
     if (node->IsOp()) {
       auto* op = node->Op();
-      if (op->HasAttr("use_quantizer") &&
-          BOOST_GET_CONST(bool, op->GetAttr("use_quantizer"))) {
-        ++use_quantizer_true_count;
+      if (op->GetAttrIfExists<std::string>("mkldnn_data_type") == "int8") {
```
grygielski: Can `HasOpINT8DataType()` be used here?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
```diff
@@ -26,7 +26,7 @@ namespace ir {
 void SetOp(ProgramDesc* prog, const std::string& type, const std::string& name,
            const std::vector<std::string>& inputs,
            const std::vector<std::string>& outputs, bool use_mkldnn,
-           bool use_quantizer = false) {
+           const std::string mkldnn_data_type = "float32") {
```
grygielski: It would be better to pass string as const reference to the function `const std::string& mkldnn_data_type` to avoid unnecessary copy.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
```diff
@@ -142,7 +142,7 @@ static const std::initializer_list<std::string> variable_names{
 // d->Dropout1->g and (g, w5, b3)->Fc1->h and (h,w3,b1,i)->Conv3->j
 //
 // (d,w4, b2)->Conv4->i
-ProgramDesc BuildProgramDesc(bool use_mkldnn, bool use_quantizer) {
+ProgramDesc BuildProgramDesc(bool use_mkldnn, std::string mkldnn_data_type) {
```
grygielski: `std::string` -> `const std::string&`

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
```diff
@@ -24,13 +24,11 @@ namespace ir {
 void SetOp(ProgramDesc* prog, const std::string& type, const std::string& name,
            const std::vector<std::string>& inputs,
            const std::vector<std::string>& outputs,
-           boost::tribool use_quantizer) {
+           const std::string mkldnn_data_type = "float32") {
```
grygielski: As before

paddle/fluid/operators/fc_op.cc
```diff
@@ -142,13 +142,12 @@ class FCOpMaker : public framework::OpProtoAndCheckerMaker {
     AddAttr<bool>(framework::kAllKernelsMustComputeRuntimeShape,
                   "Skip calling InferShape() function in the runtime.")
         .SetDefault(true);
+    AddAttr<std::string>(
+        "mkldnn_data_type",
+        "(string, default \"float32\"). Data type of mkldnn kernel")
+        .SetDefault("float32")
+        .InEnum({"float32", "int8", "bfloat16"});
```
lidanqing-vv, line 154: I wonder does it have to be string? Because string may easily cause some typos
Do you consider enum or constexpr? enum MKLDNN_DATATYPE{ FLOAT32=0, INT8=1, BFLOAT16=2} ?

grygielski: I think using enum is hard because of Python-C++ compatibility of attribute type. This is why most operators use strings with `InEnum()` check.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
```diff
@@ -250,10 +250,10 @@ static const std::initializer_list<std::string> variable_names_concat = {
 ProgramDesc BuildProgramDescConcat() {
   ProgramDesc prog;
 
-  SetOp(&prog, "pool2d", "Pool1", {"a1"}, {"b1"}, true, false);
-  SetOp(&prog, "pool2d", "Pool2", {"a2"}, {"b2"}, true, false);
-  SetOp(&prog, "concat", "Concat", {"b1", "b2"}, {"c"}, true, true);
-  SetOp(&prog, "pool2d", "Pool3", {"c"}, {"d"}, true, false);
+  SetOp(&prog, "pool2d", "Pool1", {"a1"}, {"b1"}, true, "float32");
```
lidanqing-intel, line 253: I really think using string may cause typos. Do you consider using int and enum. Because there are only 3 types right?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
wozna: For now `HasOpINT8DataType()` is defined in `cpu_quantize_pass`. Do you think to public `HasOpINT8DataType()` and put to another file like `mkldnn_helper.h`?

grygielski: I think it would improve readability. Moreover, later we will probably need something like `HasOpBF16DataType()` and having both in same place (e.g. `mkldnn_helper.h`) is a good idea.

lidanqing-vv:
LGTM.

juncaipeng:
@wozna  

For compatibility, paddle prohibits directly deleting the op attribute. So the "use_quantizer" attribute cannot be changed to "mkldnn_data_type". Otherwise, the previous paddle inference library does not support the new optimized model.  

We suggest keeping the "use_quantizer" attribute and adding "mkldnn_data_type".  The "use_quantizer" attribute of quantized op should also be True. Besides, mark the "use_quantizer" attribute as deprecated in op_maker.

luotao1:
> mark the "use_quantizer" attribute as deprecated in op_maker

Some supplements, please refer the deprecated attribute `x_data_format` and `y_data_format` in `elementwise_op.h`
https://github.com/PaddlePaddle/Paddle/blob/a7944904d3b50aeffcaae61f00ecd5d6920805b3/paddle/fluid/operators/elementwise/elementwise_op.h#L139-L142

lidanqing-vv:
@wozna  keep use_qunatizer and need to assign use_qunatizer value to mkldnn_data_type

wozna:
I restore `use_quantizer` and checked compatibility: 
1. If we have an old model and a new code.  

> copy model from old attr to `mkldnn_data_type`

2. If we have a new model and old code

> set use `use_quantizer` to true

luotao1:
Please solve the conflict

PR #25844, opened 2020-07-30
【paddle.fleet】Fleet run graph in Executor and add two more strategies

guru4elephant:
### PR types

New features

### PR changes

APIs

### Describe

Fleet distributed_optimizer.minimize function may generate graph structure for runtime, and current design make the graph structure as an attribute for Program. When we are calling Executor.run, we can check if the input program has _graph to run. 

`fuse_grad_size_in_MB` and `_fuse_grad_size_in_TFLOPS` are two properties for gradient fusion. 
`fuse_grad_size_in_MB` is the size of  Tensor in MegaBytes that will be fused during backpropagation.
`fuse_grad_size_in_TFLOPS` is the TeraBytes of Computation between two adjacent fusion point.\
Currently, these two properties are not valid and we will make the properties applicable in next couple of PRs.


python/paddle/fleet/base/distributed_strategy.py
```diff
@@ -378,6 +378,30 @@ def fuse_all_reduce_ops(self, flag):
...
             print("WARNING: fuse_all_reduce_ops should have value of bool type")
 
+    @property
+    def fuse_grad_size_in_MB(self):
+        return self.strategy.fuse_grad_size_in_MB
+
+    @fuse_grad_size_in_MB.setter
+    def fuse_grad_size_in_MB(self, value):
+        if isinstance(value, int):
+            self.strategy.fuse_grad_size_in_MB = value
```
gongweibao, line 388: Why only int?

python/paddle/fleet/base/distributed_strategy.py
```diff
@@ -378,6 +378,30 @@ def fuse_all_reduce_ops(self, flag):
...
+        else:
+            print("WARNING: fuse_grad_size_in_MB should have value of int type")
+
+    @property
+    def _fuse_grad_size_in_TFLOPS(self):
+        return self.strategy.fuse_grad_size_in_TFLOPS
+
+    @_fuse_grad_size_in_TFLOPS.setter
+    def _fuse_grad_size_in_TFLOPS(self, value):
```
gongweibao, line 397: Same as above?

python/paddle/fleet/base/distributed_strategy.py
guru4elephant: most of the models are bigger than MB?

python/paddle/fleet/base/distributed_strategy.py
guru4elephant: Oh, this should be float, I will fix it.

python/paddle/fleet/base/distributed_strategy.py
gongweibao: yeah. Float is better.

zhwesky2010 (approved):
test_fleet_meta_optimizer ==> test_fleet_recompute_meta_optimizer + test_fleet_graph_execution_meta_optimizer + test_fleet_graph_executor, So approve it!

PR #25868, opened 2020-07-31
remove copy_if_different because windows random fail

zhwesky2010:
### PR types

Others
### PR changes

Others
### Describe

move copy_if_different because windows random fail.


luotao1 (approved):
LGTM

PR #25892, opened 2020-08-03
shape_op support GPUKernel

Aurelius84:
### PR types

New features
### PR changes

OPs
### Describe

+ shape_op support GPUKernel

### DIFF
--- a/paddle/fluid/operators/shape_op.cu
+++ b/paddle/fluid/operators/shape_op.cu
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */
 
 #include <cuda.h>
+#include <vector>
 #include "paddle/fluid/operators/shape_op.h"
 
 namespace paddle {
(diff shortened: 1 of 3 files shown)


paddle/fluid/operators/shape_op.cu
```diff
@@ -12,12 +12,58 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    if (in_var->IsType<SelectedRows>()) {
+      in_dims = in_var->Get<SelectedRows>().value().dims();
+    } else {
+      in_dims = in_var->Get<LoDTensor>().dims();
+    }
+
+    auto* out_t = ctx.Output<Tensor>("Out");
+    auto dim_size = in_dims.size();
+    out_t->Resize({dim_size});
+    // The data type in DDims is int64_t. Downcasting it into int32_t
```
zhiqiu, line 41: Why not output int64_t? And it's better to add some checks if you really want to make the output int32_t.  For example,

```
auto safe_cast = [](int64_t x){
  PADDLE_ENFORCE_LE(x, std::numeric_limits<int32_t>::max(),  err_msg);
  PADDLE_ENFORCE_GE(x, std::numeric_limits<int32_t>::min(),  err_msg);
  return static_cast<int32_t>(x);
}
in_dims_data.push_back(safe_cast(in_dims[i]));
```

zhiqiu:
What about the performance of op chain: `shape->slice->xx`, before and after this PR.

paddle/fluid/operators/shape_op.cu
```diff
@@ -12,12 +12,58 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+using Tensor = framework::Tensor;
+using LoDTensor = framework::LoDTensor;
+using SelectedRows = framework::SelectedRows;
+
+template <typename T>
+class ShapeCUDAKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& ctx) const override {
+    auto* in_var = ctx.InputVar("Input");
+    framework::DDim in_dims;
```
zhhsplendid, line 31: Not important here but you should notice when you write code elsewhere. C++ is a complex language so we often have to distinguish the reference assign and value copy very clearly to make code faster.

framework::DDim overrides the operator= as a copy. So here your code initializes a DDim and then copies the DDim, which slows down this shape_op. A faster way will be using a const pointer or const reference. For example:

```
const framework::DDim &in_dims = in_var->IsType<SelectedRows>() ? in_var->Get<SelectedRows>().value().dims() : in_var->Get<LoDTensor>().dims();
```

It is not important here because shape_op usually is not the speed bottleneck. So you can choose to change or not change here but I think it is good for you to know it (maybe you already know it then it is just a reminder) so I wrote this long comment :-)

paddle/fluid/operators/shape_op.cu
```diff
@@ -12,12 +12,58 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      in_dims = in_var->Get<LoDTensor>().dims();
+    }
+
+    auto* out_t = ctx.Output<Tensor>("Out");
+    auto dim_size = in_dims.size();
+    out_t->Resize({dim_size});
+    // The data type in DDims is int64_t. Downcasting it into int32_t
+    // to keep same with most usage in framework. This is dangerous
+    // but the data value of shape seems no more than the limit of int32_t.
+    std::vector<int32_t> in_dims_data;
```
zhhsplendid, line 44: ddim.h has a function named `vectorize`, could you check whether you can reuse that code?

PR #25907, opened 2020-08-03
【paddle.fleet】add the support for multi-node training for pipeline

sandyhouse:
### PR types

New features
### PR changes

Others
### Describe

Add the support for multi-node training for pipeline with paddle.fleet api.

**How to use it**
The following simple example show how to use pipeline with paddle.fleet api. (train.py)
```
import numpy as np
import paddle.fluid as fluid
with fluid.device_guard("gpu:0"):
    data = fluid.layers.data(name='data', shape=[32, 32], dtype='float32')
    data_loader = fluid.io.DataLoader.from_generator(
        feed_list=[data],
        capacity=32,
        use_double_buffer=True,
        iterable=False)
   fc1 = fluid.layers.fc(input=data, size=1000, act='tanh')
with fluid.device_guard("gpu:0"):
    fc2 = fluid.layers.fc(intput=fc1, size=1000)
    loss = fluid.layers.mean(fc2)

strategy = paddle.fleet.DistributedStrategy()
strategy.pipeline = True
strategy.pipeline_configs = {'micro_batch': 2}
optimizer = paddle.optimizer.SGD(learning_rate=0.01)
optimizer.minimize(loss)

def reader():
    for _ in range(320):
        data = np.random.random(size=[32,32]).astype("float32")
        yield data

data_loader.set_sample_generator(reader, batch_size=1)
place = fluid.CUDAPlace(0)
exe = fluid.Executor(place)
data_loader.start()
exe.train_from_dataset(fluid.default_main_program())
```
**How to run**
On trainer 1:
```
fleetrun --ips="127.0.0.1,127.0.0.2" --gpus=0 train.py
```
On trainer 2:
```
fleetrun --ips="127.0.0.1,127.0.0.2" --gpus=0 train.py
```


python/paddle/fleet/meta_optimizers/common.py
```diff
@@ -47,10 +47,15 @@ def is_optimizer_op(op):
 
 
 class CollectiveHelper(object):
-    def __init__(self, role_maker, nrings=1, wait_port='6174'):
+    def __init__(self,
+                 role_maker,
+                 nrings=1,
+                 wait_port='6174',
+                 use_pipeline=False):
```
gavin1332: Create a new `PipelineHelper` to do compile, which could inherited from `CollectiveHelper`. As `PipelineHelper` is specific for pipeline, it could be put into corresponding meta optimizer.

sandyhouse: Done. Create a new class PipelineHelper inherited from CollectiveHelper, and put it in the file defining pipeline_metat_optimizer.

PR #25920, opened 2020-08-03
Update the code of the compare ops for the broadcast function

wawltor:
### PR types
Others

### PR changes
OPs

### Describe
Fix the broadcast function for the compare ops


FlyingQianMM (approved):
LGTM

PR #25935, opened 2020-08-04
Add limit support for load_dygraph loading jit.save result

chenwhql:
### PR types

Function optimization
### PR changes

APIs
### Describe

Add limit support for `load_dygraph` loading `jit.save` result, only supports loading default save format of `jit.save`.

In order to support other compatible formats, we may need to discuss adding new argument to load_dygraph, and this work will be discussed later.

### DIFF
--- a/python/paddle/fluid/dygraph/checkpoint.py
+++ b/python/paddle/fluid/dygraph/checkpoint.py
@@ -16,12 +16,13 @@ from __future__ import print_function
 
 import os
 import collections
-from ..framework import Variable, default_main_program, in_dygraph_mode, dygraph_only, Parameter, ParamBase
+from ..framework import Variable, default_main_program, in_dygraph_mode, dygraph_only, Parameter, ParamBase, _varbase_creator, _dygraph_tracer
 import pickle
 import six
 from . import learning_rate_scheduler
 import warnings
 from .. import core
+from paddle.fluid.dygraph.io import VARIABLE_FILENAME, EXTRA_VAR_INFO_FILENAME, _load_persistable_vars
 
 __all__ = [
     'save_dygraph',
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/dygraph/checkpoint.py
```diff
@@ -140,22 +141,84 @@ def load_dygraph(model_path, keep_name_table=False):
...
+        # 2. load `__variables.info__`
+        var_info_path = os.path.join(model_prefix, EXTRA_VAR_INFO_FILENAME)
+        if not os.path.exists(var_info_path):
+            raise RuntimeError(
+                "No target can be loaded. Now only supports loading `state_dict` from "
+                "the result saved by `imperative.save` and `imperative.jit.save`."
+            )
+        with open(var_info_path, 'rb') as f:
+            extra_var_info = pickle.load(f) if six.PY2 else pickle.load(
                 f, encoding='latin1')
```
zhhsplendid: Could you add a simple comment about using encoding='latin1'?

I didn't know it before and I don't think it is a common knowledge. I found from python doc that it is required when we unpickle NumPy arrays. Is it the reason?

https://docs.python.org/3/library/pickle.html#pickle.Unpickler

zhhsplendid (commented):
Other things LGTM

python/paddle/fluid/dygraph/checkpoint.py
chenwhql: thx, I think this is the reason, but here it is not Numpy array, but a dict, I try to remove this limit.

chenwhql: done, thx

zhhsplendid (approved):
LGTM

phlrain (approved):

PR #25951, opened 2020-08-04
Add some error messages for the op without double grads.

wzzju:
### PR types

Bug fixes
### PR changes

Others
### Describe

```python
import paddle
from paddle import fluid
import numpy as np

paddle.enable_imperative()
x = fluid.layers.ones(shape=[2, 3, 2, 2], dtype='float32') 
x.stop_gradient = False
y = paddle.fluid.layers.batch_norm(x)

dx = fluid.dygraph.grad(
        outputs=[y],
        inputs=[x], 
        create_graph=True, 
        retain_graph=True)[0]

loss = fluid.layers.reduce_mean(dx)
loss.backward()

print(x.gradient())
```
As the code described above, when an op doesn't own double grad op, the value of `x.gradient()` is `None`. After this pr, this program will throw some error messages, which is shown as below.

image


zhhsplendid (approved):
LGTM

PR #25969, opened 2020-08-05
Fix `paddle.cos` docstring

willthefrog:
### PR types
Others

### PR changes
Docs

### Describe
explain input/out put range and out of boundary behavior


jzhang533 (approved):
lgtm

PR #25986, opened 2020-08-05
refine the value parameter's Tensor support of fill_constant Op test=…

wangchaochaohu:
### PR types

Function optimization 
### PR changes

OPs
### Describe

when value parameter is a tensor and data type is different with dtype, it will raise error message in C++. fix it 
image


zhupengyang (dismissed):
LGTM

lanxianghit (approved):
LGTM

swtkiwi (approved):
LGTM

PR #25987, opened 2020-08-05
add more error info for these ops without double grad ops.

wzzju:
### PR types

Others
### PR changes

Others
### Describe

Add more error info for these ops without double grad ops.


zhhsplendid (approved):
LGTM

PR #26013, opened 2020-08-06
add cpu random Generator

yaoxuefeng6:
### PR types

Others
### PR changes

Others
### Describe

bak


jzhang533 (dismissed):
lgtm
this is a breaking change, the related OP should remove `ctx.Attr<int>("seed")` accordingly.
@yaoxuefeng6 will have instructions on how to fix the related OP.
@XiaoguangHu01 , @willthefrog , please have a look.

paddle/fluid/operators/randint_op.cc
```diff
@@ -42,15 +43,20 @@ class CPURandintKernel : public framework::OpKernel<T> {
...
     std::uniform_int_distribution<T> dist(ctx.Attr<int>("low"),
                                           ctx.Attr<int>("high") - 1);
-    for (int64_t i = 0; i < size; ++i) data[i] = dist(engine);
+    if (framework::Generator::GetInstance()->is_init_py) {
+      auto& gen_engine = framework::Generator::GetInstance()->GetCPUEngine();
+      for (int64_t i = 0; i < size; ++i) data[i] = dist(gen_engine);
+    } else {
+      unsigned int seed = static_cast<unsigned int>(ctx.Attr<int>("seed"));
+      std::minstd_rand engine;
+      if (seed == 0) {
```
willthefrog: what if use want to set seed to literal 0?

paddle/fluid/framework/generator.h
```diff
@@ -0,0 +1,96 @@
...
+#include <memory>
+#include <mutex>  // NOLINT
+#include <random>
+#include <typeinfo>
+#include <utility>
+
+namespace paddle {
+namespace framework {
+
+struct GeneratorState {
```
willthefrog, line 30: why note simply merge this into `Generator` as it is one-to-one "has a" relationship?

paddle/fluid/operators/randint_op.cc
willthefrog: what if use want to set seed to literal 0?

paddle/fluid/framework/generator.h
```diff
@@ -0,0 +1,96 @@
...
+  int64_t device = -1;
+  uint64_t current_seed = 34342423252;
+  std::mt19937_64 cpu_engine;
+};
+
+struct Generator {
+  Generator() {
+    GeneratorState default_gen_state_cpu;
+    default_gen_state_cpu.device = -1;
+    default_gen_state_cpu.current_seed = 34342423252;
```
willthefrog, line 40: default initialize with fixed seed?

paddle/fluid/operators/randint_op.cc
yaoxuefeng6: Currently It is not allowed in paddle, and related doc note this to users. If it is necessary to allow this, we can fix it.

paddle/fluid/framework/generator.h
yaoxuefeng6: To let State-related value be in this class. we can easily and clearly add things in this class later.

paddle/fluid/framework/generator.h
yaoxuefeng6: We should set a default one. maybe random?

willthefrog: initialization with random seed seems more reasonable

jzhang533 (approved):
lgtm
will merge and start testing

luotao1 (approved):
LGTM for __init__.py

PR #26018, opened 2020-08-06
Use static local variable instead of  global variable for safty

zhiqiu:
### PR types

Others
### PR changes

Others
### Describe

Use static local variable instead of  global variable for safty

### Reason:
The initialization and destruction of global and static global variable are not well ordered in c++, which may lead to hard-find-bugs. 
Use static local pointer instead, which is  initialized when first accessed, and destructed by when process exiting by the operating system, which is safe.


luotao1 (approved):
LGTM

chenwhql (approved):

PR #26036, opened 2020-08-07
[Paddle-TRT] Nvinfer plugin

zlsh80826:
### PR types

New features

### PR changes

Others

### Describe

Introduce the features that calls plugins of nvinfer_plugin directly
Currently implement gelu plugin with `CustomGeluPluginDynamic`
This feature can be enabled by passing the cmake option `WITH_NVINFER_PLUGIN=ON`


jeng1220:
@NHZlX , @cryoco,

This patch demonstrates that how to use dlopen to load plugins from libnvinfer_plugin.so. The benefit is that Paddle can use existing plugins made by NVIDIA TensorRT team to avoid reinventing the wheel again.

Also, the source code of libnvinfer_plugin.so is open, which named `TRTOSS`. Everyone is able to access it on GitHub (see https://github.com/NVIDIA/TensorRT). If Paddle has some custom requirements, for instance, use latest plugins on old TensorRT, just clone TRTOSS, then build TRTOSS on the top of old TensorRT. Or, it is possible to fork TRTOSS, then make small modification to meet Paddle's custom requirements.

It is strongly recommended to use this method to optimize Paddle-TRT in the future

PR #26052, opened 2020-08-08
Fix prelu for compatibility with saved model of old version

zhiqiu:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix prelu for compatibility with saved model of old version

In PR #25067, the shape of alpha in prelu was set to [1, channel] (instead of [1, channel, 1, 1]) if mode == 'channel', however it may result in compatibility problem with saved model of old version paddle.

It should be pointed out that [1, channel, 1, 1] is wrong since the input shape is not always of rank 4. We need to fix this API in paddle-2.0.


chenwhql (approved):
LGTM

phlrain (approved):

PR #26103, opened 2020-08-10
[API2.0] add Device api

wangchaochaohu:
### PR types

New features
### PR changes

OPs
### Describe

add set_device and get_device for Paddle


paddle/fluid/memory/memcpy.cc
```diff
@@ -70,6 +78,9 @@ void Copy<platform::CPUPlace, platform::CUDAPlace>(
       SyncCUDAStream();
     }
   }
+  if (change_device) {
+    platform::SetDeviceId(device_id);
```
willthefrog: I don't see why this is necessary, since `CUDADeviceContext` should set the proper  device for next operation.

paddle/fluid/pybind/pybind.cc
```diff
@@ -341,6 +341,11 @@ PYBIND11_MODULE(core_noavx, m) {
 
   m.def("set_num_threads", &platform::SetNumThreads);
 
+#ifdef PADDLE_WITH_CUDA
+  m.def("get_device_id", &platform::GetCurrentDeviceId);
+  m.def("set_device_id", &platform::SetDeviceId);
```
willthefrog: this is not needed either.

python/paddle/device.py
```diff
@@ -12,11 +12,82 @@
...
+    #            'cuda_places',
+    #            'CUDAPinnedPlace',
+    #            'CUDAPlace',
+    #            'is_compiled_with_cuda'
+]
+
+
+def set_device(device):
+    """
+    This function can determine whether the program is running on the CPU
```
willthefrog: Please refine the docs wording.

willthefrog (changes requested):
this PR is unnecessarily complicated for the desired purpose, these APIs are meant to be nothing more than a simple unified interface to set and get global default device.

at the lower level, paddle already has the device management logic in place, there is no need to change that.

the implementation should be very simple:
- `paddle.set_device(device)` sets a global variable for the default device (make sure user input is valid).
- `paddle.get_device()` reads this variable, if it is not set, fallback to 'gpu:0' if gpu is available ('cpu' otherwise).
- `imperative.guard(device=None)` and `fluid.Executor(device=None)`  should use `paddle.get_device()` to get the default device if `device` is not given.
- do the same for any other public API that requires user to provide a device.

willthefrog (commented):
to be more specific, paddle already has the global variable `framework._current_device` (see `device_guard`), but the dygraph tracer does not respect this variable, `Tracer::TraceOp()` needs to be modified so that it takes effect in dygraph too.

paddle/fluid/pybind/pybind.cc
wangchaochaohu: Done

python/paddle/device.py
wangchaochaohu: Done

python/paddle/fluid/framework.py
```diff
@@ -247,7 +247,19 @@ def _dygraph_tracer():
...
+    return _global_expected_place_
+
+
+def _set_dygraph_tracer_expected_place(place):
+    global _dygraph_tracer_
+    if _dygraph_tracer_ is not None:
+        _dygraph_tracer_._expected_place = place
+
+
+def _set_expected_place(place):
```
willthefrog, line 266: also call `_set_dygraph_tracer_expected_place`

python/paddle/fluid/framework.py
```diff
@@ -64,7 +64,7 @@
 CONTROL_DEP_VAR_PREFIX = core.kControlDepVarName()
 
 _dygraph_tracer_ = None
-_dygraph_current_expected_place_ = None
+_global_expected_place_ = None
 _current_device = None
```
willthefrog, line 68: remove `_current_device`, and make `device_guard` use `_set_expected_place` internally.

python/paddle/device.py
```diff
@@ -12,11 +12,90 @@
...
+     .. code-block:: python
+            
+        import paddle
+        paddle.enable_imperative()
+        device = paddle.fluid.dygraph.get_device()
+
+    """
+    device = ''
+    place = framework._current_expected_place()
+    if isinstance(place, core.CPUPlace):
```
willthefrog, line 88: move all these fallback logic into `_current_expected_place`?

python/paddle/device.py
```diff
@@ -12,11 +12,90 @@
...
+                            re.match(r'gpu:\d+', lower_device))
+        if not avaliable_device:
+            raise ValueError(
+                "The device must be a string which is like 'cpu' or 'gpu:0'")
+        device_info_list = device.split(':', 1)
+        device_id = device_info_list[1]
+        device_id = int(device_id)
+        place = core.CUDAPlace(device_id)
+        framework._set_expected_place(place)
+        framework._set_dygraph_tracer_expected_place(place)
```
willthefrog: `_set_dygraph_tracer_expected_place` should be called inside `_set_expected_place`

python/paddle/device.py
```diff
@@ -12,11 +12,90 @@
...
+     .. code-block:: python
+            
+        import paddle
+        paddle.enable_imperative()
+        paddle.fluid.dygraph.set_device("gpu:0")
+        x1 = paddle.ones(name='x1', shape=[1, 2], dtype='int32')
+        x2 = paddle.zeros(name='x2', shape=[1, 2], dtype='int32')
+        data = paddle.stack([x1,x2], axis=1)
+    """
+    lower_device = device.lower()
```
willthefrog, line 53: this should be a simple wrapper for `_set_expected_place`, move all these logic there.

python/paddle/fluid/dygraph/base.py
```diff
@@ -280,11 +275,16 @@ def guard(place=None):
     tracer = Tracer()
     VarBase = core.VarBase
 
+    expected_place = framework._current_expected_place()
+
     if place is None:
-        if core.is_compiled_with_cuda():
-            place = core.CUDAPlace(0)
+        if expected_place is None:
```
willthefrog: fold fallback logic into `_current_expected_place`, which then will never return `None`

willthefrog (changes requested):
suggestions to simplify the implementation:

- internally `_set_expected_place` and `_current_expected_place` are the sole entry points to set and get default device (which should handle all checking and fallback logic)
- `set_device` and `get_device` are the user facing frontend, which   makes use of `_set_expected_place` and `_current_expected_place` internally, and handles string -> place (and vice versa)  parsing
- `device_guard` should simply call `_set_expected_place` internally
- dygraph guard and executor should simply call `_current_expected_place` internally

paddle/fluid/pybind/pybind.cc
```diff
@@ -1321,6 +1321,8 @@ All parameter, weight, gradient are variables in Paddle.
              std::exit(-1);
 #endif
            })
+      .def("get_device_id",
```
willthefrog, line 1324: this is unnecessary.

python/paddle/fluid/framework.py
wangchaochaohu: device_guard is not the global place change, if we use _set_expected_place will change the device_gurad function

zhangting2020: It is recommended to change the device_guard in the next PR. If something goes wrong, it’s better to deal with it.

python/paddle/fluid/framework.py
```diff
@@ -247,7 +247,26 @@ def _dygraph_tracer():
 
 
 def _current_expected_place():
-    return _dygraph_current_expected_place_
+    global _global_expected_place_
+    if _global_expected_place_ is None:
+        if core.is_compiled_with_cuda():
+            _global_expected_place_ = core.CUDAPlace(0)
```
willthefrog, line 253: setting `_global_expected_place_` in a getter does not seem very clean...

willthefrog (changes requested):

python/paddle/device.py
wangchaochaohu: Done

python/paddle/device.py
wangchaochaohu: Done

python/paddle/fluid/framework.py
wangchaochaohu: Done

willthefrog (dismissed):
need subsequent PR for `device_guard` modification, otherwise LGTM

luotao1 (dismissed):
LGTM for framework.py

LielinJiang (approved):

luotao1 (approved):

PR #26105, opened 2020-08-10
add node num to RoleMaker; fix recompute Meta Optimizer;  raise error for undefined attr in distributeOptimizer 

mapingshuo:
### PR types

Bug fixes
### PR changes

APIs
### Describe

- add node num to RoleMaker; 
- fix recompute Meta Optimizer; 
- raise error for undefined attr in distributeOptimizer

### DIFF
--- a/python/paddle/fleet/base/role_maker.py
+++ b/python/paddle/fleet/base/role_maker.py
@@ -101,6 +101,14 @@ class RoleMakerBase(object):
         """
         raise NotImplementedError("Please implement this method in child class")
 
+    def node_num(self):
+        """
+        Get the training node number
+        Returns:
+            int: node num
+        """
+        raise NotImplementedError("Please implement this method in child class")
+
     def role_id(self):
         """
         Get current id.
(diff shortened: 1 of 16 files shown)


python/paddle/fleet/base/role_maker.py
```diff
@@ -286,6 +294,14 @@ def server_num(self):
             self.generate_role()
         return self._trainers_num
 
+    def node_num(self):
+        """
+        return the training node number
+        """
+        if not self._role_is_generated:
+            self.generate_role()
+        return self._node_num
```
seiriosPlus, line 303: `node_num` define twice?

python/paddle/fleet/base/role_maker.py
```diff
@@ -101,6 +101,14 @@ def server_index(self):
         """
         raise NotImplementedError("Please implement this method in child class")
 
+    def node_num(self):
+        """
+        Get the training node number
+        Returns:
+            int: node num
+        """
+        raise NotImplementedError("Please implement this method in child class")
```
seiriosPlus, line 110: what is the difference between `node_num` and `worker_num`

mapingshuo: > what is the difference between `node_num` and `worker_num`

For the machine with muti-GPU, the `node num` stands for the number of machines, and the `worker_num` stands for the number of GPU cards.

python/paddle/fleet/base/role_maker.py
mapingshuo: The first node_num function is defined in `RoleMakerBase `, and the second node_num function is defined in `PaddleCloudRoleMaker `

PR #26169, opened 2020-08-11
Fix loaded variable suffix repeat error

chenwhql:
### PR types

Bug fixes
### PR changes

Others
### Describe

when we load same model multiple times, if we only keep one `@LOADED` suffix, there will be parameters with same name, this PR fix this problem.

```
class MultiLoadingLinearNet(fluid.dygraph.Layer):
    def __init__(self, size, model_path):
        super(MultiLoadingLinearNet, self).__init__()
        self._linear = Linear(size, size)
        self._load_linear1 = fluid.dygraph.jit.load(model_path)
        self._load_linear2 = fluid.dygraph.jit.load(model_path)

    @declarative
    def forward(self, x):
        tmp1 = self._linear(x)
        tmp2 = self._load_linear1(tmp1)
        tmp3 = self._load_linear2(tmp2)
        y = self._linear(tmp3)
        return y
```

- original:
image

- new:
image

### DIFF
--- a/paddle/fluid/framework/operator.h
+++ b/paddle/fluid/framework/operator.h
@@ -64,9 +64,6 @@ constexpr char kZeroVarSuffix[] = "@ZERO";
 /// Variables with this suffix are the new Gradient.
 constexpr char kNewGradSuffix[] = "@NEWGRAD@";
 
-/// Variables with this suffix are the loaded from pre-train model.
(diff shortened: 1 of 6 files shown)


liym27 (approved):
LGTM

zhhsplendid (approved):
LGTM

luotao1 (approved):
LGTM for operator.h

PR #26180, opened 2020-08-12
[API2.0] add op for cudnn version query test=develop

wangchaochaohu:
### PR types

New features
### PR changes

APIs
### Describe

Add cudnn version query: def get_cudnn_version
this api return the version of cudnn.


python/paddle/device.py
```diff
@@ -13,10 +13,25 @@
...
+    #            'CPUPlace',
+    #            'cuda_pinned_places',
+    #            'cuda_places',
+    #            'CUDAPinnedPlace',
+    #            'CUDAPlace',
+    #            'is_compiled_with_cuda'
+]
+
+
+def cudnn_version():
```
willthefrog: export a variable instead? e.g.,

```python
cudnn_version = _cudnn_version()
```

paddle/fluid/platform/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ ENDIF()
 cc_library(cpu_info SRCS cpu_info.cc DEPS ${CPU_INFO_DEPS})
 cc_test(cpu_info_test SRCS cpu_info_test.cc DEPS cpu_info)
 
-nv_library(gpu_info SRCS gpu_info.cc DEPS gflags glog enforce monitor)
+nv_library(gpu_info SRCS gpu_info.cc DEPS gflags glog enforce monitor dynload_cuda)
```
willthefrog, line 48: any reason behind this change?

python/paddle/fluid/tests/unittests/test_query_op.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class TestCudnnVersion(unittest.TestCase):
+    def test_no_cudnn(self):
+        if not core.is_compiled_with_cuda():
+            cudnn_version = paddle.cudnn_version()
+            self.assertEqual((cudnn_version is None), True)
+
+    def test_cudnn(self):
+        cudnn_version = paddle.cudnn_version()
```
willthefrog: incomplete test?

willthefrog (changes requested):

paddle/fluid/platform/CMakeLists.txt
wangchaochaohu: this is because we use Hascudnn in gpu_info

python/paddle/fluid/tests/unittests/test_query_op.py
wangchaochaohu: wo don't know ci machine cudnn version

python/paddle/device.py
wangchaochaohu: this fucntion will be exposed to user?

willthefrog: scratch that, python does not have `@property` mechanism for module

paddle/fluid/platform/CMakeLists.txt
willthefrog: so an existing bug unrelated to this PR?

python/paddle/device.py
willthefrog: rename to `get_cudnn_version()` to signify that it is a function not a variable?

python/paddle/fluid/tests/unittests/test_query_op.py
willthefrog: at least check if it returns an integer?

python/paddle/fluid/tests/unittests/test_query_op.py
```diff
@@ -0,0 +1,33 @@
...
+
+from __future__ import print_function
+
+import unittest
+import paddle
+from paddle.fluid import core
+
+
+class TestCudnnVersion(unittest.TestCase):
+    def test_no_cudnn(self):
```
willthefrog, line 23: this can be merged into the test below, since you can not test both case in one go.

willthefrog (changes requested):

python/paddle/fluid/tests/unittests/test_query_op.py
wangchaochaohu: Done

python/paddle/fluid/tests/unittests/test_query_op.py
wangchaochaohu: Done

python/paddle/device.py
wangchaochaohu: Done

willthefrog (commented):
maybe give some explanation of the return value, something like 7600 -> 7.6?  otherwise LGTM.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -73,6 +74,10 @@ int GetCUDADeviceCount() {
   return dev_cnt;
 }
 
+/* Here is a very simple CUDA “pro tip”: cudaDeviceGetAttribute() is a much
```
willthefrog, line 77: why is this comment here?  looks like that `cudaDeviceGetAttribute` is not used anyway

willthefrog (changes requested):

python/paddle/device.py
```diff
@@ -13,10 +13,25 @@
...
+    #            'CPUPlace',
+    #            'cuda_pinned_places',
+    #            'cuda_places',
+    #            'CUDAPinnedPlace',
+    #            'CUDAPlace',
+    #            'is_compiled_with_cuda'
+]
+
+
+def get_cudnn_version():
```
willthefrog, line 36: the return value should be cached, as it will never change and this function may be called on each step for each OP that use it in dygraph mode.

willthefrog (changes requested):

paddle/fluid/platform/gpu_info.cc
wangchaochaohu: This comment is for GetCUDAComputeCapability, I change it to right position by the way

python/paddle/device.py
wangchaochaohu: Done

willthefrog (dismissed):
LGTM

heavengate (dismissed):

LielinJiang (approved):

PR #25898, opened 2020-08-03
Paddle-2.0 API directory migration

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

Paddle 2.0 mainly promotes dygraph mode. The directory level of dygraph mode needs to be moved up one level, and the APIs of static mode are moved to the 'static' directory.

1. remove imperative directory
```
# dynamic and static mode switch API
paddle.enable_imperative  -> paddle.disable_static 
paddle.disable_imperative -> paddle.enable_static
paddle.in_imperative_mode -> paddle.in_dynamic_mode

# dynamic graph related API
paddle.imperative.to_variable -> paddle.to_variable
paddle.imperative.enable -> deleted
paddle.imperative.guard -> deleted
paddle.imperative.grad -> paddle.grad
paddle.imperative.no_grad -> paddle.no_grad
paddle.imperative.save -> paddle.save
paddle.imperative.load -> paddle.load
paddle.imperative.BackwardStrategy -> paddle.BackwardStrategy

# data_parallel related API
paddle.imperative.ParallelEnv -> paddle.ParallelEnv
paddle.imperative.prepare_context -> paddle.prepare_context
paddle.imperative.DataParalell -> paddle.DataParallel

# learning_rate_decay related API
paddle.imperative.NoamDecay -> paddle.NoamDecay
paddle.imperative.PiecewiseDecay -> paddle.PiecewiseDecay
paddle.imperative.NaturalExpDecay -> paddle.NaturalExpDecay
paddle.imperative.ExponentialDecay -> paddle.ExponentialDecay
paddle.imperative.InverseTimeDecay -> paddle.InverseTimeDecay
paddle.imperative.PolynomialDecay -> paddle.PolynomialDecay
paddle.imperative.CosineDecay -> paddle.CosineDecay

# dy_to_static related API
paddle.imperative.jit -> paddle.jit
paddle.imperative.TracedLayer -> paddle.jit.TracedLayer
paddle.imperative.declarative -> paddle.jit.to_static
paddle.imperative.ProgramTranslator -> paddle.jit.PragramTranslator
paddle.imperative.TranslatedLayer -> paddle.jit.TranslatedLayer
paddle.imperative.jit.save -> paddle.jit.save
paddle.imperative.jit.load -> paddle.jit.load
paddle.imperative.jit.SaveLoadConfig -> paddle.jit.SaveLoadConfig
```

2. Static graph related API in `paddle.framework`, moved to `paddle.static`
```
paddle.save -> paddle.static.save
[...]

PR #26195, opened 2020-08-12
Remove original name for alias APIs

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

In the implementation of Paddle 2.0 directory migration（PR #25898）, we both import the alias name and original name for alias APIs. For example, in paddle/\__init__.py
```
from .fluid.dygraph.base import enable_dygraph  #DEFINE_ALIAS
from .fluid.dygraph.base import disable_dygraph  #DEFINE_ALIAS
from .fluid.dygraph.base import enable_dygraph as disable_static  #DEFINE_ALIAS
from .fluid.dygraph.base import disable_dygraph as enable_static  #DEFINE_ALIAS
```
Importing original name is unnecessary, however, the algorithm used to print all the specifications of APIs in `print_signatures.py` use the original name. If we don't import original name, CI can't pass.
After solving the problem of `print_signatures.py` in  PR #26143, we can remove the unnecessary original name. This PR is used to remove the original name for alias APIs.

### DIFF
--- a/python/paddle/__init__.py
+++ b/python/paddle/__init__.py
@@ -210,8 +210,6 @@ from .framework import BackwardStrategy  #DEFINE_ALIAS
 from .framework import to_variable  #DEFINE_ALIAS
 from .framework import grad  #DEFINE_ALIAS
 from .framework import no_grad  #DEFINE_ALIAS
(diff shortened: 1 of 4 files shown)


zhiqiu (approved):
LGTM

PR #26208, opened 2020-08-12
Change pairwise distance api name eps to epsilon

ZHUI:
### PR types

Bug fixes
### PR changes

APIs
### Describe

change the pairwise distance api name eps to epsilon


wawltor (approved):
LGTM

PR #26220, opened 2020-08-12
support the tuple for attribute of axis in min, max for api2.0

wawltor:
### PR types
Bug fixes 
### PR changes
 OPs

### Describe
support the tuple for attribute of axis in min, max


FlyingQianMM (approved):
LGTM

PR #26231, opened 2020-08-13
Use a faster algorithm for complex matmul, test=develop

kuke:
### PR types

Performance optimization

### PR changes

APIs

### Describe

For the matmul operation of complex tensors, there exist a faster algorithm which reduces the times of multiplication by one 

```
(a+bj)(c+dj)=(ac-bd)+(bc+ad)j 
set P1=ac; P2=(a+b)(c+d); P3=bd; then (a+bj)(c+dj)=(P1-P3)+(P2-P1-P3)j
```


iclementine (approved):
LGTM

PR #26238, opened 2020-08-13
Refine deprecated decorator

zhiqiu:
### PR types

Others
### PR changes

APIs
### Describe

Refine deprecated decorator, show deprecation warning once.

For the following case, the original  `deprecated` on `reduce_mean` will print too many warnings, this PR fixes that.
```
import paddle

def model(x):
    return paddle.fluid.layers.reduce_mean(x)   # reduce_mean is decorated with @deprecated

paddle.disable_static()
for epoch in range(5):
    for batch in range(100):
        x = paddle.rand([2, 10])
        loss = model(x)
        if batch % 100 == 0:
            print(loss)
```

- before
image

- after
image

### DIFF
--- a/python/paddle/utils/deprecated.py
+++ b/python/paddle/utils/deprecated.py
@@ -18,7 +18,6 @@ decorator to deprecate a function or class
 import warnings
 import functools
 import paddle
-import sys
 
 # NOTE(zhiqiu): Since python 3.2, DeprecationWarning is ignored by default,
 # and since python 3.7, it is once again shown by default when triggered directly by code in __main__.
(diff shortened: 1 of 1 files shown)


willthefrog (changes requested):
consider printing only once for each API? e.g., by adding a set that keeps track of the APIs for which warning are already printed

zhiqiu:
> consider printing only once for each API? e.g., by adding a set that keeps track of the APIs for which warning are already printed

The `default` filter of warnings provides it.

willthefrog (approved):
LGTM

PR #26246, opened 2020-08-13
【paddle.fleet】Set default value to strategy in distributed_optimizer

qjing666:
### PR types
Function optimization

### PR changes
APIs 

### Describe
Set strategy to default values when users do not want to define their own.


guru4elephant (approved):
LGTM

PR #26254, opened 2020-08-13
[OpDevOptimize] add common unary op

zhiqiu:
### PR types

Others
### PR changes

OPs
### Describe

Add common unary op

#### Usage

For example, if we want to write a new operator, `y = -(x)`.
We can simply write like this, 
```
class TestNegOpMaker : public framework::OpProtoAndCheckerMaker {
 public:
  void Make() override {
    AddInput("X", "input of test op");
    AddOutput("Out", "output of test op");
    AddComment("Out = -X");
  }
};

template <typename T>
class TestNegOpGradMaker : public framework::SingleGradOpMaker<T> {
 public:
  using framework::SingleGradOpMaker<T>::SingleGradOpMaker;

 protected:
  void Apply(GradOpPtr<T> op) const override {
    op->SetType(this->ForwardOpType() + "_grad");
    op->SetInput(framework::GradVarName("Out"), this->OutputGrad("Out"));
    op->SetOutput(framework::GradVarName("X"), this->InputGrad("X"));
    op->SetAttrMap(this->Attrs());
  }
};

}  // namespace operators
}  // namespace paddle

namespace functors = paddle::operators::functors;

REGISTER_OPERATOR(test_neg, ops::UnaryOp, ops::TestNegOpMaker,
                  ops::TestNegOpGradMaker<paddle::framework::OpDesc>,
                  ops::TestNegOpGradMaker<paddle::imperative::OpBase>);
REGISTER_OPERATOR(test_neg_grad, ops::UnaryGradOp);
REGISTER_OP_KERNEL_4(test_neg, ops::UnaryOpKernel, CPU, functors::Neg, int,
                     int64_t, float, double);
REGISTER_OP_KERNEL_4(test_neg_grad, ops::UnaryGradOpKernel, CPU, functors::Neg,
                     int, int64_t, float, double);
```
As the example shown, the implementation of `neg` op can be easy and simple enough.

Plus, if we want to implement another unary op, we only need to change the `OpMaker`, `op_name`, and `functor`.

### DIFF
--- a/paddle/fluid/framework/op_desc.cc
+++ b/paddle/fluid/framework/op_desc.cc
@@ -54,26 +54,24 @@ class CompileTimeInferShapeContext : public InferShapeContext {
   std::vector<std::string> Outputs(const std::string &name) const override;
 
   std::string GetInputNameByIdx(size_t idx) const override {
-    auto &op_proto =
-        paddle::framework::OpInfoMap::Instance().Get(op_.Type()).proto_;
-    PADDLE_ENFORCE_LT(idx, op_proto->inputs().size(),
+    auto input_names = op_.OrderedInputNames();
+    PADDLE_ENFORCE_LT(idx, input_names.size(),
                       platform::errors::OutOfRange(
                           "The index should be less than the size of inputs of "
                           "operator %s, but got index is %d and size is %d",
-                          op_.Type(), idx, op_proto->inputs().size()));
-    return op_proto->inputs()[idx].name();
+                          op_.Type(), idx, input_names.size()));
+    return input_names[idx];
   }
 
   std::string GetOutputNameByIdx(size_t idx) const override {
-    auto &op_proto =
-        paddle::framework::OpInfoMap::Instance().Get(op_.Type()).proto_;
+    auto output_names = op_.OrderedOutputNames();
     PADDLE_ENFORCE_LT(
-        idx, op_proto->outputs().size(),
+        idx, output_names.size(),
         platform::errors::OutOfRange(
             "The index should be less than the size of outputs of "
             "operator %s, but got index is %d and size is %d",
-            op_.Type(), idx, op_proto->outputs().size()));
-    return op_proto->outputs()[idx].name();
+            op_.Type(), idx, output_names.size()));
+    return output_names[idx];
   }
 
   void ShareDim(const std::string &in, const std::string &out, size_t i = 0,
@@ -303,7 +301,7 @@ OpDesc::OpDesc(const std::string &type, const VariableNameMap &inputs,
   outputs_ = outputs;
   attrs_ = attrs;
   need_update_ = true;
-  // TODO(zhiqiu): set input_names_ and output_names_ properly here
+  InitOrderedInputOutputNamesForForwardOp();
   block_ = nullptr;
 }
 
@@ -358,6 +356,7 @@ OpDesc::OpDesc(const proto::OpDesc &desc, BlockDesc *block)
     }
   }
   this->block_ = block;
+  InitOrderedInputOutputNamesForForwardOp();
 }
 
 proto::OpDesc *OpDesc::Proto() {
@@ -384,7 +383,11 @@ void OpDesc::SetInput(const std::string &param_name,
                       const std::vector<std::string> &args) {
   need_update_ = true;
   inputs_[param_name] = args;
-  input_names_.emplace_back(param_name);
+  // Only add input_names_ for grad op
+  // For forward op, use the input names in its proto directly
+  if (!IsOpInfoHasProto(Type())) {
+    input_names_.emplace_back(param_name);
+  }
 }
 
 const std::vector<std::string> &OpDesc::Output(const std::string &name) const {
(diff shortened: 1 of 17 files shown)


paddle/fluid/framework/op_desc.cc
```diff
@@ -303,6 +288,7 @@ OpDesc::OpDesc(const std::string &type, const VariableNameMap &inputs,
   outputs_ = outputs;
   attrs_ = attrs;
   need_update_ = true;
+  InitOrderedInputOutputNamesForForwardOp();
```
chenwhql, line 291: moving to the back of `block_ = nullptr;` look better?

paddle/fluid/framework/op_desc.h
```diff
@@ -27,6 +29,13 @@ namespace framework {
 
 class BlockDesc;
 class ProgramDesc;
+
+// If an op's OpInfo has proto, it is a forward op; otherwise, it is a grad op
+static inline bool IsOpInfoHasProto(std::string type) {
```
chenwhql, line 34: use `const std::string& type`?

paddle/fluid/framework/op_desc.h
```diff
@@ -138,6 +150,50 @@ class OpDesc {
...
+        }
+      }
+    }
+  }
+
+  std::string GetInputNameByIdx(size_t idx) const {
+    PADDLE_ENFORCE_LT(idx, input_names_.size(),
+                      platform::errors::OutOfRange(
+                          "The index should be less than the size of inputs of "
+                          "operator %s, but got index is %d and size is %d",
```
chenwhql, line 182: `and inputs size is %d?` is better? 

I found that `%d` may not be very good here, and there may be a lot of similar problems in the current error messgae in paddle...〒_〒

https://zh-google-styleguide.readthedocs.io/en/latest/google-cpp-styleguide/others/#id12

paddle/fluid/framework/op_desc.h
```diff
@@ -138,6 +150,50 @@ class OpDesc {
...
+                          Type(), idx, input_names_.size()));
+    return input_names_[idx];
+  }
+
+  std::string GetOutputNameByIdx(size_t idx) const {
+    PADDLE_ENFORCE_LT(
+        idx, output_names_.size(),
+        platform::errors::OutOfRange(
+            "The index should be less than the size of outputs of "
+            "operator %s, but got index is %d and size is %d",
```
chenwhql, line 192: same above

paddle/fluid/framework/op_registry.cc
```diff
@@ -21,12 +21,16 @@ namespace framework {
 
 std::unique_ptr<OperatorBase> OpRegistry::CreateOp(
     const std::string& type, const VariableNameMap& inputs,
-    const VariableNameMap& outputs, AttributeMap attrs, bool attr_check) {
+    const VariableNameMap& outputs, AttributeMap attrs,
+    const std::vector<std::string>& input_names,
```
chenwhql, line 25: Although I understand it is difficult here, but the `input_names` info already  in the `inputs` and only the order information is missing, in this case the input information is redundant and not a good design. Can we find a way to implement an efficient `ordered_map` here?

for example:
maybe we can define a `paddle::ordered_map`, it hold a `std::unordered_map<int, vector<string>> dict_` and a `std::vector<std::string> names_`, we can find name from vector and get index to map the vector<string>, then replace all `VariableNameMap`.

I just think design like this is better, but the efficiency is not necessarily the highest, maybe we can think other ways

paddle/fluid/framework/op_registry.cc
```diff
@@ -45,7 +49,8 @@ static VariableNameMap ConvertOpDescVarsToVarNameMap(
 }
 
 std::unique_ptr<OperatorBase> OpRegistry::CreateOp(
-    const proto::OpDesc& op_desc) {
+    const proto::OpDesc& op_desc, const std::vector<std::string>& input_names,
```
chenwhql, line 52: same above, `op_desc` also hold the `input_names_` info

paddle/fluid/framework/op_registry.h
```diff
@@ -90,15 +91,20 @@ class OpRegistry {
...
-                                                const VariableNameMap& inputs,
-                                                const VariableNameMap& outputs,
-                                                AttributeMap attrs,
-                                                bool attr_check = true);
-
-  static std::unique_ptr<OperatorBase> CreateOp(const proto::OpDesc& op_desc);
+  static std::unique_ptr<OperatorBase> CreateOp(
+      const std::string& type, const VariableNameMap& inputs,
+      const VariableNameMap& outputs, AttributeMap attrs,
+      const std::vector<std::string>& input_names = {},
```
chenwhql, line 97: maybe if we write this way, we can solve the current problem, but it may increase the cost of reconstruction in the future

paddle/fluid/framework/op_registry.cc
zhiqiu: Exactly, yes. It could be next work, to replace the VariableNameMap with ordered_map.
Currently, I just want to enhance the ability of `getting-name-by-index` for several `contexts` in paddle. And, only with that the abstraction of common infers hape fucntions, operators can be done.

paddle/fluid/framework/op_registry.cc
zhiqiu: `OpDesc` could be constructed without `input_names`, for example, in unit test.

paddle/fluid/framework/op_registry.h
zhiqiu: This is a way to `add` some features without changing existing operators. I agree that refactoring `VariableNameMap` may be a better way but it requires modifying all operators. It is a trade-off.

PR #26266, opened 2020-08-14
fix the margin ranking loss doc and api, test=develop

wawltor:
### PR types
Others

### PR changes
APIs

### Describe
update the target to the label, change the out to output


zhiqiu (approved):
LGTM

PR #26283, opened 2020-08-14
[Paddle-TRT] add TensorRT getInferLibVersion

jeng1220:
### PR types

Bug fixes
### PR changes

Others
### Describe

Because Paddle uses dlopen to load TensorRT at runtime, the TensoRT
version might not be as same as the version that built with Paddle.
In original implementation, Paddle always throws warning message "
The Paddle lib links the XXX version TensorRT, make sure the runtime
TensorRT you are using is no less than this verion, otherwise, there
might be Segfault!" However, it is still unclear to user.

This patch adds `getInferLibVersion` to check which version of TensorRT
is loaded by dlopen at runtime. If versions are not consistent. Paddle
will throw "This Paddle lib was built with TensorRT XXX, but this
system is using TensorRT YYY. It is recommended to use a consistent
TensorRT version. Otherwise, there might be Segfault!"


jeng1220:
@NHZlX @cryoco
Plz take a look

shangzhizhou (approved):
LGTM

jeng1220:
@cryoco ,
The develop has implemented the similar feature, so we don't need to merge this PR

PR #26294, opened 2020-08-14
Add PADDLE_ENFORCE  in nll loss cuda kernel

joey12300:
### PR types

Bug fixes 
### PR changes

OPs
### Describe

Add a check whether label value is out of bound in cuda kernel. If it is, throw an exception.


wawltor (approved):
LGTM

PR #26401, opened 2020-08-18
mean: not support int32, int64; add check for axis

zhupengyang:
### PR types

Others

### PR changes

OPs

### Describe

- The mean result of data with type int32 or int64 is wrong, so we erase the register for int32 and int64.
- add more check for axis in reduce ops


XiaoguangHu01 (approved):
LGTM

TCChenlong (approved):
LGTM

PR #26420, opened 2020-08-18
specify cuda arch when dectected failed

zhwesky2010:
### PR types

Others
### PR changes

Others
### Describe

specify cuda arch when dectected failed. 
- '61' for Nvidia GTX1080/1080Ti.
- '75' for Nvidia RTX 2080

It will reduce compile time Defaultly, about 50-60min. But if you want copilme all CUDA Architecture, you should add -DCUDA_ARCH=ALL when cmake, this will cost 90+ min.


luotao1 (approved):
LGTM

PR #26426, opened 2020-08-19
Fix erf

WuHaobo:
### PR types
Others

### PR changes
APIs

### Describe
1. add `name` args for erf
2. polish the English docs for erf


shippingwang (approved):
LG

PR #26483, opened 2020-08-20
[API2.0]Unify pooling function and add adaptive max pooling function

shippingwang:
### PR types

New features

### PR changes

APIs

### Describe

This PR
- add adaptive max pooling 2d 3d function
- Uniform paddle ```pooling``` function, extract common function 1. ```_update_padding_nd``` which is consistent with ```conv``` function, ``` 2. _expand_low_nd_padding``` to implement low dimension function by calling high dimension function(unsqueeze nd to n+1 d)
- solve #26331 and #26369 problem 
   - fix typo
   - uniform name and order
   - add deprecated fluid function
   - add ```Shape``` field in layer/pooling.py
   - add ref link in functional/pooling.py
   - use ```x``` instead of ```input```, ```out``` instead of ```output```
- uniform padding description, which is consistent with ```conv``` function.
Example:
```
padding (string|int|list|tuple): The padding size. Padding could be in one of the following fo rms.
1. A string in ['valid', 'same'].
2. An int, which means the feature map is zero-padded by the size of `padding` on every side.
3. A list[int] or tuple(int) whose length is 1, which means the feature map is zero-padded by the size of `padding[0]` on every side.
4. A list[int] or tuple(int) whose length is 2. It has the form [pad_before, pad_after].
5. A list or tuple of pairs of integers. It has the form [[pad_before, pad_after], [pad_before, pad_after], ...]. Note that, the batch dimension and channel dimension should be [0,0] or (0,0).
he default value is 0.
```

- support padding both front and back in 1d function, although it is **NOT** consistent with PyTorch, It is same as PaddlePaddle conv 1d function and other pooling Nd functions.
- separate adaptive 1/2/3d max/avg unittest


python/paddle/nn/functional/pooling.py
```diff
@@ -146,99 +168,69 @@ def avg_pool1d(x,
...
+    See more details in :ref:`api_nn_pooling_AvgPool1d` .
 
     Args:
         x (Tensor): The input tensor of pooling operator which is a 3-D tensor with
                           shape [N, C, L]. where `N` is batch size, `C` is the number of channels,
-                          `L` is the length of the feature. The data type if float32 or float64.
+                          `L` is the length of the feature. The data type is float32 or float64.
         kernel_size (int|list|tuple): The pool kernel size. If pool kernel size is a tuple or list,
-            it must contain one integers.
+            it must contain an integer.
```
willthefrog, line 180: integers?

python/paddle/nn/functional/pooling.py
```diff
@@ -146,99 +168,69 @@ def avg_pool1d(x,
...
         stride (int|list|tuple): The pool stride size. If pool stride size is a tuple or list,
-            it must contain one integers.
-        padding (string|int|list|tuple): The pool padding. If `pool_padding` is a string, either 'VALID' or
-            'SAME' which is the padding algorithm. If pool padding size is a tuple or list,
-            it could be the following forms: `[pad_left, pad_right]`. If padding is non-zero,
-            then the input is implicitly zero-padded on both sides for padding number of points.
+            it must contain an integer.
```
willthefrog, line 182: integers?

python/paddle/nn/functional/pooling.py
```diff
@@ -146,99 +168,69 @@ def avg_pool1d(x,
...
+        padding (string|int|list|tuple): The padding size. Padding could be in one of the following forms.
+            1. A string in ['valid', 'same'].
```
willthefrog, line 184: case insensitive?

python/paddle/nn/functional/pooling.py
```diff
@@ -146,99 +168,69 @@ def avg_pool1d(x,
...
-            padding = [0]
+    padding, padding_algorithm = _update_padding_nd(
+        padding, 1, channel_last=False, ceil_mode=ceil_mode)
 
-    padding = update_padding1d(padding, "avg")
+    # use 2d to implenment 1d should expand padding in advance.
+    padding = _expand_low_nd_padding(padding)
 
     if in_dygraph_mode():
         output = core.ops.pool2d(
```
willthefrog, line 237: `use_cudnn` is different? 
also,  document the reasoning of `use_cudnn` default value.

python/paddle/nn/layer/pooling.py
```diff
@@ -284,231 +119,291 @@ def forward(self, x):
...
+    def forward(self, x):
+        return F.avg_pool3d(
+            x,
+            kernel_size=self.ksize,
+            stride=self.stride,
+            padding=self.padding,
+            ceil_mode=self.ceil_mode,
+            count_include_pad=self.count_include_pad,
+            divisor_override=self.divisor,
+            data_format=self.data_format,
```
willthefrog, line 312: `data_format` works?

WuHaobo (dismissed):
LGTM

python/paddle/nn/functional/pooling.py
```diff
@@ -146,99 +168,69 @@ def avg_pool1d(x,
                count_include_pad=True,
                ceil_mode=False,
                name=None):
```
willthefrog, line 170: missing data format?

willthefrog (changes requested):
- double check `data_format` 
- double check `use_cudnn`, also document reasoning behind `use_cudnn` default value choice.
- also for future references, it would be nice to have some preliminary benchmark on 1d pooling in PR description

python/paddle/nn/functional/pooling.py
shippingwang: It looks like a merged bug https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/nn/functional/pooling.py#L247 , someone mixed up `use_cudnn` and `exclusive` params, I should and already correct it in the next commit.

python/paddle/nn/functional/pooling.py
shippingwang: yes, apply upper function after passing str params

python/paddle/nn/functional/pooling.py
shippingwang: It supposes to add in this PR https://github.com/PaddlePaddle/Paddle/pull/26331, hmm I suggest refining it in the next PR

python/paddle/nn/functional/pooling.py
shippingwang: 1d is an integer

python/paddle/nn/functional/pooling.py
shippingwang: 1d is an integer

python/paddle/nn/layer/pooling.py
shippingwang: yes

willthefrog (commented):
check weather custom pooling implementation is faster than cudnn for adaptive/global
- if both faster, set `use_cudnn` to false for adaptive and global
- if global is faster, set `global_pooling` to true and `use_cudnn` to false for adaptive pooling with output size 1

shippingwang:
> check weather custom pooling implementation is faster than cudnn for adaptive/global
> 
> * if both faster, set `use_cudnn` to false for adaptive and global
> * if global is faster, set `global_pooling` to true and `use_cudnn` to false for adaptive pooling with output size 1

I will compare the performance of  ```core.ops.[max/avg]_pool[2/3]d_with_index()```, ```core.ops.pool[2/3]d (use_cudnn=Ture, use global_pooling=True)```, and ```core.ops.pool[2/3]d(use_cudnn=False, use_global_pooling=True)``` when the output size is 1 in ```adaptive [max/avg] pool[1/2/3]d```  function in next PR. 

please feel free to review it again while CI is passing

willthefrog (approved):
need follow up for data format and cudnn, otherwise LGTM

python/paddle/fluid/tests/unittests/test_adaptive_avg_pool1d.py
```diff
@@ -0,0 +1,119 @@
...
+
+def adaptive_start_index(index, input_size, output_size):
+    return int(np.floor(index * input_size / output_size))
+
+
+def adaptive_end_index(index, input_size, output_size):
+    return int(np.ceil((index + 1) * input_size / output_size))
+
+
+def avg_pool1D_forward_naive(x,
```
XiaoguangHu01, line 35: 1D -> 1d

XiaoguangHu01 (approved):
LG API

phlrain (approved):
LGTM for core.ops

raindrops2sea (approved):

PR #25774, opened 2020-07-28
support mnist and resnet dygraph_to_static test

sfraczek:
### PR types
New features 

### PR changes
Others

### Describe
Add a test with mkldnn to `dygraph_to_static` tests of mnist and resnet. It depends on other dygraph PRs.

PR #26497, opened 2020-08-20
Add use of global flag 'use_mkldnn' to layer_helper

arlesniak:
### PR types

New features
### PR changes

Others
### Describe

Get use of global  FLAGS_use_mkldnn to append_activation in layer_helper

### DIFF
--- a/paddle/fluid/imperative/tracer.cc
+++ b/paddle/fluid/imperative/tracer.cc
@@ -50,8 +50,7 @@ void Tracer::TraceOp(const std::string& type, const NameVarBaseMap& ins,
                      const platform::Place& place, bool trace_backward) {
   VLOG(1) << "Trace Op: " << type;
   if (FLAGS_use_mkldnn) {
-    auto& mutable_op_attrs = const_cast<framework::AttributeMap&>(attrs);
-    mutable_op_attrs["use_mkldnn"] = true;
+    attrs["use_mkldnn"] = true;
   }
   auto op = framework::OpRegistry::CreateOp(type, {}, {}, {}, false);
   const auto& op_info = op->Info();
--- a/paddle/fluid/imperative/tracer.cc
+++ b/paddle/fluid/imperative/tracer.cc
@@ -21,6 +21,8 @@
 #include "paddle/fluid/platform/profiler.h"
 #include "paddle/fluid/string/string_helper.h"
 
+DECLARE_bool(use_mkldnn);
+
 namespace paddle {
 namespace imperative {
 
@@ -47,6 +49,10 @@ void Tracer::TraceOp(const std::string& type, const NameVarBaseMap& ins,
                      const NameVarBaseMap& outs, framework::AttributeMap attrs,
                      const platform::Place& place, bool trace_backward) {
   VLOG(1) << "Trace Op: " << type;
+  if (FLAGS_use_mkldnn) {
+    auto& mutable_op_attrs = const_cast<framework::AttributeMap&>(attrs);
+    mutable_op_attrs["use_mkldnn"] = true;
+  }
   auto op = framework::OpRegistry::CreateOp(type, {}, {}, {}, false);
   const auto& op_info = op->Info();
   auto* attr_checker = op_info.Checker();
--- a/python/paddle/fluid/layer_helper.py
+++ b/python/paddle/fluid/layer_helper.py
@@ -29,6 +29,8 @@ from .layer_helper_base import LayerHelperBase
 class LayerHelper(LayerHelperBase):
     def __init__(self, layer_type, **kwargs):
         self.kwargs = kwargs
+        self.debug_ = self.kwargs.get('debug', False)
+        self.appended_op = None
         name = self.kwargs.get('name', None)
         # TODO(panyx0718, minqiyang): dygraph mode
         # can not use both `layer_type` and `name`. Deprecate LayerHelper
@@ -40,7 +42,10 @@ class LayerHelper(LayerHelperBase):
             self.kwargs['name'], layer_type=layer_type)
 
     def append_op(self, *args, **kwargs):
-        return self.main_program.current_block().append_op(*args, **kwargs)
+        op = self.main_program.current_block().append_op(*args, **kwargs)
+        if self.debug_:
+            self.appended_op = op
+        return op
 
     def multiple_input(self, input_param_name='input'):
         inputs = self.kwargs.get(input_param_name, [])
@@ -151,7 +156,7 @@ class LayerHelper(LayerHelperBase):
         # priority order: use_mkldnn in kwargs then global
         use_mkldnn = self.kwargs.get(
             'use_mkldnn', core.globals().get("FLAGS_use_mkldnn", False))
-        if (use_mkldnn is not None) and use_mkldnn:
+        if use_mkldnn:
             act['use_mkldnn'] = use_mkldnn
 
         act_type = act.pop('type')
--- a/python/paddle/fluid/layer_helper.py
+++ b/python/paddle/fluid/layer_helper.py
@@ -148,9 +148,9 @@ class LayerHelper(LayerHelperBase):
         if 'use_cudnn' in self.kwargs and self.kwargs.get('use_cudnn'):
             act['use_cudnn'] = self.kwargs.get('use_cudnn')
 
-        use_mkldnn = core.globals()["FLAGS_use_mkldnn"]
-        if 'use_mkldnn' in self.kwargs:
-            use_mkldnn = self.kwargs.get('use_mkldnn')
+        # priority use_mkldnn in kwargs then global
+        use_mkldnn = self.kwargs.get(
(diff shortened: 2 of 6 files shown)


arlesniak:
PR-CI-Coverage does not update status but after clicking Details it's all green

python/paddle/fluid/layer_helper.py
```diff
@@ -147,8 +147,13 @@ def append_activation(self, input_var):
 
         if 'use_cudnn' in self.kwargs and self.kwargs.get('use_cudnn'):
             act['use_cudnn'] = self.kwargs.get('use_cudnn')
-        if 'use_mkldnn' in self.kwargs:
-            act['use_mkldnn'] = self.kwargs.get('use_mkldnn')
+
+        # priority order: use_mkldnn in kwargs then global
+        use_mkldnn = self.kwargs.get(
+            'use_mkldnn', core.globals().get("FLAGS_use_mkldnn", False))
+        if (use_mkldnn is not None) and use_mkldnn:
```
wojtuss: This is equivalent to `if use_mkldnn:`.

luotao1:
> PR-CI-Coverage does not update status but after clicking Details it's all green

Some network error. If you will fix https://github.com/PaddlePaddle/Paddle/pull/26497#discussion_r478162539, you can push a new commit to trigger it.

grygielski (dismissed):
LGTM

luotao1:
```
2020-08-27 19:32:07 ****************
2020-08-27 19:32:07 0. You must have one RD (zhiqiu (Recommend) or phlrain) approval for the api change for the opreator-related api without 'core.ops'.
2020-08-27 19:32:07 For more details, please click [https://github.com/PaddlePaddle/Paddle/wiki/paddle_api_development_manual.md]
2020-08-27 19:32:07 Related APIs: paddle.common_ops_import.LayerHelper.append_activation
2020-08-27 19:32:07 paddle.common_ops_import.LayerHelper.append_op
2020-08-27 19:32:07 paddle.fluid.layers.utils.LayerHelper.append_activation
2020-08-27 19:32:07 paddle.fluid.layers.utils.LayerHelper.append_op
2020-08-27 19:32:07 
2020-08-27 19:32:07 There are 1 approved errors.
2020-08-27 19:32:07 ****************
```

python/paddle/fluid/layer_helper.py
arlesniak: Done

python/paddle/fluid/layer_helper.py
```diff
@@ -40,7 +42,10 @@ def __init__(self, layer_type, **kwargs):
             self.kwargs['name'], layer_type=layer_type)
 
     def append_op(self, *args, **kwargs):
-        return self.main_program.current_block().append_op(*args, **kwargs)
+        op = self.main_program.current_block().append_op(*args, **kwargs)
+        if self.debug_:
+            self.appended_op = op
+        return op
```
zhiqiu: It seems `appended_op` is only used to get the operator for testing its attrs. 
I think we don't need to add a member to LayerHelper, the first appended op (in your test case, only one op is appended) can be obtained by `fluid.default_main_program().global_block().ops[0] `.

python/paddle/fluid/tests/unittests/test_imperative_basic.py
```diff
@@ -640,6 +641,37 @@ def test_append_activation_in_dygraph3(self):
...
+        with fluid.dygraph.guard():
+            a = fluid.dygraph.to_variable(a_np)
+            res1 = func(a)
+            res2 = fluid.layers.relu(a)
+            self.assertTrue(helper.appended_op.attrs.get('use_mkldnn', False))
+            self.assertTrue(np.array_equal(res1.numpy(), res2.numpy()))
+
+    def test_append_activation_in_dygraph_global_use_mkldnn(self):
+        a_np = np.random.uniform(-2, 2, (10, 20, 30)).astype(np.float32)
+        fluid.set_flags({'FLAGS_use_mkldnn': True})
```
wojtuss: @luotao1 Here I have the same question as [here](https://github.com/PaddlePaddle/Paddle/pull/25774#discussion_r462963970), whether setting the global flag can interfere with other tests running in parallel on CI.
Other than that, LGTM.

python/paddle/fluid/layer_helper.py
```diff
@@ -147,8 +152,13 @@ def append_activation(self, input_var):
...
             act['use_cudnn'] = self.kwargs.get('use_cudnn')
-        if 'use_mkldnn' in self.kwargs:
-            act['use_mkldnn'] = self.kwargs.get('use_mkldnn')
+
+        # priority order: use_mkldnn in kwargs then global
+        use_mkldnn = self.kwargs.get(
+            'use_mkldnn', core.globals().get("FLAGS_use_mkldnn", False))
+        if use_mkldnn:
+            act['use_mkldnn'] = use_mkldnn
+
```
zhiqiu: Maybe it is better to implement this functionality in C++ side. There are 2 reasons,

1. `FLAGS_use_mkldnn ` is a GFLAG, which is declared and used in c++. 
2. There may be other way to add an activation op, for example, `_append_activation_in_dygraph`.

python/paddle/fluid/layer_helper.py
arlesniak: Thank you for the comment 
> fluid.default_main_program().global_block().ops

is empty in dygraph mode, I'm working on different solution.

grygielski (approved):
LGTM

paddle/fluid/imperative/tracer.cc
```diff
@@ -47,6 +49,9 @@ void Tracer::TraceOp(const std::string& type, const NameVarBaseMap& ins,
                      const NameVarBaseMap& outs, framework::AttributeMap attrs,
                      const platform::Place& place, bool trace_backward) {
   VLOG(1) << "Trace Op: " << type;
+  if (FLAGS_use_mkldnn) {
+    attrs["use_mkldnn"] = true;
+  }
```
zhiqiu, line 54: LGTM for dygraph mode.
I don't see handling `use_mkldnn` in static mode, is it already done in other PR?

wojtuss (approved):
LGTM

luotao1:
@zhiqiu Please have a double review!

python/paddle/fluid/tests/unittests/test_imperative_basic.py
```diff
@@ -640,6 +641,31 @@ def test_append_activation_in_dygraph3(self):
...
+        helper = LayerHelper(
+            fluid.unique_name.generate("test"), act="relu", use_mkldnn=True)
+        func = helper.append_activation
+        with fluid.dygraph.guard():
+            a = fluid.dygraph.to_variable(a_np)
+            res1 = func(a)
+            res2 = fluid.layers.relu(a)
+            self.assertTrue(np.array_equal(res1.numpy(), res2.numpy()))
+
+    def test_append_activation_in_dygraph_global_use_mkldnn(self):
```
zhiqiu, line 655: LGTM for the tests to check the accuracy of the result when `use_mkldnn` is True and False.

I wonder if there are another ways that we checked the `mkldnn` is really used. For example, we can add some VLOG in the mkldnn kernel, and see if the log is printed when `fluid.set_flags({'FLAGS_use_mkldnn': True})` is called. And we can do it manually it is difficult for unittest.
Maybe you have already done that, I just wonder that.

python/paddle/fluid/tests/unittests/test_imperative_basic.py
wojtuss: I can't see any option to test global flags in a different way, also other flags are tested this way. After the last commit the time spent with the flag set is minimum.

python/paddle/fluid/layer_helper.py
arlesniak: I've changed tracer on the c++ side, so there is no need for the above changes anymore.

paddle/fluid/imperative/tracer.cc
arlesniak: If you mean running through executor, it's already done in executor.cc:185

python/paddle/fluid/tests/unittests/test_imperative_basic.py
arlesniak: The problem we had is that AFAIK there is no better way to check if mkldnn kernel has been run than to check the output of DNNL_VERBOSE for now. The other idea is to run kernel throwing Exception with extended information about type of kernel. The other problem is that AFAIK python unittest lib does not have parsing c++ stdout result out of the box. Do you have suggestion what is the best way to check what type of kernel has been run from the python level ?

luotao1: > Do you have suggestion what is the best way to check what type of kernel has been run from the python level

I think checking the output of DNNL_VERBOSE is Ok. You can do it in the next PR.

PR #26499, opened 2020-08-20
Add output scale and trt op teller support for hard_swish and hard_sigmoid

cryoco:
### PR types

Function optimization
### PR changes

Others
### Describe

Add output scale and trt op teller support for hard_swish and hard_sigmoid


NHZlX (approved):
LGTM

PR #26504, opened 2020-08-20
Revert "Update cub to 1.9.8"

chenwhql:
### PR types

Bug fixes
### PR changes

Others
### Describe

Reverts PaddlePaddle/Paddle#24895

- revert reason:

After `cub` was upgraded to version 1.9.8, a single training process will occupy all currently visible GPUs, which has two effects:

1. training start speed is significantly slower
2. make the experience of using multi-process multi-cards training worse

There may be bugs in this version of cub, so revert this commit


zhangting2020 (approved):
LGTM

PR #26516, opened 2020-08-21
fix slow var initialize of LookaheadOptimizer, test=develop

mapingshuo:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix slow var initialize


zhiqiu (approved):
LGTM for the API of static graph

PR #26032, opened 2020-08-07
Add SyncBatchNorm

ceci3:
### PR types

Others

### PR changes

APIs

### Describe

Add SyncBatchNorm API for dygraph
1. because moving mean and moving variance computed all the time in the batch norm C++ kernel , so 'track_running_stats' is not kick in the API now. need to change batch norm kernel.
2. in the train mode, use mini-batch stats, in the eval mode, use global stats, not use `use_global_stas` and `trainable_stats` to jugde whether to use global or mini-batch stats

PR #26523, opened 2020-08-21
fix syncbn, test=develop

ceci3:
### PR types

Others

### PR changes

APIs

### Describe

use_global_stats  only  support use_global_stats = false, fix it
move syncbn from fluid/dygraph to paddle/nn/norm
pre PR: https://github.com/PaddlePaddle/Paddle/pull/26032


python/paddle/nn/layer/norm.py
```diff
@@ -14,15 +14,239 @@
...
+        self._weight_attr = weight_attr
+        self._bias_attr = bias_attr
+        self._num_features = num_features
+        self._data_layout = data_format
+        self._momentum = momentum
+        self._epsilon = epsilon
+        self._track_running_stats = track_running_stats
+
+        if self._track_running_stats == False:
+            logging.warn(
```
willthefrog: use `warnings` instead of `logging`

willthefrog (changes requested):

python/paddle/nn/layer/norm.py
ceci3: done, thanks

XiaoguangHu01 (approved):
LGTM

jzhang533 (approved):
lgtm

willthefrog (approved):
LGTM

PR #26347, opened 2020-08-17
Move and rename uniform_random API

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

- Implement `uniform_random` API under `paddle/tensor/random.py`.
- Rename `uniform_random` API in `paddle/tensor/random.py` as `uniform`.
- Create alias name of `paddle.tensor.random.uniform`：`paddle.tensor.uniform`, `paddle.uniform`

PR #26531, opened 2020-08-21
Fix uniform op Doc problem

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

`uniform` op has been implemented in PR #26347. There are some problems in Doc. This PR is used to fix the `uniform` op doc.


jzhang533 (approved):
lgtm

PR #26548, opened 2020-08-21
refine deprecated decorator

zhiqiu:
### PR types

Function optimization
### PR changes

APIs
### Describe

This PR enhances `deprecated` decorator:

1. support decorating class

-before
-after

2. change filter from `default` to `once`

- before
- after

**Plus, this PR temporally disables the warnings for paddle 2.0-beta, and it should be re-enabled in the future.**


jzhang533 (approved):
lgtm

PR #26565, opened 2020-08-22
Fix the sample code of SELU/selu

hong19860320:
### PR types

Function optimization

### PR changes

APIs

### Describe

Change the data type of sample code of SELU/selu


zhupengyang (approved):
LGTM

PR #26583, opened 2020-08-24
Support fp64 in `huber_loss_op` cuda kernel

yghstill:
### PR types

Bug fixes

### PR changes

OPs

### Describe

`huber_loss_op.cu` not support double type, so should support fp64 in `huber_loss_op` cuda kernel.


jerrywgz (approved):
LGTM

PR #26614, opened 2020-08-24
reduce_mean grad error if keepdim=True and reduce_all=True

zhupengyang:
### PR types

Bug fixes

### PR changes

OPs

### Describe

- eigen Tensor(y, dy) should be flattened to 1D if reduce_all=True
- If keepdim=True and reduce_all=True, reduce_mean grad compute is crashed because y&dy is N-D but not flattened to 1-D.


hong19860320 (approved):
LGTM

qjing666 (approved):
LGTM

PR #26647, opened 2020-08-25
Fix atomicAdd in grid sample op and affine grid op

wanghaoshuang:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix atomicAdd in grid sample op and affine grid op


baiyfbupt (approved):
LGTM

PR #26664, opened 2020-08-25
[DO NOT MERGE][ONEDNN] Added prototype of midout for pool

jczaja:
This PR is to be discussed. oneDNN kernel of grad pool2d needs "MidOut" data , similar way like it is done for LRN op.
So far instead of MidOut , oneDNN cache was used. But this is very complex especially when multi-threaded training is running. It will also reduce cache operations for dygraph. 

Notes:
- MidOut is added regardless the execution (oneDNN, GPU, native CPU)
- MidOut data depends on oneDNN , so format is opaque, so we do not validate this data in UT.

@luotao1 Please take a look if this idea is acceptable.

### PR types
Breaking changes

### PR changes
Ops

### Describe
oneDNN pool2d kernel of grad op is using data from FWD. This PR is adding MidOut tensor to store that data and make it available during grad execution.

### DIFF
--- a/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
@@ -56,7 +56,10 @@ class PoolMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     if ((ctx.Attr<bool>("is_test") == false) &&
         (ctx.Attr<std::string>("pooling_type") == "max")) {
       // Training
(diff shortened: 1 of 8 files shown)


lidanqing-vv:
@donproc Can you give some suggestion to this PR?

lidanqing-vv:
@jczaja Dongdong suggest to try dispensable, https://github.com/PaddlePaddle/Paddle/wiki/OP-Input-Output-Attribute-Compatibility-Modification

PR #26682, opened 2020-08-26
[Dy2stat] Fix jit.to_static usage and Add unittests

Aurelius84:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix `paddle.jit.to_static`(alias name of `declarative`) interface unavailable bug.

In Paddle2.0，we recommend users to use `paddle.jit.to_static` but it will raise error that was not up to expectations. Because `DECORATOR_NAMES` list doesn't include `to_static`.

In addition, unittests related to functions and models are added but not all, we keep some unittsts for testing `declarative`.


liym27 (approved):
LGTM

PR #26683, opened 2020-08-26
Fix clip input check

willthefrog:
### PR types
Bug fixes

### PR changes
OPs 

### Describe

- support `max == min`
- fp16 is not supported
- fix  max/min default value  overflow error 
- allow both min and max to be None


jzhang533 (approved):
lgtm

PR #26006, opened 2020-08-06
add set default dtype, get default dtype

ForFishes:
### PR types

New features 
### PR changes

APIs
### Describe

add "set_default_dtype" ,"get_default_dtype"

PR #26695, opened 2020-08-26
support default dtype for distribution APIs

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

In Paddle 2.0-beta, PR #26006 add a new `default_dtype` API. This PR is used to support global default dtype for distribution related APIs.

#### Distribution related APIs include:

```
- op：
    - paddle.tensor.random.bernoulli
    - paddle.tensor.random.gaussian_random
    - paddle.tensor.random.standard_normal
    - paddle.tensor.random.normal
    - paddle.tensor.random.uniform
    - paddle.tensor.random.randn
    - paddle.tensor.random.rand
    - paddle.tensor.random.randint
    - paddle.tensor.random.randperm

- sample method of distribution related classes
    - paddle.distribution.Normal
    - paddle.distribution.Uniform
```

#### They are divided into the following categories:

- have `dtype` parameter and the default value is `float32`, including:
```
- paddle.tensor.random.gaussian_random
- paddle.tensor.random.standard_normal
- paddle.tensor.random.uniform
- paddle.tensor.random.randn
- paddle.tensor.random.rand
```
method: 
Set `None` to `dtype` parameter. if `dtype` is `None`, get the value of `dtype` from `global dtype`.
Because `global dtype` support [`float16`, `float32`, `float64`], but these distribution related APIs only support [`float32`, `float64`]. So if `global dtype` is `float16`, remind users that the dtype has error.

- have `dtype` parameter and the default value is `int`, including:
```
- paddle.tensor.random.randint
- paddle.tensor.random.randperm
```
Can't use `global dtype` for `global dtype` not supporting `int`.

- don't have `dtype` parameter and have input tensor, including:
```
- paddle.tensor.random.bernoulli
```
Be consistent with Pytorch. get the output value of `dtype` from dtype of input tensor.
Don't use `global dtype`.

- Call other distritbuion APIs, including:
```
- paddle.tensor.random.normal
- sample method in paddle.distribution.Normal
- sample method in paddle.distribution.Uniform
```
Let the dtype of `uniform` be consistent with the dtype of `low` parameter and `high` parameter.
Let the dtype of `normal` be consistent with the dtype of `mean` parameter and `std` parameter.
Don't use `default dtype`.


python/paddle/tensor/random.py
```diff
@@ -111,7 +111,7 @@ def gaussian_random(shape, mean=0.0, std=1.0, dtype='float32', name=None):
         seed(int, optional): ${seed_comment}
         dtype(str|np.dtype|core.VarDesc.VarType, optional): The data type of
             the output Tensor. Supported data types: float32, float64.
-            Default is float32.
+            Default is None, use ``get_default_dtype``.
```
jzhang533: 不需要支持`core.VarDesc.VarType`
use ``get_default_dtype`  -> use global default (see `get_default_dtype` for details)

下同

jzhang533 (commented):
- for APIs planned to deprecate ( in python/paddle/fluid/layers/nn.py),  use `deprecated` to decorate is enough.

XiaoguangHu01 (dismissed):
LGTM

python/paddle/tensor/random.py
pangyoki: Done

jzhang533 (approved):
lgtm

XiaoguangHu01 (approved):
LGTM

python/paddle/fluid/tests/unittests/test_uniform_random_op.py
```diff
@@ -536,5 +536,31 @@ def test_check_output(self):
...
+            paddle.framework.set_default_dtype('float32')
+            out = paddle.tensor.random.uniform([2, 3])
+            self.assertEqual(out.dtype, fluid.core.VarDesc.VarType.FP32)
+
+        def test_default_fp_64():
+            paddle.framework.set_default_dtype('float64')
+            out = paddle.tensor.random.uniform([2, 3])
+            self.assertEqual(out.dtype, fluid.core.VarDesc.VarType.FP64)
+
+        test_default_fp_64()
```
zhiqiu, line 559: Better `fp64` or `float64`, not  `fp_64`. You can refine it in the next PR.

python/paddle/tensor/random.py
```diff
@@ -401,6 +415,13 @@ def uniform(shape, dtype='float32', min=-1.0, max=1.0, seed=0, name=None):
 
 
     """
+    if dtype is None:
+        dtype = paddle.framework.get_default_dtype()
+        if dtype not in ['float32', 'float64']:
+            raise TypeError(
+                "uniform only supports [float32, float64], but the default dtype is %s"
+                % dtype)
+
```
zhiqiu, line 424: Not important, but I want to suggest you to abstract and reuse the same code, try to reduce duplicated code (maybe you already know).

zhiqiu (approved):

python/paddle/fluid/tests/unittests/test_uniform_random_op.py
pangyoki: Done, refer to PR #26951

python/paddle/tensor/random.py
pangyoki: Done, refer to PR #26951

PR #26713, opened 2020-08-26
fix sample method of Uniform and Normal class

pangyoki:
### PR types

Bug fixes

### PR changes

APIs

### Describe

Fix the `sample` method of  `Uniform` and `Normal` class .

For `Uniform` class：
- Reason
The shape of inputs of `elementwise_add` op in `sample` method is not matching. In detail, in the following code:
```output = uniform_random_tmp * (zero_tmp + self.high - self.low) + self.low```
the shape of zero_tmp is [batch_shape, sample_shape]，
the shape of self.high is [batch_shape].
When zero_tmp and self.high doing `elementwise_add` operation, it will cause "Broadcast dimension dismatch". 

- Method
To solve this problem, before doing `elementwise_add` operation, `reshape` zero_tmp. That is, change the shape of zero_tmp from [batch_shape, sample_shape] to [sample_shape, batch_shape].
At the same time, reshape uniform_random_tmp to [sample_shape, batch_shape].

For `Normal` class, it has the same problem, using the `reshape` method too.

### CODE REVIEW

**Review by zhiqiu — COMMENTED** [review]
Please add some unittests.

**Comment by pangyoki** [comment]
> Please add some unittests.

Done

**Comment by pangyoki** [comment]
> Please add some unittests.

Done

File: python/paddle/fluid/tests/unittests/test_distribution.py [hunk 1]
```diff
@@ -516,6 +594,31 @@ def compare_uniform_with_numpy(self,
...
+            output_sample_float_np_broadcast_diff.shape,
+            gt_sample_float_np_broadcast_diff.shape,
+            rtol=tolerance,
+            atol=tolerance)
+        np.testing.assert_allclose(
+            output_sample_np_diff.shape,
+            gt_sample_np_diff.shape,
+            rtol=tolerance,
+            atol=tolerance)
+        np.testing.assert_allclose(
```
zhiqiu: Better check `equal` for integer, and you can use the assertion methods in `unittest`, like `self.assertEqual(a, b)`

pangyoki: Done

zhiqiu (approved):
LGTM

PR #26729, opened 2020-08-27
[2.0 API] fix the call to core.ops.x

sandyhouse:
### PR types

Others
### PR changes

APIs
### Describe

call core.ops.xx before any check as it is unnecessary to check parameter validation in dygraph mode (it just report errors during run time.)


wangxicoding (approved):
LGTM

PR #26732, opened 2020-08-27
fix remainder, floor_div

ForFishes:
### PR types

Bug fixes
### PR changes

APIs
### Describe

1、`remainder` should have same sign as divide.
2、`floor_divide` is not correct when divided by zero.


XiaoguangHu01 (approved):
LGTM

PR #25558, opened 2020-07-16
Add iterable dataset support for multiprocess DataLoader

heavengate:
### PR types
New features

### PR changes
APIs

### Describe

**add IterableDataset support for multiprocess DataLoader**

- add `paddle.io.IterableDataset` base class
- add `paddle.io.get_worker_info` to get worker process information for data splitting in IterableDataset

屏幕快照 2020-07-30 下午9 05 38

屏幕快照 2020-07-30 下午9 07 05

屏幕快照 2020-07-30 下午9 07 20

屏幕快照 2020-07-30 下午9 07 30

PR #26739, opened 2020-08-27
fix dataloader performace decrease & unittest error

heavengate:
### PR types
Bug fixes

### PR changes
APIs

### Describe
fix DataLoader performance decrease in some reader intensive model by  #25558.

The `_send_idx` and `_rcvd_idx` checking should be only in the `IterableDataset` case, Map-Style dataset may enter this while loop in the last few samples(when some of subprocess exit and others not) and cost extra time.

In CI unittest, the time cost in while loop may lead main process check subprocess status error and exit

| Model | original Speed | #25558 Speed | This PR Speed |
| :----- | :------: | :------: | :------: |
| MobileNetV1 | 1047.742 img/s | 913.665 img/s | 1030.650 img/s |
| MobileNetV2 | 816.83 img/s | 574.08 img/s | 808.304 img/s |


qingqing01 (approved):
LGTM

PR #26749, opened 2020-08-27
add parameters check in static mode for diag op

LutaoChu:
### PR types

Bug fixes
### PR changes

APIs
### Describe

add 'offset' type check, add 'padding_value' type check, add input 'x' shape check


wawltor (approved):
LGTM

PR #26774, opened 2020-08-28
refine paddle inference api

jiweibo:
### PR types

New features
### PR changes

APIs
### Describe

**Another copy of pr https://github.com/PaddlePaddle/Paddle/pull/26098**

Due to historical reasons, there are many kinds of API in paddle inference at present, which is a great challenge for users to use. Therefore, we re optimized the API. The details are as follows:

1) We support **forward compatibility**, that is, all previous APIs can be used, but will be **officially removed** in the **next two versions**.

2) In this version，all interfaces in paddle_infer namespace.

3) Use `paddle_infer::Config` to do inference configuration.

4) Use `paddle_infer:: Tensor` to represent input and output

5) Use `paddle_infer::Predictor` to represent the inference engine.

6) Use `paddle_infer::CreatePredictor(Config&config)` to create inference engine.


Superjomn (approved):
LGTM

luotao1 (approved):
LGTM

PR #26776, opened 2020-08-28
fix dimensions error for mobilenetv1_KL_quant

sfraczek:
### PR types
Bug fixes

### PR changes
Others

### Describe
Fixed Error when running `test_quant2_int8_mobilenetv1_mkldnn` with the recieved `mobilenetv1_KL_quant model`.
The error that was fixed was `AssertionError: The size of weight scales vector (1000) does not match the number of output channels (1024) in the weights tensor fc7_weights.`
The reason was that the weights in this mul operator did not require transposition as in previous cases.


wojtuss (commented):
I think those new cases have to be covered in the test.

wojtuss (approved):
LGTM

PR #26512, opened 2020-08-20
Fix the cmake-function named inference_download_and_uncompress on Windows

Avin0323:
### PR types
Others

### PR changes
Others

### Describe
Cmake-function named ```inference_download_and_uncompress``` doesn't support download file on Windows. Unittests(```test_analyzer```, ```test_api_impl```, ```test_analysis_predictor```) that depends file word2vec.inference.model.tar.gz couldn't run on Windows.

Fix this by adding ```if```branching. And remove the limit of unittests on Windows.

PR #26839, opened 2020-08-31
Modify data download function and support unittests of inference APIs on windows

Avin0323:
### PR types
Others

### PR changes
Others

### Describe
Since the data download cmake-function did not support Windows system previously, unittests of inference APIs are disabled on Windows. This problem has been fixed in [PR26512](https://github.com/PaddlePaddle/Paddle/pull/26512) which will set `WITH_INFERENCE_API_TEST=ON` to add unittests of inference APIs on Windows.

And fix five unittest problems:

1. In file `paddle/fluid/inference/tests/api/analyzer_image_classification_tester.cc`: To create directory should use `_mkdir` instead of `mkdir` on windows.
2. In file `paddle/fluid/inference/tests/api/analyzer_ner_tester.cc`:  `std::min` should specify an explicit type, otherwise the compiler can't get correct type.
3. In file `paddle/fluid/inference/tests/api/CMakeLists.txt`: `MobileNet_qat_perf.tar.gz` does not download before run `test_analyzer_quant_performance_benchmark` on Windows, but it download successful on Linux. The reason is that data only download on Linux according to file `python/paddle/fluid/contrib/slim/tests/CMakeLists.txt`.
image
4. In file `paddle/fluid/inference/tests/api/full_pascalvoc_test_preprocess.py` and `paddle/fluid/inference/tests/api/full_ILSVRC2012_val_preprocess.py`: The import location of the `StringIO` is inconsistent between py2 and py3.
image
5.In file `paddle/fluid/inference/tests/api/analyzer_capi_tester.cc`: Float array `input` points to an invalid address, and it can run successful via `new/delete`. The real reason needs more debugging.
image


zhwesky2010 (approved):
LGTM

PR #26884, opened 2020-09-01
Fix some cast error when inserting cast op.

wzzju:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix some cast error when inserting cast op.
* Not insert cast when src_dtype is equal to dest_dtype.
* Set the stop_gradient attr value of the out as the input's value when inserting `cast` op.


Xreki (approved):
LGTM

PR #26924, opened 2020-09-02
Fix clip op attr

willthefrog:
### PR types
Bug fixes

### PR changes
OPs

### Describe
turns out attribute only support float32,  PR #26683 introduced a but that causes error when input is fp64  
this PR revert previous changes to clip op


iclementine (approved):
LGTM

PR #26932, opened 2020-09-02
fix uniform CI random fail problem

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

`uniform` op may occur CI random fail problem. To test correctness of `uniform` op, we will sample from uniform distribution and collect statistic data. If the sample number is small, it may occur occasionality, and make the statistic data different with the ideal situation.
To solve this problem, increase the sample number to avoid occasionality.


zhiqiu (approved):
LGTM

PR #26934, opened 2020-09-02
Fix capture exit code failed caused by `set -e`

Avin0323:
### PR types
Others

### PR changes
Others

### Describe
`set -e` causes capture exit code failed, then add `set +e` in `build_and_check`. And change the exit code of `check_style` to 4.


luotao1 (approved):
LGTM

PR #26940, opened 2020-09-02
[cherry-pick2.0]set_default_type only take effect on python floats or complex

zhwesky2010:
### PR types

Bug fixes
### PR changes

APIs
### Describe

``paddle.set_default_type`` should only take effect on python floats or complex.


jzhang533 (approved):
lgtm

PR #26901, opened 2020-09-02
erase Raises and refine doce of random functions

zhupengyang:
### PR types

Others

### PR changes

Docs

### Describe

erase Raises and refine doce of random functions
image
image
image
image
image

PR #26951, opened 2020-09-02
change uniform_random to uniform and optimize function names in unittest for random ops

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

PR #26695 has supported global default dtype for distribution APIs, there are some format problems to be processed. PR #26901 has fixed problems for `gaussian`, `standard_normal`, `randn` and `normal`.
This PR is used to
- change `fp_16`, `fp_32`, `fp_64` to `fp16`, `fp32`, `fp64`.
- change name of `uniform_random` to `uniform`.

### DIFF
--- a/python/paddle/tensor/random.py
+++ b/python/paddle/tensor/random.py
@@ -36,19 +36,6 @@ __all__ = [
 ]
 
 
-def set_global_dtype_for_random_apis(api_name):
-    """
-    This is a common function used in APIs that can use global default dtype.
-    Including: ``gaussian_random``, ``standard_normal``, ``randn``, ``uniform``, ``rand``.
-    """
-    dtype = paddle.framework.get_default_dtype()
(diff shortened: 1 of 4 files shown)


python/paddle/tensor/random.py
```diff
@@ -41,6 +41,19 @@
 ]
 
 
+def set_global_dtype_for_random_apis(api_name):
```
zhiqiu: Maybe the function could be named as `_infer_dtype_from_global_dtype`.

pangyoki: This function has been implemented in PR #26901

pangyoki: the dumplicate code has been deleted in PR #26901

zhiqiu (approved):
LGTM

PR #26987, opened 2020-09-03
[Dy2stat] Add descriptor cache for StaticLayer

Aurelius84:
### PR types

Bug fixes

### PR changes

Others

### Describe

Add descriptor cache for StaticLayer.

### Why need this
In StaticLayer, we implement `__get__` method to make it into `descriptor`(see [this](https://docs.python.org/3/howto/descriptor.html)).

This allows us to easily parse the instance from decorated bound method, such as `self.forward`.
```python
class A:
    @decorator
    def foo(self, x):
        return x

a = A()
a.foo(10)
```

if `decorated` is a decorator,  `a.foo` will firstly call `__get__` in `decorator`. The common implementation is:
```python
def __get__(self, instance, owner):
   # instance is `a` as above
   # owner is `A` as above
    do_somethind_useful()
```

**However why we need `self._descriptor_cache` here?**

Before this PR, consider the following example code:
```python
class Net(Layer):
    ....
    @to_static
    def forward(self, x):
         return x*2

net_1 = Net()
net_2 = Net()

isinstance(net_1.forward, StaticLayer)  # True
isinstance(net_2.forward, StaticLayer)  # True

id(net_1.forward) == id(net_2.forward)  # True !!
id(net_1.forward) == id(Net.forward)  # True !!
```
That's the problem. For different instance from same Class, they share the same StaticLayer instance, which has risk in some  operations.

### Solution
We introduce a `wreakref.WeakKeyDictionary()` as descriptor_cache to hold different StaticLayer instance.

```python
def __get__(self, instance, owner):
    if instance is not in self._descriptor_cache:
        self._descriptor_cache[instance] = self._clone()  # create a new staticLayer

    return self._descriptor_cache[instance]
```

The benefits to choose `wreakref.WeakKeyDictionary()` instead of `Dict` is if `net` is garbage collected, it will be removed from `self._descriptor_cache` to save memory.

After this PR: 
```python
id(net_1.forward) == id(net_2.forward)  # False
id(net_1.forward) == id(Net.forward)  # False
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/convert_call_func.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/convert_call_func.py
@@ -31,6 +31,7 @@ from paddle.fluid.dygraph.dygraph_to_static.convert_operators import convert_len
 from paddle.fluid.dygraph.dygraph_to_static.logging_utils import TranslatorLogger
 from paddle.fluid.dygraph.dygraph_to_static.program_translator import StaticLayer
 from paddle.fluid.dygraph.dygraph_to_static.program_translator import convert_to_static
+from paddle.fluid.dygraph.dygraph_to_static.program_translator import unwrap_decorators
 from paddle.fluid.dygraph.layers import Layer
 
 # TODO(liym27): A better way to do this.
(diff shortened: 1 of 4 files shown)


chenwhql (approved):
LGTM

liym27 (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #26995, opened 2020-09-03
fix the remainder

ForFishes:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix the remainder when the x'shape < y's shape.


jzhang533 (approved):
lgtm

PR #26999, opened 2020-09-03
Add bfloat16 passes

wozna:
### PR types

New features
### PR changes

Others 
### Describe

This PR is the next step of adding bfloat16 support to PaddlePaddle. 

This PR presents: 
- two passes for enabling mkldnn bfloat16  
- tests for those passes
- insertion passes to `CpuPassStrategy`
- there is still no support in Numpy for `bfloat16`, so added there  in `framework.py`  conversion from python `uint16`  to C++ `bfloat16`

More about passes:
After enabling bfloat16, two passes will be applied to the graph. 
- first **cpu_bfloat16_placement_pass.cc** - this pass first sets to attribute `mkldnn_data_type = bfloat16` to all operators contained in the vector `bfloat16_enabled_op_types` or if user won't set any op types all operators that support bfloat16 will be marked.
- second **cpu_bfloat16_pass.cc** - this pass first adds reorder between `float32` and `bfloat16` operators, unless `bfloat16` operator is `conv2d` (conv2d can reoder input to bfloat16 inside kernel, so reoder is no needed, what decreases number of reoders). 
Than adds reorder between `bfloat16` and `float32`  operators or use attr `force_output_fp32` if the operator has this attribute.


paddle/fluid/framework/ir/mkldnn/cpu_bfloat16_pass_tester.cc
```diff
@@ -0,0 +1,145 @@
...
+ProgramDesc BuildProgramDesc(bool use_mkldnn) {
+  ProgramDesc prog;
+  for (auto& v : variable_names) {
+    prog.MutableBlock(0)->Var(v);
+  }
+  SetOp(&prog, "dropout", "Dropout1", {"z"}, {"a"}, use_mkldnn, "float32");
+  SetOp(&prog, "conv2d", "Conv1", {"a"}, {"b"}, use_mkldnn, "bfloat16");
+  SetOp(&prog, "pool2d", "Pool1", {"b"}, {"c"}, use_mkldnn, "bfloat16");
+  SetOp(&prog, "conv2d", "Conv1", {"c"}, {"d"}, use_mkldnn, "bfloat16");
+  SetOp(&prog, "dropout", "Dropout2", {"d"}, {"e"}, use_mkldnn, "float32");
```
jczaja, line 86: perhaps in next PRs it would be good to add some Op here with int8 data type to test if bf16 is not overwritting int8 ops.

jczaja (approved):
LGTM

jczaja:

luotao1 (approved):

paddle/fluid/framework/ir/mkldnn/cpu_bfloat16_pass_tester.cc
wozna: You're right, I'll add it.

PR #27008, opened 2020-09-04
cherry-pick from the develop PR#26792, fix the argmin, argmax

wawltor:
### PR types
 Others

### PR changes
OPs

### Describe
fix the argmin,argmax op for the paddlepaddle 2.0， cherry-pick from [PR 26792](https://github.com/PaddlePaddle/Paddle/pull/26792)
- Registered the ``AddCheckPoint``  for the argmax, argmin,  add the ``flatten`` attribuate for the argmax/argmin, the flatten means that flattening the tensor as input;At the same time, change the default value from -1 to 3 in the attribute of argmax/argmin, but do not change the operator logic.

- Fixed the output of argmax/argmin, because the paddlepaddle do not support the ``scalar`` tensor,  change the output of argmin/argmax from the ``scalar``  tensor to normal tensor with shape ``[1]`` .

- Added the check for dtype of argmin/argmax, when the  num of element in argmax/argmin is larger the max value of ``int32``,  remind the user to use the dtype of ``int64``.

### DIFF
--- a/paddle/fluid/operators/arg_max_op.cc
+++ b/paddle/fluid/operators/arg_max_op.cc
@@ -12,6 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */
 
+#include "paddle/fluid/framework/op_version_registry.h"
 #include "paddle/fluid/operators/arg_min_max_op_base.h"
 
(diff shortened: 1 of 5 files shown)


XiaoguangHu01 (approved):
LGTM

Superjomn (approved):
LGTM

TCChenlong (approved):
LGTM

PR #27011, opened 2020-09-04
[Dy2stat]Reduce repeat time of Predictor in unittest

Aurelius84:
### PR types

Others
### PR changes

Others
### Describe

Reduce repeat time of Predictor in unittest.

The `repeat_time` means it will run predictor for many time to ensure no diff for same input tensor. But this will cost a lot of time leading Timeout in some model unittest. So in this PR, I reduce it from 10 into 2.


OliverLPH (approved):
LGTM

liym27 (approved):

PR #27023, opened 2020-09-04
Fix conv and summary api bug

LielinJiang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix conv and summary api bug


DDDivano (approved):
LGTM

PR #27027, opened 2020-09-04
fix cuda generator hard-coded offset step

yaoxuefeng6:
### PR types

Function optimization
### PR changes

Others
### Describe

fix cuda generator hard-coded offset step


MrChengmo (approved):
LGTM

PR #27028, opened 2020-09-04
fix normal ut by set manual_seed

zhupengyang:
### PR types

Bug fixes

### PR changes

OPs

### Describe

fix normal ut by set manual_seed


hong19860320 (approved):
LGTM

PR #27034, opened 2020-09-04
optimize the error message

kinghuin:
### PR types
Function optimization

### PR changes
OPs

### Describe
Optimize the error messages for unpooling.cc, unpooling.cu


wawltor (approved):
LGTM

PR #27040, opened 2020-09-04
[Dy2stat] support usage: to_static(model)

Aurelius84:
### PR types

New features
### PR changes

APIs
### Describe

support usage: to_static(model)

### Code Example:
Before this PR，if we want decorate a `forward()` from model, we shall write code like:
```python
model.forward = to_static(model.forward, input_spec=...)
``` 
It's not easy-to-use. In this PR，we support decorate model directly and convert call for `foward` and all inner function.

```Python
class SimpleNet(Layer):
    def __init__(self, ...):
        .......

    def forward(self, x):
        ......
        return out

net = SimpleNet()
x = paddle.to_tensor(np.ones(4, 10))

# case 1:  decorate model and feed data
net = to_static(net)
out = net(x)
print(out)

# case 2: decorate model with InputSpec
net = to_static(net, input_spec=[InputSpec([None, 8, 10])])
self.assertListEqual(list(input_shape), [-1, 8, 10])


# case 3: redecorate with new InputSpec
net = declarative(net, input_spec=[InputSpec([None, 16, 10])])
input_shape = net.forward.inputs[0].shape
self.assertListEqual(list(input_shape), [-1, 16, 10])
```

### DIFF
--- a/python/paddle/fluid/dygraph/jit.py
+++ b/python/paddle/fluid/dygraph/jit.py
@@ -211,6 +211,11 @@ def declarative(function=None, input_spec=None):
     # for usage: `declarative(foo, ...)`
     if function is not None:
         if isinstance(function, Layer):
+            if isinstance(function.forward, StaticLayer):
+                class_name = function.__class__.__name__
+                warnings.warn(
+                    "`{}.forward` has already been decorated somewhere. It will be redecorated to replace previous one.".
+                    format(class_name))
             function.forward = decorated(function.forward)
             return function
         else:
--- a/python/paddle/fluid/dygraph/jit.py
+++ b/python/paddle/fluid/dygraph/jit.py
@@ -210,7 +210,11 @@ def declarative(function=None, input_spec=None):
 
     # for usage: `declarative(foo, ...)`
     if function is not None:
-        return decorated(function)
+        if isinstance(function, Layer):
+            function.forward = decorated(function.forward)
+            return function
+        else:
+            return decorated(function)
 
     # for usage: `@declarative`
     return decorated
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/jit.py
```diff
@@ -210,7 +210,16 @@ def decorated(python_func):
 
     # for usage: `declarative(foo, ...)`
     if function is not None:
-        return decorated(function)
+        if isinstance(function, Layer):
+            if isinstance(function.forward, StaticLayer):
```
zhhsplendid, line 214: Please check whether ``function`` has ``forward``, if not, I think we should throw our exception to indicate users?

zhhsplendid: Sorry, I was wrong, ``function`` is an instance of ``Layer`` so it must have the ``forward``. I will approve this PR

zhhsplendid (approved):
LGTM

python/paddle/fluid/dygraph/jit.py
Aurelius84: It inherit `Layer` and hold `forward` function.

python/paddle/fluid/dygraph/jit.py
```diff
@@ -210,7 +210,16 @@ def decorated(python_func):
...
+                class_name = function.__class__.__name__
+                warnings.warn(
+                    "`{}.forward` has already been decorated somewhere. It will be redecorated to replace previous one.".
+                    format(class_name))
```
liym27, line 218: > 这次PR中不重要

动转静中的warning，统一了接口，后续建议使用 `logging_utils._TRANSLATOR_LOGGER.warn()`

我这边最近会统一做更改

liym27 (approved):
LGTM

python/paddle/fluid/dygraph/jit.py
Aurelius84: 赞

PR #27046, opened 2020-09-04
fix _check_values_dtype_in_probs method in Distribution class

pangyoki:
### PR types

Others

### PR changes

APIs

### Describe

`_check_values_dtypes_in_probs(param, value)` method in `Distribution` class is used to ensure dtype of `param` and `value` are consistent. (refer to PR #26767)
If in_dygraph_mode and value.dtype == param.dtype, it doesn't have return value in dygraph mode. And it has to run static mode again. This PR adds return value in dygraph mode.

### DIFF
--- a/python/paddle/distribution.py
+++ b/python/paddle/distribution.py
@@ -138,7 +138,7 @@ class Distribution(object):
         convert value's dtype to be consistent with param's dtype.
 
         Args:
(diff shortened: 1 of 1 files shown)


zhiqiu (approved):
LGTM

PR #27108, opened 2020-09-07
[2.0API] fix weight_norm support negative dim and unittest in convert_syncbn

ceci3:
### PR types

Bug fixes

### PR changes

APIs

### Describe

fix weight_norm support negative dim and unittest in convert_syncbn


baiyfbupt (approved):
LGTM

PR #27112, opened 2020-09-07
[Feature] Enhance inplace addto strategy for gradient accumulation in static graph

zhiqiu:
### PR types

New features
### PR changes

Others
### Describe

New feature, support `inplace addto strategy` for gradient accumulation, which can improve the performance of gradient accumulation.

### Background
In back-propagation，if the gradients of a tensor are generated by more than one operation, these gradients should be accumulated before back-propagating to the next layer. For example,

If the forward network contains the following part,
```
y = conv2d(x, w)
z = add(x, y)

```
Then in backward, 
```
x_grad_0 = add_grad(z_grad)
x_grad_1 = conv2d_grad(y_grad, ...)
x_grad = sum(x_grad_0, x_grad_1)

```
Traditionally, if the gradients of a tensor are generated by `n` operation, then after these gradients generated, a `sum` operation is used to sum up these gradients. `x_grad = sum(x_grad_0, x_grag_1, ..., x_grad_n)`

However, we can improve the performance by adds up these gradients one by one in their backward operations. 

For example, the backward API of conv in cudnn [cudnnConvolutionBackwardData](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnConvolutionBackwardData)  contains a `beta` arguments,
image

which means, 
`dstValue = alpha[0]*result + beta[0]*priorDstValue`

So, in the above case, if the gradient `x_grad_1` is set to `x_grad_0` before `conv2d_grad`'s execution, then we can set `beta = 1` and the result `x_grad_1` will be exactly `x_grad_0 + x_grad_1`, which is the gradient accumulation we want to do.

### Implementation

- In the backward stage, when appending grad operators automatically, use several `grad_add ops` instead of `sum op`.
```
    g = sum(g_0, g_1,..., g_n)
 ==>
    g_sum_0 = g_0
    g_sum_1 = grad_add(g_sum_0, g_1)
    g_sum_2 = grad_add(g_sum_1, g_2)
     ...
    g_sum_n = grad_add(g_sum_n-1, g_n)
```

- For each `grad_add` op with the form `out = grad_add(left, right)`
    - Make the two input tensor and the output tensor share the same memory address. 
    - Make the right input of `grad_add`  do `addto` calculation.
    - Make the `grad_add` skip execution (since its work is done by `addto`)

### Usage
To train with `inplace addto strategy`, there are two steps.
- Set `FLAGS_max_inplace_grad_add` to a positive number, for example, 8. It means if the number gradients that need to sum up is less than 8, the `grad_add` will be used.
- Set `build_strategy.enable_addto=True`, it enables the `inplace addto strategy`.

### Performance
The performance of ResNet50 with amp enabled and batch_size=128 on V100 single card.

- before: 1014 img/s
- after: 1078 img/s, **6.3% imporved.**

### Accuracy

Compare the accuracy of training ResNet-50 with amp (inplace addto  enabled) and fp32 on 2 v100 cards.
Batch_size=256.

1. train loss

image

2. test loss
image

3. test acc1
image

4. test acc5
image


zhiqiu:
This PR fails on building in some CI checks, the error message is as follows:
image

It seems that the reason is elementwise_add_grad_grad does not have MKLDNN kernel. However, `elementwise_add_grad_grad` is not modified in this PR.

luotao1:
> It seems that the reason is elementwise_add_grad_grad does not have MKLDNN kernel. However, elementwise_add_grad_grad is not modified in this PR.

@jczaja @lidanqing-intel @grygielski Could you help see this problem?

zhiqiu:
> > It seems that the reason is elementwise_add_grad_grad does not have MKLDNN kernel. However, elementwise_add_grad_grad is not modified in this PR.
> 
> @jczaja @lidanqing-intel @grygielski Could you help see this problem?

Fixed in #27310, it is a bug of Paddle's cmake file.

phlrain (dismissed):

raindrops2sea:
> * Set `FLAGS_max_inplace_grad_add` to a positive number, for example, 8. It means if the number gradients that need to sum up is less than 8, the `grad_add` will be used.

What is a reasonable default value for max_inplace_grad_add? The reason and influence of the maximum number limit need to be further analyzed in the future.

raindrops2sea (dismissed):

lanxianghit (dismissed):

zhiqiu:
> > * Set `FLAGS_max_inplace_grad_add` to a positive number, for example, 8. It means if the number gradients that need to sum up is less than 8, the `grad_add` will be used.
> 
> What is a reasonable default value for max_inplace_grad_add? The reason and influence of the maximum number limit need to be further analyzed in the future.

OK, I will do more experiments on different models to evaluate it.

lanxianghit (approved):

raindrops2sea (approved):

Xreki (approved):
LGTM for the modification of operator.h and backward.py. PS: 一些注释掉的代码后续可以清理一下。

zhiqiu:
> LGTM for the modification of operator.h and backward.py. PS: 一些注释掉的代码后续可以清理一下。

好的

PR #27121, opened 2020-09-07
Fix ERNIE GPU inference with shared params in multihead matmul 

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

If weights or biases in qkv's fc are shared by multiple multihead_matmul patterns, we do not support this kind of fusion, multihead_matmul_fuse_pass_v2 will not take effect.


jiweibo (approved):
LGTM

PR #27175, opened 2020-09-08
fix judge cache file of inference api

zhwesky2010:
### PR types

Others
### PR changes

Others
### Describe

fix judge cache file of inference API. such as ``temp/transformer_model.tar.gz`` , because cannot be hit now because ''/'' between ``temp`` and ``transformer_model.tar.gz``. 

image


luotao1 (approved):
LGTM

PR #27185, opened 2020-09-08
Move unittest test_optimizer_in_control_flow from CI multi_cards.

liym27:
### PR types

Others
### PR changes

Others
### Describe

Move unittest test_optimizer_in_control_flow from CI multi_cards.

- Why do this? 
`test_optimizer_in_control_flow.py` needs to run in multi cards to **test `NotImplementedError`**. Here, move this test from RUN_TYPE=DIST in tests/unittests/CMakeList.txt, to use multi cards **only on CPU** not GPU to reduce CI time.


zhhsplendid (approved):
LGTM

PR #27201, opened 2020-09-09
Fix test_origin_info to be compatible with PY3.8, because ast module is different in PY3.8

liym27:
### PR types

Bug fixes
### PR changes

Others
#### Fix test_origin_info to be compatible with PY3.8, because ast module is different in PY3.8

---
There are differences in ast_node.lineno between PY3.8+ and PY3.8-.
 If the first gast.FunctionDef has decorator, the lineno of gast.FunctionDef is differs.
1. PY3.8-
 its lineno equals to the lineno of the first decorator node, which is not right.
2. PY3.8 and PY3.8+
its lineno is the actual lineno, which is right.


gfwm2013 (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #27237, opened 2020-09-10
remove auto mode from localsgd optimizer

ForFishes:
### PR types

Function optimization
### PR changes

APIs
### Describe

remove auto mode from localsgd optimizer


gavin1332 (approved):
LGTM

PR #27246, opened 2020-09-10
fix unused var with zero gradient bug in fluid.gradient

Aurelius84:
### PR types

Bug fixes
### PR changes

APIs 
### Describe

fix unused var with zero gradient

Example code:
`x1`, `x2`, `x3` are output vars from same op, but only `x1` is used to calculate gradient, so `x2@GRAD`、`x3@GRAD` should be zero but not before in this PR. They are not initialized as expected.

```python
class TestGradientWithPrune(unittest.TestCase):
    def test_prune(self):
        x = fluid.data(name='x', shape=[3], dtype='float32')
        x.stop_gradient = False
        x1, x2, x3 = fluid.layers.split(x, dim=0, num_or_sections=3)
        y = x1 * 2
        x1_grad = fluid.gradients(y, x)

        exe = fluid.Executor(fluid.CPUPlace())
        main = fluid.default_main_program()
        out = exe.run(main,
                      feed={'x': np.ones([3]).astype('float32')},
                      fetch_list=[x1_grad])
        self.assertTrue(np.array_equal(out[0], [2., 0., 0.]))
```

### DIFF
--- a/python/paddle/fluid/backward.py
+++ b/python/paddle/fluid/backward.py
@@ -1751,6 +1751,11 @@ def calc_gradient(targets, inputs, target_gradients=None, no_grad_set=None):
         if input.block.program != prog:
             raise "input must be in the same program as targets"
 
+    # find no grad var by op_path
+    no_grad_vars = _find_no_grad_vars(block, op_path, targets,
+                                      block_no_grad_set)
+    block_no_grad_set.update(no_grad_vars)
+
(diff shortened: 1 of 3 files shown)


chenwhql (approved):
LGTM

liym27 (approved):
LGTM 👍

XiaoguangHu01 (approved):
LGTM

PR #27248, opened 2020-09-10
Temporally disable zero_copy

zhiqiu:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Temporally disable zero_copy of `to_variable` and `to_tensor`
### Why?
we found two problems when enable zero_copy on CPUPlace.
- Eigen requires 16-bytes alignments, but the data of numpy array may not always satisfy the condition. 
Details: https://eigen.tuxfamily.org/dox/group__TopicUnalignedArrayAssert.html
- When used in flask framework, it may result in hang.
Details: https://github.com/PaddlePaddle/Paddle/issues/26635
- zero_copy is only used in CPUPlace. For training on CPU, the computation is not as fast as on GPU, so the time consumption of doing copy may not be the major issue of performance.

So, we temporally diable the zero_copy strategy.

### DIFF
--- a/python/paddle/tensor/creation.py
+++ b/python/paddle/tensor/creation.py
@@ -208,20 +208,20 @@ def to_tensor(data, dtype=None, place=None, stop_gradient=True):
             value=data,
             place=place,
             persistable=False,
-            zero_copy=True,
+            zero_copy=False,
             stop_gradient=stop_gradient)
     else:
         name = unique_name.generate('generated_tensor')
         real_tensor = paddle.Tensor(
             value=data.real,
             place=place,
-            zero_copy=True,
+            zero_copy=False,
             name=name + ".real",
             stop_gradient=stop_gradient)
         imag_tensor = paddle.Tensor(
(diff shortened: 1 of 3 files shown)


phlrain (dismissed):

python/paddle/tensor/creation.py
```diff
@@ -208,20 +208,20 @@ def to_tensor(data, dtype=None, place=None, stop_gradient=True):
             value=data,
             place=place,
             persistable=False,
-            zero_copy=True,
+            zero_copy=False,
```
zhwesky2010, line 210: the doc:" if the data is an numpy\.ndarray of with the same ``dtype``  and the current place is cpu, no copy will be performed". should be changed too.

zhiqiu: done, thx

zhwesky2010 (approved):
LGTM

jzhang533 (approved):
lgtm

PR #27317, opened 2020-09-15
Replace the 'spawn' start method with 'fork' start method for multiprocessing, on MacOS with python>=3.8

gfwm2013:
### PR types
Bug fixes

### PR changes
Others

### Describe
On MacOS，the spawn start method is now the default in Python3.8 multiprocessing。But paddle currently can not handle this, so force process to start using 'fork' start method.

More details, please refer to:
https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
https://bugs.python.org/issue33725

### DIFF
--- a/python/paddle/reader/decorator.py
+++ b/python/paddle/reader/decorator.py
@@ -32,6 +32,21 @@ import random
 import zlib
 import paddle.compat as cpt
 
+# On macOS, the 'spawn' start method is now the default in Python3.8 multiprocessing,
+# Paddle is currently unable to solve this, so forces the process to start using 
+# the 'fork' start method.
+#
+# TODO: This solution is not good, because the fork start method could lead to 
+# crashes of the subprocess. Figure out how to make 'spawn' work.
+#
+# For more details, please refer to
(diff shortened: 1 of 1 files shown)


python/paddle/reader/decorator.py
```diff
@@ -32,6 +32,21 @@
...
+# Paddle is currently unable to solve this, so forces the process to start using 
+# the 'fork' start method.
+#
+# TODO: This solution is not good, because the fork start method could lead to 
+# crashes of the subprocess. Figure out how to make 'spawn' work.
+#
+# For more details, please refer to
+# https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+# https://bugs.python.org/issue33725
+if sys.version_info >= (3, ):
```
zhhsplendid: If this only fails for Python3.8, could we only use fork when the version_info refers to Python 3.8?

gfwm2013: Thanks, done.

zhhsplendid (approved):
LGTM

PR #27325, opened 2020-09-15
Add conv2d bfloat16 support

wozna:
### PR types

New features
### PR changes

OPs 
### Describe

This PR presents:
- fix pybind conversion from uint16 to bfloat16. Numpy doesn't support bfloat16 data type, so in python bfloat16 is represented by uint16. In UT all operations can be done on float32 and then using `convert_float32_to_uint16`  from `OpTest` converted to uint16.
- conv2d and dequantize bfloat16 support with unit tests. In both unit tests `atol` in check output is changed, because we compare native convolution that was calculated on float32 with conv2d bf16 kernel that was calculated on bfloat16. 

>  When the output from the convolution is:
> - bfloat16, in Python we get uint16 data type, so `atol = 2`
> - float32, so `atol = 1e-2` 
>

### DIFF
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -1895,7 +1895,7 @@ PDNode *patterns::QuantizePlacement::operator()(
 PDNode *patterns::Bfloat16Placement::operator()(
     const std::unordered_set<std::string> &bfloat16_enabled_op_types) {
   std::unordered_set<std::string> supported_op_types =
-      std::unordered_set<std::string>();
+      std::unordered_set<std::string>({"conv2d"});
   if (!bfloat16_enabled_op_types.empty()) {
     supported_op_types = bfloat16_enabled_op_types;
   }
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -1894,7 +1894,7 @@ PDNode *patterns::QuantizePlacement::operator()(
 
 PDNode *patterns::Bfloat16Placement::operator()(
     const std::unordered_set<std::string> &bfloat16_enabled_op_types) {
-  std::unordered_set<std::string> supported_op_types {"conv2d"};
+  std::unordered_set<std::string> supported_op_types{"conv2d"};
   if (!bfloat16_enabled_op_types.empty()) {
     supported_op_types = bfloat16_enabled_op_types;
   }
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -1894,8 +1894,7 @@ PDNode *patterns::QuantizePlacement::operator()(
 
 PDNode *patterns::Bfloat16Placement::operator()(
     const std::unordered_set<std::string> &bfloat16_enabled_op_types) {
-  std::unordered_set<std::string> supported_op_types =
-      std::unordered_set<std::string>({"conv2d"});
+  std::unordered_set<std::string> supported_op_types {"conv2d"};
   if (!bfloat16_enabled_op_types.empty()) {
     supported_op_types = bfloat16_enabled_op_types;
   }
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -51,12 +51,13 @@ inline MKLDNNMemoryFormat GetWeightsFormat(const MKLDNNMemoryFormat format,
   }
 }
 
-static mkldnn::memory::data_type GetDstType(bool is_int8,
+static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
                                             bool force_fp32_output,
                                             std::string fuse_activation,
                                             bool fuse_residual_conn,
                                             const Tensor* residual_param) {
-  auto dst_dt = mkldnn::memory::data_type::f32;  // uint8_t, int8_t, float
+  auto dst_dt =
+      mkldnn::memory::data_type::f32;  // uint8_t, int8_t, float, bfloat16
   if (is_int8) {
     dst_dt = (fuse_activation == "relu" || fuse_activation == "relu6")
                  ? mkldnn::memory::data_type::u8
@@ -67,6 +68,13 @@ static mkldnn::memory::data_type GetDstType(bool is_int8,
     if (fuse_residual_conn && residual_param) {
       auto residual_dt = framework::ToMKLDNNDataType(residual_param->type());
       if (dst_dt != residual_dt) dst_dt = residual_dt;
+    }
+  } else {
+    if (!force_fp32_output && is_bfloat16) {
+      dst_dt = mkldnn::memory::data_type::bf16;
+      if (fuse_residual_conn && residual_param) {
+        dst_dt = framework::ToMKLDNNDataType(residual_param->type());
+      }
     }
   }
   return dst_dt;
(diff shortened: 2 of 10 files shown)


jczaja (approved):
LGTM

paddle/fluid/framework/ir/graph_pattern_detector.cc
```diff
@@ -1895,7 +1895,7 @@ PDNode *patterns::QuantizePlacement::operator()(
 PDNode *patterns::Bfloat16Placement::operator()(
     const std::unordered_set<std::string> &bfloat16_enabled_op_types) {
   std::unordered_set<std::string> supported_op_types =
-      std::unordered_set<std::string>();
+      std::unordered_set<std::string>({"conv2d"});
```
wojtuss: This can be simplified to:
```
std::unordered_set<std::string> supported_op_types {"conv2d"};
```
or
```
auto supported_op_types = std::unordered_set<std::string>({"conv2d"});
```

jczaja (dismissed):

jczaja:

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -51,12 +51,13 @@ inline MKLDNNMemoryFormat GetWeightsFormat(const MKLDNNMemoryFormat format,
   }
 }
 
-static mkldnn::memory::data_type GetDstType(bool is_int8,
+static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
```
luotao1, line 54: Since there is `mkldnn_data_type`, why not `GetDstType(std::string  mkldnn_data_type, xxx)`?

python/paddle/fluid/framework.py
```diff
@@ -603,7 +603,7 @@ def convert_np_dtype_to_dtype_(np_dtype):
     elif dtype == np.bool:
         return core.VarDesc.VarType.BOOL
     elif dtype == np.uint16:
-        return core.VarDesc.VarType.INT16
+        return core.VarDesc.VarType.BF16
```
luotao1, line 608: Why change here?

luotao1 (commented):
```
2020-09-17 17:40:25 1. Developers are not allowed to set the check_dygraph field directly, which is set to True by default. If you need to change the check_dygraph field, you must have one RD (phlrain (Recommend) or lanxianghit) review and approve. 
2020-09-17 17:40:25 The code that do not meet the specification are as follows:
2020-09-17 17:40:25  python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_bf16_mkldnn_op.py : 
2020-09-17 17:40:25 +            check_dygraph=False) 
2020-09-17 17:40:25  python/paddle/fluid/tests/unittests/mkldnn/test_dequantize_mkldnn_op.py : 
2020-09-17 17:40:25 +        self.check_output(atol=1e-2, check_dygraph=False) 
2020-09-17 17:40:25 2. Please use the default precision parameters of 'atol, rtol, eps, max_relative_error'. If you don't use the default value, you must have one RD (Xreki (Recommend), luotao1, lanxianghit or phlrain) approval for the usage of other values. The detailed information is in the link: https://github.cor/PaddlePaddle/Paddle/wiki/OP-test-accuracy-requirements. The error line is 
2020-09-17 17:40:25 +        self.check_output(atol=1e-2, check_dygraph=False)
2020-09-17 17:40:25 3. It is an Op accuracy problem, please take care of it. You must have one RD (zhangting2020 (Recommend), luotao1 or phlrain) approval for the usage (either add or delete) of @skip_check_grad_ci. For more information, please refer to: https://github.com/PaddlePaddle/Paddle/wiki/Gradient-Check-Is-Required-for-Op-Test. The corresponding lines are as follows:
2020-09-17 17:40:25 python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_bf16_mkldnn_op.py
2020-09-17 17:40:25 +	@skip_check_grad_ci(
2020-09-17 17:40:25 There are 4 approved errors.
2020-09-17 17:40:25 ****************
```
Since you will add several bfloat16 op support later:
- If mkldnn bfloat16 ops don't support dygraph, don't need grad op, and have its own accuracy. you should update `op_test.py`, and not for each time approval.

python/paddle/fluid/framework.py
wozna: There is no support for bfloat16 in NumPy.  
We would like to test bfloat16 operators not only with input data float32, but also with bfloat16, so it converts `np.uint16` data to C++ `bfloat16`.  
It was also added to `tensor_py.h` https://github.com/PaddlePaddle/Paddle/blob/c0caf0e45fac2bbbae184a702a4fe0313a9c3c69/paddle/fluid/pybind/tensor_py.h#L272 
Importantly, there is no other usage of `np.uint16` in Paddle.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
wozna: You are right. I corrected it.

wozna: I checked it and unfortunately, it won't work. In int8 models during warmup all conv operators are already marked as `int8` in `mkldnn_data_type` , but during warmup, computations should be run as `float32`  that is why we need to check the tensor type for int8.

wozna:
@luotao1  @jczaja  I finished all changes after which the only one approved for `framework.py` file is needed.
Could you review those changes?

jczaja (dismissed):
LGTM

paddle/fluid/operators/mkldnn/dequantize_mkldnn_op.cc
```diff
@@ -110,4 +110,6 @@ class DeQuantOpKernel : public framework::OpKernel<T> {
 namespace ops = paddle::operators;
 
 REGISTER_OP_KERNEL(dequantize, MKLDNN, ::paddle::platform::CPUPlace,
-                   ops::DeQuantOpKernel<uint8_t>, ops::DeQuantOpKernel<int8_t>);
+                   ops::DeQuantOpKernel<uint8_t>, ops::DeQuantOpKernel<int8_t>,
+                   ops::DeQuantOpKernel<paddle::platform::bfloat16>,
+                   ops::DeQuantOpKernel<float>);
```
wojtuss: Do we need dequantize from `float`?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -380,14 +391,23 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
         std::is_same<T, int8_t>::value || std::is_same<T, uint8_t>::value;
+    bool is_BFLOAT16 = ctx.Attr<std::string>("mkldnn_data_type") == "bfloat16";
+    auto residual_param = ctx.Input<Tensor>("ResidualData");
+    bool fuse_residual_conn = ctx.Attr<bool>("fuse_residual_connection");
+    std::string fuse_activation = ctx.Attr<std::string>("fuse_activation");
+    bool force_fp32_output = ctx.Attr<bool>("force_fp32_output");
+
     if (!is_INT8) {
-      ComputeFP32<float>(ctx);
+      auto dst_dt =
```
wojtuss: This could be simplified:
```
auto dst_dt = GetDstType(is_INT8, is_BFLOAT16, force_fp32_output, fuse_activation,
                     fuse_residual_conn, residual_param);
if (!is_INT8) {
   if (dst_dt = ...) ...
} else {
   if (dst_dt = ...) ...
}
```

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -51,12 +51,13 @@ inline MKLDNNMemoryFormat GetWeightsFormat(const MKLDNNMemoryFormat format,
...
 
-static mkldnn::memory::data_type GetDstType(bool is_int8,
+static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
                                             bool force_fp32_output,
                                             std::string fuse_activation,
                                             bool fuse_residual_conn,
                                             const Tensor* residual_param) {
-  auto dst_dt = mkldnn::memory::data_type::f32;  // uint8_t, int8_t, float
+  auto dst_dt =
+      mkldnn::memory::data_type::f32;  // uint8_t, int8_t, float, bfloat16
```
wojtuss: I think this comment could be removed completely.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -167,6 +168,22 @@ def wrapper(cls):
...
+
+
+def convert_float_to_uint16(float_list):
+    new_output = []
+    for first in float_list:
+        for second in first:
+            for third in second:
+                for fourth in third:
+                    new_output.append(
+                        np.uint16(copy_bits_from_float_to_uint16(fourth)))
```
wojtuss: This could be written shorter as:
```
for x in np.nditer(float_list):
   new_output.append(np.uint16(copy_bits_from_float_to_uint16(x)))
```

paddle/fluid/operators/mkldnn/dequantize_mkldnn_op.cc
wozna: You right, we don't need it. I just missed it.

python/paddle/fluid/tests/unittests/op_test.py
wozna: Yes, it's great improvement :)

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
wozna: Done

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
wozna: Done

paddle/fluid/framework/ir/graph_pattern_detector.cc
wozna: Done

wojtuss (dismissed):
LGTM

luotao1 (approved):
LGTM

PR #27334, opened 2020-09-15
judge whether remove build dir to accelerate compile

zhwesky2010:
### PR types

Performance optimization
### PR changes

Others
### Describe

If keep build dir cache, the compilation will be fasted a lot, the following measures have been taken to reduce risk :

1. Judge whether remove build dir to accelerate compile. Any commits that change compilation organization will not use the build cache.
2. The first commit of each day will not use the build cache.
3. Add switch ``WITH_CACHE`` to control whether to use the build cache.


luotao1 (approved):
LGTM

PR #27344, opened 2020-09-16
Fix bugs in hdfs download

qjing666:
### PR types
Bug fixes

### PR changes
APIs

### Describe
Original hdfs downloader will check if local_path exists on hdfs, which will cause program timeout, remove the check. 

Fix bug: fleet._role_maker return None


guru4elephant (approved):
LGTM

PR #27355, opened 2020-09-16
python2 of windows don't need to install netifaces

zhwesky2010:
### PR types

Bug fixes
### PR changes

Others
### Describe

python2 for windows not install netifaces, because windows don't have distribution now. 

``netifaces`` must use Visual C++ tools for python2 windows. It can be remove in python2 windows.
https://www.microsoft.com/en-us/download/details.aspx?id=44266

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -20,4 +20,5 @@ prettytable
 objgraph
 astor
 pathlib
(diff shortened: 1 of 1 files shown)


luotao1 (approved):
LGTM

PR #27362, opened 2020-09-16
fix code of adaptive lsgd to avoid the conflict

ForFishes:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix code of adaptive lsgd to avoid the conflict


gavin1332 (approved):
LGTM

PR #27376, opened 2020-09-17
Remove inplace argument when calling nn.reshape()

zhiqiu:
### PR types

Others
### PR changes

OPs
### Describe

Remove inplace argument when calling `fluid.layers.reshape()`

There are 2 reasons,

1. `fluid.layers.reshape()` do not support inplace in dygraph mode, while dygraph will be default in Paddle-2.0. So we need to avoid annoying warnings in Paddle-2.0 as follows.

image

2. `fluid.layers.reshape(inplace=True)` will be replace by  `BufferSharedIdentityInplaceOpPass`, see details in #25900

### DIFF
--- a/python/paddle/fluid/dygraph/parallel.py
+++ b/python/paddle/fluid/dygraph/parallel.py
@@ -416,7 +416,7 @@ class DataParallel(layers.Layer):
                 g_var_shapes.append(g_var.shape)
                 flattened_vars.append(
                     nn.reshape(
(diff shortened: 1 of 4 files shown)


iclementine (dismissed):
LGTM

jzhang533 (commented):
￼lgtm

jzhang533 (dismissed):

jzhang533 (approved):
lgtm

PR #27405, opened 2020-09-18
fix fleetrun port_env bug

danleifeng:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix fleetrun port_env bug
os.env return string but need int value


sandyhouse (approved):
LGTM

PR #27409, opened 2020-09-18
Correct the error in decorator.py

gfwm2013:
### PR types
Bug fixes

### PR changes
 Others

### Describe
Background:
On MacOS，the spawn start method is now the default in Python3.8 multiprocessing。But paddle currently can not handle this, so force process to start using 'fork' start method. 

This method would be failed on windows platform, so add a judgment statement to limit the method to the MacOS platform.


zhhsplendid (approved):
LGTM

PR #27437, opened 2020-09-21
Fix GradientClipByGlobalNorm dtype bug

zhwesky2010:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix GradientClipByGlobalNorm dtype bug.

```
import paddle
import numpy as np
paddle.disable_static()

paddle.set_default_dtype('float64')

linear = paddle.nn.Linear(10, 10)
print(linear.weight.dtype)
print(linear.bias.dtype)

out = linear(paddle.to_tensor(np.random.rand(10, 10)))
out.backward()

optimizer = paddle.optimizer.Adam(0.1, parameters=linear.parameters(), grad_clip=paddle.nn.GradientClipByGlobalNorm(1.0))
optimizer.step()
```

image


python/paddle/fluid/clip.py
```diff
@@ -590,7 +591,9 @@ def _dygraph_clip(self, params_grads):
         global_norm_var = layers.reduce_sum(global_norm_var)
         global_norm_var = layers.sqrt(global_norm_var)
         max_global_norm = layers.fill_constant(
-            shape=[1], dtype='float32', value=self.clip_norm)
```
zhiqiu, line 593: Better infer dtype by `global_norm_var`?

python/paddle/fluid/clip.py
```diff
@@ -635,7 +638,9 @@ def _static_clip(self, params_grads):
                 global_norm_var = layers.sums(sum_square_list)
                 global_norm_var = layers.sqrt(x=global_norm_var)
                 max_global_norm = layers.fill_constant(
-                    shape=[1], dtype="float32", value=self.clip_norm)
```
zhiqiu, line 638: Same above.

zhwesky2010: Ok

python/paddle/fluid/clip.py
zhwesky2010: Done

zhiqiu (approved):
LGTM

PR #27447, opened 2020-09-21
Fix bug: The calculation result of Diag_v2 Op under large size input is wrong

LutaoChu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Fix bug: The calculation result of Diag_v2 Op under large size input is wrong. The large size here refers to more than 1024.


wawltor (approved):
LGTM

Issue #25506, opened 2020-07-13
Memory leaks when doing CPU inference with MKLDNN


cryoco:
**System information**
Paddle version: 1.8.2
Paddle With CUDA: False
OS: Ubuntu 16.04
CPU: 16  Intel(R) Xeon(R) CPU E5-2620 v3 @ 2.40GHz
Python version: 3.5.2
CUDA version: 9.0.176
cuDNN version: None.None.None
Nvidia driver version: None
API information: inference configuration
```python
config.disable_gpu()
config.enable_mkldnn()
config.set_cpu_math_library_num_threads(4)
```
**To Reproduce**
1. download models and data
- [ocr detection model](https://paddleocr.bj.bcebos.com/ch_models/ch_det_mv3_db.tar)
- [ocr recognition model](https://paddleocr.bj.bcebos.com/ch_models/ch_rec_mv3_crnn.tar)
- [test data](https://paddle-inference-dist.bj.bcebos.com/temp/ocr_test_data.tgz)
The test data pictures are binary files saved with `np.save`, and can be loaded with `np.load`.
2. perform inference with mkldnn
code to reproduce:
[test_ocr_mkldnn_mem.txt](https://github.com/PaddlePaddle/Paddle/files/4912792/test_ocr_mkldnn_mem.txt)(rename to test_ocr_mkldnn_mem.py)

- detection
```python
python3 test_ocr_mkldnn_mem.py --model_file=./ch_det_mv3_db/model --params_file=./ch_det_mv3_db/params --mode=det --mkldnn
```
- recognition
```python
python3 test_ocr_mkldnn_mem.py --model_file=./ch_rec_mv3_crnn/model --params_file=./ch_rec_mv3_crnn/params --mode=rec --mkldnn
```
  Memory usage increasing can be witnessed with `top` command.

3. perform inference without mkldnn
- detection
```python
python3 test_ocr_mkldnn_mem.py --model_file=./ch_det_mv3_db/model --params_file=./ch_det_mv3_db/params --mode=det 
```
- recognition
```python
python3 test_ocr_mkldnn_mem.py --model_file=./ch_rec_mv3_crnn/model --params_file=./ch_rec_mv3_crnn/params --mode=rec
```
Memory usage will maintain stable.

The input shapes we use in OCR are dynamic, which might be relevant to this issue.

luotao1:
Please use `SetMkldnnCacheCapacity` interface.
[...]

PR #27493, opened 2020-09-23
Fix memory leak for mkldnn by using zero_copy_run.

jiweibo:
### PR types

Bug fixes
### PR changes

APIs
### Describe

When mkldnn is enabled, for dynamic input shape, we need to set cache capacity to prevent memeory leaks.

According to the implementation of Predictor::Run, we fix the bug of Predictor::ZeroCopyRun.

Related pr: https://github.com/PaddlePaddle/Paddle/issues/25506


cryoco (approved):
LGTM

PR #27498, opened 2020-09-23
Decrease Random Failure Probability for test_parallel_executor_mnist

zhhsplendid:
### PR types

Others
### PR changes

Others
### Describe

As the title, decrease random failure probability for test_parallel_executor_mnist

The old code set larger delta when comparing reduce and all reduce, but didn't set all. I added it.

On my linux machine, I run 100 times, no failure occurs. In addition, we only saw this random failure on CI two times since I worked. I thought it was rare and I just increased the delta.


Aurelius84 (approved):

liym27 (approved):
LGTM

chalsliu (approved):

PR #27549, opened 2020-09-24
Fix batch norm double grad compute

ceci3:
### PR types

Bug fixes

### PR changes

OPs 

### Describe

fix batch norm double grad compute when use_global_state=True
add more unittest


qingqing01 (approved):
LGTM

PR #27604, opened 2020-09-27
[Cherry-Pick] Add Log double grad kernel

joey12300:
### PR types

Function optimization
### PR changes

OPs
### Describe

Add Log double grad kernel

* add log double grad functor
* register log double grad kernel for cpu and cuda
* add unittest for log double grad


wawltor (approved):
LGTM

PR #27610, opened 2020-09-27
optimize the speed&memory of matmul op

ForFishes:
### PR types

Function optimization
### PR changes

APIs
### Describe

optimize the speed&memory of matmul op


jzhang533 (approved):
lgtm

PR #27706, opened 2020-09-29
[bug fix] avoiding multiple initialization of gloo for fleet in dygraph mode

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Describe

avoiding multiple initialization of gloo for fleet in dygraph mode


python/paddle/distributed/fleet/base/role_maker.py
```diff
@@ -869,7 +870,8 @@ def _generate_role(self):
             else:
                 self._collective_env()
             self._role_is_generated = True
-            self._gloo_init()
+            if not paddle.fluid.framework.in_dygraph_mode():
```
guru4elephant, line 873: it seems in dygraph mode, we do not need to do gloo_init, but actually it is initialized in other places. In general, this is not a good design, and should be refactored with init_parallel_env in the future.

guru4elephant (approved):
LGTM, not a good design, but ok

PR #27709, opened 2020-09-29
Update initializer examples of Bilinear

Ray2020BD:
### PR types

Function optimization
### PR changes

APIs
### Describe

Update initializer examples of Bilinear

### DIFF
--- a/python/paddle/fluid/initializer.py
+++ b/python/paddle/fluid/initializer.py
@@ -734,7 +734,6 @@ class BilinearInitializer(Initializer):
             import math
             import paddle.nn as nn
 
-            paddle.disable_static()
             factor = 2
             C = 2
             B = 8
--- a/python/paddle/fluid/initializer.py
+++ b/python/paddle/fluid/initializer.py
@@ -729,36 +729,26 @@ class BilinearInitializer(Initializer):
 
         .. code-block:: python
 
-            import paddle.fluid as fluid
+            import numpy as np
+            import paddle
+            from paddle.regularizer import L2Decay
             import math
-            factor = 2
-            C = 2
-            B = 8
-            H = W = 32
-            w_attr = fluid.param_attr.ParamAttr(
-                learning_rate=0., 
-                regularizer=fluid.regularizer.L2Decay(0.),
-                initializer=fluid.initializer.Bilinear())
-            x = fluid.data(name="data", shape=[B, 3, H, W], 
-                                  dtype="float32")
-            conv_up = fluid.layers.conv2d_transpose(
-                input=x,
-                num_filters=C,
-                output_size=None,
-                filter_size=2 * factor - factor % 2,
-                padding=int(math.ceil((factor - 1) / 2.)),
-                stride=factor,
-                groups=C,
-                param_attr=w_attr,
-                bias_attr=False)
+            import paddle.nn as nn
 
-    Where, `num_filters=C` and `groups=C` means this is channel-wise transposed
-    convolution. The filter shape will be (C, 1, K, K) where K is `filer_size`,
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/initializer.py
```diff
@@ -729,36 +729,26 @@ class BilinearInitializer(Initializer):
...
-    This initializer will set a (K, K) interpolation kernel for every channel
-    of the filter identically. The resulting shape of the output feature map
-    will be (B, C, factor * H, factor * W). Note that the learning rate and the
-    weight decay are set to 0 in order to keep coefficient values of bilinear
-    interpolation unchanged during training.
+            w_attr = paddle.ParamAttr(learning_rate=0.,
+                                      regularizer=L2Decay(0.),
+                                      initializer=nn.initializer.Bilinear())
+            data = np.random.uniform(-1, 1, [30, 10, 32]).astype('float32')
+            data = paddle.to_tensor(data)
```
jzhang533: 这里直接用`paddle.rand`吧。
上面也不需要import numpy和disable_static了。
另外上面的where, ... 那部分描述还要留着呀。

Ray2020BD: 好的，已经完成了

python/paddle/fluid/initializer.py
```diff
@@ -729,31 +729,32 @@ class BilinearInitializer(Initializer):
 
         .. code-block:: python
 
-            import paddle.fluid as fluid
+            import paddle
```
zhhsplendid: Suggest to change import order. Usually we group the imports by function, then sort it by alphabetical order. Here `import math` is in the middle of import paddle, which is ugly.

Suggest to change like this:

```
import math

import paddle
import paddle.nn
from paddle.regularizer import L2Decay
```

python/paddle/fluid/initializer.py
zhhsplendid: paddle.rand返回已经是个tensor，所以不用data = paddle.to_tensor(data)，请删除745行

python/paddle/fluid/initializer.py
Ray2020BD: I've already modified.

python/paddle/fluid/initializer.py
```diff
@@ -729,31 +729,34 @@ class BilinearInitializer(Initializer):
...
-                param_attr=w_attr,
-                bias_attr=False)
-
-    Where, `num_filters=C` and `groups=C` means this is channel-wise transposed
-    convolution. The filter shape will be (C, 1, K, K) where K is `filer_size`,
+            w_attr = paddle.ParamAttr(learning_rate=0.,
+                                      regularizer=L2Decay(0.),
+                                      initializer=nn.initializer.Bilinear())
+            data = paddle.rand([B, 3, H, W], dtype='float32')
+            #data = paddle.to_tensor(data)
```
zhhsplendid: Delete, not comment. Why comment? This example code is used for user reading, if you put comment here, it may confuse user!

python/paddle/fluid/initializer.py
zhhsplendid: Could you just put in order? You changed it like:

import xxx
from xxx import
import xxx

Why can't you just make it pretty? Since all xxx are `paddle`

import xxx
import xxx
from xxx import

python/paddle/fluid/initializer.py
Ray2020BD: Already delete.

zhhsplendid (approved):
LGTM

jzhang533 (approved):
lgtm

PR #27718, opened 2020-09-29
fix pool3d bug, test=develop

LDOUBLEV:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix pool3d bug, http://newicafe.baidu.com/v5/issue/DLTP-12154/show?cid=5

### DIFF
--- a/paddle/fluid/operators/math/pooling.cc
+++ b/paddle/fluid/operators/math/pooling.cc
@@ -672,6 +672,7 @@ class Pool3dFunctor<platform::CPUDeviceContext, PoolProcess, T> {
     int dstart, dend;
     int hstart, hend;
     int wstart, wend;
+
     for (int i = 0; i < batch_size; i++) {
       for (int c = 0; c < output_channels; ++c) {
         for (int pd = 0; pd < output_depth; ++pd) {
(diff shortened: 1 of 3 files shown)


paddle/fluid/operators/math/pooling.cc
```diff
@@ -712,10 +720,9 @@ class Pool3dFunctor<platform::CPUDeviceContext, PoolProcess, T> {
                   }
                 }
               }
-              int pool_size =
-                  (exclusive || adaptive)
-                      ? (dend - dstart) * (hend - hstart) * (wend - wstart)
-                      : ksize_depth * ksize_height * ksize_width;
+              if (exclusive || adaptive) {
```
willthefrog, line 760: this check seems redundant now, isn't `pool_size` always the same value?

willthefrog (changes requested):
also fix the 2d and cuda version?

paddle/fluid/operators/math/pooling.cc
LDOUBLEV: It is not redundant to update`pool_size`  when `padding ` is not zero.
The variable `pool_size` will be updated after running the following code.
```
pool_size = (dend - dstart) * (hend - hstart) * (wend - wstart);
dstart = std::max(dstart, 0);
hstart = std::max(hstart, 0);
wstart = std::max(wstart, 0);
dend = std::min(dend, input_depth);
hend = std::min(hend, input_height);
wend = std::min(wend, input_width);
```

LDOUBLEV:
> also fix the 2d and cuda version?

Got it!

ceci3 (approved):

willthefrog (approved):
LGTM.

the cuda version seems unused, maybe remove later.

PR #27719, opened 2020-09-29
【paddle.fleet】geo send sparse optimize

123malin:
### PR types

Bug fixes

### PR changes

Others

### Describe

1. change geo_communicator send content: from sparse id to grad
2. default ps gloo init method: from file to http


seiriosPlus (approved):
LGTM

PR #27720, opened 2020-09-29
add type check

hutuxian:
### PR types

Function optimization
### PR changes

APIs
### Describe

add type check for `initializer ` and `regularizer ` in ParamAttr


gavin1332 (approved):
LGTM

Issue #27588, opened 2020-09-27
test_conv2d_bf16_mkldnn_op fails on 5117 machine


luotao1:
test_conv2d_bf16_mkldnn_op fails on 5117 machine (Nightly Job)
- error log
```
[19:20:48] :	 [Step 1/2] The following tests FAILED:
[19:20:48] :	 [Step 1/2] 	1179 - test_conv2d_bf16_mkldnn_op (Failed)
[19:20:48] :	 [Step 1/2] At least one test failed with exit code => 0
[19:20:48] :	 [Step 1/2] At least one test failed with exit code => 0
[19:20:48] :	 [Step 1/2] 1/1 Test #1179: test_conv2d_bf16_mkldnn_op .......***Failed    1.98 sec
[19:20:48] :	 [Step 1/2] W0926 19:20:48.445756 51794 operator.cc:205] conv2d raises an exception dnnl::error, could not create a primitive descriptor iterator
[19:20:48] :	 [Step 1/2] W0926 19:20:48.461982 51794 operator.cc:205] conv2d raises an exception dnnl::error, could not create a primitive descriptor iterator
[19:20:48] :	 [Step 1/2] W0926 19:20:48.662735 51794 operator.cc:205] conv2d raises an exception dnnl::error, could not create a primitive descriptor iterator
[19:20:48] :	 [Step 1/2] W0926 19:20:48.681520 51794 operator.cc:205] conv2d raises an exception dnnl::error, could not create a primitive descriptor iterator
[19:20:48] :	 [Step 1/2] W0926 19:20:48.695686 51794 operator.cc:205] conv2d raises an exception dnnl::error, could not create a primitive descriptor iterator
[19:20:48] :	 [Step 1/2] W0926 19:20:48.721101 51794 operator.cc:205] conv2d raises an exception dnnl::error, could not create a primitive descriptor iterator
[19:20:48] :	 [Step 1/2] W0926 19:20:48.740229 51794 operator.cc:205] conv2d raises an exception dnnl::error, could not create a primitive descriptor iterator
[19:20:48] :	 [Step 1/2] test_conv2d_bf16_mkldnn_op failed
[19:20:48] :	 [Step 1/2]  ...E...E.......E...E...E...E...E
[19:20:48] :	 [Step 1/2] ======================================================================
[19:20:48] :	 [Step 1/2] ERROR: test_check_output (test_conv2d_bf16_mkldnn_op.TestConv2d)
[...]

PR #27732, opened 2020-09-29
Add avx512 core instructions check

wozna:
### PR types

Bug fixes
### PR changes

OPs 
### Describe

This PR adds avx512 core instruction check in bfloat16 convolution test. Bfloat16 evaluation is possible only if the CPU has ` AVX512BW, AVX512VL, and AVX512DQ` flags. Without them, all tests with bfloat16 will fail.
It is related to https://github.com/PaddlePaddle/Paddle/issues/27588

### DIFF
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -142,6 +142,17 @@ bool IsCompiledWithMKLDNN() {
 #endif
 }
 
+bool IsBfloat16() {
+#ifndef PADDLE_WITH_MKLDNN
+  return false;
+#else
+  if (platform::MayIUse(platform::cpu_isa_t::avx512_core))
+    return true;
+  else
+    return false;
+#endif
+}
+
 bool IsCompiledWithBrpc() {
 #ifndef PADDLE_WITH_DISTRIBUTE
   return false;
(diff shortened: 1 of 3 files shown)


paddle/fluid/pybind/pybind.cc
```diff
@@ -142,6 +142,17 @@ bool IsCompiledWithMKLDNN() {
 #endif
 }
 
+bool IsBfloat16() {
```
wojtuss: I would suggest renaming it into something like "SupportsBfloat16" or "CanRunBfloat16", as `IsBfloat16` does not convey the meaning of the function.

paddle/fluid/pybind/pybind.cc
```diff
@@ -1661,6 +1672,7 @@ All parameter, weight, gradient are variables in Paddle.
   m.def("is_compiled_with_cuda", IsCompiledWithCUDA);
   m.def("is_compiled_with_xpu", IsCompiledWithXPU);
   m.def("is_compiled_with_mkldnn", IsCompiledWithMKLDNN);
+  m.def("is_bfloat16", IsBfloat16);
```
wojtuss: ditto

paddle/fluid/pybind/pybind.cc
wozna: Yeah, you are right, I like `SupportsBfloat16`  :)

wojtuss (commented):
LGTM

jczaja (approved):
LGTM

wojtuss (approved):
LGTM

jczaja:

PR #27739, opened 2020-09-30
Add an option to set number of warmup iterations

wojtuss:
### PR types
Others

### PR changes
Others

### Describe
With this patch one can set the number of iterations for warmup phase in performance tests. When the option `--warmup=true` is used, by default one iteration of inference is run as a warmup. Now, with additional option `--warmup_iters=5` the given number of iterations will be run in warmup.
This option helps properly measure performance of GRU FP32/INT8 model with oneDNN. When running GRU inference with oneDNN, the oneDNN cache needs more than one (actually close to 20) iterations to cache all the primitives required during inference of the whole dataset.

### DIFF
--- a/paddle/fluid/inference/tests/api/tester_helper.h
+++ b/paddle/fluid/inference/tests/api/tester_helper.h
@@ -65,6 +65,7 @@ DEFINE_bool(zero_copy, false, "Use ZeroCopy to speedup Feed/Fetch.");
 DEFINE_bool(warmup, false,
             "Use warmup to calculate elapsed_time more accurately. "
             "To reduce CI time, it sets false in default.");
+DEFINE_int32(warmup_iters, 1, "Number of batches to process during warmup.");
 
 DEFINE_bool(enable_profile, false, "Turn on profiler for fluid");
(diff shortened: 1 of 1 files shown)


luotao1 (commented):
Does this PR make the elapsed time of the unit-test become longer since each unit-test has an extra warmup iteration? Could the default warmup_iter be 0?

wojtuss:
> Does this PR make the elapsed time of the unit-test become longer since each unit-test has an extra warmup iteration? Could the default warmup_iter be 0?

@luotao1 
The option `--wamup_iters` only changes the number of warmup iterations if the `--warmup` option is used. If the `--warmup` option is not used, or `--warmup_iters` is not used or is set to a number <=1, then it has no effect (i.e. still the default number of warmup iterations (1) is used for the `--warmup` option). If it is set to a number >1, then yes, it has impact on execution time of the tests which use this option, because more inference iterations are executed during warmup. However, for GRU model test the warmup iterations are rather quick.

luotao1:
Got it, thanks for your explanation.

luotao1 (approved):

PR #27747, opened 2020-09-30
[oneDNN] adaptive pool support

### LINKED ISSUE #27398: rec_r34_vd_tps_bilstm_attn(from PaddleOCR repo) MKLDNN prediction failed
System information
-PaddlePaddle version: develop, `d28162b97fd2d224968c18c9da735900ef280e7c`
-CPU: Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz, mkldnn enabled
-GPU: P4, CUDA10, CUDNN7.5, compile with GPU, but run cpu inference
-OS Platform: Ubuntu1604
-Dockerfile: Paddle/tools/manylinux1/Dockerfile.cuda10_cudnn7_gcc48_ubuntu16
-Python version: 3.7
-Cmake orders
-C++version.txt
-API information

```python
config = AnalysisConfig(args.model_file, args.params_file)
config.disable_gpu()
config.switch_use_feed_fetch_ops(False)
config.switch_specify_input_names(True)
config.enable_mkldnn()
```
Note: You can get most of the information by running summary_env.py.

Steps to reproduce the behavior
[mkldnn_reproduce_code.txt](https://github.com/PaddlePaddle/Paddle/files/5244239/mkldnn_reproduce_code.txt)

```shell
python3.7 mkldnn_reproduce_code.txt --model_file=./rec_r34_vd_tps_bilstm_attn/__model__ \
                                    --params_file=./rec_r34_vd_tps_bilstm_attn/__params__
```

Describe your current behavior

```shell
I0918 07:50:21.910486 27984 analysis_predictor.cc:527] ======= optimize end =======
I0918 07:50:21.910612 27984 naive_executor.cc:102] ---  skip [feed], feed -> image
I0918 07:50:21.911798 27984 naive_executor.cc:102] ---  skip [save_infer_model/scale_0.tmp_0], fetch -> fetch
I0918 07:50:21.911818 27984 naive_executor.cc:102] ---  skip [save_infer_model/scale_1.tmp_0], fetch -> fetch
W0918 07:50:21.927631 27984 operator.cc:205] pool2d raises an exception dnnl::error, could not create a descriptor for a pooling forward propagation primitive
  File "test_zerotensor.py", line 44, in <module>
    main()
  File "test_zerotensor.py", line 20, in main
    predictor.zero_copy_run()
RuntimeError: could not create a descriptor for a pooling forward propagation primitive
```

Code to reproduce the issue
[mkldnn_reproduce_code.txt](https://github.com/PaddlePaddle/Paddle/files/5244239/mkldnn_reproduce_code.txt)

Other info / logs
image

jczaja:
### PR types
New features

### PR changes
OPs

### Describe
Added support for adaptive pooling (It was missing in oneDNN kernel) e.g. handling attribute "adaptive" and act accordingly. This is first of two fixes needed to fix #27398

### DIFF
--- a/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
@@ -126,21 +126,13 @@ class PoolMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
     UpdatePadding(&paddings, global_pooling, 0, padding_algorithm, data_dims,
                   strides, ksize);
 
-
     const bool adaptive = ctx.Attr<bool>("adaptive");
 
-    const auto src_tz = paddle::framework::vectorize(in_x->dims());
-    if (adaptive) {
-      ksize[0] = static_cast<int>(
-          ceil(static_cast<double>(src_tz[src_tz.size() - 2] / ksize[0])));
-      ksize[1] = static_cast<int>(
-          ceil(static_cast<double>(src_tz[src_tz.size() - 1] / ksize[1])));
-      strides[0] = ksize[0];
-      strides[1] = ksize[1];
-    }
+    PoolingMKLDNNHandler::ComputeAdaptivePoolParameters(
+        ctx, paddle::framework::vectorize(in_x->dims()), ksize, strides)
 
-    auto& dev_ctx =
-        ctx.template device_context<platform::MKLDNNDeviceContext>();
+        auto& dev_ctx =
+            ctx.template device_context<platform::MKLDNNDeviceContext>();
 
     std::vector<mkldnn::primitive> pipeline;
 
--- a/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
@@ -126,7 +126,7 @@ class PoolMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
     UpdatePadding(&paddings, global_pooling, 0, padding_algorithm, data_dims,
                   strides, ksize);
 
-    platform::PoolingMKLDNNHandler::ComputeAdaptivePoolParameters(
+    platform::PoolingMKLDNNHandler<T>::ComputeAdaptivePoolParameters(
         ctx, paddle::framework::vectorize(in_x->dims()), ksize, strides);
 
     auto& dev_ctx =
--- a/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
@@ -126,11 +126,11 @@ class PoolMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
     UpdatePadding(&paddings, global_pooling, 0, padding_algorithm, data_dims,
                   strides, ksize);
 
-    PoolingMKLDNNHandler::ComputeAdaptivePoolParameters(
+    platform::PoolingMKLDNNHandler::ComputeAdaptivePoolParameters(
         ctx, paddle::framework::vectorize(in_x->dims()), ksize, strides);
 
-        auto& dev_ctx =
-            ctx.template device_context<platform::MKLDNNDeviceContext>();
+    auto& dev_ctx =
+        ctx.template device_context<platform::MKLDNNDeviceContext>();
 
     std::vector<mkldnn::primitive> pipeline;
 
--- a/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
@@ -126,6 +126,19 @@ class PoolMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
     UpdatePadding(&paddings, global_pooling, 0, padding_algorithm, data_dims,
                   strides, ksize);
 
+
+    const bool adaptive = ctx.Attr<bool>("adaptive");
+
+    const auto src_tz = paddle::framework::vectorize(in_x->dims());
+    if (adaptive) {
+      ksize[0] = static_cast<int>(
+          ceil(static_cast<double>(src_tz[src_tz.size() - 2] / ksize[0])));
+      ksize[1] = static_cast<int>(
+          ceil(static_cast<double>(src_tz[src_tz.size() - 1] / ksize[1])));
+      strides[0] = ksize[0];
+      strides[1] = ksize[1];
+    }
+
     auto& dev_ctx =
         ctx.template device_context<platform::MKLDNNDeviceContext>();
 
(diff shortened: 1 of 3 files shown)


jczaja:
@wozna, @wojtuss , @arlesniak , @arogowie-intel  Please review

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -853,6 +854,16 @@ class PoolingMKLDNNHandler : public MKLDNNHandlerT<T, mkldnn::pooling_forward,
         CorrectOutputSize(src_tz, dst_tz, ksize, paddings, strides,
                           mkldnn_paddings[1]);
       }
+
+      if (adaptive) {
+        ksize[0] = static_cast<int>(
+            ceil(static_cast<double>(src_tz[src_tz.size() - 2] / ksize[0])));
+        ksize[1] = static_cast<int>(
+            ceil(static_cast<double>(src_tz[src_tz.size() - 1] / ksize[1])));
```
arogowie-intel: I guess you want to round up the kernel size, thus you need to cast to floating point at least one arg of division operator, otherwise both sides are integers and we have integer division. If integer division is that what you want, then cast to double is not necessary.
```suggestion
        ksize[0] = static_cast<int>(
            ceil(static_cast<double>(src_tz[src_tz.size() - 2]) / ksize[0]));
        ksize[1] = static_cast<int>(
            ceil(static_cast<double>(src_tz[src_tz.size() - 1]) / ksize[1]));
```

arogowie-intel: I'd suggest to have UT which distinguish those two slightly different rounding cases.

python/paddle/fluid/tests/unittests/mkldnn/test_pool2d_mkldnn_op.py
```diff
@@ -61,6 +61,22 @@ def init_data_type(self):
...
+    def init_pool_type(self):
+        self.pool_type = "avg"
+        self.pool2D_forward_naive = avg_pool2D_forward_naive
+
+    def init_kernel_type(self):
+        self.use_mkldnn = True
+
+    def init_test_case(self):
+        self.ksize = [1, 1]
+        self.strides = [1, 1]
```
arlesniak, line 77: I'm wondering if you could put test case triggering another calculation in additon to happy path in the mkldnn_reuse when adaptive is used.

jczaja: @arlesniak I have added another test. Thanks for pointing out missing scenario coverage

paddle/fluid/platform/mkldnn_reuse.h
jczaja: @arogowie-intel After some analysis of reference solution I restricted supported cases only to those when src_tz dims are divisible by corressponding ksize dims. This is because when we have some rounding applied then we may end up that during one pooling operation (multiple poolings over same input) pooling window size may vary , which is not supported by oneDNN.

arlesniak (approved):
LGTM

jczaja:

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -853,6 +853,9 @@ class PoolingMKLDNNHandler : public MKLDNNHandlerT<T, mkldnn::pooling_forward,
         CorrectOutputSize(src_tz, dst_tz, ksize, paddings, strides,
                           mkldnn_paddings[1]);
       }
+
```
lidanqing-vv, line 856: 1. why `ctx.Attr<bool>("adaptive")` is not here but inside the function `ComputeAdaptivePoolParameters`?

2. In the function `ComputeAdaptivePoolParameters` declaration, it is not clear what will be changed, what will not. Maybe we could refer to tester_helper.cc
```
void TestOneThreadPrediction(const PaddlePredictor::Config *config, const std::vector<std::vector<PaddleTensor>> &inputs, 
 std::vector<std::vector<PaddleTensor>> *outputs, int num_threads, bool use_analysis = FLAGS_use_analysis)
```
in this way from the `ComputeAdaptivePoolParameters(ctx, src_tz, &ksize, &strides);`, we know what will be changed what will not.
But maybe point 2 is not important

paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
```diff
@@ -126,6 +126,9 @@ class PoolMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
     UpdatePadding(&paddings, global_pooling, 0, padding_algorithm, data_dims,
                   strides, ksize);
 
+    platform::PoolingMKLDNNHandler<T>::ComputeAdaptivePoolParameters(
```
lidanqing-vv, line 129: Should `UpdatePadding(&paddings, global_pooling, 0, padding_algorithm, data_dims,strides, ksize);` be changed to 
`UpdatePadding(&paddings, global_pooling, ctx.Attr<bool>("adaptive"), padding_algorithm, data_dims,strides, ksize);`, or it does not matter ?

paddle/fluid/platform/mkldnn_reuse.h
jczaja: ComputeAdaptivePoolParameters is used in two places e.g. once inside PoolMKLDNNHandler and once directly in PoolGrad. So it was easiest for me to just read it inside rather than read it in two places and then pass as argument which would introduced tiny overhead as more data is passed to call.

paddle/fluid/operators/mkldnn/pool_mkldnn_op.cc
jczaja: @lidanqing-intel I think they should stay seprate as UpdatePadding is a common function used for CPU and oneDNN kernels (perhaps GPU as well) and computeAdaptivePoolParameters is just aplicable to oneDNN kernel.

paddle/fluid/platform/mkldnn_reuse.h
lidanqing-vv: Ok, it is clear. Thanks

jczaja:

python/paddle/fluid/tests/unittests/mkldnn/test_pool2d_mkldnn_op.py
```diff
@@ -160,4 +191,6 @@ def init_shape(self):
 
 
 if __name__ == '__main__':
+    from paddle import enable_static
+    enable_static()
```
luotao1, line 195: Why use `enable_static` only in this mkldnn unit-test?

jczaja: @luotao1 It should be in every test , it just I was working on that one so I added only in this one.

luotao1: Got it.

luotao1 (approved):
LGTM

PR #27750, opened 2020-10-01
Add avx512 core instructions check 2 

wozna:
### PR types

Bug fixes
### PR changes

OPs 
### Describe

In previous PR https://github.com/PaddlePaddle/Paddle/pull/27732 I added an instruction check but it only works while running from python, but it did not work if the test was run from cmake. 
This PR adds `unittest.SkipIf` what skip the test if the place does not support bfloat16 evaluation and it works for both cases. 

It also removes duplicated `self.use_mkldnn = True`


jczaja (approved):
LGTM

luotao1 (approved):

PR #27754, opened 2020-10-02
keep network mode unchange when use summary api

LielinJiang:
### PR types

Bug fixes
### PR changes

Others
### Describe

keep network mode unchange when use summary api


wangchaochaohu (approved):
LGTM

PR #27777, opened 2020-10-09
Change type of mkldnn_data_type attribute

wozna:
### PR types

Others 
### PR changes

OPs 
### Describe

This PR change type of `mkldnn_data_type` attribute from string to int. It is more elegant and less prone to errors solution. 
Unfortunately, I couldn't find a way to split this PR with fewer files.

This PR remove temporarily `test_post_training_quantization_mnist` because this test was generated with `mkldnn_data_type` as std::string. This model should be generated after this PR merge and the test should be restored.

### DIFF
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -2114,22 +2114,22 @@ PDNode *patterns::Bfloat16Placement::operator()(
 PDNode *patterns::OrphanedBfloat16::operator()() {
   auto *prev_op = pattern->NewNode(prev_op_repr())->assert_is_op();
   prev_op->assert_more([&](Node *node) {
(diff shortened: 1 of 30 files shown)


wojtuss:
@luotao1 
This PR changes the type of the `mkldnn_data_type` attribute from `std::string` to `int`. Choosing the former type was suboptimal, as operations on strings are slower than on integers and integers suite the purpose of the attribute better. Changing the type of the attribute breaks backward compatibility, but assuming this kind of changes can be introduces by 2.0 final release we consider it worth merging (after fixing CI builds). It could also be handled by the op versioning mechanism. Do you think we can proceed with this PR further?

lidanqing-vv:
I close this PR since we discussed that a lot of models maybe be influenced and also we checked in string comparison length will be compared first, and also Baidu CIs and our CIs did not detect obvious performance drop.

PR #27443, opened 2020-09-21
Use dygraph mode by default

pangyoki:
### PR types

Breaking changes

### PR changes

Others

### Describe

Paddle 2.0 uses `dygraph mode` by default. This PR change the default mode from `static mode` to `dygraph mode`.

#### 1. use dygraph mode by default
In `paddle/__init__.py`, add `disable_static()` to change the default mode to `dygraph mode`.

#### 2. manually change unittest's mode to static mode
Unittest is realized based on static mode. It will cause conflicts when dygraph mode is opened by default.
In order to pass CI, before the unittest runs, switch the mode to static mode.

- modify `test_runner.py`
When using `ctest` to execute unittest files, `test_runner.py` is the entry point for many unittest files. `test_runner.py` will call the module in unittest files to execute.
Thus, we can add `enable_static` in `test_runner.py`. Before running unittest, `test_runner.py` will provide static mode.

- modify unittest file
Some unittests are not executed based on `test_runner.py`, so their running environment is still dygraph mode, and CI wil fail.
To solve this problem, we modify all unittests in the CI error report manually.
The following files will be added `enable_static` manually.

CI-Mac
```
2020-09-17 21:17:20 	132 - test_train_recognize_digits_mlp (Failed)
2020-09-17 21:17:20 	133 - test_train_recognize_digits_conv (Failed)
2020-09-17 21:17:20 	156 - test_beam_search_decoder (Failed)
2020-09-17 21:17:20 	157 - test_data_feeder (Failed)
2020-09-17 21:17:20 	158 - test_detection (Failed)
2020-09-17 21:17:20 	159 - test_error_clip (Failed) # add enable_static
2020-09-17 21:17:20 	160 - test_if_else_op (Failed)
2020-09-17 21:17:20 	162 - test_python_operator_overriding (Failed)
2020-09-17 21:17:20 	615 - test_nan_inf (Failed) # add enable_static
2020-09-17 21:17:20 	954 - test_fit_a_line (Failed)
2020-09-17 21:17:20 	955 - test_image_classification (Failed)
2020-09-17 21:17:20 	956 - test_label_semantic_roles (Failed)
2020-09-17 21:17:20 	957 - test_machine_translation (Failed)
2020-09-17 21:17:20 	958 - test_recognize_digits (Failed)
2020-09-17 21:17:20 	959 - test_recommender_system (Failed) # add enable_static
2020-09-17 21:17:20 	960 - test_rnn_encoder_decoder (Failed)
[...]

PR #27819, opened 2020-10-10
fix Error of gpu version paddle when CUDA device is not set properly

pangyoki:
### PR types

Others

### PR changes

Others

### Describe

When users install GPU version Paddle, CPU device can't be used if CUDA device is not set properly.
```
Cannot use GPU because there is no GPU detected on your machine.
```

- Reason
To open dygraph mode by default. `disable static()` is used in `paddle/__init__.py` .  Refer to PR https://github.com/PaddlePaddle/Paddle/pull/27443 .
When Paddle is compiled with CUDA version, `CUDAPlace(0)` will be used if default place is `None`. It will raise Error if CUDA device is not set properly.
```
if core.is_compiled_with_cuda():
     _global_expected_place_ = core.CUDAPlace(0)
```

This PR is used to solve this problem. CPU device will be used by default when there's no proper CUDA devices.


python/paddle/fluid/framework.py
```diff
@@ -284,7 +284,17 @@ def _current_expected_place():
...
         if core.is_compiled_with_cuda():
-            _global_expected_place_ = core.CUDAPlace(0)
+            try:
+                device_count = core.get_cuda_device_count()
+            except Exception as e:
+                device_count = 0
+            if device_count > 0:
+                _global_expected_place_ = core.CUDAPlace(0)
+            else:
+                logging.warning(
```
zhiqiu: Use warnings instead of logging

python/paddle/fluid/framework.py
```diff
@@ -284,7 +284,17 @@ def _current_expected_place():
...
+                    "You are using GPU version Paddle, But Your CUDA Device is not set properly. CPU device will be used by default."
```
zhiqiu: But -> but

python/paddle/fluid/framework.py
pangyoki: done

python/paddle/fluid/framework.py
pangyoki: done

python/paddle/fluid/framework.py
```diff
@@ -284,7 +284,17 @@ def _current_expected_place():
     global _global_expected_place_
     if _global_expected_place_ is None:
         if core.is_compiled_with_cuda():
-            _global_expected_place_ = core.CUDAPlace(0)
+            try:
+                device_count = core.get_cuda_device_count()
+            except Exception as e:
+                device_count = 0
```
zhiqiu, line 290: I think we need to do something if an Exception happens instead of just ignore it, we can raise warnings at least. It can be refined in the next PR.

python/paddle/fluid/tests/unittests/test_gpu_package_without_gpu_device.py
```diff
@@ -0,0 +1,57 @@
...
+            _python = sys.executable
+
+            ps_cmd = '{} {}'.format(_python, test_file)
+            ps_proc = subprocess.Popen(
+                ps_cmd.strip().split(" "),
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE)
+            stdout, stderr = ps_proc.communicate()
+
+            assert 'CPU device will be used by default' in str(
```
zhiqiu, line 48: Not important.
Since `unittest` is used,  I recommend using `unittest's` API, like `self.assertXXX` instead of python's builtin `assert`.

zhiqiu (approved):

luotao1 (approved):
LGTM for framework.py

PR #27820, opened 2020-10-10
Fix summary dtype not support int

LielinJiang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix summary dtype not support int


wangchaochaohu (approved):
LGTM

PR #27840, opened 2020-10-12
Enable GLOO when compiled with WITH_DISTRIBUTE=ON

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

Enable GLOO when compiled with WITH_DISTRIBUTE=ON
For distributed training, we will initialize GLOO by default. To avoid use -DWITH_GLOO=ON option, we set WITH_GLOO=ON by default when WITH_DISTRIBUTE=ON is set.


gavin1332 (approved):
LGTM

luotao1 (approved):
LGTM

PR #27643, opened 2020-09-27
fleet combine amp dgc recompute meta optimizer

wangxicoding:
### PR types

Others
### PR changes

Others
### Describe

Refine amp&dgc&recompute&lars&lamb&local_sgd, make them can be an inner_opt of other meta optimizer.
Now the following strategies that can be combined：
``` python
import paddle.distributed.fleet as fleet
strategy = fleet.DistributedStrategy()

# 1. amp + recompute
strategy.amp = True
strategy.recompute = True

# 2. dgc + recompute
strategy.dgc = True
strategy.recompute = True

# 3. lars + recompute
strategy.lars = True
strategy.recompute = True

# 4. lamb + recompute
strategy.lamb = True
strategy.recompute = True

# 6. amp + localsgd
strategy.amp = True
strategy.localsgd = True

# 6. amp + adaptive_localsgd
strategy.amp = True
strategy.adaptive_localsgd = True

# 7. amp + lars
strategy.amp = True
strategy.lars = True

# 8. amp + lamb
strategy.amp = True
strategy.lamb = True

# amp + dgc still have a problem, not combine them for now.
# 9. need supply
```

PR #27867, opened 2020-10-12
use the Floyd algorithm to find meta optimizer max path, test=develop

wangxicoding:
### PR types

Others

### PR changes

Others

### Describe

Use the Floyd algorithm to find the max path of meta optimizer.
For example, if we have a meta a-->b-->c, a-->c, we can get the max path a-->b-->c.
Or if we have a meta a-->b-->c, a-->d, b-->d, we can get the max path a-->b-->d.

In addition to the strategies given in the https://github.com/PaddlePaddle/Paddle/pull/27643, the following strategies can also be combined.
``` python
Now the following strategies that can be combined：

import paddle.distributed.fleet as fleet
strategy = fleet.DistributedStrategy()

# 9. amp + recompute + lars
strategy.amp = True
strategy.recompute = True
strategy.lars = True

# 10. amp + recompute + lamb
strategy.amp = True
strategy.recompute = True
strategy.lamb = True
```

### DIFF
--- a/python/paddle/distributed/fleet/base/strategy_compiler.py
+++ b/python/paddle/distributed/fleet/base/strategy_compiler.py
@@ -13,24 +13,95 @@
 # limitations under the License.
 
 
-def maximum_path_len_algo(optimizer_list):
-    max_idx = 0
-    max_len = 0
-    candidates = []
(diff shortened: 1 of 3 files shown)


guru4elephant (approved):
LGTM

PR #27874, opened 2020-10-13
Remove scale loss and coll grads in doc

chenwhql:
### PR types

Others
### PR changes

Docs
### Describe

Remove scale loss and coll grads in doc


saxon-zh (approved):
LGTM

PR #27888, opened 2020-10-13
Lookup table v2 xpu

yinhaofeng:
### PR types

Function optimization
### PR changes

OPs
### Describe

add lookup table v2 xpu op and test


zhangting2020 (approved):
LGTM for skip_check_grad_ci

PR #27891, opened 2020-10-13
Fix the accuracy problem of allclose op when using float64 data type

huangxu96:
### PR types

Bug fixes
### PR changes

OPs
### Describe

This PR fixed a bug in allclose_op, which cannot get the expected output when fp64 as input in some cases.

**BUG reproduce process:**

```python
paddle.disable_static()

np_x = np.array([10.1]).astype("float64")

np_y = np.array([10]).astype("float64")

x = paddle.to_tensor (np_x)

y = paddle.to_tensor (np_y)

result = paddle.allclose(x=x, y=y, rtol=0.01, atol=0, equal_nan=False, name="ignore_nan")

result = result.numpy()

print(result)
```

This result is expected to be **True** but it returns **false**.

**Problem reason**

Floating point number cannot be determined equal directly, since floating point number cannot have a precise experssion in
Computer. For example, 0.1 might be 0.09999 or 0.100001 in computer. So when we want to determine  two "0.1" if they are equaled, the compute might executes if 0.09999 equal to 0.100001. This is how the false result comes.

**Solving approach:**

Add an extremely small value (1e-15) when determine if two double varibles are equaled.

Change the date type of rtol and atol from float32 to float64, since the accuracy of rtol and atol will also impact the final result.


paddle/fluid/operators/allclose_op.cu
```diff
@@ -12,12 +12,83 @@
...
 
+namespace paddle {
+namespace operators {
+
+template <typename T>
+struct GetTensorValue<platform::CUDADeviceContext, T> {
+  T operator()(const framework::Tensor& tensor) const {
+    const T* data = tensor.data<T>();
+    T value;
+    cudaMemcpy(&value, data, sizeof(T), cudaMemcpyDeviceToHost);
```
wzzju: Prefer to use `memory::Copy(platform::CPUPlace(), &value, gpu_place, data, sizeof(T), dev_ctx.stream());`

paddle/fluid/operators/allclose_op.cc
```diff
@@ -13,25 +13,67 @@
...
+};
+
+template <typename T>
+struct AllcloseFunctor<platform::CPUDeviceContext, T> {
+  void operator()(const platform::CPUDeviceContext& ctx,
+                  const framework::Tensor& in, const framework::Tensor& other,
+                  const double rtol, const double atol, bool equal_nan,
+                  framework::Tensor* output) {
+    auto* in_a = in.data<T>();
+    auto* in_b = other.data<T>();
```
wzzju, line 39: Use better var name instead.

paddle/fluid/operators/allclose_op.cc
```diff
@@ -13,25 +13,67 @@
...
+    *out_data = true;
+    for (int i = 0; i < in_dims; i++) {
+      const T a = in_a[i], b = in_b[i];
+      bool val;
+      if (std::isnan(a) || std::isnan(b)) {
+        val = equal_nan && std::isnan(a) == std::isnan(b);
+      } else {
+        T left = (a > b ? a - b : b - a);
+        T right = atol + (b > 0 ? rtol * b : (-rtol) * b);
+        T dif = (left > right ? left - right : right - left);
```
wzzju: Use `diff` instead.

paddle/fluid/operators/allclose_op.cc
```diff
@@ -13,25 +13,67 @@
...
+        val = a == b || left <= right || dif <= 1e-15;
+      }
+      *out_data &= val;
+    }
+  }
+};
+
+template struct AllcloseFunctor<platform::CPUDeviceContext, double>;
```
wzzju: Not needed here.

paddle/fluid/operators/allclose_op.cu
```diff
@@ -12,12 +12,83 @@
...
+  if (tid == 0) {
+    val_ = true;
+  }
+  __syncthreads();
+  if (isnan(a) || isnan(b)) {
+    val = equal_nan && isnan(a) == isnan(b);
+  } else {
+    T left = (a > b ? a - b : b - a);
+    T right = atol + (b > 0 ? rtol * b : (-rtol) * b);
+    T dif = (left > right ? left - right : right - left);
```
wzzju: Use `diff` instead.

paddle/fluid/operators/allclose_op.cu
```diff
@@ -12,12 +12,83 @@
...
+    val = a == b || left <= right || dif <= 1e-15;
+  }
+  atomicAnd(reinterpret_cast<int*>(&val_), static_cast<int>(val));
+  __syncthreads();
+  if (tid == 0) {
+    *out_data = static_cast<bool>(val_);
```
wzzju: Here `static_cast` is not needed.

paddle/fluid/operators/allclose_op.cu
```diff
@@ -12,12 +12,83 @@
...
+  T operator()(const framework::Tensor& tensor) const {
+    const T* data = tensor.data<T>();
+    T value;
+    cudaMemcpy(&value, data, sizeof(T), cudaMemcpyDeviceToHost);
+    return value;
+  }
+};
+
+template <typename T>
+__global__ void AllcloseCUDAKernel(const T* in_a, const T* in_b,
```
wzzju: The performance here is too poor.

paddle/fluid/operators/allclose_op.cu
```diff
@@ -12,12 +12,83 @@
...
+                      platform::errors::InvalidArgument(
+                          "Dims of input(a) and dims of other(b) should"
+                          "be equal, but received the dims of input is : %d ,"
+                          "received the dims of other is :%d. ",
+                          in_dims, other_dims));
+    const T* in_data = in.data<T>();
+    const T* other_data = other.data<T>();
+    bool* out_data = output->mutable_data<bool>(ctx.GetPlace());
+    int grid = 1;
+    int block = in_dims;
```
wzzju: Think over and try to instead these ridiculous codes here.

python/paddle/fluid/tests/unittests/test_allclose_op.py
```diff
@@ -22,19 +22,20 @@ class TestAllcloseOp(OpTest):
     def set_args(self):
```
wzzju, line 22: Add the test case below.
image

wzzju (changes requested):

paddle/fluid/operators/allclose_op.cc
huangxu96: Done.

paddle/fluid/operators/allclose_op.cu
huangxu96: Done.

paddle/fluid/operators/allclose_op.cu
huangxu96: Already used parallel reduction here.

paddle/fluid/operators/allclose_op.cu
huangxu96: Already used parallel reduction here.

paddle/fluid/operators/allclose_op.cu
huangxu96: Done.

paddle/fluid/operators/allclose_op.cc
huangxu96: Done.

python/paddle/fluid/tests/unittests/test_allclose_op.py
huangxu96: Done.

paddle/fluid/operators/allclose_op.cu
huangxu96: Done.

Superjomn (dismissed):
考虑到 allclose 在推理里几乎用不到

这次升级这里暂不考虑兼容

paddle/fluid/operators/allclose_op.cc
huangxu96: Done

python/paddle/fluid/tests/unittests/test_allclose_op.py
```diff
@@ -130,5 +131,24 @@ def test_equal_nan():
         self.assertRaises(TypeError, test_equal_nan)
 
 
+class TestAllcloseOpFloat64(TestAllcloseOp):
+    def set_args(self):
+        self.input = np.array([10.1]).astype("float64")
+        self.other = np.array([10]).astype("float64")
+        self.rtol = np.array([0.01]).astype("float64")
+        self.atol = np.array([0]).astype("float64")
+        self.equal_nan = False
```
wzzju, line 149: Add the same unit test for `float32`.

paddle/fluid/operators/allclose_op.cu
```diff
@@ -12,12 +12,72 @@
...
+    int block = 1024;
+    int grid = (block - 1 + num) / block;
+    grid = (grid > block) ? block : grid;
+    cudaMemset(out_data, true, sizeof(bool));
+    AllcloseCUDAKernel<T><<<grid, block, 0, dev_ctx.stream()>>>(
+        in_data, other_data, rtol, atol, equal_nan, num, out_data);
+  }
+};
+
+template struct AllcloseFunctor<platform::CUDADeviceContext, double>;
```
wzzju: Remove this line.

huangxu96: Done.

python/paddle/fluid/tests/unittests/test_allclose_op.py
huangxu96: Done.

wzzju (approved):
LGTM.

Superjomn (approved):
LGTM

PR #27901, opened 2020-10-13
Support inplace backward

liym27:
### PR types

New features
### PR changes

Others
### Describe

The next PR of https://github.com/PaddlePaddle/Paddle/pull/27471
Deal with the backward to support assignment to a Variable in dynamic mode.

---

#### 1. Add a new attribute `version` for `VarBase`
`version` is used to check whether an inplace operation will result in an incorrect gradient calculation.
 The version number is incremented whenever the current VarBase is modified through an inplace operation.

```Python
import paddle
var = paddle.ones(shape=[4, 2, 3], dtype="float32")
print(var.inplace_version)  # 0

var[1] = 2.2
print(var.inplace_version)  # 1
```
#### 2. Raise error if an inplace operation can result in incorrect gradient computation.

- What kind in-place operations will raise error in backward?
    1. The gradient computation uses the value of a Tensor and this Tensor is modified inplace **after** using it
For example:
```Python
import paddle
var_a = paddle.ones(shape=[4, 2, 3], dtype="float32")
var_a.stop_gradient = False

var_b = var_a**2

# Here, the gradient computation will use the value of var_b
var_c = var_b**2
var_b[1:2] = 3.3  # var_b is modified inplace after using it

loss = paddle.nn.functional.relu(var_c)
loss.backward()
```
```
----------------------
Error Message Summary:
----------------------
PermissionDeniedError: Tensor 'eager_tmp_2' used in gradient computation in grad op 'elementwise_pow_grad' has been modified by an inplace operation. Its version is 1 but the expected version is 0. Please fix your code to void calling an inplace operator after using the Tensor.
  [Hint: Expected variable_version == wrapper_version, but received variable_version:1 != wrapper_version:0.] (at /paddle/Paddle_0722/Paddle/paddle/fluid/imperative/basic_engine.cc:229)
```

- What kind in-place operations will  **NOT** raise error in backward?
    1. The gradient computation **does not** use the value of a Tensor through this Tensor is modified inplace
    2. The gradient computation uses the value of a Tensor but this Tensor is modified inplace **before** using it 

Example for i
```Python
var_a = paddle.ones(shape=[4, 2, 3], dtype="float32")
var_a.stop_gradient = False

var_b = var_a**2
var_b[1:2] = 3  # var_b is modified inplace before using it
var_c = var_b + var_b  # Here, the grad op of sum doesn't use the value of var_b
loss = var_c.sum()

var_b[1:2] = 3  # var_b is modified inplace after using it

loss.backward()
```

Example for ii
```Python
var_a = paddle.ones(shape=[4, 2, 3], dtype="float32")
var_a.stop_gradient = False

var_b = var_a**2
var_b[1:2] = 3  # var_b is modified inplace before using it

# Here, the gradient computation will use the value of var_b
var_c = var_b**2
loss = var_c.sum()
loss.backward()
```

#### 4. For api `assign`, call  `_bump_inplace_version()` when it's an inplace operation inn dynamic mode

--- 

#### Speed comparison

- seq2seq :
 batch num: 20*100
 batch size: 32

- resnet:
 batch num: 10*100
 batch size: 2

- ptb：
 batch num: 200*100
 batch size:100

| Modle | without inplace (s/batch) | with inplace (s/batch)||
|--- |---|---|---|
|cpu seq2se| 0.0650103 | **0.0661232** | - 1.712%
|cpu resnet| 2.53313 | **2.54028** |-0.282%
|cpu ptb|0.0550649|**0.0555787**|-0.925%
|gpu ptb|0.018393|**0.0187718**|-2.0179%

### DIFF
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -48,23 +48,30 @@ class LoDTensor;
 
  TensorInplaceVersion is a version counter and every Tensor has a version
  counter. It's used to check whether an inplace operation will result in an
- incorrect gradient calculation. Version is incemented when the data of the
+ incorrect gradient calculation. Version is incremented when the data of the
  Variable is modified in place.
 
  - Question: In what scenarios will version counters be shared?
- - Answer: When two Variables/VarBases share the same C++ Tensor(may be not the
- same Allocation), both of them share the same version counter. For examples:
+ - Answer: When two Variables/VarBases share the same C++ Tensor(its Allocation
+ may change), both of them share the same version counter. For examples:
   1. `z = paddle.assign(input=x, output=y)`, `z` shares the same version counter
     of `y` because z and y is the same VarBase;
   2. `y = x.detach()`, `y` shares the same version counter of `x`.
 
  - Question: In what scenarios will version counters NOT be shared?
- - Answer: Replacing a `Variable`'s data by calling Tensor::ShareDataWith(...)`.
-    Because `Tensor::ShareDataWith(...)` only shares Allocation but not
-    framework::Tensor.
+ - Answer: Replacing a `Variable`'s data by calling `Tensor::ShareDataWith(...)`
+ or `Tensor::ShareBufferWith(...)`. Because they share the same Allocation but
+ not framework::Tensor.
 
  - Question: Why put the inplace_version_counter_ in framework::Tensor instead
  of Allocation or Variable?
+ - Answer:
+  1. Tensor can call ResetHolder() to reset the corresponding Allocation so that
+  the inplace_version_counter_ changes if it's in Allocation, which will lead to
+  confusing information about inplace version.
+  2. If inplace_version_counter_ is in Variable, different VariableWrappers
+  should be able to share the same Variable. However, a VariableWrapper hold a
+  Variable object but not a pointer.
 */
 
 class TensorInplaceVersion {
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -43,6 +43,42 @@ namespace framework {
 
 class LoDTensor;
 
+/*
+ NOTE(liym27): [ What is TensorInplaceVersion used for? ]
+
+ TensorInplaceVersion is a version counter and every Tensor has a version
+ counter. It's used to check whether an inplace operation will result in an
+ incorrect gradient calculation. Version is incemented when the data of the
+ Variable is modified in place.
+
+ - Question: In what scenarios will version counters be shared?
+ - Answer: When two Variables/VarBases share the same C++ Tensor(may be not the
+ same Allocation), both of them share the same version counter. For examples:
+  1. `z = paddle.assign(input=x, output=y)`, `z` shares the same version counter
+    of `y` because z and y is the same VarBase;
+  2. `y = x.detach()`, `y` shares the same version counter of `x`.
+
+ - Question: In what scenarios will version counters NOT be shared?
+ - Answer: Replacing a `Variable`'s data by calling Tensor::ShareDataWith(...)`.
+    Because `Tensor::ShareDataWith(...)` only shares Allocation but not
+    framework::Tensor.
+
+ - Question: Why put the inplace_version_counter_ in framework::Tensor instead
+ of Allocation or Variable?
+*/
+
+class TensorInplaceVersion {
+ public:
+  explicit TensorInplaceVersion(uint32_t inplace_version = 0)
+      : inplace_version_(inplace_version) {}
+  bool IsUnique() const { return inplace_version_ == 0; }
+  void Bump() { ++inplace_version_; }
+  uint32_t CurrentVersion() const { return inplace_version_; }
+
+ private:
+  uint32_t inplace_version_;
+};
+
 class Tensor {
 #ifdef PADDLE_WITH_MKLDNN
 
@@ -189,6 +225,9 @@ class Tensor {
 
   void ResetHolderWithType(std::shared_ptr<memory::Allocation> holder,
                            const proto::VarType::Type type);
+  TensorInplaceVersion& InplaceVersionCounter() {
+    return inplace_version_counter_;
+  }
 
  private:
   /*! holds the memory block if allocated. */
@@ -225,6 +264,7 @@ class Tensor {
    *          PlaceHolder::ptr_ and where the tensor data really begins.
    */
   size_t offset_;
+  TensorInplaceVersion inplace_version_counter_;
 };
 
 }  // namespace framework
--- a/paddle/fluid/framework/variable.h
+++ b/paddle/fluid/framework/variable.h
@@ -115,13 +115,12 @@ class Variable {
 
 inline framework::TensorInplaceVersion* Variable::InplaceVersionCounter() {
   framework::TensorInplaceVersion* version_counter_ptr(nullptr);
-  if (IsType<framework::Tensor>()) {
-    version_counter_ptr =
-        &GetMutable<framework::Tensor>()->InplaceVersionCounter();
-
-  } else if (IsType<framework::LoDTensor>()) {
+  if (IsType<framework::LoDTensor>()) {
     version_counter_ptr =
         &GetMutable<framework::LoDTensor>()->InplaceVersionCounter();
+  } else if (IsType<framework::Tensor>()) {
+    version_counter_ptr =
+        &GetMutable<framework::Tensor>()->InplaceVersionCounter();
 
   } else if (IsType<framework::SelectedRows>()) {
     version_counter_ptr = &GetMutable<framework::SelectedRows>()
--- a/paddle/fluid/framework/variable.h
+++ b/paddle/fluid/framework/variable.h
@@ -23,6 +23,25 @@
 namespace paddle {
 namespace framework {
 
+// NOTE(liym27): [ What is VariableInplaceVersion used for? ]
+//
+// VariableInplaceVersion is a version counter and every Variable has a version
+// counter.
+// It's used to check whether an inplace operation will result in an incorrect
+// gradient calculation.
+// Version is incemented when the data of the Variable is modified in place.
+class VariableInplaceVersion {
+ public:
+  explicit VariableInplaceVersion(uint32_t inplace_version = 0)
+      : inplace_version_(inplace_version) {}
+  bool IsUnique() const { return inplace_version_ == 0; }
+  void Bump() { ++inplace_version_; }
+  uint32_t CurrentVersion() const { return inplace_version_; }
+
+ private:
+  uint32_t inplace_version_;
+};
+
 class Variable {
  public:
   template <typename T>
@@ -69,6 +88,10 @@ class Variable {
     return holder_->Type();
   }
 
+  VariableInplaceVersion& InplaceVersionCounter() {
+    return inplace_version_counter_;
+  }
+
  private:
   struct Placeholder {
     virtual ~Placeholder() PADDLE_MAY_THROW {}
@@ -102,6 +125,7 @@ class Variable {
 
   // pointers to a PlaceholderImpl object indeed.
   std::unique_ptr<Placeholder> holder_;
+  VariableInplaceVersion inplace_version_counter_;
 };
 
 }  // namespace framework
--- a/paddle/fluid/framework/variable.h
+++ b/paddle/fluid/framework/variable.h
@@ -72,7 +72,7 @@ class Variable {
  private:
   // This method hides type T, so it doesn't appear as a template parameter of
   // Variable.
-  framework::TensorInplaceVersion& InplaceVersionCounter();
+  framework::TensorInplaceVersion* InplaceVersionCounter();
 
  public:
   uint32_t CurrentInplaceVersion();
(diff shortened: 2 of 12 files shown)


python/paddle/fluid/dygraph/varbase_patch_methods.py
```diff
@@ -226,6 +226,27 @@ def grad(self):
 
         return self.gradient()
 
+    @property
+    def version(self):
```
zhhsplendid: Version is a very common name, and protobuf may have version. Suggest to change "version" to something like "inplace_version" to make it just for inplace, or "modify_version" if this version can be used to mark not only for the inplace operation

zhhsplendid: Also some corresponding names

paddle/fluid/framework/variable.h
```diff
@@ -23,6 +23,24 @@
...
 namespace framework {
 
+// NOTE(liym27): [ What is VariableVersion used for? ]
+//
+// VariableVersion is a version counter and every Variable has a version
+// counter.
+// It's used to check whether an inplace operation will result in an incorrect
+// gradient calculation.
+// Version is incemented when the data of the Variable is modified in place.
+class VariableVersion {
```
zhhsplendid: I saw you just use uint32_t in VariableWrapper but VariableVersion in Variable, can we make them same? Just uint32_t for Variable?

python/paddle/fluid/dygraph/varbase_patch_methods.py
zhhsplendid: version is a common name so it is not good here. For example, protobuf has versions. Can we change it to "inplace_version" for inplace operation or "modify_version" if you make it extendable for future modify

paddle/fluid/pybind/imperative.cc
```diff
@@ -774,8 +800,10 @@ void BindImperative(py::module *m_ptr) {
                  imperative::AllReduce(self.Var(), self.MutableVar(), strategy);
                } else {
                  PADDLE_THROW(platform::errors::Unimplemented(
-                     "Imperative SelectedRows allreduce is not supported when "
-                     "paddle is compiled with NCCL verison lower than v2.2.12. "
+                     "Imperative SelectedRows allreduce is not supported "
+                     "when "
```
zhhsplendid: Why break one word? Can you make the line pretty?

python/paddle/fluid/dygraph/varbase_patch_methods.py
zhhsplendid: Also change some corresponding names

paddle/fluid/imperative/dygraph_grad_maker.h
```diff
@@ -226,6 +225,9 @@ class TracedGradOp {
     }
 
     auto var_wrappers = ToVarWrapperList<kRole>(vars);
+
+    var_wrappers = CopyVarWrapperList(var_wrappers);
```
zhhsplendid: Why do we need to snapshot here? By the way, I say you are doing pointer operation in CopyVarWrapperList, so it is not "Copy" VarWrapperList

python/paddle/fluid/tests/unittests/test_inplace.py
```diff
@@ -0,0 +1,94 @@
...
+import unittest
+import numpy as np
+
+import paddle
+import paddle.fluid.core as core
+
+
+class TestInplace(unittest.TestCase):
+    def test_forward_version(self):
+        with paddle.fluid.dygraph.guard():
```
zhhsplendid, line 26: Should we delete these `with paddle.fluid.dygraph.guard()` since now dygraph is enabled by default

python/paddle/fluid/dygraph/varbase_patch_methods.py
liym27: Thanks. Change it to `inplace_version` and just for inplace now

paddle/fluid/imperative/dygraph_grad_maker.h
liym27: > Why do we need to snapshot here? 

Because the current `inplace_version` of `Variable` should be saved here, and it will be used to check whether there is any inplace operation affecting gradient calculation.

> By the way, I say you are doing pointer operation in CopyVarWrapperList, so it is not "Copy" VarWrapperList

Thanks, "CopyVarWrapperList" will be changed to "SnapshotVarWrapperList"

paddle/fluid/pybind/imperative.cc
liym27: Sorry, fixed it.

zhouwei25:
如果VarBase A与VarBase B共享了一块内存allocation，他们各自持有1个Variable，VarBase B inplace修改了allocation的数据，仅VarBase B->Variable的version发生了变化，VarBase A->Variable的version没有变化，但由于共享allocation，其实VarBase A也被动进行了inplace操作，但是不会检测到。是不是应该将version与最底层的allocation绑定 会更安全一些呀

paddle/fluid/framework/variable.h
```diff
@@ -102,6 +126,7 @@ class Variable {
 
   // pointers to a PlaceholderImpl object indeed.
   std::unique_ptr<Placeholder> holder_;
+  VariableVersion version_counter_;
```
zhwesky2010: 如果VarBase A与VarBase B共享了一块内存allocation，他们各自持有1个Variable，VarBase B inplace修改了allocation的数据，仅VarBase B->Variable的version发生了变化，VarBase A->Variable的version没有变化，但由于共享allocation，其实VarBase A也被动进行了inplace操作，但是不会检测到。是不是应该将version与最底层的allocation绑定 会更安全一些呀

paddle/fluid/framework/variable.h
liym27: As the discussion offline, consider using `VariableVersion` in the future and improving scalability, `VariableVersion` in `Variable` will be kept

paddle/fluid/framework/variable.h
liym27: > VarBase A与VarBase B共享了一块内存allocation

调用哪个API会发生内存功能共享的情况呢？内存共享是用户层面清楚的吗，还是只是框架在实现上为了性能而进行内存共享~

> 是不是应该将version与最底层的allocation绑定 会更安全一些呀

是的，基于给出的这个例子，确实如此. 线下讨论了解，detach 功能的实现 不是只能共享allocation，需进一步讨论。

python/paddle/fluid/tests/unittests/test_inplace.py
liym27: In `TestInplace`, different test_XX functions interact with each other and have the same `Tracer`, which will cause  the tests to fail. So call  `with paddle.fluid.dygraph.guard()` to use different `Tracer` in different test.

paddle/fluid/pybind/imperative.cc
```diff
@@ -628,6 +632,28 @@ void BindImperative(py::module *m_ptr) {
...
+             const auto &var = self.MutableVar();
+             PADDLE_ENFORCE_EQ(
+                 var->IsInitialized(), true,
+                 platform::errors::InvalidArgument(
+                     "Tensor of %s is Empty, please check if it has no data.",
+                     self.Name()));
+
+             return var->InplaceVersionCounter().CurrentVersion();
+           })
+      .def("bump_inplace_version",
```
chenwhql: Does this API need to be public, maybe `_bump_implace_version` is better?

paddle/fluid/imperative/variable_wrapper.h
```diff
@@ -168,6 +179,8 @@ class VariableWrapper {
   int overrided_stop_gradient_{-1};
   bool persistable_{false};
 
+  uint32_t wrapper_inplace_version_ = 0;
```
chenwhql: At first I didn’t understand why inplace_version in Variable is a class and here is a uint. Later, I understand that this is a snapshot and will not be changed. If the naming here reflects this attribute, it may be easier to understand,such as `inplace_version_snapshot_`?

paddle/fluid/pybind/imperative.cc
liym27: > Does this API need to be public ?

No.
>  maybe _bump_implace_version is better?

Thanks.  Done

paddle/fluid/imperative/variable_wrapper.h
liym27: > At first I didn’t understand why inplace_version in Variable is a class and here is a uint.

As we discussed offline, it’s better that `wrapper_inplace_version_` is a unit.

>  Later, I understand that this is a snapshot and will not be changed. If the naming here reflects this attribute, it may be easier to understand,such as inplace_version_snapshot_?

Thanks, `inplace_version_snapshot_ ` is better.

paddle/fluid/framework/tensor.h
```diff
@@ -43,6 +43,42 @@ namespace framework {
 
 class LoDTensor;
 
+/*
+ NOTE(liym27): [ What is TensorInplaceVersion used for? ]
+
+ TensorInplaceVersion is a version counter and every Tensor has a version
+ counter. It's used to check whether an inplace operation will result in an
+ incorrect gradient calculation. Version is incemented when the data of the
```
zhhsplendid: Typo: incemented -> incremented

paddle/fluid/imperative/dygraph_grad_maker.h
```diff
@@ -304,6 +306,33 @@ class TracedGradOp {
...
+  // in gradient compution.
+  static std::vector<std::shared_ptr<VariableWrapper>> SnapshotVarWrapperList(
+      const std::vector<std::shared_ptr<VariableWrapper>>& var_wrappers) {
+    std::vector<std::shared_ptr<VariableWrapper>> result;
+    result.reserve(var_wrappers.size());
+
+    for (auto& var : var_wrappers) {
+      // NOTE(liym27):
+      //  Emplace back original var_wrapper if its inplace_version is not
+      //  changed.
```
zhhsplendid: Format this comment

paddle/fluid/framework/tensor.h
liym27: Thanks, done

paddle/fluid/imperative/dygraph_grad_maker.h
liym27: Thanks, done

zhhsplendid (dismissed):
LGTM

paddle/fluid/framework/variable.h
```diff
@@ -101,8 +110,48 @@ class Variable {
   };
 
   // pointers to a PlaceholderImpl object indeed.
-  std::unique_ptr<Placeholder> holder_;
+  std::shared_ptr<Placeholder> holder_;
 };
 
+inline framework::TensorInplaceVersion& Variable::InplaceVersionCounter() {
+  if (IsType<framework::Tensor>()) {
+    return GetMutable<framework::Tensor>()->InplaceVersionCounter();
```
zhiqiu: Better `GetMutable` -> `Get`, since this function does not change the Variable.

paddle/fluid/framework/variable.h
```diff
@@ -101,8 +110,48 @@ class Variable {
...
+  } else if (IsType<framework::LoDTensor>()) {
+    return GetMutable<framework::LoDTensor>()->InplaceVersionCounter();
+  } else if (IsType<framework::SelectedRows>()) {
+    return GetMutable<framework::SelectedRows>()
+        ->mutable_value()
+        ->InplaceVersionCounter();
+  } else if (IsType<framework::LoDTensorArray>()) {
+    auto tensor_array = GetMutable<framework::LoDTensorArray>();
+    return tensor_array->at(0).InplaceVersionCounter();
```
zhiqiu: Do we need to check the consistency of versions of each tensor in LoDTensorArray?

paddle/fluid/framework/variable.h
```diff
@@ -101,8 +110,48 @@ class Variable {
...
+  } else {
+    // NOTE(liym27): Other types of data are not supported, like Scope.
+    PADDLE_THROW(platform::errors::Unimplemented(
+        "Only supports Tensor, LoDTensor, SelectedRows and "
+        "LoDTensorArray to have TensorInplaceVersion."));
```
zhiqiu: Can we give the current type in error message?

paddle/fluid/framework/variable.h
```diff
@@ -101,8 +110,48 @@ class Variable {
...
+  }
+}
+inline uint32_t Variable::CurrentInplaceVersion() {
+  try {
+    return InplaceVersionCounter().CurrentVersion();
+  } catch (platform::EnforceNotMet& exception) {
```
zhiqiu: Why not raise the exception and stop the execution?

paddle/fluid/framework/variable.h
```diff
@@ -101,8 +110,48 @@ class Variable {
...
+               "LoDTensorArray to have TensorInplaceVersion.";
+    return 0;
+  }
+}
+
+inline void Variable::BumpInplaceVersion() {
+  try {
+    InplaceVersionCounter().Bump();
+  } catch (platform::EnforceNotMet& exception) {
+    VLOG(4) << "Only supports Tensor, LoDTensor, SelectedRows and "
```
zhiqiu: Same above.

paddle/fluid/imperative/basic_engine.cc
```diff
@@ -214,6 +214,31 @@ void BasicEngine::Execute() {
         }
       }
 
+      VLOG(4) << "Check whether there is any inplace operation affecting "
+                 "gradient calculation.";
+      for (auto& pair : bwd_ins) {
+        for (auto& var_wrapper : pair.second) {
+          auto wrapper_version_snapshot = var_wrapper->InplaceVersionSnapshot();
+          auto tensor_version =
+              var_wrapper->MutableVar()->CurrentInplaceVersion();
```
zhiqiu, line 234: `MutableVar` -> `Var`

python/paddle/fluid/layers/tensor.py
```diff
@@ -600,6 +605,9 @@ def assign(input, output=None):
                 value_name: values
             })
 
+    if is_inplace and in_dygraph_mode():
+        output._bump_inplace_version()
+
```
zhiqiu, line 610: Why need to change the version in the python side?

paddle/fluid/imperative/layer.cc
```diff
@@ -394,5 +394,14 @@ std::shared_ptr<GradOpNode> CreateGradOpNode(
   }
 }
 
+void BumpInplaceVersion(std::shared_ptr<imperative::VarBase> var) {
```
zhiqiu: Can we make it as a member function of VarBase?

paddle/fluid/imperative/basic_engine.cc
liym27: It's not supported to call `Var()` here because that `Var()` has type `const framework::Variable` but `CurrentInplaceVersion()` is not const and `CurrentInplaceVersion()` can't be marked const.

python/paddle/fluid/layers/tensor.py
liym27: For Python api `assign(input, output=None)`,  it can only be checked on the python side that whether `output` is changed inplace. And when changed inplace,  the inplace  version of `output` should be increased by 1.

paddle/fluid/framework/variable.h
liym27: Thanks. Actually, in dynamic mode, LoDTensorArray is not used in most cases.  There is no need to support `LoDTensorArray ` here,  I will delete it.

paddle/fluid/framework/variable.h
liym27: Yes, thanks, done

paddle/fluid/framework/variable.h
liym27: Because the exception is raised in InplaceVersionCounter() when the type is not LoDTensor or SelectedRows, which is within expectations. And only returns 0 as its inplace version.

paddle/fluid/framework/variable.h
liym27: Because the exception is raised in InplaceVersionCounter() when the type is not LoDTensor or SelectedRows, which is within expectations. And only do nothing.

paddle/fluid/framework/variable.h
liym27: Calling this function can change the `Tensor`'s inplace_version of this `Variable`, so use `GetMutable` here.

paddle/fluid/imperative/layer.cc
liym27: Yes, done

zhhsplendid (approved):
LGTM

luotao1 (approved):
LGTM for tensor.h! 注释写的太太太棒了！

paddle/fluid/framework/variable.h
```diff
@@ -101,8 +110,49 @@ class Variable {
...
+  std::shared_ptr<Placeholder> holder_;
 };
 
+inline framework::TensorInplaceVersion* Variable::InplaceVersionCounter() {
+  framework::TensorInplaceVersion* version_counter_ptr(nullptr);
+  if (IsType<framework::Tensor>()) {
+    version_counter_ptr =
+        &GetMutable<framework::Tensor>()->InplaceVersionCounter();
+
+  } else if (IsType<framework::LoDTensor>()) {
```
phlrain: Tensor 和 LodTensor是不是能公用同一个分支

liym27: 不能，会报错。

zhhsplendid (approved):
LGTM

luotao1 (approved):
LGTM

phlrain (approved):

PR #27950, opened 2020-10-14
Fix dataloader when stack input data with different type

LielinJiang:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix dataloader when stack input data with different type


heavengate (dismissed):
LGTM+++

heavengate (approved):

PR #27954, opened 2020-10-14
Fix problem with flags fp32 and int8

wozna:
### PR types

Bug fixes
### PR changes

APIs 
### Describe

This PR fixes problem with different uses of flags `--enable_int8` , `--enable_fp32` and `--enable_bf16`.
Bfloat16 test for image classification models uses the same functions as int8 and quant tests and the default value for  `--enable_int8` and `--enable_fp32` is true that is why after adding `--enable_bf16` there were some conflicts. 
This PR changes the default value for  `enable_int8` and `--enable_bf16` to `false` and adds setting these flags to every test in CMake files.


jczaja (approved):
LGTM

luotao1 (approved):
LGTM

PR #27975, opened 2020-10-15
[Dy2Stat] Fix Error when generating train_program in eval mode

Aurelius84:
### PR types

Function optimization
### PR changes

Others
### Describe

Fix Error when generating train_program in eval mode. See following example code:

```python
class Net(nn.Layer):
    def __init__(self, in_channels, hidden_size):
        super(Net, self).__init__()
        self.lstm = nn.LSTM(
            in_channels, hidden_size, direction='bidirectional', num_layers=2)

    @paddle.jit.to_static
    def forward(self, x):
        x, _ = self.lstm(x)
        return x

net = Net(12, 2)
net.eval() # switch into eval mode firstly
# Here, we will get infer_program defaultly after converting dygraph layers into static.
# When initializing ParitalLayer, a train_program will be generated by `append_backward`
# from converted program(here is infer_program) with `is_test=True`, which will raise a error.
net = paddle.jit.to_static(
    net, input_spec=[paddle.static.InputSpec(shape=[-1, 10, 12])])
paddle.jit.save(net, 'simple_lstm')
```

**Solution:**
1. Add `change_is_test_status` to make sure all is_test=True  before `append_backward`
2. Introduce lazy initialization mechanism to create program while only we need it.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/partial_program.py
@@ -138,14 +138,14 @@ class PartialProgramLayer(layers.Layer):
         self.training = True
 
     @LazyInitialized
-    def infer_program(self):
+    def _infer_program(self):
         """
         Lazy initialized property of infer_program.
         """
         return self._clone_for_test(self._origin_main_program)
 
     @LazyInitialized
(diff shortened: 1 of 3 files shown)


zhhsplendid (approved):
LGTM

PR #28021, opened 2020-10-16
Upgrade shellcheck version

gongweibao:
### PR types
 New features

### PR changes
 Others

### Describe
Upgrade shellcheck version to v0.7.1 from 0.4.1 so users can get the warning wiki URL when they occured.

For example:
If we shellcheck this script,
```
#!/bin/sh
## Example: a typical script with several problems
for f in $(ls *.m3u)
do
  grep -qi hq.*mp3 $f \
    && echo -e 'Playlist $f contains a HQ file in mp3 format'
done
```

shellcheck will generate these:
```
In test.sh line 3:
for f in $(ls *.m3u)
         ^---------^ SC2045: Iterating over ls output is fragile. Use globs.
              ^-- SC2035: Use ./*glob* or -- *glob* so names with dashes won't become options.


In test.sh line 5:
  grep -qi hq.*mp3 $f \
           ^-----^ SC2062: Quote the grep pattern so the shell won't interpret it.
                   ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
  grep -qi hq.*mp3 "$f" \


In test.sh line 6:
    && echo -e 'Playlist $f contains a HQ file in mp3 format'
            ^-- SC2039: In POSIX sh, echo flags are undefined.
               ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.

For more information:
  https://www.shellcheck.net/wiki/SC2045 -- Iterating over ls output is fragi...
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, echo flags are undef...
  https://www.shellcheck.net/wiki/SC2062 -- Quote the grep pattern so the she...
```

### DIFF
--- a/tools/dockerfile/Dockerfile.ubuntu
+++ b/tools/dockerfile/Dockerfile.ubuntu
@@ -29,8 +29,12 @@ RUN apt-get update && \
     python-matplotlib \
     automake locales clang-format swig  \
     liblapack-dev liblapacke-dev \
-    net-tools libtool module-init-tools shellcheck && \
+    net-tools libtool module-init-tools && \
     apt-get clean -y
+
+RUN wget https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.linux.x86_64.tar.xz -O shellcheck-v0.7.1.linux.x86_64.tar.xz && \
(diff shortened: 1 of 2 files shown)


luotao1:
PR描述可以增强下
> Upgrade shellcheck version to v0.7.1 

原来是什么版本呢？

> so users can get the warning wiki URL when they occured.

有没有示例显示？

luotao1 (approved):
LGTM

PR #28032, opened 2020-10-16
[Paddle-TRT] Average pooling uses TRT layer instead of plugin

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

Change average pooling from TRT plugin to TRT layer


shangzhizhou (approved):
LGTM

PR #28034, opened 2020-10-16
Fix diag OP bug on Windows Python3.8

LutaoChu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Fix 2 diag OP bugs on Windows Python3.8:
1. Fix diag_v2 OP uses too much memory bug when 'offset' is not 0
2. Fix the bug of unit test failure in old diag OP


wawltor (approved):
LGTM

PR #28069, opened 2020-10-19
[Cherry-pick]Fix the accuracy problem of allclose op when using float64 data type (#27891).

huangxu96:
### PR types

Bug fixes
### PR changes

OPs
### Describe

This PR fixed a bug in allclose_op, which cannot get the expected output when fp64 as input in some cases.

**BUG reproduce process:**

```python
paddle.disable_static()

np_x = np.array([10.1]).astype("float64")

np_y = np.array([10]).astype("float64")

x = paddle.to_tensor (np_x)

y = paddle.to_tensor (np_y)

result = paddle.allclose(x=x, y=y, rtol=0.01, atol=0, equal_nan=False, name="ignore_nan")

result = result.numpy()

print(result)
```

This result is expected to be **True** but it returns **false**.

**Problem reason**

Floating point number cannot be determined equal directly, since floating point number cannot have a precise experssion in
Computer. For example, 0.1 might be 0.09999 or 0.100001 in computer. So when we want to determine  two "0.1" if they are equaled, the compute might executes if 0.09999 equal to 0.100001. This is how the false result comes.

**Solving approach:**

Add an extremely small value (1e-15) when determine if two double varibles are equaled.

Change the date type of rtol and atol from float32 to float64, since the accuracy of rtol and atol will also impact the final result.

### DIFF
--- a/paddle/fluid/operators/allclose_op.cc
+++ b/paddle/fluid/operators/allclose_op.cc
@@ -13,11 +13,48 @@
 // limitations under the License.
 
 #include "paddle/fluid/operators/allclose_op.h"
+#include <cmath>
 #include "paddle/fluid/framework/op_registry.h"
 #include "paddle/fluid/framework/operator.h"
+#include "paddle/fluid/platform/enforce.h"
 
 namespace paddle {
 namespace operators {
+
+template <typename T>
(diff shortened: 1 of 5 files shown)


wzzju (approved):
LGTM.

Superjomn (approved):
LGTM

PR #28095, opened 2020-10-19
fix the bug that WITH_GLOO takes no effect when WITH_DISTRIBUTE is set

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

fix the bug that WITH_GLOO takes no effect when WITH_DISTRIBUTE is set. The root reason is that set WITH_GLOO=ON is put after ```include(configure)``` where PADDLE_WITH_GLOO is defined resulting in PADDLE_WITH_GLOO is not defined even though WITH_GLOO=ON is set automatically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,6 +239,12 @@ if(WITH_PROFILER)
     add_definitions(-DWITH_GPERFTOOLS)
 endif()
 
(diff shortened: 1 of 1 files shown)


wangxicoding (approved):
LGTM

luotao1 (approved):
LGTM

PR #28109, opened 2020-10-20
fill_constant op supports NaN and Inf

zhiqiu:
### PR types

Bug fixes 
### PR changes

OPs
### Describe

fill_constant op supports NaN and Inf

For example,
```
import paddle
import numpy as np
print(paddle.fluid.layers.fill_constant([1], 'float32', np.nan))
print(paddle.fluid.layers.fill_constant([1], 'float32', np.inf))
```

- before
```
Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
       [0.])
Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
       [0.])
```
- after
```
Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
       [nan.])
Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
       [inf.])
```

### DIFF
--- a/paddle/fluid/operators/fill_constant_op.h
+++ b/paddle/fluid/operators/fill_constant_op.h
@@ -14,9 +14,11 @@ limitations under the License. */
 
 #pragma once
 
+#include <limits>
 #include <sstream>
 #include <string>
 #include <vector>
+
 #include "paddle/fluid/framework/data_type.h"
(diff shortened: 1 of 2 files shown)


wawltor (approved):
LGTM

PR #28111, opened 2020-10-20
Add new api: is_tensor

Ray2020BD:
### PR types

Function optimization
### PR changes

APIs
### Describe

Add new api: is_tensor

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_is_tensor_op.py
+++ b/python/paddle/fluid/tests/unittests/test_is_tensor_op.py
@@ -20,13 +20,12 @@ import paddle
 DELTA = 0.00001
 
 
-class Test_is_tensor_op(unittest.TestCase):
+class TestIsTensorApi(unittest.TestCase):
     def test_is_tensor_real(self, dtype="float32"):
         """Test is_tensor api with a real tensor
         """
         x = paddle.rand([3, 2, 4], dtype=dtype)
-        res_expect = True
-        self.assertEqual(paddle.is_tensor(x), res_expect)
+        self.assertTrue(paddle.is_tensor(x))
 
     def test_is_tensor_complex(self, dtype="float32"):
         """Test is_tensor api with a complex tensor
@@ -34,22 +33,19 @@ class Test_is_tensor_op(unittest.TestCase):
         r = paddle.to_tensor(1)
         i = paddle.to_tensor(2)
         x = paddle.ComplexTensor(r, i)
-        res_expect = True
-        self.assertEqual(paddle.is_tensor(x), res_expect)
+        self.assertTrue(paddle.is_tensor(x))
 
     def test_is_tensor_list(self, dtype="float32"):
         """Test is_tensor api with a list
         """
         x = [1, 2, 3]
-        res_expect = False
-        self.assertEqual(paddle.is_tensor(x), res_expect)
+        self.assertFalse(paddle.is_tensor(x))
 
     def test_is_tensor_number(self, dtype="float32"):
         """Test is_tensor api with a number
         """
         x = 5
-        res_expect = False
-        self.assertEqual(paddle.is_tensor(x), res_expect)
+        self.assertFalse(paddle.is_tensor(x))
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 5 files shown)


python/paddle/tensor/logic.py
```diff
@@ -372,3 +374,35 @@ def not_equal(x, y, name=None):
...
+def is_tensor(x):
+    """
+
+    This function tests whether input object is a paddle.Tensor or a paddle.ComplexTensor.
+
+    Args:
+        x (object): Object to test.
+
+    Returns:
+        Tensor: A boolean value. True if 'x' is a paddle.Tensor or a paddle.ComplexTensor.
```
zhhsplendid: 1. Returns "a Tensor" or "a boolean value" ? From code you are returning "a boolean value". Please delete "Tensor:"

2. "True if 'x' is a paddle.Tensor or a paddle.ComplexTensor, otherwise False".
Add "otherwise False"

python/paddle/tensor/logic.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import paddle
```
zhhsplendid: I'm not sure whether here will cause "import each other" problem. I saw they just used "from xxx import xxx". Let's see whether CI tests pass. If you can, please change to "from xxx import xxx" to match styles.

python/paddle/fluid/tests/unittests/test_is_tensor_op.py
```diff
@@ -0,0 +1,56 @@
...
+
+DELTA = 0.00001
+
+
+class Test_is_tensor_op(unittest.TestCase):
+    def test_is_tensor_real(self, dtype="float32"):
+        """Test is_tensor api with a real tensor
+        """
+        x = paddle.rand([3, 2, 4], dtype=dtype)
+        res_expect = True
```
zhhsplendid: You can make code shorter in the three test cases here:
```
self.assertTrue(paddle.is_tensor(x))
```

You don't need `res_expect` because `assert` means that we expect the result will be xxx

python/paddle/fluid/tests/unittests/test_is_tensor_op.py
```diff
@@ -0,0 +1,56 @@
...
+
+from __future__ import print_function
+
+import unittest
+import paddle
+
+DELTA = 0.00001
+
+
+class Test_is_tensor_op(unittest.TestCase):
```
zhhsplendid: Coding style: python class is usually CapWords.

Reference:
https://google.github.io/styleguide/pyguide.html#3162-naming-conventions

Ray2020BD: Modified.

python/paddle/fluid/tests/unittests/test_is_tensor_op.py
Ray2020BD: Modified.

python/paddle/tensor/logic.py
Ray2020BD: Modified.

python/paddle/tensor/logic.py
Ray2020BD: Modified.

jzhang533 (dismissed):
lgtm

python/paddle/tensor/logic.py
```diff
@@ -372,3 +375,35 @@ def not_equal(x, y, name=None):
     """
     out = fluid.layers.not_equal(x, y, name=name, cond=None)
     return out
+
+
+def is_tensor(x):
+    """
+
+    This function tests whether input object is a paddle.Tensor or a paddle.ComplexTensor.
```
zhhsplendid, line 383: Forget to say in last comment, since this API doc will also represent on official website, could you build a preview screenshot and attach on this PR?

Now it is LGTM for code, but I'm afraid of Doc.

zhhsplendid (commented):
LGTM

python/paddle/tensor/logic.py
zhhsplendid: Discussed offline, now the 2.0 API may have to change FluidDoc to show correct English Doc page. So we won't put English Doc preview here but in the next FluidDoc PR.

zhhsplendid (dismissed):
LGTM

XiaoguangHu01 (dismissed):
LGTM

zhhsplendid (approved):
LGTM

jzhang533 (approved):
lgtm

XiaoguangHu01 (approved):
LGTM

PR #28112, opened 2020-10-20
add static_mode_white_list

pangyoki:
### PR types

Others

### PR changes

Others

## Background

Paddle 2.0 uses dygraph mode by default. PR #27443 has changed the default mode from static mode to dygraph mode.

However, there are some problems because of unittests are implemented based on static mode. It will cause conflicts when dygraph mode is opened by default.

In order to pass CI, before the unittest runs, PR #27443 switches the mode to static mode. The specific method is to modify `test_runner.py`, and the execution environment is static mode if `ctest` is used to run.

### the behavior of using `ctest` and `python` to run unittests is inconsistent.

- The execution environment of`ctest` is static mode
- The execution environment of `python` is dygraph mode

### developers need to develop unittest in static mode

Because the execution environment of`ctest` is static mode. In order to pass CI, developers need to develop unittest in static mode, it's inconsistent with the default mode of Paddle.

## Solution

Add a `static_mode_white_list`. The white list saves the tesing modules whose execution environment is static mode.
`ctest` will change dygraph mode to static mode only if the test module is in `static_mode_white_list`. 

- If unittest module is in `static_mode_white_list`, use static mode to run `ctest`
- If unittest module is not in `static_mode_white_list`, use dygraph mode to run `ctest`

Thus, subsequent developers can use the dygraph mode to develop unittest.

## Next task

`static_mode_white_list` can't solve the problem of inconsistent environment when using `ctest` and `python` to execute.
Unittest modules in `static_mode_white_list` need to be modified so that they can be executed in dygraph mode by default.

For example, we can add `paddle.enable_static` in every unittest modules. This work will be discussed later.


zhiqiu (approved):
LGTM

PR #28116, opened 2020-10-20
[Dy2stat] Refine return mechanism in @to_static

Aurelius84:
### PR types

Bug fixes 
### PR changes

Others
### Describe

Refine return mechanism in @to_static

**sample code:**
```python

@to_static
def foo(x, label):
     y = fc(x)
     loss = calc_loss(y, label)
     return y, loss
```
In above case, the returned value is both list `[y, loss]` in dygraph and static graph. 

However, we unpack the list if the returned list only contains one element before this PR, which has inconsistent behavior between dygraph and static graph.
```python

def foo(x):
     y = fc(x)
     return [y]

dy_out = foo(x_var)  # type(list)
st_out = to_static(foo)(x_var)  # type(Tensor)

# After this PR
st_out = to_static(foo)(x_var) # type(list)
```

In this PR, we fix this problem.


liym27 (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #28118, opened 2020-10-20
Fix the name error and exit caused by judgment failed

Avin0323:
### PR types

Others

### PR changes

Others

### Describe

- Fix the name error, use `vs_installer.exe` instead of `visual_installer.exe`.
- Test `nvcc` and `java` via `cmd /C` to avoid direct failure.
- Add argument `/E /Y /R` for `xcopy` command to copy file silent.


zhwesky2010 (approved):
LGTM

PR #28171, opened 2020-10-21
[Dy2stat]Join break cond with while cond in some pattern

Aurelius84:
### PR types

Others

### PR changes

Others
### Describe

Join break cond with while cond in some pattern.

Consider the follow code:
```python
import paddle
import time

class Net(paddle.nn.Layer):
    def __init__(self):
        super(Net, self).__init__()
        self.linear = paddle.nn.Linear(224, 224)

    @paddle.jit.to_static
    def forward(self, x):
        for i in range(4):
            if x.mean() > 10:
                break
            x = self.linear(x)

        return x
```

Before this PR, `break` statement will introduce a `cond_op` and wraps all behind ops into it. It was proved very time-consuming. The transformed code is :
```python
def forward(self, x):
    __break_0 = paddle.fluid.layers.fill_constant(shape=[1], dtype='bool',
        value=False)
    i = 0

    def while_condition_0(x, __break_0, i):
        return paddle.jit.dy2static.convert_logical_and(x=i + 1 <= 4, y=
            paddle.jit.dy2static.convert_logical_not(__break_0))

    def while_body_0(x, __break_0, i):

        def true_fn_0(__break_0):
            __break_0 = paddle.fluid.layers.fill_constant(shape=[1], dtype=
                'bool', value=True)
            return __break_0

        def false_fn_0(__break_0):
            return __break_0
        __break_0 = paddle.jit.dy2static.convert_ifelse(paddle.jit.
            dy2static.convert_call(x.mean)() > 10, true_fn_0, false_fn_0, (
            __break_0,), (__break_0,), (__break_0,))

        def true_fn_1(x):
            x = paddle.jit.dy2static.convert_call(self.linear)(x)
            return x

        def false_fn_1(x):
            return x
        x = paddle.jit.dy2static.convert_ifelse(paddle.jit.dy2static.
            convert_logical_not(__break_0), true_fn_1, false_fn_1, (x,), (x
            ,), (x,))
        i += 1
        return x, __break_0, i
    [x, __break_0, i] = paddle.jit.dy2static.convert_while_loop(
        while_condition_0, while_body_0, [x, __break_0, i])
    return x
``` 

In this PR, we join break conditional statement with while to avoid the overhead of`cond_op. 
The transformed code is:
```python
def forward(self, x):
    i = 0

    def while_condition_0(x, i):
        return paddle.jit.dy2static.convert_logical_and(x=i + 1 <= 4, y=
            paddle.jit.dy2static.convert_logical_not(paddle.jit.dy2static.
            convert_call(x.mean)() > 10))

    def while_body_0(x, i):
        x = paddle.jit.dy2static.convert_call(self.linear)(x)
        i += 1
        return x, i
    [x, i] = paddle.jit.dy2static.convert_while_loop(while_condition_0,
        while_body_0, [x, i])
    return x
```

We evaluate the cost time by:
```python

net = Net()

for i in range(10):
    start = time.time()
    out = net(x)
    end = time.time()
    print('cost time: ', end - start)
```

| Before(static)| After(static) |Dygraph |Speed up|
|:----:|:----:|:----:|:----:|
|85 ms| 11 ms |9 ms|7x|


zhhsplendid (commented):
Is it easy to split optimization transform and break_continue_transformer?

I think an easy-to-manage organization in the future is that we keep the old functional transformer classes and have some optimization transformer classes (Or to make it distinguishable, we can rename "transformer" for optimization to something else)

Aurelius84:
> Is it easy to split optimization transform and break_continue_transformer?
> 
> I think an easy-to-manage organization in the future is that we keep the old functional transformer classes and have some optimization transformer classes (Or to make it distinguishable, we can rename "transformer" for optimization to something else)

That's cool. I will push a new commit to split as optimization.

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
```diff
@@ -29,6 +29,28 @@
 
 from paddle.fluid import unique_name
 
+
+class BaseNodeVisitor(gast.NodeVisitor):
```
zhhsplendid, line 33: Should be "BaseNodeTransformer" inherited from "gast.NodeTransformer" because Visitor doesn't change node.

python/paddle/fluid/dygraph/dygraph_to_static/break_continue_transformer.py
```diff
@@ -103,41 +104,22 @@ class BreakContinueTransformer(gast.NodeTransformer):
         set continue to False at the beginning of each loop
 
         TODO: more details should be summarized as design document
+
+    Note: The class is inherited from BaseNodeVisitor instead of NodeTransformer,
```
zhhsplendid, line 108: As I commented below that "BaseNodeVisitor" had better to be "BaseNodeTransformer", you probably would like to change the Note here.

python/paddle/fluid/dygraph/dygraph_to_static/break_continue_transformer.py
```diff
@@ -282,8 +265,100 @@ def _add_stmt_into_list_before_node(self, stmt_list, node, stmt_node):
...
+def _find_ancestor_loop_index(node, ancestor_nodes):
+    for i in range(len(ancestor_nodes) - 1, -1, -1):
+        if isinstance(ancestor_nodes[i], (gast.For, gast.While)):
+            return i
+    return -1
+
+
+class BreakTransformOptimizer(BaseNodeVisitor):
+    """
+    In some pattern, the transformed code could be optimized by joining the
```
zhhsplendid: Had better to say what pattern it is, then code readers can know what this class optimizes immediately. From your code, I think it is "If there is a conditional 'break' statement right after loop".

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
Aurelius84: As discuss offline

python/paddle/fluid/dygraph/dygraph_to_static/break_continue_transformer.py
Aurelius84: As discuss offline

python/paddle/fluid/dygraph/dygraph_to_static/break_continue_transformer.py
Aurelius84: Thanks, fixed.

zhhsplendid (approved):
LGTM

Issue #9540, opened 2018-03-31
NCCLTester.ncclBcastOp randomly fails


helinwang:
Sometimes I get this error:

```
[==========] Running 4 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 4 tests from NCCLTester
[ RUN      ] NCCLTester.ncclInitOp
[       OK ] NCCLTester.ncclInitOp (901 ms)
[ RUN      ] NCCLTester.ncclAllReduceOp
[       OK ] NCCLTester.ncclAllReduceOp (136 ms)
[ RUN      ] NCCLTester.ncclReduceOp
[       OK ] NCCLTester.ncclReduceOp (24 ms)
[ RUN      ] NCCLTester.ncclBcastOp
/paddle/paddle/fluid/operators/nccl_op_test.cu.cc:281: Failure
The difference between ct[j] and result is 43, which exceeds 1e-5, where
ct[j] evaluates to 85,
result evaluates to 42, and
1e-5 evaluates to 1.0000000000000001e-05.
[  FAILED  ] NCCLTester.ncclBcastOp (7 ms)
[----------] 4 tests from NCCLTester (1069 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test case ran. (1069 ms total)
[  PASSED  ] 3 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] NCCLTester.ncclBcastOp

 1 FAILED TEST
```

helinwang:
https://github.com/PaddlePaddle/Paddle/pull/9541 temporarily disable the test, but we need to fix it.

luotao1:
In #9367, ncclAllReduceOp randomly fails as well. Could it be temporarily disable too?

helinwang:
Thanks @luotao1 , it's disabled as well by https://github.com/PaddlePaddle/Paddle/pull/9541

helinwang:
Reopening, since this issue should be fixed by really solving the problem, instead of comment out the test.

Issue #9367, opened 2018-03-26
nccl_op_test fails


luotao1:
#9359 only fix typo, but unit test of Teamcity fails:
https://paddleci.ngrok.io/viewLog.html?buildId=31543&buildTypeId=Paddle_PrCi&tab=buildLog&_focus=11520
```
[11:00:17]	142/215 Test #135: nccl_op_test ....................................***Failed    8.90 sec
[11:00:17]	[==========] Running 4 tests from 1 test case.
[11:00:17]	[----------] Global test environment set-up.
[11:00:17]	[----------] 4 tests from NCCLTester
[11:00:17]	[ RUN      ] NCCLTester.ncclInitOp
[11:00:17]	[       OK ] NCCLTester.ncclInitOp (746 ms)
[11:00:17]	[ RUN      ] NCCLTester.ncclAllReduceOp
[11:00:17]	/paddle/paddle/fluid/operators/nccl_op_test.cu.cc:183: Failure
[11:00:17]	The difference between ct[j] and expected_result is 85, which exceeds 1e-5, where
[11:00:17]	ct[j] evaluates to 0,
[11:00:17]	expected_result evaluates to 85, and
[11:00:17]	1e-5 evaluates to 1.0000000000000001e-05.
[11:00:17]	[  FAILED  ] NCCLTester.ncclAllReduceOp (15 ms)
[11:00:17]	[ RUN      ] NCCLTester.ncclReduceOp
[11:00:17]	[       OK ] NCCLTester.ncclReduceOp (3 ms)
[11:00:17]	[ RUN      ] NCCLTester.ncclBcastOp
[11:00:17]	/paddle/paddle/fluid/operators/nccl_op_test.cu.cc:281: Failure
[11:00:17]	The difference between ct[j] and result is 43, which exceeds 1e-5, where
[11:00:17]	ct[j] evaluates to 85,
[11:00:17]	result evaluates to 42, and
[11:00:17]	1e-5 evaluates to 1.0000000000000001e-05.
[11:00:17]	[  FAILED  ] NCCLTester.ncclBcastOp (5 ms)
[11:00:17]	[----------] 4 tests from NCCLTester (769 ms total)
[11:00:17]	
[11:00:17]	[----------] Global test environment tear-down
[11:00:17]	[==========] 4 tests from 1 test case ran. (769 ms total)
[11:00:17]	[  PASSED  ] 2 tests.
[11:00:17]	[  FAILED  ] 2 tests, listed below:
[...]

PR #28172, opened 2020-10-21
Fix nccl op test failed

wangxicoding:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix nccl op test failed, issue https://github.com/PaddlePaddle/Paddle/issues/9367 and https://github.com/PaddlePaddle/Paddle/issues/9540 . 
The reason for the failure of nccl_op_test is that the test uses the async GPU->CPU copy, but it does not sync at last. Add sync at last to fix this problem.


gongweibao (approved):
LGTM

PR #28194, opened 2020-10-21
Fix bug of fetch_async_op_handle when fetching the feed variable

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix bug of fetch_async_op_handle when fetching the feed variable

### Why
In https://github.com/PaddlePaddle/Paddle/pull/26643, `FetchAsyncOpHandle` is added to replace `FetchOpHandle`.
The difference is,

- FetchAsyncOpHandle uses a different cuda stream (other than the execution stream) to perform memory copy when doing fetch, while FetchOpHandle uses the same cuda stream.
- FetchAsyncOpHandle uses CUDAPinnedPlace for target fetch var, while FetchOpHandle uses CPUPlace.
- FetchAsyncOpHandle uses cuda event to perform synchronization between different stream, while FetchOpHandle uses cuda stream wait.

FetchAsyncOpHandle has better performance than FetchOpHandle. 

However, there is a hard-to-detect bug that may result in data non-determination in the previous implementation.

Consider the following case, fetch the feed variable immediately,

- execute stream: CPU->GPU copy (feed)
- fetch stream: GPU->CUDAPinned (fetch)

When the two operations are performed on different streams, since the feed variable has no generated op, it has no event and cannot perform event wait. So the fetch may happen before the feed, which may result in errors.

The random failure of unit test `test_buffer_shared_memory_reuse_pass` and `test_buffer_shared_memory_reuse_pass_and_fuse_optimization_op_pass` is caused by that.

### HOW
This PR fixes that problem by adding cuda stream wait for that special case, to avoid data non-determination. 

### Related
Memcpy synchronization behavior in cuda
https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-async

### DIFF
--- a/paddle/fluid/framework/details/fetch_async_op_handle.cc
+++ b/paddle/fluid/framework/details/fetch_async_op_handle.cc
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 #include "paddle/fluid/framework/details/fetch_async_op_handle.h"
+
 #include <string>
 #include <utility>
+
 #include "paddle/fluid/platform/profiler.h"
 
 namespace paddle {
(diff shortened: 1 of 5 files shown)


wanghuancoder (approved):
LGTM

PR #28221, opened 2020-10-22
Add compile limit for PADDLE_ENFORCE without error message

chenwhql:
### PR types

Function optimization
### PR changes

Others
### Describe

Add compile limit for PADDLE_ENFORCE without error message

In the paddle framework, there are more than 5000 error check items  that have no error types or no error notifications have been completed. This PR prohibits the check that **do not write error message** from the compilation level to better manage the quality of error message.

The error of not writing the type is not prohibited, because the windows gpu does not recognize `__CUDA_ARCH__` when compiling, if it is prohibited, the compilation will fail. This is a hidden danger of windows compilation. Follow-up is still needed. The reason is currently unknown.


luotao1 (approved):
LGTM

PR #28222, opened 2020-10-22
fix windows CUDA9 error due to BuildCustomizations

zhwesky2010:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix CUDA9 error due to BuildCustomizations.

This is caused by https://github.com/PaddlePaddle/Paddle/pull/27477 , which modifie NVCC compile Flag template, but is based on CUDA10 template, it has slight difference with CUDA9. This PR fix it.


jiweibo (approved):
LGTM

PR #28260, opened 2020-10-26
[Dy2Stat-log]Call warnings.warn() to display the warn-message only once

liym27:
### PR types

Others
### PR changes

Others
### Describe

Call `warnings.warn` but not `logging_utils.warn` because by default warnings.warn(message) will show up **only once**. `StaticFunction.__call__` will run many times, it is appropriate to display the warning  message in `StaticFunction.__call__` only once.

Reference: https://docs.python.org/3.7/library/warnings.html#warnings.warn

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
@@ -20,6 +20,7 @@ import inspect
 import six
 import textwrap
 import threading
+import warnings
 import weakref
 
 from paddle.fluid import framework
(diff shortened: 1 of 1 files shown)


Aurelius84 (commented):
Should we also replace `logging_utils` in ProgramTranslator.get_output?

liym27:
> Should we also replace `logging_utils` in ProgramTranslator.get_output?

Yes, done

chenwhql (approved):
LGTM

Aurelius84 (approved):

PR #28275, opened 2020-10-27
Set static shape for shape tensor with constant [part 1]

zhiqiu:
### PR types

Function optimization
### PR changes

APIs
### Describe

Set static shape for shape tensor with constant

### Why
Sometimes, the shape cannot be fully inferred in the c++ side. 

For example, consider the following code,
```
    import paddle
    paddle.enable_static()
    data = paddle.static.data(name="x", shape=[-1, 2], dtype='float32')
    shape = paddle.shape(data)  # shape should be [-1, 2] instead of [-1, -1]
    x = paddle.uniform(shape) 
    print(x.shape) 
    # (-1, -1)
    out = paddle.static.fc(x, 10)
    # Error, since the shape of weight of fc will be (-1, 10), which is wrong. 
```
Since the given shape of `uniform_random` is got by `layers.shape`, it is a Variable(Tensor) indeed, which is not initialized at compile time. So we cannot infer the right shape for `uniform_random`.

### How
However, in the above case, we know that the shape of `data` is (-1, 2) where `-1` is used for batch size. So the shape of `x` is exactly the same as `data`, which is (-1, 2).

The shape information is actually at the python side, and we can infer it by `looking forward`. 

It means that if the shape tensor of `uniform_random` is  got by `layers.shape`, we can use the shape of input Variable of `layers.shape` as the shape of the output Variable of `uniform_random`.

This may need to be fixed in several operators that involve `shape tensor`, this PR only handles `uniform_random` as a test example.

### Result
```
    import paddle
    paddle.enable_static()
    data = paddle.static.data(name="x", shape=[-1, 2], dtype='float32')
    shape = paddle.shape(data)  # shape should be [-1, 2] instead of [-1, -1]
    x = paddle.uniform(shape) 
    print(x.shape) 
    # (-1, 2)
    out = paddle.static.fc(x, 10)
    # OK, since the shape of weight of fc will be (2, 10), which is ok. 
```

### DIFF
--- a/python/paddle/fluid/layers/utils.py
+++ b/python/paddle/fluid/layers/utils.py
@@ -398,8 +398,8 @@ def try_set_static_shape_tensor(tensor, shape):
         if -1 in tensor.shape:
             if isinstance(shape, Variable):
                 shape = try_get_constant_shape_from_tensor(shape)
-                print('shape', shape)
-                tensor.desc.set_shape(shape)
+                if shape:
+                    tensor.desc.set_shape(shape)
 
 
 def try_get_constant_shape_from_tensor(shape_tensor):
@@ -417,16 +417,15 @@ def try_get_constant_shape_from_tensor(shape_tensor):
     
     """
     if in_dygraph_mode():
-        return shape_tensor.shape
+        return shape_tensor.numpy().to_list()
 
     try:
         if shape_tensor.op is not None:
             generate_op = shape_tensor.op
             if generate_op.type == 'shape':
-                print(generate_op.input_names[0])
                 var = shape_tensor.block.vars[generate_op.input_arg_names[0]]
                 return var.shape
     except:
-        return shape_tensor.shape
+        return None
 
-    return shape_tensor.shape
+    return None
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/layers/utils.py
```diff
@@ -378,3 +378,53 @@ def check_shape(shape):
...
+
+
+def try_set_static_shape_tensor(tensor, shape):
+    """Try to set static shape of tensor from a shape tensor.
+    
+    For example,
+
+    import paddle
+    paddle.enable_static()
+    data = paddle.fluid.layers.data(name="x", shape=[-1, 2], dtype='float32')
```
lanxianghit: 不应该再使用fluid.layers.data这个API了吧，其他是不是也尽量使用2.0的API

zhiqiu: > 不应该再使用fluid.layers.data这个API了吧，其他是不是也尽量使用2.0的API

好的，我修改下。

lanxianghit (approved):

PR #28300, opened 2020-10-28
Fix lr setting of AdamW when lr is an instance of LRScheduler

guoshengCS:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix lr setting of AdamW when lr is an instance of LRScheduler.


saxon-zh (approved):
LGTM

PR #28307, opened 2020-10-29
Hide some logs of p2p

zhiqiu:
### PR types

Others
### PR changes

Others
### Describe

Hide some logs of p2p

The following log message may show when (1)numpy>=1.19; (2) python>=3.7.
image

However, it is **not a bug but a feature of NVIDIA's GPU cards that not every two cards can do peer to peer communication**.

So, I change the log message from  `LOG(WARNING)` to `VLOG(2)` to hide it by default.


iclementine (approved):
LGTM

lanxianghit (approved):
LGTM

PR #28323, opened 2020-10-30
Retry CUDA Initialization to Fix Random Failure

zhhsplendid:
### PR types

Bug fixes
### PR changes

Others
### Describe

This PR is follow up of https://github.com/PaddlePaddle/Paddle/pull/28213. On that PR we tried to decrease GPU usage, however the CI still randomly failed. So I added retry logic for the initialization of nccl and cusolver. If the initialization failed, we can retry to avoid the random failure.


liym27 (approved):
LGTM

PR #28328, opened 2020-10-30
[Dy2stat] Support to modify value of buffer tensor

Aurelius84:
### PR types

New features
### PR changes

Others
### Describe

[Dy2stat] Support to modify value of buffer tensor

### What's New?

Before this PR, the value of registered buffer tensor is not supported to modified. Now we implement it.

A simple example:

```python
class BufferNet(paddle.nn.Layer):
    def __init__(self, shape):
        super(BufferNet, self).__init__()

        self.buffer1 = paddle.zeros(shape, 'int32')
        self.buffer2 = paddle.zeros(shape, 'int32')

    @paddle.jit.to_static
    def forward(self, x):
        self.buffer1 += x
        self.buffer2 = self.buffer1 + x

        out = self.buffer1 + self.buffer2

        return out
```

### DIFF
--- a/python/paddle/fluid/dygraph/layers.py
+++ b/python/paddle/fluid/dygraph/layers.py
@@ -1023,13 +1023,20 @@ class Layer(core.Layer):
                         self._non_persistable_buffer_names_set.add(name)
                     _buffers[name] = value
                 elif _buffers is not None and name in _buffers:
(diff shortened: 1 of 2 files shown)


liym27 (approved):
LGTM

PR #28404, opened 2020-11-04
Add retry for dygraph parallel socket bind

chenwhql:
### PR types

Function optimization
### PR changes

Others
### Describe

When run dygraph parallel by `paddle.distributed.launch` or `padddle.distributed.spawn`, the launch module will find freee ports for this training default, but after getting free port set, the ports will not be used immediately, they are unbound in a short time before they are used. In this short time, these free ports may be occupied by other tasks, which will cause binding or connecting error when nccl eenvironment is initialized. The error like:

image

Strictly speaking, after we obtain free ports, we should lock these ports until they are used by epected tasks, but the current logic is flawed, and the locking behavior is also complicated.

Here, I add retry mechanism to alleviate this problem.


littletomatodonkey (commented):
LGTM

phlrain (approved):

PR #28475, opened 2020-11-08
[Fix bug] IsCompatible should return false if the pass name is not found

lidanqing-vv:
### PR types
Bug fixes

### PR changes
Others 

### Describe
if the pass name is not found but still return true, in the python fuse tests
```
      self.assertTrue(
          PassVersionChecker.IsCompatible("anything_pass"))
```
any unregistered pass name will pass UT.

@luotao1 @shangzhizhou  Could you please take a look? Thanks


shangzhizhou (dismissed):
LGTM

lidanqing-vv:
@shangzhizhou  Hi, Please review again.
1. rename file `test_conv_bias_mkldnn_fuse_pass.py` to `test_mkldnn_conv_bias_fuse_pass.py` is necessary. Because first one will still be tested even MKLDNN is OFF but it's not what we want. After changing to `test_mkldnn_conv_bias_fuse_pass.py` the all  CIs pass

2. `REGISTER_PASS_CAPABILITY(quant_conv2d_dequant_fuse_pass);` is necessary, because in file `test_trt_quant_conv2d_dequant_fuse_pass.py`, `PassVersionChecker.IsCompatible('quant_conv2d_dequant_fuse_pass')`  is called

shangzhizhou (approved):
LGTM

lidanqing-vv:
@luotao1  这个可以合入了吗？Intel我们组的可能不太清楚这个，然后shangzhizhou已经 approve了。

PR #28497, opened 2020-11-09
fix port conflict in parallel dygraph mode

danleifeng:
### PR types

Bug fixes
### PR changes

Others
### Describe

when running dygraph parallel code in local machine,  random failure will occur.  It may be caused by the port conflict when gloo and nccl initialized at the same time. 
rank0 log (send id, connect failed):
image
rank3 log (receive id, bind failed):
image

In this PR, we change the order of the gloo and nccl init like below:
HttpServer start -> NCCL init -> GLOO init

HttpServer is used for gloo initial, which will use rank0 port. After changing order, port conflict will be resolved.

### DIFF
--- a/python/paddle/distributed/parallel.py
+++ b/python/paddle/distributed/parallel.py
@@ -138,7 +138,6 @@ def init_parallel_env():
         http_server.daemon = True
         http_server_d["running"] = True
         http_server.start()
-    wait_server_ready([ParallelEnv().trainer_endpoints[0]])
 
     # 4. init NCCL ParallelStrategy
     strategy = ParallelStrategy()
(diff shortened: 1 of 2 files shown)


python/paddle/distributed/parallel.py
```diff
@@ -140,21 +140,6 @@ def _check_var_exists(var_name):
         http_server.start()
     wait_server_ready([ParallelEnv().trainer_endpoints[0]])
```
sandyhouse: This line can be moved to line #169 (just after the nccl initialization) to decrease the wait time.

chenwhql (approved):
LGTM

PR #28511, opened 2020-11-10
Fix dtype of ungenerated grad var

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix dtype of ungenerated grad var

### Why
In the following situation, there are more than one outputs generated by an operator, and only one of them is involved in the loss calculation. (eg, rnn op)

In backward execution, `out1@grad` is generated from loss@grad by BP, while `out0@grad` is not. However, the grad_op may need  `out0@grad`, `out1@grad`, and other forward input vars (x, y, maybe) as inputs. Since `out0@grad` is not generated, it will be set as 0. The problem is that we need to know the real `dtype` of `out0@grad`, otherwise it may result in dtype error.

```
out0, out1 = op(x, y)
loss = loss_fn(out1)
loss.backward()
```
### How
In the above situation, `out0@grad` will use the `dtype` of the related forward variable `out0` as real `dtype`.

### DIFF
--- a/paddle/fluid/imperative/basic_engine.cc
+++ b/paddle/fluid/imperative/basic_engine.cc
@@ -100,7 +100,11 @@ void BasicEngine::CheckBackwardInputs(const OpBase& op) {
 
       if (tensor && !tensor->IsInitialized()) {
         auto* dev_ctx = platform::DeviceContextPool::Instance().Get(op.place());
-        // NOTE(zhiqiu): since grad var is ungenerated, so
+        // NOTE(zhiqiu): since grad variable is ungenerated, so the dtype is not
+        // correct. var->DataType() returns the default dtype, which is float32.
+        // Here, we use the type of the corresponding forward variable (if
+        // exists) as the true dtype of grad variable.
+
         auto forward_var = var->GetForwardVar();
         auto dtype = forward_var ? forward_var->DataType() : var->DataType();
         VLOG(6) << "Set ungenerated Grad: " << var->Name()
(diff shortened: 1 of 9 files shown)


paddle/fluid/imperative/basic_engine.cc
```diff
@@ -85,9 +86,13 @@ void BasicEngine::CheckBackwardInputs(const OpBase& op) {
       }
 
       if (tensor && !tensor->IsInitialized()) {
-        VLOG(6) << "Set ungenerated Grad: " << var->Name() << " as zero";
         auto* dev_ctx = platform::DeviceContextPool::Instance().Get(op.place());
-        tensor->mutable_data(op.place(), var->DataType());
+        // NOTE(zhiqiu): since grad var is ungenerated, so
```
zhwesky2010: 注释没写全

paddle/fluid/imperative/layer.h
```diff
@@ -211,6 +212,7 @@ class VarBase {
   const std::shared_ptr<VariableWrapper> var_;
 
   std::shared_ptr<VarBase> grad_var_;
+  std::weak_ptr<VarBase> forward_var_;
```
zhwesky2010: 这个forward_var_  好像不需要用到

zhiqiu: removed, thx

paddle/fluid/imperative/basic_engine.cc
zhiqiu: done

iclementine (approved):
LGTM

phlrain (approved):

Xreki (approved):
LGTM for the use of `const_cast` and op benchmark ci.

PR #28547, opened 2020-11-11
fix user set CUDA_VISIBLE_DEVICES start/end with quotation marks

zhwesky2010:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix user set CUDA_VISIBLE_DEVICES start/end with quotation marks on windows. 

image

Which has no problem on Linux.


wanghuancoder (approved):
LGTM

PR #28557, opened 2020-11-11
Fix vgg error when num_classes is given

LielinJiang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix vgg error when num_classes is given


saxon-zh (approved):
LGTM

PR #28571, opened 2020-11-12
Reduce input data size for syncbn unit test by 8x

willthefrog:
### PR types
 Others

### PR changes
Others 

### Describe
this is a preemptive measure, the unit test was failing occasionally with cudnn internal error


qingqing01 (approved):
LGTM

PR #28612, opened 2020-11-13
Add checkpoint to quantize

wozna:
### PR types

Bug fixes
### PR changes

OPs 
### Describe

This PR :
- adds op version checkpoint to `quantize op` where new attribute `bfloat16` was added. 
- registers pass compatibility to the `cpu_bfloat16_pass` which is related to the change in `quantize op`.
- changes also `InferencePassTest`, which allowed applying cpu_bfloat16_pass to the graph

### DIFF
--- a/python/paddle/fluid/tests/unittests/ir/inference/test_mkldnn_cpu_bfloat16_pass.py
+++ b/python/paddle/fluid/tests/unittests/ir/inference/test_mkldnn_cpu_bfloat16_pass.py
@@ -43,14 +43,14 @@ class TestMKLDNNCpuBfloat16Pass(InferencePassTest):
     def init_data(self):
         self.bs = 8
         self.d_type = np.float32
-        self.shape_x = [12, 1, 1]
+        self.shape_x = [12, 10, 1]
         self.shape_y = [12, 1, 64]
         self.enable_mkldnn = True
+        self.enable_mkldnn_bfloat16 = True
 
     def test_check_output(self):
-        self.enable_mkldnn = True
         use_gpu = False
-        self.check_output_with_option(use_gpu, flatten=True, bfloat16=True)
+        self.check_output_with_option(use_gpu, flatten=True)
         self.assertTrue(PassVersionChecker.IsCompatible('cpu_bfloat16_pass'))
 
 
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/tests/unittests/ir/inference/test_mkldnn_cpu_bfloat16_pass.py
```diff
@@ -0,0 +1,58 @@
...
+                np.random.random([self.bs] + self.shape_x).astype(self.d_type),
+                "y":
+                np.random.random([self.bs] + self.shape_y).astype(self.d_type)
+            }
+            self.fetch_list = [out]
+
+    def init_data(self):
+        self.bs = 8
+        self.d_type = np.float32
+        self.shape_x = [12, 1, 1]
```
lidanqing-vv: Some time ago, baidu require test dimension to be over 100? Maybe `self.shape_x = [12, 1, 10]`

python/paddle/fluid/tests/unittests/ir/inference/test_mkldnn_cpu_bfloat16_pass.py
```diff
@@ -0,0 +1,58 @@
...
+        self.shape_y = [12, 1, 64]
+        self.enable_mkldnn = True
+
+    def test_check_output(self):
+        self.enable_mkldnn = True
```
lidanqing-vv: There are two  `self.enable_mkldnn = True`, the one in the `init_data(self)` is enough

The test shows that if `self.check_output_with_option(use_gpu, flatten=True, bfloat16=False)` then this test will do automatically float32? I thought settings should be done in init, before checking output. like `self.enable_mkldnn_bfloat = True`, But both are fine.

python/paddle/fluid/tests/unittests/ir/inference/test_mkldnn_cpu_bfloat16_pass.py
wozna: Ok, I changed.

python/paddle/fluid/tests/unittests/ir/inference/test_mkldnn_cpu_bfloat16_pass.py
wozna: Yes, I agree that just adding `self.enable_mkldnn_bfloat = True` will be better solution.

lidanqing-vv:
LGTM

luotao1 (approved):

PR #28633, opened 2020-11-15
Fix scaled_params append error and no_grad setting in AdamW

guoshengCS:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix scaled_params append error and no_grad setting in AdamW. 
Using `assign` to replace `.numpy()` and then `set_value` to speed up. (Bert-base 0.8step/s->3.43step/s).


wawltor (approved):
LGTM

PR #28648, opened 2020-11-16
fix the bug for send and recv op when using nccl version under 2.7

sandyhouse:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix the bug for send and recv op when using nccl version under 2.7


wangxicoding (approved):
LGTM

PR #28649, opened 2020-11-16
Fix output dtype inconsistent with input

Aurelius84:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix output dtype inconsistent with input


wangchaochaohu (approved):
LGTM

PR #28653, opened 2020-11-16
faster the compare ops dygraph model speed

wawltor:
### PR types
Performance optimization 

### PR changes
OPs

### Describe
faster the compare ops dygraph model speed


FlyingQianMM (approved):
LGTM

PR #28700, opened 2020-11-17
fix the bug in expand_as op

sandyhouse:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix the bug in expand_as op.

ExternalError:  Cuda error(77), an illegal memory access was encountered.


danleifeng (approved):
LGTM

PR #28703, opened 2020-11-18
Fix gpu memory allocation bug.

gongweibao:
### PR types
Bug fixes

### PR changes
Others

### Describe
Paddle will alloc 400+ MB GPU memory even you only import it.
This PR fixes it by init p2p lazily.


chenwhql (approved):
LGTM

Issue #28578, opened 2020-11-12
Paddle inference raise segmentation fault error for quantized ocr recognition model


juncaipeng:
Thank you for contributing to PaddlePaddle.
Before submitting the issue, you could search issue in the github in case that th
If there is no solution,please make sure that this is an inference issue including the following details :
**System information**
-PaddlePaddle version
commit id:72c78e4dbbdf74fe9995d9b0757ca6e5832c79d7
cmake: cmake -DCMAKE_BUILD_TYPE=Release -DWITH_PYTHON=OFF -DWITH_MKL=ON -DWITH_GPU=OFF -DON_INFER=ON ..
-CPU: CPUMKL/MKLDNN
-GPU: 
-OS Platform:  Ubuntu 16.04.5 LTS   gcc (Ubuntu 4.8.5-4ubuntu2) 4.8.5
-Python version: 3.7
Note: You can get most of the information by running [summary_env.py](https://github.com/PaddlePaddle/Paddle/blob/develop/tools/summary_env.py). 
**To Reproduce**
Steps to reproduce the behavior
**Describe your current behavior**
**Code to reproduce the issue**

Download code and modes(url: https://pan.baidu.com/s/1y8OLGP6WajnSkM_f4_FNKg  password: y0qp) and refer to the readme.md.

**Other info / logs**

![image](https://user-images.githubusercontent.com/52520497/98916830-9b45b380-2506-11eb-8030-11b4d4d40288.png)

lidanqing-vv:
data and model is downloaded at `shanghai-6271-ip:/data/issue_28578`

wojtuss:
@juncaipeng 
The issue is caused by the faulty `fc_lstm_fuse_pass` (some input and output variables are not set properly in the fuse). The pass is executed during transformation of the quant model into int8 model (during saving the int8 model), and is not executed during inference of the fp32, quant or int8 model - that is why the problem occurs only during int8 inference. It is not related to the quantization process nor mkldnn.
I have proposed a fix for the issue: https://github.com/PaddlePaddle/Paddle/pull/28709.

juncaipeng:
@wojtuss  Thank you for solving the segmentation fault.
Compared to the origin model, the results of quantized model is wrong, so please check again.

wojtuss:
@juncaipeng ,
After disabling quantization but enabling optimization passes (either all or only non-mkldnn) the output I get is:
```
[...]

PR #28709, opened 2020-11-18
a fix for the fc_lstm_fuse_pass

wojtuss:
### PR types
Bug fixes

### PR changes
Others

### Describe
This PR is a proposal of a fix for the issue https://github.com/PaddlePaddle/Paddle/issues/28578.
The problem was in the faulty `fc_lstm_fuse_pass` pass: intput and output variables were not set properly during the fuse.


juncaipeng (approved):
LGTM

PR #28713, opened 2020-11-18
[Dynamic-to-Static ErrorMessage] Support dy2stat error message when call jit.save and polish error message

liym27:
### PR types

Bug fixes 
### PR changes

Others
#### 1. Support dy2stat error message when call `paddle.jit.save`
- Before:
When call `paddle.jit.save(...)`, if exception is raised in dynamic-to-static, the exception is Python native exception and not the optimized dy2stat exception.

##### 2.2  "In user code:" -> "In transformed code:"

For example:

image


zhhsplendid (approved):
LGTM

PR #28730, opened 2020-11-18
Fix cpu_bfloat16_pass

wozna:
### PR types

Bug fixes
### PR changes

Others 
### Describe

This PR changes `cpu_bfloat16_pass`.  
Those changes add the ability to handle cases such as:
- adding reorder before op with two inputs such as matmul, elementwise,
- adding reorder before op with duplicate inputs such as concat and sum
- adding reorder after output linked to more than one  operator
All these cases have been added to  `cpu_bfloat16_pass_tester`

### DIFF
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -2108,6 +2108,10 @@ PDNode *patterns::Bfloat16Placement::operator()(
     supported_op_types = bfloat16_enabled_op_types;
   }
   auto *op = pattern->NewNode(op_repr())->assert_is_ops(supported_op_types);
+  op->assert_more([&](Node *node) {
+    return node->Op()->GetAttrIfExists<bool>("use_mkldnn") ||
+           node->Op()->Type() == "reshape2";
+  });
   return op;
 }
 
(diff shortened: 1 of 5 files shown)


paddle/fluid/framework/ir/graph_pattern_detector.cc
```diff
@@ -2176,6 +2180,36 @@ PDNode *patterns::FirstBfloat16Ops::operator()() {
   return op;
 }
 
+PDNode *patterns::DuplicatedInputs::operator()() {
```
jczaja, line 2183: what does it mean duplicated input and why only sum and concat (not elementwise_add, elementwise_mul ......) ?

jczaja (dismissed):
LGTM

paddle/fluid/framework/ir/graph_pattern_detector.cc
wozna: This is because concat and sum have an input vector that is marked  `.AsDuplicable()`
https://github.com/PaddlePaddle/Paddle/blob/0ed80e09fcbac3d62e35dc07fa451ce1a32d4eb3/paddle/fluid/operators/concat_op.cc#L109
At the graph_pattern_detecotr level, it's hard to find the right connection to add reorders before each concat and sum. That is why such a workaround was used.

jczaja (approved):
LGTM

luotao1 (approved):
LGTM

PR #28735, opened 2020-11-19
[Dynamic-to-Static] Fix bug in loop_transformer: loop vars should contain the var from ancestor-for-node

liym27:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix bug in loop_transformer: loop_vars should contain the var from ancestor-for-node.

For example:   `i` should be in `before_loop_vars` of loop-node `for j in range(a) ...`
```python
for k in range(x):    # if it's this "loop_node", `i` or `j` both should NOT be in `after_loop_vars `
 # do something

for i in range(a):     # if it's this "loop_node", `k` should NOT be in `before_loop_vars`, `j` should NOT be in `in_loop_vars`
    for j in range(a): # if it's this "loop_node", `k` should NOT be in `before_loop_vars`, `i` should be in `before_loop_vars`.
         # do something
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
@@ -294,11 +294,21 @@ class NameVisitor(gast.NodeVisitor):
             return True
         return False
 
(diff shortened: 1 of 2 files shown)


zhhsplendid (approved):
LGTM

PR #28741, opened 2020-11-19
Disable windows gpu static lib.

jiweibo:
### PR types

Others
### PR changes

Others
### Describe

Disable windows gpu static lib, because the size exceeds 4GB.

### DIFF
--- a/paddle/fluid/inference/api/demo_ci/run.sh
+++ b/paddle/fluid/inference/api/demo_ci/run.sh
@@ -1,4 +1,19 @@
 #!/bin/bash
+
+# Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


T8T9 (approved):

paddle/fluid/inference/api/demo_ci/run.sh
```diff
@@ -68,6 +83,10 @@ rm -rf *
 
 for WITH_STATIC_LIB in ON OFF; do
   if [ $(echo `uname` | grep "Win") != "" ]; then
+    # TODO(wilber, T8T9): Do we still need to support windows gpu static library
```
zhwesky2010, line 86: whether static lib is needed in CPU?

jiweibo: The cpu library is very small, it needs to be evaluated in the future whether to keep.

zhwesky2010 (approved):

PR #28763, opened 2020-11-20
[Inference] Fix trt delete_pass bug.

jiweibo:
### PR types

Bug fixes
### PR changes

Others
### Describe

- Fix trt delete pass bug.  Init a AnalysisConfig with a config which deletes some specified passes, the resulting AnalysiConfig should also remove these passes.
- Fix the memory increase caused by the inability to enable memory_optimize_pass.  https://github.com/PaddlePaddle/Paddle/pull/28517


cryoco (approved):
LGTM

PR #28780, opened 2020-11-20
lazily init global group in collective

chenwhql:
### PR types

Bug fixes
### PR changes

Others
### Describe

Lazily initialized global group information. If we initialize _default_group when import module, it will not update when we use spawn to run multi-process training


sandyhouse (approved):
LGTM

PR #28988, opened 2020-11-23
Fixes mkldnn dygraph learning rate scheduler crashes

arlesniak:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fixes mkldnn dygraph crashes for learning rate scheduler other than piecewise_decay. 
Enables mkldnn for dygraph se_resnext model from models repo and MobilenetV1 with cosine_decay lrs .

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1017,6 +1017,13 @@ bool OperatorWithKernel::SupportsMKLDNN() const {
                      });
 }
 
+bool OperatorWithKernel::CanMKLDNNBeUsed(
+    const framework::ExecutionContext& ctx) const {
+  bool use_mkldnn_ctx =
+      ctx.Attr<bool>("use_mkldnn") && platform::is_cpu_place(ctx.GetPlace());
+  return use_mkldnn_ctx && this->SupportsMKLDNN();
+}
+
 void OperatorWithKernel::RuntimeInferShape(const Scope& scope,
                                            const platform::Place& place,
                                            const RuntimeContext& ctx) const {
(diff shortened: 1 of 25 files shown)


paddle/fluid/framework/operator.cc
```diff
@@ -1007,6 +1007,15 @@ static void CheckTensorNANOrInf(const std::string& op_type,
                               op_type, name));
 }
 
+bool OperatorWithKernel::SupportMKLDNN() const {
```
wojtuss: I know there is already the method named `SupportGPU()`, but I would prefer naming `SupportsMKLDNN()` (indicative) instead of `SupportMKLDNN()` (imperative).

wojtuss: Also, maybe we could hide it inside the `CanMKLDNNBeUsed()` method?

arlesniak: CanMKLDNNBeUsed() function uses Runtime context at platform level and SupportsMKLDNN() is closer to OperatorWithKernel and needs **operation type** to proceed. 
I think both of them can be left separated for now, otherwise either CanMKLDNNBeUsed needs **op_type** to be passed or we start to base deciding on GetExpectedKernelType on OperatorWithKernel ::SupportsMKLDNN() as it has all the info needed and get rid of platform::CanMKLDNNBeUsed() as it is used mainly at the OperatorWithKernel level
What is your opinion on that ?

wojtuss (approved):
LGTM

arlesniak:
@raindrops2sea @luotao1  Due to info in PR-CI_APPROVAL could you start your review please?

luotao1 (approved):
LGTM

raindrops2sea (approved):
LGTM

PR #28989, opened 2020-11-23
Upgrade  string literals to raw string

zhiqiu:
### PR types

Others
### PR changes

APIs
### Describe

Upgrade  string literals to raw string

A backslash-character pair that is not a valid escape sequence generates a DeprecationWarning since python-3.6, see details in https://docs.python.org/3/whatsnew/3.6.html?highlight=escaped%20sequence#deprecated-python-behavior.

For example, `"""\a"""`, `"""Out = \lvert input - label\rvert"""`, which is often exists in comments.
image

This PR changes all the string literals in `"""string"""` to raw string `r"""string"""` to avoid such deprecation warnings.


swtkiwi (approved):
LGTM

raindrops2sea (approved):
LGTM

Xreki (approved):
LGTM for the modification in framework.py

PR #28992, opened 2020-11-23
[Dy2stat] Fix Incorrect After Node Vars in IfElseTransformer

zhhsplendid:
### PR types

Bug fixes
### PR changes

APIs
### Describe

**Problem**: our current Dy2stat will fail on a simple case:

```
    @paddle.jit.to_static
    def forward(self, x, y):
        if self.mode == 'train':
            out = x + y
        elif self.mode == 'infer':
            out = x - y
        else:
            raise ValueError('Illegal mode')
        return out
```

**Reason**: the if-elif-else will be transferred into two `ast.If` in `ast`. In our old code, we calculate the vars before/after if-else expression. While we got `after_var_name_ids` by

```
after_var_names_ids = [	
            ctx for ctx in all_name_ids[name] if ctx not in before_var_names_ids	
        ]
```
We are using all vars minus vars before the if-else expression. However, that's incorrect because: 

```
before_var_names_ids = parent_name_ids.get(name, []) + \	
                               body_name_ids.get(name, []) + orelse_name_ids.get(name, [])
```
From the code you could see `before_var_names_ids` doesn't include the `body_name_ids` of the previous `ast.If`.

**Solution**: the key fix of this PR is:
```
after_ifelse_name_ids = get_name_ids([root], after_node=node)
```
Where we get `after_ifelse_name_id` by visiting instead of set minus. To support it, we enable `NameVisitor` to search after the `after_node`.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/ifelse_transformer.py
@@ -91,22 +91,27 @@ class IfElseTransformer(gast.NodeTransformer):
 
 
 class NameVisitor(gast.NodeVisitor):
-    def __init__(self, end_node=None):
+    def __init__(self, after_node=None, end_node=None):
+        # The start node (exclusive) of the visitor
+        self.after_node = after_node
         # The terminate node of the visitor.
         self.end_node = end_node
         # Dict to store the names and ctxs of vars.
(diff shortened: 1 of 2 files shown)


liym27 (approved):
LGTM

Aurelius84 (approved):
LGTM

chenwhql (approved):

PR #28993, opened 2020-11-23
[Dynamic-to-Static] Fix bug of convert_logical_and/convert_logical_or:  the operands are executed sequentially

liym27:
### PR types

Bug fixes
### PR changes

Others
#### 1. The operands are executed sequentially according to the running logic of Python. So here the arguments
should be callable.
#### 2. If the left hand operand is True, the right hand operand should be executed.
 For example:
    `a = x > 1 or y < 1`
Transformed code:
   ` a = paddle.jit.dy2static.convert_logical_or(lambda:x>1, lambda:y<1)`

 In `convert_logical_or(lambda:x>1, lambda:y<1)`, `lambda:y<1` must be run after `lambda:x>1`. And
if `x>1` is True, `y<1` should NOT be run.

---

#### Case 1:
```python
@paddle.jit.to_static
def test1():
    a = None
    if a is not None and a > 1:
        x = x - 1
    else:
        x = x + 1

test1()
```

Error:
```python
TypeError: unorderable types: NoneType() > int()
```
Because `a > 1` is run,   but `a` is None and `a > 1`  should not have been run.

#### Case 2:
```python
@paddle.jit.to_static
def test():
    x = paddle.zeros([1])

    a = None

    if a is not None:
        z = x+1

    if a is not None and z < -1:
        x = x - 1
    return x

test2()
```

Error:
```python
InvalidArgumentError: The Tensor in the less_than Op's Input Variable X(z) is not initialized.
  [Hint: Expected t->IsInitialized() == true, but received t->IsInitialized():0 != true:1.] (at /home/teamcity/work/ef54dc8a5b211854/paddle/fluid/framework/operator.cc:1419)
  [operator < less_than > error]  [operator < run_program > error]

```
Because `z < -1` is run,  the OP `less_than`  is added in the program, but `z` is not initialized. Actually, `z < -1` should not have been run.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/convert_operators.py
@@ -56,25 +56,39 @@ def _run_py_while(cond, body, loop_vars):
     return loop_vars
 
 
-def convert_logical_and(x, y):
+def convert_logical_and(x_func, y_func):
     """
     A function representation of a Python ``and`` statement.
 
     Args:
-        x(bool|Tensor): Left hand operand of ``and`` operator.
-        y(bool|Tensor): Right hand operand of ``and`` operator.
+        x_func(callable): x_func() is the left hand operand of ``and`` operator. x_func() is bool or Tensor.
+        y_func(callable): y_func() is the right hand operand of ``and`` operator.  y_func() is bool or Tensor.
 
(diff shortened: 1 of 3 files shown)


Aurelius84 (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #29000, opened 2020-11-23
update pow

Joejiong:
### PR types

Performance optimization
### PR changes

APIs
### Describe

Performance optimization for power API and its operator.
The following code affects this version of paddle.pow api.
This block of code is meant to Align paddlepaddle pow api's functionality and behavior with numpy.power api, but it slowdowns the speed of this api call by 90%.
```
if x.dtype != y.dtype:
   y = cast(y, dtype='float64')
   x = cast(x, dtype='float64')
```

### DIFF
--- a/python/paddle/tensor/math.py
+++ b/python/paddle/tensor/math.py
@@ -185,14 +185,9 @@ def pow(x, y, name=None):
         if isinstance(y, (int, float)):
             return core.ops.pow(x, 'factor', y)
         elif isinstance(y, (paddle.Tensor, Variable)):
-
             if x.dtype != y.dtype:
                 y = cast(y, dtype='float64')
                 x = cast(x, dtype='float64')
-                out_dygraph = _elementwise_op_in_dygraph(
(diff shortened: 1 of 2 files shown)


TCChenlong (approved):
LGTM

python/paddle/tensor/math.py
```diff
@@ -172,27 +172,22 @@ def pow(x, y, name=None):
...
     """
     # in dynamic graph mode
     if in_dygraph_mode():
         if isinstance(y, (int, float)):
             return core.ops.pow(x, 'factor', y)
         elif isinstance(y, (paddle.Tensor, Variable)):
-
             if x.dtype != y.dtype:
                 y = cast(y, dtype='float64')
                 x = cast(x, dtype='float64')
```
Joejiong, line 190: The following code affects this version of paddle.pow api.
This block of code is meant to Align paddlepaddle pow api's functionality and behavior with numpy.power api, but it slowdowns the speed of this api call by 90%.

zhhsplendid (approved):
LGTM

PR #29018, opened 2020-11-24
Fix Mnist and FashionMnist bug when return pil image

LielinJiang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix Mnist and FashionMnist bug when return pil image


saxon-zh (approved):
LGTM

PR #29064, opened 2020-11-24
Split train_mode and has_grad for tracer

### LINKED ISSUE #28089: 网络train/eval模式异常
Paddle版本2.0.0b0，一个网络train/eval模式的改变会影响其他网络
demo代码输出False, False，期望为True, False
```
  0 import paddle
  1 import paddle.nn as nn
  2 
  3 class Net1(nn.Layer):
  4     def __init__(self):
  5         super(Net1, self).__init__()
  6         self.linear = nn.Linear(2, 3)
  7 
  8     def forward(self, x):
  9         return self.linear(x)
 10 
 11 class Net2(nn.Layer):
 12     def __init__(self):
 13         super(Net2, self).__init__()
 14         self.linear = nn.Linear(2, 3)
 15 
 16     def forward(self, x):
 17         return self.linear(x)
 18 
 19 paddle.disable_static()
 20 net1 = Net1()
 21 net1.eval()
 22 net2 = Net2()
 23 net2.train()
 24 to_float = lambda x: paddle.to_tensor(x, dtype='float32')
 25 print(net1(to_float([[1,2]])).stop_gradient)
 26 print(net2(to_float([[1,2]])).stop_gradient)
```

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Split train_mode and has_grad for tracer, close https://github.com/PaddlePaddle/Paddle/issues/28089

### Why
Before this PR, the behavior `train/eval mode` and `trcing grad or not` are coupled. 

When setting `layer.eval()`, it also means `do not trace grad`, which is not a good design.

### How
This PR couples `train/eval mode` and `trcing grad or not`. 

- Setting `train/eval mode` of a Layer only affects the internal logic of the specific operators (dropout and batchnorm), since they have different algorithms in train mode and eval mode.
- If really want  "do not trace grad", use `paddle.no_grad` instead.

Also, see details in https://github.com/PaddlePaddle/Paddle/pull/25024

### DIFF
--- a/python/paddle/fluid/layers/nn.py
+++ b/python/paddle/fluid/layers/nn.py
@@ -997,7 +997,10 @@ def dropout(x,
 
         .. code-block:: python
 
+            import paddle
             import paddle.fluid as fluid
+            
+            paddle.enable_static()
             x = fluid.data(name="data", shape=[None, 32, 32], dtype="float32")
             dropped = fluid.layers.dropout(x, dropout_prob=0.5)
     """
--- a/python/paddle/fluid/layers/nn.py
+++ b/python/paddle/fluid/layers/nn.py
@@ -945,7 +945,7 @@ def cos_sim(X, Y):
 @deprecated(since="2.0.0", update_to="paddle.nn.functional.dropout")
 def dropout(x,
             dropout_prob,
-            is_test=False,
+            is_test=None,
             seed=None,
             name=None,
             dropout_implementation="downgrade_in_infer"):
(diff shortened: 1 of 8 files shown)


phlrain (approved):

python/paddle/fluid/layers/nn.py
```diff
@@ -996,7 +997,10 @@ def dropout(x,
 
         .. code-block:: python
 
+            import paddle
             import paddle.fluid as fluid
+            
+            paddle.enable_static()
             x = fluid.data(name="data", shape=[None, 32, 32], dtype="float32")
```
swtkiwi, line 1004: 建议尽量不对外保留fluid目录啦？fluid.data->paddle.static.data

lanxianghit (approved):

python/paddle/fluid/layers/nn.py
zhiqiu: 好的，这个另外提PR修。

swtkiwi (approved):
LGTM，辛苦中文文档也提PR改下

Issue #28554, opened 2020-11-11
The performance of OCR detection model become slower when enable mkldnn


juncaipeng:
Thank you for contributing to PaddlePaddle.
Before submitting the issue, you could search issue in the github in case that th
If there is no solution,please make sure that this is an inference issue including the following details :
**System information**
-PaddlePaddle version: develop (commit id: 72c78e4dbbdf74fe9995d9b0757ca6e5832c79d7) or 1.8.5
-CPU: 6148 with MKLDNN
-GPU: None
-OS Platform:Ubuntu 16.04.5
-Python version:3.7
Note: You can get most of the information by running [summary_env.py](https://github.com/PaddlePaddle/Paddle/blob/develop/tools/summary_env.py). 
**To Reproduce**
Steps to reproduce the behavior
**Describe your current behavior**
**Code to reproduce the issue**
1. Download the code and model, url: https://pan.baidu.com/s/1X5jdXj6Vq4dL1Vkk_yGplw  password: kv1y

2. Refer to readme.md and run the program

![image](https://user-images.githubusercontent.com/52520497/98806947-e7382000-2454-11eb-9b9e-9f932e54849e.png)
![image](https://user-images.githubusercontent.com/52520497/98807354-7d6c4600-2455-11eb-899c-0b2dca7c53bb.png)

3. Modify the config to disable mkldnn and run the program

![image](https://user-images.githubusercontent.com/52520497/98806920-df787b80-2454-11eb-9a4c-62af16c638ae.png)

![image](https://user-images.githubusercontent.com/52520497/98807322-6f1e2a00-2455-11eb-9283-460216a8d0ac.png)

**Other info / logs**

lidanqing-vv:
model is downloaded at `shanghai-6271-ip:/data/issue_28554` (for intel colleague)

juncaipeng:
The OCR recognition model also has the same problem. 
The demo can be downloaded from #28578 , and only consider the fp32 model.

lidanqing-vv:
Fix PR: https://github.com/PaddlePaddle/Paddle/pull/29103. I am still verifying accurcay
Native
```
------------------------->     Profiling Report     <-------------------------

Place: CPU
Time unit: ms
Sorted by total time in descending order in the same thread

-------------------------     Overhead Summary      -------------------------

Total time: 7622.93
  Computation time       Total: 7594.25     Ratio: 99.6238%
[...]

PR #29103, opened 2020-11-25
Enable Interpolate mkldnn kernel

lidanqing-vv:
### PR types
Performance optimization

### PR changes
OPs 

### Describe
Enable interpolate mkldnn kernel, so that ocr_det model inference performance is improved, 
Machine: 6148 
Reproduction: #28554 
Performance: 
- CPU Native 7576 ms
- oneDNN 3700 ms

Status: The draft PR open for verifying performance and accuracy. I am polishing up the code.


lidanqing-vv:
@juncaipeng  The accuracy error should be fixed, I checked first 500 numbers, the same. I will finish all UT tomorrow.

juncaipeng:
> @juncaipeng The accuracy error should be fixed, I checked first 500 numbers, the same. I will finish all UT tomorrow.

OK, The accuracy and performance of ocr crnn model

lidanqing-vv:
@juncaipeng Since oneDNN use different formula for interpolate. [oneDNN resampling](https://docs.oneapi.com/versions/latest/onednn/dev_guide_resampling.html)
The user may get different result when running oneDNN interpolate compared with Native interpolate, but sometimes users can get same results. But I checked tensorflow, tensorflow used same formula as oneDNN. So we have a question if it is ok that applying oneDNN interpolate sometimes have different results than Native, sometimes the same.

Since other frameworks like Tensorflow is using same formula as oneDNN, the oneDNN accuracy may **slightly** differ compared to Natvie.

juncaipeng:
@lidanqing-intel 
Compared to oneDNN, the interpolate of  Paddle has align_corners parameter and the calculation method is different.
We need to check whether the slight difference affects the final result. For the test image, the ocr detection model has  right result, but we don't test the final accuracy.

lidanqing-vv:
Hi, @jczaja  could you please review this PR

lidanqing-vv:
@juncaipeng  https://github.com/oneapi-src/oneDNN/issues/653

luotao1:
@lidanqing-intel @jczaja Discussed internal, our interpolate API behaves in the same way as Pytorch, not TF. See [torch.nn.functional.interpolate](https://pytorch.org/docs/stable/nn.functional.html?highlight=interpolate#torch.nn.functional.interpolate) and its [chinese explanation](https://www.cnblogs.com/wanghui-garcia/p/11399034.html). Thus, unfortunately, we may not call the oneDNN interpolate kernel.

lidanqing-vv:
@jczaja @wojtuss  Please review

paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
```diff
@@ -0,0 +1,208 @@
...
+      this->fwd_pd_.reset(new dnnl::resampling_forward::primitive_desc(
+          resampling_d, this->engine_));
+
+      auto key_pd = this->key_ + "@fwd_pd";
+      this->dev_ctx_.SetBlob(key_pd, this->fwd_pd_);
+    }
+  }
+
+  std::shared_ptr<resampling_forward::primitive_desc>
+  AcquireForwardPrimitiveDescriptor() {
```
jczaja: 1) Why do you need to define your own AcquireForwardPrimitiveDescriptor. There is generic implementation in MKLDNNHandlerT . Why not use this base one?

2) Currently for primitivedescriptor no info on threads should be added to the key. you used key_ which (for Parallel executor, and dygraph) will contain TID which makes crash during training

paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
```diff
@@ -0,0 +1,208 @@
...
+
+    auto out_dims_vec = ComputeOutputShape(ctx);
+    framework::DDim dim_out = framework::make_ddim(out_dims_vec);
+    z->mutable_data<T>(dim_out, ctx.GetPlace());
+
+    InterpolateMKLDNNHandler<T> handler(algo, dev_ctx, mkldnn_engine,
+                                        ctx.GetPlace(), x, z,
+                                        ctx.OutputName("Out"));
+
+    auto resampling_pd = handler.AcquireForwardPrimitiveDescriptor();
```
jczaja: This is not needed here and is set (fwd_pd_) when MKLDNNHandlerT::isCached() is called.

paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
lidanqing-vv: I called AcquireForwardPrimitiveDescriptor in MKLDNNHandlerT now

PR #29131, opened 2020-11-26
Fix a bug when running on an operating system without "bash."

GeminiCarrie:
### PR types

Bug fixes

### PR changes
OPs

### Describe
replace "/bin/bash" with "/bin/sh"


chenwhql (approved):
LGTM

PR #29134, opened 2020-11-26
Remove cast from paddle.pow api

Joejiong:
### PR types

Performance optimization
### PR changes

APIs
### Describe

Performance optimization for power API and its operator.
The following code affects this version of paddle.pow api.
This block of code is meant to Align paddlepaddle pow api's functionality and behavior with numpy.power api, but it slowdowns the speed of this api call by 40%.
```
if x.dtype != y.dtype:
   y = cast(y, dtype='float64')
   x = cast(x, dtype='float64')
```
After discussion we remove the cast operation in this api.


zhhsplendid (approved):
LGTM

PR #29139, opened 2020-11-26
fix the bug of gloo hang in dygraph mode

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Describe

A http server will be started which is used by gloo initialization, but it may exit before gloo initialization. This pr ensures the http server will not exit before gloo initialization.


danleifeng (approved):
LGTM

PR #29165, opened 2020-11-27
[Dy2Stat] Fix bug: the return statement should be transformed to an equivalent Paddle/Python if statement, which depends on if conditions of the return stmt.

liym27:
### PR types

Bug fixes
### PR changes

Others
### Describe

转写 return 语句时，需转成等价的 Paddle/Python if，之前均转写成 Paddle cond，导致某些case 下组网出错（因为return 转写中创建的 Tensor变量的 shape和dtype 不满足组网要求） 。本PR 根据情况创建变量类型。

Before this PR, the `return` statement is transformed to an equivalent `paddle if` statement. But sometimes it should be transformed to `Python if`, which depends on the  `if` conditions of the `return` stmt.
For example:

- original dygraph code:
```python
def inner_func(x):
    a = 2
    if a < 0:
        y = x + 1
        return y
    y = x * 2
    return y

@to_static
def test():
    y = paddle.ones([10])

    # the shape of inner_func(y) should be [10], not [1]
    y = inner_func(y)
    y = paddle.reshape(y, [2, 5])
    return y

test()
``` 

- Return Transformed code before this PR, the type of `__return_0` is always `Paddle Tensor`
```python
def inner_func(x):
    __return_1 = paddle.fluid.layers.fill_constant(shape=[1], dtype='bool',
        value=False)
    __return_0 = paddle.fluid.layers.fill_constant(shape=[1], dtype='bool',
        value=False) 
    __return_value_init_0 = paddle.fluid.layers.fill_constant(shape=[1],
        dtype='float64', value=0.0)
    __return_value_0 = __return_value_init_0
    a = 2
    if a < 0:
        y = x + 1
        __return_0 = paddle.fluid.layers.fill_constant(shape=[1], dtype=
            'bool', value=True)
        __return_value_0 = y

   # here __return_0 is Paddle Tensor, this if stmt will be transformed to Paddle cond
    if not __return_0: 
        y = x * 2
        __return_1 = paddle.fluid.layers.fill_constant(shape=[1], dtype=
            'bool', value=True)
        __return_value_0 = y
    return __return_value_0
```
```
InvalidArgumentError: The 'shape' in ReshapeOp is invalid. The input tensor X'size must be equal to the capacity of 'shape'. But received X's shape = [1], X's size = 1, 'shape' is [2, 5], the capacity of 'shape' is 10.
  [Hint: Expected capacity == in_size, but received capacity:10 != in_size:1.] (at /home/teamcity/work/ef54dc8a5b211854/paddle/fluid/operators/reshape_op.cc:222)
  [operator < reshape2 > error]
```

- After fixed by this PR, the  type of `__return_0` is the same as ` a < 0`.
```python
def inner_func(x):
    __return_value_init_0 = paddle.fluid.layers.fill_constant(shape=[1],
        dtype='float64', value=0.0)
    __return_value_0 = __return_value_init_0
    a = 2
    __return_0 = paddle.jit.dy2static.create_bool_as_type(a < 0, False)
    if a < 0:
        y = x + 1
        __return_0 = paddle.jit.dy2static.create_bool_as_type(a < 0, True)
        __return_value_0 = y
   
    # here __return_0 is Python bool, this if stmt will still be a python if
    if not __return_0:
        y = x * 2
        __return_1 = paddle.jit.dy2static.create_bool_as_type(not
            __return_0, True)
        __return_value_0 = y
    return __return_value_0
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/return_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/return_transformer.py
@@ -20,6 +20,7 @@ from paddle.fluid import unique_name
 from paddle.fluid.dygraph.dygraph_to_static.utils import index_in_list
 from paddle.fluid.dygraph.dygraph_to_static.break_continue_transformer import ForToWhileTransformer
 from paddle.fluid.dygraph.dygraph_to_static.variable_trans_func import create_fill_constant_node
+from paddle.fluid.dygraph.dygraph_to_static.utils import ast_to_source_code
 
 __all__ = [
     'RETURN_NO_VALUE_MAGIC_NUM', 'RETURN_NO_VALUE_VAR_NAME', 'ReturnTransformer'
@@ -251,10 +252,7 @@ class ReturnTransformer(gast.NodeTransformer):
                 value=return_value_nodes)
             node.body.insert(0, assign_return_value_node)
             node.body[:0] = assign_zero_nodes
-        # Prepend control flow boolean nodes such as '__return@1 = False'
-        for name in self.return_name[node]:
-            assign_false_node = create_fill_constant_node(name, False)
-            node.body.insert(0, assign_false_node)
+
         # Prepend no value placeholders
         for name in self.return_no_value_name[node]:
             assign_no_value_node = create_fill_constant_node(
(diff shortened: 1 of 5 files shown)


zhhsplendid (approved):
LGTM

PR #29173, opened 2020-11-27
[dy2stat] Set shape for linspace to Fix dy2stat for GridGenerator Model

zhhsplendid:
### PR types

Bug fixes
### PR changes

APIs
### Describe

GridGenerator model failed because the output shape of `linspace` is (-1). The reason is that C++ InferShape fixes the shape to (-1): 

https://github.com/PaddlePaddle/Paddle/blob/5da3d514ebaa6fffd48c4a2e6bb5b16268dae92e/paddle/fluid/operators/linspace_op.cc#L49

We cannot set the shape in C++ infer shape because this Tensor may not be initialized during compile time, but when input `num` of `linspace` is an integer, we know the shape at compiler time. This PR simply set the shape in Python and add GridGenerator as unittest.

### DIFF
--- a/python/paddle/fluid/layers/tensor.py
+++ b/python/paddle/fluid/layers/tensor.py
@@ -1477,6 +1477,8 @@ def linspace(start, stop, num, dtype=None, name=None):
                 'Num': tensor_num},
         attrs={'dtype': dtype},
         outputs={'Out': [out]})
+    if isinstance(num, int):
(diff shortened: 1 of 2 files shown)


Aurelius84 (dismissed):
LGTM

liym27 (dismissed):
LGTM

liym27 (approved):

Aurelius84 (approved):

PR #29221, opened 2020-11-30
Add ReserveSpace in dygraph batch_norm.

huangxu96:
### PR types

Function optimization
### PR changes

APIs
### Describe

This PR added ReserveSpace flag for BatchNorm in dygraph model for supporting fp16 training.

### DIFF
--- a/python/paddle/fluid/dygraph/nn.py
+++ b/python/paddle/fluid/dygraph/nn.py
@@ -31,6 +31,7 @@ from ..data_feeder import check_variable_and_dtype, check_type
 import numpy as np
 import numbers
 import logging
+import os
 import paddle.utils.deprecated as deprecated
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/nn.py
```diff
@@ -1308,6 +1309,12 @@ def __init__(self,
             dtype=self._dtype)
         self._variance.stop_gradient = True
 
+        self._has_reserve_space = False
+        if data_layout == 'NHWC':
+            flag = os.environ.get('FLAGS_cudnn_batchnorm_spatial_persistent')
```
zhiqiu, line 1314: Why not use paddle.get_flags instead of os.environ.get

huangxu96: OK，I will fix it in next PR

wzzju (approved):
LGTM.

PR #29222, opened 2020-11-30
fix DATA_HOME path in win

Steffy-zxf:
### PR types

Bug fixes 

### PR changes

 APIs 

### Describe

fix DATA_HOME path in win


wawltor (approved):
LGTM

PR #29226, opened 2020-11-30
[Dy2stat] Fix PaddleGan Deoldify Model Dy2stat Problems

zhhsplendid:
### PR types

Bug fixes
### PR changes

APIs
### Describe

This PR fixes several problems in dy2stat for Deoldify model in PaddleGan.

1. In model, software engineer wrote `if x.shape == y.shape`, the Tenser shape is a tuple in dygraph so the `==` returns True/False, but in static graph the `==` becomes element-wise comparison, which is a different behavior. In this PR we reduce the element-wise comparison result.

2. If software engineer write computations which uses parameters in hooks, the static graph can loss the parameter variable because we put `param_guard` at `forward` of a `Layer`. In this PR we made `param_guard` cover pre-hook and post-hook.

3. In PaddleGan, software engineer calculated some parameter values in `__init__` by running some dygraph code. Those code also run during dy2stat. So some variables may be assign as a `VarBase` (Tensor) first and then `Variable`, which raised an error. We fixed the bug in this PR by handling the case.

TODO: We just added testcase for the 1. shape comparison. Should add test case for 2. and 3. But since we are chasing 2.0RC, I will do it in the near future PR


python/paddle/fluid/dygraph/dygraph_to_static/logical_transformer.py
```diff
@@ -18,6 +18,30 @@
...
+    if isinstance(node, gast.GtE):
+        return ">="
+    if isinstance(node, gast.Is):
+        return "is"
+    if isinstance(node, gast.IsNot):
+        return "is not"
+    if isinstance(node, gast.In):
+        return "in"
+    if isinstance(node, gast.NotIn):
+        return "not in"
```
Aurelius84: In next PR，we can consider a dict to config the relationship between `gast.node` and `cmop`, which is O(1).

Or use `if-elif-elif` to avoid execute every conditional branch.

Aurelius84 (commented):
Great!

Aurelius84 (dismissed):

python/paddle/fluid/dygraph/layers.py
```diff
@@ -865,30 +865,30 @@ def _build_once(self, *args, **kwargs):
...
-        if not self._built:
-            with program_desc_tracing_guard(False):
-                self._build_once(*inputs, **kwargs)
-                if parallel_helper._is_data_parallel_mode():
-                    parallel_helper._broadcast_parameters(
-                        self._parameters.values())
-            self._built = True
-
         with param_guard(self._parameters), param_guard(self._buffers):
+            for forward_pre_hook in self._forward_pre_hooks.values():
```
liym27, line 869: A question: Is this within expectations？
It seems like that the **parameters** in hooks can be  in `param_guard`, then the static graph can cover the parameters. But the operators in hooks may not be convered in static graph if this Layer is the outermost layer which is decorated by `paddle.jit.to_static`.

liym27 (dismissed):
LGTM

python/paddle/fluid/dygraph/layers.py
zhhsplendid: It is within expectation but you are also right. We haven't supported paddle.jit.to_static in hook now and we don't think we have enough time for 2.0RC, so this PR just moved `param_guard` without handling dy2stat for hooks. We will do that in the future.

Aurelius84 (approved):

PR #29255, opened 2020-12-01
Rebuild group automatically in dynamic graph distributed

ForFishes:
### PR types

New features
### PR changes

Others
### Describe

Add rebuild group feature in reducer.
In dynamic graph networking, if the parameter order is not declared in the forward order, performance may be affected. As shown in the figure: in the Bert model, putting the declaration of the embedding parameter of the first layer at the end will result in slower performance. The timeline is as follows:
image
Here add the automatic rebuild group strategy, in the second batch, the group ready order can be adjusted automatically, which makes the computation and communication stream overlap fully. From the experiment of 4 machines and 32 cards, the speed can be improved  about **10.6%**.

### DIFF
--- a/paddle/fluid/imperative/reducer.cc
+++ b/paddle/fluid/imperative/reducer.cc
@@ -274,12 +274,28 @@ void Reducer::FinalizeBackward() {
 std::vector<std::vector<size_t>> AssignGroupBySize(
     const std::vector<std::shared_ptr<imperative::VarBase>> &vars,
     const std::vector<bool> &is_sparse_gradient,
-    const std::vector<size_t> &group_size_limits) {
+    const std::vector<size_t> &group_size_limits,
+    const std::vector<int64_t> &tensor_indices) {
   PADDLE_ENFORCE_EQ(vars.size(), is_sparse_gradient.size(),
                     platform::errors::PreconditionNotMet(
                         "vars len must be equal to is_sparse_gradient len, but "
                         "[%lu] != [%lu]",
                         vars.size(), is_sparse_gradient.size()));
+  auto check_perm = [](const std::vector<int64_t> &x) -> bool {
+    size_t len = x.size();
+    std::vector<size_t> cnt(len, 0);
+    for (size_t i = 0; i < len; ++i) {
+      if (x[i] >= static_cast<int64_t>(len) || x[i] < 0 || cnt[x[i]]) {
+        return false;
+      }
+      cnt[x[i]]++;
+    }
+    return true;
+  };
+  PADDLE_ENFORCE_EQ(true, check_perm(tensor_indices),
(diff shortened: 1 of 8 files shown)


paddle/fluid/imperative/reducer.cc
```diff
@@ -257,10 +275,31 @@ void Reducer::MarkGroupReady(size_t group_index) {
...
+  std::reverse(rebuild_group_index.begin(), rebuild_group_index.end());
+  return rebuild_group_index;
+}
+
 void Reducer::FinalizeBackward() {
   PADDLE_ENFORCE_CUDA_SUCCESS(cudaEventRecord(comm_enent_.get(), comm_stream_));
   PADDLE_ENFORCE_CUDA_SUCCESS(
       cudaStreamWaitEvent(compute_stream_, comm_enent_.get(), 0));
+  if (!has_rebuilt_group_) {
+    auto rebuild_group_index = RebuildGruops();
```
hutuxian, line 296: const auto&?

paddle/fluid/imperative/reducer.cc
```diff
@@ -20,47 +20,98 @@ namespace imperative {
...
+      PADDLE_THROW(platform::errors::Unimplemented(
+          "Data type (%s) is not supported when it splits tensors for "
+          "allreduce.",
+          framework::DataTypeToString(dtype_)));
+  }
+}
+
+std::ostream &operator<<(std::ostream &out, const Group &group) {
+  const auto &vars = group.variable_indices_;
+  out << "numul: " << group.all_length_ << " ;is_sparse: " << group.is_sparse_
```
hutuxian, line 71: numul->numel

paddle/fluid/imperative/reducer.cc
```diff
@@ -159,17 +223,27 @@ void Reducer::InitializeGroups(
...
           .mutable_data(place_, group.dtype_);
     }
-    // Debug Message For Reducer
-    VLOG(3) << "the groups_[" << group_index << "] basic message:";
-    VLOG(3) << "numul: " << all_length << " ;is_sparse: " << group.is_sparse_
-            << " ;var number: " << group.variable_indices_.size();
+
+    // map variables to this group by VariableLocator
+    size_t inside_group_index = 0;
+    for (const auto var_index : group_indices[group_index]) {
```
hutuxian, line 235: use `variable_indices_` directly?

hutuxian (approved):
LGTM

paddle/fluid/imperative/reducer.cc
ForFishes: fix in next pr.

paddle/fluid/imperative/reducer.cc
ForFishes: fix in next pr.

PR #29267, opened 2020-12-01
[inplace] Add ShareHolderWith for class Variable and SharePlaceholderWith in VarBase.detach() to share the same Tensor/SelectedRows

liym27:
### PR types

New features
### PR changes

Others
### Describe

1. Add **`Variable::ShareHolderWith`** for class `Variable` to support two Variables share the same Placeholder whose type can be Tensor, LoDTensor, SelectedRows, LoDTensorArray, etc.

2. Use `Variable::ShareHolderWith` in VarBase.detach() to share the same Tensor/SelectedRows so that the `detach_var` can share the same `TensorInplaceVersion` with the original var.

### DIFF
--- a/paddle/fluid/framework/variable.h
+++ b/paddle/fluid/framework/variable.h
@@ -69,6 +69,16 @@ class Variable {
     return holder_->Type();
   }
 
+  /**
+   * The internal of two Variables share the same Placeholder whose type can be
+   * Tensor, LoDTensor, SelectedRows, LoDTensorArray, etc.
+   *
+   * NOTE(liym27): In dynamic mode, sharing the same Placeholder also means
+   * share the same TensorInplaceVersion, which is very important for inplace
+   * operations.
+   */
+  void SharePlaceholderWith(const Variable& var);
+
  private:
   // This method hides type T, so it doesn't appear as a template parameter of
   // Variable.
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/tests/unittests/test_detach.py
```diff
@@ -161,5 +162,47 @@ def test_detach_exception(self):
...
+            self.assertEqual(var.inplace_version, 1)
+
+            detach_var_2 = var.detach()
+            self.assertEqual(detach_var_2.inplace_version, 1)
+
+            var[0] = 3
+            self.assertEqual(detach_var_1.inplace_version, 2)
+            self.assertEqual(detach_var_2.inplace_version, 2)
+
+    def test_backward_error(self):
```
zhwesky2010, line 183: https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/test_var_base.py#L246-L250  另外一个test_detach的inplace检查需要改下

zhwesky2010: https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/pybind/imperative.cc#L772-L777 还有API文档需要改一下

liym27: 感谢建议，已修改。

paddle/fluid/framework/variable.h
```diff
@@ -113,6 +123,14 @@ class Variable {
   std::shared_ptr<Placeholder> holder_;
 };
 
+inline void Variable::SharePlaceholderWith(const Variable& var) {
```
zhhsplendid, line 126: DO NOT change in this PR because we are chasing 2.0RC1, but we should know it:

I found that variable.h doesn't split declaration and definition (i.e., .h and .cc, interface and source). In c++, it can make compilation slower, especially variable is used very common among PaddlePaddle framework.

Maybe you already know it. I mentioned it in case you don't know :-) In the future we may split it, but DO NOT change in this PR.

zhhsplendid (commented):
I have some other problems. I will discuss with you offline.

zhhsplendid (approved):
After discussed offline, I have no problem now.

LGTM

paddle/fluid/framework/variable.h
liym27: Thanks,  we may split it in the future.

PR #29272, opened 2020-12-01
[oneDNN] Making ThreadID info in caching key optional

jczaja:
### PR types
Others

### PR changes
OPs

### Describe
currently part or hashing key to oneDNN cache are ThreadID and Executor's pointer . It is not needed to have them both (unless parallel executor is used). 

**Note:**  Due to huge number of modifies, unit tests will come in next PR

### DIFF
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -181,8 +181,9 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
 
   if (in_format != out_format) {
     void* in_data = GetDataFromTensor(in, in_type);
-    const std::string key =
-        platform::CreateKey(in_tz, in_format, out_format, in_type);
+    std::string key =
+        platform::CreateKey(*dev_ctx, in_tz, in_format, out_format, in_type);
+    key = platform::ExtendKeyWithThreadingInfoIfNeeded(*dev_ctx, key);
 
     platform::ReorderMKLDNNHandler handler(in_tz, in.type(), in_type, *dev_ctx,
                                            cpu_engine, key);
--- a/paddle/fluid/framework/data_layout_transform.cc
+++ b/paddle/fluid/framework/data_layout_transform.cc
@@ -183,7 +183,6 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
     void* in_data = GetDataFromTensor(in, in_type);
     std::string key =
         platform::CreateKey(*dev_ctx, in_tz, in_format, out_format, in_type);
-    key = platform::ExtendKeyWithThreadingInfoIfNeeded(*dev_ctx, key);
 
     platform::ReorderMKLDNNHandler handler(in_tz, in.type(), in_type, *dev_ctx,
                                            cpu_engine, key);
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -183,6 +183,9 @@ void Executor::Run(const ProgramDesc& pdesc, Scope* scope, int block_id,
                    bool force_disable_gc, bool keep_kid_scopes) {
   platform::RecordBlock b(block_id);
   if (FLAGS_use_mkldnn) EnableMKLDNN(pdesc);
+#ifdef PADDLE_WITH_MKLDNN
+  platform::AttachPointerHashToMKLDNNKey(this, place_);
+#endif
   auto ctx = Prepare(pdesc, block_id, skip_ref_cnt_vars, force_disable_gc);
   RunPreparedContext(ctx.get(), scope, create_local_scope, create_vars,
                      keep_kid_scopes);
@@ -308,6 +311,9 @@ void Executor::Run(const ProgramDesc& program, Scope* scope,
                    const std::string& fetch_holder_name) {
   platform::RecordBlock b(kProgramId);
   if (FLAGS_use_mkldnn) EnableMKLDNN(program);
+#ifdef PADDLE_WITH_MKLDNN
+  platform::AttachPointerHashToMKLDNNKey(this, place_);
+#endif
   bool has_feed_ops =
       has_feed_operators(program.Block(0), *feed_targets, feed_holder_name);
   bool has_fetch_ops =
@@ -557,7 +563,6 @@ void Executor::EnableMKLDNN(const ProgramDesc& program) {
       }
     }
   }
-  platform::AttachPointerHashToMKLDNNKey(this, place_);
 #else
   LOG(WARNING)
       << "'MKLDNN' is not supported, Please re-compile with WITH_MKLDNN option";
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -183,9 +183,6 @@ void Executor::Run(const ProgramDesc& pdesc, Scope* scope, int block_id,
                    bool force_disable_gc, bool keep_kid_scopes) {
   platform::RecordBlock b(block_id);
   if (FLAGS_use_mkldnn) EnableMKLDNN(pdesc);
-#ifdef PADDLE_WITH_MKLDNN
-  platform::AttachPointerHashToMKLDNNKey(this, place_);
-#endif
   auto ctx = Prepare(pdesc, block_id, skip_ref_cnt_vars, force_disable_gc);
   RunPreparedContext(ctx.get(), scope, create_local_scope, create_vars,
                      keep_kid_scopes);
@@ -311,9 +308,6 @@ void Executor::Run(const ProgramDesc& program, Scope* scope,
                    const std::string& fetch_holder_name) {
   platform::RecordBlock b(kProgramId);
   if (FLAGS_use_mkldnn) EnableMKLDNN(program);
-#ifdef PADDLE_WITH_MKLDNN
-  platform::AttachPointerHashToMKLDNNKey(this, place_);
-#endif
   bool has_feed_ops =
       has_feed_operators(program.Block(0), *feed_targets, feed_holder_name);
   bool has_fetch_ops =
@@ -563,6 +557,7 @@ void Executor::EnableMKLDNN(const ProgramDesc& program) {
       }
     }
   }
+  platform::AttachPointerHashToMKLDNNKey(this, place_);
 #else
   LOG(WARNING)
       << "'MKLDNN' is not supported, Please re-compile with WITH_MKLDNN option";
--- a/paddle/fluid/operators/fused/mkldnn/multi_gru_mkldnn_op.cc
+++ b/paddle/fluid/operators/fused/mkldnn/multi_gru_mkldnn_op.cc
@@ -109,14 +109,8 @@ class MultiGRUHandler {
     const std::string unique_name = ctx.OutputName("Hidden");
     // Create memory key without Ti because weights, bias and h0 memories
     // do not depend on Ti size but primitive and input/output memory do
-
-    memory_key_ = CreateKey(dev_ctx, unique_name, MKLDNNGetDataType<T>());
-    if ((dev_ctx.UseThreadID() == true) &&
-        (platform::MKLDNNDeviceContext::tls().get_cur_mkldnn_session_id() ==
-         platform::MKLDNNDeviceContextThreadLocals::kMKLDNNSessionID_Default)) {
-      memory_key_ =
-          platform::ExtendKeyWithThreadingInfoIfNeeded(dev_ctx, memory_key_);
-    }
+    memory_key_ = platform::ExtendKeyWithThreadInfoIfNeeded(
+        dev_ctx, CreateKey(dev_ctx, unique_name, MKLDNNGetDataType<T>()));
     key_ = memory_key_;
     key_.append("T").append(std::to_string(Ti_));
 
--- a/paddle/fluid/operators/fused/mkldnn/multi_gru_mkldnn_op.cc
+++ b/paddle/fluid/operators/fused/mkldnn/multi_gru_mkldnn_op.cc
@@ -109,12 +109,13 @@ class MultiGRUHandler {
     const std::string unique_name = ctx.OutputName("Hidden");
     // Create memory key without Ti because weights, bias and h0 memories
     // do not depend on Ti size but primitive and input/output memory do
-    if (platform::MKLDNNDeviceContext::tls().get_cur_mkldnn_session_id() !=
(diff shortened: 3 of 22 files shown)


jczaja:
@luotao1 For some reason CI runs are not starting despite many hours in queue. Could you please advice ? (This PR is changing 22 files which may have some impact)

luotao1:
You can use lock and unlock function to trigger CI after.
image

jczaja:
@arlesniak , @arogowie-intel Please help with review

paddle/fluid/platform/device_context.h
```diff
@@ -532,6 +532,10 @@ class MKLDNNDeviceContext : public CPUDeviceContext {
   void SetKeySuffix(const std::string& suffix) { key_suffix_ = suffix; }
   const std::string& GetKeySuffix(void) const { return key_suffix_; }
 
+  // Disable adding  thread ID to the key
+  void DisableThreadInfoInKey(void) { key_attach_thread_id_ = false; };
+  bool UseThreadID(void) const { return key_attach_thread_id_; };
```
arogowie-intel: This function name suggests to me that _some action will happen_, however it won't. I understand the benefits of short function name but I'd suggest to rename it like: `DoesKeyUseThreadID` or `IsThreadIDUsedInKey` or `UseThreadIdInKey`

paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -428,32 +428,41 @@ inline void AppendKey(std::string* key, const std::vector<T>& dims) {
   }
 }
 
-inline unsigned int HashPointer(uintptr_t ptr) {
-  // Get four less meaningful digits in decimal numerals
-  return ptr % 1000;
```
arogowie-intel, line 433: Why not using some bitmask to retrieve some digits? ie: `ptr & (0xFFFF-1)`

paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -428,32 +428,41 @@ inline void AppendKey(std::string* key, const std::vector<T>& dims) {
...
+                             ArgTypes&&... args) {
   std::string key;
   key.reserve(64);
   using expand_type = int[];
   expand_type{0, (AppendKey(&key, std::forward<ArgTypes>(args)), 0)...};
+  key += dev_ctx.GetKeySuffix();
   return key;
 }
 
+inline std::string ExtendKeyWithThreadingInfoIfNeeded(
```
arogowie-intel:
```suggestion
inline std::string ExtendKeyWithThreadInfoIfNeeded(
```

paddle/fluid/framework/data_layout_transform.cc
```diff
@@ -181,8 +181,9 @@ void innerTransDataLayoutFromMKLDNN(DataLayout in_layout, DataLayout out_layout,
 
   if (in_format != out_format) {
     void* in_data = GetDataFromTensor(in, in_type);
-    const std::string key =
-        platform::CreateKey(in_tz, in_format, out_format, in_type);
+    std::string key =
+        platform::CreateKey(*dev_ctx, in_tz, in_format, out_format, in_type);
+    key = platform::ExtendKeyWithThreadingInfoIfNeeded(*dev_ctx, key);
```
arogowie-intel: This key is already extended with ThreadInfo in ` platform::ReorderMKLDNNHandler` class constructor.

paddle/fluid/framework/executor.cc
```diff
@@ -557,7 +563,6 @@ void Executor::EnableMKLDNN(const ProgramDesc& program) {
       }
     }
   }
-  platform::AttachPointerHashToMKLDNNKey(this, place_);
```
arogowie-intel: Why you have moved out this call?

paddle/fluid/platform/device_context.h
wojtuss: ... `IsUsingThreadIdInKey`, `UsesThreadIdInKey`

paddle/fluid/platform/device_context.h
```diff
@@ -553,7 +557,9 @@ class MKLDNNDeviceContext : public CPUDeviceContext {
   std::shared_ptr<BlobMap> p_blobmap_;
   std::shared_ptr<std::mutex> p_mutex_;
   bool block_next_cache_clearing_ = false;
-  std::string key_suffix_;  // Key identifying current Executor
+  std::string key_suffix_ =
+      std::string("");  // Key identifying current Executor
```
wojtuss: This is redundant.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -498,8 +498,9 @@ class ConvMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     mkldnn::memory::data_type src_dt =
         paddle::framework::ToMKLDNNDataType(input->type());
 
-    std::string key = platform::CreateKey(
-        src_tz, src_dt, ctx.InputName("Input") + ctx.InputName("Filter"));
+    std::string key =
+        platform::CreateKey(dev_ctx, src_tz, src_dt,
+                            ctx.InputName("Input") + ctx.InputName("Filter"));
```
wojtuss, line 503: It would be good to remove the `+` operation on strings here.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -941,8 +938,9 @@ class ConvMKLDNNGradOpKernel : public paddle::framework::OpKernel<T> {
     // Get an unique name from "argument" name of "input" and "Filter" variable
     // as well as attributes of primitive to be created
     // This name will be used as key when saving info into device context
-    const std::string key = platform::CreateKey(
-        src_tz, ctx.InputName("Input") + ctx.InputName("Filter"));
+    std::string key = platform::CreateKey(
+        dev_ctx, src_tz, ctx.InputName("Input") + ctx.InputName("Filter"));
```
wojtuss, line 942: It would be good to remove the + operation on strings here, too.

paddle/fluid/platform/mkldnn_helper.h
wojtuss: or maybe `ExtendKeyWithThreadIdIfNeeded`

paddle/fluid/framework/executor.cc
```diff
@@ -183,6 +183,9 @@ void Executor::Run(const ProgramDesc& pdesc, Scope* scope, int block_id,
                    bool force_disable_gc, bool keep_kid_scopes) {
   platform::RecordBlock b(block_id);
   if (FLAGS_use_mkldnn) EnableMKLDNN(pdesc);
+#ifdef PADDLE_WITH_MKLDNN
+  platform::AttachPointerHashToMKLDNNKey(this, place_);
```
arlesniak: Please guard the call with if(FLAGS_use_mkldnn)

paddle/fluid/framework/executor.cc
```diff
@@ -308,6 +311,9 @@ void Executor::Run(const ProgramDesc& program, Scope* scope,
                    const std::string& fetch_holder_name) {
   platform::RecordBlock b(kProgramId);
   if (FLAGS_use_mkldnn) EnableMKLDNN(program);
+#ifdef PADDLE_WITH_MKLDNN
+  platform::AttachPointerHashToMKLDNNKey(this, place_);
```
arlesniak: Ditto

paddle/fluid/platform/mkldnn_helper.h
jczaja: Good suggesstion, but I removed this code and pointer is used directly as part of key

paddle/fluid/framework/executor.cc
jczaja: You may have oneDNN ops executed without FLAGS_use_mkldnn. This is what happens in Unit tests for example.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: It would. But I can't remember doing this part so It has to be analysed why was it implemented that way.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: goog point . will be be analyzed in separate task

paddle/fluid/platform/device_context.h
jczaja: ok

paddle/fluid/platform/device_context.h
jczaja: ok

paddle/fluid/platform/mkldnn_helper.h
arogowie-intel: Still shorter string key would result in faster key comparison while searching through cache map.

paddle/fluid/platform/device_context.h
arogowie-intel: I like this one: `IsUsingThreadIdInKey`

paddle/fluid/platform/mkldnn_helper.h
jczaja: ok

paddle/fluid/framework/data_layout_transform.cc
jczaja: good point. Same issue in few other files. I updated all places

paddle/fluid/framework/executor.cc
jczaja: @arogowie-intel  Because it is now called independently from EnableMKLDNN. EnableMKLDNN is just setting "use_mkldnn=True" in ops. But you can set manualy this attribute without EnableMKLDNN , so there were sitatuation when oneDNN is used but not EnableMKLDNN is called.

jczaja:
@arlesniak , @arogowie-intel , @wojtuss Thanks for review!  Let's have second round of review:)  Please inspect 2,3,4 commit of this PR.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
wojtuss: There should be no difference in the result, but prformance with `,` instead of `+` will be better, unless it exceeds the reserved string buffer.

paddle/fluid/operators/fused/mkldnn/multi_gru_mkldnn_op.cc
```diff
@@ -109,13 +109,8 @@ class MultiGRUHandler {
...
     // Create memory key without Ti because weights, bias and h0 memories
     // do not depend on Ti size but primitive and input/output memory do
-    if (platform::MKLDNNDeviceContext::tls().get_cur_mkldnn_session_id() !=
-        platform::MKLDNNDeviceContextThreadLocals::kMKLDNNSessionID_Default) {
-      memory_key_ = CreateKey(unique_name, MKLDNNGetDataType<T>());
-    } else {
-      memory_key_ = CreateKey(unique_name, MKLDNNGetDataType<T>(), "-t:",
-                              platform::ThreadIDasStr());
-    }
+    memory_key_ = platform::ExtendKeyWithThreadInfoIfNeeded(
```
wojtuss, line 112: Why for some operators the `CreateKey(...)` call is inside of the `ExtendedKeyWithThreadInfoIfNeeded(...)`, and for some operators it is not?

jczaja: CreateKey is not called inside ExtendKeyWithThreadInfoIfNeeded(). Create Key is either call on initialization list of MKLDNNHandlerT , or separatly in an op . This separate part is legacy code that need to be refactored so it does use MKLDNNHandlerT based handlers and then CreateKey will not be visible.

wojtuss: OK, thank you!

lidanqing-vv:
Hi @luotao1  这个需要approve然后合入一下，我可以cherry-pick到2.0-rc1吗？

lidanqing-intel:
@luotao1  如果这个PR 不cherry-pick的话，mobilenet int8 性能有下降，这个下降在release note里可以不提及吗？就只是mobilenet INT8下降，其他性能没什么变化。只是因为mobilenet 太小了，容易受影响。

luotao1 (dismissed):
LGTM

arogowie-intel (dismissed):

raindrops2sea (dismissed):

paddle/fluid/framework/executor.cc
luotao1: > Please guard the call with if(FLAGS_use_mkldnn)

@raindrops2sea and I agree with this. Since `AttachPointerHashToMKLDNNKey` may have some cost for other models.

> You may have oneDNN ops executed without FLAGS_use_mkldnn. This is what happens in Unit tests for example.

The PR description said "Due to huge number of modifies, unit tests will come in next PR". Is the same thing?

jczaja:
@luotao1 Is there something missing to have this PR merged?

paddle/fluid/framework/executor.cc
luotao1: @jczaja Could you guard the call with if(FLAGS_use_mkldnn)?

jczaja: @luotao1 No. We cannot do it. FLAGS_use_mkldnn makes ops that are having attribute: use_mkldnn to have it set to True. But you can set those attributes manualy (unit tests are doing it) . For situation when use_mkldnn attribute is set manually (not via FLAGS_use_mkldnn) we still want to call AttachPointerHashToMKLDNNKey even that FLAGS_use_mkldnn is not set. So We cannot guard this call with if(FLAGS_use_mkldnn)

luotao1: @jczaja  Got it. But how to avoid this problem:
AttachPointerHashToMKLDNNKey may have some cost for other models which do not use mkldnn?

jczaja: @luotao1 , @raindrops2sea I analyzed it more, and actually with some other fixes (introduced earlier in this PR) changes in Executor.cc are not needed e.g. AttachPointerHashToMKLDNNKey is part of Executor::EnableMKLDNN()

jczaja:
@luotao1 , @raindrops2sea I have edited changes from executor.cc  so now AttachPointerHashToMKLDNNKey(this, place_) is part of EnableMKLDNN (last commit) . Please continue your review.

luotao1 (approved):
LGTM

raindrops2sea (approved):

luotao1:
Please cherry-pick into release/2.0

PR #29306, opened 2020-12-02
Fix bug: delete wrong check_type of paddle.concat

liym27:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Unify the old Api `paddle.fluid.layers.concat` and new 2.0 Api `paddle.concat` about `LoDTensorArray`.

Before this PR: 
`paddle.fluid.layers.concat` supports input type: list[Tensor], tuple[Tensor], Tensor( LoDTensorArray)
`paddle.concat` supports input type:  list[Tensor], tuple[Tensor]

`paddle.fluid.layers.concat` supports that input is `Tensor( LoDTensorArray)` but `paddle.concat` does not. 
This feature is supported for Dynamic-to-Static, because after transformed, the input's type of `concat` is LOD_TENSOR_ARRAY in some scenarios. Therefore, `Tensor( LoDTensorArray)` should also be supported in `paddle.concat`.

 However, it's not recommended for users to use this feature only in static mode, so in the Api document, the supported input's type of concat doesn't contain `Tensor`

### DIFF
--- a/python/paddle/fluid/layers/tensor.py
+++ b/python/paddle/fluid/layers/tensor.py
@@ -328,8 +328,8 @@ def concat(input, axis=0, name=None):
 
     if input[0].desc.type() == core.VarDesc.VarType.LOD_TENSOR_ARRAY:
         # NOTE(liym27): Don't remove this if branch!
-        # This feature is supported for Dynamic-to-Static, because after transformed, the type of inputs[0] is LOD_TENSOR_ARRAY in some scenarios.
-        # But it's not recommended for users to use this feature in static mode.
+        # This feature is supported for Dynamic-to-Static, because after transformed, the type of inputs[0]
+        # is LOD_TENSOR_ARRAY in some scenarios. And this feature can be used in static mode.
 
         assert len(input) == 1, "If the elements of 'input' in concat are Variable(LoDTensorArray), " \
                 "number of the elements must be 1, but received %s." % len(input)
--- a/python/paddle/fluid/layers/tensor.py
+++ b/python/paddle/fluid/layers/tensor.py
@@ -327,6 +327,10 @@ def concat(input, axis=0, name=None):
     out = helper.create_variable_for_type_inference(dtype=helper.input_dtype())
 
     if input[0].desc.type() == core.VarDesc.VarType.LOD_TENSOR_ARRAY:
+        # NOTE(liym27): Don't remove this if branch!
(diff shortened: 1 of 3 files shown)


TCChenlong (dismissed):
LGTM

python/paddle/fluid/layers/tensor.py
```diff
@@ -327,6 +327,10 @@ def concat(input, axis=0, name=None):
     out = helper.create_variable_for_type_inference(dtype=helper.input_dtype())
 
     if input[0].desc.type() == core.VarDesc.VarType.LOD_TENSOR_ARRAY:
+        # NOTE(liym27): Don't remove this if branch!
+        # This feature is supported for Dynamic-to-Static, because after transformed, the type of inputs[0] is LOD_TENSOR_ARRAY in some scenarios.
+        # But it's not recommended for users to use this feature in static mode.
```
zhhsplendid: Delete "not recommended for users". In my understanding, user still needs TensorArray at some time.

python/paddle/fluid/tests/unittests/test_concat_op.py
```diff
@@ -338,21 +340,44 @@ def setUp(self):
...
             if fluid.is_compiled_with_cuda() else fluid.CPUPlace()
-        self.set_program()
 
-    def set_program(self):
-        self.program = fluid.Program()
-        with fluid.program_guard(self.program):
-            input = fluid.layers.assign(self.x)
-            tensor_array = fluid.layers.create_array(dtype='float32')
-            zero = fluid.layers.fill_constant(shape=[1], value=0, dtype="int64")
+    def set_program(self, use_paddle_api):
```
zhhsplendid: `use_paddle_api` is confusing name. Maybe name it as `use_fluid_api` is better. Then `use_fluid_api = not use_paddle_api`.

I mentioned it because fluid API is also Paddle API.

liym27: Thanks, done

python/paddle/fluid/layers/tensor.py
liym27: Thanks, use "And this feature can be used in static mode." to replace " But it's not recommended for users to use this feature in static mode."

zhhsplendid (approved):
LGTM

wangchaochaohu (commented):
LGTM

TCChenlong (approved):
LGTM

wangchaochaohu (approved):

saxon-zh (approved):
LGTM

Issue #27935, opened 2020-10-14
Add command values to beginning of log and use paddle.enable_mkldnn()) to instead of FLAGS_use_mkldnn


lidanqing-vv:
Customer request:

Could you add some MKLDNN-related print log when use FLAGS_use_mkldnn?
Could you add new api like paddle.enable_static() (eg. paddle.enable_mkldnn()) to instead of FLAGS_use_mkldnn?

Dygraph training doc
[Dygraph_Training_Reproduction.txt](https://github.com/PaddlePaddle/Paddle/files/5383932/Dygraph_Training_Reproduction.txt)

luotao1:
Discussed with @OliverLPH,  in order to see whether we use the mkldnn in th log ([train_Resnet_b32_real2.log](https://github.com/PaddlePaddle/Paddle/files/5406095/train_Resnet_b32_real2.log), now we must use `export MKLDNN_VERBOSE=1`) easily
- Could you add some MKLDNN-related print log when use `FLAGS_use_mkldnn`?
- Could you add new api like `paddle.enable_static()` (eg. `paddle.enable_mkldnn()`) to instead of `FLAGS_use_mkldnn`?

OliverLPH:
@lidanqing-intel 
By following your reproduce document, here is my test results, plz have a look.

model | dataset | type | mode | acc result | comment
-- | -- | -- | -- | -- | --
Resnet_b32 | flowers | OneDNN | real | acc1: 75.10%  acc5: 93.40% | OMP=20


model | dataset | type | mode | epoch_cost(s) | RSS memory(KB) | RSS memory(MB) | comment
-- | -- | -- | -- | -- | -- | -- | --
Resnet_b32 | flowers | OneDNN | CE | 285.27 | 5152648 | 5031.88 | OMP=20
Resnet_b32 | flowers | Native | CE | 1062.77 | 3319124 | 3241.33 | OMP=20

It seems when enable mkldnn, it has rapid speed-up compared with disable it.
However, it also occpy around 2GB more memory costs.

Another thing need to notice is when disbale mkldnn, trainning process will get NaN
![image](https://user-images.githubusercontent.com/14963836/96549756-7fbe0300-12e2-11eb-88f3-ebdb3ac6f9ae.png)

lidanqing-vv:
@arlesniak  Hi, Artur, please add command values to log, and use paddle.enable_mkldnn()) to instead of FLAGS_use_mkldnn.

[...]

PR #29378, opened 2020-12-04
Added verbose oneDNN lib version

arlesniak:
### PR types

Others 
### PR changes

Others 
### Describe

Added verbose oneDNN lib version as part of customer request in [issue 27935](https://github.com/PaddlePaddle/Paddle/issues/27935)

### DIFF
--- a/paddle/fluid/platform/device_context.h
+++ b/paddle/fluid/platform/device_context.h
@@ -466,6 +466,7 @@ class MKLDNNDeviceContextThreadLocals {
 
   typedef MKLDNNDeviceContextThreadLocals self;
   struct Body {
+    bool said_once;
     size_t cur_mkldnn_session_id;
     // Current data input shape string.
     // - For fixed-shape, it's a null string in default.
@@ -485,6 +486,7 @@ class MKLDNNDeviceContextThreadLocals {
     void set_cur_input_shape_cache_capacity(int input_shape_cache_capacity);
     void set_cur_paddle_data_layout(framework::DataLayout dl);
     framework::DataLayout get_cur_paddle_data_layout(void);
+    void announce(void);
   };
   MKLDNNDeviceContextThreadLocals() = default;
   MKLDNNDeviceContextThreadLocals(const MKLDNNDeviceContextThreadLocals& c) =
--- a/paddle/fluid/platform/device_context.h
+++ b/paddle/fluid/platform/device_context.h
@@ -486,7 +486,7 @@ class MKLDNNDeviceContextThreadLocals {
     void set_cur_input_shape_cache_capacity(int input_shape_cache_capacity);
     void set_cur_paddle_data_layout(framework::DataLayout dl);
     framework::DataLayout get_cur_paddle_data_layout(void);
-    void announce(void);
+    void started_info(void);
   };
   MKLDNNDeviceContextThreadLocals() = default;
   MKLDNNDeviceContextThreadLocals(const MKLDNNDeviceContextThreadLocals& c) =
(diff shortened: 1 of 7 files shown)


paddle/fluid/platform/device_context.h
```diff
@@ -485,6 +486,7 @@ class MKLDNNDeviceContextThreadLocals {
     void set_cur_input_shape_cache_capacity(int input_shape_cache_capacity);
     void set_cur_paddle_data_layout(framework::DataLayout dl);
     framework::DataLayout get_cur_paddle_data_layout(void);
+    void announce(void);
```
arogowie-intel: Please add some doc explaining what is this function doing.

arogowie-intel (commented):
I'd try to do this the other way. IMHO adding additional variable to ` MKLDNNDeviceContextThreadLocals::Body` only to check whether some lib version information was logged once is not sufficient reason. Moreover with such solution this condition have to be checked at construction of each MKLDNN operator (which uses ~Handler[T] class ). 

I'm not aware what is the main usage and reason to print such information, but
I'd rather convert this to _helper_ function like `printDnnlLibraryVersion` and make sure it's called at place which guarantees it will be called only once. Some Executor constructor or other place?

paddle/fluid/platform/device_context.h
jczaja: @arogowie-intel  Operators can be called directly without supervision of Executor, NaiveExecutor, ParallelExecutor.. There is also dygraph with its own specifc code flow. so I'm not sure if you can make it more elegant solution and faster than this

arlesniak:
@arogowie-intel  Thank you for comment. The root of the PR is the need to tell whether training was actually run on oneDNN by inspecting logs without extra effort.
This could also be done via inspecting env variables or calling some API function from every training script but that way was rejected by customer. Also suppressing the info with VLOG levels would force customer to use extra effort i.e. special env variables to bring the information to the log. 
Having in mind dynamic and static execution modes and possibility of running ops in tests without executor, the `MKLDNNDeviceContextThreadLocals `looks like the right place. `~Handler[T]` points out that OneDNN execution path was chosen and executed.

arogowie-intel:
@arlesniak @jczaja OK, I wasn't aware that we don't need the `~Executor` object. With that information I think my comment is now irrelevant.

paddle/fluid/platform/device_context.h
```diff
@@ -485,6 +486,7 @@ class MKLDNNDeviceContextThreadLocals {
     void set_cur_input_shape_cache_capacity(int input_shape_cache_capacity);
     void set_cur_paddle_data_layout(framework::DataLayout dl);
     framework::DataLayout get_cur_paddle_data_layout(void);
+    void started_info(void);
```
arogowie-intel: I'd suggest to rename it like `log_lib_version` to name what it exactly does.

arogowie-intel (dismissed):
Just have minor suggestion for better readability. Other than that LGTM.

paddle/fluid/platform/device_context.h
arlesniak: Done

arlesniak:

luotao1:
Could you tell us how to use this verbose and paste the result?

arlesniak:
During running training scripts and tests, whenever oneDNN is to be used, the following info would be printed to stdout (timestamp is exemplary):
`I1210 09:07:51.475006 80876 device_context.cc:473] oneDNN v1.6.0`

It will be printed once per oneDNN lib initial usage in executor/process. 
In a log file it will be printed at moment (line) when the oneDNN will be actually used for the first time.
When oneDNN is enabled, common appearance will be one time per training log without any additional env variables or training script changes.

luotao1:
Please solve the conflict

luotao1 (approved):
LGTM! I test it in models/dygraph.

luotao1:
Please cherry-pick into release/2.0

PR #29387, opened 2020-12-04
support deepcopy for Layer/Tensor/Paramerbase

zhwesky2010:
### PR types

New features

### PR changes

APIs

### Describe

Support deepcopy for Layer/Tensor/Paramerbase.
===

deep copy Layer
---
```
class MLP(paddle.nn.Layer):
    def __init__(self, param_attr=None, bias_attr=None):
        super(MLP, self).__init__()

        self._fc1 = paddle.nn.Linear(784, 10)
        self._fc2 = paddle.nn.Linear(10, 10)

    def forward(self, inputs):
        y = self._fc1(inputs)
        y = self._fc2(y)
        return y

mlp = MLP()
mlp_copy = copy.deepcopy(mlp)

print(mlp.sublayers())
print(mlp_copy.sublayers())

[<paddle.nn.layer.common.Linear object at 0x7f44bdc09bf8>, <paddle.nn.layer.common.Linear object at 0x7f44bdc09d00>]
[<paddle.nn.layer.common.Linear object at 0x7f44bdc09eb8>, <paddle.nn.layer.common.Linear object at 0x7f44bdc09f10>]

```
``mlp_copy`` is completely independent from the original ``mlp`` , has same ``parameters, buffers, sublayers, forward_pre_hooks, forward_post_hooks`` value, and not share data with origin.

deep copy Tensor
---
```
x = paddle.to_tensor([2.], stop_gradient=False)
x_copy = copy.deepcopy(x)

print(x)  # 2.
print(x_copy) #2.

x[:]  =5.

print(x)  # 5.
print(x_copy) #2.
```

deep copy Parameter
---
```
linear = paddle.nn.Linear(10, 10)
param = linear.weight

param_copy = copy.deepcopy(param)

print(param)
print(param_copy)
```

``param_copy`` is completely independent from ``param`` , and can be trained independently.

### DIFF
--- a/paddle/fluid/pybind/imperative.cc
+++ b/paddle/fluid/pybind/imperative.cc
@@ -527,6 +527,13 @@ void BindImperative(py::module *m_ptr) {
       m, "VarBase", R"DOC()DOC")
       .def_static("_alive_vars", &imperative::VarBase::AliveVarNames)
       .def("__init__",
+           [](imperative::VarBase &self) {
+             std::string name =
+                 imperative::GetCurrentTracer()->GenerateUniqueName(
+                     "generated_tensor");
+             new (&self) imperative::VarBase(name);
+           })
+      .def("__init__",
            [](imperative::VarBase &self, framework::proto::VarType::Type dtype,
               const std::vector<int> &dims, const py::handle &name,
               framework::proto::VarType::Type type, bool persistable) {
@@ -1023,6 +1030,7 @@ void BindImperative(py::module *m_ptr) {
               y = x.cuda(1)
               print(y.place)        # CUDAPlace(1)
        )DOC")
+      .def("copy_", &imperative::VarBase::CopyFrom)
       .def("_copy_to",
            [](const imperative::VarBase &self, const platform::CPUPlace &place,
               bool blocking) { return self.NewVarBase(place, blocking); },
--- a/python/paddle/fluid/dygraph/varbase_patch_methods.py
+++ b/python/paddle/fluid/dygraph/varbase_patch_methods.py
@@ -263,6 +263,36 @@ def monkey_patch_varbase():
         from paddle.tensor.to_string import to_string
         return to_string(self)
 
+    def __deepcopy__(self, memo):
+        """
+        Deep copy Tensor, it will always performs Tensor copy.
+
+        Examples:
+            .. code-block:: python
+
+                import paddle
+                import copy
+                x = paddle.to_tensor(2.)
+                y = copy.deepcopy(x)
+                
+                print(x)
+                # Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
+                #        [2.])
+
+                print(y)
+                # Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
+                #        [2.])
+
+        """
+        if not self.is_leaf:
+            raise RuntimeError(
+                "Only Leaf Tensors support the deepcopy protocol at the moment, non-Leaf Tensors contains graph information that does't support deepcopy"
+            )
+        new_varbase = core.VarBase()
+        memo[id(self)] = new_varbase
+        new_varbase.copy_(self, True)
+        return new_varbase
+
     @property
     def block(self):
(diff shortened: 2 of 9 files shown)


python/paddle/fluid/dygraph/varbase_patch_methods.py
```diff
@@ -263,6 +263,36 @@ def __str__(self):
...
+                # Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
+                #        [2.])
+
+        """
+        if not self.is_leaf:
+            raise RuntimeError(
+                "Only Leaf Tensors support the deepcopy protocol at the moment, non-Leaf Tensors contains graph information that does't support deepcopy"
+            )
+        new_varbase = core.VarBase()
+        memo[id(self)] = new_varbase
```
chenwhql, line 294: this `memo` is useful?

paddle/fluid/pybind/imperative.cc
```diff
@@ -1023,6 +1030,7 @@ void BindImperative(py::module *m_ptr) {
               y = x.cuda(1)
               print(y.place)        # CUDAPlace(1)
        )DOC")
+      .def("copy_", &imperative::VarBase::CopyFrom)
```
chenwhql, line 1033: why not write `__deepcopy__` here directly?

python/paddle/fluid/dygraph/varbase_patch_methods.py
```diff
@@ -263,6 +263,36 @@ def __str__(self):
...
+                #        [2.])
+
+                print(y)
+                # Tensor(shape=[1], dtype=float32, place=CPUPlace, stop_gradient=True,
+                #        [2.])
+
+        """
+        if not self.is_leaf:
+            raise RuntimeError(
+                "Only Leaf Tensors support the deepcopy protocol at the moment, non-Leaf Tensors contains graph information that does't support deepcopy"
```
chenwhql: `Tensors` -> `Tensor`?

deepcopy is enough, what is `deep copy protocol` ?

paddle/fluid/pybind/imperative.cc
chenwhql: named `copy_`, is it will show to users?

python/paddle/fluid/dygraph/varbase_patch_methods.py
zhwesky2010: python copy.py module will call .__reduce_ex__(4) to return current object 's __new__, __dict__.  4 is protocol, also have 0,1,2,3. 
It doesn't matter, user should not pay attention to it, will be chaned to ``deepcopy``.

paddle/fluid/pybind/imperative.cc
zhwesky2010: Want to expose it to the user, user can use tensor1.copy_(tensor2) to copy another tensor to this tensor inplacely.

zhwesky2010: > why not write `__deepcopy__` here directly?

__deepcopy__(self, memo)  memo will record this copy by memo[id(self]=copy_var, need to get id(self),  if deecopy twice, a same tensor will return.  So place it in python. 
There's no other difference between the two implementations.

python/paddle/fluid/dygraph/varbase_patch_methods.py
zhwesky2010: It's important, this memo is passed into by python `copy.py module` . User can 
```
memo = {}
copy.deepcopy(x, memo)
```
Always it return a same var.

When, 
```
copy.deepcopy(x)
```
it return different var, and `copy.py module` will create a temp memo, and pass into this function.  So, Always memo exits.

python/paddle/fluid/dygraph/varbase_patch_methods.py
zhwesky2010: Done

jzhang533 (approved):
lgtm

XiaoguangHu01 (approved):
LGTM

PR #29404, opened 2020-12-06
[NVIDIA] Softmax vectorization

zlsh80826:
### PR types

Performance Optimization

### PR changes

Ops

### Describe

~8x faster by using warp reduction on softmax forward for short sequences
~4x faster by using warp reduction on softmax backward for short sequences


zhiqiu (approved):
LGTM

PR #29408, opened 2020-12-06
fix that parameters'grad has grad var

zhwesky2010:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix that parameters'grad has grad var.  It should not have grad var.

But,  Leaf Tensor `grad var` has been add `grad var` , which lead to mis-judgment.

image


ForFishes (approved):
LGTM

PR #29411, opened 2020-12-07
support clip op trt converter

cryoco:
### PR types

Function optimization
### PR changes

Others
### Describe

support clip op trt converter


shangzhizhou (approved):
LGTM

PR #29478, opened 2020-12-08
fix abs double grad unittest

ZHUI:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix abs double grad unittest


wawltor (approved):
LGTM

PR #29479, opened 2020-12-08
add fp16 check into max and avg pool

huangxu96:
### PR types

 New features
### PR changes

APIs
### Describe

Make some 2.0 API support fp16 data as input.


wzzju (approved):
LGTM.

PR #29484, opened 2020-12-08
Make gelu fp16 computing more robust

zhiqiu:
### PR types

Function optimization
### PR changes

OPs
### Describe

Make gelu fp16 computing more robust

- In order to avoid overflow, cast fp16 input to fp32 before computing and cast it back after computing
image

From the above two figures, we can see that fp16 gelu  is more robust after this PR.

- Actually, the original implementation also performs the **cast**, but it is done internally inside each operator, like `erf`, `tanh`, `operator* `.  For example,
```
template <>
HOSTDEVICE inline float16 erf(const float16& a) {
  return float16(::erff(static_cast<float>(a)));
}
```
  So, it is better to do one `cast` before computing and one after computing to avoid too many internal `cast`.  As an evidence, the AMP training speed of bert base model increase from 412 to 420 sequence/s after this PR.

For example, given `out = a+b+c if a,b,c are float16`.
The original computing needs 6 casts,
```
tmp = float16(float(a)+float(b))
out =  float16(float(tmp)+float(c))
```

The new computing needs 4 casts,
```
cast_a = float(a)
cast_b = float(b)
cast_c = float(c)
out =  float16(cast_a + cast_b + cast_c)
```

### DIFF
--- a/paddle/fluid/operators/gelu_op.h
+++ b/paddle/fluid/operators/gelu_op.h
@@ -36,10 +36,22 @@ struct GeluFunctor {
   void operator()(Device d, X x, Out out, bool approximate) const {
     if (approximate) {
       // gelu(x) = 0.5 * x * (1 + tanh(sqrt(2 / \pi) * (x + 0.044715 * x^{3})))
-      auto temp = (static_cast<T>(M_2_SQRTPI * M_SQRT1_2) *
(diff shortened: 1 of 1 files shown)


iclementine (approved):
LGTM

PR #29485, opened 2020-12-08
Fix the bug where embedding can‘t be processed correctly in reducer

ForFishes:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix the bug where embedding can‘t be processed correctly in reducer.
The sparse parameter is currently used in the following layers:
1、paddle.nn.layer.common.Embedding
2、paddle.nn.layer.common.Embedding
3、HSigmoidLoss(will be fixed in next pr)
If we do not register the layer in the reducer and the parameter is a sparse parameter, an exception will occur. As shown below:
image


hutuxian (approved):
LGTM

PR #29492, opened 2020-12-08
New UT should not exceed 15s

XieYunshen:
### PR types

Others
### PR changes

Others
### Describe

added UT should not exceed 15s


chalsliu (approved):
VERY GOOD

PR #29500, opened 2020-12-08
fix p_norm with empty shape

ZHUI:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix p_norm with empty shape


wawltor (approved):
LGTM

PR #29529, opened 2020-12-09
[New feature][Bug fix] add paddle C library workflow and fix bug

lanking520:
## Description
Add a template C build workflow for PaddlePaddle on Mac/Linux/Windows. Everything passed check on DJL C API integration.

You can also enable docker build for the future on Cent OS 7 or Ubuntu 14.04.

Also fix one bug in Paddle C API. The `model_dir` will be empty if user try to load a binded model (e.g "__model__", "__params__"). The code should fall back to use `prog_file` instead to keep a non-empty key.

This problem will appear if user try to load two models at the same time.

### DIFF
--- a/.github/workflows/c_release_paddle.yml
+++ b/.github/workflows/c_release_paddle.yml
@@ -1,0 +1,76 @@
+name: C package Release Paddle
+
+on:
+  workflow_dispatch:
+
+jobs:
+  windows-cpu:
+    runs-on: windows-2016
+    steps:
+      - uses: actions/checkout@v2
+      - run: git submodule update --init --recursive
+      - name: Set up Python3
+        uses: actions/setup-python@v1
+        with:
+          python-version: '3.6'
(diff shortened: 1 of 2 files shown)


Shixiaowei02:
> ## Description
> Add a template C build workflow for PaddlePaddle on Mac/Linux/Windows. Everything passed check on DJL C API integration.
> 
> You can also enable docker build for the future on Cent OS 7 or Ubuntu 14.04.
> 
> Also fix one bug in Paddle C API. The `model_dir` will be empty if user try to load a binded model (e.g "**model**", "**params**"). The code should fall back to use `prog_file` instead to keep a non-empty key.
> 
> This problem will appear if user try to load two models at the same time.

Windows 单元测试挂了，我试着修一下。

lanking520:
@Shixiaowei02 Interestingly, the modification on the Inference call should have no impact to the PD_Tensor, is that test flaky?

PR #29540, opened 2020-12-10
Fix Layer train eval setting failed in static mode

chenwhql:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix Layer train eval setting failed in static mode

By our design, `paddlle.nn.Layer` also can be used in static graph mode, but `Layer.train()` and `Layer.eval()` can only be set in dygraph mode, if set in static mode, it willl throw like:
```
  File "train_mode.py", line 5, in <module>
    l.train()
  File "/usr/local/python2.7.15/lib/python2.7/site-packages/paddle/fluid/dygraph/layers.py", line 135, in train
    framework._dygraph_tracer().train_mode()
AttributeError: 'NoneType' object has no attribute 'train_mode'
```

related issue: https://github.com/PaddlePaddle/Paddle/issues/29534

This may cause some problem. For example, the behavior of `paddle.nn.Dropout`in `Layer.train()` and `Layer.eval()` are diffferent, the 2.0 `paddle.nn.Dropout` implement as:

```
def forward(self, input):
        out = F.dropout(
            input,
            p=self.p,
            axis=self.axis,
            training=self.training,
            mode=self.mode,
            name=self.name)
        return out
```

It need `self.training` as argument of original `dropout`, if user want use `paddle.nn.Dropout` for test in static mode, it need to set `Layer.eval()` firstly.

So this PR fix this problem, only set `framework._dygraph_tracer().train_mode()` in dygraph mode.

### DIFF
--- a/python/paddle/fluid/dygraph/layers.py
+++ b/python/paddle/fluid/dygraph/layers.py
@@ -132,8 +132,11 @@ class Layer(core.Layer):
                 out = mylayer(x)
 
         """
-        # global setting
-        framework._dygraph_tracer().train_mode()
+        # global setting in dygraph
+        # NOTE(chenweihang): nn.Layer also can be used in static mode,
+        # but _dygraph_tracer() can not be called in static mode
+        if in_dygraph_mode():
+            framework._dygraph_tracer().train_mode()
         # Layer-level setting
         self.training = True
         for layer in self.sublayers():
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/dygraph/layers.py
```diff
@@ -132,8 +132,11 @@ def forward(self, input):
                 out = mylayer(x)
 
         """
-        # global setting
-        framework._dygraph_tracer().train_mode()
+        # global setting in dygraph
+        # NOTE(chenweihang): nn.Layer also can be used in static mode,
+        # but _dygraph_tracer() can not be called in static mode
+        if in_dygraph_mode():
+            framework._dygraph_tracer().train_mode()
```
zhiqiu, line 139: Actually, _dygraph_tracer().train_mode() can be removed.

chenwhql: agree, and need to revert `dropout(is_test=None)` to `dropout(is_test=False)` changed in https://github.com/PaddlePaddle/Paddle/pull/29064

chenwhql: for compatibility, removing it in next PR is better, this PR need to be chery-pick into 2.0 to fix this bug

zhiqiu (approved):

PR #29541, opened 2020-12-10
[Dy2Stat] Support grammar: for ele in var[idx]

liym27:
### PR types

Bug fixes 
### PR changes

Others
#### Support to transform`for ele in var` stms in which var is a slice of Tensor.

For example：
```python
def for_iter_var_idx(x_array):
    z = fluid.layers.fill_constant([1], 'int32', 0)
    x_array = fluid.dygraph.to_variable(x_array)

    for x in x_array[0:]:
        z = z + x
    return z
```

- Before this PR, `for_iter_var_idx` can't be transformed successfully. 
```
AttributeError: 'Subscript' object has no attribute 'args'
```


zhhsplendid (approved):
LGTM

PR #29543, opened 2020-12-10
Fix Windows Unittest

zhhsplendid:
### PR types

Others
### PR changes

Others
### Describe

Fix 3 Windows Unittests

test_fuse_all_reduce_pass: Paddle cannot run multiple-GPU on Windows so set single visible GPU flag
test_feed_data_check_shape_type: Paddle cannot run multiple-GPU on Windows so set single visible GPU flag
test_tsm: Winodws GPU size is not enough so decrease batch size and data size.


zhwesky2010 (approved):
LGTM

chalsliu (approved):

PR #29551, opened 2020-12-10
Fix the dowanload bug in the case of multiple machines

ForFishes:
### PR types

Bug fixes
### PR changes

APIs
### Describe

In the case of multiple machines, except for the first machine, other machines cannot identify the downloaded data according to local_rank. Here, the data download of multiple machines is handled by filtering by ip address.

### DIFF
--- a/python/paddle/utils/download.py
+++ b/python/paddle/utils/download.py
@@ -177,7 +177,7 @@ def get_path_from_url(url, root_dir, md5sum=None, check_exist=True):
     # Mainly used to solve the problem of downloading data from different 
     # machines in the case of multiple machines. Different ips will download 
     # data, and the same ip will only download data once.
-    unique_endpoints = _get_unique_endpoints(ParallelEnv.trainer_endpoints)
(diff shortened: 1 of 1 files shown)


hutuxian (approved):
LGTM

python/paddle/utils/download.py
```diff
@@ -140,6 +140,21 @@ def _map_path(url, root_dir):
     return osp.join(root_dir, fpath)
 
 
+def _get_unique_endpoints(trainer_endpoints):
+    # Sorting is to avoid different environmental variables for each card
```
hutuxian, line 144: Move the comment in line 179~181 to here.
This function is mainly used to solve the problem of downloading data from multiple nodes. We select the process with smallest port number of each node to download data, which ensures that only one process of each node downloads data.

chalsliu (approved):

python/paddle/utils/download.py
ForFishes: ok，fix in next pr.

PR #29556, opened 2020-12-10
Polish hash function of executor cache key 

Aurelius84:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Polish hash function of executor cache key 

**BUG描述：**

在CI执行部分动转静单测时，存在一定几率的随机挂，retry之后正常。

**原因：**

之前使用了 `ProgramDesc *`指针作为了 executor_cache内在unordered_map的 **Key**。可能存在两个单测用例中的ProgramDesc指针地址一样的情况，导致执行Program A时，命中了Program B的缓存Context，因此报错。

**修复：**
1. 在hash函数计算中，增加更多的Program字段以避免hash冲突。
    + 由于hash函数经常被调用，考虑性能问题，此处并没有使用program->Proto的序列化string作为字段。频繁序列化的overhead不容忽视
    + 采用了 program_desc + block_desc + op_size (每个block_desc)作为hash计算字段。

2. 移除了指针作为map的key，选择hash_value作为key，保持外层接口 `std::pair<program_desc*, bool>`不变
    + 因为program_desc指针指向物会在单元用例执行完后被析构，若以指针作为key，则存在通过key访问已析构指向物的潜在风险
    +  移除指针作为map的key，禁止通过key访问到program_desc*指针

### DIFF
--- a/paddle/fluid/framework/executor_cache.h
+++ b/paddle/fluid/framework/executor_cache.h
@@ -34,12 +34,13 @@ class ExecutorInfoCache {
    * The ExecutorPrepareContext is different while running forward program and
    * backward program. We add bool value into cached key to distinguish this.
    */
-  using KeyType = std::pair<const framework::ProgramDesc*, /*is_grad*/ bool>;
+  using KeyInfo = std::pair<const framework::ProgramDesc*, /*is_grad*/ bool>;
+  using KeyType = size_t;
 
   struct HashPair {
-    size_t operator()(const KeyType& key) const noexcept {
+    size_t operator()(const KeyInfo& key) const noexcept {
       size_t seed = 10;
-      auto& prog_desc = key.first;
+      auto* prog_desc = key.first;
       /*
        * Note(Aurelius84): DO NOT use only ProgramDesc* to calculate hash value
        * because a new program will hold same pointer address after an older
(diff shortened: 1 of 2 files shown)


paddle/fluid/framework/executor_cache.h
```diff
@@ -34,16 +34,29 @@ class ExecutorInfoCache {
...
+      auto* prog_desc = key.first;
+      /*
+       * Note(Aurelius84): DO NOT use only ProgramDesc* to calculate hash value
+       * because a new program will hold same pointer address after an older
+       * program is destructed with a small probability. Add op size while
+       * hashing because program may contains at least one block.
+       */
+      hash_combine(&seed, prog_desc);
+      for (size_t i = 0; i < prog_desc->Size(); ++i) {
+        hash_combine(&seed, &prog_desc->Block(i));
```
liym27, line 52: `hash_combine(&seed, &prog_desc->Block(i));`
It seems to use a **pointer** of a block_desc but not a BlockDesc object to compute hash value. Is this as expected ?

Aurelius84: Yes, ` prog_desc->Block(i)` returns BlockDesc, it also hold a Proto::Message.  It's not recommended to hash a Message object from Community discussion.

> It's not a very good idea to use proto message as keys because they are not true value types. With the presence of unknown fields, two equal messages in one binary might be treated as different by another, and therefore two different binaries might get two different unordered_map<> even given the exactly same input data (if the hash function is implemented using MessageDifferencer). It's probably better to write your own hash function that has your desired behavior rather than replying on a generic one.

And `Block(i)` is creted by `new`，it will not lead to hash conflict while considering both address of `ProgramDesc` and its `BlockDesc` (s) and inner `Op_size`.

zhhsplendid (approved):
LGTM while discussed TODO.

chalsliu (approved):

PR #29559, opened 2020-12-11
Fix the bug in gloo initialization

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix the bug in gloo initialization by using the raw function in GLOO.


wangxicoding (approved):
LGTM

PR #29579, opened 2020-12-14
[Dy2stat] Enable jit.save to Save Without Running

zhhsplendid:
### PR types

New features
### PR changes

APIs
### Describe

Enable jit.save to Save Without Running.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
@@ -473,8 +473,9 @@ class StaticFunction(object):
             if input_spec is None:
                 input_spec = self._function_spec.input_spec
             elif self._function_spec.input_spec is not None:
-                if not input_specs_compatible(input_spec,
-                                              self._function_spec.input_spec):
+                if not input_specs_compatible(
+                        flatten(input_spec),
+                        flatten(self._function_spec.input_spec)):
                     raise ValueError(
                         "The `input_spec`: {} used to construct concrete_program is conflict with the `input_spec`: {} in `@paddle.jit.to_static`".
                         format(input_spec, self._function_spec.input_spec))
--- a/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
@@ -40,6 +40,7 @@ from paddle.fluid.dygraph.dygraph_to_static.partial_program import partial_progr
 from paddle.fluid.dygraph.dygraph_to_static.utils import ast_to_func
 from paddle.fluid.dygraph.dygraph_to_static.utils import ast_to_source_code
 from paddle.fluid.dygraph.dygraph_to_static.utils import func_to_source_code
+from paddle.fluid.dygraph.dygraph_to_static.utils import input_specs_compatible
 from paddle.fluid.dygraph.dygraph_to_static.utils import type_name
 from paddle.fluid.dygraph.dygraph_to_static.utils import unwrap
 from paddle.fluid.dygraph.dygraph_to_static.utils import make_hashable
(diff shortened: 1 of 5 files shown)


python/paddle/fluid/tests/unittests/test_jit_save_load.py
```diff
@@ -918,6 +919,50 @@ def forward(self, x):
...
+                paddle.static.InputSpec(
+                    shape=[None, IMAGE_SIZE], dtype='float32')
+            ])
+
+        result_00 = layer_save(inps0)
+        result_01 = layer_save(inps1)
+        #load and save without running
+        with unique_name.guard():
+            layer_load = paddle.jit.load(model_path)
+            print(type(layer_load))
```
zhhsplendid: This is a debug print, I will remove in a coming commit.

hbwx24:
LGTM

hbwx24 (dismissed):
LGTM

python/paddle/fluid/tests/unittests/test_jit_save_load.py
```diff
@@ -986,5 +1030,105 @@ def test_jit_save_data_parallel_with_to_static(self):
...
+class TestInputSpecCompatibility(unittest.TestCase):
+    def _assert_input_spec_layer_return(self, expect_layer, test_layer):
+        input_x = paddle.uniform([8, 8], dtype='float32')
+        input_y = paddle.uniform([8, 1], dtype='float64')
+        expected_result = expect_layer(input_x, input_y)
+        test_result = test_layer(input_x, input_y)
+        self.assertTrue(
+            float((expected_result[0] - test_result[0]).abs().max()) < 1e-5)
+        self.assertTrue(
+            float((expected_result[1] - test_result[1]).abs().max()) < 1e-5)
```
Aurelius84: A small question: Why not to use `np.array_equal` or `np.allclose` here? They will check both `rtol` and `atol`.

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
```diff
@@ -450,13 +451,35 @@ def foo(x, y):
...
         cached_program_len = len(self._program_cache)
         # If specific `input_spec`, apply convertion from dygraph layers into static Program.
         if cached_program_len == 0:
-            input_spec = self._function_spec.input_spec
-            has_input_spec = (input_spec is not None and len(input_spec) > 0)
+            if input_spec is None:
+                input_spec = self._function_spec.input_spec
+            elif self._function_spec.input_spec is not None:
+                if not input_specs_compatible(input_spec,
+                                              self._function_spec.input_spec):
```
Aurelius84: Should we add `flatten(input_spec)` and `flatten(sellf._function_spec.input_spec)` here to make sure all them are pure list? Because we support nested input spec structure.

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
```diff
@@ -1195,3 +1196,39 @@ def _is_wrapped(f):
...
+        other_input_spec (list[InputSpec]|tuple(InputSpec)): list/tuple of
+            paddle.static.InputSpec
+    """
+    len_specs = len(src_input_specs)
+    if len_specs != len(other_input_specs):
+        return False
+
+    for i in range(len_specs):
+        src_shape = src_input_specs[i].shape
+        other_shape = other_input_specs[i].shape
```
liym27, line 1217: 1. `src_input_specs[i]` may be not a `InputSpec` but a list or dict, etc.  `src_input_specs[i].shape` may raise an exception.  Better to add  code like  `if isinstance(rc_input_specs[i], InputSpec)` to make sure it's type is `InputSpec`

2. Does the `name` in InputSpec need to check? for example, the name is `x`
```python
InputSpec(shape=[None, 8], dtype='float32', name='x')
```

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
```diff
@@ -1195,3 +1196,39 @@ def _is_wrapped(f):
...
+    """
+    Returns True if the two input specs are compatible, otherwise False.
+
+    args:
+        src_input_spec (list[InputSpec]|tuple(InputSpec)): list/tuple of
+            paddle.static.InputSpec
+        other_input_spec (list[InputSpec]|tuple(InputSpec)): list/tuple of
+            paddle.static.InputSpec
+    """
+    len_specs = len(src_input_specs)
```
liym27, line 1211: Maybe it's better to check that whether `src_input_spec`  and `other_input_specs ` have the same nested structure first. Because the items in src_input_specs may be InputSpec, dict, list, etc.

python/paddle/fluid/tests/unittests/test_jit_save_load.py
zhhsplendid: This was just copied from tests in same file, I have changed this test case.

python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
zhhsplendid: Good suggestion, done!

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
zhhsplendid: Good! I have followed @Aurelius84 's suggestion to add `flatten` for inputs to make sure those inputs are list of InputSpecs.

python/paddle/fluid/dygraph/dygraph_to_static/utils.py
zhhsplendid: 1. No, in this case users can debug the input type is incorrect. By the way, if we add something like `assert isinstance(src_input_specs[i], InputSpec)`, I have to import InputSpec at utils.py which can cause recursive import error. This may due to our current Python import dependency organization is not very good. We may change it in the future.

You can see now we have 1200+ lines in `utils.py`, this is a large utils file now. We may consider to split it in the future and handle the import dependency better.

2. No, the name doesn't check. Our input spec checks arg names when it is `dict` whose keys are arg names.

Aurelius84 (approved):
LGTM

liym27 (approved):
LGTM

PR #29784, opened 2020-12-21
Optimization grad merge performance

wangxicoding:
### PR types

Performance optimization
### PR changes

APIs
### Describe

Optimization grad merge performance when distributed.
#### 1、Communication optimization
Before this PR，gradient merge communicates every step.
image
This PR, we optimized it, only communicate after grad merge.
image
#### 2、Implementation
1. In GradientMergeOptimizer, we remove the op_role_var attribute of the original grad op.
2. Then we create gradient_merge var and add grad to it, that is merge_grad += grad.
     And we add op_role_var attr and `grad_merge_cond` attr to this add op.
3.  In PE, we add grad_merge_all_reduce_op_handle, which will executed allreduce when `grad_merge_cond` is True.
4. Add apply_gradient to control flow, which will be executed when `cond` is True.
#### 3、Other changes
1. control flow
  At present, when apply_gradient is in the control flow, there must have a grad_block. See https://github.com/PaddlePaddle/Paddle/blob/ca1185d06ba6c80792d60ab3f1ae47c4f3fcf73d/python/paddle/fluid/optimizer.py#L604-L605
  However, in GradientMergeOptimizer, only the optimizer state is in the control flow, only one block is needed. Therefore, we set the forward and grad block of this control flow block to itself.
  And change control_flow.py, we don't need to append conditional_block_grad when block's grad_block is itself.
2. Fix the bug that NAN will appear if the gradient is persistable when fuse_all_reduce is open.
- The specific reasons for NAN
  When fuse_all_reduce is open, the gradient will be coalesced into a large fused_var, and the gradient will share the memory with fused_var.
  However, in scope_buffered_ssa_graph_executor.cc, every num_iteration_per_drop_scope step, DropScope() will be executed to clean up non persistable or temporary variables.
  At this time, fused_var will also be cleaned up. When init again, coalesce_tensor op will be executed, new memory will be allocated to fused_var, and the original persistable grad will be erased, NAN appear.
- Repair method
  When coalescing gradients, if one of the grads is persistable, set the fused_var to be persistable, and add the `persist_output` attr to coalesce_tensor op, which will copy the value of the original persistable grad to fused_var.
#### 4、Performance & Accuracy
- Performance:

| bert_base128_fp32_bs32 | normal(seq/s) | GM k_steps=64 (seq/s) | improve |
| -- | -- | -- | -- |
| 32gpus | 3379.9 | 4601.32 | 36.1% |

- **Gradient merge and large batch equivalence Test**
See https://github.com/wangxicoding/PaddleTest/tree/master/gradient_merge

|  | sgd | adam | adagrad
| --|-- | -- | --
| baseline | -0.03897955 | -0.06657485 | -0.0089463
| gradient merge | -0.03897955 | -0.06657486 | -0.0089463

The test shows that the gradient merge and the big batch are mathematically equivalent; it also shows that there is no problem in the multi-card implementation of this PR.

- Resnet accuracy tests
resnet 50. reshaped imagenet, dali, fp32, NCHW, momentum, lr=0.1/256bs
1. Baseline. 32 cards, batch_size 128. `acc1=75.4%`
2. gradient merge, k_steps=4

| batchsize/card | 32 * 4 | 64 * 4 | 128 * 4 |
| -- | -- | -- | -- |
|8cards |  | 75.6 | 75.1 |
|16cards | 75.7 | 75.5 | 74.9 |
|32cards | 75.3 |`no cvg` | `no cvg` |

When there is batch_norm, gradient merge and large batch are not completely equivalent in mathematics, but the convergence problem of gradient merge is not big from the point of accuracy.
The two convergence failures in the table may be more affected by the convergence problem of the large batch, maybe need lars optimizer and tuning when large batchsize.

### DIFF
--- a/paddle/fluid/framework/details/CMakeLists.txt
+++ b/paddle/fluid/framework/details/CMakeLists.txt
@@ -28,6 +28,8 @@ if(WITH_GPU)
             dynload_cuda variable_visitor)
     nv_library(fused_all_reduce_op_handle SRCS fused_all_reduce_op_handle.cc DEPS op_handle_base scope lod_tensor ddim memory
             dynload_cuda variable_visitor place device_memory_aligment)
+    nv_library(grad_merge_all_reduce_op_handle SRCS grad_merge_all_reduce_op_handle.cc DEPS op_handle_base scope lod_tensor
+            ddim memory dynload_cuda variable_visitor place device_memory_aligment all_reduce_op_handle fused_all_reduce_op_handle)
 
     if(WITH_DGC)
         nv_library(sparse_all_reduce_op_handle SRCS sparse_all_reduce_op_handle.cc DEPS op_handle_base scope 
@@ -50,6 +52,8 @@ else()
              variable_visitor)
     cc_library(fused_all_reduce_op_handle SRCS fused_all_reduce_op_handle.cc DEPS op_handle_base scope lod_tensor ddim memory
             variable_visitor place device_memory_aligment)
+    cc_library(grad_merge_all_reduce_op_handle SRCS grad_merge_all_reduce_op_handle.cc DEPS op_handle_base scope lod_tensor
+            ddim memory variable_visitor place device_memory_aligment all_reduce_op_handle fused_all_reduce_op_handle)
     if(WITH_DISTRIBUTE)
         cc_library(reduce_op_handle SRCS reduce_op_handle.cc DEPS op_handle_base variable_visitor scope
                 ddim selected_rows_functor)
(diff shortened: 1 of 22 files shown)


gongweibao (commented):
需要对比大batch_size下的resnet的收敛性。

wangxicoding:
> 需要对比大batch_size下的resnet的收敛性。

已添加测试
1、无batch_norm下，gradient merge和大batch在数学上是等价的，详见测试https://github.com/wangxicoding/PaddleTest/tree/master/gradient_merge 
2、resnet 大batch_size已测试，精度上大batch_size偏低，可能和batch norm以及大batch本身收敛有关系，需要使用lars以及mlperf里面的一些调参手段

gongweibao (approved):
LGTM

mapingshuo (approved):
LGTM

XieYunshen (approved):

raindrops2sea (approved):

PR #29819, opened 2020-12-22
Optimize sparse allreduce using allgather

ForFishes:
### PR types

Function optimization
### PR changes

Others
### Describe

Optimize sparse allreduce using allgather


hutuxian (approved):
LGTM

PR #29830, opened 2020-12-22
fix TransferInplaceBack

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

remove TransferInplaceBack

There is a bug in the original `TransferInplaceVarsBack` when data transform happens with inplace.
For example:
```
paddle.enable_static()
with paddle.fluid.device_guard("gpu:0"):
    x = paddle.fluid.layers.fill_constant([1], "float32", 0)
with paddle.fluid.device_guard("cpu"):
    x = paddle.increment(x)
exe = paddle.static.Executor(paddle.CUDAPlace(0))
a, = exe.run(paddle.static.default_main_program(), fetch_list=[x])  # should be 1, but got 0
```

The reason is that, before `data transorm`, variable `x` is in global_scope and its value is 0. The input and output of increment op is x.
- before transform: 
global_scope:   `x(val:0)`
`x = increment(x)`

- after transform (transform only changes the input vars of an operator)
global_scope:   `x(val:0)`
transfer_scope:   `x'(val:0) ` 
`x =  increment(x') ` # x and x' is actually not inplaced.

The `increment` op takes `x'(val:0)`  as input and `x(val:0)` as output. 
After execution, `x(val:0)` will be `x(val:1)`, which means `x(val:1) =  x'(val:0)  + 1` .

There are two problems:

- Since `x` and `x'` may be on different places, the execution may fail.
- In original implementation, `TransferInplaceBack` is called to copy `x'(val:0)` to `x(val:1)` after execution, which overwrite the data of `x` incorrectly.

So, the right way is to change the input and output of an operator if `inplace` and `data transform` happens together, and do `TransferInplaceBack` after execution.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_increment.py
+++ b/python/paddle/fluid/tests/unittests/test_increment.py
@@ -40,5 +40,19 @@ class TestIncrement(unittest.TestCase):
             self.assertEqual((output.numpy() == expected_result).all(), True)
 
 
+class TestInplaceApiWithDataTransform(unittest.TestCase):
+    def test_increment(self):
+        if fluid.core.is_compiled_with_cuda():
+            paddle.enable_static()
+            with paddle.fluid.device_guard("gpu:0"):
+                x = paddle.fluid.layers.fill_constant([1], "float32", 0)
+            with paddle.fluid.device_guard("cpu"):
+                x = paddle.increment(x)
+            exe = paddle.static.Executor(paddle.CUDAPlace(0))
+            a, = exe.run(paddle.static.default_main_program(), fetch_list=[x])
+            paddle.disable_static()
+            self.assertEqual(a[0], 1)
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1336,12 +1336,6 @@ Scope* OperatorWithKernel::PrepareData(
         continue;
       }
 
-      auto out_var_names = OutputVars(true);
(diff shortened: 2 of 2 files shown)


python/paddle/fluid/tests/unittests/test_increment.py
```diff
@@ -40,5 +40,19 @@ def test_api(self):
...
 
+class TestInplaceApiWithDataTransform(unittest.TestCase):
+    def test_increment(self):
+        if fluid.core.is_compiled_with_cuda():
+            paddle.enable_static()
+            with paddle.fluid.device_guard("gpu:0"):
+                x = paddle.fluid.layers.fill_constant([1], "float32", 0)
+            with paddle.fluid.device_guard("cpu"):
+                x = paddle.increment(x)
+            exe = paddle.static.Executor(paddle.CUDAPlace(0))
```
QingshuChen, line 51: 需不需要在加一个op，比如cast或者其他使用率increment值，用新op的结果验证是否正确？
另外，inplace原来是解决什么问题的？如何验证原来的逻辑没有问题？

zhiqiu: > 需不需要在加一个op，比如cast或者其他使用率increment值，用新op的结果验证是否正确？

这个单测中，increment 后面有 fetch op，会用到increment的值，后面也会验证结果。

> 另外，inplace原来是解决什么问题的？如何验证原来的逻辑没有问题？

原来的逻辑有bug，这个PR修复。

chenwhql (approved):
LGTM

PR #29840, opened 2020-12-22
More precise mkldnn kernel rules in GetExpectedKernelType

arlesniak:
### PR types

Others
### PR changes

Others
### Describe

More precise mkldnn kernel choice in GetExpectedKernelType based also on kernel's registered data type

### DIFF
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1044,15 +1044,11 @@ bool OperatorWithKernel::SupportsMKLDNN(
     const proto::VarType::Type data_type) const {
   auto& op_kernels = OperatorWithKernel::AllOpKernels().at(type_);
   return std::any_of(op_kernels.begin(), op_kernels.end(),
-                     [](OpKernelMap::const_reference kern_pair) {
-                       //                     [data_type](OpKernelMap::const_reference
-                       //                     kern_pair) {
+                     [data_type](OpKernelMap::const_reference kern_pair) {
                        return platform::is_cpu_place(kern_pair.first.place_) &&
                               kern_pair.first.library_type_ ==
-                                  LibraryType::kMKLDNN;
-                       //                                  &&
-                       //                                  kern_pair.first.data_type_
-                       //                                  == data_type;
+                                  LibraryType::kMKLDNN &&
+                              kern_pair.first.data_type_ == data_type;
                      });
 }
 
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1040,11 +1040,11 @@ static void CheckTensorNANOrInf(const std::string& op_type,
                               op_type, name));
 }
 
-bool OperatorWithKernel::SupportsMKLDNN(const framework::ExecutionContext& ctx,
-                                        proto::VarType::Type data_type) const {
+bool OperatorWithKernel::SupportsMKLDNN(
+    const proto::VarType::Type data_type) const {
   auto& op_kernels = OperatorWithKernel::AllOpKernels().at(type_);
   return std::any_of(op_kernels.begin(), op_kernels.end(),
-                     [ctx, data_type](OpKernelMap::const_reference kern_pair) {
+                     [data_type](OpKernelMap::const_reference kern_pair) {
                        return platform::is_cpu_place(kern_pair.first.place_) &&
                               kern_pair.first.library_type_ ==
                                   LibraryType::kMKLDNN &&
@@ -1056,7 +1056,7 @@ bool OperatorWithKernel::CanMKLDNNBeUsed(const framework::ExecutionContext& ctx,
                                          proto::VarType::Type data_type) const {
   bool use_mkldnn_ctx =
       ctx.Attr<bool>("use_mkldnn") && platform::is_cpu_place(ctx.GetPlace());
-  return use_mkldnn_ctx && this->SupportsMKLDNN(ctx, data_type);
+  return use_mkldnn_ctx && this->SupportsMKLDNN(data_type);
 }
 
 void OperatorWithKernel::RuntimeInferShape(const Scope& scope,
--- a/paddle/fluid/framework/operator.cc
+++ b/paddle/fluid/framework/operator.cc
@@ -1044,11 +1044,15 @@ bool OperatorWithKernel::SupportsMKLDNN(
     const proto::VarType::Type data_type) const {
   auto& op_kernels = OperatorWithKernel::AllOpKernels().at(type_);
   return std::any_of(op_kernels.begin(), op_kernels.end(),
-                     [data_type](OpKernelMap::const_reference kern_pair) {
+                     [](OpKernelMap::const_reference kern_pair) {
+                       //                     [data_type](OpKernelMap::const_reference
+                       //                     kern_pair) {
                        return platform::is_cpu_place(kern_pair.first.place_) &&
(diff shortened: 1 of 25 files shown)


paddle/fluid/framework/operator.cc
```diff
@@ -1007,21 +1007,23 @@ static void CheckTensorNANOrInf(const std::string& op_type,
                               op_type, name));
 }
 
-bool OperatorWithKernel::SupportsMKLDNN() const {
+bool OperatorWithKernel::SupportsMKLDNN(const framework::ExecutionContext& ctx,
+                                        proto::VarType::Type data_type) const {
   auto& op_kernels = OperatorWithKernel::AllOpKernels().at(type_);
   return std::any_of(op_kernels.begin(), op_kernels.end(),
-                     [](OpKernelMap::const_reference kern_pair) {
+                     [ctx, data_type](OpKernelMap::const_reference kern_pair) {
```
arogowie-intel: The catched variable `ctx` is not used in lambda body.

paddle/fluid/framework/operator.cc
```diff
@@ -1007,21 +1007,23 @@ static void CheckTensorNANOrInf(const std::string& op_type,
                               op_type, name));
 }
 
-bool OperatorWithKernel::SupportsMKLDNN() const {
+bool OperatorWithKernel::SupportsMKLDNN(const framework::ExecutionContext& ctx,
```
arogowie-intel: The `framework::ExecutionContext& ctx` is not used in this function. Is it really needed?

arlesniak:
@luotao1 PR-CI-OP-benchmark fails with this PR, however there are no performance changes inside. I restarted it several times but every time it gives another list of errors i.e.:
previous:
2021-01-13 02:00:19 [check_op_benchmark_result.py:150] [ERROR] Check speed result with case "minimum_2(backward)" failed.
2021-01-13 02:00:19 [check_op_benchmark_result.py:154] [ERROR] Check accuracy result with case "multiply_7(backward)" failed.
2021-01-13 02:00:19 [check_op_benchmark_result.py:154] [ERROR] Check accuracy result with case "multiply_2(backward)" failed.
2021-01-13 02:00:19 [check_op_benchmark_result.py:154] [ERROR] Check accuracy result with case "divide_4(backward)" failed.
latest:
2021-01-14 22:19:03 [check_op_benchmark_result.py:150] [ERROR] Check speed result with case "subtract_1(forward)" failed.
2021-01-14 22:19:03 [check_op_benchmark_result.py:150] [ERROR] Check speed result with case "subtract_7(forward)" failed.
2021-01-14 22:19:03 [check_op_benchmark_result.py:150] [ERROR] Check speed result with case "pow_4(forward)" failed.
2021-01-14 22:19:03 [check_op_benchmark_result.py:150] [ERROR] Check speed result with case "pow_2(backward)" failed.
2021-01-14 22:19:03 [check_op_benchmark_result.py:154] [ERROR] Check accuracy result with case "multiply_7(backward)" failed.
2021-01-14 22:19:03 [check_op_benchmark_result.py:154] [ERROR] Check accuracy result with case "multiply_4(backward)" failed.
2021-01-14 22:19:03 [check_op_benchmark_result.py:154] [ERROR] Check accuracy result with case "multiply_2(backward)" failed.

AFAIK the resulting log from the benchmark machine has checks about GPU ops, in the log there is no oneDNN verbose info, so it looks that oneDNN kernels are not run, which would eventually be correlated with the PR.

Could you advise on that, please?

paddle/fluid/framework/operator.cc
arlesniak: Done

paddle/fluid/framework/operator.cc
arlesniak: Done

arlesniak:
After 10 restarts CI passed :) 
@luotao1 Could you start your review please ?

arogowie-intel (approved):

arlesniak:
@luotao1 could you please start your review?

arlesniak:
@luotao1 Could you please start your review? PR-CI-Approval will not pass because many files were modified, but most of them it is single line of change.

wojtuss (commented):
When in the class derived from the `OperatorWithKernel` class you do not need to prepend the base class' public method calls with `OperatorWithKernel::` nor `this->` (applies to `CanMKLDNNBeUsed(...)` and `IndicateVarDataType(...)`) in particular). Although there are plenty of such calls in the original code, I would stick to the cleaner approach and skip the redundant prefixes.
If you do not agree, that's fine. LGTM then :-)

paddle/fluid/operators/activation_op.cc
```diff
@@ -93,6 +93,7 @@ framework::OpKernelType GetKernelType(const framework::ExecutionContext& ctx,
                                       const std::string& name) {
   framework::LibraryType library{framework::LibraryType::kPlain};
   framework::DataLayout layout = framework::DataLayout::kAnyLayout;
+  auto data_type = oper.IndicateVarDataType(ctx, name);
```
luotao1, line 96: > AFAIK the resulting log from the benchmark machine has checks about GPU ops, in the log there is no oneDNN verbose info, so it looks that oneDNN kernels are not run, which would eventually be correlated with the PR.
Could you advise on that, please?

Although op-benchmark-ci checks about GPU ops, does line 96 have additional time cost? How about move it into line 109?
```
if (library == framework::LibraryType::kPlain && it != oper.Attrs().end()) {
  auto data_type = oper.IndicateVarDataType(ctx, name);
  if (oper.CanMKLDNNBeUsed(ctx, data_type)) {
    xxxx
  }
}
```

luotao1:
@Avin0323 @GaoWei8 Please see how to make op-benchmark-ci more stable? The same commitid have two different results.
图片

GaoWei8:
> @Avin0323 @GaoWei8 Please see how to make op-benchmark-ci more stable? The same commitid have two different results.
> 图片

The accuracy of multiply op has been exposed before, and the threshold has been changed, so no more errors will be reported.
As for the speed issue, if rerun can pass, it is not considered as a performance issue.

paddle/fluid/operators/activation_op.cc
arlesniak: @luotao1 Thank you for the comment. The `data_type` variable is used also in line 119:
 return framework::OpKernelType(`data_type`, ctx.GetPlace(), layout, library);

 So it will be needed/calculated at the end of the function, whether the condition you mentioned is True or not (despite mkldnn is to be used or not).
 
 In the way it's implemented in PR, the variable value is calculated **only once** per function as it was prior to my changes, without additional time cost.
 It applies to every occurrence in other op files.

arlesniak: @wojtuss Thank you for the comment. Having in mind baby sitting the PR-CI-OP-benchmark for more than a week on the same code, I prefer to not refactor the code in the PR. Of course if it's OK with you because I respect you opinion.

luotao1: > In the way it's implemented in PR, the variable value is calculated only once per function as it was prior to my changes, without additional time cost.

Got it.

>  I prefer to not refactor the code in the PR. Of course if it's OK with you because I respect you opinion.

@wojtuss What's your opinion?

wojtuss: Totally understandable.

wojtuss (approved):
LGTM

PR #29917, opened 2020-12-25
Fix optimizer dtype

zhiqiu:
### PR types

Bug fixes 
### PR changes

APIs
### Describe

The dtype of Optimizer should be inferred by given parameters, which will be used in learning rate and other vars in optimizer.

For example,
```
class MyLayer(paddle.nn.Layer):
    def __init__(self, dtype):
        super(MyLayer, self).__init__()
        self._w = self.create_parameter([2, 3], dtype='float64')
        self._b = self.create_parameter([2, 3], dtype='float64')

    def forward(self, x):
        return x * self._w + self._b
            model = MyLayer(dtype)

x = paddle.rand([10, 2, 3], dtype=dtype)
loss = model(x)
adam = paddle.optimizer.Adam(parameters=model.parameters())
loss.backward()
adam.step()
```

### DIFF
--- a/python/paddle/fluid/optimizer.py
+++ b/python/paddle/fluid/optimizer.py
@@ -106,8 +106,12 @@ class Optimizer(object):
         self.regularization = regularization
         self._grad_clip = grad_clip
         self._learning_rate = learning_rate
-        # the learning rate type should be inferenced from loss
+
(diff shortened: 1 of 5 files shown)


chenwhql (approved):
LGTM

PR #29927, opened 2020-12-26
Support dygraph quant model

juncaipeng:
### PR types

Others
### PR changes

Others
### Describe

Support dygraph quantized model.

Paddle 2.0 uses the dygraph model as default,  and the model quantized by Paddl2.0 is different from Paddle1.8.

In the dygraph quantized model, the inputs -> fake_quant_dequant_ops ->quantized_op. 

Besides, the weights is real float32. 

image

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
+++ b/python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
@@ -299,15 +299,15 @@ class Quant2Int8MkldnnPass(object):
                 ])
 
     def _dequantize_weights(self, graph):
-        # For the dygraph quantized model, the weights are real fp32.
-        # Only dequantize the weights of quantized op in static quantized model,
-        # in which the quantized ops have `quantization_type` attr.
+        def _is_int8_weights(op_node, weight_name):
+            weight_var_name = op_node.input(weight_name)[0]
+            weight = self._load_param(self._scope, weight_var_name)
+            return np.all(np.abs(weight % 1) < 1e-6)
+
         for op in graph.all_op_nodes():
-            if op.name() in self._conv_ops \
-                and op.op().has_attr("quantization_type"):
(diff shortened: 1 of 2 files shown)


wzzju (dismissed):
LGTM.

juncaipeng:
@wojtuss  Please review this PR. Thanks.

python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
```diff
@@ -290,10 +297,15 @@ def _swap_inputs(self, op, old_input, new_input):
                 ])
 
     def _dequantize_weights(self, graph):
+        def _is_int8_weights(op_node, weight_name):
+            weight_var_name = op_node.input(weight_name)[0]
+            weight = self._load_param(self._scope, weight_var_name)
+            return np.all(np.abs(weight % 1) < 1e-6)
```
wojtuss: I guess the `1e-6` constant comes from the precision of fake-quantization. Can it be named and connected to the precision? I do not know how likely it is for the precision to change in the future.

juncaipeng: The function analyzes the weights is real float32 or fake float32 (int8 range).
I have updated `np.all(np.abs(weight % 1) < 1e-6)`  to `np.all(np.mod(weight, 1) == 0)`.

luotao1 (approved):

python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
juncaipeng: @wojtuss   As the tag day is January 12 and the QA have to test the module, we merge the PR right now. If you have other comments, please let me know and I will fix in the next PR.

wojtuss: @juncaipeng , Thanks. LGTM

PR #29947, opened 2020-12-28
add bkcl.so in whl for kunlun

QingshuChen:
### PR types
Others

### PR changes
Others

### Describe
add bkcl.so in whl for kunlun


wangxicoding (approved):
LGTM

PR #29962, opened 2020-12-28
Set FLAGS_selected_gpus for spawn 

chenwhql:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Set FLAGS_selected_gpus for spawn.

When the child process starts, it will inherit the configuration of the main process and set the FLAGS once, but the environment variable has not been set at this time, which leads to the FLAGS_selected_gpus is keep same with mainprocess(usually empty), so manually update the flags here.

注：增加了一个单测，又移除了，单测打印显示CI机器nvidia-smi只有两张卡，需要大于两张卡才能测这个问题


hutuxian (approved):
LGTM

PR #29965, opened 2020-12-28
[Dy2Stat] Don't convert to paddle.shape if var_x.shape is not negetive

liym27:
### PR types

Performance optimization
### PR changes

Others
### Describe

1. When `x` is Variable, call nn.shape(x) only in following cases:
- The shape of `x` is used in control flow condition.
     ```python
     if x.shape[0] == 1:
         y = XX
     ```python
- The dim to be used is negetive
     ```python
     # Assume x.shape=[3, -1] in static mode
     y = paddle.reshape(x, shape=[1, x.shape[1]])
     ```
2.  When `x` is Variable, but x.shape or x.shape[idx] doesn't contain negetive value, don't convert to paddle.shape()


zhhsplendid (approved):
LGTM

PR #30035, opened 2020-12-31
fix large scale memory 

seiriosPlus:
### PR types

Performance optimization
### PR changes

Others
### Describe

优化存储结构
大规模稀疏内存占用降低4.65倍

PR #30085, opened 2021-01-04
fix large scale memory  (#30035)

seiriosPlus:
### PR types

Performance optimization
### PR changes

Others
### Describe

large scale kv memory hold optimize.


MrChengmo (approved):
LGTM

PR #30087, opened 2021-01-04
add topo-aware in heter-ps

Thunderbrook:
### PR types
Function optimization

### PR changes
Others

### Describe
add topo aware in heter-ps to speed up


danleifeng (approved):
LGTM

PR #30095, opened 2021-01-04
fix xpu pe sync

vslyu:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix  parallel executor sync when using kunlun xpu


wangxicoding (approved):
LGTM

PR #30123, opened 2021-01-05
Add callback after TensorCopy 

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Add callback after TensorCopy to make sure that the src tensor is not destructed until the copy is completed.

### Why
In dygraph mode, the `life cycle` or `duration` of a Tensor is maintained by the `reference count` in both c++ and python.

When doing async operation on cuda, the Tensor may be destructed (and thus the corresponding memory is freed) before the operation is completed.

Since paddle is single-stream-desgin, 
- If all the async operations are on cuda device, it is safe since the freed memory will not be used again before its operation is completed.
- If the operations are between different place, it may introduce data inconsistent. For example, `cudaMemcpyAsync` is async when copying from `cuda pinned place` to `cuda place`, if the memory of `cuda pinned place` (host memory, actually) is freed, it may be reused by others and the data may change before the copy is completed. 

An example code,
```
import paddle
import numpy as np
import time

start = time.time()

x = paddle.rand([4096, 4096])
for i in range(500):
    x = paddle.matmul(x, x)

a_np = np.random.rand(1024, 1024)
a = paddle.to_tensor(a_np, place=paddle.CUDAPinnedPlace())
print(a.name)
print(a)
a = paddle.to_tensor(a)   # it will copy a from CUDAPinnedPlace to CUDAPlace(0), which is async
print('cpu end', time.time()-start)
x.numpy()
print('gpu end', time.time()-start)
```
- Before
image
As shown in the figure, the tensor `generated_tensor_501` is destructed immediately after the copy(cudaMemcpyAsync) is launched.

- After
image
As shown in the figure, the tensor `generated_tensor_501` is destructed 3 seconds later after the copy(cudaMemcpyAsync) is launched.

### DIFF
--- a/paddle/fluid/imperative/tracer.cc
+++ b/paddle/fluid/imperative/tracer.cc
@@ -55,6 +55,45 @@ static void PassStopGradient(const NameVarBaseMap& outs, bool generate_grad) {
     }
   }
 }
+paddle::framework::GarbageCollector* Tracer::MutableGarbageCollectorIfNotExists(
+    const platform::Place& place) {
+  // if not exists, create a new GarbageCollector at given place
+  if (gcs_.count(place) == 0) {
+    std::unique_ptr<framework::GarbageCollector> gc;
+    if (platform::is_gpu_place(place)) {
+#ifdef PADDLE_WITH_CUDA
+      gc.reset(new framework::StreamGarbageCollector(
+          BOOST_GET_CONST(platform::CUDAPlace, place), 0));
+
+      VLOG(10) << "Created GarbageCollector at " << place;
+#else
+      PADDLE_THROW(platform::errors::PermissionDenied(
+          "Paddle can't use CUDA device since it's not compiled with CUDA,"
+          "Please recompile or reinstall Paddle with GPU support."));
+#endif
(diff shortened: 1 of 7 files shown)


paddle/fluid/imperative/tracer.cc
```diff
@@ -56,6 +56,78 @@ static void PassStopGradient(const NameVarBaseMap& outs, bool generate_grad) {
   }
 }
 
+void IncreaseVarbaseReferenceCountUntilCopyComplete(
+    const std::shared_ptr<imperative::VarBase>& var,
+    const platform::Place& place) {
+  // Note(zhiqiu): Follow the logic of TensorCopy to determine the place that we
+  // need to add callback, see tensor_utils.cc:245
```
chenwhql, line 63: not importent, `tensor_utils.cc` -> `tensor_util.cc`

chenwhql (approved):
LGTM

paddle/fluid/imperative/tracer.cc
zhiqiu: OK, I will change it in the next PR.

PR #30139, opened 2021-01-06
Improve Index select cuda kernel

123malin:
### PR types

Performance optimization

### PR changes

OPs

### Describe

Improve index_select's cuda kernel performance.


seiriosPlus (approved):
LGTM

PR #30147, opened 2021-01-06
check rank in kernel of set_value op

liym27:
### PR types

Others
### PR changes

OPs
### Describe

For op set_value, check input's rank < 7


zhhsplendid (approved):
LGTM

PR #30168, opened 2021-01-06
Fix test_slice: avoid unnecessary copying of TensorArray from subblock to parent block

liym27:
### PR types

Bug fixes 
### PR changes

Others
### Describe

Fix unitest `test_slice` to avoid runing failed.
image

In this PR, in control flow, **don't** copy TensorArray from subblock to parent block when TensorArray is created in parent block.


zhhsplendid (approved):
LGTM

chalsliu (approved):

PR #30186, opened 2021-01-07
add fp16 support for tril_triu op

windstamp:
### PR types

Function optimization

### PR changes

OPs

### Describe

add fp16 support for tril_triu op


zhangting2020 (approved):
LGTM

PR #30199, opened 2021-01-07
use cuda generator in bernoulli cuda kernel

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

In order to make paddle.bernoulli() reproducible, use cuda generator in bernoulli cuda kernel


zhiqiu:
The new implementation uses cuda generator to get seed  instead of random seed, which increases the execution time.

GaoWei8 (approved):
LGTM for op benchmark ci

PR #30201, opened 2021-01-07
shape op support int8 and uint8 tensor

jiweibo:
### PR types

Others
### PR changes

Others
### Describe

shape op support int8 and uint8 tensor


juncaipeng (approved):
lgtm

PR #30130, opened 2021-01-05
fix adamw apply gradient

wangxicoding:
### PR types

Bug fixes
### PR changes

APIs
### Describe

1. Fix Adamw weight_decay not applied when open AMP. see https://github.com/PaddlePaddle/Paddle/issues/29794
2. Fix Adamw weight_decay LR is fixed when using LRSchedule.
3. Fix Adam step() not add imperative_base.no_grad, which gradients may be generated in the optimizer.
4. Optimization calculation, the calculation formula of `param = param - param * lr * coeff` is optimized as follows
     `param = param * (1.0 - lr * coeff)`

Issue #29794, opened 2020-12-21
静态图amp与`paddle.optimizer.AdamW`不兼容


Meiyim:
[这行](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/contrib/mixed_precision/decorator.py#L222)代码中将`minimze` 拆解成`bacward`和`apply_optimize `
导致`AdamW`中进行weight_decay的[逻辑](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/optimizer/adamw.py#L195)被绕过.

诸如此类的问题应该还有其他处（如regularizer,param_attr的插入），可能amp这块需要重新考虑兼容性问题。

zhiqiu:
> [这行](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/contrib/mixed_precision/decorator.py#L222)代码中将`minimze` 拆解成`bacward`和`apply_optimize `
> 导致`AdamW`中进行weight_decay的[逻辑](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/optimizer/adamw.py#L195)被绕过.
> 
> 诸如此类的问题应该还有其他处（如regularizer,param_attr的插入），可能amp这块需要重新考虑兼容性问题。

嗯嗯，这个问题我们也注意到了，已经有计划修复。

Meiyim:
any update?

wangxicoding:
@Meiyim 已修复，试一下~

PR #30207, opened 2021-01-07
[cherry-pick 2.0] fix adamw apply gradient

wangxicoding:
### PR types

Bug fixes
### PR changes

APIs
### Describe

cherry-pick from https://github.com/PaddlePaddle/Paddle/pull/30130
1. Fix Adamw weight_decay not applied when open AMP. see #29794
2. Fix Adamw weight_decay LR is fixed when using LRSchedule.
3. Fix Adam step() not add imperative_base.no_grad, which gradients may be generated in the optimizer.
4. Optimization calculation, the calculation formula of `param = param - param * lr * coeff` is optimized as follows
     `param = param * (1.0 - lr * coeff)`


swtkiwi (approved):
LGTM

PR #30222, opened 2021-01-08
fix pad

littletomatodonkey:
### PR types
Bug fixes

### PR changes
APIs

### Describe
Fix assertion error when x dim is 2 and pad length is 4.


willthefrog (approved):
LGTM

PR #30265, opened 2021-01-11
Support inplace move of VarBase

Aurelius84:
### PR types

New features
### PR changes

APIs
### Describe

Support inplace move of VarBase

### DIFF
--- a/paddle/fluid/imperative/layer.cc
+++ b/paddle/fluid/imperative/layer.cc
@@ -17,7 +17,6 @@
 #include <queue>
 #include <utility>
 
-#include "paddle/fluid/framework/data_transform.h"
 #include "paddle/fluid/framework/framework.pb.h"
 #include "paddle/fluid/framework/op_registry.h"
 #include "paddle/fluid/framework/variable_helper.h"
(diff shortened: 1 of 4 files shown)


zhwesky2010 (dismissed):
LGTM

paddle/fluid/imperative/layer.cc
```diff
@@ -311,6 +312,42 @@ void VarBase::CopyFrom(const VarBase& src, const bool blocking) {
...
+  const framework::Tensor* src_tensor = nullptr;
+  if (Var().IsType<framework::LoDTensor>()) {
+    src_tensor = &Var().Get<framework::LoDTensor>();
+  } else {
+    src_tensor = &Var().Get<framework::SelectedRows>().value();
+  }
+
+  framework::Tensor dst_tensor;
+  dst_tensor.Resize(src_tensor->dims());
+  dst_tensor.mutable_data(dst_place, src_tensor->type());
```
liym27, line 337: Does the difference between `dst_tensor` and `src_tensor`  **only** is **the place**? 
If yes,  
1. `dst_tensor` need to share the `inplace_version_counter_` with `src_tensor` to ensure the correctness of gradient calculation
Please add code like:
```c++
dst_tensor->ShareInplaceVersionCounterWith(src_tensor);
```

2. Here only set the same `dims` and `type` as `src_tensor`, does it ensure that the other attributes of `src_tensor` and `dst_tensor` are consistent? such as `offset_`, `layout_`?

Aurelius84: Thanks to find this, I will fix this.

Aurelius84: use `src_tensor->ShareBufferWith(dst_tensor)` to reset `holder_` and `offset_`. Other information of src_tensor is still keep unchange.

PR #30338, opened 2021-01-12
Set expected place in child thread for dataloader to avoid costing cuda memory on other card

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Set expected place in child thread for dataloader to avoid costing cuda memory on other card

`cudaSetDevice()` is valid inside host thread, so the child thread in dataloader may have the default device id (0) if not set explicitly, which may cost about 500MB memory on card 0.

It is worse when doing multi-card training in dygraph, since each process will cost about 500MB memory on card 0.


chenwhql (approved):
LGTM

heavengate (approved):
LGTM for multi-process DataLoader

Xreki (approved):
LGTM for the change of framework.py

XiaoguangHu01 (approved):
LGTM

PR #30342, opened 2021-01-12
skip quantizing ops in cpu inference

juncaipeng:
### PR types

Others
### PR changes

Others
### Describe

* skip quantizing ops in cpu inference

* remove fake_channel_wise_quantize_dequantize_abs_max to support per-channel quantization for dygraph quantized model

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass.cc
@@ -42,6 +42,10 @@ void CPUQuantizePlacementPass::ApplyImpl(ir::Graph* graph) const {
       return;
     }
 
+    if (op->Op()->GetAttrIfExists<int>("skip_quant") == 1) {
+      return;
+    }
+
     if (op->Op()->HasAttr("mkldnn_data_type") ||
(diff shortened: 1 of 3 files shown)


juncaipeng:
@wojtuss @lidanqing-intel  Please review this PR, thank you.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass.cc
```diff
@@ -42,6 +42,10 @@ void CPUQuantizePlacementPass::ApplyImpl(ir::Graph* graph) const {
       return;
     }
 
+    if (op->Op()->GetAttrIfExists<int>("skip_quant") == 1) {
```
wojtuss, line 45: The attribute is set to `True`, so I would get its value as `bool`.
Apart from that LGTM.

juncaipeng: I have set its value as bool, but it raises error. I think Paddle framework changes the data type from bool to int.

The skip_quant attr in saved model as follows.

image

wojtuss: OK, that's fine then.

wojtuss (approved):
LGTM

PR #30354, opened 2021-01-12
fix quantize error in speical naming model

baiyfbupt:
### PR types

Bug fixes
### PR changes

Others
### Describe

Refine quantize find parent.
For a conv layer name like: `neck`.`yolo_block.0`.`conv_module`.`conv0`.`conv`
User named a module as `yolo_block.0`， the original code will report an error that `neck does not have "yolo_block" attr`.
This PR fixed the issue.


juncaipeng (dismissed):
LGTM

juncaipeng (approved):
LGTM

PR #30358, opened 2021-01-12
[oneDNN] Cache oneDNN stream not to recreate in each oneDNN op 

jczaja:
### PR types
Performance optimization

### PR changes
OPs

### Describe
created on dnnl::stream involves small memory allocation. stream can be created once and then reused so this PR makes stream to be created one per thread and store in per thread object.

### DIFF
--- a/paddle/fluid/operators/fused/mkldnn/fusion_gru_mkldnn_op.cc
+++ b/paddle/fluid/operators/fused/mkldnn/fusion_gru_mkldnn_op.cc
@@ -246,7 +246,8 @@ class GRUMKLDNNHandler : public platform::MKLDNNHandlerT<T, dnnl::gru_forward> {
       memory_p = std::make_shared<dnnl::memory>(this->fwd_pd_->src_iter_desc(),
                                                 this->engine_);
 
-      auto astream = paddle::platform::MKLDNNDeviceContext::tls().get_stream();
+      const auto& astream =
+          paddle::platform::MKLDNNDeviceContext::tls().get_stream();
       dnnl::reorder(user_h0_memory, *memory_p, attr_)
           .execute(astream, user_h0_memory, *memory_p);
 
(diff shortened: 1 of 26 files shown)


paddle/fluid/platform/device_context.h
```diff
@@ -525,6 +525,9 @@ class MKLDNNDeviceContextThreadLocals {
     // Recently registered data_format. This is needed to
     // know for converting MKL-DNN Tensor to non MKL-DNN
     paddle::framework::DataLayout cur_paddle_data_layout;
+    // MKL-DNN stream used for execution of primitives (per-thread)
+    mkldnn::engine aengine;
+    mkldnn::stream astream;
```
arogowie-intel: The names of these variables don't stick to current naming convention in this struct. I'd suggest to remove the `a` prefix.

paddle/fluid/operators/fused/mkldnn/fusion_gru_mkldnn_op.cc
```diff
@@ -246,7 +246,7 @@ class GRUMKLDNNHandler : public platform::MKLDNNHandlerT<T, dnnl::gru_forward> {
       memory_p = std::make_shared<dnnl::memory>(this->fwd_pd_->src_iter_desc(),
                                                 this->engine_);
 
-      dnnl::stream astream(this->engine_);
+      auto astream = paddle::platform::MKLDNNDeviceContext::tls().get_stream();
```
arogowie-intel: Please consider this change to make sure no copy is done.
```suggestion
      const auto& astream = paddle::platform::MKLDNNDeviceContext::tls().get_stream();
```
The `execute` function accepts `astream` arg as a [const ref already](https://spec.oneapi.com/versions/latest/elements/oneDNN/source/primitives/reorder.html#_CPPv4NK4dnnl7reorder7executeERK6streamR6memoryR6memory).

arogowie-intel (commented):
Please use `auto&` to make sure that you're not doing copy while reaching for stream object.

paddle/fluid/platform/device_context.h
jczaja: ok

paddle/fluid/operators/fused/mkldnn/fusion_gru_mkldnn_op.cc
jczaja: ok . done

jczaja:
@arogowie-intel I have added some additional change (last commit) to fix a problem exposed by test-analyzer-detect. If you want more details then ask me.

jakpiase (approved):
LGTM

jczaja:
@luotao1 Could you please start your review? CI will not pass without approval. This PR is changing way oneDNN stream was created , so due to API change I had do it all at once in all locations. So there many files modified, but most of them it is single line of change.

arogowie-intel (approved):

luotao1 (approved):

PR #30384, opened 2021-01-13
Fix prune input bug of jit.save

chenwhql:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix prune input bug of jit.save

`paddle.jit.save` support prune program when the input_specs of `jit.to_static` and `jit.save` are given and the input_specs are compatible, but only tested the case of both `input_spec` and `output_spec` are given previously.

This PR add unittests and fix this problem. 

Test case like:

```
class LinerNetWithUselessLabel(paddle.nn.Layer):
        super(LinerNetWithUselessLabel, self).__init__()
        self._linear = Linear(in_size, out_size)

    @paddle.jit.to_static(input_spec=[
        InputSpec(
            shape=[None, 784], dtype='float32', name="image"), InputSpec(
                shape=[None, 1], dtype='int64', name="label")
    ])
    def forward(self, x, label):
        out = self._linear(x)
        loss = fluid.layers.cross_entropy(out, label)
        avg_loss = fluid.layers.mean(loss)
        return out

layer = LinerNetWithUselessLabel(784, 1)

model_path = "test_prune_to_static_no_train/model"
paddle.jit.save(
    layer,
    model_path,
    input_spec=[
        InputSpec(
            shape=[None, 784], dtype='float32', name="image")
    ])
```

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/program_translator.py
@@ -470,19 +470,22 @@ class StaticFunction(object):
         cached_program_len = len(self._program_cache)
         # If specific `input_spec`, apply convertion from dygraph layers into static Program.
         if cached_program_len == 0:
-            if input_spec is None:
-                input_spec = self._function_spec.input_spec
-            elif self._function_spec.input_spec is not None:
-                if not input_specs_compatible(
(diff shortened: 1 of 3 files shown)


Aurelius84 (approved):
LGTM

liym27 (approved):
LGTM

PR #30395, opened 2021-01-13
fix bug of celoss when using ignore_index and reduction

chajchaj:
### PR types
Bug fixes

### PR changes
APIs 

### Describe
fix bug of celoss when using ignore_index and reduction


willthefrog (approved):
LGTM

PR #30450, opened 2021-01-14
Update voc dataset url

LielinJiang:
### PR types

Others
### PR changes

Others
### Describe

use bcebos url instead of original foreign website


saxon-zh (approved):
LGTM

PR #30465, opened 2021-01-15
add the test case for cpu with different batch size

wawltor:
### PR types
Bug fixes 

### PR changes
OPs

### Describe
add the test case for cpu with different batch size


python/paddle/fluid/tests/unittests/rnn/test_rnn_nets.py
```diff
@@ -338,6 +338,36 @@ def forward(self, input):
     paddle.disable_static()
 
 
+class TestLSTMWithSeqs(unittest.TestCase):
+    def setUp(self):
+        paddle.disable_static()
+        paddle.set_device("cpu")
```
zhiqiu, line 344: Test CUDAPlace as well.

PR #30469, opened 2021-01-15
fix range op crash in dygraph xpu place

taixiurong:
### PR types
Others

### PR changes
OPs 

### Describe
range op crash in dygraph xpu place
*test=kunlun


QingshuChen:
LGTM

fuyinno4 (approved):

PR #30484, opened 2021-01-15
Recompute Offload: fixed bug in memcpy

JZ-LIANG:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Recompute Offload: fixed bug in memcpy
Recompute Offload support Sharding


sandyhouse (approved):
LGTM

PR #30492, opened 2021-01-15
Solve inconsistent order in each card in  dynamic graph

ForFishes:
### PR types

Others
### PR changes

Others
### Describe

The arrival order of the parameters of each card in the gradient of LSTM is inconsistent, resulting in allreduce hang.

### DIFF
--- a/paddle/fluid/imperative/reducer.cc
+++ b/paddle/fluid/imperative/reducer.cc
@@ -22,11 +22,6 @@ std::shared_ptr<Reducer> Reducer::s_instance_ = NULL;
 
 // context is used to select the stream for concat
 void Group::ConcatTensors(const platform::CUDADeviceContext &context) {
-  VLOG(3) << "Before concat, set output tensor size is " << all_length_;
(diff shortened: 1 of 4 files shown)


paddle/fluid/imperative/reducer.cc
```diff
@@ -94,14 +94,15 @@ Reducer::Reducer(const std::vector<std::shared_ptr<imperative::VarBase>> &vars,
                  const std::vector<bool> &is_sparse_gradient,
                  std::shared_ptr<imperative::ParallelContext> parallel_ctx,
                  const std::vector<size_t> &group_size_limits,
-                 bool find_unused_vars)
+                 bool find_unused_vars, bool reject_rebuld_group)
```
hutuxian: `use_rebuld_group` may be better? And we can make its default value `False`

ForFishes: ok， fix it

hutuxian (dismissed):
LGTM

PR #30526, opened 2021-01-18
fix adamw lr_to_coeff is fixed when dygraph

wangxicoding:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix bug in https://github.com/PaddlePaddle/Paddle/pull/30130 
We use `_lr_to_coeff` to cache the coeff result. But in dygraph mode, apply_gradient will be executed every step and recompute a new coeff value. So need to clear `_lr_to_coeff` every step, otherwise `_lr_to_coeff` will be fixed.


gongweibao (approved):
LGTM

PR #30536, opened 2021-01-18
Fix bug: GetAttrValue should deal with attr with attrType vector<double>

liym27:
### PR types

Bug fixes
### PR changes

OPs
### Describe

As the title

Before this PR, run code as follows:
```python
model = LayerWithSetValue(input_dim=10, hidden=1)
x = paddle.full(shape=[5, 10], fill_value=5.0, dtype="float32")
paddle.jit.save(
    layer=model,
    path="./layer_use_set_value",
    input_spec=[x],
    output_spec=None)
```

RuntimeError:
```
RuntimeError: (Unavailable) Unsupport attribute type 12. (at/home/teamcity/work/ef54dc8a5b211854/paddle/fluid/framework/attribute.cc:74)
```

### DIFF
--- a/paddle/fluid/framework/attribute.cc
+++ b/paddle/fluid/framework/attribute.cc
@@ -69,6 +69,15 @@ Attribute GetAttrValue(const proto::OpDesc::Attr& attr_desc) {
       }
       return val;
     }
+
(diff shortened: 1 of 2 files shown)


zhhsplendid (approved):
LGTM

PR #30551, opened 2021-01-18
Add bf16 fast performance verification

wozna:
### PR types

Others 
### PR changes

APIs 
### Describe

This PR adds a check to see if the CPU supports bfloat16 fast execution. Not all operating systems support reading correctly the `avx512_bf16` flag, which indicates the machine supports bfloat16 calculations. Therefore, it is best to use the CPUID function for this.
This PR adds this call using the instruction documentation https://software.intel.com/content/dam/develop/external/us/en/documents-tps/architecture-instruction-set-extensions-programming-reference.pdf

Therefore, to check whether the machine supports BF16 fast execution, it is enough to call:
- *c ++* : `platform::MayIUse(platform::cpu_isa_t::avx512_bf16)`
- *python* : `core.supports_bfloat16_fast_performance()`

Should be merged after Xbyak update https://github.com/PaddlePaddle/Paddle/pull/30809


jczaja (dismissed):
LGTM

paddle/fluid/operators/jit/gen/jitcode.h
```diff
@@ -98,7 +98,7 @@ class JitCode : public GenBase, public Xbyak::CodeGenerator {
     ret();
   }
   void L(const char* label) { Xbyak::CodeGenerator::L(label); }
-  void L(const Xbyak::Label& label) { Xbyak::CodeGenerator::L(label); }
```
arogowie-intel: I'm wondering why you've removed `const` ?  Does this function modify `label` ?

wozna: Because I've changed Xbyak library version from  v5.661 to v5.81  it turned out that one of the functions has changed from `void L (const Label & label)` to `void L (Label & label)`, so we also have to remove const so that there is no error. And I checked that now indeed the Label object is modified in Xbyak L function

wozna:
@luotao1  
In this PR, I changed the Xbyak version from 5.661 to 5.81, which supports bfloat16 instruction checking.
Unfortunately, in `PR-CI-CPU-Py2` , you can see in the build log that Xbyak was built with the old 5.661 version, so this test fails. Do you know what the reason is for this? Should I change the version elsewhere as well?

luotao1:
emmm, PR_CI_CPU_PY2 runs first develop branch then user-PR branch. There maybe something error in CI configure (@tianshuo78520a Please see it). To make your PR successfully pass. You can seperate into two PRs: 
- Only update Xbyak.cmake from from 5.661 to 5.81
- Other changes.

wozna:
I split this PR that Xbyak update should be merged first: https://github.com/PaddlePaddle/Paddle/pull/30809
Then this PR.

cmake/external/xbyak.cmake
```diff
@@ -44,7 +44,7 @@ ExternalProject_Add(
     DEPENDS             ""
     PREFIX              ${XBYAK_PREFIX_DIR}
     SOURCE_DIR          ${XBYAK_SOURCE_DIR}
-    UPDATE_COMMAND      ""
+    # UPDATE_COMMAND      ""
```
wozna, line 47: I commented on this line because if UPDATE_COMMAND is explicitly blank it blocks update of this project in re-run even if GIT_TAG was changed. This problem is mentioned in this cmake file: https://github.com/Shixiaowei02/Paddle/blob/da30e49f3e30151633281c048486486934ef4a7b/cmake/external/pybind11.cmake#L37

It turned out that in PR-CI-CPU-Py2, the new PR changes are also only an incremental compilation (re-build). So one solution is to delete this command.

@jczaja @luotao1 What do you think about such a solution?

jczaja: So when this UPDATE_COMMAND is commented out then each time we do rebuild then xbyak will be checked out?

wozna: Exactly, Xbyak is going to be checked out every time.
Of course, this error was only in one CI test, because it does not build changes from scratch, which is also not clear to me why.
Of course, after adding this PR, we can remove this comment.

jczaja: @wozna xbyak is a small bunch of headers so ti shouldn't been a problem when it is checked out when rebuilding paddle. So It is good for me.

jczaja (approved):
LGTM

jczaja:

luotao1 (approved):

PR #30637, opened 2021-01-21
fix abs bug and add abs test case

MingMingShangTian:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Fix the bug of abs when in cpu mode， the input is float but the out put is int. 
This bug is only occur in gcc 4.8.5 version. The reson is that the abs function is uesd by C lib abs which returns the int type.
And add abs test case。


GaoWei8 (approved):
LGTM for op benchmark ci

chenwhql (approved):

PR #30686, opened 2021-01-25
fix test_gen_nccl_id_op failed

wangxicoding:
### PR types

Others
### PR changes

Others
### Describe

Fix test_gen_nccl_id_op failed. 
Since multiple processes cannot be covered by coverage, the previous PR https://github.com/PaddlePaddle/Paddle/pull/30455 changed the test to multithreaded mode.
But multi-threading seems to cause the test to fail, so change back to multi-process execution.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_gen_nccl_id_op.py
+++ b/python/paddle/fluid/tests/unittests/test_gen_nccl_id_op.py
@@ -16,6 +16,7 @@ import unittest
 import os
 import copy
 from launch_function_helper import wait, _find_free_port
+from multiprocessing import Pool, Process
 from threading import Thread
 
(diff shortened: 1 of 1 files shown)


python/paddle/fluid/tests/unittests/test_gen_nccl_id_op.py
```diff
@@ -16,6 +16,7 @@
 import os
 import copy
 from launch_function_helper import wait, _find_free_port
+from multiprocessing import Pool, Process
 from threading import Thread
```
sandyhouse, line 20: No need for Thread package, but still LGTM.

sandyhouse (approved):

Issue #30115, opened 2021-01-05
MKLDNN Error: few segmentation model prediction failed


OliverLPH:
Thank you for contributing to PaddlePaddle.
Before submitting the issue, you could search issue in the github in case that th
If there is no solution,please make sure that this is an inference issue including the following details :
**System information**
-PaddlePaddle version （eg.1.1）or CommitID
-CPU: MKLDNN, onednn 1.8, Intel(R) Xeon(R) Gold 6271C CPU @ 2.60GHz
-GPU: None
-OS Platform: CentOS Linux 7.6.1810
-Docker Image: hub.baidubce.com/paddlepaddle/paddle:latest-dev
-Python version: None
-Cmake orders
-C++version.txt
-API information
Note: You can get most of the information by running [summary_env.py](https://github.com/PaddlePaddle/Paddle/blob/develop/tools/summary_env.py). 
**To Reproduce**
Seg model download link: https://sys-p0.bj.bcebos.com/Paddle-UnitTest-Model/PaddleSeg.tgz
current failed models are `fastscnn`, `hrnet`, `pspnet`, you can download from link above

```shell
git clone https://github.com/PaddlePaddle/continuous_integration.git
cd continuous_integration/inference/inference_api_test/cpp_api_test
bash build.sh ${paddle_Inference_dir} OFF ON OFF
./build/test_clas_model --model_name=${model_name} \
                            --model_path=${model_path} \
                            --params_path=${params_path} \
                            --image_shape="3,512,512" \
                            --use_mkldnn=true \
                            --batch_size=1 \
                            --use_gpu=false \
                            --accuracy="1e-6" \
                            --gtest_output=xml:test_${model_name}_mkldnn_${accuracy}_bz${batch_size}.xml
```

**Describe your current behavior**
![image](https://user-images.githubusercontent.com/14963836/103619506-d8f61400-4f6c-11eb-82a1-e0b9a00c86e4.png)
```shell
    File "/workspace/pdseg/models/model_builder.py", line 167, in build_model
      logits = seg_model(image, class_num)
    File "/workspace/pdseg/models/model_builder.py", line 86, in seg_model
      logits = fast_scnn.fast_scnn(image, class_num)
    File "/workspace/pdseg/models/modeling/fast_scnn.py", line 280, in fast_scnn
      lower_res_feature = global_feature_extractor.net(higher_res_features)
    File "/workspace/pdseg/models/modeling/fast_scnn.py", line 226, in net
[...]

PR #30720, opened 2021-01-26
[WIP][oneDNN] Fallback to CPU when adaptive pooling cannot be supported by oneDNN pool kernel

jczaja:
### PR types
Bug fixes

### PR changes
OPs

### Describe
This PR is disabling pool oneDNN kernel in sitatuation when adpative pool is to be used when input_shape is not divisible by corresponding pool window dims. This PR is fixing crash repored at  #30115


jczaja:
This is no longer needed as proper support for adaptive pool (#30757) was added

PR #30733, opened 2021-01-27
fixed compilation error on gcc 4.8.x due to the usage of isfinite

alncat:
### PR types

Bug fixes
### PR changes

Others
### Describe

fixed compilation error on gcc 4.8.x due to the usage of isfinite


OliverLPH (approved):
LGTM

PR #30777, opened 2021-01-29
Fix the nan bug when passing all zero values into clip_by_norm_op.

wzzju:
### PR types

Bug fixes
### PR changes

OPs

### Describe

```cpp
0.0 / 0.0 = -nan
1.0 / 0.0 = inf
```
As described above, if all input grads are zero, the output of `clip_by_norm` will be inf or nan.  This pr is used to fix this bug.


lanxianghit (approved):
LGTM

GaoWei8 (approved):

PR #30809, opened 2021-02-01
Update Xbyak to v5.81

wozna:
### PR types

Others
### PR changes

Others
### Describe

This PR updates the Xbyak from 5.661 to 5.81. Only from this version, the instructions related to BF16  are correctly checked.
This PR is one part of the assignment to add BF16 instruction checks.
Moreover, because I've changed Xbyak library version from v5.661 to v5.81 it turned out that one of the functions has changed from `void L (const Label & label)` to `void L (Label & label)`, so we also have to remove const from jitcode.h so that there is no error.


jczaja (approved):
LGTM

wozna:
@luotao1  Could you start your review?

luotao1 (approved):

PR #30817, opened 2021-02-01
Performance optimization for dynamic setitem

liym27:
### PR types

Performance optimization

### PR changes

OPs

### Describe

Call op set_value to speed up because the original call to TensorToPyArray will introduce unnecessary data copy.

### DIFF
--- a/paddle/fluid/pybind/imperative.cc
+++ b/paddle/fluid/pybind/imperative.cc
@@ -582,26 +582,82 @@ void BindImperative(py::module *m_ptr) {
               py::object &value_obj) {
              auto self_tensor =
                  self->MutableVar()->GetMutable<framework::LoDTensor>();
-             auto self_numpy = TensorToPyArray(*self_tensor);
+             PyObject *index_ptr = !PyTuple_Check(_index.ptr())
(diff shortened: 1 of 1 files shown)


paddle/fluid/pybind/imperative.cc
```diff
@@ -582,26 +582,82 @@ void BindImperative(py::module *m_ptr) {
               py::object &value_obj) {
              auto self_tensor =
                  self->MutableVar()->GetMutable<framework::LoDTensor>();
-             auto self_numpy = TensorToPyArray(*self_tensor);
+             PyObject *index_ptr = !PyTuple_Check(_index.ptr())
```
zhhsplendid, line 585: Not Important, you can choose to change or not

Why add the "!"? Can we write simpler (prefer shorter, positive condition):

"PyTuple_Check(_index.ptr()) ? _index.ptr(): PyTuple_Pack(1, _index.ptr())"

zhhsplendid (approved):
LGTM

paddle/fluid/pybind/imperative.cc
liym27: Thanks, I will modify it next PR.

PR #30836, opened 2021-02-02
[Static setitem] Support index is ellipsis for setitem in static mode

liym27:
### PR types

New features 
### PR changes

APIs
### Describe

功能支持：静态图支持 slice 赋值，inplace操作。

本 PR 支持场景：
key支持：Ellipsis
value支持：int, float, numpy.ndarray, Paddle Tensor
dtype: int32, int64, float32, float64, bool
其他：默认支持广播

```
tensor_x = paddle.to_tensor(np.ones((2, 3)).astype(np.float32)) # [[1,1,1], [1,1,1]]
tensor_x[...] = 0   # tensor_x : [[0, 0, 0], [0, 0, 0]]
tensor_x[...] = np.array([3,3,3]) # tensor_x : [[3, 3, 3], [3, 3, 3]]
tensor_x[...] = paddle.ones([3]) # tensor_x : [[1,1,1], [1,1,1]]
```

---

**当前 setitem 功能，索引类型支持情况如下**

索引类型 | 示例 | 动态图 | 静态图
-- | -- | -- | --
标量 | x[1]   = y | √ | √
切片 | x[1:2]   = y | √ | √
跨步长 | x[1:10:3]   = y | √ | ×
反向步长 | x[1:10:-2]   = y | √ | ×
省略[Ellipsis] | x[…,   0] = y | √ | **本 PR 支持**
Tensor | x[tensor_i]   = y | × | ×
布尔 | x[x   != 1] = y | × | ×

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -1866,6 +1866,35 @@ class Variable(object):
         starts = []
         ends = []
         max_integer = sys.maxsize
+
+        def replace_ellipsis(item):
+            # Use slice(None) to replace Ellipsis.
+            # For var, var.shape = [3,4,5,6]
+            #
+            #   var[..., 1:2] -> var[:, :, :, 1:2]
+            #   var[0, ...] -> var[0]
+            #   var[0, ..., 1:2] -> var[0, :, :, 1:2]
+
+            item = list(item)
+            ell_count = item.count(Ellipsis)
+            if ell_count == 0:
+                return item
+            elif ell_count > 1:
+                raise IndexError(
+                    "An index can only have a single ellipsis ('...')")
+
+            ell_idx = item.index(Ellipsis)
+
+            if ell_idx == len(item) - 1:
+                return item[:-1]
+            else:
+                item[ell_idx:ell_idx + 1] = [slice(None)] * (
+                    len(self.shape) - len(item) + 1)
+
+            return item
+
+        item = replace_ellipsis(item)
+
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/framework.py
```diff
@@ -1866,6 +1866,35 @@ def __setitem__(self, item, value):
...
+
+        def replace_ellipsis(item):
+            # Use slice(None) to replace Ellipsis.
+            # For var, var.shape = [3,4,5,6]
+            #
+            #   var[..., 1:2] -> var[:, :, :, 1:2]
+            #   var[0, ...] -> var[0]
+            #   var[0, ..., 1:2] -> var[0, :, :, 1:2]
+
+            item = list(item)
```
zhhsplendid, line 1878: I am not sure about the purpose of line 1878:
Is `item` always a list? If yes, you don't have to do this. If not, item = replace_ellipsis(item) would change the type of `item`, which is different to past code. Will this cause some problems?

liym27: `item` is a tuple and it can't be modified. In order to modify the value of `item`, and call the metheds `count()` and `index()` of `list`, here converts it to `list`. 

> If not, item = replace_ellipsis(item) would change the type of item, which is different to past code. Will this cause some problems?

No. Here, it's ok that `item` is a list after calling `replace_ellipsis`.

zhhsplendid (approved):
LGTM

luotao1 (approved):
LGTM for framework.py

PR #30846, opened 2021-02-02
Update paddle.static.Print with paddle2.0 api

Joejiong:
### PR types

Others
### PR changes

APIs
### Describe

Update paddle.static.Print with paddle2.0 api

test output preview:
image


zhhsplendid (approved):
LGTM

PR #30861, opened 2021-02-03
remove numpy array check in single-process dataloader

heavengate:
### PR types
Bug fixes

### PR changes
APIs

### Describe
user defined collate function may return data in list of Tensor format, do not check data in numpy array format in single-process dataloader


wangchaochaohu (approved):
LGTM

XieYunshen (approved):

PR #30869, opened 2021-02-03
[CustomOp] Check Compiler ABI compatibility

Aurelius84:
### PR types

Function optimization
### PR changes

Others
### 1. Check Compiler ABI compatibility 

+ **Linux platform**
  From 2.0.1, Paddle will use GCC >= 5.4.0, so we here set `GCC_MIN_VERSION=(5,4,0)`. And we will check whether user local environments satisfy this

+ **MacOS platform**
   There is no limitation for `clang/clang++` on MacOS

+ **Windows platform**
  We will add check after support windows.

Support `export PADDLE_SKIP_CHECK_ABI=1` to skip this check.

### 2. Add `interpreter` in load API

Support users to specific python interpreter alias or full python python. Because we reuse `setup.py` logic and  call `python`  command in subprocess. If user has install multiple version locally, we support to switch it by `interpreter` argument.

### 3. Add `verbose` argument
if set `verbose=True`, we will output inner log to users.


chenwhql (approved):
LGTM

PR #30877, opened 2021-02-03
Add Conv Transpose BF16

wozna:
### PR types

Others 
### PR changes

OPs 
### Describe

This PR: 
- change conv_tranpose op  mkldnn kernel to use MKLDNNHandlerT 
- add BF16 support for conv_tranpose op


paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -25,28 +25,26 @@ namespace operators {
...
+                              const platform::MKLDNNDeviceContext& dev_ctx,
+                              const mkldnn::engine mkldnn_engine,
+                              platform::Place cpu_place, const Tensor* input,
+                              const Tensor* filter, const Tensor* bias,
+                              Tensor* output, const std::string& unique_name)
+      : platform::MKLDNNHandlerT<T, mkldnn::deconvolution_forward>(
+            dev_ctx, mkldnn_engine, cpu_place,
+            platform::CreateKey(dev_ctx, framework::vectorize(input->dims()),
+                                unique_name)) {
     const bool is_test = ctx.Attr<bool>("is_test");
```
jczaja: In general I like very much this PR . The only thing that is missing is that inside a ConvTransposeMKLDNNHandler you should call isCached() method not to create MD again. Please look to other ops implemented with MKLDNNHandlerT like pool, softmax etc.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -241,10 +241,12 @@ class MKLDNNHandlerT {
     astream.wait();
   }
 
+  template <typename F = T>
   std::shared_ptr<mkldnn::memory> AcquireMemoryWithReorder(
       const mkldnn::memory::desc& user_md,
       const mkldnn::memory::desc& target_md, void* ptr,
-      const std::string& suffix, bool is_persistent = false) {
+      const std::string& suffix, bool is_persistent = false,
+      std::function<std::shared_ptr<F>(const F*)> custom_func = {}) {
```
arogowie-intel: Consider renaming `custom_func` to `custom_reorder_func`

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -253,6 +255,11 @@ class MKLDNNHandlerT {
         std::static_pointer_cast<dnnl::memory>(dev_ctx_.GetBlob(target_key));
 
     if (target_memory_p == nullptr) {
+      if (custom_func) {
```
arogowie-intel: If I understand still after this _custom reorder_ the condition `user_md != target_md` may be true and this would result in second reorder. Is it intentionally?

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -1484,5 +1491,17 @@ static void SetDstMemoryQuantized(
       new mkldnn::memory(*dst_md, engine, to_void_cast<T>(output_data)));
 }
 
+inline void GetWeightsTz(std::vector<int64_t>& weights_tz,  // NOLINT
```
arogowie-intel: The name of this function is vague for me. Please make it more explicit and verbose. What this `Tz` stands for? Since this functions is directly related to convolutions please note this fact in its name.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -25,28 +25,26 @@ namespace operators {
...
     PADDLE_ENFORCE_EQ(is_test, true,
                       platform::errors::InvalidArgument(
                           "ConvTransposeMKLDNN works only for inference. "
                           "Set is_test = True. but got is_test=False ."));
```
arogowie-intel:
```suggestion
                          "The attribute \'is_test\' value should be set to True, but got False."));
```

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -123,65 +121,23 @@ class ConvTransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
     std::transform(dilations.begin(), dilations.end(), dilations.begin(),
                    [](int64_t i) { return i - 1; });
 
-    const T* input_data = input->data<T>();
-    const T* filter_data = filter->data<T>();
+    const auto src_tz = paddle::framework::vectorize(input->dims());
 
-    auto src_tz = paddle::framework::vectorize<int64_t>(input->dims());
     auto iohw_weights_tz =
         paddle::framework::vectorize<int64_t>(filter->dims());
```
arogowie-intel: Template parameter is not necessary. The `int64_t` is already a default value.
```suggestion
        paddle::framework::vectorize(filter->dims());
```

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -123,65 +121,23 @@ class ConvTransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
 
-    std::vector<mkldnn::primitive> pipeline;
+    auto dst_tz = paddle::framework::vectorize<int64_t>(output->dims());
 
-    auto user_src_md = platform::MKLDNNMemDesc(
-        {src_tz}, platform::MKLDNNGetDataType<T>(), input->format());
-    auto user_weights_md = platform::MKLDNNMemDesc(
-        {weights_tz}, platform::MKLDNNGetDataType<T>(),
-        (g == 1) ? MKLDNNMemoryFormat::oihw : MKLDNNMemoryFormat::goihw);
+    const mkldnn::memory::dims stride_dims = strides;
```
arogowie-intel: Why for the second time copying the strides? It's already copied once here: https://github.com/PaddlePaddle/Paddle/pull/30877/files?file-filters%5B%5D=.cc#diff-cff6cbbbf012e4b6a91e16c01308876a81a1e409bf52d29d67741a7238b48a4eR94
Since, `mkldnn::memory::dims` type is just a typedef for `std::vector<int64_t>` you could use this type from the beginning.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -123,65 +121,23 @@ class ConvTransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    const auto mkldnn_paddings = platform::ToMkldnnPadding(paddings);
+    const mkldnn::memory::dims dilations_dims = dilations;
```
arogowie-intel: Same as above.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -192,78 +148,210 @@ class ConvTransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+    auto data_type = mkldnn::memory::data_type::f32;
+    if (ctx.Attr<std::string>("mkldnn_data_type") == "bfloat16" ||
+        std::is_same<T_out, platform::bfloat16>::value)
+      data_type = mkldnn::memory::data_type::bf16;
+
+    auto src_md =
+        platform::MKLDNNMemDesc(src_tz, data_type, chosen_memory_format);
+    auto weights_md =
+        platform::MKLDNNMemDesc(weights_tz, data_type, chosen_memory_format);
     std::vector<int64_t> bias_tz;
```
arogowie-intel: This variable is used only inside `if (bias) {` if statement branch. You could move it there.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -192,78 +148,210 @@ class ConvTransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
 
-    platform::ConvTransposeMKLDNNHandler handler(dev_ctx, mkldnn_engine, key);
-    // create a deconv(conv transpose) primitive descriptor and save it for
-    // usage in backward
-    std::shared_ptr<mkldnn::deconvolution_forward::primitive_desc>
-        conv_transpose_pd;
+    const mkldnn::primitive_attr conv_trans_attr =
+        CreatePostOps(fuse_activation, fuse_alpha, fuse_beta);
     auto fwd_prop_kind = is_test ? mkldnn::prop_kind::forward_inference
                                  : mkldnn::prop_kind::forward_training;
```
arogowie-intel: There is already check forcing `is_test` attribute to be true at the beginning . So is this ternary operator needed?

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -192,78 +148,210 @@ class ConvTransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+      const K* filter_data = filter->data<K>();
+      auto iohw_weights_tz =
+          paddle::framework::vectorize<int64_t>(filter->dims());
+      auto weights_tz = iohw_weights_tz;
+
+      // IOHW -> OIHW
+      weights_tz[0] = iohw_weights_tz[1];
+      weights_tz[1] = iohw_weights_tz[0];
+      int g = std::max(groups, 1);
+      platform::GetWeightsTz(weights_tz, g);
```
arogowie-intel: This is exactly the same code as in lines 126-134. Please make a function or reuse the result somehow.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
```diff
@@ -192,78 +148,210 @@ class ConvTransposeMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
+
+      auto user_src_md = platform::MKLDNNMemDesc(
+          weights_tz, platform::MKLDNNGetDataType<K>(),
+          (g == 1) ? filter->format() : MKLDNNMemoryFormat::goihw);
+
+      // Custom Reorder from IOHW to OIHW
```
arogowie-intel, line 257: Doesn't oneDNN support such reorder?

python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_transpose_bf16_mkldnn_op.py
```diff
@@ -0,0 +1,203 @@
...
+
+        if not self.force_fp32_output:
+            output = convert_float_to_uint16(output, self.attrs['data_format'])
+
+        self.outputs['Output'] = output
+
+
+class TestMKLDNNFuseBias(TestConv2DTransposeBF16MKLDNNOp):
+    def init_test_case(self):
+        TestConv2DTransposeBF16MKLDNNOp.init_test_case(self)
```
arogowie-intel: From py3 you can just call:
```suggestion
        super().init_test_case(self)
```
And in py2 it's rather better to use following syntax:
```
super(TestMKLDNNFuseBias, self).<called_method_from_base_class>(...)
```
Take a look here: https://stackoverflow.com/a/33469090

arogowie-intel (commented):
In general there is huge overlap with oneDNN conv kernel (not surprisingly) so it would be good to have this common part in one place. But such refactoring is rather better for another PR.

paddle/fluid/platform/mkldnn_reuse.h
wozna: Yes, it is intentionally. *Custom reorder function* is used to set an appropriate reorder for `data_format`. However, `user_md! = target_md` may differ in data type. In the case of bf16, the original weights (user_md) are float, while for calculations (target_md) we need them in bf16 and it is converted in this reorder.

paddle/fluid/platform/mkldnn_reuse.h
wozna: Good point. I think that `tz` is a shortcut to tensor and this is the nomenclature for the vector containing the dimensions. Function `GetWeightsTz` is related to the `group` attribute which for grouped convolution has to adapt this weight dimensions vector. I will rename this feature and add appropriate comments.

paddle/fluid/platform/mkldnn_reuse.h
arogowie-intel: Ok, then if the _additional_ thing you want to do is data type conversion I'd suggest to create a helper function for this task explicitly named anything close to `ConvertMemDataType`. With it there would be clear intention of control-flow logic: 1st reorder data, 2nd convert memory data type. This `AcquireMemoryWithReorder` function is already very complicated, very hard to understand it's control-flow and I suppose hard to debug. Not mentioning it's maintenance and testing. IMHO this function is doing to much things.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
wozna: We will be changing operators anyway to support training, so I think it's worth leaving it.

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
wozna: This is also related to group convolution, so we have to specify that format explicitly.

lidanqing-vv:
@wozna  
```
2021-02-10 02:15:11 ****************
2021-02-10 02:15:11 0. Unittest is not allowed to be disabled.
2021-02-10 02:15:11 You must have one RD (kolinwei(Recommend), or luotao1) approval for the usage of @unittest.skip or @unittest.skipIf.
2021-02-10 02:15:11 +@unittest.skipIf(not core.supports_bfloat16(),
2021-02-10 02:15:11 1. The error message you wrote in PADDLE_ENFORCE{_**} or PADDLE_THROW does not meet our error message writing specification. Possible errors include 1. the error message is empty / 2. the error message is too short / 3. the error type is not specified. Please read the specification [ https://github.com/PaddlePaddle/Paddle/wiki/Paddle-Error-Message-Writing-Specification ], then refine the error message. If it is a mismatch, please request chenwhql (Recommend), luotao1 or lanxianghit review and approve.
2021-02-10 02:15:11 The PADDLE_ENFORCE{_**} or PADDLE_THROW entries that do not meet the specification are as follows:
2021-02-10 02:15:11 PADDLE_ENFORCE_NE(input->format(), MKLDNNMemoryFormat::undef, + "Got wrong format for Input tensor.")); 
2021-02-10 02:15:11 2. Developers are not allowed to set the check_dygraph field directly, which is set to True by default. If you need to change the check_dygraph field, you must have one RD (phlrain (Recommend), fuyinno4 (Recommend for kunlun) or lanxianghit) review and approve. 
2021-02-10 02:15:11 The code that do not meet the specification are as follows:
2021-02-10 02:15:11  python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_transpose_bf16_mkldnn_op.py : 
2021-02-10 02:15:11 +        self.check_output(check_dygraph=(self.use_mkldnn == False)) 
2021-02-10 02:15:11 There are 3 approved errors.
2021-02-10 02:15:11 ****************
```

paddle/fluid/operators/mkldnn/conv_transpose_mkldnn_op.cc
jczaja: @arogowie-intel This code is reflecting the diffrence among oneDNN and PaddlePaddle in implementing groups. in oneDNN groups are another dimension e.g. shape without groups OIWH (weights) when there are more than one groups then It becomes GOIHW (5 dimensional). In PaddlePaddle OIHW and GOIHW are expressed in 4D data. It just weights of second group are glued (concatenated) to the end of weights of first group. That is why when groups are present we cannot rely on format inside tensor and we need to change from OIHW to GOIHW

paddle/fluid/platform/mkldnn_reuse.h
jczaja: @arogowie-intel custom function was introduced because there was no NCWH format enum in oneDNN so we needed to do reorder ourselves. This is outdated as after we impleented custom reorder relevant enum was added. So actual task is to remove custom reorder and there is a issue for that in our tracker.

jczaja (dismissed):
LGTM

arogowie-intel:
Looks good. Just a general note on `auto` - it deduces _plain_ type if you pass it a reference. Thus it entails a copy. Please pay attention to use `auto&` or `const auto&` whenever possible to avoid those copies.

jczaja (approved):

luotao1 (approved):

PR #30904, opened 2021-02-04
fix dataloader collate return list mix tensor and numpy array

heavengate:
### PR types
Bug fixes

### PR changes
APIs

### Describe
fix dataloader collate return list mix tensor and numpy array


wangchaochaohu (approved):
LGTM

PR #30929, opened 2021-02-05
support label with float input of cross_entropy, test=develop

chajchaj:
### PR types
Function optimization

### PR changes
APIs 

### Describe
support label with float input of cross_entropy


XiaoguangHu01 (approved):
LGTM

PR #31022, opened 2021-02-18
add trt transpose and flatten converter

cryoco:
### PR types

Performance optimization
### PR changes

Others
### Describe

1. add trt transpose and flatten converter;
2. make op teller able to utilize node ir infomation.


shangzhizhou (approved):
LGTM

PR #31034, opened 2021-02-18
Unification of BF16 enablement process 

wozna:
### PR types

Function optimization

### PR changes

Others 

### Describe

This PR unifies the addition of bfloat16 so that the process of adding quantize and dequantize reorders is similar to the quantization process. This way, pass `cpu_quantize_squash_pass.cc` can also be used in the bfloat16 process.

Therefore, the following changes have been added:
- `cpu_bfloat16_pass.cc` - adds quantize and dequantize before and after each operator which is marked as bfloat16
- `cpu_quantzie_squash_pass.cc`
- checks where we have `dequantize` and `quantize, then inserts `requantize` if scales are different or removes both operators when scale is the same (the latter is the most common in the case of bfloat16)
- removes dequantize op if it occurs before operators that has `force _fp32_output` attribute

Additionally, to `cpu_quantzie_squash_pass.cc` has been added:
- squash, where Scale operator and Quantize operator are merged
- squash where if we have Quantize operator before Conv2d bfloat16 op then remove Quantize because the input reorder can be executed inside the kernel

### DIFF
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -1829,9 +1829,8 @@ PDNode *patterns::OpDequant::operator()() {
   auto any_op = pattern->NewNode(any_op_repr())
                     ->assert_is_op()
                     ->assert_more([&](Node *node) {
-                      return (node->Op()->Type() == "matmul" ||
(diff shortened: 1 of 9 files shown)


jczaja (approved):
LGTM

luotao1 (approved):

PR #31040, opened 2021-02-19
[Paddle-TRT] support group_norm

cryoco:
### PR types

Function optimization
### PR changes

Others
### Describe

support group_norm op converter in paddle-TRT.


jiweibo (approved):
LGTM

PR #31049, opened 2021-02-19
[CustomOp] Refine name argument in setup

Aurelius84:
### PR types

New features
### PR changes

APIs
### What's New?

Refine name argument in setup.

Before this PR, users must specific two `name` arguments in `setup`, which will make them confuse the difference. Actually, the only important `name` is Package name, users don't need to know the shared libaray name from extension and it shares same name with Package name now.

**Example(Before):**
```python
setup(
    name='custom_relu2',   # Package Name
    ext_modules=[CUDAExtension(  
        name='custom_relu2',  # Shared library Name
        sources=['relu_op.cc', 'relu_op.cu'], 
       )]
)
```

**Example(After):**
```python
setup(
    name='custom_relu2',
    ext_modules=CUDAExtension(sources=['relu_op.cc', 'relu_op.cu'])
)
```

+ Only one name for package name
+ ext_modules support list or just one value

### DIFF
--- a/python/paddle/fluid/tests/custom_op/setup_build.py
+++ b/python/paddle/fluid/tests/custom_op/setup_build.py
@@ -23,10 +23,9 @@ use_new_custom_op_load_method(False)
 file_dir = os.path.dirname(os.path.abspath(__file__))
 
 setup(
-    name='relu2_op_shared',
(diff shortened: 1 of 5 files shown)


chenwhql (approved):
LGTM

PR #31051, opened 2021-02-19
Fix that convert_var_shape doesn't support slice like [0:]

zhhsplendid:
### PR types

Bug fixes
### PR changes

APIs
### Describe

As the title, when `slice_node` like `1:3` being passed to `idx` of `convert_var_shape`, it will cause syntax error because a function cannot take this as argument. This PR fixed it.


liym27 (approved):
LGTM

PR #31083, opened 2021-02-20
opt matmul and matmul_v2 on kunlun, *test=kunlun

wangleilei001:
### PR types
Function optimization

### PR changes
OPs

### Describe
opt matmul and matmul_v2 on kunlun, *test=kunlun


wangleilei001:
reopen one clean PR later

PR #31128, opened 2021-02-22
[CustomOp] Support attributes as func input in custom op

chenwhql:
### PR types

New features
### PR changes

Others
### Describe

[CustomOp] Support attributes as func input in custom op

Supported attr types include:
- `bool`
- `int`
- `float`
- `int64_t`
- `std::string`
- `std::vector<int>`
- `std::vector<float>`
- `std::vector<int64_t>`
- `std::vector<std::string>`

The backward func's attributes MUST be the subset of forward func's attributes.


Aurelius84 (approved):
LGTM

XieYunshen (approved):

PR #31136, opened 2021-02-23
add warning message when dtypes of operator are not same

MingMingShangTian:
### PR types

Others

### PR changes

Others

### Describe

When the dtypes of left variable and right variable are not the same, the operator compute (such as +-*/) will follow with the left variable dtypes. This PR add the warning message to the users.

The problem:
image
 
image


chenwhql (approved):
LGTM

PR #31141, opened 2021-02-23
[CustomOp] Add new paddle custom op so

chenwhql:
### PR types

Function optimization
### PR changes

Others
### Describe

[CustomOp] Add new paddle custom op so

Because the original `paddle_framework.so` constains some unneeded targets, such as `executor`, `recurrent_op` etc, which will cause inference loading so failed, so we add a new `custom_op.so` for new custom op module, and keep the original `paddle_framework.so` unchanged. 

we will remove `paddle_framework.so` in the future.


Aurelius84 (dismissed):

JiabinYang (dismissed):
LGTM

Shixiaowei02 (dismissed):

chalsliu (dismissed):

zhwesky2010 (approved):
LGTM

chalsliu (approved):

PR #31143, opened 2021-02-23
support adding correct npu op in pybind.h

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

support adding correct npu op in pybind.h
image


liym27 (approved):
LGTM

PR #31157, opened 2021-02-23
Add cublas_handle() to expose cublas_handle to ops

FrostML:
### PR types

Others

### PR changes

Others

### Describe

Add `cublas_handle()` to expose `cublas_handle` for ops.
This is used for custom op faster transformer.


paddle/fluid/platform/device_context.cc
```diff
@@ -459,6 +459,10 @@ cudnnHandle_t CUDADeviceContext::cudnn_handle() const {
   return context()->CudnnHandle();
 }
 
+cublasHandle_t CUDADeviceContext::get_cublas_handle() const {
```
zhiqiu: Suggest get_cublas_handle() -> cublas_handle(), keep consistent with other function.

FrostML: Done. Thanks.

zhiqiu (approved):
LGTM

PR #31209, opened 2021-02-25
add int pad support for Pad1D/2D/3D

littletomatodonkey:
### PR types
Function optimization
### PR changes
APIs
### Describe
Add support for `int type pad` in nn.Pad1D/2D/3D.

### DIFF
--- a/python/paddle/nn/layer/common.py
+++ b/python/paddle/nn/layer/common.py
@@ -38,6 +38,13 @@ __all__ = [
 ]
 
 
+def _npairs(x, n):
+    if isinstance(x, (paddle.Tensor, list)):
+        return x
+    x = [x] * (n * 2)
+    return x
+
+
 class Linear(layers.Layer):
     r"""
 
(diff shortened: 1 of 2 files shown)


python/paddle/nn/layer/common.py
```diff
@@ -915,7 +922,8 @@ class Pad1D(layers.Layer):
     If mode is 'reflect', pad[0] and pad[1] must be no greater than width-1.
 
     Parameters:
-        padding (Tensor | List[int32]): The padding size with data type int32. [len(padding)/2] dimensions
+        padding (Tensor | List[int32], int): The padding size with data type int32. If is int, use the
```
willthefrog: unrelated to current PR but `int32` does not seem correct, by default python 3 int is int64

willthefrog (approved):
LGTM

python/paddle/nn/layer/common.py
littletomatodonkey: done, thx~

TCChenlong (approved):
LGTM

ceci3 (approved):
LGTM

PR #31231, opened 2021-02-25
New bf

jakpiase:
### PR types
[ New features]

### PR changes
[ OPs ] 

### Describe
Added support for LSTM bf16, fixed GRU bf16 tests so weights are also in bf16 format.


jakpiase:
Need to do rebase, because github shows also files that were not changed.

PR #31257, opened 2021-02-26
[Custom OP] Support stream set on Custom Op

JiabinYang:
### PR types

Function optimization
### PR changes

APIs
### Describe

This PR support stream get from Tensor which has been set stream during RunKernelFunc. This may solve performance issue caused by stream sync in GPU mode


chenwhql (approved):
LGTM

Aurelius84 (approved):

PR #31342, opened 2021-03-01
support trt serialize when load model from memory

shangzhizhou:
### PR types
Others

### PR changes
Others

### Describe
support tensorRT serialization when load model from memory。


cryoco (approved):
LGTM

PR #31139, opened 2021-02-23
add trt transpose and flatten converter (#31022)

cryoco:
### PR types

Performance optimization
### PR changes

Others
### Describe

cherry-pick [PR#31022](https://github.com/PaddlePaddle/Paddle/pull/31022)

PR #31343, opened 2021-03-01
Revert "add trt transpose and flatten converter (#31022) (#31139)"

cryoco:
### PR types

Others
### PR changes

Others
### Describe

Revert "add trt transpose and flatten converter (#31022) (#31139)"


shangzhizhou (approved):
LGTM

PR #31367, opened 2021-03-02
[Paddle-TRT] Add Paddle-TRT affine channel support (duplicated)

zlsh80826:
### PR types

New features

### PR changes

OPs

### Describe

Support affine channel ops on paddle-trt


zlsh80826:
update pr in #31628

PR #31379, opened 2021-03-02
[Paddle-TRT] conv2d converter support SAME padding

cryoco:
### PR types

Performance optimization
### PR changes

Others
### Describe

trt conv2d converter support SAME padding


shangzhizhou (approved):
LGTM

PR #31411, opened 2021-03-03
Fix bug for set_value op when input dtype is not float32

liym27:
### PR types

Bug fixes
### PR changes

OPs
### Describe

As the title


zhhsplendid (approved):
LGTM

PR #31415, opened 2021-03-03
[Dy2stat] Fix Read-Only Attribute as while_loop Output

zhhsplendid:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix Read-Only Attribute as while_loop Output:

Usually, our convert_while_loop will be like:
```
    [a, b, c] = paddle.jit.dy2static.convert_while_loop(
            condition_name, body_name, [a, b, c])
```
where a, b, c are in loop_var_names.

However, if loop_var_names contains property such as foo.x, we cannot
assign the attribute as output of convert_while_loop because Python
property is a kind of read-only attribute. To handle the case, we replace
the attributes which are output of convert_while_loop with generated
variables, then if we know the attribute is not read-only at runtime, we
assign the attribute. The created statements are like:
```
    [a, b, __attribute_variable_1] = paddle.jit.dy2static.convert_while_loop(
            condition_name, body_name, [a, b, foo.x])
    if not isinstance(getattr(type(foo), x, None), property): foo.x = __attribute_variable_1
```


python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
```diff
@@ -39,23 +39,73 @@
...
+    controlflow while_loop.
+
+    Usually, the list just contain 1 statement such as:
+
+    [a, b, c] = paddle.jit.dy2static.convert_while_loop(
+            condition_name, body_name, [a, b, c])
+
+    where a, b, c are in loop_var_names.
+
+    However, if loop_var_names contains attribute such as foo.x, we cannot
```
zhhsplendid, line 57: Here I had a comment error: "attribute" -> "property"

I will change it in the near future (for example, after reviewer reviews it, or make it as test=document_fix in the next PR) because current commit passes CI tests.

liym27 (approved):
LGTM

PR #31417, opened 2021-03-03
Update hapi to  support AMP

LiuChiachi:
### PR types

New features
### PR changes

APIs
### Describe

Add AMP support for High-level API.

#### API 
```python
def prepare(self,
     optimizer=None, loss=None, metrics=None,
     amp_configs=None): # new  parameters for `Model.prepare()`
```
`amp_configs (str|dict|None)`: AMP configurations. If AMP or pure float16 training is used, the key 'level' of 'amp_configs' should be set to 'O1' or 'O2' respectively. Otherwise, the value of 'level' defaults to 'O0', which means float32 training. In addition to 'level', users could pass in more parameters consistent with mixed precision API. The supported keys are: 'init_loss_scaling', 'incr_ratio', 'decr_ratio', 'incr_every_n_steps', 'decr_every_n_nan_or_inf', 'use_dynamic_loss_scaling', 'custom_white_list', 'custom_black_list', and 'custom_black_varnames'or 'use_fp16_guard' is only supported in static mode. Users could refer to mixed precision API documentations :ref:`api_paddle_amp_auto_cast` and  :ref:`api_paddle_amp_GradScaler` for details. For convenience, 'amp_configs' could be set to 'O1' or 'O2' if no more
parameters are needed. 'amp_configs' could be None in float32 training. Default: None.

#### API description

When training on GPU, auto mixed precision (AMP) training is supported, and pure float16 training is also supported in static mode while using Adam, AdamW and Momentum optimizer. Before using pure float16 training, `multi_precision` could be set to True when creating optimizer, which can avoid poor accuracy or slow convergence in a way, and inputs of dtype float
should be cast to float16 by users. Users should also use `paddle.static.amp.fp16_guard` API to limit the range of pure float16 training, otherwise, 'use_fp16_guard' should be set to False by users. However, limiting the range of is not supported during training using AMP.

#### Example code:
```python

import paddle
import paddle.nn as nn
import paddle.vision.transforms as T

def run_example_code():
  device = paddle.set_device('gpu')

  net = nn.Sequential(nn.Flatten(1), nn.Linear(784, 200), nn.Tanh(),
                      nn.Linear(200, 10))

  model = paddle.Model(net)
  optim = paddle.optimizer.SGD(learning_rate=1e-3, parameters=model.parameters())

  amp_configs = {
      "level": "O1",
      "custom_white_list": {'conv2d'},
      "use_dynamic_loss_scaling": True
  }
  model.prepare(optim,
      paddle.nn.CrossEntropyLoss(),
      paddle.metric.Accuracy(),
      amp_configs=amp_configs)

  transform = T.Compose([T.Transpose(), T.Normalize([127.5], [127.5])])
  data = paddle.vision.datasets.MNIST(mode='train', transform=transform)
  model.fit(data, epochs=2, batch_size=32, verbose=1)

# mixed precision training is only supported on GPU now.
if paddle.is_compiled_with_cuda():
  run_example_code()
```


python/paddle/hapi/model.py
```diff
@@ -1255,7 +1348,19 @@ def prepare(self, optimizer=None, loss=None, metrics=None):
                 It can be None when there is no loss.
             metrics (Metric|list of Metric|None): If metrics is set, all
                 metrics will be calculated and output in train/eval mode.
-
+            amp_configs (dict|None): AMP configurations. If AMP or pure
```
zhiqiu: I think it is better to make a class for amp_config, which can limit the key-values that are accepted by amp_configs.

python/paddle/hapi/model.py
```diff
@@ -649,19 +681,30 @@ def train_batch(self, inputs, labels=None):
         labels = labels or []
         labels = [to_variable(l) for l in to_list(labels)]
 
-        if self._nranks > 1:
-            outputs = self.ddp_model.forward(* [to_variable(x) for x in inputs])
-        else:
-            outputs = self.model.network.forward(
-                * [to_variable(x) for x in inputs])
+        scaler = paddle.amp.GradScaler(**self._amp_configs)
```
zhiqiu: Are all key-values in self._amp_configs accepted by GradScaler?

python/paddle/hapi/model.py
```diff
@@ -649,19 +681,30 @@ def train_batch(self, inputs, labels=None):
...
+        with paddle.amp.auto_cast(
+                enable=False if self._amp_level == 'O0' else True,
+                **self._amp_custom_lists):
```
zhiqiu: similar above

python/paddle/hapi/model.py
```diff
@@ -649,19 +681,30 @@ def train_batch(self, inputs, labels=None):
...
-        losses = to_list(losses)
-        final_loss = fluid.layers.sum(losses)
-        final_loss.backward()
+            losses = self.model._loss(*(to_list(outputs) + labels))
+            losses = to_list(losses)
+            final_loss = fluid.layers.sum(losses)
 
-        self.model._optimizer.minimize(final_loss)
-        self.model.network.clear_gradients()
+        if self._amp_level != "O0":
```
zhiqiu, line 696: dygraph does not support 02 yet?

LiuChiachi: When O2 is chosen in dygraph, it  would be treated as O1, and users would be warned. Is this correct?

python/paddle/hapi/model.py
LiuChiachi: Thanks. If some key-values are not accepted, the program will break down here. It is necessary to catch the wrong key-values earlier.

python/paddle/hapi/model.py
```diff
@@ -852,9 +907,12 @@ class Model(object):
           model = paddle.Model(net, input, label)
           optim = paddle.optimizer.SGD(learning_rate=1e-3,
               parameters=model.parameters())
+
+          amp_configs = {'level': 'O0'}
           model.prepare(optim,
                         paddle.nn.CrossEntropyLoss(),
-                        paddle.metric.Accuracy())
+                        paddle.metric.Accuracy(),
+                        amp_configs=amp_configs)
```
qingqing01: better to add another example for amp

python/paddle/hapi/model.py
```diff
@@ -1255,7 +1370,22 @@ def prepare(self, optimizer=None, loss=None, metrics=None):
...
+                should be set to 'O1' or 'O2' respectively. Otherwise, the
+                value of 'level' defaults to 'O0', which means training without
+                using AMP or pure float training. In addition to 'level',
+                users could pass in more parameters consistent with low-level
+                API. The supported keys are: 'init_loss_scaling', 'incr_ratio',
+                'decr_ratio', 'incr_every_n_steps', 'decr_every_n_nan_or_inf',
+                'use_dynamic_loss_scaling', 'custom_white_list',
+                'custom_black_list', and 'custom_black_varnames',
+                'use_pure_fp16' or 'use_fp16_guard' is only supported in static
+                mode. Users could refer to low-level API documentations
```
qingqing01: remove  'low-level'

python/paddle/hapi/model.py
```diff
@@ -1241,7 +1329,64 @@ def parameters(self, *args, **kwargs):
...
+                    "Grad clip is not supported in pure float16 training now, and it will be supported in future version."
+
+        if self._adapter._amp_level != 'O0' and self._adapter._amp_configs:
+            for param_name in [
+                    'custom_white_list', 'custom_black_list',
+                    'custom_black_varnames'
+            ]:
+                if param_name in self._adapter._amp_configs:
+                    self._adapter._amp_custom_lists[
+                        param_name] = amp_configs.pop(param_name)
```
guoshengCS: Better not to pop from input dict `amp_configs` and treat it as constant, since the user defined input might be reused elsewhere, such as another amp model.

python/paddle/hapi/model.py
```diff
@@ -1255,7 +1400,22 @@ def prepare(self, optimizer=None, loss=None, metrics=None):
                 It can be None when there is no loss.
             metrics (Metric|list of Metric|None): If metrics is set, all
                 metrics will be calculated and output in train/eval mode.
-
+            amp_configs (dict|None): AMP configurations. If AMP or pure
+                float16 training is used, the key 'level' of 'amp_configs'
+                should be set to 'O1' or 'O2' respectively. Otherwise, the
+                value of 'level' defaults to 'O0', which means training without
```
guoshengCS: Maybe also need support for string `O1` and `O2` directly to keep concise,  if it is the most common case.

python/paddle/hapi/model.py
```diff
@@ -1255,7 +1423,23 @@ def prepare(self, optimizer=None, loss=None, metrics=None):
                 It can be None when there is no loss.
             metrics (Metric|list of Metric|None): If metrics is set, all
                 metrics will be calculated and output in train/eval mode.
-
+            amp_configs (str|dict|None): AMP configurations. If AMP or pure
+                float16 training is used, the key 'level' of 'amp_configs'
+                should be set to 'O1' or 'O2' respectively. Otherwise, the
+                value of 'level' defaults to 'O0', which means training without
```
wzzju: Maybe `which means float32 training. ` is more suitable.

python/paddle/hapi/model.py
```diff
@@ -598,6 +620,10 @@ def _compile_and_initialize(self, prog, mode):
                 startup_prog = self._startup_prog._prune(uninitialized)
                 self._executor.run(startup_prog)
 
+        if self._amp_level == "O2" and mode == 'train' and core.is_compiled_with_cuda(
+        ) and self._nranks <= 1:
```
wzzju: If `self._nranks > 1`, `self.model._optimizer.amp_init(place)` is also required under "O2".

python/paddle/hapi/model.py
```diff
@@ -1241,7 +1333,83 @@ def parameters(self, *args, **kwargs):
...
+                if param_name in amp_config_key_set:
+                    self._adapter._amp_custom_lists[param_name] = amp_configs[
+                        param_name]
+                    amp_config_key_set -= {param_name}
+
+        def _check_amp_configs(amp_config_key_set):
+            accepted_param_set = {
+                'init_loss_scaling', 'incr_ratio', 'decr_ratio',
+                'incr_every_n_steps', 'decr_every_n_nan_or_inf',
+                'use_dynamic_loss_scaling', 'use_fp16_guard', 'use_pure_fp16'
```
wzzju: If the user set `level` to `O2`, but set `use_pure_fp16` to `False`, what will happen?
Or, set `level` to `O1`, but set `use_pure_fp16` to `True`,  what will happen?

LiuChiachi: The value of  `use_pure_fp16` would be ignored, and the true `level` only depends on what the user pass to `amp_configs['level']` or `amp_configs(str)`.  I will disable `use_pure_fp16` this parameters.
```python
  if 'use_pure_fp16' in amp_configs:
      raise ValueError("''use_pure_fp16' is an invalid parameter,"
          "the level of mixed precision training only depends on 'O1' or 'O2'.")
```

python/paddle/hapi/model.py
LiuChiachi: Thanks. Done. I upgraded the usage of Fleet API and fix this.
Before this, old Fleet APIs we used don't have `amp_init` function and I solved this in a wrong way.
Thanks:)

python/paddle/hapi/model.py
LiuChiachi: Thanks! Done.

python/paddle/hapi/model.py
LiuChiachi: Thanks! Done.

python/paddle/hapi/model.py
LiuChiachi: Thanks! Done.

python/paddle/hapi/model.py
LiuChiachi: Thanks! Done.

python/paddle/hapi/model.py
LiuChiachi: Thanks! Done.

python/paddle/hapi/model.py
LiuChiachi: Thanks! Done. I added input check for `amp_configs`. Thanks:)

saxon-zh (approved):
LGTM

zhiqiu (approved):
LGTM

XiaoguangHu01 (approved):
LG API

kolinwei (approved):

qingqing01 (approved):
LGTM

PR #31433, opened 2021-03-04
[NPU] Use FillD instead of Broadcast

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

It seems `Broadcast` is not worked on CANN 20.1, so use `FillD` instead.

image


liym27 (approved):
LGTM

PR #31438, opened 2021-03-04
[Paddle-TRT] Fix TRT serialization on windows

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix TRT serialization on windows by changing fstream mode to std::ios::binary.
Linux seems also good with std::ios::binary, so there's no need to set seperately.


shangzhizhou (approved):
LGTM

PR #31482, opened 2021-03-08
[Bug fix] remove md5 check since different machine generate different binary file

lidanqing-vv:
### PR types
Bug fixes

### PR changes
Others

### Describe
Found that in icelake, generate binary file is slightly different in digits (represented float value should be very close), so the md5 is different. We can not use same md5 for different machines.


lidanqing-vv:
@wojtuss Hi, Please review

wojtuss (approved):
LGTM

PR #31491, opened 2021-03-09
[ROCM] fix softmax with loss nan in HIP platform, test=develop

qili93:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix loss nan in ROCM platform. dim size and thread size should less than 256, or will have loss nan.


GaoWei8 (approved):
LGTM for op benchamrk ci

PR #31507, opened 2021-03-09
[DO NOT MERGE] Support acl sum op

oyxuan-11:


paddle/fluid/operators/sum_op_npu.cc
```diff
@@ -0,0 +1,65 @@
...
+      x_2[i] = *x[i];
+    }
+
+    auto* out = ctx.Output<Tensor>("Out");
+
+    auto place = ctx.GetPlace();
+
+    out->mutable_data<T>(ctx.GetPlace());
+
+    auto runner = NpuOpRunner("AddN", x_2, {*out}, {});
```
liym27, line 45: Why not use "Add" according to `SumParser` ? 
https://github.com/PaddlePaddle/Paddle/blob/ascendrc/python/paddle/distributed/fleet/meta_optimizers/ascend/ascend_parser.py#L446

PR #31513, opened 2021-03-10
[Paddle-TRT] Fix engine key in trt int8 calibration

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

TRT serialization is affected by max_batch_size and pricision, but calibration is not. So it is necessary to set different engine key in serialization and calibration.


shangzhizhou (approved):
LGTM

PR #31558, opened 2021-03-11
[oneDNN] lookup_table op with support for BF16 data type.

arogowie-intel:
### PR types
New features

### PR changes
OPs

### Describe
This PR adds support for BF16 data type in `lookup_table` op for both forward and backward pass.


arogowie-intel:
@jczaja @wozna @wojtuss @arlesniak @lidanqing-intel Please start your review.

python/paddle/fluid/tests/unittests/test_lookup_table_bf16_op.py
```diff
@@ -0,0 +1,193 @@
...
+    def setUp(self):
+        self.op_type = "lookup_table"
+        self.dtype = np.uint16
+
+        table = np.random.random((17, 31)).astype("float32")
+        self.ids = np.random.randint(0, 17, (4, 1)).astype("int64")
+
+        self.w_bf16 = convert_float_to_uint16(table)
+        self.out_bf16 = self.w_bf16[self.ids.flatten()]
+        self.out_fp32 = table[self.ids.flatten()]
```
arlesniak: Please make an extra variable with ids already flattened as it's unnecessary flattened a couple of times around the code. It can be passed to _get_grad as well. For example:
self.ids_flattened = self.id.flatten()
self.out_bf16 = self.w_bf16[self.ids_flattened]
self.out_fp32 = table[self.ids_flattened]
_get_grad(table, self.ids_flattened) and inside _get_grad there would be no need calling flatten

python/paddle/fluid/tests/unittests/test_lookup_table_bf16_op.py
```diff
@@ -0,0 +1,193 @@
...
+
+def _lookup(weights, ids):
+    w_shape = weights.shape
+    out_shape = list(ids.shape[:-1])
+    out_shape.append(w_shape[-1])
+    out = weights[ids.flatten()].reshape(out_shape)
+    return out
+
+
+def _get_grad(weights, ids):
```
arlesniak: Please take into consideration moving _get_grad and lookup functions and common parts of TestLookupTableBF16Op and TestLookupTableBF16OpIds4D to a new parent Test class  just like you did nicely with TestLookupTableBF16OpWIsSelectedRows.

arlesniak (changes requested):

python/paddle/fluid/tests/unittests/test_lookup_table_bf16_op.py
arogowie-intel: Did some refactoring to more reuse code, however the two functions `_lookup` and `_get_grad` are more handy as standalone functions.

wozna (approved):
Well done python tests :) 
LGTM

paddle/fluid/operators/lookup_table_op.h
```diff
@@ -102,7 +102,8 @@ class LookupTableKernel : public framework::OpKernel<T> {
             auto id_index = table_t.GetIndexFromId(ids[i]);
 
             if (id_index != -1) {
-              if (input_data_type == framework::proto::VarType::INT8) {
+              if (input_data_type == framework::proto::VarType::INT8 ||
+                  input_data_type == framework::proto::VarType::BF16) {
```
luotao1, line 106: I wonder why do you change `lookup_table_op` but not `lookup_table_v2_op`?

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -33,10 +33,19 @@
 from paddle.fluid.op import Operator
 from paddle.fluid.executor import Executor
 from paddle.fluid.framework import Program, OpProtoHolder, Variable
-from testsuite import create_op, set_input, append_input_output, append_loss_ops
+from paddle.fluid.tests.unittests.testsuite import (
```
luotao1, line 36: why change this file?

paddle/fluid/operators/lookup_table_op.h
arogowie-intel: That's because in the model "word2vec" there is `lookup_table` op used. Moreover [`paddle.fluid.layers.embedding`](https://github.com/PaddlePaddle/Paddle/blob/d3ec88ad2b22515038618df4404955711650f92d/python/paddle/fluid/layers/nn.py#L510-L520) Py API creates `lookup_table` op.

python/paddle/fluid/tests/unittests/op_test.py
arogowie-intel: Since this is the recommended Python style for importing multiple modules from single package [PEP328](https://www.python.org/dev/peps/pep-0328/#id7) and generally it looks more readable. Moreover this way you can have multiline import statement.

arogowie-intel:
@luotao1 regarding issues rised in PR-CI-APPROVAL:

0.  > `@unittest.skipIf(not core.supports_bfloat16(),`

    I need to use this decorator, since otherwise the CI configurations with GPU will throw errors about no kernel registered for this datatype for GPU.
1. | `self.check_output_with_place(core.CPUPlace(), check_dygraph=False)`

    The `check_dygraph=False` is needed to pass tests. The BF16 data type is not supported in dygraph mode.
2. | `max_relative_error=1.5e-2,`

    Usage of BF16 data type might incur slightly lower accuracy in comparison to FP32.
3. | `@skip_check_grad_ci(`

    This is analogous as in tests [here](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/test_lookup_table_op.py#L60) - citing: "gradient of paddings makes no sense."

arlesniak (approved):
Good job!

luotao1 (approved):

PR #31566, opened 2021-03-12
[3D-parallel] add 1f1b scheduler for pipeline

sandyhouse:
### PR types

New features
### PR changes

Others
### Describe

add 1f1b scheduler for pipeline, which intervals forward phase and backward phase to save memory consumption. 

Now, two pipeline schedulers are supported: F-then-B and 1F1B.
With F-then-B scheduler, the scheduler does forward for all microbatches and then does backward for all microbatches, and then does updating.

With 1F1B scheduler, after a startup stage, the scheduler does one forward then one backward alternatively which may save GPU memory.

How to use:

```
dist_strategy = DistributedStrategy()
dist_strategy.pipeline = True
dist_strategy.pipelie_configs = {
  'schedule_mode': '1F1B', # or 'F-then-B'
}
```

TODO:
1. split program into forward program, backward program and update program
2. check all ops have the attribute of op_role.

### DIFF
--- a/paddle/fluid/framework/section_worker.cc
+++ b/paddle/fluid/framework/section_worker.cc
@@ -113,7 +113,8 @@ void SectionWorker::TrainFiles() {
   }
 
   if (schedule_mode_ == 0) {
-    // F-then-B scheduler
+    // F-then-B scheduler which runs Forward phase for all microbatches,
+    // then runs Backward phase for all microbatches.
     // step1: run forward
     for (int i = 0; i < num_microbatches_; ++i) {
       RunForward(i, gc, unused_vars_);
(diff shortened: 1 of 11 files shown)


paddle/fluid/framework/section_worker.cc
```diff
@@ -22,15 +22,79 @@ class TrainerDesc;
...
 
+void SectionWorker::RunForward(
+    int micro_id, std::unique_ptr<GarbageCollector> &gc,
+    std::unordered_map<const OperatorBase *, std::vector<std::string>>
+        &unused_vars_) {
+  for (auto &op : ops_) {
+    int op_role = op->Attr<int>(std::string("op_role"));
+    // We run op with op_role = kLRSched only for the first microbatch
+    // to avoid increasing the @LR_DECAY_STEP@ multiple times.
+    bool run_first_mbatch = op_role == static_cast<int>(OpRole::kForward) ||
```
wangxicoding, line 42: 最好加上括号确保优先级的清晰明了

wangxicoding (approved):
LGTM

paddle/fluid/framework/section_worker.cc
sandyhouse: I will do it in the next pr.

PR #31588, opened 2021-03-12
[CustomOp] Normalized function parameter writing

chenwhql:
### PR types

Function optimization
### PR changes

APIs
### Describe

[CustomOp] Normalized function parameter writing

For all function inputs, change to use `const &` format:
- Make the function writing conform to the programming style guide
- Reduce copying and improve efficiency
- At the same time, it is compatible with the written method that has been released in 2.0.1


JiabinYang (approved):
LGTM

PR #31589, opened 2021-03-12
【3D-parallel】Add SocketServer to persist server 

wangxicoding:
### PR types

Bug fixes
### PR changes

Others
### Describe

When using c_gen_nccl_id op multi times, the port may be occupied by others, so persist the server to prevent this problem.


sandyhouse (approved):
LGTM

PR #31607, opened 2021-03-12
[NPU] Fix gather_grad bug and add test

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

[NPU] Fix gather_grad bug and add test

The input X of gather_grad is set as NoNeedBuffer and it may result in bugs in the original implementation.


xymyeah (approved):
LGTM

PR #31627, opened 2021-03-15
[Paddle-TRT] Support batch axis concatenation when using dynamic shape

zlsh80826:
### PR types

New features
### PR changes

OPs
### Describe

support batch axis concatenation when using dynamic shape


cryoco (approved):
LGTM

PR #31631, opened 2021-03-15
[Paddle-TRT] Run radix sort of proposals layers on the context stream

zlsh80826:
### PR types

Performance optimization

### PR changes

OPs
### Describe

1. Run radix sort of proposals layers on the context stream. Without indicating the stream, functions will be placed on default stream, which introduces additional implicit synchronization overhead.
2. The is ~1.5% e2e performance gain on trt-fp16/maskrcnn inference


jerrywgz (commented):
./paddle/fluid/operators/detection/generate_proposals_v2_op.cu also uses SortDescending, which may applied this optimization as well?

zlsh80826:
> ./paddle/fluid/operators/detection/generate_proposals_v2_op.cu also uses SortDescending, which may applied this optimization as well?

Yes, the modification is on bbox_utils.cu, which is used by `generate_proposals` and `generate_proposals_v2`

jerrywgz (approved):
LGTM

PR #31632, opened 2021-03-15
[NPU] fix allocator min chunk size

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

[NPU] fix allocator min chunk size, from 256 -> 512 bytes, required by Ascend NPU


pangyoki (approved):
LGTM

PR #31634, opened 2021-03-15
[Paddle-TRT] NMS Performance Optimization

zlsh80826:
### PR types

Performance optimization

### PR changes

OPs

### Describe

1. Replace mask vector to raw ptr in NMS:
Original implementation introduces an unnecessary memory copy of mask from host to device before launching NMSKernel
Origin:
image
This PR:
image

2. Launch NMSKernel on context stream to prevent implicit synchronization with default stream
3. The is ~1% e2e performance gain on trt-fp16/maskrcnn inference

### DIFF
--- a/paddle/fluid/operators/detection/bbox_util.cu.h
+++ b/paddle/fluid/operators/detection/bbox_util.cu.h
@@ -275,7 +275,6 @@ static void NMS(const platform::CUDADeviceContext &ctx, const Tensor &proposals,
 
   const T *boxes = proposals.data<T>();
   auto place = BOOST_GET_CONST(platform::CUDAPlace, ctx.GetPlace());
-  framework::Vector<uint64_t> mask(boxes_num * col_blocks);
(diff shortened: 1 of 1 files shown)


cryoco (dismissed):
LGTM

paddle/fluid/operators/detection/bbox_util.cu.h
```diff
@@ -276,14 +276,19 @@ static void NMS(const platform::CUDADeviceContext &ctx, const Tensor &proposals,
   const T *boxes = proposals.data<T>();
   auto place = BOOST_GET_CONST(platform::CUDAPlace, ctx.GetPlace());
   framework::Vector<uint64_t> mask(boxes_num * col_blocks);
```
jerrywgz: mask could be removed

zlsh80826: done

jerrywgz (approved):
LGTM

PR #31695, opened 2021-03-17
[NPU] add npu kernel for concat op

zhiqiu:
### PR types

New features
### PR changes

OPs
### Describe

[NPU] add npu kernel for concat op

image
image


liym27 (approved):
LGTM

PR #31701, opened 2021-03-17
Polish tensor pipeline

heavengate:
### PR types
Function optimization

### PR changes
APIs

### Describe
**Polish tensor pipeline**
get shared memory LoDTensor from paddle.Tensor directly, do not need to call `.numpy()`


qingqing01 (approved):
LGTM

LielinJiang (approved):

PR #31713, opened 2021-03-18
[CustomOp] Support attribute in infershape function

chenwhql:
### PR types

New features
### PR changes

Others
### Describe

[CustomOp] Support attribute parameter in infershape function

InferShapeFn's attribute list must be exactly same as Forward KernelFn's attribute list except `std::vector<int64_t>` attribute. InferShape can't support `std::vector<int64_t>` attr type, because the input type is `std::vector<int64_t>`, only can use one rule to parse `std::vector<int64_t>` parameter


Aurelius84 (approved):
LGTM

zhwesky2010 (approved):

PR #31800, opened 2021-03-23
Fix the problem when rnns compatible with cudnn's parameters are created again

iclementine:
### PR types

Bug fixes

### PR changes

Others

### Describe

Test a typical case in static graph usage: create two nearly identical program with a shared startup program to share their parameters.
 
When creating a parameter, the name is checked. If there is already a parameter with the same name, which is the output of a operator(i.e. its creator), its re-creation is skipped.

But if that parameter has been the output of more than one operators, an exception is raised. For special cases, white list is added. Flattening rnn's parameters for the need to call cudnn kernel is such a case. When a SimpleRNN/GRU/LSTM is compatible with cudnn kernel, the parameters are packed together for the need of cudnn kernel. These parameters are first created by their creators, then they are also the ouput of the **packing** operator (a.k.a coalesce_tensor). 

The parameters' data are possibly copied to a new memory block(packed together), a new tensor pointing to the memory block is created. Also the original parameters' data pointer is pointed to the new address.

So we need to add a white list when creating parameter with the same name twice.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -3031,7 +3031,11 @@ class Block(object):
                         # In startup_program, "c_broadcast" and "c_sync_comm_stream"
                         # are treated as initialization ops that cause error.
                         # Think of "c_broadcast" and "c_sync_comm_stream" as a special case here.
-                        if op.type in ["c_broadcast", "c_sync_comm_stream"]:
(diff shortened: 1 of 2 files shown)


chenwhql (approved):
LGTM

guoshengCS (approved):
LGTM

PR #31806, opened 2021-03-23
[Dy2stat] Fix the bug that loop_body_func may return single element

zhhsplendid:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Our old `loop_body` function may return single element when loop_vars just contains only 1 element, which can cause bug. Let's see an example, the origin code is the new testcase TestForwardContainsForLayer of this PR. The translated code would be like below before this PR

```
def forward(self, x):
    y = paddle.zeros([10, 2, 3])
    z = []
    i = 0

    def for_loop_condition_0(i):
        return i < self.high - self.low

    def for_loop_body_0(i):
        paddle.jit.dy2static.convert_call(z.append)(paddle.jit.dy2static.
            convert_call(y[i].clone)())
        i += 1
        return i # This PR changes `return i` to `return i,` as a tuple

    [i] = paddle.jit.dy2static.convert_while_loop(for_loop_condition_0,
        for_loop_body_0, [i])
    return z
```

You could notice the `for_loop_body_0` returns `i`, then when the python for loop is running, the loop_vars `[i]` would be updated to `i`, which can cause bug.

The key point of this PR is forcing `loop_body` functions always return tuple.

### DIFF
--- a/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
+++ b/python/paddle/fluid/dygraph/dygraph_to_static/loop_transformer.py
@@ -594,7 +594,7 @@ class LoopTransformer(gast.NodeTransformer):
         # append return values for loop body
         body_stmts.append(
             gast.Return(value=generate_name_node(
-                loop_var_names, ctx=gast.Load())))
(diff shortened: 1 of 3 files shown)


liym27 (approved):
LGTM

PR #31819, opened 2021-03-23
LOG CLEAN

seiriosPlus:
### PR types

Function optimization
### PR changes

Others
### Describe

1. vlog clean.
2. brpc log clean
3. train from dataset fetch log optimize:
image

teamwork with @a6802739


123malin (approved):
LGTM

PR #31823, opened 2021-03-23
Change trainable to stop_gradient in optimizer

chenwhql:
### PR types

Function optimization
### PR changes

Others
### Describe

Change trainable to stop_gradient in optimizer.

Now users need to optimize a `input Tensor` but not `Parameter`, case like:
```
# 输入
img = paddle.to_tensor(img, dtype='float32', place=paddle.CUDAPlace(0), stop_gradient=False)

#设置为不保存梯度值 自然也无法修改
for param in model.parameters():
    param.stop_gradient = True
    
# 设置优化器
optimizer = paddle.optimizer.Adam(learning_rate=0.001, parameters=img)
```
But our optimizer now only support to optimize `Parameter`. Because now our `ParamBase.trainable` is equal to `not ParamBase.stop_gradient`, so we can use stop_gradient directly in optimzier to support optimizer `Tensor`. 

the source code are:

```python
    @property
    def trainable(self):
        return not self.stop_gradient

    @trainable.setter
    def trainable(self, trainable):
        if isinstance(trainable, bool):
            self.stop_gradient = not trainable
        else:
            raise ValueError(
                "The type of trainable MUST be bool, but the type is ",
                type(trainable))
```

### DIFF
--- a/python/paddle/optimizer/adam.py
+++ b/python/paddle/optimizer/adam.py
@@ -351,7 +351,7 @@ class Adam(Optimizer):
         """
         params_grads = []
         for param in self._parameter_list:
-            if not param.trainable:
+            if param.stop_gradient:
                 continue
             if param._grad_ivar() is not None:
                 grad_var = param._grad_ivar()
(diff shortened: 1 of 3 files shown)


JiabinYang (approved):
LGTM

phlrain (approved):

PR #31834, opened 2021-03-24
[NPU] fix bug of lookup_table_v2_grad

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

[NPU] fix bug of lookup_table_v2_grad

The original implementation has two problems
- The input "W" is set NoNeedBuffer so we can not use its content,
- The scatter should be done at a zero tensor  

And, this PR also adds synchronization in fetch op to avoid data error.

image
image


pangyoki (approved):
LGTM

liym27 (approved):
LGTM

PR #31836, opened 2021-03-24
[ROCM] added a cudnn switch of conv2d for rocm platform

ronny1996:
### PR types

Others
### PR changes

Others
### Describe

added a cudnn switch of conv2d for rocm platform

### DIFF
--- a/python/paddle/fluid/layers/nn.py
+++ b/python/paddle/fluid/layers/nn.py
@@ -1603,6 +1603,10 @@ def conv2d(input,
 
     pre_bias = helper.create_variable_for_type_inference(dtype)
 
+    if (core.is_compiled_with_cuda() and paddle.fluid.get_flags(
+            "FLAGS_conv2d_disable_cudnn")["FLAGS_conv2d_disable_cudnn"]):
+        use_cudnn = False
+
     helper.append_op(
         type=l_type,
         inputs={
--- a/python/paddle/nn/layer/conv.py
+++ b/python/paddle/nn/layer/conv.py
@@ -25,6 +25,7 @@ __all__ = [
 
 import numpy as np
 
+from ...fluid import get_flags
 from ...fluid import core
 from ...device import get_cudnn_version
 from ...fluid.dygraph import layers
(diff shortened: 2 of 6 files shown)


python/paddle/nn/layer/conv.py
```diff
@@ -629,6 +633,11 @@ def __init__(self,
                  weight_attr=None,
                  bias_attr=None,
                  data_format="NCHW"):
+        use_cudnn = True
```
qili93: 这里写成类似这样的形式
use_cudnn  = True if (core.is_compiled_with_cuda() and
                                   cudnn_version is not None) else False

python/paddle/nn/layer/conv.py
```diff
@@ -153,6 +156,7 @@ def _get_default_param_initializer():
                                           out_channels % in_channels == 0):
             self._op_type = 'depthwise_conv2d'
             self._use_cudnn = False
+        self._use_cudnn = self._use_cudnn and use_cudnn
```
qili93: 加一行注释，解释一下这里这么修改的原因，例如
// add environment of ROCM_CONV2D_DISABLE_CUDNN to disable use_cudnn in Conv2D
// If use_cudnn is not defined, default value is True, then self._use_cudnn = self._use_cudnn

qili93 (dismissed):
LGTM, api test will be covered by CPU CI and CUDA CI

python/paddle/fluid/layers/nn.py
```diff
@@ -1498,6 +1498,9 @@ def conv2d(input,
           conv2d = paddle.static.nn.conv2d(input=data, num_filters=2, filter_size=3, act="relu")
           print(conv2d.shape) # [-1, 2, 30, 30]
     """
+    use_cudnn = False if (
+        core.is_compiled_with_rocm() and
+        bool(os.getenv("ROCM_CONV2D_DISABLE_CUDNN"))) else use_cudnn
```
zhiqiu: If ROCM_CONV2D_DISABLE_CUDNN is not a built-in variable in ROCM but created by Paddle, I suggest following the naming rule like other environments like "FLAGS_xxx_xxx".

zhiqiu: Plus, it's better to get the value by `fluid.get_flags` instead of `os.getenv`

ronny1996: update

qili93 (dismissed):
LGTM

zhiqiu (approved):
LGTM for paddle.fluid.layers.conv2d

PR #31864, opened 2021-03-25
【NPU】Add int dtype kernel for reshape2 op

frankwhzhang:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Add int dtype kernel for reshape2 op


liym27 (approved):
LGTM

PR #31875, opened 2021-03-26
Fix bug of set_value op when assign a tensor to a tensor

liym27:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Fix bug of set_value op when assign a tensor to a tensor.

For example:
```python
x = paddle.ones(shape=[3, 4], dtype="int32")
x[:, 1] = x[:,1] * 2

print(x)
```

Before this PR, 
```
Tensor(shape=[3, 4], dtype=int32, place=CPUPlace, stop_gradient=True,
       [[1, 2, 1, 1],
        [1, 4, 1, 1],
        [1, 4, 1, 1]])
```

After this PR, 
```
Tensor(shape=[3, 4], dtype=int32, place=CPUPlace, stop_gradient=True,
       [[1, 2, 1, 1],
        [1, 2, 1, 1],
        [1, 2, 1, 1]])
```

### DIFF
--- a/paddle/fluid/operators/set_value_op.h
+++ b/paddle/fluid/operators/set_value_op.h
@@ -212,13 +212,16 @@ class SetValueKernel : public framework::OpKernel<T> {
     // set_value is what we want.
     TensorCopy(*in, place, out);
 
-    Tensor slice_t(dtype), pad_t(dtype);
+    Tensor slice_t(dtype), pad_t(dtype), zeros_t(dtype);
     slice_t.mutable_data<T>(slice_dims, place);
     pad_t.mutable_data<T>(in_dims, place);
+    zeros_t.mutable_data<T>(slice_dims, place);
 
     auto pad_e = framework::EigenTensor<T, D>::From(pad_t, in_dims);
     auto out_e = framework::EigenTensor<T, D>::From(*out);
     auto slice_e = framework::EigenTensor<T, D>::From(slice_t, slice_dims);
+    auto zeros_e = framework::EigenTensor<T, D>::From(zeros_t, slice_dims);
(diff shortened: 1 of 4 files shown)


zhhsplendid (dismissed):
LGTM

paddle/fluid/operators/set_value_op.h
```diff
@@ -248,7 +251,7 @@ class SetValueKernel : public framework::OpKernel<T> {
     if (value_tensor != nullptr) {
       // ElementwiseComputeEx can do broadcasting
       ElementwiseComputeEx<SubFunctor<T>, DeviceContext, T>(
-          ctx, &slice_t, value_tensor, -1, SubFunctor<T>(), &slice_t);
+          ctx, &zeros_t, value_tensor, -1, SubFunctor<T>(), &slice_t);
```
zhiqiu: The shape of slice_t and value_tensor may not equal, which result in the error, please check it.

zhiqiu (changes requested):

paddle/fluid/operators/set_value_op.h
liym27: Yes, but `ElementwiseComputeEx` support broadcast and the check will be done when call `ElementwiseComputeEx`.

zhiqiu: If slice_t.shape=[3,1] and zeros.shape=[3,1] and value.shape=[3], then the out tensor of ElementwiseComputeEx should be of shape [3,3]. If slice_t not re-allocate memory, it may result in error.

liym27: Thanks, done

zhiqiu (approved):
LGTM

PR #31935, opened 2021-03-29
Add BF16 Constant Initializer and support for other initializer 

wozna:
### PR types

Others 
### PR changes

Others 
### Describe

This PR: 
- adds BF16 Constant Initializer
- adds support when initializer input is BF16 or `dtype` is set to BF16. Then all initializers except the `Constant Initializer` will initialize the data in FP32 and then ` cast` operator FP32 -> BF16 will be added, so the output from this operator will be BF16.

### DIFF
--- a/paddle/fluid/operators/math/math_function.cc
+++ b/paddle/fluid/operators/math/math_function.cc
@@ -27,6 +27,7 @@ limitations under the License. */
 #include <vector>
 #include "paddle/fluid/framework/data_type.h"
 #include "paddle/fluid/operators/math/math_function_impl.h"
+#include "paddle/fluid/platform/bfloat16.h"
 #include "paddle/fluid/platform/float16.h"
 #include "unsupported/Eigen/CXX11/Tensor"
 
@@ -49,7 +50,6 @@ template struct SetConstant<platform::CPUDeviceContext, platform::complex128>;
 
 #ifdef PADDLE_WITH_XPU
 template struct SetConstant<platform::XPUDeviceContext, platform::float16>;
-template struct SetConstant<platform::XPUDeviceContext, platform::bfloat16>;
 template struct SetConstant<platform::XPUDeviceContext, float>;
 template struct SetConstant<platform::XPUDeviceContext, double>;
 template struct SetConstant<platform::XPUDeviceContext, uint8_t>;
--- a/paddle/fluid/operators/math/math_function.cc
+++ b/paddle/fluid/operators/math/math_function.cc
@@ -49,6 +49,7 @@ template struct SetConstant<platform::CPUDeviceContext, platform::complex128>;
 
 #ifdef PADDLE_WITH_XPU
 template struct SetConstant<platform::XPUDeviceContext, platform::float16>;
+template struct SetConstant<platform::XPUDeviceContext, platform::bfloat16>;
 template struct SetConstant<platform::XPUDeviceContext, float>;
 template struct SetConstant<platform::XPUDeviceContext, double>;
 template struct SetConstant<platform::XPUDeviceContext, uint8_t>;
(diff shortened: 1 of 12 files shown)


wozna:
@arlesniak  @arogowie-intel  Could you please start your review?

jczaja (dismissed):
LGTM

paddle/fluid/operators/math/math_function.cc
```diff
@@ -49,6 +49,7 @@ template struct SetConstant<platform::CPUDeviceContext, platform::complex128>;
 
 #ifdef PADDLE_WITH_XPU
 template struct SetConstant<platform::XPUDeviceContext, platform::float16>;
+template struct SetConstant<platform::XPUDeviceContext, platform::bfloat16>;
```
arogowie-intel, line 53: You will need to include header for bfloat16:
```
#include "paddle/fluid/platform/bfloat16.h"
```

python/paddle/fluid/tests/unittests/test_fill_constant_op.py
```diff
@@ -431,5 +431,36 @@ def test_shape_tensor_list_dtype():
...
+        self.dtype = np.uint16
+        self.mkldnn_data_type = "bfloat16"
+
+    def test_check_output(self):
+        self.check_output_with_place(core.CPUPlace())
+
+    def test_error(self):
+        if core.is_compiled_with_cuda():
+            with self.assertRaises(Exception):
+                self.check_output_with_place(core.CUDAPlace(0))
```
arogowie-intel: Why are you testing the throwing exception and not using 'unittest.skipif' as in our other bf16 tests?

wozna: I added it to check PADDLE_THROW in `fill_constant_op.h` file, but I removed it in commit https://github.com/PaddlePaddle/Paddle/pull/31935/commits/5cf7140fcd299967e06079440a382579cf5d4e5f. 
I will correct it. .

wozna:
@jczaja @arogowie-intel  Could you please repeat your review?

paddle/fluid/operators/math/math_function.cu
```diff
@@ -33,6 +34,7 @@ using complex64 = paddle::platform::complex64;
 using complex128 = paddle::platform::complex128;
 
 template struct SetConstant<platform::CUDADeviceContext, platform::float16>;
+template struct SetConstant<platform::CUDADeviceContext, platform::bfloat16>;
```
arlesniak, line 37: Are that changes needed here?  BF16 is enabled only for CPU.

wozna: Unfortunately, because of fill_constant_op.h implementation it is needed to add this change.
But thanks to https://github.com/PaddlePaddle/Paddle/blob/e2d4ef9d4d8cfb278da3e3860fd503ed639a540c/paddle/fluid/operators/fill_constant_op.h#L109  only the CPU version will be enabled.

paddle/fluid/operators/math/math_function.cc
arogowie-intel: That's strange but I do not see added header? Did you add it?

python/paddle/fluid/tests/unittests/test_fill_constant_op.py
arogowie-intel: Unfortunately I do not see changes

arogowie-intel (approved):

paddle/fluid/operators/math/math_function.cu
wozna: To clarify, the point is that in `fill constant op` we can choose the place of calculations (DeviceContext) also using the "place_type" attribute, then in `fill_constant_op.h` we call `math::SetConstant` function where we set DeviceContext and date type. When I registered bfloat16 kernel, during compilation it is checked if math::SetConstant call is supported for all data types.  And I didn't find a better solution not to have to refactor the whole thing.

arlesniak: LGTM

arlesniak (approved):
LGTM

wozna:
In PR-CI-CPU-Py2 approval is needed due to changes in `core.ops` where BF16 support has been added.

luotao1 (approved):

zhiqiu (approved):
LGTM for paddle.fluid.layers.embedding

PR #31951, opened 2021-03-30
Add LayerDict class

MingMingShangTian:
### PR types

New features
### PR changes

APIs
### Describe

Add LayerDict to store the layers with its name.

image

image
image

### DIFF
--- a/python/paddle/nn/layer/container.py
+++ b/python/paddle/nn/layer/container.py
@@ -19,6 +19,50 @@ __all__ = ['LayerDict', ]
 
 
 class LayerDict(Layer):
+    """
+    LayerDict holds sublayers in the ordered dictionary, and sublayers it contains are properly registered.
+    Holded sublayers can be accessed like a regular ordered python dictionary. 
+
+    Parameters:
+        sublayers (iterable of Layer, optional): sublayers to hold
+
+    Examplex:
+        .. code-block:: python
+
+            import paddle
+            import numpy as np
+            from collections import OrderedDict
+
+            sublayers = OrderedDict([
+                ('conv1d', paddle.nn.Conv1D(3, 2, 3)),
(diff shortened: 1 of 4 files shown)


python/paddle/nn/layer/container.py
```diff
@@ -0,0 +1,278 @@
...
+                
+                for k, v in layer_dict.items():
+                    print(k, ":", v)
+                #conv1d : Conv1D(3, 2, kernel_size=[3], data_format=NCL)
+                #conv2d : Conv2D(4, 2, kernel_size=[4, 4], data_format=NCHW)
+                #conv3d : Conv3D(4, 6, kernel_size=[3, 3, 3], data_format=NCDHW)
+                #relu : ReLU()
+
+        """
+        for key, layer in sublayers.items():
```
chenwhql: Do we need add checking for inputs?

python/paddle/nn/layer/container.py
```diff
@@ -0,0 +1,278 @@
...
+            layers_dict.clear()
+            len(layers_dict)
+            #0
+
+    """
+
+    def __init__(self, sublayers=None):
+        super(LayerDict, self).__init__()
+        if sublayers is not None:
+            self._sub_layers.update(sublayers)
```
chenwhql: here need call `self.update`, and the input should be checked in `update` method, otherwise here also need to check input

MingMingShangTian: Done

python/paddle/nn/layer/container.py
MingMingShangTian: add the inputs checking

chenwhql (approved):
LGTM

ForFishes (approved):
LGTM

PR #31956, opened 2021-03-30
[NPU] enable async copy and  add wait before sync operation

zhiqiu:
### PR types

Performance optimization
### PR changes

Others
### Describe

- enable async copy and  add wait before sync operation
- add new API async "FillNpuTensorWithConstant" to replace  sync "TensorFromVector"

TODO: replace all TensorFromVector used in NPU op

image
image
image


pangyoki (approved):
LGTM

PR #31976, opened 2021-03-31
[CustomOp]Support MacOS platform and Remove libpaddle_custom_op.so dependency

Aurelius84:
### PR types

New features
### PR changes

Others
### What's New?

**Custom Op supports to compile on MacOS platform**

#### 1. Remove libpaddle_custom_op.so

After this PR, we no long link `libpaddle_custom_op.so` on Linux and `libpaddle_custom_op.dylib` on Mac platform. Of course, we don't need to `cc_library` them  and inlcude into bdist whl. The main benefit is to decrease the size of Paddle whl package.

#### 2. Only Need core_(no)avx.so

We finally decouple the third DSO lib and only rely on `core_(no)avx.so`. According our experiment, the cost time is almost same as before. 

zh doc: https://github.com/PaddlePaddle/FluidDoc/pull/3413


python/paddle/utils/cpp_extension/cpp_extension.py
```diff
@@ -367,11 +368,14 @@ def finalize_options(self):
             self.build_lib = self.output_dir
 
     def build_extensions(self):
+        if OS_NAME.startswith("darwin"):
```
chenwhql, line 371: why only need to check mac compiler

python/paddle/utils/cpp_extension/extension_utils.py
```diff
@@ -406,6 +474,30 @@ def normalize_extension_kwargs(kwargs, use_cuda=False):
...
+    assert OS_NAME.startswith('darwin')
+
+    raw_core_name = _get_core_name()
+    core_path = os.path.join(_get_fluid_path(), raw_core_name)
+    new_lib_core_path = _get_lib_core_path()
+
+    # create symbol link
+    if not os.path.exists(new_lib_core_path):
+        try:
+            os.symlink(core_path, new_lib_core_path)
```
chenwhql, line 490: Will it be impossible to create symlnik here due to permission issues?

chenwhql (approved):
LGTM

python/paddle/utils/cpp_extension/cpp_extension.py
Aurelius84: Because `self.compiler` will use `gcc` on some mac devices.

python/paddle/utils/cpp_extension/extension_utils.py
Aurelius84: If users can install paddle, there will be  no permission issues.

PR #31989, opened 2021-03-31
upgrade the TraceLayer.save_inference_model method with add file suffix automatically

CtfGo:
### PR types

Bug fixes

### PR changes

APIs

### Describe

update the TraceLayer.save_inference_model method:
(1) change the param "dirname" to "path" to indicate it is a path prefix of the saved model, and update docs
(2) save the program and params on the path with ".pdmodel" and ".pdiparams" suffix seperately

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_traced_layer_err_msg.py
+++ b/python/paddle/fluid/tests/unittests/test_traced_layer_err_msg.py
@@ -115,36 +115,36 @@ class TestTracedLayerErrMsg(unittest.TestCase):
             dygraph_out, traced_layer = fluid.dygraph.TracedLayer.trace(
                 self.layer, [in_x])
 
-            dirname = './traced_layer_err_msg'
+            path = './traced_layer_err_msg'
             with self.assertRaises(TypeError) as e:
                 traced_layer.save_inference_model([0])
             self.assertEqual(
-                "The type of 'dirname' in fluid.dygraph.jit.TracedLayer.save_inference_model must be <{} 'str'>, but received <{} 'list'>. ".
+                "The type of 'path' in fluid.dygraph.jit.TracedLayer.save_inference_model must be <{} 'str'>, but received <{} 'list'>. ".
                 format(self.type_str, self.type_str), str(e.exception))
             with self.assertRaises(TypeError) as e:
-                traced_layer.save_inference_model(dirname, [0], [None])
+                traced_layer.save_inference_model(path, [0], [None])
             self.assertEqual(
                 "The type of 'each element of fetch' in fluid.dygraph.jit.TracedLayer.save_inference_model must be <{} 'int'>, but received <{} 'NoneType'>. ".
                 format(self.type_str, self.type_str), str(e.exception))
             with self.assertRaises(TypeError) as e:
-                traced_layer.save_inference_model(dirname, [0], False)
+                traced_layer.save_inference_model(path, [0], False)
             self.assertEqual(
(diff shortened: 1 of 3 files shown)


chenwhql (approved):
LGTM

这个更改了一个参数名，dirname -> path，可能会造成2.1和2.0在使用了这个API的部分场景中不兼容，但我个人支持做不兼容修改，有几方面原因：
1. 目前paddle存储多个结果的save load接口全局都统一成存储后缀的形式了，这个接口属于漏网之鱼，有必要统一
2. dirname这个名字表意明确，与前缀的含义不兼容，更改为path更加通用，长远看比较好
3. 2.1和2.0属于大版本不同，少量不兼容问题也合理
4. 这个接口是低频接口，使用场景不多，而且参数位置也没有改变，只要用户没有明确使用参数名，就可以继续使用

仅代表个人看法，具体还需要 @lanxianghit @jzhang533 @XiaoguangHu01 看一下

saxon-zh (approved):
LGTM

python/paddle/fluid/tests/unittests/test_traced_layer_err_msg.py
```diff
@@ -174,5 +180,25 @@ def test_linear_net_with_none(self):
                                                                         [in_x])
 
 
+class TestTracedLayerSaveInferenceModel(unittest.TestCase):
+    """test save_inference_model will automaticlly create non-exist dir"""
+
+    def setUp(self):
+        self.save_path = "./nonexist_dir/fc"
+        import shutil
```
zhhsplendid, line 188: NIT (Not Important): However, since it is not so important and this PR needs many approves. I will approve it and you can change in next PR.

Python officially suggests to put imports at top of file, I think you should follow it in your case. 
https://www.python.org/dev/peps/pep-0008/#imports

image

Let me explain my understanding of the pros and cons of importing at top of a file or in a function.

Importing at top of a file:
1. More consistent with other programming language.
2. We group imports all together and easy to organize

There are several cases I think we can import in a function:
1. Conditional import, e.g.
```
if something:
    import xxx
```
2. You would like your function code to work correctly after copy the function code to other file (other file may not contain that import)

But I don't think here we meet the cases for importing in a function :-) so I would suggest to import at the top.

python/paddle/fluid/tests/unittests/test_traced_layer_err_msg.py
```diff
@@ -174,5 +180,25 @@ def test_linear_net_with_none(self):
...
+        if os.path.exists(os.path.dirname(self.save_path)):
+            shutil.rmtree(os.path.dirname(self.save_path))
+
+    def test_mkdir_when_input_path_non_exist(self):
+        fc_layer = SimpleFCLayer(3, 4, 2)
+        input_var = paddle.to_tensor(np.random.random([4, 3]).astype('float32'))
+        with fluid.dygraph.guard():
+            dygraph_out, traced_layer = fluid.dygraph.TracedLayer.trace(
```
zhhsplendid, line 196: We should write new test cases for Paddle 2.0 path. So `fluid.dygraph.TracedLayer` should be changed to `paddle.jit.TracedLayer`. However, since it is not so important and this PR needs many approves. I will approve it and you can change in next PR.

zhhsplendid (approved):

python/paddle/fluid/tests/unittests/test_traced_layer_err_msg.py
CtfGo: Thank you for your suggestion，I will change these two in next pr.

lanxianghit (approved):

XiaoguangHu01 (approved):
LGTM

PR #32006, opened 2021-04-01
fix dynamic graph to static graph bug of expand op when shape = -1.

HexToString:
### PR types

Bug fixes
### PR changes

OPs
### Describe

gx原本shape为(-1, 1, -1)，
t=paddle.expand(gx, [-1, -1, -1])，此时t的shape为(-1, 1, -1)，不符预期，应为(-1, -1, -1)。
因为运行期间，如果expand中shape为（2，480，480），那么输出的shape应为（2，480，480），可实际的shape为（2，1，480）。
shape=-1时不能直接拷贝原shape，应在运行期间决定shape。
与李龙修复的shape中包含变量，应在运行期决定shape，而不是编译期。逻辑相同。

### DIFF
--- a/python/paddle/tensor/manipulation.py
+++ b/python/paddle/tensor/manipulation.py
@@ -1447,7 +1447,7 @@ def expand(x, shape, name=None):
     def get_attr_expand_shape(list_expand_shape):
         attrs_expand_shape = []
         for idx, shape in enumerate(list_expand_shape):
-            if isinstance(shape, Variable):
+            if isinstance(shape, Variable) or shape == -1:
                 attrs_expand_shape.append(-2)
             else:
                 attrs_expand_shape.append(shape)


python/paddle/tensor/manipulation.py
```diff
@@ -1447,7 +1447,7 @@ def expand(x, shape, name=None):
     def get_attr_expand_shape(list_expand_shape):
         attrs_expand_shape = []
         for idx, shape in enumerate(list_expand_shape):
-            if isinstance(shape, Variable):
+            if isinstance(shape, Variable) or shape == -1:
```
qingqing01, line 1450: Add comments for what meaning of -2 ?

zhhsplendid, line 1450: Should we also consider `shape == None` ? Because we also used `None` to represent unknown shape.

HexToString: if shape == None means dynamic shape, maybe we should add that.
i will confirm that.

HexToString: this is add by Lilong.
-2 is only a sign which means dynamic shape.

HexToString: i will add some comment.

HexToString:
问题原因不是上述分析导致的，该问题已由李龙同学修复，故关闭该pr

PR #32064, opened 2021-04-04
[PaddleTRT] Yolov3 bugfix

zlsh80826:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix two problem:
1. teller should not return true before the tell function end
2. Fix double cuda free problem in yolo_box plugin


cryoco (approved):
LGTM

PR #32081, opened 2021-04-06
remove pass restrictions for skip-ln pass

cryoco:
### PR types

Function optimization
### PR changes

Others
### Describe

skip-ln pass does not need emb-ln pass and multihead-matmul pass both


shangzhizhou (approved):
LGTM

PR #32084, opened 2021-04-06
fix fc doc

Joejiong:
### PR types

Others
### PR changes

Others
### Describe

Fix doc fc with unmatch weight_attr
if x is a list of two tensors, the weight_attr if provided, should also be a list of same lenght.


swtkiwi (approved):
LGTM

PR #32102, opened 2021-04-06
The unsupported_fp16_list using in AMP will be created automatically during the runtime.

wzzju:
### PR types

Function optimization

### PR changes

Others
### Describe

After this pull request, the `unsupported_fp16_list` using in AMP will be created automatically during the runtime.


paddle/fluid/pybind/pybind.cc
```diff
@@ -181,6 +184,53 @@ bool SupportsBfloat16FastPerformance() {
 #endif
 }
 
+std::tuple<std::unordered_set<std::string>, std::unordered_set<std::string>,
```
zhhsplendid, line 197: Something for the readability, please add simple comment.

The function checks all ops supports input place and dtype, returns all_ops, supported_ops, and unsupported_ops. The function name and return type are not obvious enough for above information.

paddle/fluid/pybind/pybind.cc
```diff
@@ -181,6 +184,53 @@ bool SupportsBfloat16FastPerformance() {
...
+           std::unordered_set<std::string>>
+OpSupportedInfos(std::string place, framework::proto::VarType::Type dtype) {
```
zhhsplendid: Would you mind input a `const std::string& place` ?

This is just following C++ style:
https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

If the `std::toupper` causes the trouble for using const reference, you can do a copy inside the function.

paddle/fluid/pybind/pybind.cc
wzzju: Done.

paddle/fluid/pybind/pybind.cc
wzzju: Done.

zhhsplendid (dismissed):
LGTM

zhhsplendid (dismissed):
LGTM

zhhsplendid (approved):
LGTM

Issue #31992, opened 2021-03-31
Deploying multiple models at the same time will raise MKLDNN error


juncaipeng:
* Info

C++ API
Ubuntu 16.04
CPU MKLDNN
GCC 8.2.0

* Prepare

Please contact danqing to download the demo. The demo only has two group models for test.

Download [model_test.cc.zip](https://github.com/PaddlePaddle/Paddle/files/6248677/model_test.cc.zip), unzip model_test.cc.zip, use the new model_test.cc to update the old model_test.cc file in the demo.

Make and install paddle release2.0 (commit id: `c7a6a1f9610a9ee018c19d89950d76b38f33aed1`).
```
cmake -DCMAKE_BUILD_TYPE=Release -DWITH_PYTHON=OFF -DWITH_MKL=ON -DWITH_GPU=OFF -DON_INFER=ON .. && make -j && make inference_lib_dist -j
```

Set LIB_DIR as the path of PaddleInference in build.sh.

Run `sh build.sh`.

Run `ulimit -c unlimited` enable save core file.

* The first test 

Run `./build/model_test --test_groups=0 --single_instance=true` , it does not raise error.
If set `single_instance` as true, every model only has one predictor. Otherwise, some models will have several predictors by calling predictor.clone().

Run `./build/model_test --test_groups=0 --single_instance=false`, it raises segmentation fault error.  
Run `gdb ./build/model_test core_file` get the following error.

![image](https://user-images.githubusercontent.com/52520497/113134005-e4e33080-9252-11eb-9f79-71e4e31e2253.png)

* The second test 

Run `./build/model_test --test_groups=1 --single_instance=true` , it does not raise error.

Run `./build/model_test --test_groups=1 --single_instance=false`, it also raises segmentation fault error.  

![image](https://user-images.githubusercontent.com/52520497/113134330-3c819c00-9253-11eb-9a34-2128656f5bbd.png)

Sometimes, the above demo raises different error, such as

![image](https://user-images.githubusercontent.com/52520497/113137943-c469a500-9257-11eb-93a1-5f5e911a39d4.png)

* The third test

Run `./build/model_test --test_groups="4 5 6" --single_instance=true`. The demo loads several group models and every model has one predictor, and it also raises error as following. The demo only has  two group models for now, we will provide other models later.

![image](https://user-images.githubusercontent.com/52520497/113283388-d06a6b00-931a-11eb-8ed4-6ec80a1f4147.png)

[...]

Issue #32087, opened 2021-04-06
Improve Multi-threading, multi-instance UT


lidanqing-vv:
There are not enough UTs for many threading tests and many instance cases.

@juncaipeng Please help figure out all use cases for multi-threading and multi-instances, about how users will start these scenarios, and through which functions. 

After figuring out all scenarios, we could discuss and prepare UTs together

lidanqing-vv:
Hi @juncaipeng After finishing current 2 issues, we may work on this together. Jacek and Adam could join discussion too.

lidanqing-vv:
```
void TestMultiThreadPrediction(
    const PaddlePredictor::Config *config,
    const std::vector<std::vector<PaddleTensor>> &inputs,
    std::vector<std::vector<PaddleTensor>> *outputs, int num_threads,
    bool use_analysis = true) {
  std::vector<std::thread> threads;
  std::vector<std::unique_ptr<PaddlePredictor>> predictors;
  predictors.emplace_back(CreateTestPredictor(config, use_analysis));
  for (int tid = 1; tid < num_threads; tid++) {
    predictors.emplace_back(predictors.front()->Clone());
  }

  for (int tid = 0; tid < num_threads; ++tid) {
    threads.emplace_back([&, tid]() {
      // Each thread should have local inputs and outputs.
      // The inputs of each thread are all the same.
      std::vector<std::vector<PaddleTensor>> outputs_tid;
      auto &predictor = predictors[tid];
      if (FLAGS_warmup) {
        PredictionWarmUp(predictor.get(), inputs, &outputs_tid, num_threads,
                         tid);
      }
      PredictionRun(predictor.get(), inputs, &outputs_tid, num_threads, tid);
    });
  }
  for (int i = 0; i < num_threads; ++i) {
    threads[i].join();
  }
}
```


```
  std::vector<std::thread> run_threads;
  std::vector<std::shared_ptr<Predictor>> predictor_list;
  std::vector<Params> tmp_model_params;

  for (int i = 0; i < test_model_params.size(); i++) {
    auto& param = test_model_params[i];

    auto predictor = CreatePredictor(param);
    
    predictor_list.push_back(predictor);
    tmp_model_params.push_back(param);

    for (int j = 1; j < param.instance && !FLAGS_single_instance; j++) {
      predictor_list.emplace_back(std::move(predictor->Clone()));
      tmp_model_params.push_back(param);
    }
  }

[...]

PR #32136, opened 2021-04-07
[oneDNN] Candidate fix to #31992

jczaja:
### PR types
Bug fixes

### PR changes
Others

### Describe
This is a fix to #31992 . It extends for mulit-threaded & multi-executor scenario caching keys with info on Threads and Executors. There are also some fixes related to data races in MKLDNNDeviceContext

### DIFF
--- a/paddle/fluid/platform/device_context.h
+++ b/paddle/fluid/platform/device_context.h
@@ -600,6 +600,8 @@ class MKLDNNDeviceContextThreadLocals {
     // MKL-DNN stream used for execution of primitives (per-thread)
     mkldnn::engine cur_engine;
     mkldnn::stream cur_stream;
+    std::string key_suffix;  // Key identifying current Executor
+    bool key_attach_thread_id = true;
 
     Body();
     ~Body();
@@ -612,6 +614,10 @@ class MKLDNNDeviceContextThreadLocals {
     void log_lib_version(void);
     const mkldnn::engine& get_engine(void);
     mkldnn::stream& get_stream(void);
+    void set_key_suffix(const std::string& suffix) { key_suffix = suffix; }
+    const std::string& get_key_suffix(void) const { return key_suffix; }
+    void disable_tid_in_key(void) { key_attach_thread_id = false; }
+    bool is_tid_used_in_key(void) const { return key_attach_thread_id; }
   };
   MKLDNNDeviceContextThreadLocals() = default;
   MKLDNNDeviceContextThreadLocals(const MKLDNNDeviceContextThreadLocals& c) =
@@ -655,14 +661,6 @@ class MKLDNNDeviceContext : public CPUDeviceContext {
   // Remove all entries from the blob map
   void ResetBlobMap();
 
-  // Set a suffix to be added to key
-  void SetKeySuffix(const std::string& suffix) { key_suffix_ = suffix; }
-  const std::string& GetKeySuffix(void) const { return key_suffix_; }
-
-  // Disable adding  thread ID to the key
-  void DisableThreadInfoInKey(void) { key_attach_thread_id_ = false; }
-  bool IsThreadIdUsedInKey(void) const { return key_attach_thread_id_; }
-
   // Prevent next ResetBlobMap()
   void BlockNextCacheClearing();
 
@@ -686,8 +684,6 @@ class MKLDNNDeviceContext : public CPUDeviceContext {
   std::shared_ptr<BlobMap> p_blobmap_;
   std::shared_ptr<std::mutex> p_mutex_;
   bool block_next_cache_clearing_ = false;
-  std::string key_suffix_;  // Key identifying current Executor
-  bool key_attach_thread_id_ = true;
 };
 #endif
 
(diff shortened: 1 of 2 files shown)


jczaja:
@arogowie-intel Please review

arogowie-intel (commented):
In general I have no objections. Could you please prepare some UT to check for correctness of this behavior?

jczaja:
@arogowie-intel That is good point. There is request by @lidanqing-intel  to get more tests on that : https://github.com/PaddlePaddle/Paddle/issues/32087 so it seems there is some work going on that one. I think I would like  to see what they will add and eventually build on top of it. Our current tests (test_mkldnn_caching) are testing only caching in context of single ops, not executors and predictors, so cannot be easily extended. Anyway I will try to implement little UT on that one , but due to fact that this is fix to urgent issue #31992 I would like to do it in next PR

arogowie-intel (approved):

jczaja:
@luotao1 Could you please start you review?

paddle/fluid/platform/device_context.h
```diff
@@ -600,6 +600,8 @@ class MKLDNNDeviceContextThreadLocals {
     // MKL-DNN stream used for execution of primitives (per-thread)
     mkldnn::engine cur_engine;
     mkldnn::stream cur_stream;
+    std::string key_suffix;  // Key identifying current Executor
```
lidanqing-vv, line 603: Hi, is this `key_suffix` only related with executor identifier ? maybe rename it `executor_suffix` then you don't need the comment any more.

and if `key_attach_thread_id` is a boolean value, maybe rename it to `is_thread_id_attached`

paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -457,13 +466,14 @@ inline std::string CreateKey(const platform::MKLDNNDeviceContext& dev_ctx,
...
   expand_type{0, (AppendKey(&key, std::forward<ArgTypes>(args)), 0)...};
-  key += dev_ctx.GetKeySuffix();
+  key += paddle::platform::MKLDNNDeviceContext::tls().get_key_suffix();
   return key;
 }
 
 inline std::string ExtendKeyWithThreadInfoIfNeeded(
     const platform::MKLDNNDeviceContext& dev_ctx, const std::string& key) {
-  return ((dev_ctx.IsThreadIdUsedInKey() == true) &&
+  return ((paddle::platform::MKLDNNDeviceContext::tls().is_tid_used_in_key() ==
```
lidanqing-vv, line 475: Could here exist some comments?
I also have a question, following
```
platform::MKLDNNDeviceContext::tls().get_cur_mkldnn_session_id() ==
           platform::MKLDNNDeviceContextThreadLocals::kMKLDNNSessionID_Default
``` 
means it is not limited_cache mode (cache clearing mode)
together means
```
if [is_tid_used_in_key (which is default true and for first thread, false) and unlimited cache capacity mode]:
    thread_id will be added into the key
```
but if it is cache clearing mode, thread id will not be added into the key?  Will there be some risks? If many threads execute same executor and use limited cache, why thread_id not needed in the key ?

paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -439,14 +439,23 @@ inline void AppendKey(std::string* key, const std::vector<T>& dims) {
...
+    static auto first_thread = ThreadIDasStr();
+    static_vars_barrier.unlock();
+
+    if (first_exec != ptr) {
+      paddle::platform::MKLDNNDeviceContext::tls().set_key_suffix(
+          "E" + std::to_string(reinterpret_cast<uintptr_t>(ptr)));
+    }
+    // For first thread
+    if (first_thread == ThreadIDasStr()) {
+      paddle::platform::MKLDNNDeviceContext::tls().disable_tid_in_key();
```
lidanqing-vv, line 457: Please correct me if I am wrong.

Before, the use case is: Each thread will use one executor, so executor_ptr as key suffix is enough.

Now different thread may execute one executor, so executor ptr and thread is are both needed in the key. Only for first thread it can be exempted.

lidanqing-vv (commented):
Hi, it will be good if review could be considered

paddle/fluid/platform/device_context.h
jczaja: Currently only Executor info goes there but It may change e.g. some more info may go there as well if needed , so I prefer the name stay like this

paddle/fluid/platform/mkldnn_helper.h
jczaja: That is a valid concern. I remember that ClearCache mode was used only for single threaded execution, but when I look at code now , I do not see this limitation anywhere, so perhaps we should investgate this clearing cache functionality and if keys should be updated for it as well. I'm not going to look to that in this PR, so please file an issue in our internal backtracker.

paddle/fluid/platform/mkldnn_helper.h
jczaja: yes

paddle/fluid/platform/mkldnn_helper.h
lidanqing-vv: Ok I will make a jira task for this case.

lidanqing-vv:
LGTM

paddle/fluid/platform/device_context.h
lidanqing-vv: Ok.

paddle/fluid/platform/mkldnn_helper.h
lidanqing-vv: Ok. Thanks~

luotao1 (approved):

PR #32159, opened 2021-04-08
[Dy2Stat] Support DictCmp and zip grammer

Aurelius84:
### PR types

New features
### PR changes

Others
#### 1. Support DictCmp

A simple case:

```python
for i, val in iter_list:
    dict_val = {k: v for k, v in another_dict}
```

#### 2. Support iter `zip` in For

+ zip does not support `len()` and `zip(x,y)[index]` slice.
+ Solution: we create a new `gast.Assign` node to convert `zip` into `list`

A simple case:
```python
# original code:

for (x, y) in zip(list_a, list_b):
    XXXX

# transformed codes:
__for_loop_iter_zip_0 = list(zip(list_a, list_b))

for (x, y) in __for_loop_iter_zip_0:
    XXXX

```


liym27 (approved):
LGTM

PR #32171, opened 2021-04-09
Add inner register backward hook method for Tensor

chenwhql:
### PR types

New features
### PR changes

APIs
### Describe

Add inner register backward hook method for Tensor

- remove InplaceVariableWrapperHook class and related method
- add simple void Hook for reducing in distributed training
- add `Tensor._register_backward_hook` method in python
  - used for reducer like in c++ for distributed training
```
var->GradVarBase()->AddVoidHook(std::make_shared<std::function<void()>>(
        [=]() { this->AddDistHook(global_var_index); }));
```

- example
```python
HOOK_INIT_VALUE = 10

def global_void_hook():
    global HOOK_INIT_VALUE
    HOOK_INIT_VALUE *= 2

x = paddle.to_tensor(5., stop_gradient=False)
x._register_backward_hook(global_void_hook)

for i in range(5):
    y = paddle.pow(x, 4.0)
    y.backward()
```


ForFishes (approved):
LGTM

PR #32176, opened 2021-04-09
to avoid bug while value is a lambda expression; for binary ops dim a…

zhiboniu:
### PR types

Bug fixes
### PR changes

Others
### Describe

to avoid bug while value is a lambda expression; 
for binary ops dim and ndimension:
        ('dim', lambda x: len(x.shape)),
        ('ndimension', lambda x: len(x.shape)),

while api example check in CI test,Error occurs:
2021-04-01 20:44:25 ++ python sampcd_processor.py cpu
2021-04-01 20:44:26 Traceback (most recent call last):
2021-04-01 20:44:26   File "sampcd_processor.py", line 592, in <module>
2021-04-01 20:44:26     filenames = get_filenames()
2021-04-01 20:44:26   File "sampcd_processor.py", line 473, in get_filenames
2021-04-01 20:44:26     module = eval(api).__module__
2021-04-01 20:44:26   File "<string>", line 1
2021-04-01 20:44:26     paddle.tensor.logic.Tensor.\<lambda\>
2021-04-01 20:44:26                                ^
2021-04-01 20:44:26 SyntaxError: invalid syntax

### DIFF
--- a/tools/print_signatures.py
+++ b/tools/print_signatures.py
@@ -76,7 +76,6 @@ def format_spec(spec):
 def queue_dict(member, cur_name):
     if cur_name in omitted_list:
         return
-
     doc_md5 = md5(member.__doc__)
 
     if inspect.isclass(member):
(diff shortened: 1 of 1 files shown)


willthefrog (commented):
not sure this is the right way to proceed, also the syntax is already handled in [sampcd](https://github.com/PaddlePaddle/Paddle/blob/develop/tools/sampcd_processor.py#L561), therefore should not cause CI failure anymore ( you may need to rebase your PR as `sampcd` is updated **very** recently )

PR #32181, opened 2021-04-10
fix bug in Sequential getitem while input is str

zhiboniu:
### PR types

Bug fixes 
### PR changes

OPs
### Describe

fix bug in Sequential getitem while input is str
comparison operators between str and int cause error


willthefrog (commented):
sequential is supposed be equivalent to a python list, not a dict. therefore, it is not clear why should it support string indexing...

PR #32198, opened 2021-04-12
fix NPUDeviceContext in all c++ unittest

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

make sure that the same context is used in op kernel and test body


pangyoki (approved):
LGTM

PR #32224, opened 2021-04-13
update `get_api_md5`, using the real api name as the map's key

wadefelix:
### PR types
Bug fixes

### PR changes
Others

### Describe

function `get_api_md5` may use the alias name as the map's key. It causes some in-sub-module's alias name detected as the changed API.  so update the `get_api_md5`'s logic, use the API' real name as the key.

bug found in https://github.com/PaddlePaddle/Paddle/pull/31946


lelelelelez (approved):
LGTM

PR #32243, opened 2021-04-13
fix conditional block op

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix conditional block op


pangyoki (approved):
LGTM

PR #32343, opened 2021-04-19
Fix bug in amp O2

huangxu96:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix a bug in amp O2 mode. This bug will be triggered when the var is not in the block. Using the function of  _find_var_recursive to find var instead.


wzzju (approved):
LGTM.

PR #32375, opened 2021-04-20
Do not define and save reserve_space for inference.

Xreki:
### PR types

Bug fixes

### PR changes

APIs

### Describe

Do not define and save reserve_space for inference. If the inference_program is defined explicitly, this PR can avoid defining and saving reserve_space for inference. If the inference_program is cloned from train_program, this PR cannot make any difference. It need another method.


wzzju (approved):
LGTM.

qingqing01 (approved):

PR #32378, opened 2021-04-20
fix Windows CI MP compile and environment install script and openblas CI

zhwesky2010:
### PR types

Breaking changes
### PR changes

Others
### Describe

fix Windows CI MP compile and environment install script and openblas CI.


wanghuancoder (approved):
LGTM

PR #32414, opened 2021-04-21
Modify some contents for elementwise op impl

ZzSean:
### PR types

Others
### PR changes

OPs
### Describe

Modify some contents for elementwise op impl


Xreki (approved):
LGTM

PR #32428, opened 2021-04-21
let paddle.utils.install_check support CPU package with GPU device

pangyoki:
### PR types

Bug fixes

### PR changes

Others

### Describe

- logic problem in checking

In original `paddle.utils.install_check`, if users install the CPU package on the GPU device, PADDLE will check the installation using GPU logic.
This PR fixes this problem. If the user installs the CPU package, it will only check whether it can be performed well on the CPU. 

- add checking in dygraph mode

### DIFF
--- a/python/paddle/utils/install_check.py
+++ b/python/paddle/utils/install_check.py
@@ -82,6 +82,10 @@ def _run_dygraph_single(use_cuda):
         use_cuda (bool): Whether running with CUDA.
     """
     paddle.disable_static()
+    if use_cuda:
+        paddle.set_device('gpu')
+    else:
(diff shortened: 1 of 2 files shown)


python/paddle/utils/install_check.py
```diff
@@ -74,6 +74,30 @@ def _is_cuda_available():
         return False
 
 
+def _run_dygraph_single(use_cuda):
```
zhiqiu, line 77: use_cuda is not used.

pangyoki: done

zhiqiu (approved):
LGTM

PR #32468, opened 2021-04-22
Add BF16 uniform random initializer

wozna:
### PR types

New features
### PR changes

OPs 
### Describe

This PR adds support for the BF16 datatype for the uniform_random operator.
Therefore, in the initializers that I use uniform_random_op, i.e. Uniform, Xavier (uniform) and MSRA (uniform), it will not be necessary to add the cast operators before and after the initializer if the input is BF16, which was changed in the file `initalizer.py

### DIFF
--- a/paddle/fluid/operators/uniform_random_op.h
+++ b/paddle/fluid/operators/uniform_random_op.h
@@ -16,10 +16,8 @@
 #include <algorithm>
 #include <utility>
 #include <vector>
-#include "paddle/fluid/framework/generator.h"
 #include "paddle/fluid/framework/op_registry.h"
 #include "paddle/fluid/framework/operator.h"
-#include "paddle/fluid/platform/bfloat16.h"
 
 namespace paddle {
 namespace operators {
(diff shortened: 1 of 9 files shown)


jczaja (dismissed):
LGTM

paddle/fluid/operators/uniform_random_op.h
```diff
@@ -98,5 +100,33 @@ inline std::vector<int64_t> GetNewDataFromShapeTensorList(
...
+                                    const unsigned int& seed) {
+  VLOG(4) << "[CPU] UniformRandomKernel<bfloat16>";
+  std::uniform_real_distribution<float> dist(min, max);
+  auto engine = framework::GetCPURandomEngine(seed);
+
+  for (int64_t i = 0; i < size; ++i) {
+    data[i] = static_cast<platform::bfloat16>(dist(*engine));
+  }
+}
+
```
arogowie-intel: Those two function templates are an implementation detail, thus should not be placed in header file. Please move them into the source file and put into anonymous namespace or make them static at global namespace to prevent cluttering global/local namespaces.

arogowie-intel (commented):
Good job. Please just take into account my one comment.

paddle/fluid/operators/uniform_random_op.h
wozna: Yes, you are right it will be a better solution.

arogowie-intel (approved):

wozna:
@luotao1  Could you please start your review? 
There is an error in PR-CI-OP-benchmark: missing test script of "randint" and "uniform_random" for cuda file. 
Should I cover it even if changes are connected only to CPU?

luotao1 (approved):
LGTM

luotao1:

PR #32469, opened 2021-04-22
Fix bug that epsilon become 0 using power

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Fix bug that epsilon become 0 using power

We found that in ernie-3.0 model, when epsilon is 1e-8, `FillNpuTensorWithConstant`(which use power) returns 0 and may result in -nan on Parameters.


pangyoki (approved):
LGTM

liym27 (approved):
LGTM

PR #32470, opened 2021-04-22
[BUG FIX] when x.dim < y.dim, the result of compare_op is inverse 

wawltor:
### PR types
Bug fixes 

### PR changes
 OPs

### Describe
 BUG_FIX： when x.dim < y.dim, the result of compare_op is inverse

### DIFF
--- a/paddle/fluid/operators/controlflow/compare_op.h
+++ b/paddle/fluid/operators/controlflow/compare_op.h
@@ -68,7 +68,7 @@ struct NotEqualFunctor {
   }
 };
 
-template <typename DeviceContext, typename Functor>
+template <typename DeviceContext, typename Functor, typename InverseFunctor>
 class CompareOpKernel
     : public framework::OpKernel<typename Functor::ELEM_TYPE> {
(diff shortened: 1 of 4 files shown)


ZHUI (approved):
LGTM

paddle/fluid/operators/controlflow/compare_op.h
```diff
@@ -80,21 +80,33 @@ class CompareOpKernel
     auto* y = context.Input<Tensor>("Y");
     auto* z = context.Output<Tensor>("Out");
     int axis = context.Attr<int>("axis");
-    ElementwiseComputeEx<Functor, DeviceContext, T, bool>(context, x, y, axis,
-                                                          Functor(), z);
+
+    auto x_dims = x->dims();
+    auto y_dims = y->dims();
+    if (x_dims.size() >= y_dims.size()) {
```
ZHUI, line 86: It will be better approach to fix the problem in broadcast function. 

There should be some notes to report  the broadcast issues.

PR #32486, opened 2021-04-22
add c_concat and c_split ops

sandyhouse:
### PR types

New features
### PR changes

OPs
### Describe

add c_concat op, which allgather tensors and concats them along the last axis
add c_split op, which split a tensor according to its rank along the last axis


ForFishes (approved):
LGTM

XieYunshen (approved):

PR #32550, opened 2021-04-25
fix dataloader exit error

heavengate:
### PR types
 Bug fixes

### PR changes
APIs

### Describe
**fix dataloader exit error if user exit program when dataloader is still iterating**
- add `thread_done_event` check in `thead_loop` in single-process mode
- register `__del__` to `atexit` to ensure `__del__` is called when python program exit

example code:
```python
import paddle
import numpy as np
from paddle.io import DataLoader, BatchSampler

class DemoDataset(paddle.io.Dataset):
    def __init__(self):
        self.data = np.random.random((10, 20))

    def __getitem__(self, idx):
        return self.data[idx]

    def __len__(self):
        return len(self.data)

train_dataset = DemoDataset()

batch_sampler = BatchSampler(
    train_dataset, batch_size=1, shuffle=True, drop_last=True)

loader = DataLoader(
    train_dataset,
    batch_sampler=batch_sampler,
    places=paddle.CUDAPlace(0),
    # num_workers设置为大于0就没有问题
    num_workers=0,
    return_list=True,
)

for idx, data in enumerate(loader):
    # 数据未遍历完退出
    if idx == 2:
        break
    print(idx)
```

error:
da14fbb22dbe6f6328ee47617

### DIFF
--- a/python/paddle/fluid/dataloader/dataloader_iter.py
+++ b/python/paddle/fluid/dataloader/dataloader_iter.py
@@ -35,7 +35,7 @@ else:
 import paddle
 from .. import core, layers
 from ..framework import in_dygraph_mode
-from ..multiprocess_utils import _set_SIGCHLD_handler, MP_STATUS_CHECK_INTERVAL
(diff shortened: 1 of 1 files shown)


chenwhql (approved):
LGTM

PR #32556, opened 2021-04-26
fix CheckDims issue of elementwise

b3602sss:
### PR types
Bug fixes

### PR changes
OPs

### Describe
we can't get dimension if it's dynamic shape.


shangzhizhou (approved):
LGTM
临时修复，继续跟进trt elementwise layer的问题

PR #32579, opened 2021-04-26
[HybridParallel] Fix amp bug in ModelParallel

ForFishes:
### PR types

Bug fixes 
### PR changes

Others
### Describe

[HybridParallel] Fix amp bug in ModelParallel.
 Since the minimize call in the optimizer is after the gradscaler, check_finite needs to synchronize global information. In the future, we should use check_group


sandyhouse (approved):
LGTM

PR #32592, opened 2021-04-26
Reduce grad fix

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fixed reduce backward kernel by changing the source of mkldnn memory format tag. Now it is taken from forward primitive descriptor.


jczaja (dismissed):
LGTM

paddle/fluid/operators/reduce_ops/reduce_op.h
```diff
@@ -590,7 +581,6 @@ class ReduceGradOp : public framework::OperatorWithKernel {
     }
 #endif
 
-    int in_dtype = ctx.Attr<int>("in_dtype");
     if (in_dtype >= 0) {
```
arlesniak: I think the if block (584-588)  is not needed anymore because the condition is already checked above (563-566).

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -659,24 +660,12 @@ class BroadcastDataMKLDNNHandler
           y->format(), MKLDNNMemoryFormat::undef,
           platform::errors::InvalidArgument("Wrong format set for Y tensor."));
 
-      auto src1_tz = framework::vectorize(y->dims());
       const auto src0_tz = framework::vectorize(x->dims());
```
arlesniak, line 663: Maybe you could be so kind to unify way of naming of variables in this constructor please? Mixing x, y, src_0_tz, src1_md, input_dims is a bit misleading.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -659,24 +660,12 @@ class BroadcastDataMKLDNNHandler
...
-
-      for (size_t i = src1_tz.size(); i < src0_tz.size(); ++i) {
-        src1_tz.insert(src1_tz.begin(), 1L);
-      }
-
       const auto src0_md = dnnl::memory::desc(
           src0_tz, platform::MKLDNNGetDataType<T>(), x->format());
       const auto src1_md = dnnl::memory::desc(
-          src1_tz, platform::MKLDNNGetDataType<T>(), x->format());
+          input_dims, platform::MKLDNNGetDataType<T>(), x->format());
```
arlesniak, line 668: If input_dims represents y dims, so maybe here should be y->format(). What would you suggest ?

paddle/fluid/platform/mkldnn_reuse.h
jakpiase: You're right, I'll unify my naming conventions

paddle/fluid/platform/mkldnn_reuse.h
jakpiase: I cannot use y->format(), because it has fewer dimensions than than x->format(), and oneDNN's reduction primitive does not allow that kind of operation, the number of dimensions must be equal. I am extending y dims outside this constructor and passing extended dimensions as an argument to handle that.

paddle/fluid/operators/reduce_ops/reduce_op.h
jakpiase: You're right, I'll change that. Thank you

paddle/fluid/platform/mkldnn_reuse.h
arlesniak: OK

arlesniak (approved):
LGTM

jczaja (approved):

jakpiase:

luotao1 (approved):

luotao1:

PR #32596, opened 2021-04-26
[OPs] Bug fix, fix the segment mean for illegal syncthreads usage.

ZHUI:
### PR types

Bug fixes
### PR changes

OPs
### Describe

The syncthreads only sync in block. However there should be global sync in the cuda kernel. 
So, we fixed it by launch two kernels.


chenwhql (approved):
LGTM for PADDLE_ENFORCE

wawltor (approved):
LGTM

PR #32628, opened 2021-04-27
Reduce the time cost for the elementwise_add test case

wawltor:
### PR types
Performance optimization

### PR changes
APIs 

### Describe
Reduce the time cost for the elementwise_add test case


ZHUI (approved):
LGTM

PR #32631, opened 2021-04-27
Sum kernel for CPU supporting BF16 and SelectedRows 

arogowie-intel:
### PR types
New features 

### PR changes
 OPs

### Describe
Introduce CPU native kernel supporting SUM operator with BF16 data type and `SelectedRows` input tensors.


paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -961,6 +961,74 @@ class TransposeMKLDNNHandler : public MKLDNNHandler {
   std::vector<int> logical_axis_;
 };
 
+template <typename T>
```
jczaja: I have a sentiment to this Handler but It is not used in this PR

paddle/fluid/operators/math/blas_impl.h
```diff
@@ -28,6 +29,27 @@
 namespace paddle {
 namespace operators {
 namespace math {
+namespace detail {
+
+template <typename T>
+static void axpy(int n, const T alpha, const T *x, const int incx, T *y,
+                 const int incy) {
+  // Y = Y + apha * X
```
jczaja: "apha" -> "alpha"

python/paddle/fluid/tests/unittests/test_sum_op.py
```diff
@@ -324,4 +384,6 @@ def test_list_of_none_input():
 create_test_sum_fp16_class(TestLoDTensorAndSelectedRowsOp)
 
 if __name__ == "__main__":
+    from paddle import enable_static
```
arlesniak: I think it's better to move the import to the top of the file.

python/paddle/fluid/tests/unittests/test_sum_op.py
```diff
@@ -141,6 +143,64 @@ def test_w_is_selected_rows(self):
...
 
+@unittest.skipIf(not core.supports_bfloat16(),
+                 'place does not support BF16 evaluation')
+class TestSelectedRowsSumBF16Op(TestSelectedRowsSumOp):
+    def setUp(self):
+        self.height = 10
+        self.row_numel = 12
+        self.rows = [0, 1, 2, 3, 4, 5, 6]
+        self.dtype = np.uint16
+        self.init_kernel_type()
```
lidanqing-vv, line 160: Hi, 
self.dtype = np.uint16
self.init_kernel_type()
but this function is not defined here so it will use parent's function which will set dtype to np.float64?

lidanqing-vv: sorry my mistake, I mistaken with another class. This class parent function is pass

paddle/fluid/operators/math/selected_rows_functor.cc
```diff
@@ -294,21 +296,17 @@ template struct SelectedRowsAddToTensor<platform::CPUDeviceContext, int64_t>;
...
-    std::is_floating_point<T>::value &&
-    std::is_same<DeviceContext, platform::CPUDeviceContext>::value>::type
-elementwise_add_to(const DeviceContext& ctx, BlasT<DeviceContext, T>* blas,
-                   size_t data_len, const T* in, T* out) {
-  blas->AXPY(data_len, 1., in, out);
+template <typename T>
+typename std::enable_if<std::is_floating_point<T>::value>::type
+elementwise_add_to(BlasT<platform::CPUDeviceContext, T>* blas, size_t data_len,
+                   const T* in, T* out) {
+  blas->AXPY(data_len, T(1.), in, out);
```
arlesniak: maybe 1.0f would be better

paddle/fluid/operators/math/selected_rows_functor.cc
```diff
@@ -412,7 +410,7 @@ struct MergeAdd<platform::CPUDeviceContext, T> {
       out.set_rows(merge_rows);
 
       math::SetConstant<platform::CPUDeviceContext, T> constant_functor;
-      constant_functor(context, out.mutable_value(), 0.0);
+      constant_functor(context, out.mutable_value(), static_cast<T>(0.0));
```
arlesniak: maybe 0.0f  would be better

paddle/fluid/platform/mkldnn_reuse.h
arogowie-intel: @jczaja Do you think that I should remove this handler since as you said "It is not used in this PR" ?

paddle/fluid/operators/math/selected_rows_functor.cc
arogowie-intel: There won't be much difference, because this class template is instantiated both for `float` and `double` so either way in one case there would be implicit cast.

paddle/fluid/operators/math/selected_rows_functor.cc
arogowie-intel: There won't be much difference, because this function template is instantiated both for `float` and `double` so either way in one case there would be implicit cast.

paddle/fluid/operators/math/selected_rows_functor.cc
arlesniak: I was just wondering that with double the converting constructor (bfloat.h:91) would be called, which for double is static_cast to float at first then float one called. Maybe with explicit float we could skip intermediate conversion step, one static_cast, just for performance . What's you opinion?

arogowie-intel: For bfloat16 you're probably right with respect to less conversions. Although for overall performance this would rather won't make much difference. But thanks for pointing out!

paddle/fluid/platform/mkldnn_reuse.h
jczaja: Long term we hope to use it so I would leave it like that , but I'm not sure if coverage CI will allow it to stay not used.

jczaja (dismissed):
LGTM

arlesniak (dismissed):
LGTM

arogowie-intel:
@luotao1 The CI failures are as following:

1. **PR-CI-APPROVAL** is because I'm skipping UT on platforms not supporting BF16 data type.
2. **PR-CI-Coverage** is IMHO a false positive, since I can set a breakpoint and stop in gdb in lines which it doesn't see.  Attached file with backtrace: [sum_op_py_bt.log](https://github.com/PaddlePaddle/Paddle/files/6425774/sum_op_py_bt.log) command: `gdb --args python python/paddle/fluid/tests/unittests/test_sum_op.py TestSelectedRowsSumOpInt`
3. **PR-CI-OP-benchmark** is running GPU benchmarks and this PR's changes are not related to GPU.

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -950,6 +950,74 @@ class TransposeMKLDNNHandler : public MKLDNNHandler {
...
+class AXPYMKLDNNHandler : public platform::MKLDNNHandlerT<T, dnnl::reorder> {
+ public:
+  AXPYMKLDNNHandler(const MKLDNNDeviceContext& dev_ctx,
```
lidanqing-vv: Hi, As I understand, this AXPYMKLDNNHandler is defined but never used and tested? There should be a sum mkldnn kernel or AXPY mkldnn kernel?

The test_sum_op is testing CPU naive sum op not mkldnn sum op right?

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -950,6 +950,74 @@ class TransposeMKLDNNHandler : public MKLDNNHandler {
...
+                    const mkldnn::engine mkldnn_engine,
+                    platform::Place cpu_place, int n, float alpha)
+      : platform::MKLDNNHandlerT<T, dnnl::reorder>(
+            dev_ctx, mkldnn_engine, cpu_place,
+            platform::CreateKey(dev_ctx, static_cast<int64_t>(n),
```
lidanqing-vv: Why this int n has to be cast to int64_t as below if it defined an int ?
     platform::CreateKey(dev_ctx, static_cast<int64_t>(n),
                                platform::MKLDNNGetDataType<T>(), alpha,
                                "-axpy")

lidanqing-vv (commented):
I have some doubts. Thanks

paddle/fluid/platform/mkldnn_reuse.h
arogowie-intel: I have removed this handler as it's not used.

paddle/fluid/platform/mkldnn_reuse.h
arogowie-intel: I have removed this handler as it's not used.

lidanqing-vv:
LGTM.

arogowie-intel:
@luotao1 Could you start your review please? I've explained in earlier comment why some CI runs failed.

luotao1 (approved):

luotao1:

PR #32648, opened 2021-04-28
skip fusing repeated fc when the fc with weight padding

juncaipeng:
### PR types

Bug fixes
### PR changes

Others
### Describe

skip fusing repeated fc when the fc with weight padding


jiweibo (approved):
LGTM

PR #32673, opened 2021-04-29
avoid polluting logging's root logger

iclementine:
### PR types

Bug fixes

### PR changes

Others

### Describe

PaddlePaddle **POLLUTE**s logging library by changing the config of the root logger and adding StreamHandler to it.
Logging library has a hierarchy of loggers. On top of every logger exists the root logger. Every logger's logging is also logged to its ancestors.

This can be done unexpectedly or unintentionally in two ways.

1. `logging.getLogger()`, getting a logger withou passing a name or using `name='root'`. The root logger is returned in this case. Any settings to this logger will have a global effect;
2. calling `logging.basicConfig(*args, **kwargs)`. It's possible to change the formatter and other settings of the root logger in this way, but it's better not to. Also calling `loggig.basicConfig` without passing filename or stream will add a StreamHandler to the root logger.

If the root logger's settings are changed or stream handlers are added to the root logger when paddle is imported, all subsequent use of logging will get some unexpected behavior, redundant screen logging, undesired format, etc.

So it is better to avoid using the root logger, unless you have a good reason to do so.

details: https://stackoverflow.com/questions/24750523/disable-output-of-root-logger

This PR remove all the changes to the roor logger when paddle is imported.

### DIFF
--- a/python/paddle/distributed/fleet/meta_optimizers/sharding_optimizer.py
+++ b/python/paddle/distributed/fleet/meta_optimizers/sharding_optimizer.py
@@ -29,9 +29,12 @@ from paddle.fluid.framework import Program, Variable, name_scope, default_main_p
 from paddle.fluid import layers
 
 import logging
-logging.basicConfig(
-    format='%(asctime)s %(levelname)-8s %(message)s',
-    datefmt='%Y-%m-%d %H:%M:%S')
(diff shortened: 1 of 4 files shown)


jzhang533 (approved):
lgtm

PR #32676, opened 2021-04-29
[Paddle-TRT] remove check for optim_cache_dir in trt slim int8

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

remove check for optim_cache_dir in trt slim int8


jiweibo (approved):
LGTM

PR #32680, opened 2021-04-29
[cherry-pick] skip fuse repeated fc when the fc with weight padding

juncaipeng:
### PR types

Bug fixes
### PR changes

Others
### Describe

skip fuse repeated fc when the fc with weight padding


Superjomn (approved):
LGTM

PR #32713, opened 2021-04-30
[Cherry-pick][Dy2stat] Fix to_tensor Bug Reported from QA (#32701)

zhhsplendid:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Cherry-pick of https://github.com/PaddlePaddle/Paddle/pull/32701

Dy2stat failed when user writes return paddle.to_tensor(xxx), the reason is that visit_Expr doesn't work when the Expr is in return. Some other statements may trigger same bug. To fix it, we re-wrote a transformer to transform paddle.to_tensor to paddle.assign for all Call nodes.


liym27 (approved):
LGTM

lanxianghit (approved):

PR #32715, opened 2021-04-30
Fix traverse graph bug in reducer. 

ForFishes:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix traverse graph bug in reducer. 
During the forward execution, the graph will be traversed to find the unused parameter. Because the intermediate variable clerar of the graph has been changed in the previous reverse direction, some of the checker with input and output failed.


sandyhouse (approved):
LGTM

PR #32716, opened 2021-04-30
Fix the bug in pipeline for dygraph mode

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

Fix the bug in pipeline for dygraph mode to enable dp (data parallel) and mp (model parallel).


ForFishes (approved):
LGTM

PR #32720, opened 2021-04-30
Mechanism that converts startup_program initializers to BF16

wozna:
### PR types

New features
### PR changes

OPs 
### Describe

This PR adds a mechanism to BF16 training that converts initializers from startup_program to BF16. 
The mechanism is added only to pure_bf16 mode. 

The important thing is that if you want to change the initiators to bf16, you need to use the startup_program argument when defining the model in the minimize function.

### DIFF
--- a/python/paddle/fluid/contrib/mixed_precision/bf16/amp_utils.py
+++ b/python/paddle/fluid/contrib/mixed_precision/bf16/amp_utils.py
@@ -232,7 +232,7 @@ def bf16_guard():
         yield
 
 
-def correct_post_op(post_ops, keep_fp32_ops):
+def are_post_ops_bf16(post_ops, keep_fp32_ops):
     for post_op in post_ops:
         for op in post_op:
             if op.type in keep_fp32_ops:
@@ -263,7 +263,7 @@ def cast_initializers_to_bf16(startup_prog,
                     op_post_ops.append(post_op)
                     op_out_vars.append(out_var)
 
-            if change_op and correct_post_op(op_post_ops, keep_fp32_ops):
+            if change_op and are_post_ops_bf16(op_post_ops, keep_fp32_ops):
                 for out_var in op_out_vars:
                     if out_var.dtype == core.VarDesc.VarType.FP32:
                         out_var.desc.set_dtype(core.VarDesc.VarType.BF16)
--- a/python/paddle/fluid/contrib/mixed_precision/bf16/amp_utils.py
+++ b/python/paddle/fluid/contrib/mixed_precision/bf16/amp_utils.py
@@ -232,7 +232,52 @@ def bf16_guard():
         yield
 
 
-def cast_model_to_bf16(program, amp_lists=None, use_bf16_guard=True):
+def correct_post_op(post_ops, keep_fp32_ops):
+    for post_op in post_ops:
+        for op in post_op:
+            if op.type in keep_fp32_ops:
+                return False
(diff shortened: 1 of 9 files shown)


python/paddle/fluid/contrib/mixed_precision/bf16/amp_utils.py
```diff
@@ -232,7 +232,52 @@ def bf16_guard():
         yield
 
 
-def cast_model_to_bf16(program, amp_lists=None, use_bf16_guard=True):
+def correct_post_op(post_ops, keep_fp32_ops):
```
arogowie-intel: The name of this function doesn't align with what it does. Please rename it.

python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
```diff
@@ -209,19 +209,23 @@ def find_true_prev_op(ops, cur_op, var_name):
...
         cur_op (Operator): Current operator which has var_name variable.
         var_name (string): Variable name.
+        search_all (bool): The type of operator search. Use if \"cur_op\" is not in the \"ops\" set. 
     """
     post_op = []
-    for idx, op in enumerate(ops):
-        if op == cur_op:
-            break
+    if search_all:
+        idx = -1
```
arogowie-intel: Please add some comment explaining why you need to go through all ops instead of only looking after current op node.

python/paddle/fluid/contrib/tests/test_bf16_utils.py
```diff
@@ -139,6 +139,26 @@ def test_find_true_post_op(self):
...
+            outputs={"Out": var1},
+            attrs={"shape": var1.shape,
+                   "dtype": var1.dtype,
+                   "value": 1.0})
+
+        op1 = block.append_op(
+            type="abs", inputs={"X": [var1]}, outputs={"Out": [var2]})
+        res = amp.bf16.amp_utils.find_true_post_op(
+            block.ops, inititializer_op, "X", search_all=True)
+        assert (res == [op1])
```
arogowie-intel: You can also add an assert that without `search_all=True` the result is an empty array.

python/paddle/fluid/contrib/mixed_precision/fp16_utils.py
wozna: I added a comment here in the function description, 
https://github.com/PaddlePaddle/Paddle/blob/4cfeb39fa5ba2c7964f14efedac7ccad9ab3e118/python/paddle/fluid/contrib/mixed_precision/fp16_utils.py#L220
Do you think I should expand it?

arogowie-intel: I rather thought about inplace comment explaining why you set `idx = -1` and in which situations you have to search in the way you do it right now. That would be useful note for future developers to understand the behavior. 
Actually the comment you wrote is misleading since even you use `search_all` you are still going through `ops` set, but from the beginning. I thought about noting why you do it like that, just for future.

arlesniak (approved):
LGTM

jczaja (approved):
LGTM

arlesniak:
@luotao1 Could you start your review, please?

arogowie-intel (approved):

luotao1:
@lidanqing-intel Does this PR cherry-pick to release/2.1?

PR #32726, opened 2021-05-03
fix the example code of send/recv api

sandyhouse:
### PR types

Others
### PR changes

Docs
### Describe

fix the example code of send/recv api and remove send_v2 from op_function_generator.cc (https://github.com/PaddlePaddle/Paddle/pull/32504)


zhiqiu (approved):
LGTM for op_function_generator.cc

PR #32742, opened 2021-05-05
[CHERRY-PICK] Reduce grad fix cherrypick

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
[CHERRY-PICK] Fixed reduce backward kernel by changing the source of mkldnn memory format tag. Now it is taken from forward primitive descriptor.
Original commit: [#32592](https://github.com/PaddlePaddle/Paddle/pull/32592)


jczaja (approved):
LGTM

PR #32762, opened 2021-05-06
CPU memory allocator using Transparent Huge Pages

jczaja:
### PR types
Performance optimization

### PR changes
Others

### Describe
CPUAllocator (SystemAllocator for CPU) is now for Linux/MacOS is advised to use Huge Pages when doing allocation. For Operating systems with Huge Pages support , 2MB memory pages may be used to fullfill allocation rather than standard 4KB. This may improve performance(CPU and oneDNN execution) in some scenarios

### DIFF
--- a/paddle/fluid/memory/detail/system_allocator.cc
+++ b/paddle/fluid/memory/detail/system_allocator.cc
@@ -23,6 +23,8 @@ limitations under the License. */
 #include <windows.h>  // VirtualLock/VirtualUnlock
 #else
 #include <sys/mman.h>  // for mlock and munlock
+#include <errno.h>
+#include <string.h>
 #endif
 #include "gflags/gflags.h"
 #include "paddle/fluid/memory/allocation/allocator.h"
@@ -70,8 +72,8 @@ void* AlignedMalloc(size_t size) {
   PADDLE_ENFORCE_EQ(
       error, 0,
       platform::errors::InvalidArgument(
-          "Fail to set advice to CPU memory of %ld size. Error code is  %d.",
-          size, error));
+          "Fail to set advice to CPU memory of %ld size. Error: %s.",
+          size, strerror(errno)));
 #endif
 
 #endif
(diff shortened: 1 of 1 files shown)


jczaja:
@pawelpiotrowicz  Could you please review that one?

pawelpiotrowicz:
@jczaja  Short comment, In my opinion you should keep old alignment (32,4096?) if THP won’t be applicable, it happens when requested size is less  than 2MB.
Besides I could consider skip  madvise()  if requested size is less than 2MB.  I don't know what kind of overhead comes from MADV_SEQUENTIAL or do you have use-case for this in PP. 
Code looks fine 😊

arogowie-intel (commented):
I have two thoughts:
* I'm wondering whether this should be a default behavior? 
* Also I'm thinking whether it's possible to test this code? All that comes to my mind is that you could at least check whether the allocated pointer alignment requirement is satisfied. Although as you said this only _advice_ for system...

jczaja:
@pawelpiotrowicz My bad for not explaining the context. So this code : system_allocator is used as part of PaddlePaddle own CPU allocator. So typically there is one allocation: 500MB and then all data takes any amount of memory when needed for some tensors etc. so that is why no check whether allocation of size less than 2MB is present because it will not happen. technically user can ask for preallocation of 2MB , but at this scenario user will point to use on the fly allocations (FLAGS_use_system_allocator) .

jczaja:
@arogowie-intel 
- As for checking alignment: **posix_memalign** is having second argument a value specifying requested alignment.So this makes sure that if only memory is allocated that it will be aligned to requested size.
- Regarding , testing. As two runs of same model may look diffrently e.g. one will get huge pages used and other will not due to decision of Linux Kernel it is not very obvious from experiments what is performance improvement. I gathered some data that it should be 1% across tested models. Yet when this PR is merged then perf CI will test it longer (a week) and we will see more precise performance impact

jakpiase:
@jczaja In my opinion there should be an easy way to disable this behavior, since as you have said sometimes huge pages will be enabled and sometimes not and it is nondeterministic. Probably at some point somebody may want to test performance without huge pages to get deterministic results.

lidanqing-vv:
@jczaja  Did you check performance with our CI. And there are may CIs failed, could you please maintain it?

jczaja:
@lidanqing-intel Performance gain was around 0 to 1% so I decided to work on more beneficial tasks.

PR #32605, opened 2021-04-27
[NPU] add input EpsilonTensor for adam

zhiqiu:
### PR types

Performance optimization
### PR changes

OPs
### Describe

[NPU] add input EpsilonTensor for adam

The timeline below show `adam npu kernel  on ernie-3.0 model`, we can see that there are several `time bubbles`  between two adam kernels.

image

One of the major reason, the npu op `ApplyAdamD` only aceepts `bata1`, `beta2`, `epsilon` as tensor Input.
However, in paddle, these are commonly passed as attribute (single float number).

Actually, `bata1`, `beta2` can  be passed by `Beta1Tensor`, `Beta2Tensor`, so this PR adds  `EpsilonTensor` for better 
performance.

### performance
- before
image

- after
image

**Speed up: 19448 tokens/s -> 23033 tokens/s, + 18.4 %**

PR #32780, opened 2021-05-07
[NPU] support global accumulator for adam

zhiqiu:
### PR types

Performance optimization
### PR changes

OPs
### Describe

[NPU] support global accumulator for adam

As described in https://github.com/PaddlePaddle/Paddle/pull/32605,  on ernie-3.0 model, we can see that there are several time bubbles between two adam kernels.
image

There are mainly two problems that cost much time, 

1. update beta1/beta2 pow

2. convert beta1/beta2/epsilon to NPU tensor

PR https://github.com/PaddlePaddle/Paddle/pull/32605 solves one problem 2, and this PR tries to solve problem 1.

### why
The original implementation of AdamOptimizer creates beta1_pow and beta2_pow for each parameter and updates them in adam op for each parameter. 
**It SHOULD be pointed out that, actually, the value of the beta1_pow and beta2_pow of each parameter is the same.**

This works fine in adam CUDA kernel, since the beta1_pow and beta2_pow can be updated fast. However, in NPU kernel, it requires to call two `mul` op and cost much time.

### How
So, we introduce `global beta_pow`, which means only creates one beta1_pow and one beta2_pow for all the parameters of the whole model. 
Specificlly,

- Add bool attribute `use_global_beta_pow` to adam op. If true, the outputs(Beta1PowOut, Beta2PowOut) will not be used in adam op, "
and beta_pow will be updated after all adam op in the model.
- add bool parameter `use_global_beta_pow` to `paddle.fluid.optimizer.Adam`, If true, Adam will use global beta_pow for whole model instead of creating beta_pow for each parameter.

As can be seen in the timeline, there is no mul between two `ApplyAdam`
image

### Performance
- before
image

- after
image

22211 -> 24481 tokens/s, +10 %

### DIFF
--- a/paddle/fluid/operators/optimizers/adam_op.cc
+++ b/paddle/fluid/operators/optimizers/adam_op.cc
@@ -198,6 +198,13 @@ class AdamOpMaker : public framework::OpProtoAndCheckerMaker {
                   "(bool, default false) "
                   "Whether to use multi-precision during weight updating.")
         .SetDefault(false);
+    // TODO(zhiqiu): We could set Beta1Pow, Beta2Pow, Beta1PowOut, Beta2PowOut
+    // as dispensable since they are not used when use_global_beta_pow is true.
+    AddAttr<bool>("use_global_beta_pow",
+                  "(bool, default false) "
+                  "Whether to use global beta_pow for whole model instead of "
+                  "creating beta_pow for each parameter.")
+        .SetDefault(false);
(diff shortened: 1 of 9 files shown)


liym27 (dismissed):
LGTM

wanghuancoder (approved):
LGTM

phlrain (approved):

PR #32629, opened 2021-04-27
[Paddle-TRT] Implement MHA fp16 order same as training

zlsh80826:
### PR types

Others

### PR changes

OPs

### Describe

[DEMO ONLY]
TRT MHA plugin change to apply scale before QK multiplication

PR #32785, opened 2021-05-07
[Paddle-TRT] Implement MHA fp16 order same as training (#32629)

shangzhizhou:
### PR types
Bug fixes

### PR changes
Others

### Describe
cherry pick #32629  Implement MHA fp16 order same as training ，to avoid overflow of fp16.

* implement MHA order same as training

* fix fp16 compile issue on old architecture

* fix format

* fix format


Superjomn (approved):
LGTM

PR #32858, opened 2021-05-12
[NPU] use SparseSoftmaxCrossEntropyWithLogits in npu kernel of softmax_with_cross_entropy

zhiqiu:
### PR types

Function optimization
### PR changes

OPs
### Describe

[NPU] refine npu kernel of softmax_with_cross_entropy

The original implementation of `softmax_with_cross_entropy` uses 5 NPU operators (`SoftmaxV2, Cast, OneHotD, Cast, SoftmaxCrossEntropyWithLogits`),
which may introduce too much kernel launch cost (as we know, NPU kernel launch cost 100us on average).
Same for the grad op.

The new implementation uses only 1 operator(SparseSoftmaxCrossEntropyWithLogits) for forward and 1 operator(Mul) for backward.

Plus, this PR adds a dispensable output `Backprop` for softmax_with_cross_entropy to do the backward, which is only used when compiled with NPU support.

image
image

### DIFF
--- a/paddle/fluid/operators/slice_op_npu.cc
+++ b/paddle/fluid/operators/slice_op_npu.cc
@@ -29,13 +29,10 @@ void UpdateAttr(const framework::DDim in_dims, const std::vector<int> axes,
                 const std::vector<int> starts, const std::vector<int> ends,
                 std::vector<int>* offsets, std::vector<int>* size) {
   int cnt = 0;
-  VLOG(4) << in_dims;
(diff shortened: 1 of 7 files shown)


zhhsplendid (approved):
LGTM for ShareDataWith because the usage is similar to old softmax_with_cross_entropy_op

zhangting2020 (approved):
LGTM

phlrain (approved):

PR #32859, opened 2021-05-12
[Dy2Static]Add param_guard in ParameterList to support @to_static

Aurelius84:
### PR types

Bug fixes
### PR changes

APIs
### Add param_guard in ParameterList to support @to_static

Before this PR， if Model codes use `ParameterList` as container in `__init__()` function, it will cause `No Input(X)` error while applying @to_static to convert into static program. 

See the following code, this PR fixed this bug.

```python
class NetWithParameterList(paddle.nn.Layer):
        super(NetWithParameterList, self).__init__()
        weight = self.create_parameter([in_size, out_size])
        bias = self.create_parameter([out_size], is_bias=True)
        self.params = paddle.nn.ParameterList([weight, bias])

    @to_static
    def forward(self, x):
        out = paddle.matmul(x, self.params[0])   # Cause `No Input(X)` error 
        out = paddle.add(out, self.params[1])
        out = paddle.tanh(out)
        return out
```

### DIFF
--- a/python/paddle/fluid/dygraph/container.py
+++ b/python/paddle/fluid/dygraph/container.py
@@ -15,6 +15,7 @@
 from collections import OrderedDict
 from ..framework import Parameter
 from .layers import Layer
+from .base import param_guard
 
 __all__ = [
     'Sequential',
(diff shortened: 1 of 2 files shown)


liym27 (approved):
LGTM

PR #32868, opened 2021-05-12
Polish code for _getitem_impl_

liym27:
### PR types

Function optimization
### PR changes

APIs
### Describe

As the title

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -792,29 +792,16 @@ def _getitem_impl_(var, item):
     if not isinstance(item, tuple):
         item = [item]
 
-    decrease_axis = []
-    slice_axis = []
-    slice_start = []
-    slice_end = []
-    slice_step = []
+    decrease_axes = []
+    axes = []
+    starts = []
+    ends = []
+    steps = []
+
     use_strided_slice = False
     reverse_axis = []
-    target_block = default_main_program().current_block()
 
-    def fill_constant(shape, value, force_cpu=False, out=None):
-        var.block.append_op(
-            type='fill_constant',
-            inputs={},
-            outputs={'Out': [out]},
-            attrs={
-                'shape': shape,
-                'dtype': out.dtype,
-                'value': float(value),
-                'force_cpu': force_cpu
-            })
-        out.stop_gradient = True
-        return out
-
+    max_integer = 2**31 - 1
     for dim, slice_item in enumerate(item):
         if isinstance(slice_item, slice):
             start = slice_item.start
@@ -824,8 +811,7 @@ def _getitem_impl_(var, item):
             if start is None and end is None and step is None:
                 continue
 
-            if step is None:
-                step = 1
+            step = 1 if step is None else step
 
             if start is None and end is None:
                 assert (step == -1)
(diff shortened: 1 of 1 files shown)


zhiqiu (approved):
LGTM

python/paddle/fluid/framework.py
```diff
@@ -836,106 +822,56 @@ def fill_constant(shape, value, force_cpu=False, out=None):
...
-                attrs['ends'].append(dim)
-    else:
-        attrs['ends'] = slice_end
 
-    # strides
-    if use_strided_slice == True:
-        if contain_var(slice_step):
-            inputs['StridesTensorList'] = get_new_list_tensor(slice_step)
-            for i, dim in enumerate(slice_step):
+    infer_flags = list(1 for i in range(len(axes)))
```
zhhsplendid, line 849: Would you like to get a list of 1s and the length of the list is same as len(axes)?

If so, you could simply this line as:
`infer_flags = list(1 for i in axes)`

python/paddle/fluid/framework.py
```diff
@@ -836,106 +822,56 @@ def fill_constant(shape, value, force_cpu=False, out=None):
...
+    from .layers import utils
```
zhhsplendid, line 850: Can we import at the beginning of the python file?

python/paddle/fluid/framework.py
```diff
@@ -836,106 +822,56 @@ def fill_constant(shape, value, force_cpu=False, out=None):
...
     attrs = {
-        'axes': slice_axis,
+        'axes': axes,
         'starts': [],
         'ends': [],
-        'decrease_axis': decrease_axis
+        'decrease_axis': decrease_axes
     }
-    if (use_strided_slice == True):
+    if use_strided_slice == True:
```
zhhsplendid, line 846: Can we write `if use_strided_slice:` ?

python/paddle/fluid/framework.py
```diff
@@ -948,7 +884,7 @@ def get_new_list_tensor(old_list):
             attrs=attrs)
 
         out = slice_out_var
-    elif use_strided_slice == True and len(slice_axis) > 0:
+    elif use_strided_slice == True and len(axes) > 0:
```
zhhsplendid, line 887: Can we replace `elif use_strided_slice == True and len(axes) > 0` with `elif use_strided_slice and len(axes) > 0` ?

Usually if a variable `x` is boolean, we write `if x` instead of `if x is True` or `if x == True`

Reference from official programming recommendations:
https://www.python.org/dev/peps/pep-0008/#programming-recommendations

image

python/paddle/fluid/framework.py
liym27: Thanks. I use `infer_flags = [1] * len(axes)` to replace it for better readability.

python/paddle/fluid/framework.py
liym27: No. Because it will cause a circular reference.

python/paddle/fluid/framework.py
liym27: Yes. Fixed.

python/paddle/fluid/framework.py
liym27: Thanks, done.

python/paddle/fluid/framework.py
zhhsplendid: OK to me

zhhsplendid (approved):
LGTM

PR #32872, opened 2021-05-12
change unique op VisitaDataType from small to tiny

MingMingShangTian:
### PR types

Performance optimization
### PR changes

OPs
### Describe

Change the unique op visiting data strategy from VisitDataTypeSmall to  VisitDataTypeTiny.
VisitDataTypeTiny visit only int32 and int64, which is fitting the type of Index.


chenwhql (approved):
LGTM

PR #32890, opened 2021-05-13
fix the bug in the creation of pp groups to avoid hang

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix the bug in the creation of pp groups to avoid hang


ForFishes (approved):
LGTM

PR #32911, opened 2021-05-14
Polish code for setitem and getitem

liym27:
### PR types

Function optimization
### PR changes

APIs
### Describe

As the title


python/paddle/fluid/variable_index.py
```diff
@@ -0,0 +1,306 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import numpy as np
+from . import unique_name
+from . import core
+
+MAX_INTEGER = 2**31 - 1
```
zhhsplendid, line 20: Should we use `sys.maxsize`?

liym27: No. 
`sys.maxsize is 2**63 - 1`.  Because the maximum of  c++ `int` is `2**31 - 1`,  pass `2**63 - 1` from Python to c++ will result in data out of bounds.

zhhsplendid (approved):
LGTM

zhiqiu (approved):
LGTM

PR #32941, opened 2021-05-17
fix the paddle compare op for the broadcast when the element equal

wawltor:
### PR types
 Bug fixes

### PR changes
 OPs 

### Describe
fix the paddle compare op for the broadcast when the element equal


paddle/fluid/operators/controlflow/compare_op.cc
```diff
@@ -131,18 +131,18 @@ class CompareOp : public framework::OperatorWithKernel {
 
 REGISTER_COMPARE_OP(less_than, "Out = X < Y");
 REGISTER_COMPARE_KERNEL(less_than, CPU, paddle::operators::LessThanFunctor,
-                        paddle::operators::GreaterEqualFunctor);
```
ZHUI, line 134: Better to have more annotation.

ZHUI (approved):
LGTM

PR #32967, opened 2021-05-18
Add warning for dataloader incompatable upgrade

heavengate:
### PR types
Others

### PR changes
APIs

### Describe
Add warning for DataLoader output data format incompatable upgrade when dataset only contains single filed.

屏幕快照 2021-05-18 下午7 27 39

test code:
```
import numpy as np
import paddle
from paddle.io import DataLoader, Dataset

class RandomDataset(Dataset):
    def __getitem__(self, idx):
        return np.random.random((2, 3)).astype('float32')

    def __len__(self):
        return 10

dataset = RandomDataset()
loader = DataLoader(dataset, batch_size=1)
for data in loader():
    print(data)
```


python/paddle/fluid/dataloader/fetcher.py
```diff
@@ -12,18 +12,54 @@
...
 # limitations under the License.
 
+import logging
+from ..log_helper import get_logger
+
+try:
+    from collections.abc import Sequence
+except:
+    from collections import Sequence
+
```
jzhang533, line 19: python2 和 python 3.5都已经drop了，这里可以不需要再顾忌`collections.abc`了吧。

heavengate: Done, thanks!

chenwhql (approved):
LGTM

PR #32972, opened 2021-05-18
Add leaf enforce check for set_value

chenwhql:
### PR types

Bug fixes
### PR changes

Others
### Describe

Add leaf enforce check for set_value to polish error message, the leaf Tensor can't use inplace strategy.

original error message:

image

after changing:

image


pangyoki (approved):
LGTM

liym27 (approved):

PR #33000, opened 2021-05-19
Support index is tensor or list for getitem in static mode 

liym27:
### PR types

New features
### PR changes

APIs
### Describe

As the title

For example:
```
tensor_x = paddle.to_tensor(np.array([[1,2],[3,4]]).astype(np.float32))
tensor_x[[0,0]] 
```


zhhsplendid (approved):
LGTM

PR #33016, opened 2021-05-20
[NPU] cast indices and label if their type is not consistent in accuracy npu op

pangyoki:
### PR types

Bug fixes

### PR changes

OPs

### Describe

The inputs `indices` and `label` in `accuracy` op may be different types, resulting in an error when doing equal operations.
In this PR, compare the types of `indices` and `label` before the `equal` operation. If they are inconsistent, cast them to int32.

unittest result:
图片

GLOG:

图片


zhiqiu (approved):
LGTM

zhangting2020 (approved):

PR #33019, opened 2021-05-20
replace complex64/128 with complex template in cast Op 

MingMingShangTian:
### PR types

Others
### PR changes

Others
### Describe

* replace with paddle.to_tensor and print.
* fix the bug in cast OP when cast between complex64 and complex128 types, and add test cases.


chenwhql (approved):
LGTM

PR #33024, opened 2021-05-20
Add eccl and cross stack profiler 

lw921014:
### PR types

New features

### PR changes

OPs and Profiler tools

### Describe

1、we replace hccl for paddle with eccl
image
2、we add a cross stack profiler tools to help train super ernie


cmake/external/ascend.cmake
```diff
@@ -65,6 +65,7 @@ endif()
 if(WITH_ASCEND_CL)
   set(ASCEND_CL_DIR ${ASCEND_DIR}/ascend-toolkit/latest/fwkacllib/lib64)
 
+  set(inf_eccl_lib /home/liuwei88/01_eccl/ECCL/build/lib/libeccl.so)
```
void-main: let's try not use hard coded path in code

paddle/fluid/operators/collective/c_broadcast_op_npu.cc
```diff
@@ -48,30 +49,28 @@ class CBroadcastOpASCENDKernel : public framework::OpKernel<T> {
...
-    PADDLE_ENFORCE_NPU_SUCCESS(platform::dynload::HcclBroadcast(
-        ptr, numel, dtype, (uint32_t)root, comm->comm(), stream));
+    PADDLE_ENFORCE_NPU_SUCCESS(platform::dynload::eccl_broadcast(
+        ptr, out_ptr, numel, dtype, root, comm->comm().c_str(), stream, AUTO));
 
     VLOG(3) << "rank " << comm->rank() << " invoke Bcast. recieved "
             << framework::product(out->dims());
 
-    dev_ctx->Wait();
+    // dev_ctx->Wait();
```
void-main: how about remove these useless code?

paddle/fluid/operators/collective/c_comm_init_hccl_op.cc
```diff
@@ -42,21 +42,21 @@ class CCommInitOpAscend : public framework::OperatorBase {
                       platform::errors::PreconditionNotMet(
                           "CCommInitOpAscend can run on npu place only."));
 
-    auto var = scope.FindVar(Input("X"));
-    PADDLE_ENFORCE_NOT_NULL(
-        var, platform::errors::InvalidArgument("Input con not be empty."));
+    // auto var = scope.FindVar(Input("X"));
```
void-main: same as above

paddle/fluid/operators/collective/c_comm_init_hccl_op.cc
```diff
@@ -67,7 +67,7 @@ class CCommInitOpAscend : public framework::OperatorBase {
 class CCommInitOpAscendMaker : public framework::OpProtoAndCheckerMaker {
  public:
   void Make() override {
-    AddInput("X", "Raw variable contains a NCCL UniqueId instaces.");
+    // AddInput("X", "Raw variable contains a NCCL UniqueId instaces.");
```
void-main: same as above

void-main (commented):
Comm ops part LGTM

cmake/external/ascend.cmake
lw921014: how about use the same path as hccl.so, then we need to place this package there at advance?

PR #33025, opened 2021-05-20
fix path error on windows when precision switch is turn on

XieYunshen:
### PR types

Others
### PR changes

Others
### Describe

fix path error on windows when precision switch is turn on
image


lelelelelez (approved):
LGTM

PR #33060, opened 2021-05-23
Support OutType tmeplate argument in broadcast  branch

JamesLim-sy:
### PR types

Others

### PR changes

OPs

### Describe

1. Adding OutType template argument into  elementwise_broadcast  template function, for supporting the CUDA OPs below:
    - equal
    - not euqal
    -  large_than
    -  less_than
    -  large_equal
    -  less_equal
The return type of above OP is boolean.


Xreki (approved):
LGTM

PR #33112, opened 2021-05-25
add cross stack profiler to profile super ernie

lw921014:
### PR types

New features
### PR changes

Others
### Describe

We develop a **C**ross **S**tack **P**rofile **(CSP**) tool to help those SOTA models obtain best performance, such as super ernie, GPT-3 and so on,  especially when leveraging many sophisticated distributed strategies inside a large-scale cluster which may has thousands of GPUs. 

**CSP** is the first profile toolkit in the industry, which
1. can collect cross stack data, include **system** resources, **framework** runtime info and **user**-defined  description;
2. can convert **gigebytes raw data** from thousands of nodes into **useful knowledge** with **little overhead and  code intrusion**;
3.  can achieve **nearly real-time display** by making full use of **multi-core processor** to generate **friendly tracing file**. 

Following is an example, and we try to make sure that our model run the same mode  as that mentioned in the paper.

figure 1 running model mentioned in paper
image
figure 2 running model  from CSP
image

### DIFF
--- a/tools/CrossStackProfiler/CspChromeTraceFormatter.py
+++ b/tools/CrossStackProfiler/CspChromeTraceFormatter.py
@@ -1,0 +1,140 @@
+# Copyright (c) 2021 PaddlePaddle Authors. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


gongweibao (dismissed):
LGTM

raindrops2sea (approved):

PR #33133, opened 2021-05-26
Modify Ops from complex64/128 to complex<float/double> types.

MingMingShangTian:
### PR types

Others
### PR changes

OPs
### Describe

Modify Ops from complex64/128 to complex<float/double> types.
Details as follows.
* kron
* reshpae
* slice
* trace
* transpose


chenwhql (approved):
LGTM

PR #33152, opened 2021-05-27
support input integer with shape parameter of reshape

yghstill:
### PR types

Function optimization
### PR changes

APIs
### Describe

Support input integer with shape parameter of reshape, such as:
```python
a = paddle.ones([2,5]).reshape(-1)
```


JZ-LIANG (dismissed):
LGTM

PR #33162, opened 2021-05-27
[quant] Add quant wrap for functional api and refine the qat

juncaipeng:
### PR types

Others
### PR changes

Others
### Describe

* Add layer wrapper for functional api
* Refine dygraph qat
  * For conv2d and linear, apply fake quant to the inputs
  * For the ops of which the output scale is different from the input scale, apply fake quant to the outputs
  * For the ops of which the output scale is the same as the input scale, calculate the moving average threshold of the outputs


DannyIsFunny (approved):
LGTM

XieYunshen (approved):

PR #33212, opened 2021-05-31
Add the op def for conv2d, hard_swish, leaky_relu, relu and swish

juncaipeng:
### PR types

Others
### PR changes

Others
### Describe

Add the op def for conv2d, hard_swish, leaky_relu, relu and swish


Superjomn (approved):
LGTM

PR #33265, opened 2021-06-01
Add assertion to ensure that size of each dim of paramerter is larger than 0

iclementine:
### PR types

Bug fixes

### PR changes

Others

### Describe

Add an assertion to ensure that the size of each dim of the parameter to be created is larger than 0.

When trying to create a parameter that has some dim's size == 0, currently an exception would be raised when the initializer is calculating some shape related statistics. But this exception is not always raised because some dimensions' sizes are not relavant to these statistics and some initializers do not compute such statistics.

So we add an exception to ensure that size of each dimension of the parameter to be created is larger than 0 to avoid consequent errors. A parameter with num_element == 0 is meaningless in this case.

### DIFF
--- a/python/paddle/fluid/layer_helper_base.py
+++ b/python/paddle/fluid/layer_helper_base.py
@@ -312,6 +312,10 @@ class LayerHelperBase(object):
         if not attr:
             return None
         assert isinstance(attr, ParamAttr)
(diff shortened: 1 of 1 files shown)


guoshengCS (approved):
LGTM

PR #33269, opened 2021-06-01
Fix jetson cuda arch configure when compiling with single arch

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix jetson cuda arch configure when compiling with single arch


shangzhizhou (approved):
LGTM

PR #33291, opened 2021-06-02
fix ctc loss first input as logits, not log_probs

zh794390558:
### PR types
Bug fixes

### PR changes
Docs

### Describe
fix ctc loss first input as logits, not log_probs


iclementine (approved):
LGTM

PR #33298, opened 2021-06-02
Support index is list bool for getitem in static mode

liym27:
### PR types

New features
### PR changes

APIs
### Describe

As the title

For example:
```
tensor_x = paddle.to_tensor(np.array([[1,2],[3,4]]).astype(np.float32))
tensor_x[[True,False]] 
```

### DIFF
--- a/python/paddle/fluid/variable_index.py
+++ b/python/paddle/fluid/variable_index.py
@@ -140,19 +140,34 @@ def _getitem_impl_(var, item):
             end = MAX_INTEGER if end is None else end
 
         elif isinstance(slice_item, list):
+            is_bool_list = False
             for i in slice_item:
-                if not isinstance(i, int):
-                    raise TypeError("Only support int value in list")
+                if not isinstance(i, (int, bool)):
+                    raise TypeError("Only support int or bool in list.")
+
+                if isinstance(i, bool):
+                    is_bool_list = True
+                    break
 
             if len(item) != 1:
                 raise IndexError(
                     "When index contains a list, its length must be 1, but received {}".
                     format(len(item)))
 
+            if is_bool_list:
+                new_slice_item = []
+                for idx, ele in enumerate(slice_item):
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/variable_index.py
```diff
@@ -139,19 +139,34 @@ def _getitem_impl_(var, item):
...
             if len(item) != 1:
                 raise IndexError(
                     "When index contains a list, its length must be 1, but received {}".
                     format(len(item)))
 
+            if is_bool_list:
+                new_slice_item = []
+                for idx, ele in enumerate(slice_item):
+                    if not isinstance(ele, bool):
+                        raise TypeError("Only support bool in list.")
```
zhhsplendid: Here the TypeError should be "Mixed bool index with other types is not supported"

"Only support bool in list." may confuse users that we can only use boolean index.

python/paddle/fluid/variable_index.py
```diff
@@ -139,19 +139,34 @@ def _getitem_impl_(var, item):
             end = MAX_INTEGER if end is None else end
 
         elif isinstance(slice_item, list):
+            is_bool_list = False
             for i in slice_item:
-                if not isinstance(i, int):
-                    raise TypeError("Only support int value in list")
+                if not isinstance(i, (int, bool)):
+                    raise TypeError("Only support int or bool in list.")
```
zhhsplendid: "Only support int or bool in list" -> "Only support int or bool in index list"

Otherwise users may be confused about the list thing, them may not know we refer to the index list.

python/paddle/fluid/variable_index.py
liym27: Thanks, done.

python/paddle/fluid/variable_index.py
liym27: Thanks, done.

zhhsplendid (approved):
LGTM

PR #33320, opened 2021-06-03
Convert all blocks in program into SSAgraphs.

levi131:
### PR types
New features

### PR changes
Others

### Describe
This is a draft for convert all blocks in program into SSAgraphs.


zhhsplendid (commented):
Any unit test?

levi131:
I'll add some unit tests. @zhhsplendid

paddle/fluid/framework/ir/graph.cc
```diff
@@ -17,10 +17,110 @@ limitations under the License. */
 #include "paddle/fluid/framework/ir/graph.h"
 #include "paddle/fluid/framework/operator.h"
 
+DEFINE_bool(ssa_program, false, "Convert all blocks in program into SSAgraphs");
```
zhhsplendid: `ssa_graph` may not be a good name. You can name as what it does: `convert_subblocks`

paddle/fluid/framework/ir/graph.cc
```diff
@@ -175,6 +276,56 @@ void Graph::ResolveHazard(
...
 }
 
+#if FLAGS_ssa_program
+std::shared_ptr<Graph> Graph::Clone() {
+  PADDLE_ENFORCE_EQ(
+      this->parent_, nullptr,
+      platform::errors::InvalidArgument(
+          "This graph is a subgraph, and can't be cloned individually"));
+  auto cloned_graph = std::make_shared<Graph>(this->program_);
+  cloned_graph->ReleaseSubgraphs();
```
zhhsplendid: Is making a Graph from Program enough? If it is enough, we don't have to release and add here.

levi131: I think the graph could be changed by some optimization pass, and is different form the initial graph constructed from program. So I release and add subgraph here.

paddle/fluid/framework/ir/graph.cc
levi131: I'll rename 'ssa_program'.

paddle/fluid/framework/ir/graph.h
```diff
@@ -252,11 +258,29 @@ class Graph {
...
+  std::unique_ptr<Graph> CloneSubgraph(const size_t idx);
+#else   // FLAGS_ssa_program
   std::map<std::string, std::vector<ir::Node *>> InitFromProgram(
       const ProgramDesc &program);
+#endif  // FLAGS_ssa_program
 
   // NOTE: program_ shouldn't be exposed to user.
   const ProgramDesc program_;
+#if FLAGS_ssa_program
+  const Graph *parent_;  // not owned.
```
zhhsplendid: Call it parent may not be so good.

In program, global block is special. Can we call it `global_graph_`?

paddle/fluid/framework/ir/graph.cc
```diff
@@ -17,10 +17,110 @@ limitations under the License. */
...
 #include "paddle/fluid/framework/operator.h"
 
+DEFINE_bool(ssa_program, false, "Convert all blocks in program into SSAgraphs");
+
 namespace paddle {
 namespace framework {
 namespace ir {
 
+#if FLAGS_ssa_program
+Graph::Graph(const ProgramDesc &program) : program_(program) {
```
zhhsplendid: It's unclear to me that when you construct a `Graph` from `Program`, you convert all blocks into sub_graphs, then what's the use of `parent_`? I would suggest to put `global_block` to `parent_`(you can rename it as 'global_graph_' to mapping 'global block'), then other `sub_blocks` into `sub_graphs`

paddle/fluid/framework/ir/graph.h
levi131: As far as I know, there are at least two global programs in the ladder runtime. I think it would lead to some misunderstanding.

Another consideration is that, as the basic function of converting program into graph, I think it could be better to regard it as an independent function rather than limited to the case of global program to global graph.

paddle/fluid/framework/ir/graph.h
```diff
@@ -89,11 +93,26 @@ class Graph {
   }
 
   bool Has(const std::string &attr_name) const {
+    if (FLAGS_convert_all_blocks) {
+      if (IsMainGraph()) {
+        LOG(WARNING) << "This graph is main_graph, which doesn't have nodes or "
```
zhhsplendid: You don't have to put a warning on it since it is a normal behavior, same for other APIs

paddle/fluid/framework/ir/graph_test.cc
```diff
@@ -264,5 +264,140 @@ TEST(GraphTest, TestAttrCopy) {
   ASSERT_FALSE(dst_g.Has(kFloatValue));
 }
 
+TEST(GraphTest, TestMultiBlock) {
+  if (FLAGS_convert_all_blocks) {
```
zhhsplendid: If you would like unit test to run automatically in every CI PR, you should set FLAGS_convert_all_blocks to true here, and set it back to false at the end of this test case.

paddle/fluid/framework/ir/graph.h
levi131: I have deleted the warnings.

paddle/fluid/framework/ir/graph_test.cc
levi131: I have modified the codes to ensure GraphTest.TestMultiBlock works.

zhhsplendid (dismissed):
LGTM

paddle/fluid/framework/ir/graph.h
```diff
@@ -251,12 +339,50 @@ class Graph {
...
+    sub_graphs_.clear();
+  }
+
+  void AddSubGraph(std::unique_ptr<Graph> sub_graph) {
+    PADDLE_ENFORCE_EQ(
+        this->IsMainGraph(), true,
+        platform::errors::InvalidArgument("This graph is not main_graph"));
+    sub_graphs_.push_back(std::move(sub_graph));
+  }
+
```
thisjiang, line 399: 麻烦加个`size_t SubGraphSize() const`接口吧~

paddle/fluid/framework/ir/graph.cc
```diff
@@ -17,15 +17,114 @@ limitations under the License. */
...
+  Set<const std::vector<OpDesc *>>(details::kStaleProgramOpDescs,
+                                   new std::vector<OpDesc *>(block.AllOps()));
+  return var_nodes;
+}
+
+// TODO(levi): delete this interface after when we can convert all
+// blocks into sub_graphs.
 std::map<std::string, std::vector<ir::Node *>> Graph::InitFromProgram(
     const ProgramDesc &program) {
   VLOG(3) << "block in program:" << program_.Size();
```
thisjiang, line 86: 这儿是否可以改为调用`InitFromBlock`呢？即改为
```
return InitFromBlock(program.Block(0));
```

levi131: done

paddle/fluid/framework/ir/graph.h
levi131: Add interface: size_t Graph::SubGraphsSize() const {}

zhhsplendid (dismissed):
LGTM

Aurelius84 (approved):
LGTM

zhhsplendid (approved):
LGTM

Xreki (approved):

PR #33327, opened 2021-06-03
[sharding] bugfix for group init hang

JZ-LIANG:
### PR types
Bug fixes

### PR changes
OPs

### Describe
add sync calc stream after allreduce to make a cpu barrier for  group init


wangxicoding (approved):
LGTM

PR #33377, opened 2021-06-07
Add 'self' parameters to function Cluster::update_pods, use variable name 'cur_proxy' instead of 'proxy' in file 'tools/get_pr_ut.py'

Jiangxinz:
### PR types
    Bug fixes 
### PR changes
    Others

### Describe
1. In 'python/paddle/distributed/utils.py', function ***update_pods*** of class ***Cluster*** lacks first parameters 'self' as it is neccesary as member function of class, just like *this* pointer in cpp.
2. In 'tools/get_pr_ut.py', there is a bug of ***undefined variable***: In function ***__urlretrieve*** of class ***PRChecker***, use variable name **cur_proxy** instead of **proxy**.

### DIFF
--- a/python/paddle/distributed/utils.py
+++ b/python/paddle/distributed/utils.py
@@ -166,7 +166,7 @@ class Cluster(object):
     def __ne__(self, cluster):
         return not self.__eq__(cluster)
 
-    def update_pods(cluster):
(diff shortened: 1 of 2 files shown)


lelelelelez (approved):
LGTM

PR #33381, opened 2021-06-07
bugfix: param init with fill constant str_value

JZ-LIANG:
### PR types
Bug fixes

### PR changes
OPs

### Describe
in fp64 setting, user set beta1_pow_acc = 0.9, beta2_pow_acc = 0.999 for adamw,

before bugfix the actual value are: beta1_pow_acc = 0.8999999761581421, beta2_pow_acc = 0.9990000128746033
after bugfix the actual value are: beta1_pow_acc = 0.9000000000000000, beta2_pow_acc = 0.9990000000000000

fill constant should set str_value attr to preserve precision in fp64 setting

### DIFF
--- a/python/paddle/fluid/initializer.py
+++ b/python/paddle/fluid/initializer.py
@@ -152,6 +152,7 @@ class ConstantInitializer(Initializer):
             out_dtype = var.dtype
             out_var = var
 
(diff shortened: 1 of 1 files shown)


ForFishes (approved):
LGTM

PR #33394, opened 2021-06-07
[Paddle-TRT] add support for trt dynamic shape flatten op

cryoco:
### PR types

Function optimization
### PR changes

Others
### Describe

add support for trt dynamic shape flatten op


shangzhizhou (approved):
LGTM

PR #33413, opened 2021-06-08
Fix gather infer shape using axis

ForFishes:
### PR types

 Bug fixes
### PR changes

Others
### Describe

Fix gather infer shape using axis

### DIFF
--- a/paddle/fluid/operators/gather_op.cu
+++ b/paddle/fluid/operators/gather_op.cu
@@ -31,52 +31,82 @@ class GatherOpCUDAKernel : public framework::OpKernel<T> {
     auto *index = ctx.Input<Tensor>("Index");
     auto *output = ctx.Output<Tensor>("Out");
 
+    int axis = ctx.Attr<int>("axis");
+
+    // get axis from tensor
     if (ctx.HasInput("Axis")) {
-      const Tensor *axis = ctx.Input<Tensor>("Axis");
-      const auto &index_type = index->type();
-      const auto &axis_type = axis->type();
-      auto place = ctx.GetPlace();
-      if (index_type == framework::proto::VarType::INT32 &&
-          axis_type == framework::proto::VarType::INT32) {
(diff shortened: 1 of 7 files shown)


paddle/fluid/operators/gather_op.cu
```diff
@@ -31,47 +31,33 @@ class GatherOpCUDAKernel : public framework::OpKernel<T> {
...
+      } else if (index_type == framework::proto::VarType::INT64) {
+        GatherV2CUDAFunction<T, int64_t>(x, index, axis, output, place, ctx);
       }
       return;
     }
+
     output->mutable_data<T>(ctx.GetPlace());
     if (x->numel() == 0) return;
-    const auto &index_type = index->type();
-    bool index_type_match = index_type == framework::proto::VarType::INT32 ||
```
sandyhouse, line 64: Is this enforcement needed?

python/paddle/tensor/manipulation.py
```diff
@@ -862,34 +862,39 @@ def gather(x, index, axis=None, name=None):
...
+            attrs={'axis': axis,
+                   'overwrite': False},
+            outputs={"Out": out})
+    else:
+        helper.append_op(
+            type="gather",
+            inputs={"X": x,
+                    "Index": index,
+                    "Axis": axis},
+            attrs={"overwrite": False},
```
sandyhouse, line 895: Why set the overwrite to False as the default one is True?

paddle/fluid/operators/gather_op.cu
ForFishes: not need, it should be only INT32/INT64. If not, it will be cast int dtype

python/paddle/tensor/manipulation.py
ForFishes: This will only appear when there are duplicate values in the index. The behavior of 2.0api removes the parameter interface of overwrite, so its default value must be Flase, or the reverse calculation is wrong.

sandyhouse (approved):
LGTM

PR #33416, opened 2021-06-08
fix undefined variable in optimizer

zhiqiu:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix undefined variable in optimizer


pangyoki (approved):
LGTM

PR #33439, opened 2021-06-09
Class for  processing program 

huangxu96:
### PR types

New features
### PR changes

Others
### Describe

This PR creates a  class to process the program at the C++ level. Currently, this class has two class methods:
        GetInputsOutputsInBlock()
              This method can accumulate all ops in/outputs varibles of a block.
              e.g. :

              Block0: 
                var1 va2 var3
                op1 input: {var1, var2}
                op1 output:{var3}
                var4 var5 
                op2 input : {var3, var4}
                op2 output: {var5}

        After calling this method:
             inputs, outputs = nullptr;
             GetInputsOutputsInBlock(Block0, &inputs, &outputs);
             inputs = {var1, var2, var3, var4}
             outputs = {var3, var5}

        AddDepToBlockOp()
              This method can add all input/output variables of a block into one Op (typically a control flow Op), we call this process is adding a dependency. The purpose of doing so is for the convenience of distributed training.
               e.g. :

               Block0:                                                            Block1:
                   var1, var2, var3                                           var5, var6, var7
                   op1 input: {var1, var2}                                op2 input: {var5, var6}
                   op1 output:{var3}                                       op2 output : {var7}
                   var4
                   while op -------------> sub_block:
                   while op input: {var3}
                   while op output: {var4}
   
         After calling this method:
              AddDepToBlockOp(block0)
             
                 Block0:                                                           Block1:
                   var1, var2, var3                                           var5, var6, var7
                   op1 input: {var1, var2}                                op2 input: {var5, var6}
                   op1 output:{var3}                                       op2 output : {var7}
                   var4
                   while op -------------> sub_block:
                   while op input: {var3, var5, var6}
                   while op output: {var4, var7}


paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,84 @@
...
+          (*inner_inputs).insert(in_var_name);
+      }
+
+    for (auto oname : op->OutputNames())
+      for (auto out_var_name : op->Outputs().find(oname)->second) {
+        (*inner_outputs).insert(out_var_name);
+      }
+  }
+
+  // Step2: Remove LOD_TENSOR_ARRAY created in current control flow block.
```
zhhsplendid: Why comment LOD_TENSOR_ARRAY but no related code around?

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,84 @@
...
+  // Step1: update inner_inputs and inner_outputs
+  // NOTE: Here assumes that all variables are input or output of Ops,
+
+  std::set<std::string> removed_inner_inputs;
+
+  for (OpDesc *op : current_block.AllOps()) {
+    for (auto iname : op->InputNames())
+      for (auto in_var_name : op->Inputs().find(iname)->second) {
+        if ((*inner_outputs).find(in_var_name) == (*inner_outputs).end())
+          (*inner_inputs).insert(in_var_name);
```
zhhsplendid: `inner_inputs->insert(in_var_name) ` is a common coding pattern in this case. Same for other similar places:

`(*pointer).` can be replaced by `pointer->`

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+}
+
+void ProgramProcessor::AddDepToBlockOp(const BlockDesc &block) {
+  VLOG(3) << "Op size:" << block.AllOps().size();
+  for (OpDesc *op : block.AllOps()) {
+    if (op->HasAttr("sub_block")) {
+      auto sub_block = op->GetAttr("sub_block");
+      // TODO(huangxu96): sub_block is an Attr, how to get a BlockDesc*?
```
zhhsplendid: Comment to your comment :-)

If you read the code of `GetAttr`, you can find that it returns `Attribute`, which is defined by:
```
using Attribute = boost::variant<
    boost::blank, int, float, std::string, std::vector<int>, std::vector<float>,
    std::vector<std::string>, bool, std::vector<bool>, BlockDesc*, int64_t,
    std::vector<BlockDesc*>, std::vector<int64_t>, std::vector<double>>;
```
So, you could convert the `Attribute` to `BlockDesc*` by handling boost variant.

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,139 @@
...
+      // recursively processing
+      // VLOG(3)<<"sub_block" << sub_block;
+      // AddDepToBlockOp(sub_block);
+
+      VariableNameMap sub_inputs;
+      VariableNameMap sub_outputs;
+      ProgramProcessor::GetInputsOutputsInBlock(block, &sub_inputs,
+                                                &sub_outputs);
+      VLOG(3) << "sub_inputs.size:" << sub_inputs.size();
```
zhhsplendid, line 89: This log information may not be enough for your future debug, I suggest to also log the block op type, so that it help you to debug.

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,139 @@
...
+  :param inner_outputs: Output VariableNameMap of ops in current block. */
+
+  // Step1: update inner_inputs and inner_outputs
+  // NOTE: Here assumes that all variables are input or output of Ops,
+
+  std::set<std::string> inner_inputs_var;
+  std::set<std::string> inner_outputs_var;
+
+  for (OpDesc *op : current_block.AllOps()) {
+    for (auto iname : op->InputNames()) {
```
zhhsplendid, line 33: It seems wrong...

In my understanding, Paddle's op->Input has `input` parameter like 'X', 'Y', then the VariableNameMap mappings from 'X', 'Y' to real variable argument.

Let me give `elementwise_add` as an example: 

`{Out = ['out_var'] = elementwise_add{inputs={X=['feed_1'], Y=['feed_2']}, ...some attrs...}`

The real argument variable names are 'feed_1', 'feed_2', not 'X', 'Y'

In your code, `iname` is the 'X', 'Y', I am afraid of `inner_inputs->emplace(iname, op->Input(iname))` can cause problem when some ops have same input parameter names like 'X', 'Y'

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,139 @@
...
+          inner_inputs_var.insert(in_var_name);
+          AddInput = true;  // here we assume if one var is not in the
+                            // outputs_var list, we add this input into the
+                            // inputs list.
+        }
+      }
+      if (AddInput) inner_inputs->emplace(iname, op->Input(iname));
+    }
+
+    for (auto oname : op->OutputNames()) {
```
zhhsplendid, line 40: Why is logic of output names different to input names?

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,139 @@
...
+      for (auto out_var_name : op->Output(oname)) {
+        VLOG(3) << "out_var_name:" << out_var_name;
+        inner_outputs_var.insert(out_var_name);
+      }
+      inner_outputs->emplace(oname, op->Output(oname));
+    }
+  }
+
+  // Step2: Remove LOD_TENSOR_ARRAY created in current control flow block.
+  BlockDesc *parent_block = current_block.ParentBlock();
```
zhhsplendid, line 49: I am still not understand these code and you haven't replied my previous comment. Could you tell me why and how did you handle LOD_TENSOR_ARRAY here?

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,139 @@
...
+      for (auto in_var_name : op->Input(iname)) {
+        VLOG(3) << "in_var_name:" << in_var_name;
+        if (inner_outputs_var.find(in_var_name) == inner_outputs_var.end()) {
+          inner_inputs_var.insert(in_var_name);
+          AddInput = true;  // here we assume if one var is not in the
+                            // outputs_var list, we add this input into the
+                            // inputs list.
+        }
+      }
+      if (AddInput) inner_inputs->emplace(iname, op->Input(iname));
```
zhhsplendid: Since we only care about variable names, not op input parameter names like 'X', 'Y', I suggest you can use `std::set<std::string>` instead of the `VariableNameMap`

paddle/fluid/framework/program_processing.cc
huangxu96: Here we should determine if a variable is created in the current block. If a var cannot be found in the parent block, it should be an inner var and cannot be block input. LOD_TENSOR_ARRAY here just meas this variable should be type of LOD_TENSOR_ARRAY.

paddle/fluid/framework/program_processing.cc
huangxu96: OK, actually I didn't commit some log information.

paddle/fluid/framework/program_processing.cc
huangxu96: COOL! I notified this, I will dive deep to see how to convert it.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
zhhsplendid: Good! But I don't think we should handle LOD_TENSOR_ARRAY specially and I saw your code had changed.

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,135 @@
...
+  VarDesc *current_block_var;
+
+  if (parent_block) {
+    for (auto in_var_name : *inner_inputs) {
+      VLOG(3) << "recursively find var:" << in_var_name;
+      VarDesc *parent_block_var = parent_block->FindVarRecursive(in_var_name);
+      if (current_block.HasVar(in_var_name))
+        current_block_var = current_block.FindVar(in_var_name);
+      if (parent_block_var == nullptr && current_block_var &&
+          current_block_var->GetType() == proto::VarType::LOD_TENSOR) {
```
zhhsplendid: Remove the condition: `&& current_block_var->GetType() == proto::VarType::LOD_TENSOR`.

In our case, we don't have to handle special type.

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,135 @@
...
+        VLOG(3) << "remove inner var:" << in_var_name;
+        removed_inner_inputs.insert(in_var_name);
```
zhhsplendid: Code is correct but not beautiful enough. You can write:

line 58: 
```
for (auto iter = inner_inputs->begin(); iter != inner_inputs->end(); ++iter) {
    const std::string& in_var_name = *iter;
    ...
    // line 66
    inner_inputs->erase(iter);
}
```

In this case, you can omit the code of `removed_inner_inputs` and `set_difference`

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,135 @@
...
+      for (auto out_var_name : op->Output(oname)) {
+        VLOG(3) << "insert out_var_name:" << out_var_name;
+        inner_outputs->insert(out_var_name);
+      }
+    }
+  }
+
+  // Step2: Remove variable created in current control flow block.
+  BlockDesc *parent_block = current_block.ParentBlock();
+  VarDesc *current_block_var;
```
zhhsplendid: Dangerous code: you should write
`VarDesc *current_block_var = nullptr;`
in your case because C++ doesn't initialize the value, and you have a condition `&&current_block_var` at line 63.

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,135 @@
...
+      VLOG(3) << "recursively find var:" << in_var_name;
+      VarDesc *parent_block_var = parent_block->FindVarRecursive(in_var_name);
+      if (current_block.HasVar(in_var_name))
+        current_block_var = current_block.FindVar(in_var_name);
+      if (parent_block_var == nullptr && current_block_var &&
+          current_block_var->GetType() == proto::VarType::LOD_TENSOR) {
+        VLOG(3) << "remove inner var:" << in_var_name;
+        removed_inner_inputs.insert(in_var_name);
+      }
+    }
```
zhhsplendid, line 70: I think we should also remove the output variables created in current control flow block, current code just removes the input variables

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,135 @@
...
+
+      auto *op_inputs = op->MutableInputs();
+      std::vector<std::string> *op_input_var_vec;
+      VLOG(3) << "op_type:>>>>>>" << op_type;
+      if (op_type.compare("while") == 0)
+        op_input_var_vec = &((*op_inputs)["kX"]);
+      else if (op_type.compare("conditional_block") == 0)
+        op_input_var_vec = &((*op_inputs)["kInputs"]);
+      else
+        // Only support while_op and conditinal_block_op now
```
zhhsplendid, line 100: Please log a warning for this case, if our program runs here, it means something is wrong.

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,135 @@
...
+      std::set<std::string> sub_outputs;
+      ProgramProcessor::GetInputsOutputsInBlock(*sub_block, &sub_inputs,
+                                                &sub_outputs);
+      VLOG(3) << "sub_inputs.size:" << sub_inputs.size();
+      VLOG(3) << "sub_outputs.size:" << sub_outputs.size();
+
+      auto *op_inputs = op->MutableInputs();
+      std::vector<std::string> *op_input_var_vec;
+      VLOG(3) << "op_type:>>>>>>" << op_type;
+      if (op_type.compare("while") == 0)
```
zhhsplendid: Please add braces due to code style.

Reference: https://google.github.io/styleguide/cppguide.html#Conditionals

paddle/fluid/framework/program_processing.cc
```diff
@@ -0,0 +1,135 @@
...
+        op_input_var_vec = &((*op_inputs)["kX"]);
+      else if (op_type.compare("conditional_block") == 0)
+        op_input_var_vec = &((*op_inputs)["kInputs"]);
+      else
+        // Only support while_op and conditinal_block_op now
+        continue;
+
+      for (auto sub_input : sub_inputs) {
+        if (std::find(op_input_var_vec->begin(), op_input_var_vec->end(),
```
zhhsplendid, line 107: Please write `op_input_var_vec.find(sub_input) == op_input_var_vec->end()`

std::find takes O(N), but container.find may take less time for some containers. So I would suggest you to use container.find when position is the whole container.

BTW, code is shorter, personally.

paddle/fluid/framework/program_processing_test.cc
```diff
@@ -0,0 +1,221 @@
...
+          << while_op->Output("kOutputs").size();
+
+  program_processor.AddDepToBlockOp(*global_block);
+  VLOG(3) << "After AddDependency, while op's input kX size:"
+          << while_op->Input("kX").size();
+  VLOG(3) << "After AddDependency, while op's output kOutPuts size:"
+          << while_op->Output("kOutputs").size();
+
+  ASSERT_EQ(8UL, while_op->Input("kX").size());
+  ASSERT_EQ(6UL, while_op->Output("kOutputs").size());
```
zhhsplendid: Besides the size, could you also check Var name is correct?

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing.cc
huangxu96: Done.

paddle/fluid/framework/program_processing_test.cc
huangxu96: Tried it, but failed. I wrote the var name in the commnet.

huangxu96: Done

paddle/fluid/framework/program_processing.cc
zhhsplendid: 1. You shouldn't `continue` after `throw`
2. I suggest to log warning instead of throwing an exception in case of other operators may contain "sub_block" attributes.

zhhsplendid (approved):
LGTM

zhiqiu (dismissed):
LGTM

zhhsplendid (dismissed):
LGTM

zhiqiu (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #33440, opened 2021-06-09
Check the installed openblas version in cmake

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

We need to use openblas with version 0.3.7, some old version may lead to compilation errors. For example,
image

However, if the openblas is already installed in the system, the original `cblas.cmake` does not check the version and use it directly.

This PR fixes that, and skip the installed openblas if its version != 0.3.7

Pasted Graphic 1

### DIFF
--- a/cmake/cblas.cmake
+++ b/cmake/cblas.cmake
@@ -69,15 +69,21 @@ if(NOT DEFINED CBLAS_PROVIDER)
     PATHS ${OPENBLAS_LIB_SEARCH_PATHS})
 
   if(OPENBLAS_LAPACKE_INC_DIR AND OPENBLAS_INC_DIR AND OPENBLAS_LIB)
-    set(CBLAS_PROVIDER OPENBLAS)
-    set(CBLAS_INC_DIR ${OPENBLAS_INC_DIR} ${OPENBLAS_LAPACKE_INC_DIR})
-    set(CBLAS_LIBRARIES ${OPENBLAS_LIB})
+    file(READ "${OPENBLAS_INC_DIR}/openblas_config.h" config_file)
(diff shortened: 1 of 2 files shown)


wanghuancoder (approved):
LGTM

cmake/cblas.cmake
```diff
@@ -69,15 +69,21 @@ if(NOT DEFINED CBLAS_PROVIDER)
...
-    add_definitions(-DPADDLE_USE_OPENBLAS)
-    add_definitions(-DLAPACK_FOUND)
-
-    message(STATUS "Found OpenBLAS (include: ${OPENBLAS_INC_DIR}, library: ${CBLAS_LIBRARIES})")
-    message(STATUS "Found lapack in OpenBLAS (include: ${OPENBLAS_LAPACKE_INC_DIR})")
+    file(READ "${OPENBLAS_INC_DIR}/openblas_config.h" config_file)
+    string(REGEX MATCH "OpenBLAS ([0-9]+\.[0-9]+\.[0-9]+)" tmp ${config_file})
+    string(REGEX MATCH "([0-9]+\.[0-9]+\.[0-9]+)" ver ${tmp})
+    
+    if (${ver} VERSION_EQUAL "0.3.7")
```
jzhang533, line 76: 使用系统自带的openblas的版本的话，也锁定在了0.3.7，是否过于严格了？

我在mac上编译的时候，用git clone下来的0.3.7版本编译，会出现openblas导致的链接错误：`ld: symbol(s) not found for architecture x86_64`，使用`brew install openblas`装的0.3.15版本的openblas，才能编过。

zhiqiu: 已改为限制 >=0.3.7, https://github.com/PaddlePaddle/Paddle/pull/33626

PR #33461, opened 2021-06-09
[NPU] flatten params and grads, fuse grad_clip and optimizer op

zhiqiu:
### PR types

Performance optimization
### PR changes

OPs
### Describe

[NPU] flatten params and grads, fuse grad_clip and optimizer op

For example, `ernie-3.0` model has 300+ parameters, and thus 300+ gradients of parameters.

Each training step, the program has to perform grad_clip the gradient and update the parameter. So, there are 300+ grad_clip operators and 300+ optimizer operators.
image

This PR tries to flatten all the parameters into one continuous memory space and also flatten the gradients. After that, some of the gradient clip and optimizer can be done by 1 time on the flattened parameter/gradient.
image

Currently, `Adam + ClipByGlobalNorm` is supported. 

#### Performance
ernie-3.0, bs=20480, training speed： 20684->23583 tokens/s, +13.8%


zhangting2020:
paddle/optimzier/optimizer.py是不是也需要同步修改？

python/paddle/fluid/optimizer.py
```diff
@@ -68,7 +69,15 @@ def __init__(self,
                  parameter_list=None,
                  regularization=None,
                  grad_clip=None,
+                 flatten_param_grads=False,
+                 align_size=-1,
                  name=None):
+        """
+        Args:
+            flatten_param_grads (bool, optional): Whether to flatten all the parameters and grads. 
+                If true, the parameters and gradients will be coalesce to continue mempry,
```
zhangting2020: continue mempry -> contiguous memory

zhiqiu: done, thanks

phlrain (approved):

Issue #33021, opened 2021-05-20
config. SetMkldnnCacheCapacity is useless 


juncaipeng:
Ubuntu
C++
MKLDNN

Download the demo (Link:https://dubox.com/s/1RVo80hVAC0zr5iZ2EuG0xQ Password:3v5m ) and refer to the readme.

lidanqing-vv:
Reproduced

lidanqing-intel:
Not Jacek's commit caused the problem. I m using bisect to find which commit caused memory rising.

lidanqing-intel:
a5627df331b913c047f78789c4c9b3ee5edfe65c caused thie memory rising.

lidanqing-vv:
I just tested again, From #32664 there are memory leak @jczaja Please fix this issue, thank you very much!

jczaja:
@luotao1 ,
While working on this issue I had remove(#33174, still testing) some code you added at https://github.com/PaddlePaddle/Paddle/pull/18580 . My question is do you remember why restoring oneDNN caching was needed as done here:
https://github.com/PaddlePaddle/Paddle/blob/481ee79fc92304f33165f7ed0679f16c36862cea/paddle/fluid/inference/api/analysis_predictor.cc#L332-L335

luotao1:
I remember it is used to fix a detect model https://github.com/PaddlePaddle/Paddle/pull/18081#issuecomment-502101243 which input dims are dynamic, the model is in `paddle/fluid/inference/tests/api/analyzer_detect_tester.cc`

`config. SetMkldnnCacheCapacity` is used to balance the memory and speed. You can see the following data from the detect model. (#18372) 

 | speed(ms) | memory | CacheCapacity | 
|---|---|---|
| 26.1437 | 1.5G | CacheCapacity=10 | 
| 24.0129 | 3.5G | CacheCapacity=50 |

jczaja:
@juncaipeng Just to post status of work on that one. There are at least three sources of memory leaks. We understand two of them. Fix to first one is https://github.com/PaddlePaddle/Paddle/pull/33174.

jczaja:
@juncaipeng Two PRs with fixes to this issue were merged. Please test.

lidanqing-vv:
@OliverLPH  Hi I talked with Juncai, please test this issue, since two PRs are merged and should fix this.

PR #33471, opened 2021-06-09
[oneDNN] Second fix to #33021

jczaja:
### PR types
Bug fixes

### PR changes
Others

### Describe
It is second part of fixes to #33021. Problem addressed here was that Predictor after finishing execution of Run() is restoring default settings of cache. Those default settings are unrestriced size of cache (cache_capacity = 0). but after Run() is completed , There are sometimes Tensor:: CopyToCpu() methods called which are using oneDNN objects and could make cache growing.

### DIFF
--- a/paddle/fluid/inference/api/analysis_predictor.cc
+++ b/paddle/fluid/inference/api/analysis_predictor.cc
@@ -341,8 +341,6 @@ void AnalysisPredictor::MkldnnPreSet(
     platform::MKLDNNDeviceContext::tls().set_cur_mkldnn_session_id(
         platform::MKLDNNDeviceContextThreadLocals::
             kMKLDNNSessionID_CacheClearing);
-    platform::MKLDNNDeviceContext::tls().set_cur_input_shape_cache_capacity(
-        config_.mkldnn_cache_capacity_);
     // Set current_input_shape for caching dynamic shape.
     std::stringstream ss;
     for (size_t i = 0; i < inputs_shape.size(); ++i) {
@@ -353,6 +351,9 @@ void AnalysisPredictor::MkldnnPreSet(
     VLOG(2) << "Set input shape=" << ss.str();
     platform::MKLDNNDeviceContext::tls().set_cur_input_shape_str(ss.str());
   }
+  platform::MKLDNNDeviceContext::tls().set_cur_input_shape_cache_capacity(
+      config_.mkldnn_cache_capacity_);
+
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


paddle/fluid/inference/api/analysis_predictor.cc
```diff
@@ -353,6 +351,9 @@ void AnalysisPredictor::MkldnnPreSet(
     VLOG(2) << "Set input shape=" << ss.str();
     platform::MKLDNNDeviceContext::tls().set_cur_input_shape_str(ss.str());
   }
+  platform::MKLDNNDeviceContext::tls().set_cur_input_shape_cache_capacity(
+      config_.mkldnn_cache_capacity_);
+
```
lidanqing-vv, line 356: Is there difference to move position of this line of code?  What is influenced ?

paddle/fluid/inference/tests/api/analyzer_detect_functional_mkldnn_tester.cc
```diff
@@ -120,6 +120,19 @@ void validate_cache_onednn(int cache_capacity = 1) {
...
+  auto output_names = predictor->GetOutputNames();
+  auto output_t = predictor->GetOutputTensor(output_names[0]);
+  std::vector<int> output_shape = output_t->shape();
+  size_t out_num = std::accumulate(output_shape.begin(), output_shape.end(), 1,
+                                   std::multiplies<int>());
+  std::vector<float> out_data;
+  out_data.resize(out_num);
+  output_t->CopyToCpu(out_data.data());
+
+  // Release predictor (relevant cache should be emptied)
```
lidanqing-vv, line 135: "predictor.reset(nullptr);" I feel we don't call often call this in API tests. Usually we just do predictor.Run(), that's all and seems assuming that cache will be release during deconstructor of some object? 
What will be the result if we remove this predictor.reset(nullptr) ?

paddle/fluid/inference/api/analysis_predictor.cc
jczaja: Previously , cache_cpacity was only set when it was non-zero (cache clearing mode) . But as zero value is default it was
not set. But if you have two predictors : one working in restriced cache (clear mode) and other without this limitation
then if you run executor with restricted cache size and then go to execution of seconf executor that is being run on default settings then you would actually use restriced caching again. I mean we were doing reverting to defaults in MKLDNNPostReset() , but I we cannot do it anymore so I had to move this resetting to when execution begins.

paddle/fluid/inference/tests/api/analyzer_detect_functional_mkldnn_tester.cc
jczaja: predictor.reset(nullptr)  This enforces that predictor is released and its cached objects as well. Normally you do not need to do it, as it is released when get out of scope, but I want to check if cache is released so I could not wait till end of scope.

jakpiase (approved):
LGTM

paddle/fluid/inference/api/analysis_predictor.cc
lidanqing-vv: Oh ok it is now in preset function. Sorry I did not notice. Thanks for your explanation

lidanqing-vv:
LGTM.

lidanqing-vv (approved):
LGTM

jczaja:
As agreed with @juncaipeng After passing internal review (two reviewers) for PRs specific to oneDNN we are allowed to merge .

PR #33473, opened 2021-06-09
[CP] fix the bug in the creation of pp groups to avoid hang #32890

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix the bug in the creation of pp groups to avoid hang
fix a typos


ForFishes (approved):
LGTM

PR #33476, opened 2021-06-09
fix the bug of barrier op: sync with calculation stream

sandyhouse:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix the bug of barrier op: sync with calculation stream


ForFishes (approved):
LGTM

PR #33478, opened 2021-06-09
fix inaccurate parameter count (resolve #33477)

### LINKED ISSUE #33477: Inaccurate model parameter count
**System information**
****************************************
Paddle version: 2.0.0
Paddle With CUDA: True

OS: Ubuntu 16.04
Python version: 3.7.10

CUDA version: 10.2.89
cuDNN version: 7.6.5
Nvidia driver version: 440.82
****************************************
**problem file**
`python/paddle/hapi/model_summary.py`

**To Reproduce**
    - get summer of any model with  `batch norm` layer
    - you will find that all trainable parameters in the batch norm will not be count

**Other info**
    - this is because there are parameters in BN that are not trainable, and that breaks the counting logic

DarthThomas:
### PR types
Bug fixes

### PR changes
OPs

### Describe
resolve #33477

Treat different weights separately in layers. So the case when a layer contains 'trainable' and 'not trainable' parameters at the same time would be handled correctly.


LielinJiang (dismissed):

luotao1:
@DarthThomas Sorry for too late to notice this PR:
- please resolve the conflict first
- please don't contain other people's commits, here only https://github.com/PaddlePaddle/Paddle/pull/33478/commits/6bff2b690ba32bded39f8e1dcdab23be3504a618 is useful.

PR #33484, opened 2021-06-10
Combine amp and qat

juncaipeng:
### PR types

Others
### PR changes

Others
### Describe

* Register FP16 kernels for fake_quant_dequant_ops
* Add  fake_quant_dequant_ops to the WHITE_LIST of amp
* Add the unit test of combining amp and qat


paddle/fluid/imperative/amp_auto_cast.cc
```diff
@@ -184,6 +184,17 @@ NameVarBaseMap AutoCastInputs(const std::string& op_type,
...
+
+    if (op_type == "moving_average_abs_max_scale") {
+      for (const auto& pair : ins) {
+        if (pair.first == "X" &&
+            pair.second.front()->DataType() ==
+                framework::proto::VarType::FP16) {
+          dst_type = framework::proto::VarType::FP16;
+        }
+      }
+    }
```
zhiqiu: can it be handled like "batch_norm", as the code below?

juncaipeng: moving_average_abs_max_scale only considers the dtype of input(X) to determine the dest_type. However, "batch_norm" just cast the input(X) to FP16 or FP32. These are two different cases.

zhiqiu (approved):
LGTM

PR #33492, opened 2021-06-10
[HeterPs] add delta score, scale show

Thunderbrook:
### PR types
Others

### PR changes
Others

### Describe
add delta score
scale grad by show
log format


danleifeng (approved):
LGTM

PR #33499, opened 2021-06-10
[Dy2stat] Add Support for a, b = static_variable Grammar

zhhsplendid:
### PR types

New features
### PR changes

Others
### Describe

For python, if users write `a, b = var`, the `__getitem__` method will iterate through 0, 1, 2 ... until `__getitem__` throws an IndexError, then stop. The var[0], var[1] will be given to a, b respectively. If more values are given, the unpack size would cause error.

We didn't raise the IndexError in the past and we add statement in `__getitem__` to raise IndexError here to support grammar like `a, b = var` in this PR.


Aurelius84 (approved):
LGTM

liym27 (approved):

PR #33535, opened 2021-06-11
1, remove layernorm dynamic fp16; 2, let reshape out in dynamic shape

shangzhizhou:
### PR types
Others

### PR changes
Others

### Describe
1. remove useless code（layernorm dynamic plugin's fp16 precision not enough yet）
2. let reshape plugin out with dynamic shape


cryoco (approved):
LGTM

PR #33543, opened 2021-06-12
Use importlib to detect availability of cv2.

bwoodsend:
### PR types
Function optimization

### PR changes
Others

### Describe

This PR replaces the ugly *FIXME* labeled check for whether or not `cv2` is installed with an importlib call which accomplishes the same thing as the original.

The main motivation for getting rid of the original `subprocess.Popen([sys.executable, ...])` is because:
- It fails for embedded Python (where `sys.executable` is typically `None`)
- It causes an infinite loop of subprocesses when frozen with PyInstaller where `sys.executable` is an executable equivalent to `python main-code.py`. In other words, any python script containing `import paddle` will launch itself in a subprocess, then that subprocess will launch itself again in another subprocess, and so on until your computer crashes.

### DIFF
--- a/python/paddle/dataset/image.py
+++ b/python/paddle/dataset/image.py
@@ -34,18 +34,14 @@ from __future__ import print_function
 
 import six
 import numpy as np
-# FIXME(minqiyang): this is an ugly fix for the numpy bug reported here
+
+# Test for cv2 avaibility. Do so without importing it so as to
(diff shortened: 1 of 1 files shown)


bwoodsend:
I see that that cmake build error on Windows also happens in other PRs so I very much doubt that it's relevant.

PR #33335, opened 2021-06-03
make paddle.to_tensor() copy if data is tensor

zhiqiu:
### PR types

Bug fixes
### PR changes

APIs
### Describe

- make paddle.to_tensor() copy if data is varbase, so the computation graph is splited from which of data.
- make the  stop_gradient of input(data) not changed

PR #33570, opened 2021-06-15
Add return when input is tensor

zhiqiu:
### PR types

Performance optimization
### PR changes

APIs
### Describe

Add return when input is tensor, related #33335


pangyoki (approved):
LGTM

PR #33576, opened 2021-06-15
Change dtype of flops and params used in flops calculation to prevent integer overflow (from int32 -> int64)

yingyibiao:
### PR types

Bug fixes
### PR changes

APIs
### Describe

#33481


guoshengCS (approved):
LGTM

PR #33662, opened 2021-06-18
[Paddle-TRT] TensorRT8 void** compatibility

zlsh80826:
### PR types

Others

### PR changes

Others

### Describe

TensorRT8 change the API for all argument with ``void**`` (https://docs.nvidia.com/deeplearning/tensorrt/release-notes/tensorrt-8.html#rel_8-0-0-EA)
>Methods that take parameters of type void** where the array of pointers is unmodifiable are now changed to take type void* const*

So this PR modifies the ``void**`` to ``void* const*`` if the compiled TRT version is not less than 8.0


shangzhizhou (approved):
LGTM

cryoco (approved):
LGTM

raindrops2sea (approved):
LGTM

PR #33666, opened 2021-06-18
fix the but that concat op can't support uint8

youth123:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix the but that concat op can't support uint8


vslyu (approved):
LGTM

sandyhouse (approved):
LGTM

PR #33745, opened 2021-06-23
[Dy2Stat]Support Python3 type hint

Aurelius84:
### PR types

Others
### PR changes

Others
### Describe

Support Python3 type hint

Example：

```python
class LinearNetWithTuple2(BaseLayer):
        super(LinearNetWithTuple2, self).__init__(in_size, out_size)

    def forward(self, x) -> Tuple[paddle.Tensor, np.array]:
        out1, out2 = self.build(x)
        return (out2, np.ones([4, 16]))


class LinearNetWithDict(BaseLayer):
        super(LinearNetWithDict, self).__init__(in_size, out_size)

    def forward(self, x) -> Dict[str, paddle.Tensor]:
        out1, out2 = self.build(x)
        return {'out': out2}
```

### DIFF
--- a/python/paddle/fluid/tests/unittests/dygraph_to_static/test_origin_info.py
+++ b/python/paddle/fluid/tests/unittests/dygraph_to_static/test_origin_info.py
@@ -65,7 +65,7 @@ class TestOriginInfo(unittest.TestCase):
         self.func = simple_func
 
     def set_static_lineno(self):
-        self.static_abs_lineno_list = [3, 4, 5]
+        self.static_abs_lineno_list = [5, 6, 7]
 
     def set_dygraph_info(self):
         self.line_num = 3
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/tests/unittests/dygraph_to_static/test_origin_info.py
```diff
@@ -65,7 +65,7 @@ def set_test_func(self):
         self.func = simple_func
 
     def set_static_lineno(self):
-        self.static_abs_lineno_list = [3, 4, 5]
+        self.static_abs_lineno_list = [5, 6, 7]
```
Aurelius84, line 68: Because we add two `import` statement, so here we adjust line infos by +2

zhhsplendid (approved):
LGTM

PR #33748, opened 2021-06-23
fix bug when the cuda kernel config exceeds dims max

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix bug when the cuda kernel config exceeds dims max
close https://github.com/PaddlePaddle/PaddleClas/pull/685, #33107

When use 3D dim3 config for the blocks of cuda kernel, the limit of each dim is (1024, 1024, 64). For example
```
dim3 num_blocks(x,y,z), num_threads(32, 8, 1);
kernel<<<num_blocks, num_threads>>>(param)
```
The condition shouble be: x <=1024, y <=1024, z <=64, otherwise, "cudaErrorInvalidConfiguration" will be raised.

In layer_norm_grad, it uses 3D dim3 (1, batch_size, 1) as blocks config. While, batch_size may > 1024.
This PR change `3D dim3 (1, batch_size, 1)` to `1D dim3 (batch_size)` to solve the problem.

### DIFF
--- a/paddle/fluid/operators/layer_norm_op.cu
+++ b/paddle/fluid/operators/layer_norm_op.cu
@@ -398,9 +398,9 @@ __global__ void LayerNormBackwardComputeGradInput(
     const U *__restrict__ mean, const U *__restrict__ var, const float epsilon,
     const U *gamma, T *grad_input) {
 #ifdef __HIPCC__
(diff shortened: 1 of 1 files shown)


pangyoki (approved):
LGTM

PR #33749, opened 2021-06-23
Add fusion_gru and multi_gru to PTQ (Post-Training Quantization)

wozna:
### PR types

New features

### PR changes

OPs 
### Describe

This PR: 
- adds support for `fusion_gru` and `multi_gru` to Post-Training Quantization process
- adds gathering scales based on the max absolute value per output channel for fusion_gru/multi_gru operators
- adds GRU model test for PTQ for with `multi_gru_fuse_pass` and without it
- fix in the requantize operator so that there is no hard-coded data format but that it is set based on input


wozna:
I have a problem that in `PR-CI-Coverage` for  file paddle/fluid/inference/api/mkldnn_quantizer.cc coverage is 0% even if all UT added in this PR  run this code multiple times. I added the same changes in PR https://github.com/PaddlePaddle/Paddle/pull/33912 where I added a few prints and made the test not pass, then everything is written in the log and as you can see in the logs in `PR-CI-Coverage` the tests go through these parts of the code. I still don't know why only this file has line coverage equal to 0.

lidanqing-vv (approved):
LGTM

wozna:
The recent changes are related to cmake files. It turned out that `mkldnn_quantizer.cc` was duplicated. It was added to the `analysis_predictor` library and then it was separately added to the `paddle_inference_shared` library where the `analysis_predictor` library was added as a dependency. That's why I left this file only in `analysis_predictor` lib.

PR #33768, opened 2021-06-24
[Paddle-TRT]Fix flatten converter when batch_size > 1

cryoco:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix flatten converter crash when batch_size > 1


jiweibo (approved):
LGTM

PR #33797, opened 2021-06-28
Add QuantizeFusionLSTM pass and collect lstm scales

lidanqing-vv:
### PR types
New features

### PR changes
Others

### Describe
Add QuantizeFusionLSTM pass and add LSTM scales

### DIFF
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -2249,9 +2249,10 @@ PDNode *patterns::MultipleQuantize::operator()() {
 PDNode *patterns::QuantizePlacement::operator()(
     const std::unordered_set<std::string> &quantize_enabled_op_types) {
   std::unordered_set<std::string> supported_op_types =
-      std::unordered_set<std::string>(
-          {"concat", "conv2d", "elementwise_add", "fc", "matmul", "pool2d",
-           "prior_box", "reshape2", "transpose2", "fusion_gru", "multi_gru"});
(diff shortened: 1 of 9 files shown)


paddle/fluid/framework/ir/graph_pattern_detector.cc
```diff
@@ -2251,7 +2251,8 @@ PDNode *patterns::QuantizePlacement::operator()(
   std::unordered_set<std::string> supported_op_types =
       std::unordered_set<std::string>(
           {"concat", "conv2d", "elementwise_add", "fc", "matmul", "pool2d",
-           "prior_box", "relu", "reshape2", "transpose2", "fusion_gru"});
+           "prior_box", "relu", "reshape2", "transpose2", "fusion_gru",
+           "multi_gru", "fusion_lstm"});
```
wozna: For now, `multi_gru` is supported in QAT, but it is not in PTQ (it'll be added https://github.com/PaddlePaddle/Paddle/pull/33749). As well as, here you add support for `fusion_lstm` for PTQ but not for QAT. I think that we should separate these lists because it is very illegible and I think it might cause bugs.
Maybe in `quant2_int8_mkldnn_pass.py` we should define default quant2 list by setting `quantize_enabled_op_types` to `cpu_quantize_placement_pass`.

lidanqing-vv: Hi 
- now I only add fusion_lstm now, I removed multi_gru
- fusion_lstm here is used for QAT, I m not sure if it is for PTQ. But QAT also need this pattern detection too. I need to detect the pattern and then substitue with fusion_lstm mkldnn kernel.

PR #33798, opened 2021-06-28
Update paddle_build.sh for remove op grad

OliverLPH:
### PR types

Others

### PR changes

Others

### Describe

enable remove op grad with build inference lib


XieYunshen (approved):
LGTM

PR #33819, opened 2021-06-28
[NPU] remove duplicated stream sync in fetch op

zhiqiu:
### PR types

Performance optimization
### PR changes

OPs
### Describe

remove duplicated stream sync in fetch op, since `TensorCopySync` has `stream sync` Inside the function. 

- before 
image

- after
image


pangyoki (approved):
LGTM

PR #33847, opened 2021-06-29
Fix the tril op does not stop gradient

youth123:
### PR types

Others
### PR changes

OPs
### Describe

Fix the tril op does not stop gradient.


sandyhouse (approved):
LGTM

PR #33852, opened 2021-06-30
fix typo in norm.py

parap1uie-s:
### PR types
Bug Fix
### PR changes
OPs

### Describe

`x != None` is not regular rule in python. Use `x is not` None instead.


parap1uie-s:
@TCChenlong kunlun的CI挂了，看日志好像不是我变更的问题？

PR #33872, opened 2021-06-30
[NPU] delete useless GELU in gelu grad npu op

pangyoki:
### PR types

Performance optimization

### PR changes

OPs

### Describe

In the original implementation of `gelu grad` op, `GELU` NPU OP is used to calculate an intermediate variable named `out`. However, we found that this variable was not actually used. In order to improve performance, the useless GELU operation was deleted. 

- unittest result
图片
图片

- timeline of gelu grad op
Before:
图片
After:
图片

- model performance
29491 tokens/s -> 30101 tokens/s, about 2.1% performance improvement.


zhiqiu (approved):
LGTM

phlrain (approved):

Issue #33685, opened 2021-06-21
Add MKLDNN Kernel  for PRelu


juncaipeng:
Download and unzip the demo (Link:https://dubox.com/s/1iudoIXxOAyJoo9Qqaxz_Kg Password:csrp).

Run `sh run.sh`

```
------------------------->     Profiling Report     <-------------------------

Place: CPU
Time unit: ms
Sorted by total time in descending order in the same thread

-------------------------     Overhead Summary      -------------------------

Total time: 20980.3
  Computation time       Total: 18912.5     Ratio: 90.1441%
  Framework overhead     Total: 2067.8      Ratio: 9.8559%

-------------------------     GpuMemCpy Summary     -------------------------

GpuMemcpy                Calls: 0           Total: 0           Ratio: 0%

-------------------------       Event Summary       -------------------------

Event                        Calls       Total       Min.        Max.        Ave.        Ratio.
thread0::prelu               16830       17924.6     0.173186    10.5915     1.06504     0.854354
  ext_reorder                16830       669.975     0.007853    0.922596    0.31244     0.0373774*
thread0::conv2d              25500       2825.22     0.039415    9.08506     0.110793    0.134661
  int_reorder                17390       577.194     0.00928     2.28289     1.4228      0.2043*
thread0::load_combine        1           187.326     187.326     187.326     187.326     0.00892868
thread0::reshape2            510         22.038      0.03293     0.257293    0.0432117   0.00105041
  ext_reorder                510         2.19305     0.003012    0.029513    0.006885    0.0995125*
thread0::scale               510         18.7042     0.029137    0.312968    0.036675    0.000891514
thread0::ext_reorder         510         2.39843     0.003014    0.011713    0.0047028   0.000114318*
```

For the above model, the details of the inference time as follows
![image](https://user-images.githubusercontent.com/52520497/122703877-19302100-d285-11eb-8151-e4785eb3350f.png)

Add MKLDNN Kernel for PRelu to reduce the inference time. Other activatios may also need to add MKLDNN kernels.

Hope to fix this issue before 2.2RC.

PR #33172, opened 2021-05-27
Add lookup_table_v2 BF16 op

wozna:
### PR types

New features
### PR changes

OPs
### Describe

This PR adds BF16 support for FWD and BWD lookup_table_v2 op. 
Added UT that reuses test `test_lookup_table_bf16_op`.

PR #33878, opened 2021-06-30
Added PRelu BF16/FP32 FWD/BWD kernels

jakpiase:
### PR types
New features

### PR changes
OPs

### Describe
Added PRelu BF16/FP32 FWD/BWD kernels

Profiling before this PR:
image
Profiling after this PR:
image

So it made ~88% speedup, measured with Intel(R) Core(TM) i9-9940X CPU @ 3.30GHz on mobilefacenet model
It's also a solution for this [issue](https://github.com/PaddlePaddle/Paddle/issues/33685)

### DIFF
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -2262,11 +2262,26 @@ PDNode *patterns::QuantizePlacement::operator()(
 PDNode *patterns::Bfloat16Placement::operator()(
     const std::unordered_set<std::string> &bfloat16_enabled_op_types) {
   std::unordered_set<std::string> supported_op_types =
-      std::unordered_set<std::string>(
-          {"concat", "conv2d", "conv2d_transpose", "elementwise_add",
-           "elementwise_mul", "fc", "fusion_gru", "fusion_lstm", "gelu",
-           "layer_norm", "matmul", "matmul_v2", "pool2d", "prelu", "relu",
-           "reshape2", "softmax", "split", "sum", "transpose2"});
+      std::unordered_set<std::string>({"concat",
+                                       "conv2d",
+                                       "conv2d_transpose",
+                                       "elementwise_add",
+                                       "elementwise_mul",
+                                       "fc",
+                                       "fusion_gru",
+                                       "fusion_lstm",
+                                       "gelu",
+                                       "layer_norm",
+                                       "matmul",
+                                       "matmul_v2",
+                                       "pool2d",
+                                       "prelu",
+                                       "relu",
+                                       "reshape2",
+                                       "softmax",
+                                       "split",
+                                       "sum",
+                                       "transpose2"});
   if (!bfloat16_enabled_op_types.empty()) {
     supported_op_types = bfloat16_enabled_op_types;
   }
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -2265,8 +2265,8 @@ PDNode *patterns::Bfloat16Placement::operator()(
       std::unordered_set<std::string>(
           {"concat", "conv2d", "conv2d_transpose", "elementwise_add",
            "elementwise_mul", "fc", "fusion_gru", "fusion_lstm", "gelu",
-           "layer_norm", "matmul", "matmul_v2", "pool2d", "relu", "reshape2",
-           "softmax", "split", "sum", "transpose2"});
+           "layer_norm", "matmul", "matmul_v2", "pool2d", "prelu", "relu",
+           "reshape2", "softmax", "split", "sum", "transpose2"});
   if (!bfloat16_enabled_op_types.empty()) {
     supported_op_types = bfloat16_enabled_op_types;
   }
--- a/paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
@@ -24,6 +24,7 @@ using platform::MKLDNNDeviceContext;
 using platform::MKLDNNGetDataType;
 using platform::to_void_cast;
 
+namespace {
 template <typename T>
 class PReluMKLDNNHandler
     : public platform::MKLDNNHandlerT<T, dnnl::prelu_forward,
@@ -70,6 +71,7 @@ class PReluMKLDNNHandler
                                             output_data, "@diff_weights_mem_p");
   }
 };
+}  // anonymous namespace
 
 template <typename T>
 class PReluMKLDNNKernel : public framework::OpKernel<T> {
--- a/paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
@@ -32,7 +32,7 @@ class PReluMKLDNNHandler
   PReluMKLDNNHandler(const MKLDNNDeviceContext& dev_ctx,
                      const mkldnn::engine engine, platform::Place cpu_place,
                      const Tensor* x, const Tensor* weights,
-                     const std::string& uniq_name)
+                     const std::string& uniq_name, bool is_test)
       : platform::MKLDNNHandlerT<T, dnnl::prelu_forward, dnnl::prelu_backward>(
             dev_ctx, engine, cpu_place,
             platform::CreateKey(dev_ctx, framework::vectorize(x->dims()),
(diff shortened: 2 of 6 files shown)


lidanqing-vv:
@jakpiase  Please remove the `check_dygraph=False`, because it will be set to False by default since following function in op_test.py

image

You can also refer to this PR https://github.com/PaddlePaddle/Paddle/pull/33172

jakpiase:
Thank you, I'll delete that. I am glad to know that information

jakpiase:
Unfortunately disabling check_dygraph that you have mentioned is only in check_output_with_place. The code doesn't work since in check_grad_with_place check_dygraph is not disabled.

jakpiase:
Since tests will not pass without that line, I will add it back
https://xly.bce.baidu.com/paddlepaddle/paddle/newipipe/detail/3077208/job/5391455

lidanqing-vv:
> @jakpiase Please remove the `check_dygraph=False`, because it will be set to False by default since following function in op_test.py
> 
> image
> 
> You can also refer to this PR #33172

@jakpiase  What about add this pattern into check_grad_with_place ?

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
```diff
@@ -0,0 +1,161 @@
...
+    if (!this->isCached()) {
+      auto x_md = memory::desc(framework::vectorize(x->dims()),
+                               MKLDNNGetDataType<T>(), x->format());
+      auto weights_md =
+          memory::desc(framework::vectorize(weights->dims()),
+                       MKLDNNGetDataType<T>(), memory::format_tag::any);
+
+      this->AcquireForwardPrimitiveDescriptor(dnnl::prop_kind::forward_training,
+                                              x_md, weights_md);
+      this->AcquireBackwardPrimitiveDescriptor(x_md, weights_md, x_md,
```
jczaja: If you could read is_test value here and based on it skip creation of Backward PD that would be good (to avoid unneccessery allocation during inference)

jakpiase: Nice idea

jczaja (dismissed):
LGTM

paddle/fluid/framework/ir/graph_pattern_detector.cc
```diff
@@ -2262,11 +2262,26 @@ PDNode *patterns::QuantizePlacement::operator()(
...
+                                       "matmul",
+                                       "matmul_v2",
+                                       "pool2d",
+                                       "prelu",
+                                       "relu",
+                                       "reshape2",
+                                       "softmax",
+                                       "split",
+                                       "sum",
+                                       "transpose2"});
```
arogowie-intel, line 2284: I'm very glad to see such formatting :)

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
```diff
@@ -0,0 +1,162 @@
...
+      : platform::MKLDNNHandlerT<T, dnnl::prelu_forward, dnnl::prelu_backward>(
+            dev_ctx, engine, cpu_place,
+            platform::CreateKey(dev_ctx, framework::vectorize(x->dims()),
+                                uniq_name)) {
+    if (!this->isCached()) {
+      auto x_md = memory::desc(framework::vectorize(x->dims()),
+                               MKLDNNGetDataType<T>(), x->format());
+      auto weights_md =
+          memory::desc(framework::vectorize(weights->dims()),
+                       MKLDNNGetDataType<T>(), memory::format_tag::any);
```
arogowie-intel:
```suggestion
                       MKLDNNGetDataType<T>(), format_tag::any);
```

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
```diff
@@ -0,0 +1,162 @@
...
+        this->AcquireBackwardPrimitiveDescriptor(x_md, weights_md, x_md,
+                                                 weights_md);
+    }
+  }
+
+  std::shared_ptr<memory> AcquireWeightsMemoryWithReorder(const Tensor* input,
+                                                          const bool is_test) {
+    const T* input_data = input->data<T>();
+    auto user_weights_md =
+        memory::desc(framework::vectorize(input->dims()),
```
arogowie-intel, line 81:
```suggestion
        desc(framework::vectorize(input->dims()),
```

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
```diff
@@ -0,0 +1,162 @@
...
+};
+
+template <typename T>
+class PReluMKLDNNKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& ctx) const override {
+    this->RunKernel(ctx);
+  }
+
+  void RunKernel(const framework::ExecutionContext& ctx) const {
```
arogowie-intel, line 104: Why are you calling `RunKernel` instead of putting `RunKernel`'s body into `Compute` function? What's the advantage? Do you reuse this function somewhere else?

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
```diff
@@ -0,0 +1,162 @@
...
+
+using dnnl::memory;
+using framework::Tensor;
+using platform::GetMKLDNNFormat;
+using platform::MKLDNNDeviceContext;
+using platform::MKLDNNGetDataType;
+using platform::to_void_cast;
+
+template <typename T>
+class PReluMKLDNNHandler
```
arogowie-intel, line 29: You can put this class into anonymous namespace or make it static to get internal linkage and thus not clutter symbol table.

python/paddle/fluid/tests/unittests/mkldnn/test_prelu_mkldnn_op.py
```diff
@@ -0,0 +1,149 @@
...
+    def test_check_grad(self):
+        self.check_grad(['X', 'Alpha'], 'Out')
+
+
+class TestPReluModeAllOneDNNOp(TestPReluModeChannelOneDNNOp):
+    def init_attrs(self):
+        self.mode = "all"
+        self.alpha = np.random.random((1, 1, 1, 1)).astype("float32")
+
+    # Skip 'Alpha' input check because in mode = 'all' it has to be a single 1D value, so checking if it has at least 100 values will cause an error
```
arogowie-intel: Please wrap this long error line.

python/paddle/fluid/tests/unittests/mkldnn/test_prelu_mkldnn_op.py
```diff
@@ -0,0 +1,149 @@
...
+            self.dalpha = self.alpha.copy()
+
+            if self.mode == "all":
+                self.dx = np.where(self.x > 0, dout, dout * self.alpha[0])
+            elif self.mode == "channel":
+                for i in range(self.x.shape[1]):
+                    self.dx[:, i] = np.where(self.x[:, i] > 0, dout[:, i],
+                                             dout[:, i] * self.alpha[0, i])
+            elif self.mode == "element":
+                self.dx = np.where(self.x[:] > 0, dout[:], dout[:] * self.alpha)
```
arogowie-intel: Isn't it conceptually the same code as for `ref_prelu` ?

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
jakpiase: That's some kind of semantics which I use in every operator. It will be easier for future to add functionality like "force_fp32_output" when we have two different functions for that or if OneDNN will have more support for mixed-precision, but if you want I can merge these two functions

python/paddle/fluid/tests/unittests/mkldnn/test_prelu_mkldnn_op.py
jakpiase: It's similiar, but arguments of numpy functions are different. I think that merging them together won't result in clear code

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
jakpiase: Ok

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
jakpiase: I cannot do that, desc() is not visible without memory namespace

paddle/fluid/operators/mkldnn/prelu_mkldnn_op.cc
jakpiase: I cannot do that, format_tag is not visible without memory namespace

lidanqing-vv (dismissed):
LGTM.

lidanqing-vv (approved):
LGTM

juncaipeng (approved):
lgtm

PR #33046, opened 2021-05-21
Add AXPY oneDNN handler

arogowie-intel:
### PR types
Performance optimization

### PR changes
OPs

### Describe
This PR introduces oneDNN AXPY handler providing BLAS::AXPY interface. The use of this handler is enabled at this moment only for `bfloat16` data type for `sum` operator with `SelectedRows` input tensors.  It gives almost 10% performance improvement on CPX machine.

Additionally this PR speeds up SGD operator with Selected rows with AXPY handler.

PR #33879, opened 2021-06-30
Use CBLAS for SelectedRows elementwise add operation.

arogowie-intel:
### PR types
Performance optimization

### PR changes
Others

### Describe
This PR gets back CBLAS library usage for `float` data type for `SelectedRows` tensors for elementwise add operation as it is faster.

### DIFF
--- a/paddle/fluid/operators/math/selected_rows_functor.cc
+++ b/paddle/fluid/operators/math/selected_rows_functor.cc
@@ -302,15 +302,15 @@ namespace scatter {
 
 #ifdef PADDLE_WITH_MKLDNN
 template <typename T>
-typename std::enable_if<std::is_same<T, float>::value ||
-                        std::is_same<T, platform::bfloat16>::value>::type
+typename std::enable_if<std::is_same<T, platform::bfloat16>::value>::type
 elementwise_add_to(BlasT<platform::CPUDeviceContext, T>* blas, size_t data_len,
                    const T* in, T* out) {
   onednn_handler_axpy(data_len, T(1.f), in, out);
 }
 
 template <typename T>
-typename std::enable_if<std::is_same<T, double>::value ||
+typename std::enable_if<std::is_same<T, float>::value ||
+                        std::is_same<T, double>::value ||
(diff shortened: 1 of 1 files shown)


arlesniak (approved):
LGTM

wozna (approved):
LGTM.  
Could you please add to the PR description, what is the approximate percentage difference in performance? This could be useful information in the future.

arogowie-intel:
@wozna 
This is rather the _revert_ to the state that was earlier. I just mistakenly switched float implementation from CBLAS to ours. However the difference is significant around 25%.

lidanqing-vv (approved):
LGTM

paddle/fluid/operators/math/selected_rows_functor.cc
```diff
@@ -302,15 +302,15 @@ namespace scatter {
...
   onednn_handler_axpy(data_len, T(1.f), in, out);
 }
 
 template <typename T>
-typename std::enable_if<std::is_same<T, double>::value ||
+typename std::enable_if<std::is_same<T, float>::value ||
+                        std::is_same<T, double>::value ||
                         std::is_same<T, platform::complex<float>>::value ||
                         std::is_same<T, platform::complex<double>>::value>::type
 elementwise_add_to(BlasT<platform::CPUDeviceContext, T>* blas, size_t data_len,
```
lidanqing-vv, line 316: 1. It seems using following is enough
```
#ifdef PADDLE_WITH_MKLDNN
template <typename T>
typename std::enable_if<std::is_same<T, platform::bfloat16>::value>::type
elementwise_add_to(BlasT<platform::CPUDeviceContext, T>* blas, size_t data_len,
                   const T* in, T* out) {
  onednn_handler_axpy(data_len, T(1.f), in, out);
}
#endif
```

2. Otherwise, among these 3 functions, second function looks copy of third function https://github.com/PaddlePaddle/Paddle/pull/33046/files#, but in third function `is_floating_point<T>` for `<long double>` is also true, but in second function only float and double are included maybe not enough.

PR #33895, opened 2021-07-01
optimize for elastic 

kuizhiqing:
### PR types

Others

### PR changes

Others

### Describe

* add more random for node tag
* prevent deadlock in extreme condition


danleifeng (approved):
LGTM

PR #33929, opened 2021-07-02
fix bug of p2p in PY36 ce

ForFishes:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix bug of p2p in PY36 ce


DDDivano (approved):
LGTM

PR #33937, opened 2021-07-02
fix decoding error when clip grad op and python==2

jiweibo:
### PR types

Others
### PR changes

Others
### Describe

https://github.com/PaddlePaddle/Paddle/pull/33919

兼容python2


OliverLPH (approved):
LGTM

Issue #33918, opened 2021-07-01
MKLDNN error for ppyolov2 model


xiegegege:
Paddle Test Environment
****************************************
Paddle version: 0.0.0
Paddle With CUDA: True

OS: Ubuntu 16.04
Python version: 3.7.0

CUDA version: 10.1.243
cuDNN version: 7.6
Nvidia driver version: None
CPU: Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz
****************************************

install paddle develop (commit id:1e62c239d323354eccfc974d4e2e6496f93d848e)
```shell
git clone https://github.com/PaddlePaddle/PaddleDetection.git 
cd PaddleDetection
git checkout develop

python -m pip install -r requirments.txt
python tools/export_model.py -c configs/ppyolo/ppyolov2_r50vd_dcn_365e_coco.yml --output_dir=inference_model -o weights=https://paddledet.bj.bcebos.com/models/ppyolov2_r50vd_dcn_365e_coco.pdparams
python deploy/python/infer.py --model_dir=./inference_model/ppyolov2_r50vd_dcn_365e_coco/ --image_file=demo/000000570688.jpg --device=CPU --enable_mkldnn=True
```

<img width="877" alt="767953fbf406aa18918dda343fd810c0" src="https://user-images.githubusercontent.com/46314656/124121722-d0a80d00-daa7-11eb-8530-e0badc8c1152.png">
I have tested this case and it works well in paddle2.1.0 and paddle2.1.1， so this should be  the new problem caused by paddle develop.
The model ttfnet_darknet53_1x_coco has the same problem.

[ppyolov2_cpu_mkldnnn_log.txt](https://github.com/PaddlePaddle/Paddle/files/6748005/ppyolov2_cpu_mkldnnn_log.txt)

lidanqing-vv:
1. 6148 is capable with AVX512, You had this issue because you used some virtual machine ? @xiegegege

xiegegege:
@lidanqing-intel Yes, it's a virtual machine. I tested in some other physical machine, and it works well.

jakpiase:
Hi @xiegegege
#33944 will fix your issue. Your virtual machine probably uses AVX2 and we recommend switching to AVX512, because our performance improvements with oneDNN library are made mainly for that instruction set.

xiegegege:
@jakpiase OK, got it, thank you.

lidanqing-vv:
@xiegegege 
[...]

PR #33944, opened 2021-07-02
Split op oneDNN AVX2 fix

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fixed bug that was occuring when AVX2 instruction set was used. In that scenario blocking format was chosen and split op doesn't support that. It is a fix for [#33918](https://github.com/PaddlePaddle/Paddle/issues/33918)

### DIFF
--- a/paddle/fluid/operators/split_op.cc
+++ b/paddle/fluid/operators/split_op.cc
@@ -14,6 +14,7 @@ limitations under the License. */
 
 #include "paddle/fluid/operators/split_op.h"
 #include <string>
+#include "paddle/fluid/platform/cpu_info.h"
 
 namespace paddle {
 namespace operators {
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/split_op.cc
```diff
@@ -78,9 +78,18 @@ class SplitOp : public framework::OperatorWithKernel {
...
-                                     framework::DataLayout::kMKLDNN,
-                                     framework::LibraryType::kMKLDNN);
+      // OneDNN uses blocking format, which cannot be always
+      // supported with reorders, because if blocked dimension is not divisible
+      // by
+      // 8 or 16(depending on which blocking format is used) submemory cannot be
+      // created, so in that scenario a fallback is needed
+      auto tmp_md = dnnl::memory::desc(
+          framework::vectorize(ctx.Input<Tensor>("X")->dims()),
+          dnnl::memory::data_type::f32, ctx.Input<Tensor>("X")->format());
```
arlesniak, line 88: Will it always be `dnnl::memory::data_type::f32` when checking inner_nblks ?

jakpiase: This datatype doesn't matter. I only need inner_nblks and it will be the same for any datatype. I have asked oneDNN team if I can check if a format_tag is blocking in any simpler way, but they said that there is no easy way to do that, so I am constructing temporary memory descriptor to get inner_nblks

arlesniak (approved):
LGTM

paddle/fluid/operators/split_op.cc
```diff
@@ -78,9 +78,18 @@ class SplitOp : public framework::OperatorWithKernel {
 
 #ifdef PADDLE_WITH_MKLDNN
     if (this->CanMKLDNNBeUsed(ctx, input_data_type)) {
-      return framework::OpKernelType(input_data_type, ctx.GetPlace(),
-                                     framework::DataLayout::kMKLDNN,
-                                     framework::LibraryType::kMKLDNN);
+      // OneDNN uses blocking format, which cannot be always
```
lidanqing-vv, line 81: As you described in teams, oneDNN split only work well when avx512 instructions available, in this way why not check if avx512 is supported in the condition together with this->CanMKLDNNBeused ?

lidanqing-vv: Currently under avx512 instructions, what ops are using block format ? If under avx512 there are still ops using block format, maybe this way is good.

jakpiase: On avx512 blocking format may be chosen later, that way it will still be working

lidanqing-vv (approved):
LGTM

lidanqing-vv:
Without oneDNN split op, #33918 will be slower than naive version? Did you test #33918 with your PR ?

lidanqing-vv:
#33944 will solve this error on AVX2. But testing this model on AVX2 environment may not speedup as good as on AVX512 (6148 itself suppots AVX512). Because oneDNN improve very well on AVX512. But in AVX2, some oneDNN op will call reference version (slower).

So to achieve expected performance, please use avx512. But this fix will solve error on AVX2.

@juncaipeng  Please approve and merge this PR

juncaipeng (approved):
LGTM

PR #33957, opened 2021-07-05
[NPU] change Add to AddN in sum npu op

pangyoki:
### PR types

Performance optimization

### PR changes

OPs

### Describe

use `AddN` instead of `Add`.

- before (use Add)

图片

As shown in the timeline, the time consumption of `ClipGrad` is 144ms (439ms - 583ms), and the time consumption of `Add` in `ClipGrad` is 23ms (521ms - 544ms), which accounts for 16.0%. 

- after (use AddN)
图片

As shown in the timeline, the time consumption of `ClipGrad` is 138ms (471ms - 609ms), and the time consumption of `AddN` in `ClipGrad` is 3ms (562ms - 565ms), which accounts for 2.2%.

### DIFF
--- a/paddle/fluid/operators/sum_op_npu.cc
+++ b/paddle/fluid/operators/sum_op_npu.cc
@@ -35,16 +35,10 @@ class SumNPUKernel : public framework::OpKernel<T> {
     auto place = ctx.GetPlace();
 
     int n = static_cast<int>(x.size());
-    PADDLE_ENFORCE_EQ(n > 0, true,
-                      platform::errors::InvalidArgument(
-                          "The size of Input(x) list must be larger than 0."));
     if (n == 1) {
       TensorCopy(*x[0], place, out);
       return;
     }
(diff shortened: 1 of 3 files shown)


paddle/fluid/operators/sum_op_npu.cc
```diff
@@ -35,23 +35,28 @@ class SumNPUKernel : public framework::OpKernel<T> {
...
+      }
+    }
+
     auto stream =
         ctx.template device_context<paddle::platform::NPUDeviceContext>()
             .stream();
-
-    const auto& runner = NpuOpRunner("Add", {*x[0], *x[1]}, {*out}, {});
-
+    NpuOpRunner runner{"AddN", {inputs}, {*out}, {{"N", n}}};
```
zhiqiu, line 57: keep const auto& runner

zhiqiu (approved):
LGTM

PR #33959, opened 2021-07-05
make stop_gradient=True for random op

zhiqiu:
### PR types

Function optimization
### PR changes

APIs
### Describe

make stop_gradient=True for random op in static graph

image


pangyoki (approved):
LGTM

PR #33970, opened 2021-07-05
[OP] fix histogram op when input tensor is empty, test=develop

qili93:
### PR types

Bug fixes

### PR changes

OPs

### Describe

fix histogram op when input tensor is empty, keep the same behavior with torch.


zhangting2020 (approved):
LGTM

PR #33977, opened 2021-07-05
Added OpTestTool for BF16

arlesniak:
### PR types

Others
### PR changes

Others
### Describe

Added OpTestTool for BF16 convenience

### DIFF
--- a/python/paddle/fluid/tests/unittests/op_test.py
+++ b/python/paddle/fluid/tests/unittests/op_test.py
@@ -1780,12 +1780,12 @@ class OpTest(unittest.TestCase):
 
 class OpTestTool:
     @classmethod
-    def skipIf(cls, condition: object, reason: str):
+    def skip_if(cls, condition: object, reason: str):
         return unittest.skipIf(condition, reason)
 
     @classmethod
-    def skipIfNotBF16(cls):
-        return OpTestTool.skipIf(
+    def skip_if_not_bf16(cls):
+        return OpTestTool.skip_if(
             not (isinstance(_current_expected_place(), core.CPUPlace) and
                  core.supports_bfloat16()),
             "Place does not support BF16 evaluation")
--- a/python/paddle/fluid/tests/unittests/op_test.py
+++ b/python/paddle/fluid/tests/unittests/op_test.py
@@ -1784,7 +1784,7 @@ class OpTestTool:
         return unittest.skipIf(condition, reason)
 
     @classmethod
-    def skip_if_not_bf16(cls):
+    def skip_if_not_cpu_bf16(cls):
         return OpTestTool.skip_if(
             not (isinstance(_current_expected_place(), core.CPUPlace) and
                  core.supports_bfloat16()),
--- a/python/paddle/fluid/tests/unittests/op_test.py
+++ b/python/paddle/fluid/tests/unittests/op_test.py
@@ -32,7 +32,7 @@ import paddle.fluid.core as core
 from paddle.fluid.backward import append_backward
 from paddle.fluid.op import Operator
 from paddle.fluid.executor import Executor
-from paddle.fluid.framework import Program, OpProtoHolder, Variable
+from paddle.fluid.framework import Program, OpProtoHolder, Variable, _current_expected_place
 from paddle.fluid.tests.unittests.testsuite import (
     create_op,
     set_input,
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -1776,3 +1776,16 @@ def _get_gradient(self,
...
+
+class OpTestTool:
+    @classmethod
+    def skipIf(cls, condition: object, reason: str):
+        return unittest.skipIf(condition, reason)
+
+    @classmethod
+    def skipIfNotBF16(cls):
+        return OpTestTool.skipIf(
+            not (isinstance(_current_expected_place(), core.CPUPlace) and
```
jakpiase, line 1789: Shouldn't it be like: skipIf(not CPU or not SupportsBF16)?

arlesniak: We want the test to be run with both CPU and BF16 support, otherwise (logical not) skip.

jakpiase: My bad, we meant the same thing but I didn't see one parenthesis

jakpiase (dismissed):
LGTM

python/paddle/fluid/tests/unittests/op_test.py
lidanqing-vv: Hi, I have two questions:
1. It looks 
```
    @classmethod
    def skipIf(cls, condition: object, reason: str):
        return unittest.skipIf(condition, reason)
```
is not really needed, why don't you call this unittest.skipIf in skipIfNotBF16 ?
2. Why do you create a new class `OpTestTool`, why don't you just put this class decorator function `def skipIfNotBF16(cls):` in OpTest ?

lidanqing-vv (commented):
Hi, I asked questions above

python/paddle/fluid/tests/unittests/op_test.py
arlesniak: `OpTestTool` was intended to be namespace for convenient toolbox. It contains classmethods that can be used as decorators which gives flexibility we need.

lidanqing-intel: Got explanation that keeping following in OpTestTool because we make all possible skip in one class, it will be easy to use and maintain.
```
@classmethod
    def skipIf(cls, condition: object, reason: str):
```
Thanks !

lidanqing-vv (dismissed):
LGTM. Thank you very much !

More info: Artur made this OpTestTool because he thinks that there maybe other decorator functions could be added, not just skip if functions. Currently we use inheritance for UTs, but for inheritance we have to watch out carefully parent's functions, decorator is more portable and flexible.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -1776,3 +1776,16 @@ def _get_gradient(self,
                              fetch_list,
                              scope=scope,
                              return_numpy=False)))
+
+
+class OpTestTool:
+    @classmethod
+    def skipIf(cls, condition: object, reason: str):
```
wzzju: Please use the `snake case` name instead of the `camel case` name.

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -1776,3 +1776,16 @@ def _get_gradient(self,
...
+        return unittest.skipIf(condition, reason)
+
+    @classmethod
+    def skipIfNotBF16(cls):
```
wzzju: Here, you just deal with CPU, please reflect this in the naming.

Please use the `snake case` name instead of the `camel case` name.

arlesniak: OK

python/paddle/fluid/tests/unittests/op_test.py
arlesniak: OK

wzzju (approved):
LGTM

arlesniak:
PR-CI-Windows shows the following error, not corelated with the PR code:
`fatal error C1083: Cannot open include file: 'google/protobuf/stubs/scoped_ptr.h': No such file or directory`

arlesniak:
Looks like PR-CI-ROCM-Compile is blocked.

lidanqing-vv:
This CI machine shut down in weekend so this CI is rerun.  @arlesniak Please do not rerun this CI anymore, they already move this task ahead now.

PR #33984, opened 2021-07-06
exclude lr scheduler's state from accumulators_holder

iclementine:
### PR types

Bug fixes

### PR changes

APIs

### Describe

Fix the bug that "an optimizer that has a learning rate scheduler who has called `set_state_dict()` before any update would cause an error in `step()` afterwards."

### DIFF
--- a/python/paddle/optimizer/optimizer.py
+++ b/python/paddle/optimizer/optimizer.py
@@ -286,7 +286,10 @@ class Optimizer(object):
         if isinstance(self._learning_rate, LRScheduler):
             self._learning_rate.set_state_dict(state_dict["LR_Scheduler"])
 
-        state_dict.pop("LR_Scheduler")
+        # NOTE: exclude learning rate scheduler's state from 
+        # _accumulators_holder.
+        if "LR_Scheduler" in state_dict:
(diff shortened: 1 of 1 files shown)


python/paddle/optimizer/optimizer.py
```diff
@@ -286,6 +286,10 @@ def set_state_dict(self, state_dict):
         if isinstance(self._learning_rate, LRScheduler):
             self._learning_rate.set_state_dict(state_dict["LR_Scheduler"])
 
+        # NOTE: exclude learning rate scheduler's state from 
+        # _accumulators_holder.
+        if "LR_Scheduler" in state_dict:
+            state_dict.pop("LR_Scheduler")
```
guoshengCS, line 293: 这里是否最好是重新创建一个state_dict以避免对输入的修改呢

iclementine: 好的

iclementine: 已经修改

guoshengCS (approved):
LGTM

chenwhql (approved):
LGTM

PR #33992, opened 2021-07-06
[NPU] NpuOpRunner supports host tensor as input

zhiqiu:
### PR types

New features
### PR changes

OPs
### Describe

NpuOpRunner supports host tensor as input

CANN-5.0.2 support input tensors on host. Specifically, the tensor of shape, tensor of dims, etc, which are are small vector/list. 
For example, the input `axis` in `GatherV2` is a tensor  of shape [1]. It should be const tensor, otherwise, the op cannot be compiled due to dynamic shape. 
Now, we can use `aclSetTensorPlaceMent` to declare `axis` as host tensor to solve the problem.

image


pangyoki (approved):
LGTM

PR #33996, opened 2021-07-07
fix the bug in pipeline that the variable of send is released by gc before send is completed.

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix the bug in pipeline that the variable of send is released by gc before send is completed. The reason is that the variable for the corresponding c_sync_comm_stream is modified.


wangxicoding (approved):
LGTM

PR #34021, opened 2021-07-07
Added printing tensor's format if oneDNN is used

jakpiase:
### PR types
Others
### PR changes
Others
### Describe
Added printing tensor's format if oneDNN is used. That functionality will be helpful when debugging. All oneDNN format tags are supported for printing.
Example:
image


paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -533,5 +533,709 @@ inline bool HasOpFLOAT32DataType(const paddle::framework::OpDesc* op) {
...
+    case MKLDNNMemoryFormat::gOIw4o8i8o4i:
+      return "gOIw4o8i8o4i";
+    case MKLDNNMemoryFormat::gOIdhw4o8i8o4i:
+      return "gOIdhw4o8i8o4i";
+    case MKLDNNMemoryFormat::gOdhwI16o2i:
+      return "gOdhwI16o2i";
+    case MKLDNNMemoryFormat::OhwI16o4i:
+      return "OhwI16o4i";
+  }
+  return "unknown";
```
jakpiase: This option is necessary because if oneDNN would add a new format tag compilation would result in a failure

arlesniak (dismissed):
LGTM

lidanqing-vv (commented):
1. Could you try to << directly the format variable, because if we could see in DNNL_VERBOSE log the format name, it means printing out format function should be done already?
2. If the format could not be printed out directly, could you please try to use macro
```
#include<stdio.h>
#include<iostream>
using namespace std;

enum class format_tag {
      undef,
      any,
      a,
      ab,
      ba
  };

  #define GET_VARIABLE_NAME(Variable) (#Variable)

  int main(){
      std::cout<<GET_VARIABLE_NAME(format_tag::undef)<<std::endl;
      return 0;
  }
```

jakpiase:
You were right @lidanqing-intel, I have used a function that I have found in oneDNN source code

lidanqing-vv (approved):
LGTM. Thank you very much!

PR #34026, opened 2021-07-08
Add paddle.inv API

haolin-nju:
### PR types

New features

### PR changes

APIs

### Describe

Add paddle.inv API, which reuses paddle.inverse api.


haolin-nju:
Someone has implemented paddle.linalg.inv api.

PR #34032, opened 2021-07-08
[NPU] Fix vector overflow in slice grad npu op

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

[NPU] Fix vector overflow in slice grad npu op


pangyoki (approved):
LGTM

PR #34040, opened 2021-07-09
[Paddle-TRT] Fix output data type selection

zlsh80826:
### PR types

Bug fixes

### PR changes

OPs

### Describe

The output type in plugins should be same as input type instead of `data_type_`.
Or it will make trt fix the network output to `data_type_`.


jiweibo (approved):
LGTM

PR #34052, opened 2021-07-09
refine varbase init function

zhiqiu:
### PR types

Function optimization
### PR changes

Others
### Describe

- Remove checks on kwargs in `core.VarBase()`. When use kwargs, it means the arguments is not required, but we add check on `value`, it is not needed.
 before
image
after
image

- Code refinement, split common function and reuse it.


sneaxiy (approved):
LGTM.

PR #34091, opened 2021-07-12
fix no value for parameter

Jiangxinz:
### PR types
Others

### PR changes
Others

### Describe
In FIle: python/paddle/tests/test_model.py, Line 129
Error: No value for argument 'reason' in unbound method call
Fix: **cls.skipTest(...** --> **cls().skipTest(...**


lelelelelez (approved):
LGTM

PR #34121, opened 2021-07-13
Add Python Helper Function to Get Block Inputs/Outputs

zhhsplendid:
### PR types

New features
### PR changes

APIs
### Describe

As the title, add Python helper function to get block Inputs/Outputs. This function is helpful to distributed computing for sharding.

### DIFF
--- a/python/paddle/fluid/layers/utils.py
+++ b/python/paddle/fluid/layers/utils.py
@@ -17,7 +17,7 @@ import collections
 import copy
 import six
 import numpy as np
-from ..framework import Variable, in_dygraph_mode
+from ..framework import Block, Variable, in_dygraph_mode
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/layers/utils.py
```diff
@@ -429,3 +429,31 @@ def try_get_constant_shape_from_tensor(shape_tensor):
             return None
 
         return None
+
+
+def get_inputs_outputs_in_block(block, program):
```
thisjiang: The parameter `program` not use?

zhhsplendid: Thanks, deleted

Aurelius84 (approved):
LGTM

PR #34134, opened 2021-07-13
Elastic fix

kuizhiqing:
### PR types

Others

### PR changes

Others
### Describe

some fix under extreme condition
* kill process by group, more robust in multiprocess scenario
* add cmd entry to manipulate etcd, just set np for now
* add service discovery for k8s env


danleifeng (approved):
LGTM

PR #34136, opened 2021-07-13
[Dy2Stat] Ignore to convert paddle.no_grad

Aurelius84:
### PR types

 Bug fixes
### PR changes

Others
### Describe

In https://github.com/PaddlePaddle/Paddle/pull/33110 , we support to convert `paddle.grad` into `paddle.static.gradients`. But it also convert `paddle.no_grad` by mistake. So in this PR, we fix it


zhhsplendid (approved):
LGTM

PR #34140, opened 2021-07-13
fix DataLoader memory leak

heavengate:
### PR types
Bug fixes

### PR changes
APIs

### Describe
**fix DataLoader memory leak introduced by https://github.com/PaddlePaddle/Paddle/pull/32550**
- register `__del__` to `atexit` may remain a link to DataLoader object, which make DataLoader cannot be GCed when epoch finish
- simply add `__del__` to release resource can also work to fix DataLoader exit error,  for `__del__` will be called when DataLoader object GCed, so remove `atexit` register here


chenwhql (approved):
LGTM

PR #34157, opened 2021-07-14
[Paddle-TRT] Add noexcept on methods inherited from TensorRT

zlsh80826:
### PR types

Breaking changes

### PR changes

Others

### Describe

TensorRT8 requires all class methods inherited from TensorRT should add noexcept keyword.
Paddle-TRT can support TensorRT8 after this PR.

[Reference](https://docs.nvidia.com/deeplearning/tensorrt/release-notes/tensorrt-8.html)
> TensorRT now declares API’s with the noexcept keyword to clarify that exceptions must not cross the library boundary. All TensorRT classes that an application inherits from (such as IGpuAllocator, IPluginV2, etc…) must guarantee that methods called by TensorRT do not throw uncaught exceptions, or the behavior is undefined.


shangzhizhou (approved):
LGTM

raindrops2sea (approved):

PR #34192, opened 2021-07-15
Added matmul_v2 BF16/FP32 BWD kernel

jakpiase:
### PR types
New features

### PR changes
OPs

### Describe
Added matmul_v2 grad kernel for enabling BF16 word2vec training.


jczaja (dismissed):
LGTM

paddle/fluid/operators/matmul_v2_op.cc
```diff
@@ -62,10 +62,15 @@ class MatMulV2Op : public framework::OperatorWithKernel {
...
+    if (ndims_x > ndims_y) {
       new_dims.assign(dims_x.begin(), dims_x.end() - 2);
-    } else {
+    } else if (ndims_x < ndims_y) {
       new_dims.assign(dims_y.begin(), dims_y.end() - 2);
+    } else {
+      new_dims.reserve(ndims_x);
+      for (size_t i = 0; i < ndims_x - 2; ++i) {
+        new_dims.push_back(std::max(dims_x[i], dims_y[i]));
+      }
```
arogowie-intel, line 73: Are you sure that this is the correct behavior? This significantly change the ops behavior by adding broadcasting. Moreover all already written software already relies on fact that the `dims_x` is taken when X input tensor has bigger rank.  
Besides above note you should probably check whether the processed axes are broadcastable according to NumPy rules.

jakpiase: I am sure about that.
PaddlePaddle matmul_v2 native kernel does that in Compute() method instead of in InferShape()(I have no idea why) and this incorrect behavior of InferShape resulted in BF16 tests failure.
image
They also have test for broadcasting both X and Y:
image

arogowie-intel (dismissed):

arogowie-intel:
My comments regarding the exceeded library binary size:
* When you move sth to the header file please make sure that you expose there **only** symbols which are used somewhere else.
* Please split [matmul_mkldnn_op.h](https://github.com/PaddlePaddle/Paddle/blob/17fe96b72a6a6814856d3db1ac2cebc844f050d0/paddle/fluid/operators/mkldnn/matmul_mkldnn_op.h) header file into header with **just** necessary classes and functions declarations and the source file (`matmul_mkldnn_op.cc`) with respective classes and functions implementations. 
* Hide everything that is implementation detail (ie. not a part of a official public API) inside private/protected part for class members and under anonymous namespaces in source files. 
* Since kernel classes are rather implementation detail and used only in kernel registration hide them in unnamed namespaces.
* Regarding [matmul_v2_mkldnn_op.cc](https://github.com/PaddlePaddle/Paddle/blob/2104d0d7f7897d9c7360cab47f06dbb57af6cb81/paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc) file: since it's already a source file just move out everything to the unnamed namespaces or make it a global static. 
* Kernel registration should be inside source file.
* Put class/function [explicit template instantiation](https://en.cppreference.com/w/cpp/language/class_template) only for the types used in kernel registration.

General rule of thumb. Please strive to do not clutter any paddle namespaces and hide each implementation detail in source files. Remember that each symbol (both class or function) you would place in public namespace will be visible to the end user.

jczaja (dismissed):

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.h
```diff
@@ -0,0 +1,42 @@
+/* Copyright (c) 2020 PaddlePaddle Authors. All Rights Reserved.
```
lidanqing-vv: Maybe change it to 2021

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace paddle {
+namespace operators {
+
+using platform::MKLDNNDeviceContext;
+using framework::ExecutionContext;
+using Tensor = framework::Tensor;
+
+template <typename T>
+class MatMulGradMKLDNNKernel : public framework::OpKernel<T> {
```
lidanqing-vv, line 28: Why `MatMulGradMKLDNNKernel` need a separate header file? and this header file is only used in its cc file?

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_v2_mkldnn_op.py
```diff
@@ -23,14 +24,12 @@
 import paddle.fluid as fluid
 import paddle.fluid.framework as framework
 
-paddle.enable_static()
 
-
-def reference_matmul(X, Y, transpose_X=False, transpose_Y=False):
+def reference_matmul(X, Y, trans_x=False, trans_y=False):
```
lidanqing-vv: Why do you change transpose_x to trans_x ? I feel keeping code self explaining is good. With trans_x, if somebody remove below comments,  may get confused with trans_x. 

Just my thought.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.h
jakpiase: I'll do that

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.h
jakpiase: Because there was a problem with increased binary size, so I have implemented Adam's advices and left only reused symbols in ".h" file. Other classes are in anonymous namespaces to take advantage of internal linkage

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_v2_mkldnn_op.py
jakpiase: That is a typo. I have probably changed all transpose_X into trans_x by a regex and changed the parameter name by a mistake. I'll change that

lidanqing-vv (approved):
LGTM. Thanks for explanations !

jczaja (approved):

PR #34197, opened 2021-07-15
move the recv op the beginning of the forward/backward phase for pipeline

sandyhouse:
### PR types

Function optimization
### PR changes

Others
### Describe

move the recv op the beginning of the forward/backward phase for pipeline


ForFishes (approved):
LGTM

PR #34209, opened 2021-07-16
Fix cast op that can not cast the arrays that  the size of arrays is beyond int32

youth123:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix cast op that can not cast the arrays that  the size of arrays is beyond int32


sandyhouse (approved):
LGTM

PR #34228, opened 2021-07-19
[cherry-pick] [Paddle-TRT] Fix output data type selection

jiweibo:
### PR types

Others
### PR changes

Others
### Describe

The output type in plugins should be same as input type instead of data_type_.
Or it will make trt fix the network output to data_type_.

cherry-pick https://github.com/PaddlePaddle/Paddle/pull/34040


Superjomn (approved):
LGTM

PR #34234, opened 2021-07-19
[NPU] add is_empty_op_npu, test=develop

qili93:
### PR types

New features

### PR changes

OPs

### Describe

Add new op of is_empty for NPU.

Note: aclrtMemcpy will throw error if dst is null, while cudaMemcpy will NOT.

```log
[ERROR] ASCENDCL(32786,python):2021-07-19-15:28:03.398.113 [/home/jenkins/agent/workspace/Compile_ACL_Centos_X86/acl/runtime/memory.cpp:225]32786 aclrtMemcpy: param [dst] must not be null.
```


zhiqiu (commented):
Suggest to resue `IsEmptyOpKernel`, just need to register for NPU.

zhiqiu (approved):

PR #34240, opened 2021-07-19
[NPU] Refine npu unit tests

zhiqiu:
### PR types

Function optimization
### PR changes

Others
### Describe

Refine npu unit tests
- using `@unittest.skip()` and `check_dygraph=False` needs approval
image

However,
1. `@unittest.skip()` is not needed, since the unittests in `.../npu` will not be gererated if paddle is not compiled with NPU, so it is not needed.
2. Runing npu op in dygraph mode is supported now.

So, this PR removes all  `@unittest.skip()` and `check_dygraph=False` in npu tests.


zhwesky2010 (approved):
LGTM for change unittests

qili93 (approved):
LGTM

raindrops2sea (approved):
LGTM

PR #34256, opened 2021-07-20
Add Dependency to Fix Random Compilation Failure

zhhsplendid:
### PR types

Bug fixes
### PR changes

Others
### Describe

Add boost as dependency to fix random compilation failure. This is due to program_processing.cc used boost but didn't write boost into DEPS in the CMakeLists.txt


Aurelius84 (approved):
LGTM

PR #34261, opened 2021-07-20
Optimize inference fusion pass logs to avoid duplication

cryoco:
### PR types

Others
### PR changes

Others
### Describe

Optimize inference fusion pass logs to avoid duplication for passes  with multiple gpds.
Logs show the actual fusion count in total, rather than the number of gpds applied.

Taking embedding_eltwise_layernorm_fuse_pass for example:
Before:
image
After:
image


shangzhizhou (approved):
LGTM

PR #34273, opened 2021-07-20
Fix the save logic for the qat saving unit test.

wzzju:
### PR types

Bug fixes

### PR changes

Others

### Describe

Fix the save logic for the qat saving unit test. The error that occurred before is described as following.
image


juncaipeng (approved):
LGTM

PR #34274, opened 2021-07-20
copy found_inf to cpu in advance to improve performance

zhiqiu:
### PR types

Performance optimization
### PR changes

Others
### Describe

Since found_inf needs to be on cpu in adam op, we copy it in advance to avoid multiple time copies.

### DIFF
--- a/paddle/fluid/operators/matmul_op_npu.cc
+++ b/paddle/fluid/operators/matmul_op_npu.cc
@@ -1,0 +1,183 @@
+/* Copyright (c) 2021 PaddlePaddle Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


pangyoki (approved):
LGTM

chenwhql (approved):
LGTM for PADDLE_ENFORCE

zhhsplendid (approved):
LGTM for ShareDataWith, but discussed that name like "x_" can be changed because it usually looks like class private data member name. We will change in the future PR.

phlrain (approved):

zhiqiu:
> LGTM for ShareDataWith, but discussed that name like "x_" can be changed because it usually looks like class private data member name. We will change in the future PR.

Refined in https://github.com/PaddlePaddle/Paddle/pull/34330, thanks.

PR #34281, opened 2021-07-20
Get env variable

parulgrover20:


luotao1:
Thanks for your contribution, https://github.com/PaddlePaddle/Paddle/pull/45229 has already resolved the problem.

PR #34311, opened 2021-07-21
Fix hapi fleet bug in static mode

LiuChiachi:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix hapi fleet bug in static mode,  remove ``_init_context `` when using fleet API


wangxicoding (approved):
LGTM

PR #34316, opened 2021-07-21
FP32 LearningRate input for SGD with bfloat16.

arogowie-intel:
### PR types
New features

### PR changes
OPs

### Describe
This PR introduce FP32 data type for LearningRate input Tensor for SGD operator when used bfloat16 data type (for other inputs) to preserve better accuracy.


arogowie-intel:
@jczaja @arlesniak  Do we want to have LR Tensor always of FP32 type when used BF16 or maybe this should be controlled by ie. _mixed_precision_ SGD's attribute?

jczaja:
@arogowie-intel regardin type of LR Tensor, As fit a line (the easiest possible model) got some problem due to LR being in bf16 and tensorflow got LR always in floating point then I think it should be always float. I guess actual research engineers are better to comment on that one

PR #34319, opened 2021-07-21
Fix concat bug when used for tensors with large shapes

wuhuachaocoding:
### PR types

Bug fixes
### PR changes

 OPs
### Describe

in the past, using the concat for tensors with large shapes was easily out of the integer range.   
Fix concat bug when used for tensors with large shapes


gongweibao (approved):
LGTM

PR #34329, opened 2021-07-22
Fix for failing CI(test_activation_mkldnn_op.py)

jakpiase:
### PR types
Bug fixes

### PR changes
Others

### Describe
This PR fixes CI problem with test_activation_mkldnn_op.py failing
There is this "@OpTestTool.skip_if_not_cpu_bf16()" decorator which was misplacedly used inside a class, so base class was running and failing(that's why the log was like "exp op" instead of "sigmoid op". In this PR I have moved all my tests to separate file and disabled whole class(outside of class instead of inside the class, so base class will be skipped as well)


jczaja (approved):
LGTM

python/paddle/fluid/tests/unittests/mkldnn/test_activation_mkldnn_op.py
```diff
@@ -79,90 +79,6 @@ def setUp(self):
         self.attrs = {"use_mkldnn": True, "approximate": True}
 
 
-#Use it as a base class for BF16 activation tests, just override necessary functions
-class TestMKLDNNSigmoidBF16Op(TestActivation):
-    @OpTestTool.skip_if_not_cpu_bf16()
```
lidanqing-vv, line 84: I have two questions.
- Is it enough to just switch line 83 and line 84? Moving to another file is for clearness or other purpose? 
- Also you wrote in description: moving the decorator outside the class will also make the base class skip bf16 related functions? You mean TestActivation will also skip bf16 related functions?

jakpiase: I have moved it for clearness. I mean that both TestMKLDNNSigmoidBF16Op and TestActivation will be skipped in this solution. Earlier only TestMKLDNNSigmoidBF16Op was skipped.

lidanqing-vv: - Thanks
- TestActication class is mkldnn related class? Why it need to skip bf16? Also, I consulted Artur, he said decorator could be inherited(From parent to child), but could not be passed from child to parent. Is it real TestActication will skip BF16 functions?

arlesniak (approved):
LGTM

lidanqing-vv (approved):
LGTM

python/paddle/fluid/tests/unittests/mkldnn/test_activation_mkldnn_op.py
jakpiase: Artur was absolutely right. Decorator will be passed from parent to child but not from child to parent. It was my misplacement of that decorator. I am sorry for that situation.

PR #34381, opened 2021-07-26
Fix lrn bug when shape=0

huangjun12:
### PR types
Bug fixes

### PR changes
APIs

### Describe
Fix lrn bug when shape=0


jeff41404 (approved):
lgtm

PR #34386, opened 2021-07-26
install lsof in npu docker

zhiqiu:
### PR types

Others
### PR changes

Others
### Describe

install `lsof` in npu docker, so users can use it to check which card is used.

usage: `lsof /dev/davinci_manager`


qili93 (approved):
LGTM

PR #34387, opened 2021-07-26
Fix local variable in check_file_diff_approvals

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix local variable usage in check_file_diff_approvals

`ERROR_LINES` in `if_statement` is not a local variable in the shell.
It is used in two checks, if not initialized, it may result in the wrong message. 

For example,
image


lelelelelez (approved):
LGTM

PR #34396, opened 2021-07-26
【cherrypick】fix concat bug when used for tensors with large shapes.

wuhuachaocoding:
### PR types

Bug fixes
### PR changes

 OPs
### Describe

int the past, using concat API for tensors with large shapes will be out of the integer range.
so, I have fixed concat bug when used for tensors with large shapes.


wangxicoding (approved):
LGTM

PR #34431, opened 2021-07-27
fix optimizer.py

sljlp:
### PR types
  
 Bug fixes
### PR changes  
  
APIs 
### Describe

if op type is clear_float_status, 
it's device should be set self._device:all


wangxicoding (approved):
LGTM

PR #34475, opened 2021-07-29
[NPU] Avoid cpu tensor freed before copying to npu completed

zhiqiu:
### PR types

Function optimization
### PR changes

Others
### Describe

[NPU] Avoid cpu tensor freed before copying to npu completed


gongweibao (approved):
LGTM

PR #34488, opened 2021-07-29
fix force kill for elastic

kuizhiqing:
### PR types

Bug fixes

### PR changes

Others

### Describe

healthy process may not behave well in the case other process killed with -9 in the elastic scenario.


danleifeng (approved):
LGTM

PR #34505, opened 2021-07-30
Bug fix : Can't load multiple modules of custom c++ op 

zyfncg:
### PR types

Bug fixes
### PR changes

Others
### Describe

The framework will throw an error when user load custom c++ op from different modules, the PR fix this bug, so now user can load multiple modules of custom c++ op


chenwhql (approved):
LGTM

PR #34519, opened 2021-07-30
[hybrid] optimize pipeline performance with recompute and amp

wangxicoding:
### PR types

Performance optimization
### PR changes

Others
### Describe

optimize pipeline performance with recompute and amp.
Main optimization points
1. Recompute doesn't set op_device attr to backward op, will led Pipeline send/recv sub grad.
image
After we add op_device attr to recompute backward op, Pipeline send/recv is ok.
image
2. Add concat and split into amp gray_list. Before we add, concat only run fp32.
image
After we add concat into amp gray_list, concat can run will fp16 if inputs is fp16
image
3. When use Pipeline, if op's device is all, this op will compute in all PipelineStage. While when we use AMP,  AMP will insert cast op after this device:all op, and cast op's device is not all, the Pipeline will send cast var.
image
For optimize, we set this cast op as device:all if prev op is device:all, and Pipeline will not send this var.
image

#### Performance test
Test with Ernie3.0,  "hidden_size": 4096,  "num_attention_heads": 128,  "num_hidden_layers": 76,  "num_sharing_layers": 64,
16cards 32G V100，mp=8, pp=2, amp, recompute, gbs=32, micro_bs=2
| develop(tokens/s) | PR(tokens/s) | improve |
| -- | -- | -- |
| 2164.5 | 2445.2 | 12.96% |

### DIFF
--- a/python/paddle/fluid/backward.py
+++ b/python/paddle/fluid/backward.py
@@ -944,6 +944,13 @@ def _append_backward_ops_with_checkpoints_(
         for op_desc in reversed(added_descs):
             grad_op_desc, op_grad_to_var = core.get_grad_op_desc(
                 op_desc, cpt.to_text(no_grad_dict[block.idx]), [])
+
+            # Set device for grad_op according to forward Op
+            if op_desc.has_attr(device_attr_name):
+                op_device = op_desc.attr(device_attr_name)
+                for g_op_desc in grad_op_desc:
+                    g_op_desc._set_attr(device_attr_name, op_device)
+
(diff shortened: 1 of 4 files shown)


gongweibao (commented):
介绍一下做了哪些改动？为何能提升性能？提升多少性能？

sandyhouse (approved):
LGTM

JZ-LIANG (approved):
LGTM

zhiqiu (approved):
LGTM for backward.py

PR #34521, opened 2021-07-31
Fix Inference CE Error by Topo Order

zhhsplendid:
### PR types

Bug fixes
### PR changes

Others
### Describe

This PR changed Topological Order Sort from Graph to Program. The TopologySortGraphByDescOrder caused Inference CE error. After debug, we used an image to illustrate the reason:

47905d81fd23548b6b6db38b4d7dfd0b

From above sub-graph, some fusions didn't solve Hazard like WAW or WAR (reference: https://en.wikipedia.org/wiki/Hazard_(computer_architecture)#Write_after_read_(WAR) ) for the Var Node "res2c.add.output.5.tmp_0" which makes different topological orders can have different outputs because Inference translates from Graph to Program to run.

Why didn't the old system have the problem? The reason is that old Topological Sort also had fixed order by node id! (see: https://github.com/PaddlePaddle/Paddle/blob/8b72a1a73d3913d4d8aca88e6c47988278c392d4/paddle/fluid/framework/ir/graph_helper.cc#L129). The old system combined wrong Graph format and special Topological Sort Order to run the correct output.

I don't have time to debug on lots of fusions to resolve Hazard, so this PR just make our new Topological Sort Order to be close to the old Topological Sort Order.

### DIFF
--- a/paddle/fluid/framework/ir/graph.cc
+++ b/paddle/fluid/framework/ir/graph.cc
@@ -17,7 +17,7 @@ limitations under the License. */
 #include "paddle/fluid/framework/ir/graph.h"
 #include "paddle/fluid/framework/operator.h"
 
-DEFINE_bool(convert_all_blocks, false,
+DEFINE_bool(convert_all_blocks, true,
             "Convert all blocks in program into SSAgraphs");
 
 namespace paddle {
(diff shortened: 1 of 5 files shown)


sneaxiy (approved):
LGTM.

phlrain (approved):

PR #34524, opened 2021-07-31
Add basic functions of Program Pass

sneaxiy:
### PR types
New features

### PR changes
Others

### Describe
Add basic functions for `Pass` applied to `ProgramDesc`. More codes would be added in the next PRs.

### DIFF
--- a/paddle/fluid/framework/ir/graph.h
+++ b/paddle/fluid/framework/ir/graph.h
@@ -47,6 +47,10 @@ constexpr char kStaleProgramOpDescs[] = "stale_program_op_descs";
 
 namespace ir {
 
+const char kGraphToProgramVarsToRemove[] =
+    "__graph_to_program_vars_to_remove__";
+const char kGraphToProgramSortKind[] = "__graph_to_program_sort_kind__";
+
 /*
  * The graph is a Directed Acyclic Single Static Assignment Graph.
  *
--- a/paddle/fluid/framework/ir/graph.h
+++ b/paddle/fluid/framework/ir/graph.h
@@ -47,10 +47,6 @@ constexpr char kStaleProgramOpDescs[] = "stale_program_op_descs";
 
 namespace ir {
 
-const char kGraphToProgramVarsToRemove[] =
-    "__graph_to_program_vars_to_remove__";
-const char kGraphToProgramSortKind[] = "__graph_to_program_sort_kind__";
-
 /*
  * The graph is a Directed Acyclic Single Static Assignment Graph.
  *
--- a/paddle/fluid/framework/ir/graph_helper.cc
+++ b/paddle/fluid/framework/ir/graph_helper.cc
@@ -15,7 +15,9 @@ limitations under the License. */
 #include "paddle/fluid/framework/ir/graph_helper.h"
 #include <queue>
 #include <stack>
+#include "paddle/fluid/framework/op_proto_maker.h"
 
+DECLARE_bool(convert_all_blocks);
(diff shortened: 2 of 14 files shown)


paddle/fluid/framework/ir/graph_helper.cc
```diff
@@ -475,6 +477,117 @@ std::vector<ir::Node *> TopologySortGraphByDescOrder(const Graph &graph) {
...
+    } else if (n->Op()) {
+      ops->emplace_back(*n->Op());
+    }
+    // delete no OpDesc op
+  }
+}
+
+static void GraphToBlock(const Graph &graph, proto::BlockDesc *block,
+                         const SortKind *sort_kind) {
+  // Remove the unneeded variables after memory optimization.
```
zhhsplendid, line 514: Just for you information, nothing to change.

I don't know where did we use the kGraphToProgramVarsToRemove and `var2remove` ... I thought we can delete these code, but I am not sure so I didn't do it in my past PRs.

paddle/fluid/framework/ir/graph.h
```diff
@@ -47,6 +47,10 @@ constexpr char kStaleProgramOpDescs[] = "stale_program_op_descs";
 
 namespace ir {
 
+const char kGraphToProgramVarsToRemove[] =
```
zhhsplendid: Should we put it at `graph.h`? It is only used in GraphToProgram and your PR moved the GraphToProgram to graph_helper, can we put those constant at `graph_helper.h`?

python/paddle/fluid/framework.py
```diff
@@ -4148,6 +4164,91 @@ def __init__(self):
         # compiled program, i.e. Graph
         self._graph = None
 
+    def _find_var_ctor_kwargs(self, new_desc):
```
zhhsplendid: what's the meaning of `ctor`? From the code, you are finding 'class' and 'kwargs', can we rename the method name as `_find_var_class_kwargs`?

paddle/fluid/framework/ir/graph_helper.cc
sneaxiy: Thanks for your FYI! I agree with you. Just keep it as the old logic.

paddle/fluid/framework/ir/graph.h
sneaxiy: Done.

python/paddle/fluid/framework.py
sneaxiy: Done.

zhhsplendid (approved):
LGTM

python/paddle/fluid/framework.py
```diff
@@ -4148,6 +4164,91 @@ def __init__(self):
...
+    def _find_var_class_kwargs(self, new_desc):
+        old_desc = self.desc
+        all_new_vars = []
+        block_num = new_desc.num_blocks()
+        for idx in range(block_num):
+            new_block_desc = new_desc.block(idx)
+            all_new_vars.append([])
+            block_new_vars = all_new_vars[-1]
+            for new_var_desc in new_block_desc.all_vars():
+                if self.blocks[idx].has_var(new_var_desc.name()):
```
zhiqiu, line 4176: The block number of new_desc and old_desc is always equal?

zhiqiu (approved):
LGTM for the framework.py

python/paddle/fluid/framework.py
sneaxiy: For now, it is true. We can fix this if this is not in the future.

PR #34619, opened 2021-08-04
[NPU]Use another method to void c_allreduce_sum core!

gongweibao:
### PR types
Function optimization

### PR changes
Others

### Describe
1. Checknumeric is very slow and
2. ReaduceMeanD will trigger `sum` op error under fp16. So cast input tensor to fp32 to avoid this.
3. `c_allreduce_sum` doesn't core under fp16, but core under fp32.


zhiqiu (approved):
LGTM

chalsliu (approved):

PR #34624, opened 2021-08-05
Fix paddle buffered data reader bug

DesmonDay:
### PR types
Bug fixes
### PR changes
APIs
### Describe
When a class object compares with a data array, it may raise problems. Therefore, fix the bug by changing buffered data reader's `EndSignal` to `None`.
屏幕快照 2021-08-04 下午9 03 22


ZHUI (changes requested):
Please add some tests for the code.

ZHUI (approved):
LGTM

PR #34664, opened 2021-08-06
Recompute: fix bug with transformer attention mask

JZ-LIANG:
### PR types
Bug fixes

### PR changes
OPs

### Describe
  NOTE In Transformer-like network, if user put the attention mask into the recompute segment output,
  pylayer will force the stop_gradient of attention mask to be False, which will make the number of 
  tensor that need grad does not match.
  the backward_inputs_with_grad is used to avoid this case.


ForFishes (approved):
LGTM

PR #34668, opened 2021-08-06
support batch_norm multi-threading on cpu

Zjq9409:
### PR types

Performance optimization
### PR changes

OPs
### Describe

Support batch_norm multi-thread on cpu, the performance test data of OMP==20 is as follows：
image


paddle/fluid/operators/batch_norm_op.cc
```diff
@@ -353,17 +352,45 @@ class BatchNormKernel<platform::CPUDeviceContext, T>
...
           ConstEigenArrayMap<T> x_arr(x->data<T>(), sample_size, N * C);
+          // temporary variables for storing intermediate values
+          framework::Tensor tmp_tensor;
+          auto *tmp = tmp_tensor.mutable_data<T>(
+              framework::make_ddim({1, N * C}), ctx.GetPlace());
+          EigenVectorArrayMap<T> tmp_arr(tmp, 1, N * C);
+
+#ifdef PADDLE_WITH_MKLML
+#pragma omp parallel for
+#endif
```
JamesLim-sy: 既然源语的格式固定，又始终跟for循环配合使用，可以尝试用宏封装一下这部分：
```
#ifdef PADDLE_WITH_MKLML
 #pragma omp parallel for
 #endif

for(int nc = 0; nc < N * C; ++nc) {
   ....;
}
```

PR #34692, opened 2021-08-07
enable train thread larger than read thread

kuizhiqing:
### PR types

New features

### PR changes

Others

### Describe

Training thread can be set large than read thread when using QueueDataSet, this feature will be used in Steam training scenario.


yaoxuefeng6 (approved):
LGTM

PR #34700, opened 2021-08-09
[NPU] add lock for npu_pinned_allocator

zhiqiu:
### PR types

Function optimization
### PR changes

Others
### Describe

add lock for npu_pinned_allocator to support multi-thread usage, for example, Parallel executor.


pangyoki (approved):
LGTM

PR #34719, opened 2021-08-09
Fix error of HSigmoidLoss

linjieccc:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix bugs of HSigmoidLoss where memory overflow occurs if the first dimension of input and label is not equal.


wawltor (approved):
LGTM

PR #34773, opened 2021-08-10
add (N,C,*) input support for GroupNorm

zoooo0820:
### PR types
  Bug fixes

### PR changes
OPs

### Describe
The nn.GroupNorm is only support 4-D input now. However, there is no shape check for input tensor.  So there are two type problems.
1. When input dimension is less than 4-D, it may cause errors and other Paddle operations will also crash.
2. When input dimension is larger than 4-D, the output result is not correct.

What this PR did:
1. add (N, C, *) (>= 2-D) input support.
2. change API description in python file.


python/paddle/nn/layer/norm.py
```diff
@@ -338,8 +338,8 @@ class GroupNorm(Layer):
         name(str, optional): Name for the GroupNorm, default is None. For more information, please refer to :ref:`api_guide_Name`..
 
     Shape:
-        - x: 4-D tensor with shape: (batch, num_features, height, weight).
-        - output: 4-D tensor with same shape as input x.
+        - x: Tensor with shape: (N, C, *), where N is batch_size, C is num_features.
```
jerrywgz: 这里需要说明得更清晰些，比如tensor size的要求，不同的layout对应的shape也应该是不一样的。另外需要更新下fluiddoc下的中文文档

zoooo0820: 目前该API只支持NCHW layout输入，参考相似API，中文文档已修改。

fluid doc预览
gn

jerrywgz (approved):

dingjiaweiww (approved):

Issue #34574, opened 2021-08-03
quantized segmentation model raises relu error


juncaipeng:
Download the demo(url:https://pan.baidu.com/s/1NW7L4-fggH-nnJE9SNu63g  passwd:kf1l) and refer to the readme.

Error as follows:
```
NotFoundError: Operator (relu) does not have kernel for data_type[int8_t]:data_layout[ANY_LAYOUT]:place[CPUPlace]:library_type[PLAIN].
```

wozna:
@juncaipeng I just started investigating this problem. This model has an inplace operation with elementwise_add and relu. It turned out that the quantization process is not immune to this, hence this error. Therefore I will start working to find the best solution for this.

wozna:
@juncaipeng  I found the problem, it turned out we had an error in elementwise_add quantization because we set the `force_fp32_output` attribute if the output scale was not found. Unfortunately, this operator does not have this attribute or this functionality. I fixed it in PR https://github.com/PaddlePaddle/Paddle/pull/34820

juncaipeng:
Thank you, I will test the PR.

lidanqing-vv:
@juncaipeng 你有时间测了吗？

PR #34820, opened 2021-08-11
Fix elementwise_add quantization

wozna:
### PR types

Bug fixes
### PR changes

OPs 
### Describe

This PR:
- removes `relu` from the list of quantizable operators because so far `relu` has no quantization support,
- modifies quantization process for `elementwise_add`: if there is no output tensor scale, the operator is not quantized. Previously, when there was no output scale, the `force_fp32_output` attribute was set, unfortunately, it was an error because the operator does not have such an attribute or functionality. 

This PR should fix the problem https://github.com/PaddlePaddle/Paddle/issues/34574

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
@@ -770,7 +770,8 @@ void CPUQuantizePass::QuantizeElementwiseAdd(Graph* graph) const {
     GET_IR_NODE_FROM_SUBGRAPH(elementwise_add_out, elementwise_add_out,
                               elementwise_add_pattern);
 
-    if (!AreScalesPresentForNodes({elementwise_add_x, elementwise_add_y})) {
+    if (!AreScalesPresentForNodes({elementwise_add_x, elementwise_add_y}) ||
+        !AreScalesPresentForNodes({elementwise_add_out})) {
       LogCannotQuantizeOp(elementwise_add_op);
       return;
     }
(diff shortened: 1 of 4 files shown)


jczaja (dismissed):
LGTM

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -770,7 +770,8 @@ void CPUQuantizePass::QuantizeElementwiseAdd(Graph* graph) const {
     GET_IR_NODE_FROM_SUBGRAPH(elementwise_add_out, elementwise_add_out,
                               elementwise_add_pattern);
 
-    if (!AreScalesPresentForNodes({elementwise_add_x, elementwise_add_y})) {
+    if (!AreScalesPresentForNodes({elementwise_add_x, elementwise_add_y}) ||
+        !AreScalesPresentForNodes({elementwise_add_out})) {
```
sfraczek: Consider calling this method only once with all three nodes in arguments.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass_tester.cc
```diff
@@ -131,13 +131,13 @@ TEST(QuantizerPlacementPass, enabled_conv_excluded_one) {
...
-  MainTest({}, {}, 6);
+  // all quantized operators, except relu
+  MainTest({}, {}, 5);
 }
 
 TEST(QuantizerPlacementPass, default_attr_value) {
-  //  all operators quantized
-  DefaultAttrTest(6);
+  //  all quantized operators, except relu
+  DefaultAttrTest(5);
```
sfraczek: I think that reversing order of words "quantized" and "operators" might be confusing.  
You could consider making it a full sentence for example:
_All operators except relu should be quantized_

sfraczek (approved):
LGTM

lidanqing-vv (approved):
LGTM

PR #34830, opened 2021-08-11
[oneDNN ] disabling more ops caching

jczaja:
### PR types
Bug fixes

### PR changes
OPs 

### Describe
Recently oneDNN introduced caching of its elements: primitives and primitives descriptors so we can disable partially our own caching. This PR disabled caching of LRN, Transpose, Layer Norm and Sum oneDNN ops


paddle/fluid/operators/mkldnn/sum_mkldnn_op.cc
```diff
@@ -45,82 +45,63 @@ using paddle::platform::MKLDNNDeviceContext;
...
-    const std::string key_pd = this->key_ + "@fwd_pd";
-    this->fwd_pd_ = std::static_pointer_cast<dnnl::sum::primitive_desc>(
-        this->dev_ctx_.GetBlob(key_pd));
-    if (this->fwd_pd_ == nullptr) {
-      this->fwd_pd_.reset(new dnnl::sum::primitive_desc(dst_md, scales, srcs_md,
-                                                        this->engine_));
-      this->dev_ctx_.SetBlob(key_pd, this->fwd_pd_);
-    }
+    this->fwd_pd_.reset(
+        new dnnl::sum::primitive_desc(dst_md, scales, srcs_md, this->engine_));
```
jakpiase, line 85: You are using "new" here. Is it later destroyed anywhere?

jczaja: this->fwd_pd_ is a shared pointer so when reference count will get to zero then deallocation of  of object will happen . So we do not have to worry about memory leaks.

jakpiase: Ok, thanks. I didn't see that it is a smart pointer

jakpiase (approved):
LGTM

lidanqing-vv (approved):
LGTM.

PR #34870, opened 2021-08-13
[hybrid] refine pipeline stage and mp send/recv check

wangxicoding:
### PR types

Others
### PR changes

Others
### Describe

1. Refine pipeline stage check, move the check into insert_send_recv. Even if the stage is unordered, as long as send/recv inserted in order, it can pass the check.
2. Remove the check of that size of send/recv var must be divisible by mp_degree. If it cannot be divisible, send/recv full data.


sandyhouse (approved):
LGTM

PR #34884, opened 2021-08-13
[CustomOp] Fix ext_tensor.cast failed bug

chenwhql:
### PR types

Bug fixes
### PR changes

Others
### Describe

The code in macro `__NVCC__` need to put into `.cu` file and compiled by `nvcc`.

```
#if defined(__NVCC__) || defined(__HIPCC__)
    } else if (platform::is_gpu_place(in_.place())) {
      platform::Transform<platform::CUDADeviceContext> trans;
      auto *context = static_cast<const platform::CUDADeviceContext *>(ctx_);
      trans(*context, in_begin, in_end, out_begin,
            CastDataTypeFunctor<InType, OutType>());
      context->Wait();
#endif
```

Before this PR, if we call `tensor.cast(x)` and x.place is kGPU, it will throw error.

e72962a35b5a5e853a5ddc84e251c33d

2e39c354ef9e4a5c6eb151b7a7efe45e


zhwesky2010 (approved):
LGTM

paddle/fluid/framework/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ function(windows_symbolic TARGET)
     add_custom_command(OUTPUT ${final_path}/.${src}.cu
             COMMAND ${CMAKE_COMMAND} -E copy_if_different "${final_path}/${src}.cc" "${final_path}/.${src}.cu"
             COMMENT "create hidden file of ${src}.cu")
-    add_custom_target(${TARGET} ALL DEPENDS .${src}.cu)
+    add_custom_target(${TARGET} ALL DEPENDS ${final_path}/.${src}.cu)
```
jim19930609, line 20: LoL, so we have a file named "ext_tensor.cu" for Linux Symlink, and a hidden ".ext_tensor.cu"(actually .ext_tensor.cu.cu) for Windows Symlink. What if there were a third/fourth... platform?

jim19930609 (approved):

paddle/fluid/framework/CMakeLists.txt
zhwesky2010: `src` is `ext_tensor` here. If system is based on Unix, these is no problem generally.

chenwhql: we should remove these symlink cases later, include data_type_transform, tensor_utils, etc.

PR #34888, opened 2021-08-13
Fix random generator thread safety bug

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
The `thread_offset` should be read when the lock is held.


zhiqiu (approved):
LGTM

PR #34906, opened 2021-08-15
Weighted CEloss bug

HydrogenSulfate:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Reason: According to the issue reported by PR https://github.com/PaddlePaddle/PaddleSeg/pull/1240. It is found that the cross entropy loss will have two possible error reporting bugs under the condition of using the combination of hardlabel+weight parameters.
1. If the input matrix is ​​too large, it may cause an access error when calculating the maximum/minimum value of the label, which is related to the implementation of paddle.min/max OP
2. There is a value that exceeds the number of categories in the input hardlabel, but the value is marked as ignore_index, and an error will still be reported at this time.

solution:
1. First, ensure that the number of categories of weight must be aligned with the number of categories of the input matrix, otherwise an unequal number of categories exception will be thrown
2. Modify the judgment logic of whether the maximum and minimum values ​​of the hardlabel exceed the upper or lower bounds of the number of categories: first ignore the ignore_index in the hardlabel, and then call paddle.nonzero (hardlabel<0) and paddle.nonzero (hardlabel>0) to determine instead, if The nonzero function returns at least one element index that meets the conditions, which means that at least one value of hardlabel exceeds the upper/lower bounds, and an exception should be thrown at this time.
3. After excluding the possible abnormalities of (1.) and (2.), use gather_nd to sample the weight of each category from the input weight tensor. At this time, the weight of the ignore_index position in the sampling result is temporarily from the 0th category, so also need to use a mask to reset the ignore_index position's weight to 0.0 .

Modification points:
1. Modified part of the logic of cross-entropy loss under the dynamic graph and static graph to avoid using the paddle.min/max function temporarily caused by the problem to determine the upper and lower bounds
2. Added several corresponding unit tests


LutaoChu:
LGTM

python/paddle/fluid/tests/unittests/test_cross_entropy_loss.py
```diff
@@ -1013,7 +1034,7 @@ def test_cross_entropy_loss_1d_with_weight_none_func(self):
     def test_cross_entropy_loss_1d_mean(self):
         input_np = np.random.random([100, 200]).astype(self.dtype)  #N,C
         label_np = np.random.randint(0, 100, size=(100)).astype(np.int64)  #N,1
-        weight_np = np.random.random([200]).astype(self.dtype)  #C
+        # weight_np = np.random.random([200]).astype(self.dtype)  #C
```
chajchaj: 多余代码的话删掉

python/paddle/fluid/tests/unittests/test_cross_entropy_loss.py
```diff
@@ -1022,7 +1043,7 @@ def test_cross_entropy_loss_1d_mean(self):
         with fluid.program_guard(prog, startup_prog):
             input = fluid.data(name='input', shape=[100, 200], dtype=self.dtype)
             label = fluid.data(name='label', shape=[100], dtype='int64')
-            weight = fluid.data(name='weight', shape=[100], dtype=self.dtype)
+            # weight = fluid.data(name='weight', shape=[100], dtype=self.dtype)
```
chajchaj: 多余代码的话删掉

chajchaj (changes requested):

python/paddle/fluid/tests/unittests/test_cross_entropy_loss.py
HydrogenSulfate: done

python/paddle/fluid/tests/unittests/test_cross_entropy_loss.py
HydrogenSulfate: done

chajchaj (approved):
LGTM

PR #34915, opened 2021-08-16
Enhance tensor shape check for dist op.

ZHUI:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Enhance tensor shape check for dist op.


wawltor (approved):
LGTM

PR #34948, opened 2021-08-16
fix the bug of channel-wise quantization for ernie

XGZhang11:
### PR types

Bug fixes
### PR changes

OPs 
### Describe

support channel-wise quantization of ernie
Dequant the output of conv or mul at the "x_num_col_dims" dim. For example, if "x_num_col_dims"  is 1, we will perform the channel wise dequant at the first dim of output. If "x_num_col_dims"  is 2, we will perform the channel wise dequant at the second dim of output.


qingqing01 (approved):

Superjomn (approved):
LGTM

PR #34972, opened 2021-08-17
Add a check for multiplex op

aoyulong:
### PR types
Others

### PR changes
Ops

### Describe
The paddle.multiplex will break down when the size of candidate tensors is zero and doesn't give useful an error message. This pr will check the situation and gives user more information.


paddle/fluid/operators/multiplex_op.cu
```diff
@@ -29,6 +29,10 @@ class MultiplexGPUKernel : public framework::OpKernel<T> {
     auto* out = ctx.Output<Tensor>("Out");
     out->mutable_data<T>(ctx.GetPlace());
 
+    PADDLE_ENFORCE_GT(ins[0]->numel(), 0,
```
ForFishes: Why only check the first tensor of the input? Shouldn't it be checked more than one becasue of MultiInput type?

paddle/fluid/operators/multiplex_op.h
```diff
@@ -31,6 +31,10 @@ class MultiplexCPUKernel : public framework::OpKernel<T> {
 
     out->mutable_data<T>(ctx.GetPlace());
 
+    PADDLE_ENFORCE_GT(ins[0]->numel(), 0,
```
ForFishes: Same as above.

paddle/fluid/operators/multiplex_op.h
```diff
@@ -31,6 +31,10 @@ class MultiplexCPUKernel : public framework::OpKernel<T> {
...
+                      platform::errors::PreconditionNotMet(
+                          "indexing will be out of bounds with size 0."));
```
ForFishes: indexing->index ? PreconditionNotMet-> OutOfRange?

aoyulong: Done.

paddle/fluid/operators/multiplex_op.h
aoyulong: Done.

paddle/fluid/operators/multiplex_op.cu
aoyulong: Done.

ForFishes (approved):
LGTM

PR #34978, opened 2021-08-17
add the safe check for the some ops

wawltor:
### PR types
Function optimization

### PR changes
OPs 

### Describe
add the safe check for the some ops


joey12300 (approved):
LGTM

PR #34981, opened 2021-08-18
Support CostInfo and MemProfiler in InterpreterCore

Aurelius84:
### PR types

New features
### PR changes

APIs
### Describe

Support CostInfo in New Executor

### Why modify into Atomic?

+ [Which is more efficient, basic mutex lock or atomic integer?](https://stackoverflow.com/questions/15056237/which-is-more-efficient-basic-mutex-lock-or-atomic-integer)
> If you have a counter for which atomic operations are supported, it will be more efficient than a mutex.
> 1. Atomic operations leverage processor support (compare and swap instructions) and don't use locks at all, whereas locks are more OS-dependent and perform differently on, for example, Win and Linux.


paddle/fluid/framework/new_exec.h
```diff
@@ -112,6 +156,36 @@ struct OpFuncNode {
...
+    } else {
+      total_time += cost_timer.ElapsedMS();
+    }
+  }
+
+  double total_time;           // ms
+  size_t host_memory_bytes;    // bytes
+  size_t device_memory_bytes;  // bytes
+  platform::Timer cost_timer;
+  platform::Timer exclude_timer;
```
aoyulong: What's the function of the exclude_timer? Besides, Why does CostInfo contain Timer? It will be better to decouple these two or rename CostInfo to something else such as CostCollector.

Aurelius84: Thanks, fixed

wanghuancoder (approved):
LGTM

PR #35030, opened 2021-08-19
[oneDNN]  disable caching for interpolate and batch Norm

jczaja:
### PR types
Others

### PR changes
OPs

### Describe
PR is disabling caching oneDNN objects  by PaddlePaddle cache (cache of oneDNN library will be used instead)

### DIFF
--- a/paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
@@ -148,7 +148,7 @@ class InterpolateMKLDNNKernel : public framework::OpKernel<T> {
 
     auto out_dims_vec = ComputeOutputShape(ctx);
     framework::DDim dim_out = framework::make_ddim(out_dims_vec);
-    z->reshape(dim_out);
+    z->Resize(dim_out);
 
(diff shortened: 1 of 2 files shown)


arlesniak (approved):
LGTM

paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
```diff
@@ -155,11 +148,10 @@ class InterpolateMKLDNNKernel : public framework::OpKernel<T> {
 
     auto out_dims_vec = ComputeOutputShape(ctx);
     framework::DDim dim_out = framework::make_ddim(out_dims_vec);
-    z->mutable_data<T>(dim_out, ctx.GetPlace());
+    z->Resize(dim_out);
```
jakpiase, line 151: Why did you earlier need this "mutable_data" call and why don't you need it now?

jczaja: Great question. mutable data is not needed here as it is called internally inside AcquireDstMemory so having it here(as deleted) was creating situation that it was called twice which was not needed and in some situations it could cause performance drop

jakpiase (approved):
LGTM

lidanqing-vv (approved):
LGTM

PR #35045, opened 2021-08-20
[NPU] add int64_t kernels for YoloV3, test=develop

qili93:
### PR types

Bug fixes

### PR changes

OPs

### Describe

YOLOV3 need NPU kernels for dtype of int64_t, update the following ops:

1. Add `int64_t` NPU kernel and update unit test to add int32 and int64 test cases
2. Fix some ops:
    - fill_constant_op_npu.cc:  change NPU op from `FillD` to `Fill` as need to support `int64_t`, while `FillD` has issue in dtype of `int64_t`;
    - scale_op_npu.cc: No need to call Add OP, just need to update bias value and call `Power` OP only.
    - meshgrid_op_npu.cc: change NPU op from  `BroadcastToD` to `BroadcastTo` as need to support `int64_t`, as `BroadcastToD` cannot support dtype of `int64_t`;

- Compile when WITH_ASCEND_INT64 OFF

```
python -c $'import paddle\nfor k in paddle.fluid.core._get_all_register_op_kernels()["fill_constant"]:print(k)' | grep NPUPlace
```

image

- Compile when WITH_ASCEND_INT64 ON

image

Add NPU UT results here:

image
image

image
image

image
image

image
image

image

image
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,6 +224,7 @@ option(WITH_UNITY_BUILD "Compile with UnityBuild mode"             OFF)
 option(WITH_STRIP       "Strip so files of Whl packages"         OFF)
 option(NEW_RELEASE_CUBIN   "PaddlePaddle next-level release strategy for pypi cubin package"             OFF)
 option(NEW_RELEASE_JIT   "PaddlePaddle next-level release strategy for backup jit package"             OFF)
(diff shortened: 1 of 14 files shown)


zhiqiu (approved):
LGTM

wanghuancoder (approved):
LGTM

PR #35062, opened 2021-08-22
Fix a bug of strided_slice op, about the axes parameter access memory out of bounds

TeslaZhao:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix a bug of strided_slice op, about the axes parameter access memory out of bounds.


paddle/fluid/operators/strided_slice_op.cc
```diff
@@ -62,6 +62,20 @@ class StridedSliceOp : public framework::OperatorWithKernel {
     auto ends_size = ends.size();
     auto strides_size = strides.size();
 
+    for (size_t i = 0; i < axes.size(); ++i) {
+      PADDLE_ENFORCE_GE(axes[i], 0,
+                        platform::errors::InvalidArgument(
+                            "The axis should be greater than or equal to 0."
+                            "But received %d of axes[%d]",
+                            axes[i], i));
+      PADDLE_ENFORCE_LE(
```
Superjomn: axis[i] < size ?

so LT not LE

TeslaZhao: LT is right. Thanks for this advice.

Superjomn (approved):
LGTM

PR #35079, opened 2021-08-23
Fix a bug of transpose op, about accessing memory out of bounds of the perm param

TeslaZhao:
### PR types

Bug Fixes

### PR changes

OPs

### Describe

Each of perm params must be greater than or equal to 0, and less than input's dim size.


Superjomn (approved):
LGTM

PR #35083, opened 2021-08-23
add bilinear_interp trt converter

zhouping828:
add tensorRT converter for billinear_interp op


zhouping828:
use PR template and move to https://github.com/PaddlePaddle/Paddle/pull/35181

PR #35098, opened 2021-08-23
fix bmm bug

firestonelib:
### PR types

Bug fixes 
### PR changes

APIs
### Describe

Fix the bmm address points to the zero page error.


ForFishes (approved):
LGTM

PR #35116, opened 2021-08-24
[NPU] Fix the performance problem of elementwise_add when 'axis' is not specified

ronny1996:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix the performance problem of elementwise_add when 'axis' is not specified
373310ab6a54301ced349f6cc6465f3b
4a1d6f0f212db93290a98d1c016ee598


qili93 (approved):
LGTM

PR #35124, opened 2021-08-24
fix cpu adamw problem for np.float64

zhaoyinglia:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix cpu adamw problem when parameter's type is np.float64.


JZ-LIANG (approved):
LGTM

PR #35212, opened 2021-08-27
[Dy2Stat]Support append method and initialized value for List in ControlFlow

Aurelius84:
### PR types

Bug fixes
### PR changes

Others
### Describe

Support append method and initialized value for List in ControlFlow.

**Code Example:**
```python

def test_list_pop_in_for_loop(x, iter_num):
    # Use `fill_constant` so that static analysis can analyze the type of iter_num is Tensor
    iter_num = paddle.ones(shape=[1], dtype="int32") * iter_num  
    a = []
    b = [x - 1, x + 1]
    for i in range(iter_num):
        a.append(x + i)
        b.append(x * 2)

    one = paddle.ones(shape=[1], dtype="int32")
    for i in range(one.numpy()[0]):
        item = a.pop()

    return a[0], item, b[1]
```


TCChenlong (approved):
LGTM

XiaoguangHu01 (approved):
LGTM

PR #35223, opened 2021-08-27
[NPU] update batch norm op, test=develop

qili93:
### PR types

Performance optimization

### PR changes

OPs

### Describe

Update batch norm NPU op for better performance.


zhiqiu (approved):
LGTM

PR #35233, opened 2021-08-28
[Op Def] Add extra def of linear_interp & linear_interp_v2 & addmm

Ray2020BD:
### PR types

Bug fixes
### PR changes

Others
### Describe

Add extra and quant define for linear_interp & linear_interp_v2 & addmm


winter-wang (approved):
LGTM

PR #35275, opened 2021-08-30
[Don't merge] use link flags --exclude-libs to reduce size of paddle inference shar…

ZyhHelen:
### PR types
 Others

### PR changes
 Others

### Describe
use link flags --exclude-libs to reduce size of paddle inference shared library


luotao1:
Sorry for too late to notice this PR, please resolve conflicts first. @ZyhHelen

PR #35286, opened 2021-08-31
fix xpu adamw

zhaoyinglia:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix adamw api for xpu


JZ-LIANG (approved):
reverse xpu adamw to combination of ops version. LGTM

PR #35315, opened 2021-08-31
label the attrs of prelu op as extra

juncaipeng:
### PR types

Others
### PR changes

Others
### Describe

label the attrs of prelu op as extra


winter-wang (approved):
LGTM

PR #35326, opened 2021-08-31
bugfix for Tensor Parallel accuracy

JZ-LIANG:
### PR types
Bug fixes

### PR changes
Others

### Describe
Tensor Parallel need to broadcast parameters that are not splited when initialization.


sandyhouse (approved):
LGTM

PR #35335, opened 2021-09-01
[NPU] set constant before copy data

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
For NPU, the nan/inf check will check the operand for every operators. If the align parts of the coalesce tensor contain nan/inf, it will trigger the NPU's nan/inf check. This update is mainly for c_allreduce_sum fusion.

image


gongweibao (approved):
LGTM

wangxicoding (approved):
LGTM

PR #35342, opened 2021-09-01
[hybrid performance] reduce the allreduce times by sorting vars before fusing

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
reduce the allreduce times by sorting vars by its' dtypes before fusing


FeixLiu:
not useful

FeixLiu:
not useful since the allreduce_vars are all in the same type

PR #35358, opened 2021-09-01
add AsExtra to the reshape op

zmxdream:
### PR types

Function optimization
### PR changes

APIs
### Describe

add AsExtra to the following attribute of reshape op:
 1. use_mkldnn


winter-wang (approved):
LGTM

PR #35359, opened 2021-09-01
[NPU] Shard index op for npu

sljlp:
### PR types

Others
### PR changes

OPs
### Describe

prettify codes from https://github.com/PaddlePaddle/Paddle/pull/35281
add UT for int 64


wangxicoding (approved):
LGTM

PR #35367, opened 2021-09-01
Replace "pass" with "error exception"

xingfeng01:
### PR types
Others

### PR changes
APIs

### Describe

This PR adds error exception in fused transformer python interface. 
The function body are not implemented (will be implemented later). 
Following zhiqiu's comment in previous PR-35206 (merged already), it is better to raise an exception instead of using "pass". 

image


lanxianghit (approved):
LGTM

PR #35370, opened 2021-09-01
[NPU] release gil before op run

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

release gil before op run

### why
Ascend op will create python threads during running, if GIL is not realsed, a segment fault error will be raised.
image

Before this PR, the `op_test` uses `cpu place` to calculate numeric gradients in npu grad test.
After this PR, the `op_test` uses `npu place` to calculate numeric gradients in npu grad test directly.


liym27 (dismissed):

qili93 (approved):
LGTM

PR #35380, opened 2021-09-01
add AsExtra to concat op

zmxdream:
### PR types

Function optimization
### PR changes

APIs
### Describe

add AsExtra to the  following attribute of concat op:
  1. use_mkldnn
  2. use_quantizer
  3. mkldnn_data_type


winter-wang (approved):
LGTM

PR #35381, opened 2021-09-01
add AsExtra to partial_sum op

zmxdream:
### PR types

Function optimization
### PR changes

 APIs
### Describe

add AsRExtra to the following attribute of partial_sum op:
 1.use_mkldnn


winter-wang (approved):
LGTM

PR #35387, opened 2021-09-02
[NVIDIA] Fused multi-head attention

zlsh80826:
### PR types

Performance optimization

### PR changes

APIs

### Describe

Add NV FMHA **APIs**. Though I use Paddle OP for functional tests temporary, we should integrate the FMHA APIs to transformer.


paddle/fluid/operators/contrib/fmha_op.cu
```diff
@@ -0,0 +1,382 @@
...
+        reinterpret_cast<void *>(const_cast<int32_t *>(d_cu_seqlen)),
+        reinterpret_cast<void *>(const_cast<int32_t *>(d_seqlen)),
+        reinterpret_cast<void *>(d_output),
+        reinterpret_cast<void *>(d_softmax_mask), gen_cuda,
+        device_ctx.stream());
+  }
+};
+
+template <typename DeviceContext, typename T>
+class FMHAGradGPUKernel : public framework::OpKernel<T> {
```
limin2021, line 333: 可不可以把FMHAGradGPUKernel和run_fmha_dgrad_fp16的实现不要放在op的.cu文件中，可以挪到.cu.h/.h中，方便其他op对它的调用。

python/paddle/fluid/tests/unittests/test_fmha_op.py
```diff
@@ -0,0 +1,86 @@
+#   Copyright (c) 2018 PaddlePaddle Authors. All Rights Reserved.
```
limin2021: Copyright 2021

paddle/fluid/operators/contrib/fmha_op.cu
zlsh80826: Implementations should not be in the header files, which will cause redundant compilations. I add the cpp interface in fmha.h, users can call the FMHA implementations through include fmha.h and correctly linking.

PR #35412, opened 2021-09-02
add backward inplace for dygraph

zhiqiu:
### PR types

New features
### PR changes

Others
### Describe

add backward inplace for dygraph

### how
In dygraph mode, the forward part is not determined, while the backward computation graph is determined after forward. So the graph optimization can be performed. And, the reference count used to guide inplace optimization can be easily get `shared_ptr` of input tensor.

Two benefits: (1) reduce d2d copy and speed up training, (2) reduce memory usage

### performance
GPT model, speed up **+2.8%**.
f5f2b2cb62e7531cf3d5f698e4de5734

### DIFF
--- a/paddle/fluid/imperative/basic_engine.cc
+++ b/paddle/fluid/imperative/basic_engine.cc
@@ -314,6 +314,17 @@ static std::shared_ptr<NameVarMap<VariableWrapper>> CallGradientHooks(
   return tmp_ins_ptr;
 }
 
+bool IsInputCanInplace(const std::shared_ptr<VariableWrapper>& var) {
+  auto* inner_var = var->MutableVar();
+  if (inner_var->IsInitialized() && inner_var->IsType<framework::LoDTensor>()) {
+    auto tensor = inner_var->GetMutable<framework::LoDTensor>();
+    if (tensor->IsInitialized()) {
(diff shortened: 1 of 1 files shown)


paddle/fluid/imperative/basic_engine.cc
```diff
@@ -314,6 +314,68 @@ static std::shared_ptr<NameVarMap<VariableWrapper>> CallGradientHooks(
...
+}
+
+static void PerformBackwardInplace(const std::string& op_type,
+                                   const NameVarMap<VariableWrapper>& ins,
+                                   NameVarMap<VariableWrapper>* outs) {
+  auto& infer_inplace =
+      paddle::framework::OpInfoMap::Instance().Get(op_type).infer_inplace_;
+
+  if (infer_inplace) {
+    auto in_to_outs = infer_inplace(true);
```
sneaxiy, line 335: Why always be `true`?

zhiqiu: Discussed offline, the parameter being `true` and `false` has no difference now.

sneaxiy (approved):
LGTM

chenwhql (approved):
LGTM

PR #35434, opened 2021-09-03
Implement FunctionTraits to support two kinds of elementwise functor

Xreki:
### PR types

Function optimization

### PR changes

OPs

### Describe

rt


Xreki:
本PR的修改在https://github.com/PaddlePaddle/Paddle/pull/35487 中一起review，故关闭该PR。

PR #35445, opened 2021-09-03
[RC22] Fix linear with matmul_op replace

Ray2020BD:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix paddle.nn.linear by replace matmul op with matmul_v2 op


jeff41404 (approved):
lgtm

PR #35453, opened 2021-09-03
[hybrid] check pipeline persist var which changed in forward and used in backward

wangxicoding:
### PR types

Others
### PR changes

Others
### Describe

The pipeline requires multiple forward calculations before backward, so when the persistable var is changed in the forward, it may cause errors in the backward calculation who using this persistable var. However, some backward op don't need this var(NoNeedBufferVars), there will be no error at this time.
This PR add the check of these persistable vars which changed in forward and used in backward.


sandyhouse (approved):
LGTM

Issue #34723, opened 2021-08-09
[develop branch] - Deploying ppyolo, solov2,fcos models will raise MKLDNN error


OliverLPH:
**System information**
-PaddlePaddle version: https://github.com/PaddlePaddle/Paddle/commit/22c4c189ee5d6e09c3b56f908411b2913ab57c80, https://github.com/PaddlePaddle/Paddle/commit/41c4f7238f7ca6bf35cd6741762831e3e619a087
-CPU: Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz
-GPU: None
-OS Platform: Ubuntu1604(image: paddlepaddle/paddle:latest-dev-cuda10.1-cudnn7-gcc82)
-Python version: python 3.7
-Cmake orders
-C++version.txt
-API information
Note: You can get most of the information by running [summary_env.py](https://github.com/PaddlePaddle/Paddle/blob/develop/tools/summary_env.py). 
**To Reproduce**
cmake commands
```shell
cmake .. -DPYTHON_EXECUTABLE:FILEPATH=/usr/local/bin/python3.7 -DPYTHON_INCLUDE_DIR:PATH=/usr/local/include/python3.7m -DPYTHON_LIBRARIES:FILEPATH=/usr/local/lib/libpython3.7m.so -DWITH_FLUID_ONLY=ON -DWITH_GPU=ON -DWITH_TESTING=OFF -DCMAKE_BUILD_TYPE=Release -DON_INFER=ON -DWITH_INFERENCE_API_TEST=OFF -DWITH_MKL=ON -DWITH_MKLDNN=ON -DWITH_AVX=ON -DWITH_TENSORRT=ON -DWITH_PYTHON=ON -DCUDA_ARCH_NAME=All -DWITH_DISTRIBUTE=OFF -DWITH_STRIP=ON
```

**Describe your current behavior**
current error
<img width="819" alt="图片" src="https://user-images.githubusercontent.com/14963836/128683522-e426289d-904a-4415-a02a-0d091a0ec302.png">

**Code to reproduce the issue**
Run test with PaddleDetection codes

```
# clone PaddleDetection
git clone https://github.com/PaddlePaddle/PaddleDetection.git 
cd PaddleDetection
git checkout develop
python -m pip install -r requirments.txt

# download and export ppyolo_r50vd
[...]

PR #35455, opened 2021-09-03
Fix for reshape2 oneDNN op

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix for reshape2 oneDNN op which was causing errors in fcos model mentioned in #34723

### DIFF
--- a/paddle/fluid/operators/mkldnn/reshape_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/reshape_mkldnn_op.cc
@@ -22,6 +22,24 @@ using paddle::framework::LoDTensor;
 using platform::to_void_cast;
 using platform::GetMKLDNNFormat;
 
+inline std::vector<int> extract_shape(
+    const std::vector<const Tensor*>& list_new_shape_tensor) {
+  std::vector<int> vec_new_shape;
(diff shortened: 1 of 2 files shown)


jczaja (dismissed):
LGTM

jczaja:
@piotrekobiIntel  Please review

lidanqing-vv (dismissed):
LGTM

ghost:
I would suggest using a range-based for loop instead of a classic one like this:
for (const auto& tensor : list_new_shape_tensor) 

Thanks to this we do not have to copy the tensor pointer to a new variable, which can slightly improve performance.
It also saves one line of code and makes it clear that we don't really need the index "i" for anything.

Thanks to the fact that we know the size of the vec_new_shape vector,
we can reserve the memory that it needs right after defining it like this:
vec_new_shape.reserve(list_new_shape_tensor.size());

That way there is no time wasted on allocation every iteration.

After using reserve(), it also would be better to use emplace_back() instead of push_back.

Also, I'm not sure if it is necessary to declare the function as inline, as it is not a very small function.
Chances are that the compiler simply ignores the inline keyword.

jczaja (approved):

arlesniak (approved):
LGTM

PR #35471, opened 2021-09-05
operators/flatten_op.cc add AsExtra

dyning:
### PR types
Function optimization 

### PR changes
OPs

### Describe
add AsExtra


winter-wang (approved):
LGTM

PR #35474, opened 2021-09-05
add asExtra for nce op

pangyoki:
### PR types

Function optimization

### PR changes

OPs

### Describe

add asExtra for nce op


winter-wang (approved):
LGTM

PR #35480, opened 2021-09-06
fix the bug of layer_norm  when batch_size=1

zkh2016:
### PR types

Bug fixes
### PR changes

OPs
### Describe

The bug is that access to mean and var is incorrect, and the array will be out of bounds:  the shape of mean and var is [batch_size], and the range of thread idx is 0~feature_size, so mean[idx] and var[idx] is incorrect.

When batch_size=1, the correct access is mean[0] and var[0], and a unit test with batch_size=1 is added.

### DIFF
--- a/paddle/fluid/operators/layer_norm_kernel.cu.h
+++ b/paddle/fluid/operators/layer_norm_kernel.cu.h
@@ -705,7 +705,7 @@ __global__ void LayerNormBackwardWhenBatchSizeIsOne(
   int64_t idx = threadIdx.x + blockIdx.x * blockDim.x;
   if (idx < feature_size) {
     auto var_val =
-        static_cast<U>(real_sqrt(static_cast<float>(var[idx]) + epsilon));
+        static_cast<U>(real_sqrt(static_cast<float>(var[0]) + epsilon));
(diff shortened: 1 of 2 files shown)


xingfeng01:
LGTM

xingfeng01 (approved):

paddle/fluid/operators/layer_norm_kernel.cu.h
```diff
@@ -705,7 +705,7 @@ __global__ void LayerNormBackwardWhenBatchSizeIsOne(
   int64_t idx = threadIdx.x + blockIdx.x * blockDim.x;
   if (idx < feature_size) {
     auto var_val =
-        static_cast<U>(real_sqrt(static_cast<float>(var[idx]) + epsilon));
+        static_cast<U>(real_sqrt(static_cast<float>(var[0]) + epsilon));
```
lanxianghit, line 708: 请添加注释进行说明

zkh2016: 该kernel在调用端进行了判断，只有batch_size=1的时候会被调用。

lanxianghit (approved):

PR #35493, opened 2021-09-06
add as-extra for softplus/leaky_relu/softmax

cryoco:
### PR types

Function optimization
### PR changes

OPs
### Describe

add as-extra for softplus/leaky_relu/softmax


winter-wang (approved):
LGTM

PR #35502, opened 2021-09-06
fix bn/in/squeeze/syncbn extra

ceci3:
### PR types

Function optimization
### PR changes

OPs
### Describe

fix bn/in/squeeze/syncbn extra


winter-wang (approved):
LGTM

PR #35550, opened 2021-09-07
Added requirements_first

jakpiase:
### PR types
Others

### PR changes
Others

### Describe
Added requirements_first to fix dependency of bfloat16 package. It is needed as a seperate PR because CI-Coverage would fail without it as it resets itself to paddle/develop at the end of CI.


jakpiase:
Closing, I am aborting this idea, because I have found a better one.

PR #35569, opened 2021-09-08
add layerwise learning rate for adamw

zhaoyinglia:
### PR types

 New features
### PR changes

OPs
### Describe

add layerwise learningrate feature for adamw

### DIFF
--- a/paddle/fluid/operators/optimizers/adam_op.cc
+++ b/paddle/fluid/operators/optimizers/adam_op.cc
@@ -236,6 +236,10 @@ class AdamWOpMaker : public AdamOpMaker {
  public:
   void Make() {
     AdamOpMaker::Make();
+    AddAttr<float>("lr_ratio",
+                   "(float, default 1.0) "
+                   "layerwise learning rate decay")
+        .SetDefault(1.0f);
     AddAttr<float>("coeff",
                    "(float, default 0.01) "
                    "coeff of the weight decay")
--- a/python/paddle/optimizer/adamw.py
+++ b/python/paddle/optimizer/adamw.py
@@ -64,6 +64,10 @@ class AdamW(Adam):
         epsilon (float, optional): A small float value for numerical stability.
             The default value is 1e-08.
         weight_decay (float|Tensor, optional): The weight decay coefficient, it can be float or Tensor. The default value is 0.01.
+        lr_ratio (function|None, optional): If it is not None, 
+            the learning rate will be updated with layerwise learning rate ratio.
+            Otherwise, the learning rate is the original.
+            Default: None.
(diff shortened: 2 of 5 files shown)


paddle/fluid/operators/optimizers/adam_op.cc
```diff
@@ -236,6 +236,10 @@ class AdamWOpMaker : public AdamOpMaker {
  public:
   void Make() {
     AdamOpMaker::Make();
+    AddAttr<float>("lr_ratio",
```
JZ-LIANG, line 239: why add this argument to adam? is that adamw and adam share the same .cc file ?

in this case, adamw should have its own .cc file

python/paddle/optimizer/adamw.py
```diff
@@ -163,6 +165,9 @@ def __init__(self,
         self._apply_decay_param_fun = apply_decay_param_fun
         self._coeff = coeff
         self._lr_to_coeff = dict()
+        if lr_ratio is not None:
+            assert isinstance(lr_ratio, Callable)
+        self._lr_ratio = lr_ratio
```
JZ-LIANG, line 177: you should think about how many kernel will be affected by  "lr_ratio". 
if you only want the lr_ratio the affect gpu and cpu kernel, you should raise an Unimplement Error for xpu and npu here.

python/paddle/optimizer/adamw.py
```diff
@@ -163,6 +165,9 @@ def __init__(self,
         self._apply_decay_param_fun = apply_decay_param_fun
         self._coeff = coeff
         self._lr_to_coeff = dict()
+        if lr_ratio is not None:
```
JZ-LIANG, line 172: should add explanation for the new lr_ration argument, which should follow the explanation for "apply_decay_param_fun"

paddle/fluid/operators/optimizers/adam_op.cc
zhaoyinglia: AdamWOpMaker inherits AdamOpMaker, and they use the same InferShape function of AdamOp.
In this case, 'lr_ratio' has no effect on Adam.

python/paddle/optimizer/adamw.py
zhaoyinglia: done.

python/paddle/optimizer/adamw.py
zhaoyinglia: done.

PangHua (dismissed):

PangHua (approved):

JZ-LIANG (approved):
LGTM

PR #31752, opened 2021-03-19
gradient merge: to support save

JZ-LIANG:
### PR types
Bug fixes 

### PR changes
OPs

### Describe
fixed bug for fleet gradient merge to support save api

PR #35578, opened 2021-09-08
Set gradient_merge_cond persistable to false

ZHUI:
### PR types

Bug fixes
### PR changes

Others
### Describe

1. gradient_merge_cond is not need to save as persistable.
2. save op is not supported to save bool type

Old PR tries to fix this issue: https://github.com/PaddlePaddle/Paddle/pull/31752


wangxicoding (approved):
LGTM

PR #35585, opened 2021-09-08
add dist_attr for dist op and var

zhaoyinglia:
### PR types

New features
### PR changes

OPs
### Describe

add dist_attr for newly added dist op and var.


python/paddle/distributed/auto_parallel/operators/common.py
```diff
@@ -114,3 +114,44 @@ def find_best_compatible_distributed_operator_impl(name, op_dist_attr,
         best_compatible_impl, idx = None, -1
 
     return best_compatible_impl, idx
+
+
+def set_distributed_attr_for_var(src_op_dist_attr, var, src_var):
```
JZ-LIANG: function naming could be more self-explained.
here you actually copied all the dist_attr info from src to dst var, it is more close to  a "copy" than a "setting" semantically. 
use something like "copy_distributed_attr_for_var" or check if a similar function has already been implemented in context.py

python/paddle/distributed/auto_parallel/operators/common.py
```diff
@@ -114,3 +114,44 @@ def find_best_compatible_distributed_operator_impl(name, op_dist_attr,
...
+    auto_paralle_context = src_op_dist_attr.get_owner_context()
+    dist_attr = copy.deepcopy(
+        auto_paralle_context.get_tensor_distributed_attr_for_program(src_var))
+    dist_attr._owner_tensor = var
+    dist_attr._owner_context = auto_paralle_context.get_tensor_distributed_attr_for_program(
+        src_var)._owner_context
+    auto_paralle_context.set_tensor_distributed_attr_for_program(var, dist_attr)
+
+
+def set_distributed_attr_for_dist_op(dist_op, dst_block, src_op_dist_attr):
```
JZ-LIANG: the same as above one, but here you just copy partial of the dist_attr info for the src

python/paddle/fluid/tests/unittests/test_auto_parallel_partitioner.py
```diff
@@ -135,6 +135,110 @@ def initialization_check(mode, dist_context, dist_startup_prog,
...
+            equal = check_equal_var_dist_attr(serial_out_dist_attr,
+                                              out_dist_attr)
+
+        # check op's dist_attr 
+        equal = check_equal_dist_op_attr(dist_context, dist_main_prog,
+                                         serial_op, dist_ops, dist_op_idx[i])
+
+    return equal
+
+
```
JZ-LIANG, line 241: you should add a new test for all case: checking all vars and  ops have dist attr

python/paddle/distributed/auto_parallel/operators/common.py
zhaoyinglia: done.

python/paddle/distributed/auto_parallel/operators/common.py
zhaoyinglia: done.

python/paddle/fluid/tests/unittests/test_auto_parallel_partitioner.py
zhaoyinglia: done.

JZ-LIANG (approved):
LGTM

PR #35588, opened 2021-09-08
[HybridParallel]Support fp16 in flatten op to avoid error in recompute

ForFishes:
### PR types

Bug fixes 
### PR changes

Others
### Describe

[HybridParallel]Support fp16 in flatten op to avoid error in recompute


JZ-LIANG (approved):
LGTM

PR #35590, opened 2021-09-08
Change depth as 1 when cloning benchmark code

xiegegege:
### PR types

Others
### PR changes

Others
### Describe

Changing depth as 1 when cloning benchmark code, to accelerate code download.


OliverLPH (approved):
LGTM

PR #35595, opened 2021-09-08
Fix scatter and gather bug

sneaxiy:
### PR types
Bug fixes

### PR changes
OPs

### Describe
- `gather` should support empty `index`.
- `scatter` should support larger shapes. Use `int64_t` instead of `int` to avoid overflow.


ForFishes (approved):
LGTM

Xreki (approved):
LGTM for op benchmark ci

PR #35597, opened 2021-09-08
change trt_tile_op half diff and add some func for CE

fengxiaoshuai:
### PR types
Others

### PR changes
Others

### Describe
change trt_tile_op half diff and add some func for CE


shangzhizhou (approved):
LGTM

PR #35599, opened 2021-09-08
Update scales when variable is unsigned

wozna:
### PR types

Bug fixes
### PR changes

Others 
### Describe

This PR:
- fixes updating scales in the quant-aware training quantization process. In this process we tag vars which are after relu op, to `unsigned`. But if this var was before tagged as `signed`, it means that the scales for this var were already computed for signed data. To resolve that, scales for this var must be multiplied by 2 to fill the entire range of uint8. This change should improve accuracy in QAT models.
- changes batch size in `inference_quant2_int8_image_classification_test` because the above fix improve acc in resnet50 test and acc comparison QAT with PTQ  int8 failed. 

Graph showing the change:
Scales


jczaja (approved):
LGTM

wozna:
Please wait with the merging until I finish functional tests for all models.

lidanqing-vv (approved):
LGTM. It solved QAT LSTM accuracy diff issue. Thanks

PR #35661, opened 2021-09-10
Add paddle.device.cuda.get_device_properties

Yanxing-Shi:
### PR types

New features

### PR changes

APIs

### Describe

Add paddle.device.cuda.get_device_properties


paddle/fluid/platform/gpu_info.cc
```diff
@@ -297,6 +314,64 @@ std::vector<int> GetSelectedDevices() {
   return devices;
 }
 
+#ifdef PADDLE_WITH_CUDA
+cudaDeviceProp *GetDeviceProperties(int id) {
+  std::call_once(init_flag, [&] {
```
sneaxiy: Too many duplicate codes. I think we can unify the codes of CUDA and HIP by some kinds of techniques shown in `paddle/fluid/platform/type_defs.h`.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -39,6 +44,18 @@ DECLARE_uint64(gpu_memory_limit_mb);
 
 constexpr static float fraction_reserve_gpu_memory = 0.05f;
 
+static std::once_flag init_flag;
+static std::deque<std::once_flag> device_flags;
+static int gpu_num = -1;
```
sneaxiy: I do not think you should add this global variable `gpu_num`. As I mentioned offline yesterday, you do not need to record the `gpu_num`. You can use `device_props.size()` below.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -39,6 +44,18 @@ DECLARE_uint64(gpu_memory_limit_mb);
...
+
+#ifdef PADDLE_WITH_CUDA
+static std::vector<cudaDeviceProp> device_props;
```
sneaxiy: `device_props` -> `g_device_props`. Here, the prefix `g_` means global variables.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -39,6 +44,18 @@ DECLARE_uint64(gpu_memory_limit_mb);
 
 constexpr static float fraction_reserve_gpu_memory = 0.05f;
 
+static std::once_flag init_flag;
+static std::deque<std::once_flag> device_flags;
```
sneaxiy: Too fancy codes to use `std::deque::resize` for non-copyable and non-moveable types. 

Although `std::deque::resize` can accept the non-copyable and non-moveable types like `std::once_flag` (note that `std::once_flag` is neither copyable nor moveable, see https://en.cppreference.com/w/cpp/thread/once_flag), many of us should think that `std::deque` should accept the moveable types at least. I do not think it is a good idea to use `std::deque<std::once_flag>`.

Alternatively, you use use `std::vector<std::unique_ptr<std::once_flag>>`.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -297,6 +314,64 @@ std::vector<int> GetSelectedDevices() {
...
+      platform::errors::OutOfRange(
+          "The device id %d exceeds out of range [0, the number of "
+          "devices %d on this machine), Because the gpu id should be "
+          "smaller than the number of gpus. Please input appropriate "
+          "device id again!",
+          id, gpu_num));
+
+  std::call_once(device_flags[id], [&] {
+    cudaDeviceProp device_prop;
+    PADDLE_ENFORCE_CUDA_SUCCESS(cudaGetDeviceProperties(&device_prop, id));
```
sneaxiy: You need not to use the temporary variable `device_prop`. You can write the more direct codes:

```C++
PADDLE_ENFORCE_CUDA_SUCCESS(cudaGetDeviceProperties(&device_props[id], id));
```

paddle/fluid/platform/gpu_info.h
```diff
@@ -67,6 +67,15 @@ dim3 GetGpuMaxGridDimSize(int);
 //! Get a list of device ids from environment variable or use all.
 std::vector<int> GetSelectedDevices();
 
+//! Get the properties of device.
+#ifdef PADDLE_WITH_CUDA
+cudaDeviceProp *GetDeviceProperties(int id);
```
sneaxiy: I have emphasized the Google coding style for a long time. Use const reference for unchanged variables, and use non-const pointer for changeable variables.

You should not return non-const pointer here, because one should think that returning non-const pointer means the users may change the returned value. However, I do not think that `cudaDeviceProp` should be changed.

paddle/fluid/pybind/pybind.cc
```diff
@@ -2267,6 +2267,71 @@ All parameter, weight, gradient are variables in Paddle.
 #endif
 #endif
 
+#ifdef PADDLE_WITH_CUDA
+  m.def("get_device_properties",
+        [](int id) -> cudaDeviceProp * {
```
sneaxiy: Duplicate codes for CUDA and HIP.

python/paddle/device/cuda/__init__.py
```diff
@@ -24,6 +24,7 @@
     'synchronize',
     'device_count',
     'empty_cache',
+    'get_device_properties'
```
sneaxiy: Add comma in the end.

python/paddle/device/cuda/__init__.py
```diff
@@ -141,3 +141,37 @@ def empty_cache():
...
+    device_id = -1
+    if device is not None:
+        if isinstance(device, int):
+            device_id = device
+        elif isinstance(device, core.CUDAPlace):
+            device_id = device.get_device_id()
+        else:
+            raise ValueError("device type must be int or paddle.CUDAPlace")
+
+    return core.get_device_properties(device_id)
```
sneaxiy, line 273: Not friendly error message. If the users install the CPU version and he/she calls `paddle.device.cuda.get_device_properties()`, he/she would get the error message like `"core" has no attribute "get_device_properties"`. It is not friendly to users.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -22,6 +22,11 @@ limitations under the License. */
 #else
 #include "paddle/fluid/platform/dynload/cudnn.h"
 #endif
+
+#include <deque>
```
sneaxiy: Remove useless headers.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -39,6 +44,10 @@ DECLARE_uint64(gpu_memory_limit_mb);
 
 constexpr static float fraction_reserve_gpu_memory = 0.05f;
 
+static std::once_flag g_init_flag;
+static std::vector<std::unique_ptr<std::once_flag>> g_device_flags;
```
sneaxiy: The variable names `g_init_flag` and `g_device_flags` are not readable. Nobody knows what they mean if he/she does not read the following codes.

Try:
- `g_init_flag` -> `g_device_props_size_init_flag`
- `g_device_flags` -> `g_device_prop_init_flags`

paddle/fluid/platform/gpu_info.cc
```diff
@@ -297,6 +306,43 @@ std::vector<int> GetSelectedDevices() {
   return devices;
 }
 
+const gpuDeviceProp *GetDeviceProperties(int id) {
```
sneaxiy: Have you read my comment https://github.com/PaddlePaddle/Paddle/pull/35661#discussion_r708798805 carefully? I mean **const reference**.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -297,6 +306,43 @@ std::vector<int> GetSelectedDevices() {
...
+  int gpu_num = 0;
```
sneaxiy: No need to put this variable outside the `std::call_once`. I mean this line can be put inside the following `std::call_once`.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -297,6 +306,43 @@ std::vector<int> GetSelectedDevices() {
...
+  std::call_once(g_init_flag, [&] {
+    gpu_num = platform::GetCUDADeviceCount();
+    g_device_flags.resize(gpu_num);
+    g_device_props.resize(gpu_num);
+    for (int i = 0; i < gpu_num; ++i) {
+      g_device_flags[i] = std::unique_ptr<std::once_flag>(new std::once_flag());
```
sneaxiy: `g_device_flags[i] = std::make_unique<std::once_flag>()` would be simpler.

paddle/fluid/platform/gpu_info.cc
```diff
@@ -297,6 +306,43 @@ std::vector<int> GetSelectedDevices() {
...
+  if (id == -1) {
+    id = platform::GetCurrentDeviceId();
+  }
+
+  if (id < 0 || id >= static_cast<int>(g_device_props.size())) {
+    PADDLE_THROW(platform::errors::OutOfRange(
+        "The device id: %d exceeds out of range [0, the number of devices: %d "
+        "on this machine). Because the device id should be greater than or "
+        "equal to zero and smaller than the number of gpus. Please input "
+        "appropriate device again!",
```
sneaxiy, line 326: Your error message would be like:

```
The device id: 8 exceeds out of range [0, the number of devices: 4 on this machine).
```

This is: 
- in the wrong format in math.
- wrong grammar in English. The word `exceed` has the meaning of `out of range`!


I prefer that the error message would be like:

```
The device id 8 is out of range [0, 4), where 4 is the number of devices on this machine.
```

python/paddle/device/cuda/__init__.py
```diff
@@ -14,7 +14,9 @@
 
 import paddle
 from paddle.fluid import core
+from paddle.fluid import framework
 from paddle.fluid.wrapped_decorator import signature_safe_contextmanager
+from paddle.device import is_compiled_with_cuda
```
sneaxiy: I guess this line would cause circular dependencies. Are you sure it is right? If you have checked, please tell me.

My thought on why there are circular dependencies: 
- When users import `paddle.device.cuda`, Python would import `python/paddle/device/cuda/__init__.py` and reach this line `from paddle.device import is_compiled_with_cuda`. 
- Then, Python would try to import `paddle.device`. Inside `python/paddle/device/__init__.py`, there is the code `from . import cuda`. So, Python would import `python/paddle/device/cuda/__init__.py` again. That is why I think the circular dependencies occur.

python/paddle/device/cuda/__init__.py
```diff
@@ -204,3 +207,63 @@ def stream_guard(stream):
...
+            import paddle
+            paddle.device.cuda.get_device_properties()
+            paddle.device.cuda.get_device_properties(0)
+            paddle.device.cuda.get_device_properties('gpu:0')
+            paddle.device.cuda.get_device_properties(paddle.CUDAPlace(0))
+
+    '''
+
+    place = framework._current_expected_place()
+    if not isinstance(place, core.CUDAPlace) or not is_compiled_with_cuda():
```
sneaxiy: I do not think that you should check `isinstance(place, core.CUDAPlace)`.

The `framework._current_expected_place()` may return `core.CPUPlace()`. Even if users install the GPU version Paddle, they can use CPU instead of GPU.

python/paddle/device/cuda/__init__.py
```diff
@@ -204,3 +207,63 @@ def stream_guard(stream):
...
+        raise ValueError(
+            "The current device: {} is not expected. Because paddle.device.cuda."
+            "get_device_properties only support cuda device Please change device"
+            "and input device again!".format(place))
```
sneaxiy: I am confused with this sentence. What are the differences between `device` and `input device`?

python/paddle/device/cuda/__init__.py
```diff
@@ -204,3 +205,69 @@ def stream_guard(stream):
...
         finally:
             stream = _set_current_stream(pre_stream)
+
+
+def get_device_properties(device=None):
+    '''
+    Return the properties of given CUDA device.
+
+    Args:
+        device(paddle.CUDAPlace() or int or str): The device, the ID of the device
```
sneaxiy: `paddle.CUDAPlace()` -> `paddle.CUDAPlace`

python/paddle/device/cuda/__init__.py
```diff
@@ -204,3 +205,69 @@ def stream_guard(stream):
...
+    Return the properties of given CUDA device.
+
+    Args:
+        device(paddle.CUDAPlace() or int or str): The device, the ID of the device 
+            or the string name of device like 'gpu:x' which to get the properties of 
+            the device from. If device is None, the device is the current device. 
+            Default: None.
+
+    Returns:
+        _CudaDeviceProperties: the properties of the device which include ASCII string
```
sneaxiy: `_CudaDeviceProperties` -> `_gpuDeviceProperties`? I guess you can fix this in the next PR with `test=document_fix`.。。

python/paddle/device/cuda/__init__.py
```diff
@@ -204,3 +205,69 @@ def stream_guard(stream):
...
+            # _CudaDeviceProperties(name='A100-SXM4-40GB', major=8, minor=0, total_memory=40536MB, multi_processor_count=108)
+
+            paddle.device.cuda.get_device_properties(paddle.CUDAPlace(0))
+            # _CudaDeviceProperties(name='A100-SXM4-40GB', major=8, minor=0, total_memory=40536MB, multi_processor_count=108)
+
+    '''
+
+    if not core.is_compiled_with_cuda():
+        raise ValueError(
+            "The current device {} is not expected. Because paddle.device.cuda."
```
sneaxiy: It seems that the error message is not right. It should be something like `The API paddle.device.cuda.get_device_properties() is not supported in CPU-only PaddlePaddle. Please reinstall PaddlePaddle with GPU support to call this API`.

python/paddle/device/cuda/__init__.py
```diff
@@ -204,3 +205,69 @@ def stream_guard(stream):
...
+            "get_device_properties only support cuda device. Please change device"
+            "and input device again!".format(device))
+
+    device_id = -1
+
+    if device is not None:
```
sneaxiy, line 252: What if `device is None`?

python/paddle/device/cuda/__init__.py
```diff
@@ -204,3 +205,69 @@ def stream_guard(stream):
...
+        elif isinstance(device, core.CUDAPlace):
+            device_id = device.get_device_id()
+        elif isinstance(device, str):
+            if device.startswith('gpu:'):
+                device_id = int(device[4:])
+            else:
+                raise ValueError(
+                    "The current string {} is not expected. Because paddle.device."
+                    "cuda.get_device_properties only support string which is like 'gpu:x'. "
+                    "Please input appropriat string again!".format(device))
```
sneaxiy: Wrong word `appropriat`.

sneaxiy (approved):
LGTM.

zhhsplendid (approved):
LGTM

TCChenlong (approved):

lanxianghit (approved):

PR #35678, opened 2021-09-13
no necessary to add scale_op to stop inplace op from being pruned. fix prune instead.

hp03:
### PR types
Bug fixes

### PR changes
Others

### Describe
when the _prune_with_input method (python/paddle/fluid/framework.py) is implemented, there is no static graph and no inplace op. after inplace op is introduced, some graph will not save, cause the inplace op will be pruned. to stop this, a scale_op is added to each output var. this is a trick, it works sometime, but not always, cause scale_op doesn't support all the datatypes. another problem introduced by inplace op is when a var is both feedvar and fetchvar, prune will also prune it. 

this PR just fix the prune method to support inplace op, so there will be no necessary to add scale_op after output var. this PR just remove the tricks.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -5031,12 +5031,6 @@ class Program(object):
                         "All targets of Program._prune_with_input() can only be "
                         "Variable or Operator, but received %s." % type(t))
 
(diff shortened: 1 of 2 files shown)


Superjomn (approved):
LGTM

PR #35700, opened 2021-09-13
[NPU] fix overflow bug for elementwise_div npu op

pangyoki:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Use `Div` CANN OP to achieve `1/y` instead of `Power` CANN OP. Because `Power` will cause precision overflow, that is, `float_status` will be set to 1.


zhiqiu (approved):
LGTM

PR #35718, opened 2021-09-14
[HETERPS]fix make error when pscore equals on

yaoxuefeng6:
### PR types

Others
### PR changes

Others
### Describe

fix make error when pscore=on


danleifeng (approved):
LGTM

PR #35732, opened 2021-09-14
refine auto_growth allocator

zhiqiu:
### PR types

Performance optimization
### PR changes

Others
### Describe

refine auto_growth allocator

CUDA does address alignment itself, so in most case we do not need to manually perform address alignment.
refer: https://stackoverflow.com/questions/14082964/cuda-alignment-256bytes-seriously

The problem is that, the original implementation may alloc more 256 bytes due to AlignedAllocator and the 256 bytes may put into `free blocks`. Which may introduce more memory fragmentation and increase the time to find fit block when allocing new allocation.

- before
image

- after
image

The max number of free blocks in standalone_executor_test(single thread, no gc version, for consistance) decreases: 200->185

### DIFF
--- a/paddle/fluid/memory/allocation/aligned_allocator.cc
+++ b/paddle/fluid/memory/allocation/aligned_allocator.cc
@@ -20,6 +20,7 @@ namespace paddle {
 namespace memory {
 namespace allocation {
 
+// For memory address alignment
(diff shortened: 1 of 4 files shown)


sneaxiy (approved):
LGTM

wanghuancoder (approved):
LGTM

PR #35733, opened 2021-09-14
fix dim check of class center sample

GuoxiaWang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix dim check of class center sample


sandyhouse (approved):
LGTM

PR #35735, opened 2021-09-14
[Auto Parallel] Generalization for Partition and Completion

JZ-LIANG:
### PR types
Function optimization

### PR changes
OPs

### Describe
- add default dist op to allow each op to handle all communications it need by its dist op, therefore we could remove the "Data Parallel"(grad sync comm) & "Tensor Parallel"(parallel matmul allreduce comm) logic in Auto Parallelism.
- make a mapping for op in forward segment to corresponding op in backward & update segment, which will make the auto completion for backward & update more robust.
- revise and generalize the auto completion logic for backward, update and startup program, which allows later module to access distribute attribute information in this part.
- unify the forward and backward API of dist op.
- revise the unitest for parallelizer to truly guard the logic of communication initialization.
- revise the unitest for reshard to allow more general communication op order.
- bugfix for namemapping bug in dist embedding op


python/paddle/distributed/auto_parallel/context.py
```diff
@@ -429,3 +429,77 @@ def _get_data_parallel_info(self):
     def _get_model_parallel_info(self):
         # This function is a hard code, and will be obsoleted in the future
         return self._model_parallel_axis, self._process_mesh
+
+
+class DistOpContext:
```
aoyulong: Please consider to merge the function of DistOpContext into the DistributedContext. At the beginning, we can keep our classes as less as possible. In the future, we can split a large class into several classes to represent different abstract concepts.

python/paddle/distributed/auto_parallel/operators/dist_default.py
```diff
@@ -0,0 +1,96 @@
...
+from ..utils import compute_compatible_and_update_dim_mapping
+from paddle.fluid import core, unique_name
+from paddle.fluid.framework import in_dygraph_mode
+from paddle.fluid.framework import Program, Parameter, Variable, program_guard
+from paddle.fluid.data_feeder import check_variable_and_dtype, check_dtype
+from ..process import new_process_group
+from ..utils import _get_comm_group
+
+
+class DistributedDefault(DistributedOperator):
```
aoyulong, line 35: Please rename DistributedDefault to DistributedDefaultOperator or DistributedDefaultOp since DistributedDefault is a little ambiguous.

python/paddle/distributed/auto_parallel/operators/dist_default.py
```diff
@@ -0,0 +1,96 @@
...
+    def is_output_compatible(self, op_dist_attr):
+        raise NotImplementedError("Please Implement this method.")
+
+    def update_dims_mapping(self, op_dist_attr):
+        raise NotImplementedError("Please Implement this method.")
+
+    @staticmethod
+    def forward(ctx, *args, **kwargs):
+
+        dst_block = ctx.get_dst_main_program().global_block()
```
aoyulong: To get dst_block, we must first call get_dst_main_program. Can we directly to store the dst_block in the context? If we can, the dist op developers don't need to know the dst_main_program concept.

python/paddle/distributed/auto_parallel/operators/dist_default.py
```diff
@@ -0,0 +1,96 @@
...
+        src_op = ctx.get_cur_src_op()
+        varname_mapping = ctx.get_varname_mapping()
```
aoyulong: Varname_mapping is not used in this function.

python/paddle/distributed/auto_parallel/operators/dist_default.py
```diff
@@ -0,0 +1,96 @@
...
+            assert output_name in kwargs, "input [{}] is not given".format(
+                output_name)
+            assert len(kwargs[output_name]) == len(
+                src_op.desc.output(output_name)
+            ), "number of tensor for input [{}] is not match".format(
+                output_name)
+
+        # replicate op in dist program
+        dist_op_desc = dst_block.desc.append_op()
+        dist_op_desc.copy_from(src_op.desc)
```
aoyulong, line 91: Please check the id of the new dist_op after copy_from src_op.desc. And think whether the relationship of the id of the new dist_op and the id of the src_op is OK.

python/paddle/distributed/auto_parallel/operators/dist_embedding.py
```diff
@@ -102,118 +102,107 @@ def update_dims_mapping(self, op_dist_attr):
...
-            return static_handle
+            relative_idx = rank_id % num_partition
+
+        per_part_size = Weight_var.shape[0]
+        relative_idx = relative_idx * per_part_size
+
+        # TODO caculate ring id 
+        model_parallel_axis, process_mesh = op_dist_attr.get_owner_context(
+        )._get_model_parallel_info()
+        group_ranks = _get_comm_group(process_mesh.process_group,
```
aoyulong: Can the  _get_comm_group function become a member function of ProcessMesh？Since the first arguments are related to the same ProcessMesh object.

Aurelius84 (dismissed):
LGTM for modification in `append_backward`

XieYunshen (approved):
LGTM for `list(REMOVE_ITEM TEST_OPS test_auto_parallel_parallelizer)`
单测修改为多卡，只有GPU环境下需要执行

TCChenlong (approved):

python/paddle/distributed/auto_parallel/operators/dist_default.py
JZ-LIANG: recorded, will create a new bugfix pr to fixe it.

python/paddle/distributed/auto_parallel/operators/dist_default.py
JZ-LIANG: yes, it will be a problem that the serial op and its correspond dist op share the same id, it is ok for current cases. but we need redesign here later.

XiaoguangHu01 (approved):
LG API

raindrops2sea (approved):

Issue #35719, opened 2021-09-14
Deploying swin_transformer will raise reshape2 mkldnn error.


baoachun:
![图片](https://user-images.githubusercontent.com/22114318/133208275-99503c1c-254e-439b-8902-12129c4c6baf.png)

![图片](https://user-images.githubusercontent.com/22114318/133208316-5a49ce5b-7181-44bb-b1e2-b26faeeb5fcf.png)

lidanqing-vv:
@baoachun 
Hi, #35719, #35717 and #34723 three issues will be take care be @jakpiase

jakpiase:
Hi @baoachun, could you please advise me how can I run swin_transformer?

jakpiase:
I have reproduced this error, working on it now

jakpiase:
Hi @baoachun my fix to this issue has been merged. Could you please tell me if this issue still exist?

baoachun:
@jakpiase 
This problem has been solved.

PR #35740, opened 2021-09-14
Fix for swin_transformer(matmul+transpose+reshape)

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix for #35719. Problem was caused because matmul+transpose+reshape was not handling dim inference which is used by reshape op. Other problem was caused because we weren't handling 6 dim tensors, but now we do.

### DIFF
--- a/paddle/fluid/operators/matmul_op.cc
+++ b/paddle/fluid/operators/matmul_op.cc
@@ -655,6 +655,24 @@ class MatMulOp : public framework::OperatorWithKernel {
           platform::errors::InvalidArgument("reshape_out supported rank is 3, "
                                             "received %d",
                                             reshape_out_size));
+
+      auto it = std::find(reshape_out.begin(), reshape_out.end(), -1);
+
+      // if "-1" is present then one of reshape dims must be infered
+      if (it != reshape_out.end()) {
+        int index = std::distance(reshape_out.begin(), it);
+
+        auto ddim_out_vec = framework::vectorize(ddim_out);
+
+        int ddim_out_product =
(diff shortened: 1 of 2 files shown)


jczaja (commented):
LGTM

jakpiase:
@piotrekobiIntel @tsocha Could you please review and maybe advise me if that operation of infering missing dim can be done more concisely?

lidanqing-vv (approved):
LGTM

ghost:
I would suggest using a simple for loop to check if "-1" is present. That way you don't need to iterate over the array twice to find presence and later distance. I don't know how often "-1" will be present, but if it's often enough it might even be better to calculate the product in that for loop instead of iterating over the array once again.

paddle/fluid/operators/matmul_op.cc
```diff
@@ -655,6 +655,24 @@ class MatMulOp : public framework::OperatorWithKernel {
...
+        auto ddim_out_vec = framework::vectorize(ddim_out);
+
+        int ddim_out_product =
+            std::accumulate(ddim_out_vec.begin(), ddim_out_vec.end(), 1,
+                            std::multiplies<int>());
+        int reshape_out_product = std::accumulate(
+            reshape_out.begin(), reshape_out.end(), -1, std::multiplies<int>());
+
+        reshape_out[index] = ddim_out_product / reshape_out_product;
+      }
```
tsocha, line 674: You modify a reshape_out dim which is already pointed by iterator.
There is no need to calculate an index via std::distance, just use 'it'.

I wonder if it is worth to check shapes compatibility:
'''
ddim_out_product % reshape_out_product == 0
'''

PR #35769, opened 2021-09-15
fix py api for paddle.inference.contrib.utils

shangzhizhou:
### PR types
Others

### PR changes
APIs

### Describe
fix python api "paddle.inference.contrib.utils"


jiweibo (approved):
LGTM

PR #35774, opened 2021-09-15
Basic PR on Cost Model

zhhsplendid:
### PR types

New features
### PR changes

APIs
### Describe

Add basic Cost Model, it uses executor to run program and profile it to get op time.

This is an early basic version, we will add more functions in the future.


paddle/fluid/framework/ir/cost_model.cc
```diff
@@ -0,0 +1,250 @@
...
+  Executor executor(place);
+  Scope scope;
+  executor.Run(startup_program, &scope, /*block_id = */ 0);
+
+  // TODO(zhhsplendid): handle the case that Profiler is already enabled
+  SetTracerOption(platform::TracerOption::kAllOpDetail);
+  EnableProfiler(profiler_state);
+  executor.Run(main_program, &scope, /*block_id = */ 0);
+
+  std::vector<std::vector<Event>>* time_events =
```
thisjiang: I think the best idea is using `unique_ptr` or `shared_ptr` rather than raw pointer.

paddle/fluid/framework/ir/cost_model.cc
```diff
@@ -0,0 +1,250 @@
...
+  while (event_index < main_thread_events.size()) {
+    if (main_thread_events[event_index].name() == "_start_profiler_") {
+      start_profiler_idx = event_index;
+    } else if (main_thread_events[event_index].name() == "_stop_profiler_") {
+      stop_profiler_idx = event_index;
+      break;
+    }
+    ++event_index;
+  }
+  if (start_profiler_idx != 0 && stop_profiler_idx != 0) {
```
thisjiang: `start_profiler_idx != 0`? Why can not `main_thread_events[0].name() == "_start_profiler_"`, I think it is possible.

zhhsplendid: There is a little background, profiler would have some initialize events which aren't "_start_profiler_", but to support the `start_profiler_idx == 0` case, I changed the related code. I compare it with -1;

paddle/fluid/framework/ir/cost_model.cc
zhhsplendid: Fine to me, change done.

paddle/fluid/framework/ir/cost_model_test.cc
```diff
@@ -0,0 +1,209 @@
...
+      cost_model.ProfileMeasure(empty_program, empty_program, "cpu", {"time"});
+  EXPECT_EQ(cost_data.GetWholeTimeMs(), 0);
+}
+
+TEST(CostModelTest, TestProfileMeasure_Program) {
+  CostModel cost_model;
+  ProgramDesc program = CreateTestProgram();
+  ProgramDesc empty_program;
+  CostData cost_data =
+      cost_model.ProfileMeasure(program, empty_program, "cpu", {"time"});
```
thisjiang, line 126: I didn't find "gpu" version test, should add it for coverage CI?

paddle/fluid/framework/ir/cost_model_test.cc
```diff
@@ -0,0 +1,209 @@
...
+               const platform::Place &place) const override {
+    // Fake RunImpl, for test only
+    Variable *var = scope.FindVar("X");
+    if (var != nullptr) {
+      LoDTensor *tensor = var->GetMutable<LoDTensor>();
+      tensor->mutable_data<float>(place);
+    }
+    int count = 0;
+    while (count <= 1000) {
+      ++count;
```
thisjiang, line 54: Does this useless loop necessary?

paddle/fluid/framework/ir/cost_model.cc
```diff
@@ -0,0 +1,256 @@
...
+    if (main_thread_events[event_index].name() == "_start_profiler_") {
+      start_profiler_idx = event_index;
+    } else if (main_thread_events[event_index].name() == "_stop_profiler_") {
+      stop_profiler_idx = event_index;
+      break;
+    }
+    ++event_index;
+  }
+  if (start_profiler_idx != -1 && stop_profiler_idx != -1) {
+    double cpu_time_ms = main_thread_events[start_profiler_idx].CpuElapsedMs(
```
thisjiang, line 138: Add `PADDLE_ENFORCE_GE(start_profiler_idx, 0)` and `PADDLE_ENFORCE_GT(stop_profiler_idx, start_profiler_idx)` here?

paddle/fluid/framework/ir/cost_model.cc
```diff
@@ -0,0 +1,256 @@
...
+using platform::MemEvent;
+
+const double CostData::NOT_MEASURED = -1;
+
+CostData::~CostData() {
+  // TODO(zhhsplendid): when we save a copy of program/graph, we should delete
+  // here.
+}
+
+double CostData::GetOpTimeMs(int op_id) const { return op_time_ms_.at(op_id); }
```
thisjiang, line 37: I think it's better to use `size_t` replace `int` because the `global_block.OpSize()` return `size_t`.

paddle/fluid/framework/ir/cost_model_test.cc
zhhsplendid: Necessary to occupy test time

paddle/fluid/framework/ir/cost_model_test.cc
zhhsplendid: You can see it on python test ^_^

paddle/fluid/framework/ir/cost_model.cc
zhhsplendid: Not good now, because we are not sure about profiler

paddle/fluid/framework/ir/cost_model.cc
zhhsplendid: A little background, we would also support graph in the future, then there will be int node id. We are not sure whether there is negative graph id

thisjiang (approved):
LGTM

phlrain (approved):

XiaoguangHu01 (approved):
LGTM

PR #35787, opened 2021-09-16
fix the memory leak for the static.auc operators

wawltor:
### PR types
Bug fixes 

### PR changes
OPs

### Describe
fix the memory leak for the static.auc


joey12300 (approved):
LGTM

PR #35791, opened 2021-09-16
add the shape check for the matmul

wawltor:
### PR types
Function optimization 

### PR changes
OPs 

### Describe
add the shape check for the matmul


joey12300 (approved):
LGTM

PR #35793, opened 2021-09-16
Initial support lazy launch for auto-parallel

bliu3650:
### PR types
New features

### PR changes
Others

### Describe
Lazy launch for starting un-balanced subprocesses on each node. This feature is required for auto parallel scenario which will split the user task automatically and dispatch sub-task to distributed nodes/ranks. Partial ranks of some nodes will be chosen to execute the related sub-tasks.
This PR is a customized implementation on distributed.launch and is in experimental.


python/paddle/distributed/fleet/launch_utils.py
```diff
@@ -798,6 +799,94 @@ def cloud_ps_heter_env_set(args):
...
+    node_rank = node_ips.index(node_ip)
+
+    assert args.ranks is not None, \
+        "ranks should be provided when enable_rank_mapping is set."
+    node_mapping_ranks_tmp = [x.strip() for x in args.ranks.split(';')]
+    node_mapping_ranks = []
+    gpus_num = fluid.core.get_cuda_device_count()
+    for s in node_mapping_ranks_tmp:
+        ranks_in_node = [int(x.strip()) for x in s.split(',')]
+        assert len(ranks_in_node) <= gpus_num, \
```
bliu3650: 只需要检查传递给当前 node 的 ranks 数与当前 node 的物理卡数。
assert len(node_mapping_ranks[node_rank]) <= gpus_num

python/paddle/distributed/fleet/launch_utils.py
```diff
@@ -798,6 +800,97 @@ def cloud_ps_heter_env_set(args):
...
+                 "node_rank:{} node_mapping_ranks:{}".format(
+                     node_ips, node_ip, node_rank, node_mapping_ranks[
+                         node_rank]))
+
+    # NOTE: there are different number of global mapped ranks on each node.
+    trainer_endpoints = []
+    for ip in node_ips:
+        node_rank = node_ips.index(ip)
+        free_ports = None
+        if not cloud_utils.use_paddlecloud() and len(
```
bliu3650: This if branch reused previous get_cluster_from_args implementation, and it's used for finding free port for local single node mode when there was no START_PORT provided. Seems good.

python/paddle/distributed/fleet/launch.py
```diff
@@ -157,6 +157,20 @@ def _parse_args():
         type=str,
         default="127.0.0.1",
         help="Paddle cluster nodes ips, such as 192.168.0.16,192.168.0.17..")
+    collective_group.add_argument(
+        "--ranks_mapped",
```
aoyulong: Please rename `ranks_mapped` to `rank_mapping`.

python/paddle/distributed/fleet/launch.py
```diff
@@ -157,6 +157,20 @@ def _parse_args():
...
+        type=str,
```
aoyulong: I think it is better to store the `rank_mapping` in a json file since this mapping may be very long. Another benefit of using a standard file format is that user can directly give the rank_mapping instead of using our auto mapping function. Our implementation should be compatible with this requirement.

python/paddle/distributed/fleet/launch.py
```diff
@@ -157,6 +157,20 @@ def _parse_args():
...
+        default="",
+        help="This mapped ranks information is used specifically for "
+        "lazy launch for auto parallel. Some of the ranks in each node "
+        "may not be used, and the indices of rank should be kept the same "
+        "as the indices of sub-task splited by auto parallel. "
+        "Paddle cluster nodes-ranks mapping example, such as 0,1;4,5,6;3,7..")
+    collective_group.add_argument(
+        "--enable_rank_mapping",
```
aoyulong: Please rename `enable_rank_mapping` to `enable_auto_mapping`.

python/paddle/distributed/fleet/launch_utils.py
```diff
@@ -798,6 +800,97 @@ def cloud_ps_heter_env_set(args):
...
+        pod.device_mode = device_mode
+        cur_node_endpoints = trainer_endpoints[node_rank]
+
+        # choose rank from global mapped ranks and set it to the trainer.
+        ranks_per_node = node_mapping_ranks[node_rank]
+        for i in range(len(ranks_per_node)):
+            trainer = Trainer()
+            # change global rank(mapped) to relative rank within each node.
+            # e.g. mapped ranks of node: 3,4,7 -> 0,1,2
+            relative_rank = ranks_per_node.index(ranks_per_node[i])
```
aoyulong: Please rename ` relative_rank` to `local_rank`.

python/paddle/distributed/fleet/launch_utils.py
```diff
@@ -798,6 +800,97 @@ def cloud_ps_heter_env_set(args):
...
+        "ranks_mapped should be provided when enable_rank_mapping is set."
+    # for example: 0,1;4,5,6;3,7 -> [[0,1], [4,5,6], [3,7]]
+    node_mapping_ranks_tmp = [x.strip() for x in args.ranks_mapped.split(';')]
+    node_mapping_ranks = []
+    for s in node_mapping_ranks_tmp:
+        ranks_in_node = [int(x.strip()) for x in s.split(',')]
+        node_mapping_ranks.append(ranks_in_node)
+
+    assert len(node_mapping_ranks[node_rank]) <= gpus_num, \
+        "number of ranks mapped to one node should not exceed the max."
```
aoyulong: Change `exceed the max` to `exceed the number of the available GPUs.`

python/paddle/fluid/tests/unittests/test_fleet_launch_mapped_ranks.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
aoyulong: Please rename `test_fleet_launch_mapped_ranks.sh` to `test_fleet_launch_rank_mapping.sh`.

python/paddle/distributed/fleet/launch.py
```diff
@@ -243,16 +257,19 @@ def launch_collective(args):
...
-    elif device_mode == DeviceMode.ASCEND_NPU:
-        # for ascend
+    # for ascend
+    if device_mode == DeviceMode.ASCEND_NPU:
         cluster, pod = ascend_utils.get_cloud_cluster(
             rank_table_file=os.getenv("RANK_TABLE_FILE", None),
             device_mode=device_mode,
             start_port=start_port)
+    # lazy launch for auto-parallel
+    elif args.enable_rank_mapping == True:
```
aoyulong: It is better to move this new branch to the last second one instead of in the middle of this if-else statement. If I understand right, the enable_rank_mapping does not distinguish local or cloud server.

python/paddle/distributed/fleet/launch.py
bliu3650: Agree. And changed it to rank_mapping_file, as the input in json format is better.

python/paddle/distributed/fleet/launch.py
bliu3650: Agree and done.

python/paddle/distributed/fleet/launch.py
bliu3650: Yes. This `elif args.enable_rank_mapping == True:` branch is shared with both the local and paddlecloud environment, so put it here. And the reason for not separating them is that this lazy launch (temporary solution) need additional step to collect the ip-ranks mapping information which should be provided by users or calculated automatically by preceding logic (which might be something like placement). Of course, when the environment on the paddlecloud can prepare this rank mapping information, this lazy launch process can add logic to get cluster from paddlecloud automatically then.

bliu3650:
Use https://github.com/PaddlePaddle/Paddle/pull/36570 instead which corrected the author's address email.

PR #35802, opened 2021-09-16
[CustomOp] Remove cuda sync in ext_tensor copy_to

chenwhql:
### PR types

Performance optimization
### PR changes

Others
### Describe

[CustomOp] Remove cuda sync in ext_tensor copy_to

Custom Op `copy_to` will run on stream from the current devicecontext, no need sync here


JiabinYang (approved):
LGTM

PR #35887, opened 2021-09-20
Add quant2 int8 lstm model test

wozna:
### PR types

New features
### PR changes

Others 
### Describe

This PR:
- adds a test with quant-aware training lstm model
- disable the function `self._update_relu_output_scales`. It causes nondeterministic quantization behaviour because it converts the quantization data range to uint8 when all scales were calculated for int8.
- refactors `save_quant_model`, so that you can call the graph transform function from the unit test


jczaja (approved):
LGTM

lidanqing-vv (approved):
LGTM

PR #35890, opened 2021-09-21
support ernie-int8 test and prune op attribute test

OliverLPH:
### PR types

Others

### PR changes

Others

### Describe

1. add macro of TensorRT version
2. add ernie-int8 precision test, only support after trt version > 7.2
3. add tensorrt tuned test for op attribute pruned model


jiweibo (approved):
LGTM

PR #35892, opened 2021-09-21
Added flatten and flatten2 BF16/FP32 FWD/BWD kernels

jakpiase:
### PR types
New features

### PR changes
OPs

### Describe
Added(currently disabled) flatten and flatten2 BF16/FP32 FWD/BWD kernels


jczaja:
@tsocha Please review this PR

jczaja (approved):
LGTM

jakpiase:
FYI:
Flatten, reshape and squeeze ops are all disabled and will be re-enabled when memory descriptor will be added to Tensor class. It was my intention to keep them disabled in this PR

lidanqing-vv (commented):
- If you commented the mkldnn kernels calling part, how does mkldnn kernels code coverage passed? It looks now the mkldnn UT will always only call native kernels, is it?
- When will be memory descriptor being added into Tensor? What is the cause to add mem descriptor into Tensor?

jakpiase:
Yes, for now only native kernels are called in mkldnn UTs. I will be adding md into tensor at the beginning of Q4, so it is only a temporary solution. We need to add memory descriptor into tensor class, because it is impossible to support all kind of data layouts through the memory format tags.

lidanqing-vv (approved):
LGTM

PR #35917, opened 2021-09-22
Add functional autograd API: jacobian

levi131:
### PR types
New features

### PR changes
APIs

### Describe
Add functional autograd API: jacobian.
The API signature is:
     `def jacobian(func, inputs, create_graph=False, allow_unused=False)`

This API computes the Jacobian matrix of `func` with respect to `inputs`.
This API is ONLY available in imperative mode.

 Parameters:

            list/tuple as inputs and returns a Tensor or a Tensor tuple.

 Returns:

        Jacobian (Tensor or nested tuple of Tensors): if function ``func``
        takes a Tensor as inputs and returns a Tensor as outputs, Jacobian
        will be a single Tensor containing the Jacobian matrix for the
        linearized inputs and outputs. If one of the inputs and outputs is
        a Tensor, and another is a Tensor list/tuple, then the Jacobian will
        be a tuple of Tensors. If both of inputs and outputs are Tensor
        list/tuple, then the Jacobian will be a tuple of tuple of Tensors
        where ``Jacobian[i][j]`` will contain the Jacobian matrix of the
        linearized ``i``th output and ``j``th input and will have same
        dtype and device as the corresponding input. ``Jacobian[i][j]`` will
        have as size ``m * n``, where ``m`` and ``n`` denote the numbers of
        elements of ``i``th output and ``j``th input respectively.

TODO:
        1、We find a bug about `grad` interface when running testcase TestJacobianFloat64::test_create_graph_true, will fix it later.
        2、May add some pre-process on `inputs` to support jacobian API when inputs.stop_gradients = True.
        3、Vectorize the computing process.


JiabinYang (approved):
LGTM

python/paddle/autograd/functional.py
```diff
@@ -0,0 +1,110 @@
...
+    return paddle.stack(
+        origin_list, axis=0) if isinstance(origin_list[0],
+                                           paddle.Tensor) else None
+
+
+@framework.dygraph_only
+def jacobian(func, inputs, create_graph=False, allow_unused=False):
+    ''' 
+    .. note::
+        **This API is ONLY available in Dygraph mode.**
```
phlrain: imperative mode

JiabinYang (changes requested):
Fix bugs first

python/paddle/autograd/functional.py
levi131: done

XieYunshen (dismissed):
LGTM for set_tests_properties(test_jacobian PROPERTIES TIMEOUT 20)

python/paddle/autograd/functional.py
```diff
@@ -0,0 +1,110 @@
...
+        linearized inputs and outputs. If one of the inputs and outputs is
+        a Tensor, and another is a Tensor list/tuple, then the Jacobian will
+        be a tuple of Tensors. If both of inputs and outputs are Tensor
+        list/tuple, then the Jacobian will be a tuple of tuple of Tensors
+        where ``Jacobian[i][j]`` will contain the Jacobian matrix of the
+        linearized ``i``th output and ``j``th input and will have same
+        dtype and device as the corresponding input. ``Jacobian[i][j]`` will
+        have as size ``m * n``, where ``m`` and ``n`` denote the numbers of
+        elements of ``i``th output and ``j``th input respectively.
+    '''
```
TCChenlong, line 153: 缺少示例代码

TCChenlong (dismissed):

python/paddle/autograd/functional.py
levi131: 已经补充

phlrain (approved):

CheQiXiao (approved):

XieYunshen (approved):

PangHua (approved):

JiabinYang (dismissed):
LGTM

TCChenlong (approved):

PR #35978, opened 2021-09-23
Add Basic CINN Runner Class

zhhsplendid:
### PR types

New features
### PR changes

APIs
### Describe

Add basic CINN entrance class.

There are 3 classes in this PR: `CinnRunner`, `CinnCacheKey`, `CinnCompiledObject`

`CinnRunner` is the main tool class which stores a cache from `CinnCacheKey` to `CinnCompiledObject` .
`CinnCacheKey` use `Graph`, input names, input shapes as cache key.
`CinnCompiledObject` will be the communication between Paddle and CINN, we will add more functions here after we have CINN interface.


paddle/fluid/framework/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_subdirectory(details)
 add_subdirectory(fleet)
 add_subdirectory(io)
 add_subdirectory(new_executor)
+add_subdirectory(cinn)
```
CtfGo: I think the name of this subdirectory is not clear enough to indicate we are the user of cinn and link it, someone like` pd2cinn` or `compiler_adaptor` maybe better.

CtfGo (commented):
In General, this pr LGTM, though it may need a little update afterwards to match cinn interface.

paddle/fluid/framework/CMakeLists.txt
zhhsplendid: Agree, but rename to `paddle2cinn` to match naming style of other modules.

zhhsplendid:
> In General, this pr LGTM, though it may need a little update afterwards to match cinn interface.

Yes, I agree

Superjomn (approved):
LGTM

PR #35985, opened 2021-09-23
fix pad tuple

littletomatodonkey:
### PR types
Bug fixes

### PR changes
APIs

### Describe
Support pad on all dims using tuple input


ceci3 (approved):
LGTM

PR #36033, opened 2021-09-23
Added shape (U)INT8/BF16/FP32 oneDNN kernel

jakpiase:
### PR types
New features

### PR changes
OPs

### Describe
Added shape (U)INT8/BF16/FP32 OneDNN kernel. It should speed up some models because it will remove unnecessary external reorders. It can work with any datatype supported by OneDNN, because shape kernel works in the same way, no matter what kind of datatype does the Input tensor have. External reorders are removed because shape kernel always results in plain memory format tag (NCHW/ABCD), so native kernel that is executed afterwards won't need a reorder. It speeds up faster_rcnn_r50_fpn_1x by 1%


paddle/fluid/operators/mkldnn/shape_mkldnn_op.cc
```diff
@@ -0,0 +1,58 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */
+
+#include "paddle/fluid/framework/op_registry.h"
+#include "paddle/fluid/operators/shape_op.h"
+#include "paddle/fluid/platform/mkldnn_helper.h"
+#include "paddle/fluid/platform/mkldnn_reuse.h"
```
arogowie-intel: You're probably not using this header.

paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -328,6 +328,23 @@ inline mkldnn::memory::format_tag GetMKLDNNFormat(
   return mkldnn::memory::format_tag::undef;
 }
 
+inline mkldnn::memory::format_tag GetPlainMKLDNNFormat(int tensor_rank) {
```
arogowie-intel: Just a note since I have a feeling that I saw this kind of code multiple times in many PR's. Please make sure that it is reused anywhere it could be.

paddle/fluid/platform/mkldnn_helper.h
```diff
@@ -328,6 +328,23 @@ inline mkldnn::memory::format_tag GetMKLDNNFormat(
...
+      return mkldnn::memory::format_tag::ab;
+    case 3:
+      return mkldnn::memory::format_tag::abc;
+    case 4:
+      return mkldnn::memory::format_tag::abcd;
+    case 5:
+      return mkldnn::memory::format_tag::abcde;
+  }
+
+  return mkldnn::memory::format_tag::abcdef;
```
arogowie-intel: You can make it `default` `switch` statement as well. As always in such _default_ cases I would recommend to check whether the input tensor rank is really what you're assuming to - here 6D. You never know how it may be used in the future. So it would make much faster to spot any errors if there would have been at least some warning or better assertion.

paddle/fluid/operators/mkldnn/shape_mkldnn_op.cc
```diff
@@ -0,0 +1,58 @@
...
+namespace operators {
+
+using paddle::framework::Tensor;
+using paddle::framework::LoDTensor;
+
+template <typename T>
+class ShapeMKLDNNKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext& ctx) const override {
+    auto* in_var = ctx.InputVar("Input");
```
arogowie-intel:
```suggestion
    const auto* in_var = ctx.InputVar("Input");
```

paddle/fluid/operators/shape_op.cc
```diff
@@ -58,6 +73,14 @@ Shape Operator.
...
 Return the shape of the input.
 )DOC");
+    AddAttr<bool>("use_mkldnn",
+                  "(bool, default false) Only used in mkldnn kernel")
+        .SetDefault(false);
+    AddAttr<std::string>(
+        "mkldnn_data_type",
+        "(string, default \"float32\"). Data type of mkldnn kernel")
+        .SetDefault("float32")
+        .InEnum({"float32", "bfloat16", "int8"});
```
arogowie-intel: I wonder whether this attribute is necessary?  I do not see you use it.

python/paddle/fluid/tests/unittests/mkldnn/test_shape_mkldnn_op.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def config(self):
+        self.shape = [5, 7, 4]
+        self.dtype = np.float32
+
+    def test_check_output(self):
+        self.check_output_with_place(core.CPUPlace())
+
+
+class TestShape1DOneDNNOp(TestShapeOneDNNOp):
```
arogowie-intel: Every time I see such almost identical test classes I wonder why python `unittest` package does not support parameterization. This is rather just a note for you to be aware of those things. The best solution for such cases would be to use `pytest` package since it supports [parameterized test cases](https://docs.pytest.org/en/latest/parametrize.html). However uncle Google says that there is also package like [ddt](https://stackoverflow.com/a/50514671) which looks promising and can be used for `unittest` package.

arogowie-intel (changes requested):

paddle/fluid/platform/mkldnn_helper.h
lidanqing-vv: Also, is 6 dimensions biggest dimension it could be, because there used to be some issues caused by unsupported format.

paddle/fluid/operators/shape_op.cc
jakpiase: I think that it is needed for bf16 inference pass

lidanqing-vv (dismissed):
LGTM

paddle/fluid/operators/shape_op.cc
wozna: Exactly, this is needed for bfloat16 inference pass and for future quantization.

jakpiase:
I will continue working on this PR after md-in-tensor related PR will be merged, because this PR has a dependency on that PR

jczaja (approved):
LGTM

jczaja:
@baoachun Please start your review

paddle/fluid/operators/shape_op.cc
```diff
@@ -58,6 +73,16 @@ Shape Operator.
 
 Return the shape of the input.
 )DOC");
+    AddAttr<bool>("use_mkldnn",
```
baoachun, line 76: @lidanqing-intel  Do we still support modifying the properties of native operators now?

jczaja: @baoachun Plan is (as executed in https://github.com/PaddlePaddle/Paddle/pull/36541) that the only new attribute to be added in native ops is "use_mkldnn" . That is becasue in training there are no passes, so we still need an attribute (set with env var FLAGS_use_mkldnn) so that PaddlePaddle runs oneDNN kerenels.

jakpiase: `use_mkldnn` attribute is necessary for running mkldnn kernels during training since there are not inference passes in that mode. This attribute is used to determine whether mkldnn kernel should be run or not and that behavior is checked in `GetExpectedKernelType()`. It is a common behavior for all mkldnn kernels and that attribute must be included in original operator version but I can create another onednn-based op and move `mkldnn_data_type` attribute into it, but that cannot be done with `use_mkldnn` attribute. Personally, I would like to keep both of these attributes in native op to allow common inference infrastructure usage but if that's a problem then I'll create a separate op for that.
Here's code for shape op's `GetExpecterKernelType()` which uses `use_mkldnn`  in `this->CanMKLDNNBeUsed()` function https://github.com/PaddlePaddle/Paddle/blob/c4755323b2719692b938e5cb5b6900e2cdd116d3/paddle/fluid/operators/shape_op.cc#L38-L51

jczaja: @baoachun , @jakpiase 
Training does no use passes. So Executor used for training is having a mechanism that when FLAGS_use_mkldnn=True then iteration through all ops is done and those having use_mkldnn attribute are having this attribute set to True. So that is the way how during training to let know Paddle to use oneDNN kernels.

Regarding GPU and its "use_cudnn". I looked to code and most of operators (older ones like, conv2d, softmax, gelu, activation, abs,....) are having "use_cudnn" defined. Probably due to API compatibility reasons. 
Some newer ops e.g.  Matmul . Are not having use_cudnn. But matmul CPU and GPU kernel are the same method and execution decision is made based on parameter Place e.g. CPU or GPU.

Some others like affine_channel are having their own kernel for GPU , different from CPU. And decision which run is made based on searching through registers kernels when Key is OpKernelType using following comparison: 

https://github.com/PaddlePaddle/Paddle/blob/1252f4bb3e574df80aa6d18c7ddae1b3a90bd81c/paddle/fluid/framework/op_kernel_type.cc#L55-L60

So you can see that Kernel is chosen if required type matches .e.g library, place, type .  

For GPU kernels there is GPU place. So library plain and GPU place inidicate that GPU kerenels should be used.
For  oneDNN kernels it is a bit harder as there is only CPU place which is used for CPU kernels and oneDNN kernels.
So to pick oneDNN kernels we also need to set library to MKLDNN. And to set library to MKLDNN we currently use "use_mkldnn".  

What I'm trying to write is that if there were oneDNN place then we could do the same as for GPU e.g. discard use_mkldnn attribute. But since there is no oneDNN place and we use CPU place then we need some way to indicate
during the training that library should be MKLDNN and this is done using attribute use_mkldnn.

baoachun (approved):
LGTM

PR #35909, opened 2021-09-22
Replace Eigen with Lapack library for "eigvals" OP kernel

From00:
### PR types
Function optimization

### PR changes
OPs

### Describe
This PR implements the kernel of "eigvals" OP with the Lapack library, which has a better performance than the previous Eigen library.

Here are some preliminary test results comparing between two library:
image

PR #36038, opened 2021-09-24
[cherry-pick] Replace Eigen with Lapack library for eigvals OP kernel

From00:
### PR types
Function optimization

### PR changes
OPs

### Describe
cherry-pick (#35909)

This PR implements the kernel of "eigvals" OP with the Lapack library, which has a better performance than the previous Eigen library.

Here are some preliminary test results comparing between two library:
image

In general, for small-scale float32 matrices, the accuracy of Lapack implementation may slightly worse than that of Eigen in some cases. However, in most cases, Lapack implementation is closer to NumPy. And more crucially, in terms of computing time cost, Eigen is significantly slower than Lapack. Therefore, it is necessary to replace Eigen with Lapack library for linalg.eigvals OP.

### DIFF
--- a/paddle/fluid/operators/eigvals_op.h
+++ b/paddle/fluid/operators/eigvals_op.h
@@ -14,60 +14,44 @@
 
 #pragma once
 
-#include <complex>
+#include <string>
 #include <vector>
-#include "Eigen/Dense"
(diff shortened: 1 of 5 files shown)


zhiqiu (approved):
LGTM

lanxianghit (approved):

PR #36063, opened 2021-09-24
Remove the compute functor from reduce_functor_op.h  to helper_primitives.h

AnnaTrainingG:
### PR types

Others
### PR changes

APIs 
### Describe

Remove compute functor from reduce_functor to helper_primitives.h


zhiqiu (dismissed):
LGTM

PR #36102, opened 2021-09-26
py bug fix for pslib

kuizhiqing:
### PR types

Bug fixes

### PR changes

Others

### Describe

* py2 to py3 should use int wrap float division
* default _iface for gloo if preferred other than "lo"
* FLAGS_selected_gpus may not be set


danleifeng (approved):
LGTM

PR #36108, opened 2021-09-26
Add functional autograd API:hessian

levi131:
### PR types
New features

### PR changes
APIs
### Describe
1、Add functional autograd API: hessian.
2、Move some helper interfaces together into utils.py.

The API signature is:
     `def hessian(func, inputs, create_graph=False, allow_unused=False)`

This API computes the hessian matrix of `func` with respect to `inputs`.
This API is ONLY available in imperative mode.

 Parameters:

            list/tuple as inputs and returns a Tensor with a single element.

 Returns:

        Hessian (Tensor or a tuple of tuple of Tensors): if function ``func``
        takes a Tensor as ``inputs``, Hessian will be a single Tensor containing
        the Hessian matrix for the linearized ``inputs`` Tensor. If function
        ``func`` takes a Tensor list/tuple as ``inputs``, then the Hessian will
        be a tuple of tuple of Tensors where ``Hessian[i][j]`` will contain the
        Hessian matrix of the ``i``th input and ``j``th input with size ``m * n``.
        Here ``m`` and ``n`` denote the number of elements of the ``i`` th input
        and the ``j`` th input respectively.

TODO:
        1、enable test case test_create_graph_true when 3nd order grad op  matmul_grad_grad_grad is ok.


phlrain (approved):

XieYunshen (approved):
LGTM for set_tests_properties(test_hessian PROPERTIES TIMEOUT 20)

JiabinYang (approved):
LGTM

PR #36119, opened 2021-09-26
support saving model defined parameters without add scale_op

hp03:
### PR types
Bug fixes

### PR changes
Others

### Describe
pr 35730 removes scale_op for target_vars. sometimes user need to export parameters defined in model, just like the code  below.
```
import paddle
import math
import numpy as np

class ONNXModel(paddle.nn.Layer):
    def __init__(self):
        super(ONNXModel, self).__init__()
        self.x2paddle_transitions_0 = self.create_parameter(shape=[12, 12], attr='x2paddle_transitions_0', dtype='float32', default_initializer=paddle.nn.initializer.Constant(value=0.0))
        self.conv0 = paddle.nn.Conv2D(in_channels=512, out_channels=256, kernel_size=[3, 3], padding=[1, 1, 1, 1], bias_attr=True)

    def forward(self, input_tensor):
        x2paddle_transitions_0 = self.x2paddle_transitions_0
        output = self.conv0(input_tensor)
        return x2paddle_transitions_0, output

model = ONNXModel()
model.eval()

sepc_list = list()
sepc_list.append(paddle.static.InputSpec(shape=[1, 512, 256, 256], dtype="float32", name="x2paddle_input_mask_0"))
static_model = paddle.jit.to_static(model, input_spec=sepc_list)

print("static fininsed!")

paddle.jit.save(static_model,"inference_model/model")

print("save fininsed!")
```


Superjomn (approved):
LGTM

zhiqiu (approved):

PR #36146, opened 2021-09-27
Lml/vhp

levi131:
### PR types
New features
### PR changes
APIs
### Describe
Add autograd functional API: vhp
This function computes the product between a vector ``v`` and the
Hessian matrix of `func` with respect to `inputs`.

    Parameters:
            list/tuple as inputs and returns a Tensor with a single element.
        v (Tensor|list(Tensor)|tuple(Tensor)|None, optional): the vector used
            to compute vector hessian product. ``v`` should have same shape
            and dtype with ``inputs``. If ``v`` is None, it will be set as
            Tensor|list(Tensor) with all elements 1. Defaults to "None".
    Returns:
        output (tuple): tuple with:
            func_output (Tensor): output of ``func(inputs)``
            vhp (list(Tensor)): result of the vector hessian product
            with the same shape and dtype as the inputs.


python/paddle/autograd/functional.py
```diff
@@ -285,3 +285,115 @@ def jac_func(*ins):
 
     return jacobian(
         jac_func, inputs, create_graph=create_graph, allow_unused=allow_unused)
+
+
+@framework.dygraph_only
+def vhp(func, inputs, v=None, create_graph=False, allow_unused=False):
+    ''' 
+    .. note::
+        **This API is ONLY available in imperative mode.**
```
tongxin: in the imperative mode.

tongxin: This function computes..

tongxin: The product of ``v`` and the hessian of ``func``. It may not be a dot product.

python/paddle/autograd/functional.py
```diff
@@ -285,3 +285,115 @@ def jac_func(*ins):
...
+            vy = paddle.ones(shape=[2, 2], dtype='float32') * 3
+            vhp_rslt = paddle.autograd.vhp(func, [x, y], v=[vx, vy], allow_unused=True)
+            print(vhp_rslt)
+            # (Tensor(shape=[1], dtype=float32, place=CUDAPlace(0), stop_gradient=False,
+            #        [8.]),
+            # (Tensor(shape=[2, 2], dtype=float32, place=CUDAPlace(0), stop_gradient=True,
+            #        [[8., 8.],
+            #         [8., 8.]]), None))
+    '''
+    inputs = _check_tensors(inputs, "inputs")
```
tongxin: I think the main reason of using _check_tensors here is to turn the arguments to a proper tensor list.  in that sense I'd prefer to use the name something similar to ``to_tensorlist``. To eliminate redundancy, I can replace the function with same name with this one.

python/paddle/autograd/functional.py
```diff
@@ -285,3 +285,115 @@ def jac_func(*ins):
...
+    outputs = func(*inputs)
+    assert isinstance(outputs, paddle.Tensor) and outputs.shape == [
+        1
+    ], "The function to compute vhp should return a Tensor with a single element"
```
tongxin: Can we do a function with tensor output?

python/paddle/autograd/utils.py
```diff
@@ -47,3 +47,9 @@ def _replace_none_with_zero_tensor(t, spec_t):
         return zero_t
     else:
         return t
+
+
+def _construct_one_tensor(spec_t):
```
tongxin: How about just calling paddle.ones_like()? There's a tendency of over abstraction here.

python/paddle/autograd/functional.py
```diff
@@ -285,3 +285,115 @@ def jac_func(*ins):
...
+    if v is None:
+        v = list(_construct_one_tensor(input) for input in inputs)
```
tongxin: I think it suffices to use ones_like(). And, if use vjp, jvp, we can avoid resetting stop_gradient.

python/paddle/autograd/functional.py
```diff
@@ -285,3 +285,115 @@ def jac_func(*ins):
...
+    else:
+        v = _check_tensors(v, "v")
+
+    jac = paddle.grad(
+        outputs,
+        inputs,
+        create_graph=True,
+        retain_graph=True,
+        allow_unused=allow_unused)
```
tongxin: Shall we shrink the number of lines here?

python/paddle/autograd/functional.py
```diff
@@ -285,3 +285,115 @@ def jac_func(*ins):
...
+    jac = tuple(
+        _replace_none_with_zero_tensor(jac[i], inputs[i])
```
tongxin: grads = [zeros_like(x) if grad is None else grad for grad, x in zip(jac, inputs)]

python/paddle/autograd/functional.py
levi131: Many thanks for giving these suggestions

python/paddle/autograd/functional.py
levi131: We only support function who returns a Tensor with a single element.

python/paddle/autograd/functional.py
levi131: I' would polish the code here after the pr https://github.com/PaddlePaddle/Paddle/pull/36020 is merged.

python/paddle/autograd/functional.py
levi131: This line is too long to be wrote in one line.

python/paddle/autograd/functional.py
levi131: I' would polish the code here after the pr https://github.com/PaddlePaddle/Paddle/pull/36020 is merged.

python/paddle/autograd/utils.py
levi131: I' would polish the code here after the pr https://github.com/PaddlePaddle/Paddle/pull/36020 is merged.

XieYunshen (approved):
LGTM for `set_tests_properties(test_vhp PROPERTIES TIMEOUT 50)`

JiabinYang (approved):
LGTM

PR #36150, opened 2021-09-27
Remove wait if no fetch data in PE

sneaxiy:
### PR types
Performance optimization

### PR changes
Others

### Describe
Remove device context wait if there is no fetch data in ParallelExecutor.


chenwhql (approved):
LGTM

PR #36162, opened 2021-09-27
Rename ScaleLossGradOpHandle::Name() return value for better log analysis

sneaxiy:
### PR types
Others

### PR changes
Others

### Describe
The `ScaleLossGradOpHandle::Name()` returns a name with blank, which makes log analysis difficult (for example using `awk`).


chenwhql (approved):
LGTM

PR #36175, opened 2021-09-28
[NPU] fix bug of top_k npu op

pangyoki:
### PR types

Bug fixes

### PR changes

OPs

### Describe

`N` is `the last dimension of input` rather than `the size of input`.
`assist_seq`: A 1D tensor of type float16. with size of 2N, which `"N" is the last dimension`. The first N numbers is indices, and the next N numbers is deviation of casting float16 to int32.


zhiqiu (approved):
LGTM

PR #36187, opened 2021-09-28
[New Feature] Support triple grad in Paddle

JiabinYang:
### PR types

 New features
### PR changes

OPs 
### Describe

This PR is trying to support higher order differentiation (3 order) in operators:

1. Support it by refactor rename strategy in double grad mechanism
2. Support triple grad gradient checker
3. Support Sigmoid triple grad kernel


python/paddle/fluid/backward.py
```diff
@@ -408,7 +413,9 @@ def _strip_grad_suffix_(name):
     """
     name = cpt.to_text(name)
     pos = name.find(core.grad_var_suffix())
-    return name[:pos] if pos != -1 else name
+    new_name = name[:pos] if pos != -1 else name
+    new_pos = name.rfind('grad/')
+    return new_name[new_pos + 5:] if new_pos != -1 else new_name
```
chenwhql, line 418: why add 5? do we need a comment here?

phlrain: len('grad/') = 5

chenwhql (approved):
LGTM

python/paddle/fluid/backward.py
JiabinYang: Sure, can I add comment in next pr

PR #36190, opened 2021-09-28
Add basic support for CUDA Graph

sneaxiy:
### PR types
New features

### PR changes
APIs
### Describe
Add basic support for CUDA Graph, including:

- Memory pool for CUDA Graph capturing. CUDA Graph needs to cache all Tensor address without calling any `cudaFree`.
- Basic APIs like `CUDAGraph.capture_begin/capture_end/replay/reset`. Notice that this API is in the experimental stage, and may be changed in the future.
- Use `CUDAGraphCaptureModeGuard` to switch to `cudaStreamCaptureModeRelaxed` to skip the unsupported `cudaMalloc` during capturing.

Usage:
```
from paddle.device.cuda.graphs import CUDAGraph

input = ... # define the input tensor. The definition of the input tensor should be before the `CUDAGraph.capture_start()`.

graph = CUDAGraph()
graph.capture_start()
output = ... # do some GPU operations here
graph.capture_end()

for _ in range(BATCH_NUM):
    input.copy_(input_tensor, False) # input_tensor is the input data of the model, may be from DataLoader
    graph.replay()
    print(output)

graph.reset() # it is not required, but it is better to call this method to release cached memory asap
```

TODO: when using CUDA Graph,
- Cache CuDNN descriptor. Otherwise, errors would raise during capturing.
- Disable CuDNN exhaustive search. Otherwise, errors would raise during capturing.
- Disable `FLAGS_sync_all_reduce` when using distributed training. The `FLAGS_sync_all_reduce` would call `cudaStreamSynchronize`, which is not supported during capturing.
- Modify ParallelExecutor to support CUDA Graph.


paddle/fluid/memory/allocation/allocator_facade.cc
```diff
@@ -279,12 +351,57 @@ class AllocatorFacadePrivate {
...
+    for (auto& item : allocator->allocators_) {
+      auto& old_allocator = item.second;
+      old_allocator = CUDAGraphAllocator::Create(old_allocator);
+    }
+    VLOG(10) << "Prepare memory pool for CUDA Graph with ID " << id;
+  }
+
+  void RemoveMemoryPoolOfCUDAGraph(CUDAGraphID id) {
+    auto iter = cuda_graph_allocator_map_.find(id);
+    PADDLE_ENFORCE_EQ(iter != cuda_graph_allocator_map_.end(), true,
```
zhiqiu: Can use PADDLE_ENFORCE_NE  directly

paddle/fluid/platform/cuda_graph.h
```diff
@@ -0,0 +1,116 @@
...
+  static std::unique_ptr<CUDAGraph> capturing_graph_;
+};
+
+#if CUDA_VERSION >= 10010
+class CUDAGraphCaptureModeGuard {
+  DISABLE_COPY_AND_ASSIGN(CUDAGraphCaptureModeGuard);
+
+ public:
+  explicit CUDAGraphCaptureModeGuard(cudaStreamCaptureMode new_mode) {
+    old_mode_ = new_mode;
```
zhiqiu: why old_mode_ = new_mode?

paddle/fluid/platform/cuda_graph.h
```diff
@@ -0,0 +1,116 @@
...
+    PADDLE_ENFORCE_CUDA_SUCCESS(
+        cudaThreadExchangeStreamCaptureMode(&old_mode_));
+  }
+
+  ~CUDAGraphCaptureModeGuard() PADDLE_MAY_THROW {
+    PADDLE_ENFORCE_CUDA_SUCCESS(
```
zhiqiu: Is it ok to raise exception in the destructor?

paddle/fluid/platform/gpu_info.cc
```diff
@@ -557,6 +558,7 @@ class RecordedCudaMallocHelper {
 #ifdef PADDLE_WITH_HIP
     auto result = hipMalloc(ptr, size);
 #else
+    CUDAGraphCaptureModeGuard capture_mode_guard{cudaStreamCaptureModeRelaxed};
```
zhiqiu, line 561: imho, call this func when IsCUDAGraphCapturing is true.

paddle/fluid/platform/cuda_graph.h
sneaxiy: Yes. Although it is not recommended to raise exception in the destructor, I think that we should not hide the exception. If exception is raised in the destructor, `std::terminate` would be called to stop the process immediately.

paddle/fluid/platform/gpu_info.cc
sneaxiy: Done.

paddle/fluid/memory/allocation/allocator_facade.cc
sneaxiy: Done.

paddle/fluid/platform/cuda_graph.h
sneaxiy: Change the variable name and add some comments for better understanding.

paddle/fluid/platform/cuda_graph.h
zhiqiu: > Yes. Although it is not recommended to raise exception in the destructor, I think that we should not hide the exception. If exception is raised in the destructor, `std::terminate` would be called to stop the process immediately.

Maybe we need a `PADDLE_WARN_CUDA_SUCCESS`, etc.

sneaxiy: > > Yes. Although it is not recommended to raise exception in the destructor, I think that we should not hide the exception. If exception is raised in the destructor, `std::terminate` would be called to stop the process immediately.
> 
> Maybe we need a `PADDLE_WARN_CUDA_SUCCESS`, etc.

Hard to do this. Suppose that we have a common method `void func()`. The `func` may be called anywhere, inside destructor or outside destructor, but we have to only write one of `PADDLE_ENFORCE_CUDA_SUCCESS` and  `PADDLE_ENFORCE_WARN_SUCCESS` inside its implementation.

zhiqiu (approved):
LGTM

PR #36258, opened 2021-10-04
[NPU] fix elementwise_mul to support broadcast, test=develop

qili93:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix elementwise_mul to support broadcast.


python/paddle/fluid/tests/unittests/npu/test_elementwise_mul_op_npu.py
```diff
@@ -18,147 +18,208 @@
...
-        x = np.random.uniform(1, 2, [3, 4]).astype(self.dtype)
-        y = np.random.uniform(1, 2, [3, 4]).astype(self.dtype)
-        out = np.multiply(x, y)
+        self.outputs = {
+            'Out': self.inputs['X'] * self.inputs['Y'].reshape(1, 1, 100)
+        }
+
+
+# TODO(qili93): Paddle enformce not met: 
+# dimension mismatch at elementwise_op_function.h:240
```
zhiqiu: assertRaises can check exception

qili93: Fixed by adding axis support.

In paddle 2.0, axis is not supported as parameter in multiply api, while axis is able to set to non -1 in paddle.fluid.layers.elementwise_mul. Fix elementwise_mul_op_npu.cc to support axis = -1. The previous error is caused by input is broadcast case while axis is always set to -1 in 2.0 api.

zhiqiu (approved):
LGTM

zhangting2020 (approved):
LGTM for skip_check_grad_ci

PR #36284, opened 2021-10-08
Fix for oneDNN conv op

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix for conv op

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -219,7 +219,8 @@ class ConvMKLDNNHandlerT
 
       float sum_scale;
       std::vector<float> output_shift_scale;
-      std::tie(sum_scale, output_shift_scale) = get_int8_scales(ctx);
+      if (platform::is_int8<T>() || fuse_residual_conn)
+        std::tie(sum_scale, output_shift_scale) = get_int8_scales(ctx);
 
       const mkldnn::primitive_attr conv_attr = CreatePostOps(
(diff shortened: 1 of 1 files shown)


jczaja (dismissed):
LGTM

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -219,7 +219,8 @@ class ConvMKLDNNHandlerT
 
       float sum_scale;
       std::vector<float> output_shift_scale;
-      std::tie(sum_scale, output_shift_scale) = get_int8_scales(ctx);
+      if (platform::is_int8<T>() || fuse_residual_conn)
```
wozna: Do you need to add this `|| fuse_residual_conn` to the condition? Checking if it is int8 is not enough?

jakpiase: Unfortunately, `TestWithFuse` of `test_conv2d_mkldnn_op.py` was giving me wrong results when there were no `|| fuse residual conn`. To be honest, I do not fully understand that behavior

wozna: Okey, you can remove this `fuse_residual_conn` from the condition, just add the default value to` sum_scale = 1.0f`. And now it will work, because in fact an uninitialized value was set to the residual_conn scale.

wozna (approved):
LGTM

jczaja (approved):

PR #36298, opened 2021-10-09
Add const for OpDesc::id() and VarDesc::id()

sneaxiy:
### PR types
Others

### PR changes
Others

### Describe
Add `const` to `OpDesc::id()` and `VarDesc::id()` method.


zhiqiu (approved):
LGTM

PR #36321, opened 2021-10-11
fix fft axis

iclementine:
### PR types

Bug fixes

### PR changes

APIs

### Describe

fix fft axis: 
Before: When axis evaluates to False, `-1` is used instead.
After: Only when axis is `None` will `-1` be used.


jeff41404 (approved):
lgtm

PR #36332, opened 2021-10-11
fix bug of clear third_party cache every 10 days

betterpig:
### PR types

Bug fixes
### PR changes

Others
### Describe

There are two places write and read day.txt, so when the second operation set param day_before by reading day.txt, it actually is day_now, which was write by the first operation. So that third_party cache can't be removed. 
To fix the bug I put these two rmdir operation together, since they can share the same code of day comparing.


zhwesky2010 (approved):
LGTM

PR #36344, opened 2021-10-11
Fix the CE's bug when axis is specified and weight is provided

HydrogenSulfate:
### PR types
Bug fixes

### PR changes
APIs

### Describe
**Background**: 
CrossEntropy loss function uses hard label, and when the weight is received at the same time, then if specify the axis other than -1, an error will be reported in the calculation process

**Problem location:** 
gather_nd will be used when calculating the intermediate variable weight_gather, but when the coordinate value is not in the last dimension, gather_nd will make an error. When the input shape is as described in the background, it will cause the above problem

**Solution:**
When the axis is specified as a dimension other than -1, manually construct a correct permutation and pass it to the subsequent gather_nd function to use, so as to get the weight_gather with the correct shape

**Other modifications:**
Fixed some error judgment conditions related to axis, and add an test case in unittest


chajchaj (approved):
LGTM

PR #36351, opened 2021-10-11
Refine LarsOptimizer

sneaxiy:
### PR types
Function optimization

### PR changes
OPs

### Describe
Replace LARS with Momentum when weight decay is excluded.


JamesLim-sy (approved):
Good work.

PR #36368, opened 2021-10-12
unify usage of tuple and list

levi131:
### PR types
Others

### PR changes
APIs

### Describe
1、Fix a mistake in docstring of API grad.
        * The return value of grad API is a list but described as tuple in docstring.
2、Unify usage of list and tuple in functial aitograd API vjp/jvp.
        * The argument `func` may use Tensor|list[Tensor]|tuple[Tensor] as inputs and outputs;
        * The `vjp/jvp` part of return value should always be list[Tensor], which is same type as the return value of grad API;
3、Merge two interfaces `_check_tensors` and `to_tensorlist` into one interface and rename as `_tensors`.


TCChenlong (approved):

XiaoguangHu01 (approved):
LGTM

PR #36385, opened 2021-10-13
Remove RunFromCinn in PE because We Will Call CinnRunner in Compute of SubgraphOp 

zhhsplendid:
### PR types

Function optimization
### PR changes

Others
### Describe

Remove `RunFromCinn` method in PE because We Will Call `CinnRunner` in `Compute` method of SubgraphOp


wzzju (approved):
LGTM

CtfGo (approved):
LGTM

PR #36397, opened 2021-10-13
adjust parallel UT rule

betterpig:
### PR types

Others
### PR changes

Others
### Describe

- reduce some unittest's parallel number to avoding timeout failure
- in paddle_build.bat, add /d after cd to change diretory safely


zhwesky2010 (approved):
LGTM

PR #36431, opened 2021-10-14
Fix LARS wrong thread synchronization

sneaxiy:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix LARS wrong thread synchronization


Xreki (approved):
LGTM

PR #36466, opened 2021-10-15
fix out of range for area interp

tink2123:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix out of range for area interp


ceci3 (approved):
LGTM

PR #36492, opened 2021-10-17
Use master parameter to calculate L2NormKernel of LARSOptimizer

sneaxiy:
### PR types
Function optimization

### PR changes
OPs

### Describe
To make more precise results, use master parameter to calculate the `p_norm` when AMP is enabled.


sneaxiy:
Close because of no precision improvement.

PR #36511, opened 2021-10-18
fix out_of_range bug of multinomial op's cuda kernel

pangyoki:
### PR types

Bug fixes

### PR changes

OPs

### Describe

A bug in `sample` method (`multinomial` op) of `Categorical` API is mentioned in issue https://github.com/PaddlePaddle/Paddle/issues/36401.

#### issue

- test case

```python
import paddle
with paddle.no_grad():
    actor=paddle.nn.Sequential(paddle.nn.Linear(20, 2800))
    logits=actor(paddle.rand([20]))
    cat=paddle.distribution.Categorical(logits.exp())
    print(cat.sample([1]))
```

- error message

```python
Error: /paddle/paddle/fluid/operators/multinomial_op.cu:42 Assertion `in_data[id] >= 0.0` failed. The input of multinomial distribution should be >= 0, but got -0.038249.
（省略类似错误）
Error: /paddle/paddle/fluid/operators/multinomial_op.cu:42 Assertion `in_data[id] >= 0.0` failed. The input of multinomial distribution should be >= 0, but got -0.044257.
  File "bug.py", line 6, in <module>
    print(cat.sample([1]))
  File "/opt/conda/envs/python35-paddle120-env/lib/python3.7/site-packages/paddle/distribution.py", line 771, in sample
    sample_index = multinomial(logits, num_samples, True)
  File "/opt/conda/envs/python35-paddle120-env/lib/python3.7/site-packages/paddle/tensor/random.py", line 133, in multinomial
    'replacement', replacement)
SystemError: (Fatal) Operator multinomial raises an thrust::system::system_error exception.
The exception content is
:transform: failed to synchronize: cudaErrorLaunchFailure: unspecified launch failure. (at /paddle/paddle/fluid/imperative/tracer.cc:192)
```

#### Reason for error

In the cuda kernel implementation, the number of threads exceeding the size of the input array will be used to perform calculations (the reason is that the `block` size is limited, when setting the `grid` size, more threads will be set for `rounding`).

However, when the cuda kernel calculates, it does not limit the array subscripts. As a result, when calculating in the thread, the space exceeding the size of the array is accessed, causing an error.

#### bug fix

Restrictions on the subscripts of the accessed arrays.

图片


zhiqiu (approved):
LGTM

chenwhql (approved):
LGTM for PADDLE_ENFORCE

Avin0323 (approved):
LGTM for PR-CI-OP-benchmark

PR #36568, opened 2021-10-20
fix fc fuse compat problem

jiweibo:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix fc fuse compat problem

image


winter-wang (approved):
LGTM

PR #36595, opened 2021-10-20
[WIP] partial disabling of caching for conv2d, conv_transpose, quantize and pool2d

### LINKED ISSUE #34554: Deploying models with multi-thread will raise MKLDNN error
**System information**
-PaddlePaddle version: develop(393a0b166a380e85550a639e603a7ff44c0a440e) or release/2.1 
-CPU: MKLDNN v2.2.1
-GPU: None
-OS Platform: ubuntu1604
-Python version: None
-Cmake orders
-C++version.txt
-API information
Note: You can get most of the information by running [summary_env.py](https://github.com/PaddlePaddle/Paddle/blob/develop/tools/summary_env.py). 
**To Reproduce**
reproduce procedure is same as https://github.com/PaddlePaddle/Paddle/issues/31992

cmake commands
```shell
cmake .. -DWITH_FLUID_ONLY=ON -DWITH_GPU=OFF -DWITH_TESTING=OFF -DCMAKE_BUILD_TYPE=Release -DON_INFER=ON -DWITH_INFERENCE_API_TEST=OFF -DWITH_MKL=ON -DWITH_MKLDNN=ON -DWITH_AVX=ON -DWITH_DISTRIBUTE=OFF -DWITH_STRIP=ON -DWITH_PYTHON=OFF
```

* The first test
Run `./build/model_test --single_thread=true --single_instance=false --test_groups="1"`,  it does not case error.

* The second test
Run `./build/model_test --single_thread=true --single_instance=true --test_groups="1"`,  it does not case error.

* The third test
Run `./build/model_test --single_thread=false --single_instance=true --test_groups="1"`,  it will raise an error as followed
image

this error might caused by https://github.com/PaddlePaddle/Paddle/pull/33549/files#diff-3363ff47a1f22a111386eeff77a8572f6618d7d0602ffe71e92ecf8682faa84eR2343

**Other info / logs**

jczaja:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Disables some more caching of oneDNN objects in order to fix #34554

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -433,11 +433,9 @@ class ConvMKLDNNHandlerT
             ? (groups > 1 ? (weights_tz)[1] * (weights_tz)[0] : (weights_tz)[0])
             : 1;
 
-    std::vector<float> v;
-    v.reserve(count);
-    bias_scale_tuple = std::make_shared<std::tuple<float, std::vector<float>>>(std::tuple<float, std::vector<float> >(mask_reorder, std::vector<float>(count)));
+    bias_scale_tuple = std::make_shared<std::tuple<float, std::vector<float>>>(std::make_tuple(static_cast<float>(mask_reorder), std::vector<float>(count)));
     for (int i = 0; i < count; i++) {
-      bias_scale_tuple->get<1>[i] = scale_in_data * scale_weights_data[i];
+      bias_scale_tuple->get<1>()[i] = scale_in_data * scale_weights_data[i];
     }
 
     dev_ctx.SetBlob(key_bs, bias_scale_tuple);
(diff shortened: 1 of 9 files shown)


paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -63,330 +63,306 @@ static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
...
-              framework::DataLayout::kMKLDNN, in->layout()));
-      PADDLE_ENFORCE_NE(in->format(), MKLDNNMemoryFormat::undef,
-                        platform::errors::InvalidArgument(
-                            "Got wrong format for Input tensor."));
+                     Tensor* in_x_grad)
+      : platform::MKLDNNHandlerNoCachingT<T, mkldnn::convolution_forward,
+                                          mkldnn::convolution_backward_data,
+                                          mkldnn::convolution_backward_weights>(
+            mkldnn_engine, ctx.GetPlace()),
+        is_test_(false) {
```
wozna, line 246: Shouldn't there be a check for the `is_test` attribute as it is done for forward `is_test_(ctx.Attr<bool>("is_test"))`? Then we also check if `is_test` is false, which if we set false here, it doesn't make sense

wozna (commented):
The rest of the changes look good for me

jczaja:
depracated

PR #36643, opened 2021-10-22
Add distributed pass framework: including PassBase/PassTest/PassUtils

sneaxiy:
### PR types
New features

### PR changes
Others

### Describe

- Add the distributed pass framework, including:
    - `PassContext`: a helper class to record the applied pass history. 
    - `PassBase`: the super class of all distributed passes. The developers should override the `_check_self`, `_check_conflict`, `_apply_impl/_apply_single_impl` methods to implement their own passes.
    - `ParallelOptPass`, `CalcOptPass` and `CommOptPass`: three classes that are derived from `PassBase`. The developers should derive one of them instead of `PassBase` to implement their own passes.
    - `CPPPassWrapper`: a class that is derived from `PassBase`. It is used to wrap the existing C++ graph-based passes. Developers can call these passes using the same method as other Python passes.
    - `PassManager`: a helper class to do the automatic pass combination and solving conflicts.
    - Python pass example: `FuseAllReducePass`.
    - C++ graph-based example: `FuseElementwiseAddActPass`.
- Add the distributed pass test framework `DistPassTestBase`.
- Add `split_program` utils to show how to write the pass utility tools.


python/paddle/distributed/passes/pass_base.py
```diff
@@ -0,0 +1,273 @@
...
+
+    @abstractmethod
+    def _check_conflict(self, other_pass):
+        pass
+
+    def _check_conflict_including_common_rules(self, other_pass):
+        return self._check_conflict(other_pass) and all(
+            [r(other_pass, self) for r in PassBase._COMMON_RULES])
+
+    def apply(self, main_programs, startup_programs, context=None):
```
JZ-LIANG, line 74: what about to have 4 apply methods: apply_forward, apply_backward, apply_optimize, and apply(include above three) 
1: it would more nature to for the implementation of amp & recompute pass, since they will modify the forward and backward program with different logic.
2: if we want to generate a program with forward ops only (for inference ), it will more convenient we could call amp.apply_forward for this program instead of amp.apply

python/paddle/distributed/passes/pass_base.py
```diff
@@ -0,0 +1,273 @@
...
+
+        assert isinstance(main_programs, list)
+        assert isinstance(startup_programs, list)
+        assert len(main_programs) == len(startup_programs)
+        self._apply_impl(main_programs, startup_programs, context)
+        context._add_pass(self)
+        return context
+
+    def _apply_impl(self, main_programs, startup_programs, context):
+        for main_program, startup_program in zip(main_programs,
```
JZ-LIANG, line 95: it is a little bit weird that handle the traversal of rank inside PassBase, it should hand over to user or outside framework logic which call instantiate PassBase. 

for instance, in Autoparallel there will be a Parallelizer to hanlde this logic.

python/paddle/distributed/passes/pass_base.py
```diff
@@ -0,0 +1,273 @@
...
+
+import six
+import sys
+from abc import ABC, abstractmethod
+from paddle.fluid.framework import program_guard, _apply_pass as _apply_cpp_pass
+
+
+class PassContext:
+    def __init__(self):
+        self._applied_passes = []
```
JZ-LIANG, line 23: it would be better that _applied_passes is a  dict instead of a list, so as that we cold call the applied pass later. 
for example sharding pass might need some infomation from amp pass, so it call fetch that info from passcontext by name of amp pass

fuyinno4 (approved):

python/paddle/distributed/fleet/launch_utils.py
```diff
@@ -518,7 +530,11 @@ def start_local_trainers(cluster,
 
         current_env.update(proc_env)
 
-        cmd = [sys.executable, "-u", training_script] + training_script_args
+        coverage_args = []
+        if run_with_coverage():
```
aoyulong, line 534: What is the `run_with_coverage` used for? Only used for pass testing？

python/paddle/distributed/passes/pass_base.py
```diff
@@ -0,0 +1,273 @@
...
+    def _check_conflict(self, other_pass):
+        return True
+
+    def _apply_single_impl(self, main_program, startup_program, context):
+        _apply_cpp_pass(main_program, startup_program, self.cpp_name,
+                        self._attrs, self.cpp_attr_types)
+
+
+# Like AutoParallel/HybridParallel, etc. 
+class ParallelOptPass(PassBase):
```
aoyulong, line 146: From the current situation of the auto-parallel pass optimizations, we tend to directly inherit the PassBase and not use the `ParallelOptPass`, `CalcOptPass` and `CommOptPass`.

chenwhql (approved):

aoyulong (approved):

python/paddle/fluid/tests/unittests/distributed_passes/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+file(GLOB TEST_OPS RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" "test_*.py")
+string(REPLACE ".py" "" TEST_OPS "${TEST_OPS}")
+
+foreach(TEST_OP ${TEST_OPS})
+    py_test_modules(${TEST_OP} MODULES ${TEST_OP})
+    list(APPEND DIST_TEST_OPS ${TEST_OP})
+    set_tests_properties(${TEST_OP} PROPERTIES TIMEOUT 120)
```
XieYunshen, line 7: TIMEOUT值下个PR改为60S

XieYunshen (approved):
LGTM for `set_tests_properties(test_split_program PROPERTIES TIMEOUT 120)`

python/paddle/distributed/fleet/launch_utils.py
sneaxiy: Yes.

python/paddle/distributed/passes/pass_base.py
sneaxiy: Would fix it in the next PR.

python/paddle/fluid/tests/unittests/distributed_passes/CMakeLists.txt
sneaxiy: Would fix it in the next PR.

raindrops2sea (approved):

PR #36658, opened 2021-10-23
[bugfix]remove ctc grad norm flag, which is not needed

zh794390558:
### PR types

Bug fixes
### PR changes

APIs
### Describe

1. remove ctc grad norm flag, which is not needed


zh794390558:
https://github.com/PaddlePaddle/docs/pull/3709

TCChenlong (approved):

PR #36667, opened 2021-10-23
fix pool2d convert case

JZZ-NOTE:
### PR types

Others
### PR changes

Others
### Describe

pool2d convert case 添加 skip，需要后续修复


jiweibo (approved):
LGTM

PR #36698, opened 2021-10-25
Fix the null ptr bug in build_cinn_pass.

wzzju:
### PR types

Bug fixes

### PR changes

Others

### Describe

Fix the null ptr bug in build_cinn_pass.


thisjiang (approved):
LGTM

PR #36711, opened 2021-10-25
Correct conv2d int8 mkldnn UT

wozna:
### PR types

Bug fixes
### PR changes

OPs 
### Describe

In this PR the unit test of the covolution has been fixed:
- There was an error in the test that in the calculation of the reference convolution the `input` and `input_shift` were multiplied by `Scale_in`, and for the tested OneDNN, the input was not multiplied. This caused this test to not work properly when `Scale_in` was other than 1.0. Therefore, multiplication was removed, so we assume that input is already quantized.
- In the test, the weight scale was multiplied by 0.5. However, this situation should only occur when using Intel AVX2 or Intel AVX512 to avoid a potential overflow. If the processor supports VNNI instructions, the weight scales should remain unchanged.

The changes were adjusted based on documentation https://docs.oneapi.io/versions/latest/onednn/dev_guide_int8_computations.html#doxid-dev-guide-int8-computations

### DIFF
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -226,6 +226,17 @@ bool SupportsBfloat16FastPerformance() {
 #endif
 }
 
+bool SupportsVNNI() {
+#ifndef PADDLE_WITH_MKLDNN
+  return false;
+#else
+  if (platform::MayIUse(platform::cpu_isa_t::avx512_core_vnni))
+    return true;
+  else
+    return false;
+#endif
+}
+
 // According to the input `place` and `dtype`, this function returns a tuple
 // consists of three sets:
 // 1) All operators registered in the Paddle framework.
@@ -2113,6 +2124,7 @@ All parameter, weight, gradient are variables in Paddle.
   m.def("_is_compiled_with_heterps", IsCompiledWithHETERPS);
   m.def("supports_bfloat16", SupportsBfloat16);
   m.def("supports_bfloat16_fast_performance", SupportsBfloat16FastPerformance);
+  m.def("supports_vnni", SupportsVNNI);
   m.def("op_supported_infos", OpSupportedInfos);
   m.def("is_compiled_with_brpc", IsCompiledWithBrpc);
   m.def("is_compiled_with_dist", IsCompiledWithDIST);
--- a/paddle/fluid/pybind/pybind.cc
+++ b/paddle/fluid/pybind/pybind.cc
@@ -226,14 +226,20 @@ bool SupportsBfloat16FastPerformance() {
 #endif
 }
 
+bool SupportsInt8() {
+#ifndef PADDLE_WITH_MKLDNN
+  return false;
+#else
+  return (platform::MayIUse(platform::cpu_isa_t::avx2) ||
+          platform::MayIUse(platform::cpu_isa_t::avx512f));
+#endif
+}
+
 bool SupportsVNNI() {
 #ifndef PADDLE_WITH_MKLDNN
   return false;
 #else
-  if (platform::MayIUse(platform::cpu_isa_t::avx512_core_vnni))
-    return true;
-  else
-    return false;
+  return platform::MayIUse(platform::cpu_isa_t::avx512_core_vnni);
 #endif
 }
 
@@ -2124,6 +2130,7 @@ All parameter, weight, gradient are variables in Paddle.
   m.def("_is_compiled_with_heterps", IsCompiledWithHETERPS);
   m.def("supports_bfloat16", SupportsBfloat16);
   m.def("supports_bfloat16_fast_performance", SupportsBfloat16FastPerformance);
+  m.def("supports_int8", SupportsInt8);
   m.def("supports_vnni", SupportsVNNI);
   m.def("op_supported_infos", OpSupportedInfos);
   m.def("is_compiled_with_brpc", IsCompiledWithBrpc);
--- a/python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_int8_mkldnn_op.py
+++ b/python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_int8_mkldnn_op.py
@@ -23,8 +23,7 @@ from paddle.fluid.tests.unittests.test_conv2d_op import conv2d_forward_naive, Te
 
 
 def conv2d_forward_refer(input, filter, group, conv_param):
-    out, in_n, out_h, out_w, out_c = conv2d_forward_naive(input, filter, group,
-                                                          conv_param)
+    out, _, _, _, _ = conv2d_forward_naive(input, filter, group, conv_param)
     return out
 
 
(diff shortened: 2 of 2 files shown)


wozna:
@piotrekobiIntel @sfraczek Please review

paddle/fluid/pybind/pybind.cc
```diff
@@ -226,6 +226,17 @@ bool SupportsBfloat16FastPerformance() {
...
 }
 
+bool SupportsVNNI() {
+#ifndef PADDLE_WITH_MKLDNN
+  return false;
+#else
+  if (platform::MayIUse(platform::cpu_isa_t::avx512_core_vnni))
+    return true;
+  else
+    return false;
```
ghost:
```suggestion
  return platform::MayIUse(platform::cpu_isa_t::avx512_core_vnni);
```

python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_int8_mkldnn_op.py
```diff
@@ -94,32 +99,30 @@ def setUp(self):
                     output = np.round(output1 - output2).astype(self.dsttype)
 
         else:
+            input = np.random.randint(0, 10,
```
ghost: If I understand correctly, the implementation for uint8 differs from the one for int8 mainly by the fact that int8 needs to be offset by an array stored in output2 and a different input range. Have you considered refactoring the code by creating a function that takes the input type and input range as arguments? If the type is uint8 then output2 can simply  be set to np.zeros(len(output1)). This should make the code much shorter.

sfraczek (dismissed):
LGTM

paddle/fluid/pybind/pybind.cc
```diff
@@ -226,6 +226,17 @@ bool SupportsBfloat16FastPerformance() {
 #endif
 }
 
+bool SupportsVNNI() {
+#ifndef PADDLE_WITH_MKLDNN
+  return false;
+#else
+  if (platform::MayIUse(platform::cpu_isa_t::avx512_core_vnni))
```
sfraczek: or AVX2_VNNI?

python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_int8_mkldnn_op.py
```diff
@@ -53,39 +52,45 @@ def setUp(self):
             'pad': self.pad,
             'dilation': self.dilations
         }
+        # This implementation of convolution quantization is based on OneDNN documentation
```
arlesniak, line 55:
```suggestion
# This implementation of convolution quantization is based on OneDNN documentation
# https://oneapi-src.github.io/oneDNN/dev_guide_int8_computations.html#doxid-dev-guide-int8-computations-1dg-i8-comp-s11
scale_output_shift = (self.scale_out /
                      (self.scale_in * self.scale_weights[0]))

filter = np.random.random(self.filter_size).astype(self.weighttype)

# When the Intel AVX2 or Intel AVX512 Instruction Set is used
# the reorder additionally scales the weights by 0.5
# to overcome the potential overflow issue.
avx_scale = 0.5 if self.srctype == np.int8 and core.supports_vnni() else 1

filter_int = np.round(filter * self.scale_weights[0] * avx_scale).astype(np.int32)
scale_output_shift = scale_output_shift / avx_scale

def conv2d_forward_refer_helper(input_):
    return conv2d_forward_refer(input_.astype(np.int32), filter_int, self.groups, conv2d_param).astype(np.float32) * scale_output_shift

def residual_helper(init_low, init_high, output_):
    input_residual = np.random.randint(init_low, init_high, self.input_residual_size).astype(self.srctype)
    return output_ + input_residual * (
            self.scale_out / self.scale_in_eltwise)

if self.srctype == np.int8:
    init_low, init_high = (-5, 5)
    input = np.random.randint(init_low, init_high, self.input_size).astype(self.srctype)
    input_shift = (np.ones(self.input_size) * 128).astype(np.uint8)

    output1 = conv2d_forward_refer_helper(np.round(input + input_shift))
    output2 = conv2d_forward_refer_helper(np.round(input_shift))
    output = output1 - output2
else:
    init_low, init_high = (0, 10)
    input = np.random.randint(init_low, init_high, self.input_size).astype(self.srctype)
    output = conv2d_forward_refer_helper(input)

if self.fuse_residual:
    output = residual_helper(init_low, init_high, output)

output = np.round(output)

if self.fuse_activation == "relu":
    output = np.maximum(output, 0)

output = output.astype(self.dsttype)
```

arlesniak: Maybe some refactoring would make it a bit more readable. Perhaps as above, please treat it just as a suggestion. Of course it can be done in another PR. What is your opinion ?

wozna: Yes you are right, your changes make it more readable. I will do it in this PR

python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_int8_mkldnn_op.py
wozna: You're right, I'll try this approach

paddle/fluid/pybind/pybind.cc
wozna: You're right, avx2_VNNI should also be here. To add this, I have to modify the MayIUse function as it does not have such a check.

wozna: AVX2_VNNI checking is not available in the Xbyak version, which is now available in Paddle. In order to do this, you need to upgrade Xbyak. I will do a separate task to do this.

sfraczek (approved):
LGTM

arlesniak (approved):
LGTM

python/paddle/fluid/tests/unittests/mkldnn/test_conv2d_int8_mkldnn_op.py
```diff
@@ -23,13 +23,12 @@
...
 def conv2d_forward_refer(input, filter, group, conv_param):
-    out, in_n, out_h, out_w, out_c = conv2d_forward_naive(input, filter, group,
-                                                          conv_param)
+    out, _, _, _, _ = conv2d_forward_naive(input, filter, group, conv_param)
     return out
 
 
-@unittest.skipIf(not core.supports_bfloat16(),
-                 "place does not support BF16 evaluation")
+@unittest.skipIf(not core.supports_int8(),
```
jakpiase, line 30: If we don't want to be stopped by CI-Approval, you can use OpTestTool to skip these tests if int8 is not supported by place. The behavior is identical to using pure "@unittest.skipIf", but since it is a wrapper regex cannot catch that and CI-Approval does not complain about that
`
@OpTestTool.skip_if(not core.supports_int8(), "place does not support int8 computation")
`

lidanqing-vv (approved):
LGTM

kolinwei (approved):

PR #36714, opened 2021-10-25
add new API/OP:  paddle.linalg.triangular_solve

zhwesky2010:
### PR types

New features
### PR changes

APIs
### Describe

Computes the solution of a system of equations with a triangular coefficient matrix `x` and
multiple right-hand sides `y` .

Input `x` and `y` is 2D matrices or batches of 2D matrices. If the inputs are batches, the outputs
is also batches.

Args:

    x (Tensor): The input triangular coefficient matrix. Its shape should be [*, M, M], where * is zero or
        more batch dimensions. Its data type should be float32 or float64.
    y (Tensor): Multiple right-hand sides of system of equations. Its shape should be [*, M, K], where * is 
        zero or more batch dimensions. Its data type should be float32 or float64.
    upper (bool, optional): Whether to solve the upper-triangular system of equations (default) or the lower-triangular 
        system of equations. Default: True.
    transpose (bool, optional): whether x should be transposed before calculation. Default: False.
    unitriangular (bool, optional): whether x is unit triangular. If True, the diagonal elements of x are assumed 
        to be 1 and not referenced from x . Default: False.
    name(str, optional): Name for the operation (optional, default is None).

Returns:

    Tensor: The solution of the system of equations. Its data type should be the same as that of `x`.

Examples:

    # A · X = b
    #
    # a square system of linear equations:
    # x1 +   x2  +   x3 = 0
    #      2*x2  +   x3 = -9
    #               -x3 = 5

    import paddle

    x = paddle.to_tensor([[1, 1, 1], 
                          [0, 2, 1],
                          [0, 0,-1]], dtype="float64")
    y = paddle.to_tensor([[0], [-9], [5]], dtype="float64")
    out = paddle.linalg.triangular_solve(x, y, upper=True, transpose=False, unitriangular=False)

    print(out)
    # [7, -2, -5]


Avin0323 (dismissed):
LGTM for PR-CI-OP-benchmark.

paddle/fluid/operators/math/matrix_solve.cc
```diff
@@ -34,6 +34,45 @@ class MatrixSolveFunctor<platform::CPUDeviceContext, T> {
...
+    CBLAS_DIAG diag = unitriangular ? CblasUnit : CblasNonUnit;
+
+    const T* a_data = a->data<T>();
+    T* b_data = b->mutable_data<T>(context.GetPlace());
+
+    int a_dim_size = a->dims().size();
+    int b_dim_size = b->dims().size();
+
+    int M = b->dims()[b_dim_size - 2];
+    int N = b->dims()[b_dim_size - 1];
```
jeff41404: Tensor->dims()[] wll return int64_t, use static_cast ?

paddle/fluid/operators/math/matrix_solve.cu.cc
```diff
@@ -163,6 +163,68 @@ class MatrixSolveFunctor<platform::CUDADeviceContext, T> {
...
+    CBLAS_DIAG diag = unitriangular ? CblasUnit : CblasNonUnit;
+
+    const T* a_data = a->data<T>();
+    T* b_data = b->mutable_data<T>(context.GetPlace());
+
+    int a_dim_size = a->dims().size();
+    int b_dim_size = b->dims().size();
+
+    int M = b->dims()[b_dim_size - 2];
+    int N = b->dims()[b_dim_size - 1];
```
jeff41404: Tensor->dims()[] wll return int64_t, use static_cast ?

paddle/fluid/operators/math/matrix_solve.cc
zhwesky2010: Done, thx

paddle/fluid/operators/math/matrix_solve.cu.cc
zhwesky2010: Done, thx

jeff41404 (approved):
lgtm

TCChenlong (approved):
LGTM

python/paddle/tensor/linalg.py
```diff
@@ -2315,6 +2315,79 @@ def solve(x, y, name=None):
...
+    Examples:
+    .. code-block:: python
+
+        # a square system of linear equations:
+        # x1 +   x2  +   x3 = 0
+        #      2*x2  +   x3 = -9
+        #               -x3 = 5
+
+        import paddle
+        import numpy as np
```
jzhang533, line 2356: numpy没用到。

zhwesky2010: 下个PR顺带删掉~ thx

XiaoguangHu01 (approved):
LG API

jzhang533 (approved):
LGTM

raindrops2sea (approved):
LGTM

PR #36727, opened 2021-10-26
roll_op: support Tensor as input for shifts

iclementine:
### PR types

Function optimization

### PR changes

Ops

### Describe

roll_op: support Tensor as input for shifts

Roll_op does not support "shifts" as a tensor input, since it is defined as an attribute, which has to ben known at graph assembly stage. This restrict the expressiveness of this function, becase the amount of shits may depend on the shape of the input or other runtime information.

```python
import paddle
paddle.enable_static()

x = paddle.arange(24).reshape([4, 6])
shifts = paddle.shape(x) // 2
y = paddle.roll(x, shifts=x, axis=[0, 1])

exe = paddle.static.Executor()
y_np = exe.run(fetch_list=[y])
```


jeff41404 (approved):
lgtm

PR #36744, opened 2021-10-26
[Auto Parallel] Completion Dist Attribute for Backward & Update stage

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Describe
Completion Dist Attribute for Backward & Update stage
Make sure all op and var have Dist Attribute


python/paddle/distributed/auto_parallel/completion.py
```diff
@@ -710,39 +690,60 @@ def _get_op_by_id(ops, id):
...
+                        input_var.name)
+                    ref_dims_mapping = input_var_dist_attr.get_dims_mapping()
+                    ref_process_mesh = input_var_dist_attr.get_process_mesh()
+
+                    # tensor dist attr
+                    output_var = vars[grad_op.desc.output(output_name)[0]]
+                    output_var_attr = TensorDistributedAttribute(output_var,
+                                                                 dist_context)
+                    output_var_attr.set_dims_mapping(ref_dims_mapping)
+                    output_var_attr.set_process_mesh(ref_process_mesh)
```
Caozhou1995: the grad var process mesh should be the same with grad op instead of forward var?

JZ-LIANG: yes， here is a bug and had been fixed

aoyulong (commented):
I have tested this for the new implementation and it works well for me now. In the future, we should give clear and reasonable rules to complete these distributed attributes for the backward and update tensors and ops.

python/paddle/distributed/auto_parallel/completion.py
```diff
@@ -659,90 +670,80 @@ def _get_op_by_id(ops, id):
...
-                if "@GRAD" in var_name:
-                    dims_mapping = dist_context.get_tensor_distributed_attr_for_program(
-                        vars[var_name]).get_dims_mapping()
-                    grad_op_attr.set_input_dims_mapping(var_name, dims_mapping)
+            grad_op_attr.set_process_mesh(forward_op_process_mesh)
+
+            # var 
+            for output_name in grad_op.desc.output_names():
+                assert len(grad_op.desc.output(output_name)) in [0, 1]
+                # if grad_op.type == "cast":
```
wangxicoding, line 697: 没用可以干掉

wangxicoding (approved):

PR #36747, opened 2021-10-26
GeneratePass support attr condition and mapping

Avin0323:
### PR types

New features

### PR changes

Others

### Describe

GeneratePass支持辅助类型下属性限定及映射。


paddle/fluid/framework/pass_desc.proto
```diff
@@ -16,20 +16,68 @@ package paddle.framework.proto;
 
 // Describes one subsitute subgraph.
 message PassDesc {
```
zhhsplendid, line 18: Note: proto can be added, but cannot be deleted after release.

It is OK to you in this PR because now we shouldn't have users, but once it is released, you cannot change it.

Please decide whether you will cherry-pick this PR, otherwise the release 2.2 will contain your old proto and it will be conflict to your new proto

zhhsplendid (commented):
Other things LGTM

paddle/fluid/framework/pass_desc.proto
zhhsplendid: My mistake, this proto is used only for Paddle data transformation, nothing will cause version mismatch during usage. So it can change at any time.

zhhsplendid (approved):
LGTM

PR #36748, opened 2021-10-26
fix fftshift/ifftshift on static mode

iclementine:
### PR types

Bug fixes
### PR changes

APIs

### Describe

fix fftshift/ifftshift on static mode. 

`fftshift`/`ifftshift` uses roll to move frequency components, thus affected by an issue that roll does not support Tensor as input ("shifts")(which is solved in #36727)

Related PR https://github.com/PaddlePaddle/Paddle/pull/36727


jeff41404 (approved):
lgtm

PR #36758, opened 2021-10-26
bugfix: only check backend when mode == Collecive

2742195759:
### PR types

Bug fixes

### PR changes

APIs

### Describe

bugfix: only check backend when mode == Collecive


Aurelius84 (approved):
LGTM

PR #36759, opened 2021-10-26
Added reshape+transpose+matmul_v2 fuse pass

sfraczek:
### PR types
Performance optimization

### PR changes
Others

### Describe
This is reshape+transpose+matmul_v2 fuse pass which is based on previous identical fuse for matmul_v1: https://github.com/PaddlePaddle/Paddle/pull/23754, this fuse will speedup bert-like models.


paddle/fluid/operators/matmul_op.cc
```diff
@@ -324,8 +324,8 @@ class MatMulGradKernel : public framework::OpKernel<T> {
   }
 };
 
-framework::DDim GetDimForInput(const framework::InferShapeContext &ctx,
-                               std::string input_name) {
+static framework::DDim GetDimForInput(const framework::InferShapeContext &ctx,
+                                      std::string input_name) {
```
jakpiase:
```suggestion
                                      const std::string& input_name) {
```
That might not be performance-critical, but it still may save us some time

paddle/fluid/operators/matmul_v2_op.cc
```diff
@@ -19,6 +19,36 @@
 namespace paddle {
 namespace operators {
 
+static framework::DDim GetDimForInput(const framework::InferShapeContext& ctx,
+                                      std::string input_name) {
```
jakpiase: Same as above

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
```diff
@@ -28,6 +28,119 @@ using Tensor = paddle::framework::Tensor;
 using paddle::framework::vectorize;
 using paddle::framework::make_ddim;
 using paddle::framework::GradVarName;
+using paddle::framework::DDim;
+
+static DDim GetDimForInput(const ExecutionContext& ctx,
+                           std::string input_name) {
```
jakpiase: Same as above

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
```diff
@@ -28,6 +28,119 @@ using Tensor = paddle::framework::Tensor;
...
+
+  std::vector<int64_t> new_x(x.size());
+  for (size_t i = 0; i < x.size(); i++) {
+    new_x[i] = x[axis[i]];
+  }
+  return new_x;
+}
+
+std::vector<int64_t> GetInputStrides(const ExecutionContext& ctx,
+                                     std::string input_name) {
```
jakpiase: Same as above

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
```diff
@@ -64,16 +179,24 @@ class MatMulV2MKLDNNHandler
     y_strides.reserve(x_dims.size());
     out_strides.reserve(x_dims.size());
 
-    if (!trans_x) {
-      x_strides.insert(x_strides.end(), {M * K, K, 1});
+    if (!strides_x.empty()) {
+      x_strides = strides_x;
```
jakpiase: Could you please change the names of these variables? Since x_strides and strides_x sounds like it's exactly the same

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
```diff
@@ -64,16 +179,24 @@ class MatMulV2MKLDNNHandler
...
+        x_strides.insert(x_strides.end(), {M * K, K, 1});
+      } else {
+        x_strides.insert(x_strides.end(), {M * K, 1, M});
+      }
     }
 
-    if (!trans_y) {
-      y_strides.insert(y_strides.end(), {N * K, N, 1});
+    if (!strides_y.empty()) {
+      y_strides = strides_y;
```
jakpiase: Same as with x_strides and strides_x

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_v2_mkldnn_op.py
```diff
@@ -434,6 +438,61 @@ def set_op_type(self):
...
+    def set_op_type_and_transpose_y_name(self):
+        self.op_type = "matmul_v2"
+        self.transpose_y_name = "trans_y"
+
+
+class TestMatMulV2OpReshapeTranspose3DYFloat(
+        TestReshapeTransposeMatMulOp3DYFloat):
+    def set_op_type_and_transpose_y_name(self):
+        self.op_type = "matmul_v2"
+        self.transpose_y_name = "trans_y"
```
jakpiase, line 493: Why only transpose_y is tested?

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
```diff
@@ -28,6 +28,119 @@ using Tensor = paddle::framework::Tensor;
...
+                        "Axis values must be ranging from 0 to (dims - 1)."));
+
+  std::vector<int64_t> new_x(x.size());
+  for (size_t i = 0; i < x.size(); i++) {
+    new_x[i] = x[axis[i]];
+  }
+  return new_x;
+}
+
+std::vector<int64_t> GetInputStrides(const ExecutionContext& ctx,
```
jakpiase, line 33:
```suggestion
static std::vector<int64_t> GetInputStrides(const ExecutionContext& ctx,
```

jakpiase: Since matmul and matmul_v2 are sharing these functions, maybe it would be nice to include just the signature in "matmul_mkldnn_op.h", so we won't need to have two copies of exactly the same function, what do you think about that?

jakpiase (commented):
Thank you very much for your contribution!

paddle/fluid/operators/matmul_op.cc
sfraczek: AFAIK GetDimForInput is always called with const char*.
I think that the string construction happens in place here based on char ptr so there is no copying of string but passing a char*. If so, then it's not necessarily slower. I will use a char argument instead 
and create a string based on it inside the function explicitly. It will show better the intent behind it. I will also validate if it's exactly either 'X' or 'Y'.

paddle/fluid/operators/matmul_v2_op.cc
sfraczek: same reply as above

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
sfraczek: same reply as above

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
sfraczek: I can move some function declarations there but some function signatures are different - those I won't move unless we figure out something else.

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
sfraczek: I was actually using input_name as read/write variable so I liked the copy/constructed object from const char*. 
I will change it to char instead as mentioned in above comment.

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
sfraczek: Ok

paddle/fluid/operators/mkldnn/matmul_v2_mkldnn_op.cc
sfraczek: ok

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_v2_mkldnn_op.py
sfraczek: Maybe it is because the only use case for bert like models had transposition of y so it was prepared for that only.

paddle/fluid/operators/matmul_op.cc
jakpiase: That sounds nice, thank you!

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -502,6 +464,85 @@ class MatMulMKLDNNKernel : public paddle::framework::OpKernel<T> {
...
+                        "Axis values must be ranging from 0 to (dims - 1)."));
+
+  std::vector<int64_t> new_x(x.size());
+  for (size_t i = 0; i < x.size(); i++) {
+    new_x[i] = x[axis[i]];
+  }
+  return new_x;
+}
+
+static framework::DDim GetDimForInput(const framework::InferShapeContext& ctx,
```
wozna: This is a similar question that Jakub drew attention to. Couldn't we dump this function to _matmul_mkldnn_op.h_ to implement mkldnn? I can see that for _matmul_mkldnn_op.cc_,  _matmul_v2_mkldnn_op.cc_ we have different contexts: `const ExecutionContext & ctx` and` const framework::InferShapeContext & ctx`, do you know why they are different?

Actually, this function has the same logic for _matmul_op.cc_, _matmul_v2_op.cc_, _matmul_mkldnn_op.cc_, _matmul_v2_mkldnn_op.cc_, which gives us four copies of this code. So maybe at least for mkldnn we can optimize it.

wozna (commented):
Great job, only have one question

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
sfraczek: I tried moving functions to header by it doesn't build on CI saying that there's duplicate definition and other problems. I cannot reproduce it on my machine, maybe when I got cuda/cudnn installed and configured then I could try to reproduce that error. Otherwise testing by  pushing to a PR is not a good way for debugging this. There are hard to understand include patterns for matul matmul v2 and mkldnn counterparts. Some functions are the same in matmul_xpu functions also. I couldn't figure it out given the time I had. I would suggest trying refactoring later in a separate PR depending on priorities. I think we could re-think the namespaces in those files.

They have different context I needed to use the same function in a place where I had access to found different context variable so I made it this way.

lidanqing-vv:
@sfraczek  Hi, more CIs failed, do you know why ?

sfraczek:
> @sfraczek Hi, more CIs failed, do you know why ?

Hi, yes I have reproduction thanks to Joanna and I'm working on it. The problem is with WITH_UNITY_BUILD

lidanqing-vv (approved):
LGTM. Thank you very much !

lidanqing-vv:
@sfraczek CheckPRTemplate passed. I will try to ask for approval now

lidanqing-vv:
This PR will need two approvals. 
Because it exceeds 20 files and added new attributes in the native op.
```
2021-11-05 22:08:10 2021-11-05 22:08:10 (3.12 MB/s) - 已保存 “bk.txt” [5/5])
2021-11-05 22:08:17 ****************
2021-11-05 22:08:17 0. You must have Dianhai approval for change 20+ files or add than 1000+ lines of content.
2021-11-05 22:08:17 1. You must have one RD (Avin0323(Recommend) or zhouwei25 or wanghuancoder or luotao1) approval for modifying unity_build_rule.cmake which the rules of Unity Build.
2021-11-05 22:08:17 There are 2 approved errors.
2021-11-05 22:08:17 ****************
```

sfraczek:
on fp32 model by running cpu_infer.py from https://github.com/PaddlePaddle/Paddle/issues/36962 with commented out passes
```
      // "map_matmul_v2_to_mul_pass",              //
      // "map_matmul_v2_to_matmul_pass",           //
      // "map_matmul_to_mul_pass",                 //
```
Built with RelWithDebInfo
after (this branch)
acc: 0.5567, time:  41.98263645172119
before (develop):
acc: 0.5567, time: 42.94174814224243
on i9 desktop machine

From log of fuses:
I1110 10:33:40.420223 28072 fuse_pass_base.cc:57] ---  detected 6 subgraphs
---    Fused 6 ReshapeTransposeMatmul patterns for matmul_v2 Op with reshape's xshape with transpose's xshape

lidanqing-vv:
@baoachun Hi, 这个PR可以approve然后合入吗？这个PR实现了matmul_v2 fuse, 而且已经在真实Bert 模型#36962 提供的模型上测过了，对性能有提升，对精度无影响。这个PR超过20个文件，因为涉及的文件比较多，但是单测和真实模型都测过了，可以合其实。

jakpiase (dismissed):
LGTM, really good work :)

jczaja (dismissed):
LGTM

lidanqing-vv (dismissed):
LGTM

lidanqing-vv:
@baoachun  Please send more models with broadcasting. Thanks

lidanqing-vv:
@sfraczek  Baidu require to split this PR

sfraczek:
opened new version https://github.com/PaddlePaddle/Paddle/pull/37847

PR #36775, opened 2021-10-27
static cost model

huangxu96:
### PR types

New features
### PR changes

Others
### Describe

Add Static CostModel. Static data is based on op benchmark system


python/paddle/cost_model/cost_model.py
```diff
@@ -53,17 +54,33 @@ def profile_measure(self,
...
-        print("core:<<<<<<<")
 
         cost_model = core.CostModel()
         cost_data = cost_model.ProfileMeasure(device)
-        # cost_list = self.stop_cost_model()
-        # return cost_list
 
+    def static_cost_data(self):
+        static_cost_data_path = os.path.join(
+            os.path.dirname(__file__), "static_op_benchmark.json")
```
zhhsplendid, line 63: A little suggestion, "static_op_benchmark.json" can be set as a constant at head of file. In the code, we may not find it easily. However, you can do it in next PR

python/paddle/cost_model/cost_model.py
```diff
@@ -53,17 +54,33 @@ def profile_measure(self,
...
+        with open(static_cost_data_path, 'r') as load_f:
+            load_dict = json.load(load_f)
```
zhhsplendid, line 65: Just a question, nothing to change: how large is "static_op_benchmark.json" ?

python/paddle/cost_model/cost_model.py
```diff
@@ -53,17 +54,33 @@ def profile_measure(self,
...
+        if op_name == None:
+            raise ValueError(
+                'op_name should not be empty when you want to get static op time'
+            )
 
-cost_model.profile_measure(startup_program, main_program)
+        op_cost = {}
+        for op_data in self._static_cost_data:
+            if (op_data["op"] == op_name) and (dtype in op_data["config"]):
+                if (forward == True):
```
zhhsplendid: Clean the code:

`if forward:`

is enough

python/paddle/cost_model/cost_model.py
```diff
@@ -53,17 +54,33 @@ def profile_measure(self,
...
+        static_cost_data_path = os.path.join(
+            os.path.dirname(__file__), "static_op_benchmark.json")
+        with open(static_cost_data_path, 'r') as load_f:
+            load_dict = json.load(load_f)
+        self._static_cost_data = load_dict
+        # return all static cost data
+        return load_dict
 
-cost_model = CostModel()
+    def get_static_op_time(self, op_name, forward=True, dtype="float32"):
```
zhhsplendid, line 70: Please add comment about `forward`, otherwise the parameter is not obvious.

huangxu96: Done.

python/paddle/cost_model/cost_model.py
huangxu96: Done.

python/paddle/cost_model/cost_model.py
huangxu96: 550KB.

python/paddle/cost_model/cost_model.py
huangxu96: Cool, will do it in the next PR.

python/paddle/cost_model/cost_model.py
zhhsplendid: OK to me

zhhsplendid (approved):
LGTM

PR #36778, opened 2021-10-27
[Bug Fix]: "epsilon" attribute cannot parse into cpp code from python API

JamesLim-sy:
### PR types

Bug fixes
### PR changes

OPs
### Describe

The `epsilon` attribute is not supported in parserd into cpp codes.


sneaxiy (approved):
LGTM

PR #36797, opened 2021-10-27
Modify Struct into Class to improve encapsulation and Polish code exception

Aurelius84:
### PR types

Others
### PR changes

Others
### Describe

Modify Struct into Class to improve encapsulation and Polish code exception

+ `VariableScope` 、`Insttruction`  et.al  are modified from struct into Class
+ Polish code in executor.
+ Reuse `OpFuncNode` in `Instruction`


wanghuancoder (approved):
LGTM

PR #36837, opened 2021-10-28
[Bug fixes] Add default arg to enhance varbase ClearGradient func

veyron95:
### PR types
Bug fixes 

### PR changes
APIs 

#### 1. Add arg to enhance varbase ClearGradient func,
use case: 1
```
 linear.weight.clear_gradient() #  same as the original version.
```
use case: 2
```
 linear.weight.clear_gradient(False)   # False means to real clear gradient
 # default args: set_to_zero = True
```

#### 2. Expose two func to set and get the " is_empty_ " attribute of grad_var_
- _is_gradient_set_empty()
- _gradient_set_empty(is_empty = False) or _gradient_set_empty(is_empty = True) , default=True
```
@dygraph_guard
    def test_tensor_method_clear_gradient_case2(self):
        input = paddle.uniform(self.input_shape)
        linear = paddle.nn.Linear(2, 3)
        out = linear(input)
        out.backward()
        # default arg set_to_zero is true
        # so, False means real clear gradient
        linear.weight.clear_gradient(False)

        # before ._gradient_set_empty(False), 
        # the return of ._is_gradient_set_empty() should be True
        self.assertTrue(linear.weight._is_gradient_set_empty())

        # reset, because ClearGradient will call SetIsEmpty(True), but this is not our expectation.
        linear.weight._gradient_set_empty(False)
        # after ._gradient_set_empty(False), 
        # the return of ._is_gradient_set_empty() should be False
        self.assertFalse(linear.weight._is_gradient_set_empty())

        # actual result
        gradient_actual = linear.weight.grad
        # expected result
        self.assertTrue(np.empty(gradient_actual))

```

### DIFF
--- a/paddle/fluid/imperative/layer.cc
+++ b/paddle/fluid/imperative/layer.cc
@@ -186,7 +186,7 @@ size_t VarBase::GradOpNum() const {
   return grad_node_ ? grad_node_->size() : 0;
 }
 
-void VarBase::ClearGradient() {
+void VarBase::ClearGradient(bool set_to_zero) {
   VLOG(4) << "ClearGradient " << Name();
   if (grad_var_) {
     if (grad_var_->Var().IsType<framework::SelectedRows>()) {
@@ -204,7 +204,7 @@ void VarBase::ClearGradient() {
       auto* grad_t =
           grad_var_->MutableVar()->GetMutable<framework::LoDTensor>();
       if (grad_t->IsInitialized()) {
-        if (!FLAGS_real_release) {
+        if (set_to_zero) {
           auto* dev_ctx =
               platform::DeviceContextPool::Instance().Get(grad_t->place());
           operators::math::set_constant(*dev_ctx, grad_t, 0.0);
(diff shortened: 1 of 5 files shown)


Baibaifan (dismissed):
LGTM for collective

paddle/fluid/imperative/layer.cc
```diff
@@ -219,6 +222,23 @@ void VarBase::ClearGradient() {
   }
 }
 
+void VarBase::_GradientSetEmpty(bool set_is_empty) {
```
JiabinYang: `set_is_empty` -> `is_empty`

paddle/fluid/imperative/layer.cc
```diff
@@ -219,6 +222,23 @@ void VarBase::ClearGradient() {
...
+  VLOG(4) << "Gradient " << Name() << " SetIsEmpty " << set_is_empty;
```
JiabinYang: same

paddle/fluid/imperative/layer.cc
```diff
@@ -219,6 +222,23 @@ void VarBase::ClearGradient() {
...
+  if (grad_var_) {
+    grad_var_->SharedVar()->SetIsEmpty(set_is_empty);
+  }
+}
+
+bool VarBase::_IsGradientSetEmpty() {
+  bool res = true;
+  if (grad_var_) {
+    res = grad_var_->SharedVar()->is_empty_;
```
JiabinYang: not safe an access of SharedVar, check if it's nullptr first

JiabinYang (commented):
some comments

paddle/fluid/imperative/layer.cc
veyron95: Done.

paddle/fluid/imperative/layer.cc
veyron95: Done.

paddle/fluid/imperative/layer.cc
veyron95: Done.

TCChenlong (approved):
LGTM

JiabinYang (approved):
LGTM

PR #36861, opened 2021-10-29
heterogenous mix training under dygraph mode

bliu3650:
### PR types
New features

### PR changes
Others

### Describe

Heterogenous mix training represents the model training with heterogenous hardwares. Dygraph mode is only supported now. GPU/NPU/XPU are targeting devices for this prototype work.

The basic idea is very similar as the use of hierarchical communication topology. The low layer reduce the data within each node, while the upper layer reduce across all global nodes. Each heterogenous node has different (same first for simplicity) number of cards with same architecture. 

bj-65ac514c594f76ec4ecb447946df75e8e866f95c

The general communication flow are as followings:
- Gradient tensors would be synchronized internally first with Reduce;
- Device tensors on local rank 0 of each node would be copied to the host memory of that node;
- CPU tensors would be synchronized across all different nodes with AllReduce operation with Gloo;
- CPU tensors would be then copied back to the device memory of the same local rank 0;
- Gradient tensor finally would be broadcasted to all other ranks internally.

Users only need to change one line in their distributed training script to enable this feature:
`  - fleet.init(is_collective=True)`
`  + dist_strategy = fleet.DistributedStrategy()`
`  + dist_strategy.heter_ccl_mode = True`
`  + fleet.init(is_collective=True, strategy=dist_strategy)`

Instructions to launch the heterogenous mix training from every individual node: (e.g. two nodes with two cards on each node)
`fleetrun --backend heter --ips "gpu_node_ip,npu_node_ip" --gpus "0,1" train_fleet_dygraph.py`
`fleetrun --backend heter --ips "gpu_node_ip,npu_node_ip" --npus "0,1" train_fleet_dygraph.py`

https://user-images.githubusercontent.com/21085266/140508741-bcded441-1986-4d5c-a189-d65b6a27f3c7.mp4


paddle/fluid/imperative/gloo_context.cc
```diff
@@ -195,6 +201,18 @@ void GLOOParallelContext::AllReduce(const framework::SelectedRows &src,
   }
 }
 
+void GLOOParallelContext::InterReduce(const framework::Variable &src,
+                                      framework::Variable *dst, int ring_id) {
+  // no need to implement.
```
gongweibao: 1.CPU InterReduce
2.other ERROR?

paddle/fluid/imperative/gloo_context.cc
```diff
@@ -195,6 +201,18 @@ void GLOOParallelContext::AllReduce(const framework::SelectedRows &src,
...
+  return;
+}
+
+void GLOOParallelContext::InterBroadCast(framework::Variable *src,
+                                         int ring_id) {
+  // no need to implement.
```
gongweibao: 同上

paddle/fluid/imperative/gloo_context.cc
```diff
@@ -195,6 +201,18 @@ void GLOOParallelContext::AllReduce(const framework::SelectedRows &src,
   }
 }
 
+void GLOOParallelContext::InterReduce(const framework::Variable &src,
+                                      framework::Variable *dst, int ring_id) {
+  // no need to implement.
+  return;
+}
+
+void GLOOParallelContext::InterBroadCast(framework::Variable *src,
```
gongweibao, line 210: 静态图怎么用它。

gongweibao (changes requested):

PR #36891, opened 2021-11-01
negative label in softmax cross entropy

xingfeng01:
### PR types
Function optimization

### PR changes
OPs

### Describe
Op softmax_with_cross_entropy, "labels" is one input. In doc, this should be zero or positive integer. This PR enforce the behavior when labels is negative. In previous code, the address is computed using labels, when it is negative, the address idx could be out of bound. 

The behavior is as following in this PR.
- labels is not negative: no modifications compared to previous code
- labels is negative: return loss[ids] as zero

In code, firstly check if label is zero, if yes, return 0.


yghstill (approved):

XiaoguangHu01 (approved):
LGTM

PR #36917, opened 2021-11-01
Optimized and refactored oneDNN layer_norm kernel

jakpiase:
### PR types
Others

### PR changes
OPs

### Describe
Layernorm was copying scaleshift data two times, which was unnecessary. I have deleted that and I have also refactored this code. Mutable->data was also called two times, so I have removed it as well

### DIFF
--- a/paddle/fluid/operators/mkldnn/layer_norm_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/layer_norm_mkldnn_op.cc
@@ -44,15 +44,18 @@ class LayerNormMKLDNNHandler : public platform::MKLDNNHandlerNoCachingT<
   }
 
   std::shared_ptr<dnnl::memory> AcquireScaleShiftMemory(
-      std::vector<float>& scaleshift_data) {
-    // scaleshift_data comes from temporary buffer so we need to copy it into
-    // created memory primitivie
-    auto scaleshift_mem =
+      const Tensor* scale, const Tensor* shift) {
+    // OneDNN requires a single piece of memory for scale and shift data
+    const unsigned int C = framework::vectorize(scale->dims())[0];
+
+    auto scaleshift_memory =
         this->AcquireMemoryFromPrimitive(this->fwd_pd_->weights_desc());
-    auto data_ptr = scaleshift_mem->get_data_handle();
-    std::size_t num_bytes = scaleshift_data.size() * sizeof(float);
-    std::memcpy(data_ptr, scaleshift_data.data(), num_bytes);
-    return scaleshift_mem;
+
+    auto mem_data_handle =
+        reinterpret_cast<float*>(scaleshift_memory->get_data_handle());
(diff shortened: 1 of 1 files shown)


jakpiase:
@tsocha Please review that PR

paddle/fluid/operators/mkldnn/layer_norm_mkldnn_op.cc
```diff
@@ -133,20 +132,7 @@ class LayerNormMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
 
     std::shared_ptr<mkldnn::memory> scaleshift_memory;
```
lidanqing-vv: Can you put this `std::shared_ptr<mkldnn::memory> scaleshift_memory;` inside `if (with_scaleshit)` ?

paddle/fluid/operators/mkldnn/layer_norm_mkldnn_op.cc
```diff
@@ -133,20 +132,7 @@ class LayerNormMKLDNNOpKernel : public paddle::framework::OpKernel<T> {
...
-      // data
-      std::vector<float> scaleshift_data;
-      scaleshift_data.reserve(2 * C);
-      scaleshift_data.insert(scaleshift_data.begin(), scale->data<float>(),
-                             scale->data<float>() + C);
-
-      scaleshift_data.insert(scaleshift_data.end(), bias->data<float>(),
-                             bias->data<float>() + C);
-
-      scaleshift_memory = handler.AcquireScaleShiftMemory(scaleshift_data);
```
lidanqing-vv, line 149: I saw 
```
// TODO(jczaja): Enable FP32 when performance is good
REGISTER_OP_KERNEL(layer_norm, MKLDNN, ::paddle::platform::CPUPlace,
                   ops::LayerNormMKLDNNOpKernel<paddle::platform::bfloat16>);
```
in the last few lines of this file. So does this PR improve the performance and hence could register FP32 kernel layer_norm ?

paddle/fluid/operators/mkldnn/layer_norm_mkldnn_op.cc
jakpiase: Ok

paddle/fluid/operators/mkldnn/layer_norm_mkldnn_op.cc
jakpiase: Unfortunately not. I have deleted one unnecessary copying of scaleshift tensor, but that have improved the performance by around ~1%, so this PR is more like refactoring. Our layernorm is still almost 2x slower than native kernel and that will need further investigation

jczaja (approved):
LGTM, you could generate flamegraph (without LN vs with LN) to see if any problem is revealed. And if not then maybe run Top down microarchitectue analysis to see where is ti bound. Anyway nice PR.

tsocha (approved):

jakpiase:
@jczaja please merge that PR

PR #36958, opened 2021-11-03
[AutoParallel] Add AutoConvert

zhaoyinglia:
### PR types

New features
### PR changes

Others
### Describe

Support parameters auto-convert when parallel degree or the number of devices change.


python/paddle/distributed/auto_parallel/utils.py
```diff
@@ -472,28 +539,408 @@ def load_distributed_checkpoint(checkpoint_path,
...
+
     else:
         raise ValueError("'checkpoint_path' can not be None.")
 
-    program.set_state_dict(state_dict) if program else \
-        warnings.warn("'Program' is None, parameters will not be loaded.")
 
-    if "addition_info" not in state_dict_info:
-        return
+def load_parameter_into_program(param_dict, program):
```
Caozhou1995, line 584: Would it be better to check param_dict whether None? and if param_dict is None, this api will raise an error or init the param value?

python/paddle/distributed/auto_parallel/utils.py
```diff
@@ -362,138 +362,614 @@ def make_data_unshard(dist_main_prog, dist_startup_prog, dist_context=None):
...
+        for name, attr in dist_attr["model"].items():
+            if name not in total_dist_attr:
+                total_dist_attr[name] = attr
+
+    return total_dist_attr
+
+
+def _save_distributed_state_dict(program, addition_info, checkpoint_path):
+    """ Save parameters' state_dict """
+    rank = paddle.distributed.get_rank()
```
JZ-LIANG, line 629: aware that  the paddle.distributed.get_rank() method return the physical rank.
by now in autoparallel we don't distinguish with physical rank id and logical rank id, they are the same value.
but later when we the physical process mapping module is merged, here should be update accordingly.

python/paddle/distributed/auto_parallel/utils.py
```diff
@@ -362,138 +362,614 @@ def make_data_unshard(dist_main_prog, dist_startup_prog, dist_context=None):
...
+        param_dict(dict): parameters' name and value.
+        program(Program): the program to be updated
+    """
+    _check_param_dict(param_dict)
+    assert program and isinstance(program, paddle.fluid.framework.Program)
+
+    program.set_state_dict(param_dict)
+
+
+def _save_distributed_attribute(program, dist_attr_path):
```
JZ-LIANG: should have dist_context as an argument for save_distributed_checkpoint and _save_distributed_attribute (at least optional)
since dist_context is an optional argument for inner function "get_dist_attr", you should allow user to give the dist_context they want to save. 
using the global default dist_context could not cover all scenario.

python/paddle/distributed/auto_parallel/utils.py
```diff
@@ -362,138 +362,614 @@ def make_data_unshard(dist_main_prog, dist_startup_prog, dist_context=None):
...
+    """
+    assert _check_valid_path(checkpoint_path), \
+        "'checkpoint_path' cannot be None."
+    assert _check_valid_path(dist_attr_path), \
+        "'dist_attr_path' cannot be None."
+
+    state_dict_info = _load_distributed_state_dict(checkpoint_path)
+    dist_attr = _load_distributed_attribute(dist_attr_path)
+    param_dict = state_dict_info["model"]
+    addition_info = state_dict_info["addition_info"]
```
JZ-LIANG, line 530: should also assert that if  addition_info is the same among all rank

python/paddle/distributed/auto_parallel/utils.py
```diff
@@ -362,138 +362,614 @@ def make_data_unshard(dist_main_prog, dist_startup_prog, dist_context=None):
...
+    return param_dict, dist_attr, addition_info
+
+
+def load_checkpoint_into_program(checkpoint_path, dist_attr_path, program):
```
JZ-LIANG: should also has dist_context as an optional argument.
global default dist_context would not cover all the case.

python/paddle/fluid/tests/unittests/CMakeLists.txt
```diff
@@ -1035,6 +1037,7 @@ if(WITH_DISTRIBUTE AND WITH_GPU AND WITH_NCCL)
     set_tests_properties(test_parallel_margin_cross_entropy PROPERTIES TIMEOUT 120)
     set_tests_properties(test_auto_parallel_data_unshard PROPERTIES TIMEOUT 120)
     set_tests_properties(test_auto_parallel_save_load PROPERTIES TIMEOUT 120)
+    set_tests_properties(test_auto_parallel_autoconvert PROPERTIES TIMEOUT 120)
```
XieYunshen, line 1040: 可以将timeout值改小一些，根据任务中的执行时间(最大14S左右)，可以调整为30S

XieYunshen (approved):
LGTM，
下一个PR将会修改timeout值

python/paddle/distributed/auto_parallel/utils.py
zhaoyinglia: Done.

python/paddle/distributed/auto_parallel/utils.py
zhaoyinglia: ok, it will be updated after merging physical process mapping module.

python/paddle/distributed/auto_parallel/utils.py
zhaoyinglia: Done.

python/paddle/distributed/auto_parallel/utils.py
zhaoyinglia: addition_info is loaded from the first file of checkpoint_path, so it is the same among all ranks.

JZ-LIANG (approved):
LGTM

PR #36960, opened 2021-11-03
Fix PTen thread safety error

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
The double checking does not work! The PaddleNLP BERT model hangs because of core dump. The double checking does not guarantee:
- thread safety.
- thread visibility.

Considering the following codes:

```C++
void KernelSignatureMap::Emplace(const std::string& op_type,
                                 KernelSignature&& signature) {
  if (!Has(op_type)) { // code C1
    std::unique_lock<std::mutex> lock(mutex_);
    if (!Has(op_type)) {
      map_.emplace(op_type, signature); // code C2
    }
  }
}
```

The problems would be:
- thread safety: when C1 and C2 are executed concurrently, the data structure of the `map_` may be corrupted when C2 is running (C2 is not an atomic operation). Therefore, C1 (C1 is not an atomic operation) may occur core dump.
- thread visibility: even though C1 and C2 are not executed concurrently, the side-effect after L2 is executed (i.e., a new key-value pair is inserted) may not be seen when C1 is executed in another thread. It is because: (1) the compiler/CPU may reorder the CPU instruction. The actual execution order of the codes may not be the same as what we expect. (2) the CPU has cache for memory. Only when memory barrier is used (for example, use `std::mutex`), we would get the right value (which is changed in another thread) in the current thread.

### DIFF
--- a/paddle/fluid/framework/CMakeLists.txt
+++ b/paddle/fluid/framework/CMakeLists.txt
@@ -399,7 +399,7 @@ cc_library(save_load_util SRCS save_load_util.cc DEPS tensor scope layer)
 cc_test(save_load_util_test SRCS save_load_util_test.cc DEPS save_load_util tensor scope layer)
 cc_library(generator SRCS generator.cc DEPS enforce place)
 
-cc_library(pten_utils SRCS pten_utils.cc DEPS lod_tensor selected_rows place pten var_type_traits pten_hapi_utils)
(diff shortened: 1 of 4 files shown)


chenwhql (dismissed):
LGTM

chenwhql (approved):

PR #36980, opened 2021-11-03
Added caching of scales for bias in conv2d int8

jczaja:
### PR types
Performance optimization

### PR changes
OPs

### Describe
Currently scaling of bias data for conv2d int8 happens in every iterations. This is not needed and scaling cna be done once and then stored. This PR is caching ones computed scaled bias. 

_Perf improvement:_ Mobilenet_v1 int8 **~2%** improvement on whole model.
_processor:_ Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz

### DIFF
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -411,10 +411,17 @@ class ConvMKLDNNHandlerT
 
     bool is_multi_channel = scale_weights_data.size() > 1;
     int mask_reorder = is_multi_channel ? 1 << 0 : 1;
-    const int count =
-        is_multi_channel
-            ? (groups > 1 ? (weights_tz)[1] * (weights_tz)[0] : (weights_tz)[0])
-            : 1;
+
+    int count;
+    if (is_multi_channel) {
+      if (groups > 1) {
+        count = (weights_tz)[1] * (weights_tz)[0];
+      } else {
+        count = (weights_tz)[0];
+      }
+    } else {
+      count = 1;
+    }
 
     bias_scale_tuple =
         std::make_shared<std::tuple<float, std::vector<float>>>(std::make_tuple(
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -834,7 +834,7 @@ class ConvMKLDNNOpKernel : public framework::OpKernel<T> {
       auto p_scales_tuple = handler.get_int8_bias_scales(ctx);
 
       auto bias_memory_p = handler.AcquireBiasMemoryWithReorder(
-          bias, is_test, std::get<1>(*p_scales_tupple),
+          bias, is_test, std::get<1>(*p_scales_tuple),
           std::get<0>(*p_scales_tuple));
       args.insert({MKLDNN_ARG_BIAS, *bias_memory_p});
     }
--- a/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
@@ -389,6 +389,45 @@ class ConvMKLDNNHandlerT
     }
   }
 
+  std::shared_ptr<std::tuple<float, std::vector<float>>> get_int8_bias_scales(
+      const framework::ExecutionContext& ctx) {
+    // Get scales int8 bias key
+    const std::string key_bs = this->key_ + "@bs";
+
+    // Scales for int8 bias are to be cached to avoid
+    // computing them each iteration
+    auto bias_scale_tuple =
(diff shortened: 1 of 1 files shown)


paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -389,6 +389,45 @@ class ConvMKLDNNHandlerT
...
+    const auto* filter = ctx.Input<Tensor>("Filter");
+    const auto& weights_tz = framework::vectorize(filter->dims());
+    const int groups = std::max(ctx.Attr<int>("groups"), 1);
+
+    const auto& scale_weights_data =
+        ctx.Attr<std::vector<float>>("Scale_weights");
+    const auto& scale_in_data = ctx.Attr<float>("Scale_in");
+
+    bool is_multi_channel = scale_weights_data.size() > 1;
+    int mask_reorder = is_multi_channel ? 1 << 0 : 1;
```
jakpiase, line 413: What is the reason for bit shifting "1" zero bits? Isn't "1 << 0" equal to 1? For now that ternary operation is not needed, since no matter what value "is_multi_channel" has, the result will be equal to 1

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -389,6 +389,45 @@ class ConvMKLDNNHandlerT
...
+    const int count =
+        is_multi_channel
+            ? (groups > 1 ? (weights_tz)[1] * (weights_tz)[0] : (weights_tz)[0])
```
jakpiase: Nesting ternary operators produces code that is not necessarily easy to understand for me, could you please change that to normal if statement to avoid nesting ternary ops?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -389,6 +389,45 @@ class ConvMKLDNNHandlerT
...
+            : 1;
+
+    bias_scale_tuple =
+        std::make_shared<std::tuple<float, std::vector<float>>>(std::make_tuple(
+            static_cast<float>(mask_reorder), std::vector<float>(count)));
+    for (int i = 0; i < count; i++) {
```
jakpiase, line 426: Some time ago I would complain about "i++", but since I have read that compiler will optimize it anyway(this knowledge comes from the book that you have recommended about modern CPUs, so thank you!)

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: @wozna Could you please take this question?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: ok. Will do that

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jakpiase: From what I understand "mask_reorder" is used as flags, but even with that knowledge that doesn't seem like it's easy to understand at the first glance

jczaja: I disscussed this with @wozna and there is probably some problem here that require separate investigation (testing accuracy etc.). It will not be  solved in this PR

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: done

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -411,10 +411,17 @@ class ConvMKLDNNHandlerT
 
     bool is_multi_channel = scale_weights_data.size() > 1;
     int mask_reorder = is_multi_channel ? 1 << 0 : 1;
-    const int count =
-        is_multi_channel
-            ? (groups > 1 ? (weights_tz)[1] * (weights_tz)[0] : (weights_tz)[0])
-            : 1;
+
+    int count;
```
arlesniak: Maybe if you initiate with :
`int count =1;`
then
```
else {
      count = 1;
    }
```
won't be necessary. What is your opinion ?

jakpiase (dismissed):
LGTM

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: Ok. I incorporated your suggesstion while keeping if-else blocks as requested by @jakpiase

jakpiase (approved):
LGTM

lidanqing-vv (approved):
LGTM

PR #36987, opened 2021-11-04
Try to fix CUDA Graph H2D copy bug

sneaxiy:
### PR types
New features

### PR changes
APIs

### Describe
When there is H2D copy and the host pointer would be destructed, the CUDA Graph capturing is wrong.

This PR refines the `CUDAGraph` class to make it support skipping some GPU operations using the `platform::SkipCUDAGraphCaptureGuard`. The CUDA Graph is separated into several segments when there is H2D copy.

This PR also adds `CUDAGraph.print_to_dot_files` to print the CUDA Graph as the DOT format files.


zhiqiu (approved):
LGTM

PR #37078, opened 2021-11-09
Enable FC int8

jczaja:
### PR types
Performance optimization

### PR changes
Others

### Describe
It does enable mkldnn FC functionality be default for int8 classification workloads (internal tests)


wozna (approved):
LGTM

jakpiase (commented):
Ars you sure that fc+act pass will work correctly for int8?

wozna:
> Ars you sure that fc+act pass will work correctly for int8?

Yes, it is the same as convolution, because data is multiplied by output_scale and then all activations are applied.

jakpiase (approved):
LGTM. Thank you very much @wozna for explanation

lidanqing-vv (approved):
LGTM

PR #37079, opened 2021-11-09
Fix to #34554

jczaja:
### PR types
Bug fixes 

### PR changes
OPs

### Describe
This disabled caching of oneDNN primitives so that oneDNN cache its own elements. This change fixes the problems reported in #34554


jczaja:
@pawelpiotrowicz , @Silv3s, @tsocha , @zuzg, @piotrekobiIntel  Please review

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -61,414 +61,428 @@ static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
...
+      std::make_shared<std::tuple<float, std::vector<float>>>(std::make_tuple(
+          static_cast<float>(mask_reorder), std::vector<float>(count)));
+  for (int i = 0; i < count; i++) {
+    std::get<1>(*bias_scale_tuple)[i] = scale_in_data * scale_weights_data[i];
+  }
+
+  dev_ctx.SetBlob(key_bs, bias_scale_tuple);
+
+  return bias_scale_tuple;
+}
```
ghost, line 122: Maybe merging these 2 functions into one and doing something different depending on the result of:
`if(std::is_same<T, int8_t>::value || std::is_same<T, uint8_t>::value`
would be a better solution?
It would allow to do get rid of the remap structs that seemed quite confusing to me at first glance.

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -61,414 +61,428 @@ static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
...
+      std::get<1>(*scale_tuple)[i] =
+          static_cast<float>(static_cast<double>(scale_out_data) /
+                             (static_cast<double>(scale_in_data) *
+                              static_cast<double>(scale_weights_data[i])));
+  }
+
+  dev_ctx.SetBlob(key_s, scale_tuple);
+
+  return scale_tuple;
+}
```
piotrekobiIntel, line 183: Same here. What is your opinion?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: I have diffrent observation e.g. I think remapping is way easier for me to understand:) And also "if.." is a runtime check as c++14 does not have if contrexpr (c++17) , and remapping mechanism makes a check evaluated in compile time

ghost (dismissed):
LGTM

Silv3S (dismissed):
LGTM

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -61,414 +61,428 @@ static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
...
+    if (platform::is_int8<T>()) {
+      src_md = platform::MKLDNNMemDesc(
+          src_tz, framework::ToMKLDNNDataType(input->type()),
+          chosen_memory_format);
+      weights_md = platform::MKLDNNMemDesc(
+          weights_tz, mkldnn::memory::data_type::s8, chosen_memory_format);
+    } else {
+      src_md = platform::MKLDNNMemDesc(src_tz, data_type, chosen_memory_format);
+      weights_md = platform::MKLDNNMemDesc(weights_tz, data_type,
+                                           MKLDNNMemoryFormat::any);
```
jakpiase, line 320: You are using "chosen memory format" everywhere except that place, please unity that

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -61,414 +61,428 @@ static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
...
+    int g = std::max(groups, 1);
+    platform::GetGroupConvWeightsTz(weights_tz, g);
+    auto dst_tz = framework::vectorize(out_grad->dims());
+
+    /* create memory descriptor for conv backward without specified format
+     * ('any') which lets a primitive (conv backward in this case) choose
+     * the memory format preferred for best performance
+     */
+    const auto chosen_memory_format = MKLDNNMemoryFormat::any;
+    const auto weights_format = MKLDNNMemoryFormat::any;
```
jakpiase: What is the point of having two identical variables here? Every format im conv should use any, do I think that this redundancy is not necessary

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: good catch. Fixed

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: some legacy remaining. I removed those

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -61,414 +61,425 @@ static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
...
+      fuse_residual_conn ? scale_out_data / scale_in_eltwise_data : 1.0f;
+  int count = is_multi_channel ? (groups > 1 ? (weights_tz)[1] * (weights_tz)[0]
+                                             : (weights_tz)[0])
+                               : 1;
+  std::vector<float> output_shift_scale(count);
+
+  scale_tuple =
+      std::make_shared<std::tuple<float, std::vector<float>>>(std::make_tuple(
+          static_cast<float>(sum_scale), std::vector<float>(count)));
+  for (int i = 0; i < count; i++) {
```
lidanqing-vv, line 168: for loop could use #pragma omp parallel for?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -61,414 +61,425 @@ static mkldnn::memory::data_type GetDstType(bool is_int8, bool is_bfloat16,
...
-      const std::string fuse_activation =
-          ctx.Attr<std::string>("fuse_activation");
-      const float fuse_alpha = ctx.Attr<float>("fuse_alpha");
-      const float fuse_beta = ctx.Attr<float>("fuse_beta");
-      const bool fuse_residual_conn =
-          ctx.Attr<bool>("fuse_residual_connection");
-      const int groups = ctx.Attr<int>("groups");
-      const std::string padding_algorithm =
-          ctx.Attr<std::string>("padding_algorithm");
+    const std::string fuse_activation =
```
lidanqing-vv, line 256: Will `const auto&` avoid copy here and following assignment ?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: This generation of scales is performed only once as it is cached now. So there is no need adding openmp as it would just create multiple threads . each thread with some resources that would just be not used later on. So that is why I do not add omp parallel here

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
jczaja: This code I just touched due to indentation changes so I do not know. It sounds like a good idea to check that but I will not do that in this PR. Please add to our bugtracker.

jakpiase (dismissed):
LGTM

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
lidanqing-vv: Ok ~

lidanqing-vv (dismissed):
LGTM

jczaja:
@chenwhql  Please review and approve PR-CI-APPROVAL. Problem is again with PADDLE_ENFORCE checker that works only on 2 out of 3 lines of PADDLE_ENFORCE.

lidanqing-vv:
@baoachun Please review this

lidanqing-vv (approved):
LGTM

baoachun:
@jczaja @lidanqing-intel, This pr needs to be blocked temporarily because it involves incompatible upgrades.

jczaja:
@baoachun  There is no API change in this PR. let me know if I can merge it

jczaja:
@baoachun I want to clarify. There is no API change in this PR. SetMkldnnCacheCapacity() is still changing the capacity  of cache as it used to be. With time we will update this function : SetMkldnnCacheCapacity to also change capacity of oneDNN internal cache. Even in a future it will not impact API e.g. SetMkldnnCacheCapacity will remain.

baoachun:
@jczaja , then can I understand it like this: What you are modifying is the internal cache of oneDNN, and SetMkldnnCacheCapacity() is the cache of the modified paddle oneDNN, they are two types of cache, so the function of the SetMkldnnCacheCapacity() api will not change?

jczaja:
@baoachun Yes. this is correct

baoachun:
Hi @jczaja @lidanqing-intel,  This pr will cause the performance of the quantitative model to decrease, we need to evaluate the risks brought by the code merge.

PR #37080, opened 2021-11-09
Fix rnn grad bug in cpu when dropout is zero

joey12300:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix rnn grad bug in cpu when num_layers is set larger than 1 and dropout_prob is set 0


ZeyuChen (approved):
LGTM

PR #37082, opened 2021-11-10
[fleet_executor] fix brpc depends, more compile option

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
more brpc depend options, fix sometimes user will only compiled with distributed but not with pscore.


wangxicoding (approved):
LGTM

LiYuRio (approved):

PR #37086, opened 2021-11-10
[Cherry pick to 2.2.1]Fix rnn grad bug in cpu when dropout is zero (#37080)

joey12300:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix rnn grad bug when num_layers is set 2 and dropout_prob is set 0


ZeyuChen (approved):
LGTM

PR #37087, opened 2021-11-10
[fleet_executor] Add retry to the message bus's send. Use unique_lock instead of calling lock().

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
Add retry to the message bus's send. Use unique_lock instead of calling lock().
Lower the level of send fail's vlog to avoid too many log when retry.


wangxicoding (approved):
LGTM

PR #37091, opened 2021-11-10
[Auto Parallel]  Add the unified cluster representation

aoyulong:
### PR types
New features

### PR changes
Others

### Describe

The Cluster is used for representing the underlying diverse resources in a united way. It not only store the topology of the devices within the cluster, but also store the quantitative measurements of the device capability about computation, storage and communication.


JZ-LIANG (approved):
LGTM

PR #37094, opened 2021-11-10
[Auto Parallel] Do the physical mapping between the process graph and the cluster graph

aoyulong:
### PR types
New features

### PR changes
Others

### Describe

* First, build the cluster graph from the cluster representation.
* Second, build the process graph from the partitioned program.
* Third, do the mapping between the two graphs.


JZ-LIANG (approved):
LTGM

PR #37141, opened 2021-11-11
Add backward function hook to dygraph

jim19930609:
### PR types
New features

### PR changes
Others

### Describe
This PR enables the registration of "backward function post hook" on an OpBase. "backward function post hook" has the form of std::function<void()>, which is called immediately after the execution of the corresponding OpBase.

*This is a secrete interface to satisfy requests from the distribution team, which will not serve as a public API


JiabinYang (approved):
LGTM

wzy-99:
请问paddlepaddle什么时候支持反向传播hook函数

PR #37155, opened 2021-11-12
reduce graph-engine cache memory

seemingwang:
### PR types
Performance optimization

### PR changes
Others

### Describe
reduce graph-engine cache memory


yaoxuefeng6 (approved):
LGTM

PR #37167, opened 2021-11-13
[fleet_executor] Add sync method

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
Add sync method.

If Carrier receives messages during creating interceptors, it may encounter a situation where the target interceptor has not been initialized, then it will fall into a RunTime error (dst interceptor was not found). To avoid this happen, we add a new flag `creating_interceptors` which indicating all interceptors all still under initialization. 

For now, if Carrier  receives messages during creating interceptors, it will first push those messages into a tmp vector. After creating all interceptors, the Carrier will handle all messages in the vector.

Three usages:
#### No. 1 
User provides an init config, and won't set new interceptors later.
No extra codes are needed.

#### No. 2
User doesn't provide an init config, and will set interceptors later by calling `SetInterceptor`.
After adding all interceptors, method `SetCreatingFlag(false)` needs to be called.

#### No. 3
User provides an init config, and will set more interceptors later by calling `SetInterceptor`.
Before adding new interceptors, method `SetCreatingFlag(true)` needs to be called.
After adding all interceptors, method `SetCreatingFlag(false)` needs to be called.


paddle/fluid/distributed/fleet_executor/carrier.cc
```diff
@@ -70,16 +77,44 @@ Interceptor* Carrier::SetInterceptor(int64_t interceptor_id,
...
+    // finish create interceptors outside, handle tmp messsages
+    HandleTmpMessages();
+  }
+}
+
+void Carrier::HandleTmpMessages() {
+  VLOG(3) << "Carrier has received " << message_tmp_.size()
+          << " messages during creating interceptors.";
+  for (const auto& msg : message_tmp_) {
+    EnqueueInterceptorMessage(std::move(msg));
```
wangxicoding: 这个为啥用std::move

wangxicoding (dismissed):
LGTM

wangxicoding (approved):

PR #37181, opened 2021-11-14
update pscore fleet&comm&common_dense_table

zhaocaibei123:
### PR types

Function optimization
### PR changes

Others
### Describe

1. update push sparse/dense and client2client connection in distributed::fleet & communicator
2. add save/load in common dense table for save model in server


yaoxuefeng6 (approved):
LGTM

PR #37194, opened 2021-11-15
update dataset

zhaocaibei123:
### PR types

Function optimization
### PR changes

Others
### Describe

1. update init_param
2. fix comm=nullptr in fleet&multi trainer
3. update dataset for global shuffle


yaoxuefeng6 (approved):
LGTM

PR #37222, opened 2021-11-15
[Einsum] correct output dimension errors.

tongxin:
### PR types
Bug fixes

### PR changes
APIs

### Describe

Einsum returns a tensor with wrong dimensions when an operand has only a single element but is multidimensional. 

Example.   

```
x = paddle.rand([1, 1])
y = paddle.rand([1])
z = paddle.einsum('ij, j', x, y)
```

z is a tensor with shape []. The correct result should be in shape [1].

This bug gets fixed in this PR.


zhiqiu (approved):
LGTM

levi131 (approved):
LGTM

PR #37225, opened 2021-11-16
Add log when EnableMKLDN() and EnableMemoryOptim() are both set

lidanqing-vv:
### PR types
Others

### PR changes
Others

### Describe
When `config.EnableMKLDNN()` and `config.EnableMemoryOptim()` are both set ON, "memory_optimize_pass" will not be added, also notify users to use `config.SetMkldnnCacheCapacity(xx)` when mkldnn enabled.


lidanqing-vv:
@baoachun I checeked yesterday, when use_mkldnn_ is on, setting enable_memory_optim will not work, I added a log here, what do you think?

lidanqing-vv:
@wozna Could you please review this ? We can discuss about it too.

jakpiase (dismissed):
LGTM

wozna (dismissed):
LGTM

jczaja (dismissed):
LGTM

PR #37245, opened 2021-11-16
copy beta pow to same place when skip_update=1

zhiqiu:
### PR types

Others
### PR changes

OPs
### Describe

By default, `beta_pow` of `adam` optimizer is initialized on CPU Place.
When training on CUDA Place, the original implementation will copy it to CUDA Place when skip_update=1.

This PR fixes that.


Aurelius84 (approved):
LGTM

PR #37283, opened 2021-11-17
Fix the slow running speed of kl_div when option 'reduction' is set

LielinJiang:
### PR types

Performance optimization
### PR changes

Others
### Describe

Fix the slow running speed of kl_div when option reduction is set

image


qingqing01 (approved):
LGTM

PR #37293, opened 2021-11-17
Fix bug of paddle::experimental::MovesStorage

YuanRisheng:
### PR types

Bug fixes
### PR changes

Others
### Describe

1. Fix bugs when user moves storage from pten::DenseTensor(Src) to fluid::framework::Tensor(Dst) and Src has different type with Dst
 
image
2. support the attr that is vector<int64_t> .

### DIFF
--- a/paddle/fluid/framework/tensor.cc
+++ b/paddle/fluid/framework/tensor.cc
@@ -209,5 +209,7 @@ void Tensor::ResetHolderWithType(std::shared_ptr<memory::Allocation> holder,
   type_ = type;
 }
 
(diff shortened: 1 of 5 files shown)


paddle/fluid/framework/tensor.cc
```diff
@@ -209,5 +209,7 @@ void Tensor::ResetHolderWithType(std::shared_ptr<memory::Allocation> holder,
   type_ = type;
 }
 
+void Tensor::set_type(const proto::VarType::Type type) { type_ = type; }
```
chenwhql: const proto::VarType::Type -> const proto::VarType::Type&

PR #37296, opened 2021-11-17
[Dy2stat]Fix error when calling sublayer's non-forward func in dy2stat

0x45f:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix error when calling sublayer's non-forward func in dy2stat


Aurelius84 (approved):
LGTM

PR #37312, opened 2021-11-17
fix for cufft: some early versions of cufft do not define CUFFT_VERSION in the header

iclementine:
### PR types

Bug fixes
### PR changes

Others

### Describe

Some early versions of cufft do not define CUFFT_VERSION in the header. This PR add a branching for whether CUFFT_VSERSION is defined or not to avoid compiling error with these cufft versions.


Xreki (approved):
LGTM for op benchmark ci

jeff41404 (approved):
lgtm

PR #37317, opened 2021-11-18
Optimize cinn_cache_key by replace GraphToProgram to Dot string

thisjiang:
### PR types

Performance optimization
### PR changes

Others
### Describe

优化`cinn_cache_key` 的生成方式，通过替换非常耗时的`GraphToProgram`为简约的`Dot`字符串


paddle/fluid/framework/paddle2cinn/cinn_cache_key.cc
```diff
@@ -39,13 +44,42 @@ CinnCacheKey::CinnCacheKey(const ir::Graph& graph,
   this->SetKey(graph, input_shapes, arch_str);
 }
 
+std::string CinnCacheKey::HashGraph(const ir::Graph& graph) {
+  // using Dot to unqiue graph
+  inference::analysis::Dot dot;
+  std::unordered_map<const ir::Node*, std::string> node2dot;
```
zhhsplendid, line 48: Just a question, can we reuse id of `ir::Node` instead of creating your own id?

In that case we can save the `std::unordered_map` data structure and corresponding insert/find operations

paddle/fluid/framework/paddle2cinn/cinn_cache_key.cc
```diff
@@ -39,13 +44,42 @@ CinnCacheKey::CinnCacheKey(const ir::Graph& graph,
   this->SetKey(graph, input_shapes, arch_str);
 }
 
+std::string CinnCacheKey::HashGraph(const ir::Graph& graph) {
```
zhhsplendid: Hash usually return hash type. Can we call it `GraphHashString` or `GraphHashStr`?

Or you just return hash type at line 73 while let caller of this function make it to string.

thisjiang: Done.Changed to return `size_t`.

paddle/fluid/framework/paddle2cinn/cinn_cache_key.cc
zhhsplendid: `node.h` states that DO NOT USE node id. To be discussed in the future.

zhhsplendid (approved):
LGTM

Xreki (approved):
LGTM for op benchmark ci

PR #37332, opened 2021-11-18
make third_party's cmake get source code directly

betterpig:
### PR types

Others
### PR changes

Others
### Describe

make third_party's cmake get source code directly, since cache_third_party only wraps download command simply


cmake/external/eigen.cmake
```diff
@@ -46,9 +41,10 @@ ExternalProject_Add(
     extern_eigen3
     ${EXTERNAL_PROJECT_LOG_ARGS}
     ${SHALLOW_CLONE}
-    "${EIGEN_DOWNLOAD_CMD}"
-    PREFIX          ${EIGEN_PREFIX_DIR}
-    SOURCE_DIR      ${EIGEN_SOURCE_DIR}
+    GIT_REPOSITORY    ${EIGEN_REPOSITORY}
+    GIT_TAG           ${EIGEN_TAG}
+    PREFIX            ${EIGEN_PREFIX_DIR}
+    SOURCE_DIR        ${EIGEN_SOURCE_DIR}
```
zhwesky2010, line 47: EIGEN_SOURCE_DIR 这些也可以删掉了，这个只是EIGEN_PREFIX_DIR的拓展

zhwesky2010 (approved):

cmake/external/eigen.cmake
betterpig: OK. Only the URL download method checks whether this directory is either missing or empty before initiating the download, stopping with an error if it is not empty. All other download methods silently discard any previous contents of the source directory.

PR #37351, opened 2021-11-18
Changed second batch of deprecated mkldnn header and function names to new oneDNN names

ghost:
### PR types

Others
### PR changes

Others
### Describe

This PR changes the second batch of old mkldnn header and function names to new oneDNN ones.
These changes are required to build PaddlePaddle with future oneDNN versions (2.5+). Without this PR changes, we can not upgrade oneDNN to v2.5+ version.

The changes are as follows:
Main oneDNN include file changed from "mkldnn.hpp" to "dnnl.hpp"
mkldnn:: namespaces changed to dnnl::
MKLDNN_ARG... macros changed to DNNL_ARG...

### DIFF
--- a/paddle/fluid/framework/tensor.h
+++ b/paddle/fluid/framework/tensor.h
@@ -90,9 +90,9 @@ class Tensor {
 #ifdef PADDLE_WITH_MKLDNN
 
  public:
-  inline mkldnn::memory::format_tag format() const { return format_; }
+  inline dnnl::memory::format_tag format() const { return format_; }
 
-  inline void set_format(const mkldnn::memory::format_tag format) {
+  inline void set_format(const dnnl::memory::format_tag format) {
     format_ = format;
   }
 
@@ -106,7 +106,7 @@ class Tensor {
    *       this field.
    */
 
-  mkldnn::memory::format_tag format_ = mkldnn::memory::format_tag::undef;
+  dnnl::memory::format_tag format_ = dnnl::memory::format_tag::undef;
 #endif
 
  public:
--- a/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/activation_mkldnn_op.cc
@@ -30,9 +30,9 @@ namespace operators {
 
 using framework::DataLayout;
 using framework::Tensor;
-using mkldnn::memory;
-using mkldnn::primitive;
-using mkldnn::stream;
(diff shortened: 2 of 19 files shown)


paddle/fluid/framework/tensor.h
```diff
@@ -106,7 +106,7 @@ class Tensor {
    *       this field.
    */
 
-  mkldnn::memory::format_tag format_ = mkldnn::memory::format_tag::undef;
+  dnnl::memory::format_tag format_ = dnnl::memory::format_tag::undef;
```
ghost, line 109: @XiaoguangHu01  @chenwhql  @zhiqiu  @Xreki  @luotao1
These are the only changes made to tensor.h. Please approve this simple refactor.

lidanqing-vv:
@baoachun @chenwhql This PR is not related with function changes. But it is required if we want to upgrade oneDNN to v2.5. Please approve reasons are as below.

Since mkldnn has been renamed as dnnl long time ago, but until v2.4, there is mapping https://github.com/oneapi-src/oneDNN/blob/v2.4/include/mkldnn_dnnl_mangling.h file which could aviod us to rewrite dnnl to mkldnn, that is why all mkldnn_xx works in Paddle. But starting from oneDNN v2.5, mkldnn_dnnl_mangling.h is removed from https://github.com/oneapi-src/oneDNN/tree/v2.5-rc/include, and we need to rewrite manually all mkldnn to dnnl, which is this PR.

Silv3S (approved):

jczaja (approved):
LGTM

baoachun:
@lidanqing-intel  Follow-up changes to the framework require someone to review.

chenwhql (approved):
LGTM for tensor.h

paddle/fluid/framework/tensor.h
chenwhql: Will the concept of mkldnn be deprecated in the future? Do we need to change the kMKLDNN of the DataLayout to kDNNL? and do the naming of some MKLDNN functions and directories need to be changed?

ghost: The names of these elements don't NEED to be changed for the framework to properly work with future oneDNN versions, but I feel like changing them would still be healthy and would eliminate some confusion related to the naming. Having said that, changing them is definitely not of high priority right now.

chenwhql: I see, thanks

PR #37352, opened 2021-11-18
Add corner case in PTQ scale calculation

wozna:
### PR types

Bug fixes
### PR changes

Others 
### Describe

During post-training quantization of the Retinanet model, I encountered the problem in KL scaling factor calculation. It turned out that it was not checked whether the iterator, set by the algorithm, was a positive number. This PR adds a condition that will prevent us from reaching the vector value with a negative index.


lidanqing-vv (approved):
LGTM

jczaja (approved):
LGTM

sfraczek (approved):
LGTM! Good job

PR #37371, opened 2021-11-19
Moved Pten Module Declaration inside Kernel APIs

jim19930609:
### PR types
Function optimization

### PR changes
Others

### Describe
Pten module declaration in terms of "PT_DECLARE_MODULE()" used to be placed inside its internal test files. However, this will result in "low-level kernel not found" error for any higher-level tests that make use of kernel apis.

This change moved the use of "PT_DECLARE_MODULE()" from test files to where kernel api functions defined. In that case, along with including kernel api headers, low-level kernel functions will also be imported automatically.


chenwhql (approved):
LGTM

PR #37389, opened 2021-11-19
add new API paddle.nn.initializer.Dirac

zhwesky2010:
### PR types

New features
### PR changes

APIs
### Describe

paddle.nn.initializer.Dirac
---
Initialize the 3D/4D/5D Tensor with Dirac delta function. It can reserve the feature of convolution layer input, which means that as many channels are reserved as possible.
```
attr = paddle.ParamAttr(initializer=paddle.nn.initializer.Dirac())
conv = paddle.nn.Conv1D(3, 2, 3, weight_attr=attr)
# conv.weight:
# Tensor(shape=[2, 3, 3], dtype=float32, place=CPUPlace, stop_gradient=False,
#       [[[0., 1., 0.],
#         [0., 0., 0.],
#         [0., 0., 0.]],
# 
#        [[0., 0., 0.],
#         [0., 1., 0.],
#         [0., 0., 0.]]])

input = paddle.rand([8, 3, 10])
output = conv(input)
input.shape  # [8, 3, 10]
output.shape # [8, 2, 8]
paddle.equal_all(output, input[:, 0:2, 1:9])
# [True]
# It means output is almost the same with input, 2 channels are reserved.
```

FireShot Capture 262 - Dirac-API Document-PaddlePaddle Deep Learning Platform - 10 136 157 23


jeff41404 (approved):
lgtm

python/paddle/fluid/initializer.py
```diff
@@ -1039,9 +1039,10 @@ def calculate_gain(nonlinearity, param=None):
     Get the recommended gain value of some nonlinearity function.
 
     Args:
-        nonlinearity(str): nonlinearity function.
-        param(bool|int|float, optional): optional parameter for somme nonlinearity function. Now, it only applies to 'leaky_relu'. Default: None,  
-        it will be calculated as 0.01 in the formula.
+        nonlinearity(str): name of nonlinearity activation function. If it is a linear function, which is one of 
+        "linear/conv1d/conv2d/conv3d/conv1d_transpose/conv2d_transpose/conv3d_transpose" , will return 1.0
```
TCChenlong, line 1043: will return 1.0？

TCChenlong (approved):
LGTM

python/paddle/fluid/initializer.py
zhwesky2010: This API returns a float number.

lanxianghit (approved):

PR #37410, opened 2021-11-22
Bug fix for snapshotting VariableWrapper with initialized tensor but e…

jim19930609:
### PR types
Bug fixes

### PR changes
Others

### Describe
When snapshotting variable wrapper during GradNodeCreation, we used to create a temporary variable wrapper as long as:
1. it's underlying "tensor" member is initialized
2. its inplace version is bumped. 

We changed this condition to:
1. it's underlying "tensor" is initialized
2. it's underlying "allocation" is initialized
2. its inplace version is bumped. 

After this patch, inplace-version-bumped variable wrapper with initialized tensor but empty allocation will not be treated as "inplaced", and we avoided creating new variable wrapper during the snapshotting.


JiabinYang (approved):
LGTM

PR #37445, opened 2021-11-22
[new-exec] sync scope and variable_scope when init executor

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

- sync scope and variable_scope when init executor
- fix bug in data_transform


Aurelius84 (approved):
LGTM

PR #37490, opened 2021-11-23
Adapt auto search

zhaoyinglia:
### PR types

Bug fixes
### PR changes

Others
### Describe

* fix new var'shape and new op's dist_attr
* adapt auto search

### DIFF
--- a/python/paddle/distributed/auto_parallel/operators/common.py
+++ b/python/paddle/distributed/auto_parallel/operators/common.py
@@ -111,37 +111,27 @@ def find_best_compatible_distributed_operator_impl(name, dist_op, fwd=True):
     return best_compatible_impl, idx
 
 
-def copy_distributed_attr_for_var(dist_context, dst_var, src_var):
-    """
-    copy src var's dist_attr to dst var
-    """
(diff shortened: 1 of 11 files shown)


python/paddle/distributed/auto_parallel/operators/common.py
```diff
@@ -111,37 +111,27 @@ def find_best_compatible_distributed_operator_impl(name, dist_op, fwd=True):
...
-
-    for output_varname in dist_op.desc.output_arg_names():
-        output_var = dst_block.var(output_varname)
-        tensor_dist_attr = dist_context.get_tensor_dist_attr_for_program(
-            output_var)
-        op_dist_attr.set_output_dist_attr(output_varname, tensor_dist_attr)
-
-    dist_context.set_op_dist_attr_for_program(dist_op, op_dist_attr)
-    op_dist_attr = dist_context.get_op_dist_attr_for_program(dist_op)
+def infer_shape(block, src_var, src_var_dist_attr, op_input_dist_attr):
```
JZ-LIANG, line 114: this function is only used for reshard case where input var's mapping is not the same with op's input's mapping. should change new name in next pr.

JZ-LIANG (approved):
LGTM

PR #37500, opened 2021-11-24
[new-exec] fix program cache key

zhiqiu:
### PR types

Bug fixes 
### PR changes

Others
### Describe

[new-exec] fix program cache key

The original implementation use `str(program)` as cache key, which has two problems:

- some attributes can not be decoded,

> in load_inference_mode
> ...
> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 24: invalid start byte

- the program's content may be `vary large`, thus cost much time


wanghuancoder (dismissed):
LGTM

Aurelius84 (dismissed):
LGTM

Aurelius84 (approved):

PR #37511, opened 2021-11-24
bring forward check added_ut

betterpig:
### PR types

Others
### PR changes

Others
### Describe

bring forward check added_ut since if adde_ut failed there is no need to go on.


zhwesky2010 (approved):
LGTM

PR #37573, opened 2021-11-26
Added interface reset_grad_inplace_version

jim19930609:
### PR types
Bug fixes

### PR changes
Others

### Describe
reset_inplace_version removes all inplace related records to VarBase/VariableWrapper, the essential purpose of which is to let you use inplace operations as if using its non-inplaced version, which of course will cause unexpected consequences if not used with care.

This is essentially a hack interface to satisfy one specific request


JiabinYang (approved):
LGTM

Issue #36962, opened 2021-11-03
Bert模型在Intel CPU上量化预测精度异常


yghstill:
- 版本、环境信息：
    1）PaddlePaddle版本：请提供您的PaddlePaddle版本号（如1.1）或CommitID
    2）CPU：Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz
    3）GPU：-
    4）系统环境：ubuntu 16，Python 3.7

- 复现信息：如为报错，请给出复现环境、复现步骤
（1）得到离线量化后的模型后，使用附件中save_quant_model.py对量化模型进行转换和优化操作导出Intel CPU上可以使用的量化模型 
（2）使用PaddleInference 部署量化模型：运行cpu_infer.py执行预测 （注意修改代码中设置的模型等路径），预测精度如下图，正确acc应该0.5以上，不符预期：
- 问题描述：我们在PaddlePaddle中适配支持Bert模型时，模型动转静后，再离线量化，模型在TRT int8上精度正常，在Intel CPU上预测流程可以跑通，但是精度异常
![image](https://user-images.githubusercontent.com/15628872/140010481-76fcb127-50e0-42b6-8139-3f1a36324f58.png)

执行代码在code.zip

[code.zip](https://github.com/PaddlePaddle/Paddle/files/7465321/code.zip)

lidanqing-vv:
@yghstill  收到

wozna:
@yghstill In the _cpu_infer.py_ file, the module `task_distill_zh` is called, which is not available. Where can I find it?

lidanqing-vv:
@wanghaoshuang  
We can not reproduce the issue, error: ModuleNotFoundError: No module named 'task_distill_zh', could you please help? Thanks !

```
(py3.7) li@c3gpo:~/repo/Issue_Bert_36962$ python cpu_infer.py --model_path=$PWD/tnews_quant_models/model
Traceback (most recent call last):
  File "cpu_infer.py", line 26, in <module>
    from task_distill_zh import convert_example, METRIC_CLASSES, MODEL_CLASSES
ModuleNotFoundError: No module named 'task_distill_zh'
(py3.7) li@c3gpo:~/repo/Issue_Bert_36962$ pip install task_distill_zh
ERROR: Could not find a version that satisfies the requirement task_distill_zh
ERROR: No matching distribution found for task_distill_zh
```

yghstill:
[...]

PR #37618, opened 2021-11-26
dequantize matmul and matmul_v2 Y weights in quant2_int8

sfraczek:
### PR types
Bug fixes

### PR changes
Others

### Describe
fix for a problem in issue: https://github.com/PaddlePaddle/Paddle/issues/36962.
There, matmul_v2 has weights quantized but they are not dequantized during conversion of qat model to fp32 model. later matmul_v2 fp32 is used with those quantized weights.
I separated also conv2d and matmul unit tests for quant2_int8_mkldnn_pass

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
+++ b/python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
@@ -333,12 +333,11 @@ class Quant2Int8MkldnnPass(object):
             weight = self._load_param(self._scope, weight_var_name)
             return np.all(np.mod(weight, 1) == 0)
 
+        mul_and_matmul_ops = self._mul_ops + self._matmul_ops
         for op in graph.all_op_nodes():
             if op.name() in self._conv_ops and _is_int8_weights(op, "Filter"):
                 self._dequantize_op_weights(graph, op, "Filter", "Output")
-            elif op.name() in self._mul_ops and _is_int8_weights(op, "Y"):
-                self._dequantize_op_weights(graph, op, "Y", "Out")
-            elif op.name() in self._matmul_ops and _is_int8_weights(op, "Y"):
+            elif op.name() in mul_and_matmul_ops and _is_int8_weights(op, "Y"):
                 self._dequantize_op_weights(graph, op, "Y", "Out")
 
         return graph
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
```diff
@@ -336,6 +338,9 @@ def _is_int8_weights(op_node, weight_name):
                 self._dequantize_op_weights(graph, op, "Filter", "Output")
             elif op.name() in self._mul_ops and _is_int8_weights(op, "Y"):
                 self._dequantize_op_weights(graph, op, "Y", "Out")
+            elif op.name() in self._matmul_ops and _is_int8_weights(op, "Y"):
```
wozna: It looks like you can combine this part into one and just check `op.name() in [self._mul_ops, self._matmul_ops]`. What do you think?

wozna (dismissed):

python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
sfraczek: I can apply your idea, but I will concatenate those lists before for loop

sfraczek: I wonder if we even need matmul_ops variable separate from mul_ops

wozna: You're right, it looks like we are doing exactly the same things for them, so for me you can combine it.

lidanqing-vv:
@baoachun We have big concern about matmul_v2 and related passes. Let's discuss it when you got time
1. Should we enable all fuses of mul, matmul_v1, fc for matmul_v2
2. Could we just keep matmul_v2 as the only interface and mapping it into different ops: mul, matmul_v1, fc according to different situations

baoachun:
> @baoachun We have big concern about matmul_v2 and related passes. Let's discuss it when you got time
> 
>     1. Should we enable all fuses of mul, matmul_v1, fc for matmul_v2
> 
>     2. Could we just keep matmul_v2 as the only interface and mapping it into different ops: mul, matmul_v1, fc according to different situations

I'm sorry I don't understand what you mean very much. What is the problem now?

lidanqing-vv (approved):
LGTM

lidanqing-vv (approved):
LGTM

wozna (approved):
LGTM

PR #37626, opened 2021-11-26
pscore global shuffle&default accessor config

zhaocaibei123:
### PR types

Function optimization
### PR changes

Others
### Describe

1. add global shuffle function for pscore
2. add default accessor config for memory sparse table


yaoxuefeng6 (approved):
LGTM

PR #37664, opened 2021-11-29
Changed default implementation of bfloat16 to eigen

ghost:
### PR types
Others

### PR changes
Others

### Describe
Changed the default implementation of the bfloat16 type to the one provided by the eigen library.


ghost:
@chenwhql @Aganlengzi 
We would like to change the implementation of the bfloat16 datatype from our custom one to the one provided by the Eigen library. Your #37122 PR caused the custom_op tests to fail because during the JIT compilation they can't find the required Eigen header. I tried adding the required include to utils.py by using relative paths, but some CI machines still don't have access to these header files. Where should the header file be added so that it's always visible to the compiler for these tests?

Aganlengzi:
@piotrekobiIntel @lidanqing-intel 
I think using eigen3 files in compiling directories is a workaround, `paddle_includes` in utils.py is used for custom_op to add include directories and different environments(Linux, Windows and Mac) have different compiling paths.  
But I am not sure if we can make all CI machines installed eigen3 and then use them like other system libs, or if PaddlePaddle can put eigen3 files into package. I will ask.

python/paddle/fluid/tests/custom_op/utils.py
```diff
@@ -27,7 +37,15 @@
...
 paddle_includes = [
     os.path.join(site_packages_path, 'paddle', 'include'),
-    os.path.join(site_packages_path, 'paddle', 'include', 'third_party')
+    os.path.join(site_packages_path, 'paddle', 'include', 'third_party'),
+    os.path.join(site_packages_path, 'paddle', 'include', 'third_party',
+                 'eigen3', 'src', 'extern_eigen3'),
+    os.path.join(paddle_path),
+    os.path.join(paddle_path, 'paddle', 'fluid', 'platform'),
+    eigen_path,
+    "C:\\home\\workspace\\cache\\third_party\\cpu\\42f1341df65c796cf2b261c10cd2e4af\\eigen3\\src\\extern_eigen3",
```
Aganlengzi: maybe IS_WINDOWS can be used here.

ghost: Currently I'm still testing which of these paths are required. I'm adding as many as possible to try to make all the CI's pass. After that I'll remove the unnecessary ones and add IS_WINDOWS checks as you suggest.

chenwhql:
We can replace the implementation of bfloat, but not in this way. This will cause the external user's custom operator to cannot use bfloat16. bfloat16.h file has be exposed to users as an external C++ API header file. If it is directly changed to include eigen, Users will not find eigen related symbols.

We suggest to organize the implementation of eigen's bfloat16 to bfloat16.h separately, retain the license, and place it in the paddle/utils directory to replace the original bfloat16.h in paddle, so as to ensure the availability of custom operators. You can refer to paddle/utils/any.h

Aganlengzi:
> We suggest to organize the implementation of eigen's bfloat16 to bfloat16.h separately, retain the license, and place it in the paddle/utils directory to replace the original bfloat16.h in paddle, so as to ensure the availability of custom operators. You can refer to paddle/utils/any.h

@piotrekobiIntel @lidanqing-intel Please consider this suggestion to replace the implementation of bfloat16.

ghost:
> We can replace the implementation of bfloat, but not in this way. This will cause the external user's custom operator to cannot use bfloat16. bfloat16.h file has be exposed to users as an external C++ API header file. If it is directly changed to include eigen, Users will not find eigen related symbols.
> 
> We suggest to organize the implementation of eigen's bfloat16 to bfloat16.h separately, retain the license, and place it in the paddle/utils directory to replace the original bfloat16.h in paddle, so as to ensure the availability of custom operators. You can refer to paddle/utils/any.h

@chenwhql @Aganlengzi Is the solution implemented in 4480afafa3cfee04f19e8a394e290bd4a9d8220e satisfactory? It doesn't add any hard-coded include directiories to the python test code, but instead uses cmake to create a file that provides the correct eigen include directory.

python/paddle/fluid/tests/unittests/test_activation_op.py
```diff
@@ -2973,30 +2973,32 @@ def test_check_grad(self):
...
+            place = core.CUDAPlace(0)
+            self.check_grad_with_place(
+                place, ['X'], 'Out', max_relative_error=grad_atol)
+
+    cls_name = "{0}_{1}".format(parent.__name__, "bf16")
+    TestActBF16.__name__ = cls_name
+    globals()[cls_name] = TestActBF16
+
+
+create_test_act_bf16_class(TestRelu)
```
ghost: @chenwhql @Aganlengzi This bfloat16 test located in test_activation_op.py is failing on the Py3 CI compiled with CUDA. I suspect #37202 caused it to fail. The test is failing only on the Py3 CI. It passes locally with PP compiled with CUDA as well as on the Coverage CI that uses CUDA. It isn't skipped on any of them, as the only skip condition is `@unittest.skipIf(not paddle.is_compiled_with_cuda(), "core is not compiled with CUDA")`, which is not true for PP compiled with GPU. The error shown in the CI log isn't really helpful at all:
image
Could you please look into this and suggest what has to be changed to fix the failing CI?

Aganlengzi: @piotrekobiIntel did you merge the develop? I checked other PRs and test_activation_op seems good in their Py3 CI.

Aganlengzi:
> We suggest to organize the implementation of eigen's bfloat16 to bfloat16.h separately, retain the license, and place it in the paddle/utils directory to replace the original bfloat16.h in paddle, so as to ensure the availability of custom operators. You can refer to paddle/utils/any.h

The suggested way is followed in Paddle in similar needs and BFloat16.h is a separate file in Eigen. In this way, we do not rewrite Eigen/Core to eigen_include.h and need not introduce other changes. Could you please consider the suggestion above?

lidanqing-vv:
The one UT CI that stably failed always passed in other PRs' CIs recently. So the failed UT most probably are caused by this PR.

lidanqing-vv:
@piotrekobiIntel Hi, to summarise, what need to be done is
- bfloat16.h file has be exposed to users as an external C++ API header file, not included in Paddle as this PR do
- move the eigen bfloat implementation into Paddle/paddle/utils/bfloat16.h, retain the license, (and delete the current bfloat16.h in paddle/fluid/platform/bfloat16.h). 
- Reference: Paddle/paddle/utils/any.h

If anything unclear, let's have discussion.

python/paddle/fluid/tests/unittests/test_activation_op.py
ghost: @Aganlengzi I did just now, and the test is still failing. I'd like to fix this CI before changing the location of bfloat16.h, just to be sure that everything else is working correctly. Do you have any suggestions on what can be done or who I should ask for help? As I said above, I believe that #37202 caused the test to fail.

Aganlengzi: @ronny1996 will help check #37202 for UT fail in test_activation_op.py.

ghost:
@chenwhql @Aganlengzi @lidanqing-intel
Could you elaborate about the issue with missing symbols and give some reproduction steps to it? The way we implemented it currently using a cmake command (4480afafa3cfee04f19e8a394e290bd4a9d8220e) is working for us - PP is compiling correctly and all the tests (including the new custom_op JIT tests) are passing locally.

ghost:
The only remaining issue is the Py3 CI failing, and we are waiting for feedback from @ronny1996 as we are not able to reproduce it locally.

ghost:
@chenwhql @Aganlengzi @lidanqing-intel
Eigens BFloat16.h uses many dependencies from the Eigen library, so this would require copying many more files, or copying their code to BFloat16.h. Copying all these files to paddle/utils causes redefinition errors because of other parts of PaddlePaddle that use the Eigen library. These can be fixed for now, but any future updates to the Eigen library can cause more issues, that would need to be looked into and fixed manually. Because of this, I really think that the solution from 4480afafa3cfee04f19e8a394e290bd4a9d8220 is better.

cmake/external/eigen.cmake
```diff
@@ -20,6 +20,8 @@ set(EIGEN_SOURCE_DIR ${THIRD_PARTY_PATH}/eigen3/src/extern_eigen3)
 set(EIGEN_REPOSITORY https://gitlab.com/libeigen/eigen.git)
 set(EIGEN_TAG        f612df273689a19d25b45ca4f8269463207c4fee)
 
+file(WRITE ${PADDLE_SOURCE_DIR}/paddle/pten/api/include/eigen_include.h "#include \"${EIGEN_SOURCE_DIR}/Eigen/Core\"")
+
```
lidanqing-vv, line 24: Is it possible to copy only bfloat16 header related file into this location ?

ghost: No. The bfloat16 implementation of eigen requires a big part of the eigen library to work and trying to copy more files can cause issues as mentioned above.

ghost:
@chenwhql
The current solution of including eigen to make it visible to custom_op tests is creating a file using cmake that always includes the correct absolute path of Eigen.
I've compared the size of the .whl file between current develop and this PR. The sizes in bytes are:
current develop: 112541339
develop + this PR: 112737124
The difference is less than 200KB, (the total size is over 100MB) so this should not be an issue.
We are still waiting for reproduction steps of the missing symbols problem, as on our side we can't see any issues related to this.

PR #37730, opened 2021-11-30
Remove cpp layer

JiabinYang:
### PR types

Function optimization
### PR changes

Others
### Describe

This PR is trying to remove cpp `core.Layer`,  since there is not necessary to have this


zhiqiu (approved):
LGTM

PR #37749, opened 2021-12-01
fix typos: is_integer in attribute.py

iclementine:
### PR types

Others

### PR changes

Others

### Describe

fix typos: misspelling of `is_integer` to `is_interger`.
Reason: It is not yet a public API, so changing its name is okay. It may become a public API in the future, so it is better to fix the typo.

### DIFF
--- a/python/paddle/fft.py
+++ b/python/paddle/fft.py
@@ -15,7 +15,7 @@
 from typing import Sequence
 import numpy as np
 import paddle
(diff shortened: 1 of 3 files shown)


jeff41404:
add is_integer to public API?

iclementine:
> add is_integer to public API?

It would become a public API, along with `is_floating_point` and `is_complex`. But this woudl be done in another pull request.

jeff41404 (approved):
lgtm

PR #37774, opened 2021-12-01
Fix CUDA Graph H2D bug by restore host memory

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
PR https://github.com/PaddlePaddle/Paddle/pull/36987 is wrong, because the device memory which is to be copied into may be reused by other tensors. H2D copy must be captured into the CUDA Graph instead of skipping.

### DIFF
--- a/paddle/fluid/operators/math/concat_and_split.cu
+++ b/paddle/fluid/operators/math/concat_and_split.cu
@@ -287,13 +287,11 @@ class ConcatFunctor<platform::CUDADeviceContext, T> {
     const T** dev_ins_data = nullptr;
     if (!has_same_shape || in_num < 2 || in_num > 4) {
       tmp_dev_ins_data = memory::Alloc(context, in_num * sizeof(T*));
-      {
-        platform::SkipCUDAGraphCaptureGuard guard;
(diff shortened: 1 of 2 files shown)


paddle/fluid/operators/math/concat_and_split.cu
```diff
@@ -287,13 +287,11 @@ class ConcatFunctor<platform::CUDADeviceContext, T> {
...
     if (!has_same_shape || in_num < 2 || in_num > 4) {
       tmp_dev_ins_data = memory::Alloc(context, in_num * sizeof(T*));
-      {
-        platform::SkipCUDAGraphCaptureGuard guard;
-        memory::Copy(BOOST_GET_CONST(platform::CUDAPlace, context.GetPlace()),
-                     tmp_dev_ins_data->ptr(), platform::CPUPlace(),
-                     static_cast<void*>(inputs_data), in_num * sizeof(T*),
-                     context.stream());
-      }
+      auto* restored =
```
zhiqiu, line 290: Will the data in `restored` be changed every batch iteration?

sneaxiy: No. Because the `restored` data is exactly the tensor pointer address of the input/output data.

zhiqiu (approved):

PR #37777, opened 2021-12-01
[new-exec] use stream safe allocator in memcpy_h2d

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

[new-exec] use stream safe allocator in memcpy_h2d

Some operators in paddle may alloc temporal cuda memory during computing and release it intermediately, and the memory may be reallocated by memcpy_h2d. Since most operators use compute_stream and memcpy_h2d uses h2d_stream, some errors may occur.

fix cuda error (700) when training mask_rcnn model in PaddleDetection
image


Aurelius84 (approved):
LGTM

XiaoguangHu01 (approved):
LGTM

PR #37811, opened 2021-12-03
Buf fix for reset grad inplace version

jim19930609:
### PR types
Bug fixes

### PR changes
Others

### Describe
Previously, we took advantage of ResetInplaceVersion() to implement "reset_grad_inplace_version()" interface so as to avoid VariableWrapper copy during AutoGrad, and the way ResetInplaceVersion() works is to set the value of VariableWrapper's "snap_shot_inplace_version" to that of its underlying Tensor's "inplace_version". 

However, an issue occurs once "reset_grad_inplace_version()" is used together with "clear_gradient()" and "cross-batch accumulation", which essentially caused an underlying Tensor copy that set the "inplace_version" to 0 but leave the "snap_shot_inplace_version" unchanged. 

To resolve the mismatch between "snap_shot_inplace_version" and "inplace_version", we added a boolean to ResetInplaceVersion(set_to_zero = false) which set both "snap_shot_inplace_version" and "inplace_version" to 0. In this way, "reset_grad_inplace_version()" interface will become insensitive to any underlying Tensor copy.


zhiqiu (approved):
LGTM

PR #37814, opened 2021-12-03
Support Thread Local Allocator for CUDA Graph capturing

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
Why we need thread local allocator for CUDA Graph capturing? Supposing that we use the `thread_local` mode to capture the CUDA Graph and we use `DataLoader` with `BufferedReader` to read the input data. The `BufferedReader` may allocate GPU memory in another thread when the CUDA Graph is capturing. This is wrong. Therefore, if the capture mode is `thread_local`, we should make the thread local CUDA Graph allocator that would only allocate memory for the certain thread.

Note that this PR make `FLAGS_use_stream_safe_cuda_allocator=false` because the PR https://github.com/PaddlePaddle/Paddle/pull/37290 make the CUDA Graph allocator breaks down.


zhiqiu (approved):
LGTM

PR #37895, opened 2021-12-06
Refine CUDA atomicAdd for FP16 by CUDA primitive methods

sneaxiy:
### PR types
Performance optimization

### PR changes
Others

### Describe
The CUDA function `atomicAdd` for `__half` is supported after CUDA 10.0. We can use this primitive function to replace our self-implemented CAS.

Testing when `vocab_size = 30528, batch_size = 56, seq_len = 512, hidden_size = 1024, repeat_test_times = 1000`, the backward op runs faster on 80G A100 machine:

| Before(ns) | After(ns) | Improvement | 
|---|---|---|
|6045901.697|5118376.15| +15% |


wzzju (approved):
LGTM

PR #37901, opened 2021-12-07
fix softmax max dim

Yanxing-Shi:
### PR types

Performance optimization

### PR changes

OPs

### Describe

When testing the performance of Bert Base model with sequence 512, the maximum dimensions of softMax forward and backward were changed from 318 to 512, and the ips was increased from 287 sequence/s to 322 sequence/s.

max_dim = 318:
image

max_dim = 512:
image


sneaxiy (approved):
LGTM.

PR #37921, opened 2021-12-07
Put_along_axis

huangxu96:
### PR types

New features
### PR changes

APIs
### Describe

Paddle new APIs: put_along_axis


paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,209 @@
...
+          bool is_scatter_like = true>
+struct cpu_gather_scatter_functor {
+  template <typename func_t>
+  void operator()(Tensor self, int dim, const Tensor& index, const Tensor& src,
+                  const std::string& method_name, const func_t& reduce_op,
+                  const platform::DeviceContext& ctx) {
+    if (index.numel() == 0) {
+      return;
+    }
+    auto* self_data = self.data<tensor_t>();  // problem occour here
```
zhhsplendid: What problem? Your comment may confuse future developer.

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,209 @@
...
+    int64_t index_size = index.numel();
+    int64_t src_size = src.numel();
+    auto self_dims = self.dims();
+    VLOG(3) << "index_size" << index_size;
+    VLOG(3) << "self_size" << self_size;
+    auto index_dims = index.dims();
+    auto src_dims = src.dims();
+    if (self_size == 0 || src_size == 0 || index_size == 0) {
+      VLOG(3) << "zero size input found";
+      platform::errors::InvalidArgument(
```
zhhsplendid: `platform::errors::InvalidArgument` only gives a class containing error message. You should throw or enforce something for this error class.

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,209 @@
...
+namespace operators {
+
+using Tensor = framework::Tensor;
+
+class TensorAssign {
+ public:
+  template <typename tensor_t>
+  void operator()(tensor_t* self_data, tensor_t* src_data) const {
+    *self_data = *src_data;
+    VLOG(3) << "self_data assigned:" << *self_data;
```
zhhsplendid: Please remove some debug logs.

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,209 @@
...
+    int64_t src_size = src.numel();
+    auto self_dims = self.dims();
+    VLOG(3) << "index_size" << index_size;
+    VLOG(3) << "self_size" << self_size;
+    auto index_dims = index.dims();
+    auto src_dims = src.dims();
+    if (self_size == 0 || src_size == 0 || index_size == 0) {
+      VLOG(3) << "zero size input found";
+      platform::errors::InvalidArgument(
+          "self_size, src_size, index_size cannot be 0");
```
zhhsplendid: Can we split and tell user which is 0 and triggers this error?

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,209 @@
...
+  void operator()(Tensor self, int dim, const Tensor& index, const Tensor& src,
+                  const std::string& method_name, const func_t& reduce_op,
+                  const platform::DeviceContext& ctx) {
+    if (index.numel() == 0) {
+      return;
+    }
+    auto* self_data = self.data<tensor_t>();  // problem occour here
+    VLOG(3) << "self_data:" << *self_data;
+    auto* index_data = index.data<index_t>();
+    VLOG(3) << "index_data:" << *index_data;
```
zhhsplendid: Same as above, consider to remove some debug logs.

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,209 @@
...
+    int64_t inner_dim_size = 1;
+    int64_t outer_dim_size = 1;
+    for (int64_t i = 0; i < dim; ++i) {
+      inner_dim_size *= index_dims[i];
+    }
+
+    for (int i = dim + 1; i < index_dims.size(); i++) {
+      outer_dim_size *= index_dims[i];
+    }
+    // Debug function
```
zhhsplendid: Remove this kind of debug comment when you are submitting a formal PR

paddle/fluid/operators/gather_scatter_kernel.cu
```diff
@@ -0,0 +1,194 @@
...
+                            Tensor src, const platform::DeviceContext& ctx) {
+  VLOG(3) << "start scatter mul kernel";
+
+  gpu_gather_scatter_functor<tensor_t, index_t,
+                             /*is_scatter_like=*/true>()(
+      self, dim, index, src, "scatter_mul_gpu", reduce_mul, ctx);
+  VLOG(3) << "<<<< Done gpu_scatter_mul_kernel <<<<<";
+}
+
+namespace plat = paddle::platform;
```
zhhsplendid: Can we remove this namespace line?

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,209 @@
...
+                            Tensor src, const platform::DeviceContext& ctx) {
+  VLOG(3) << "start scatter mul kernel";
+  VLOG(3) << "src data scatter:" << *(src.data<tensor_t>());
+  cpu_gather_scatter_functor<tensor_t, index_t,
+                             /*is_scatter_like=*/true>()(
+      self, dim, index, src, "scatter_mul_cpu", reduce_mul, ctx);
+  VLOG(3) << "<<<< Done cpu_scatter_assign_kernel <<<<<";
+}
+
+namespace plat = paddle::platform;
```
zhhsplendid: Can we remove this namespace line?

paddle/fluid/operators/gather_scatter_kernel.cu
```diff
@@ -0,0 +1,194 @@
...
+
+namespace paddle {
+namespace operators {
+
+using Tensor = framework::Tensor;
+
+class TensorAssign {
+ public:
+  template <typename tensor_t>
+  constexpr void operator()(tensor_t* self_data, tensor_t* src_data) const {
```
zhhsplendid: Can we remove the constexpr for this void function?

paddle/fluid/operators/gather_scatter_kernel.h
```diff
@@ -0,0 +1,72 @@
...
+limitations under the License. */
+
+#include "paddle/fluid/framework/tensor.h"
+
+#pragma once
+
+namespace paddle {
+namespace operators {
+
+#define Instantiate_Template_Funtion(func)                              \
```
zhhsplendid: Typo: Funtion -> Function

paddle/fluid/operators/gather_scatter_kernel.h
```diff
@@ -0,0 +1,72 @@
...
+  Instantiate_Template_Funtion_index_t(                                 \
+      func, int) Instantiate_Template_Funtion_index_t(func, float)      \
```
zhhsplendid: Could you split the macro so that each line has one `Instantiate_Template_Funtion_index_t`?

In that case the code would be beautiful.

paddle/fluid/operators/gather_scatter_kernel.h
```diff
@@ -0,0 +1,72 @@
...
+      Instantiate_Template_Funtion_index_t(func, double)                \
+          Instantiate_Template_Funtion_index_t(func, int64_t)           \
+              Instantiate_Template_Funtion_index_t(func, plat::float16) \
+                  Instantiate_Template_Funtion_index_t(func, unsigned char)
+
+#define Instantiate_Template_Funtion_index_t(func, tensor_t)                \
```
zhhsplendid: Typo: Funtion -> Function

paddle/fluid/operators/gather_scatter_kernel.cu
```diff
@@ -0,0 +1,194 @@
...
+
+  gpu_gather_scatter_functor<tensor_t, index_t,
+                             /*is_scatter_like=*/true>()(
+      self, dim, index, src, "scatter_mul_gpu", reduce_mul, ctx);
+  VLOG(3) << "<<<< Done gpu_scatter_mul_kernel <<<<<";
+}
+
+namespace plat = paddle::platform;
+Instantiate_Template_Funtion(gpu_gather_kernel)
+    Instantiate_Template_Funtion(gpu_scatter_assign_kernel)
```
zhhsplendid: Can we put these lines to align for the beginning of the lines?

paddle/fluid/operators/put_along_axis_op.cu
```diff
@@ -0,0 +1,130 @@
...
+    *result = *input;  // inplace opeartion
+  }
+};
+
+template <typename T>
+class PutAlongAxisGradOpCUDAKernel : public framework::OpKernel<T> {
+ public:
+  void Compute(const framework::ExecutionContext &ctx) const override {
+    PADDLE_ENFORCE_EQ(
+        platform::is_cpu_place(ctx.GetPlace()), true,
```
zhhsplendid: It is a CUDA Kernel, but you ask users to run in a CPU Place.

zhhsplendid: Your test coverage may not be enough so this enforce doesn't give you the error during test. Please add tests to make sure your code is correct.

paddle/fluid/operators/put_along_axis_op.cu
```diff
@@ -0,0 +1,130 @@
...
+        platform::errors::PreconditionNotMet("This kernel only runs on CPU."));
+
+    auto input_grad = ctx.Output<Tensor>(framework::GradVarName("Input"));
+    auto index = ctx.Input<Tensor>("Index");
+    auto result_grad = ctx.Input<Tensor>(framework::GradVarName("Result"));
+    auto axis = ctx.Attr<int>("Axis");
+    // We need know the shape of input matrix to determine the shape of grad
```
zhhsplendid: Grammar: "We need to know"

paddle/fluid/operators/put_along_axis_op.h
```diff
@@ -0,0 +1,115 @@
...
+    auto index = ctx.Input<Tensor>("Index");
+    auto result_grad = ctx.Input<Tensor>(framework::GradVarName("Result"));
+    auto axis = ctx.Attr<int>("Axis");
+    // We need know the shape of input matrix to determine the shape of grad
+    // matrix of input.
+    auto input = ctx.Input<Tensor>("Input");
+    input_grad->Resize(input->dims());
+    input_grad->mutable_data<T>(ctx.GetPlace());
+    auto input_grad_data = input_grad->data<T>();
+    for (int64_t i = 0; i < input_grad->numel(); ++i) {
```
zhhsplendid: Can we use memset to set 0? It is faster than manual loop.

paddle/fluid/operators/take_along_axis_op.h
```diff
@@ -0,0 +1,85 @@
...
+    auto index = ctx.Input<Tensor>("Index");
+    auto result_grad = ctx.Input<Tensor>(framework::GradVarName("Result"));
+    auto axis = ctx.Attr<int>("Axis");
+    // We need know the shape of input matrix to determine the shape of grad
+    // matrix of input.
+    auto input = ctx.Input<Tensor>("Input");
+    input_grad->Resize(input->dims());
+    input_grad->mutable_data<T>(ctx.GetPlace());
+    auto input_grad_data = input_grad->data<T>();
+    for (int64_t i = 0; i < input_grad->numel(); ++i) {
```
zhhsplendid: Can we use memset to set 0? It is faster than manual loop.

paddle/fluid/operators/take_along_axis_op.cu
```diff
@@ -0,0 +1,102 @@
...
+  void Compute(const framework::ExecutionContext &ctx) const override {
+    PADDLE_ENFORCE_EQ(
+        platform::is_gpu_place(ctx.GetPlace()), true,
+        platform::errors::PreconditionNotMet("This kernel only runs on GPU."));
+
+    auto input_grad = ctx.Output<Tensor>(framework::GradVarName("Input"));
+    auto index = ctx.Input<Tensor>("Index");
+    auto result_grad = ctx.Input<Tensor>(framework::GradVarName("Result"));
+    auto axis = ctx.Attr<int>("Axis");
+    // We need know the shape of input matrix to determine the shape of grad
```
zhhsplendid: Grammar: we need to know

paddle/fluid/operators/gather_scatter_kernel.cu
zhhsplendid: I saw you are using `plat::float16` in Instantiate_Template_Funtion. 

I would suggest to change `plat::float16` to `paddle::platform::float16`. Then everyone use the macro doesn't have to remember that he has to set `namespace plat = paddle::platform`.

This is an important concept in Software Engineering: less background knowledge for a function call, less error

python/paddle/fluid/tests/unittests/test_put_along_axis_op.py
```diff
@@ -0,0 +1,127 @@
...
+from op_test import OpTest, skip_check_grad_ci
+import paddle
+import paddle.fluid as fluid
+from paddle.framework import core
+from paddle.fluid.dygraph.base import switch_to_static_graph
+
+paddle.enable_static()
+
+
+@skip_check_grad_ci("backward calculation has wrong logic in op test")
```
zhhsplendid: Please make sure the backward is correct instead of skip the check.

python/paddle/fluid/tests/unittests/test_put_along_axis_op.py
```diff
@@ -0,0 +1,127 @@
...
+class TestPutAlongAxisOp(OpTest):
+    def setUp(self):
+        self.init_data()
+        self._cpu_only = True
```
zhhsplendid: Should we also check gpu? As I found your GPU kernel bug (see my comment above)

python/paddle/fluid/tests/unittests/test_take_along_axis_op.py
```diff
@@ -0,0 +1,111 @@
...
+paddle.enable_static()
+
+
+class TestTakeAlongAxisOp(OpTest):
+    def setUp(self):
+        self.init_data()
+        self.op_type = "take_along_axis"
+        self.xnp = np.random.random(self.x_shape).astype(self.x_type)
+        self.target = np.take_along_axis(self.xnp, self.index, self.axis)
+        braoadcast_shape_list = list(self.x_shape)
```
zhhsplendid: Typo: braoadcast -> broadcast

zhhsplendid (commented):
Already a good job. However the code is too long so that I left a lot of comments. Can you split the PR in the future? (You don't have to do it in this PR) It seems the PR exceeding 1000 lines would require Dianhai to review (that may be slow ...)

python/paddle/fluid/tests/unittests/test_put_along_axis_op.py
huangxu96: Done.

python/paddle/fluid/tests/unittests/test_take_along_axis_op.py
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.cu
huangxu96: Done. Got new knowledge~~

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: It's a fixed problem, already remove this comment.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.cu
huangxu96: It was aligned, but it was modified by the pre-commit. I will try to align it again.

paddle/fluid/operators/gather_scatter_kernel.h
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.h
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.h
huangxu96: Done.

paddle/fluid/operators/put_along_axis_op.cu
huangxu96: > It is a CUDA Kernel, but you ask users to run in a CPU Place.

Done.

paddle/fluid/operators/put_along_axis_op.cu
huangxu96: Done.

paddle/fluid/operators/take_along_axis_op.cu
huangxu96: Done.

paddle/fluid/operators/take_along_axis_op.cc
```diff
@@ -0,0 +1,142 @@
...
+
+namespace paddle {
+namespace operators {
+
+class TakeAlongAxisOp : public framework::OperatorWithKernel {
+ public:
+  using framework::OperatorWithKernel::OperatorWithKernel;
+
+  void InferShape(framework::InferShapeContext* ctx) const override {
+    PADDLE_ENFORCE_EQ(
```
iclementine: What is the reason to deviate from paddle's naming convention?

paddle/fluid/operators/take_along_axis_op.cc
```diff
@@ -0,0 +1,142 @@
...
+};
+
+class TakeAlongAxisGradOp : public framework::OperatorWithKernel {
+ public:
+  using framework::OperatorWithKernel::OperatorWithKernel;
+
+  void InferShape(framework::InferShapeContext* ctx) const override {
+    ctx->SetOutputDim(framework::GradVarName("Input"),
+                      ctx->GetInputDim("Input"));
+    ctx->ShareLoD("Input", /*-->*/ framework::GradVarName("Input"));
```
iclementine: Does thid implementation handles LoDTensor correctly? Or otherwise, it is recommended not to do anything with lod.

paddle/fluid/operators/take_along_axis_op.cc
```diff
@@ -0,0 +1,142 @@
...
+                  ops::TakeAlongAxisGradOpMaker<paddle::framework::OpDesc>,
+                  ops::TakeAlongAxisGradOpMaker<paddle::imperative::OpBase>);
+
+REGISTER_OPERATOR(take_along_axis_grad, ops::TakeAlongAxisGradOp);
+
+REGISTER_OP_CPU_KERNEL(take_along_axis, ops::TakeAlongAxisOpKernel<float>,
+                       ops::TakeAlongAxisOpKernel<double>,
+                       ops::TakeAlongAxisOpKernel<int>,
+                       ops::TakeAlongAxisOpKernel<uint8_t>,
+                       ops::TakeAlongAxisOpKernel<int64_t>);
```
iclementine: Is it possible to register for complex64 and complex128?

paddle/fluid/operators/take_along_axis_op.cc
```diff
@@ -0,0 +1,142 @@
...
+        platform::errors::InvalidArgument(
+            "Input(Index) of TakeAlongAxisOp should not be null."));
+    PADDLE_ENFORCE_EQ(
+        ctx->HasOutput("Result"), true,
+        platform::errors::InvalidArgument(
+            "Output(Result) of TakeAlongAxisOp should not be null."));
+
+    auto index_dim = ctx->GetInputDim("Index");
+
+    ctx->SetOutputDim("Result", index_dim);
```
iclementine: So Broadcasting along `index` & `input'` shape (expect for the 'Axis'） is not supported?

paddle/fluid/operators/take_along_axis_op.cc
```diff
@@ -0,0 +1,142 @@
...
+    return framework::OpKernelType(expected_kernel_type.data_type_,
+                                   tensor.place(), tensor.layout());
+  }
+};
+
+class TakeAlongAxisOpMaker : public framework::OpProtoAndCheckerMaker {
+ public:
+  void Make() override {
+    AddInput("Input", "The input tensor of TakeAlongAxisOp");
+    AddInput("Index", "The index tensor of TakeAlongAxisOp");
```
iclementine, line 67: I suggest to add "index" both as an input & as an attr, which enables directly pass what is received from python to the operator without any wrapping or unwrapping(between python scalar & paddle Tensor).

Same for other operators.

paddle/fluid/operators/take_along_axis_op.cc
```diff
@@ -0,0 +1,142 @@
...
+    AddOutput("Result", "The result tensor of TakeAlongAxisOp");
+    AddAttr<int>("Axis",
```
iclementine: I suggest adding "index" both as an input & as an attr, which enables directly pass what is received from python to the operator without any wrapping or unwrapping(between python scalar & paddle Tensor).

If 'Axis' is only added as an attribute, user cannot pass a scalar Tensor as axis to this operator. 
Same for other operators.

paddle/fluid/operators/take_along_axis_op.h
```diff
@@ -0,0 +1,85 @@
...
+#include "paddle/fluid/framework/op_registry.h"
+#include "paddle/fluid/operators/gather_scatter_kernel.h"
+
+namespace paddle {
+namespace operators {
+
+using Tensor = framework::Tensor;
+
+template <typename T>
+class TakeAlongAxisOpKernel : public framework::OpKernel<T> {
```
iclementine: Why not make TakeAlongAxisOpKernel & TakeAlongAxisOpCUDAKernel two partial specialization of a same primary template?

paddle/fluid/operators/take_along_axis_op.h
```diff
@@ -0,0 +1,85 @@
...
+        platform::errors::PreconditionNotMet("This kernel only runs on CPU."));
+
+    auto input = ctx.Input<Tensor>("Input");
+    auto axis = ctx.Attr<int>("Axis");
+    auto index = ctx.Input<Tensor>("Index");
+    auto result = ctx.Output<Tensor>("Result");
+    result->Resize(index->dims());
+    result->mutable_data<T>(ctx.GetPlace());
+
+    const auto &index_type = index->type();
```
iclementine: It's better to return a value, for a small object. 

Though binding a const l-value reference to a temporary is okay.

Tensor.type() returns a value anyway.

paddle/fluid/operators/take_along_axis_op.h
```diff
@@ -0,0 +1,85 @@
...
+    auto index = ctx.Input<Tensor>("Index");
+    auto result_grad = ctx.Input<Tensor>(framework::GradVarName("Result"));
+    auto axis = ctx.Attr<int>("Axis");
+    // We need to know the shape of input matrix to determine the shape of grad
+    // matrix of input.
+    auto input = ctx.Input<Tensor>("Input");
+    input_grad->Resize(input->dims());
+    input_grad->mutable_data<T>(ctx.GetPlace());
+    auto input_grad_data = input_grad->data<T>();
+    for (int64_t i = 0; i < input_grad->numel(); ++i) {
```
iclementine: +1. You csn use `paddle::operators::math::SetConstantSetConstant`

paddle/fluid/operators/take_along_axis_op.cu
```diff
@@ -0,0 +1,102 @@
...
+    auto input = ctx.Input<Tensor>("Input");
+    input_grad->Resize(input->dims());
+    input_grad->mutable_data<T>(ctx.GetPlace());
+
+    // Set to zero tensor.
+    auto &dev_ctx = ctx.template device_context<platform::CUDADeviceContext>();
+    math::SetConstant<platform::CUDADeviceContext, T> functor;
+    functor(reinterpret_cast<const platform::CUDADeviceContext &>(dev_ctx),
+            input_grad, static_cast<T>(0));
+    const auto &index_type = index->type();
```
iclementine: The same. Better to return to init a value rather than bind a reference to a temporary.

paddle/fluid/operators/put_along_axis_op.cc
```diff
@@ -0,0 +1,146 @@
...
+};
+
+class PutAlongAxisGradOp : public framework::OperatorWithKernel {
+ public:
+  using framework::OperatorWithKernel::OperatorWithKernel;
+
+  void InferShape(framework::InferShapeContext* ctx) const override {
+    ctx->SetOutputDim(framework::GradVarName("Input"),
+                      ctx->GetInputDim("Input"));
+    ctx->ShareLoD("Input", /*-->*/ framework::GradVarName("Input"));
```
iclementine: Does put along axis handles LOD correctly? 
If it is not the case, is better not to shared lod.

paddle/fluid/operators/put_along_axis_op.h
```diff
@@ -0,0 +1,115 @@
...
+      }
+    } else {
+      platform::errors::InvalidArgument(
+          "can not suppor reduce_op: (%s) for scatter kernel, only "
+          "support reduce op: 'add‘, 'assign', 'multiply', the defalut reduce "
+          "op is 'assign' ",
+          reduce_op);
+      return;
+    }
+    *result = *input;  // inplace opeartion
```
iclementine: Does this violate the Rule that "Share Data between input & ouptu inside an operator is not allowed."

https://github.com/PaddlePaddle/Paddle/wiki/ShareDataWith-and-ShareBufferWith-are-prohibited-in-OP

paddle/fluid/operators/gather_scatter_kernel.h
```diff
@@ -0,0 +1,72 @@
...
+                                    const Tensor& index, Tensor result,     \
+                                    const platform::DeviceContext& ctx);    \
+  template void func<tensor_t, int64_t>(Tensor input, int dim,              \
+                                        const Tensor& index, Tensor result, \
+                                        const platform::DeviceContext& ctx);
+
+using Tensor = framework::Tensor;
+
+template <typename tensor_t, typename index_t>
+void cpu_gather_kernel(Tensor self, int dim, const Tensor& index, Tensor result,
```
iclementine, line 41: Suggestion:
1. const Tensor& self as const input parameter;
2. Tensor* as output parameter;
3. device context as the first parameter(a convention in paddle followed by many functions that take tensor as parameter in paddle).

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,203 @@
...
+static ReduceAdd reduce_add;
+
+class ReduceMultiply {
+ public:
+  template <typename tensor_t>
+  void operator()(tensor_t* self_data, tensor_t* src_data) const {
+    *self_data *= *src_data;
+  }
+
+  void operator()(bool* self_data, bool* src_data) const {
```
iclementine: Consider making this as a specialization of `template <typename tensor_t> void operator()` rather than an overload of `operator()`.

Template specialization resolution and overloaded function resolution are different stage.

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,203 @@
...
+    *self_data = *self_data && *src_data;
+  }
+};
+
+static ReduceMultiply reduce_mul;
+
+template <typename tensor_t, typename index_t = int64_t,
+          bool is_scatter_like = true>
+struct cpu_gather_scatter_functor {
```
iclementine, line 50: I think it is better to split it into two classes; one for gather and the other for scatter. Reasons:

1. don's pass arguments that is not used, for gather, reduction can be eliminated;
2. better semantics, parameters that does not need to be mutated shoule be made a const parameter.(So we can pass const Tensor to that parameter).
3. readability.

paddle/fluid/operators/gather_scatter_kernel.cc
```diff
@@ -0,0 +1,203 @@
...
+  template <typename func_t>
+  void operator()(Tensor self, int dim, const Tensor& index, const Tensor& src,
```
iclementine: Regular functions that operates on Tensor should take a Tensor either as 
1. const Tensor& (const input);
2. Tensor* (output).

Pass a tensor by-value would cause Tensor assignment, which is a shallow copy. It would have the same effect as `ShareDataWith`, which is not encouraged.

python/paddle/tensor/manipulation.py
```diff
@@ -2492,7 +2492,55 @@ def _var_to_list(var):
...
+    broadcast_shape_list[axis] = 1
+    broadcast_shape = tuple(broadcast_shape_list)
+    if in_dygraph_mode():
+        indices = paddle.broadcast_to(indices, broadcast_shape)
+        return _C_ops.take_along_axis(arr, indices, 'Axis', axis)
+    check_variable_and_dtype(
+        arr, 'x',
+        ['float16', 'float32', 'float64', 'int32', 'int64', 'uint8'], 'take_along_axis')
+    check_variable_and_dtype(indices, 'index', ['int32', 'int64'], 'take_along_axis')
+    indices = paddle.broadcast_to(indices, broadcast_shape) # broadcast to shape of the input array first.
```
iclementine: Oh I see it! broadcasting in not down in the take along axis operator, but a separate operator here.

python/paddle/tensor/manipulation.py
```diff
@@ -2492,7 +2492,55 @@ def _var_to_list(var):
...
+    helper = LayerHelper('put_along_axis', **locals())
+    dtype = helper.input_dtype()
+    helper.append_op(
+        type="put_along_axis",
+        inputs={"Input": arr,
+                "Index": indices,
+                "Value": values},
+        attrs= {"Axis": axis},
+        outputs={"Result": arr})
+    return arr
```
iclementine: a blank line shoud be added here.

python/paddle/tensor/manipulation.py
```diff
@@ -2492,7 +2492,55 @@ def _var_to_list(var):
     out = x.matmul(y).reshape(shape_out)
     return out
 
+def take_along_axis(arr, indices, axis):
```
iclementine: docstring to add

paddle/fluid/operators/put_along_axis_op.h
```diff
@@ -0,0 +1,115 @@
...
+    // matrix of input.
+    auto input = ctx.Input<Tensor>("Input");
+    input_grad->Resize(input->dims());
+    input_grad->mutable_data<T>(ctx.GetPlace());
+    auto input_grad_data = input_grad->data<T>();
+    for (int64_t i = 0; i < input_grad->numel(); ++i) {
+      *(input_grad_data + i) = 0;
+    }
+    const auto &index_type = index->type();
+    if (index_type == framework::proto::VarType::INT32) {
```
iclementine, line 103: I think there may be something wrong here.

Though put along axis is implemented as inplace operation. The gradient operation should compute the gradient of Input and Value separately as needed.

So not only is taking output's gradient along axis with index to fill Value's gradient needed, copying gradients of the output to the input's gradient among those positions that is not gathered and masking (for scatter assign) gradient of the input amopng those gathered positions is also needed.

iclementine: I think there may be something wrong here.

Though put along axis is implemented as inplace operation. The gradient operation should compute the gradient of Input and Value separately as needed.

So not only is taking output's gradient along axis with index to fill Value's gradient needed, copying gradients of the output to the input's gradient among those positions that is not gathered and masking (for scatter assign) gradient of the input amopng those gathered positions is also needed.

iclementine (commented):
1. There are some suggesting on code style and sementic issues;
2. implementation of the gradient of put along aixs seems errorous(both the gradient of input & value should be computed, correctly). I think this is also the reason why it cannot pass grad check.
3. add docstring for APIs.

paddle/fluid/operators/put_along_axis_op.cc
```diff
@@ -0,0 +1,146 @@
...
+  using framework::SingleGradOpMaker<T>::SingleGradOpMaker;
+
+ protected:
+  void Apply(GradOpPtr<T> op) const override {
+    op->SetType("put_along_axis_grad");
+    op->SetInput("Index", this->Input("Index"));
+    op->SetInput("Input", this->Input("Input"));
+
+    op->SetInput(framework::GradVarName("Result"), this->OutputGrad("Result"));
+    op->SetOutput(framework::GradVarName("Input"), this->InputGrad("Input"));
```
iclementine, line 108: The gradient of Value should also be computed.

paddle/fluid/operators/take_along_axis_op.h
huangxu96: Done.

python/paddle/fluid/tests/unittests/test_put_along_axis_op.py
huangxu96: Done.

paddle/fluid/operators/take_along_axis_op.cc
huangxu96: Removed it

paddle/fluid/operators/take_along_axis_op.cc
huangxu96: Doing broadcast in python level

paddle/fluid/operators/take_along_axis_op.h
huangxu96: Done.

paddle/fluid/operators/put_along_axis_op.cc
huangxu96: removed it

python/paddle/tensor/manipulation.py
huangxu96: Added.

paddle/fluid/operators/put_along_axis_op.cc
huangxu96: Done.

paddle/fluid/operators/put_along_axis_op.h
huangxu96: Gradient bug fixed

python/paddle/tensor/manipulation.py
```diff
@@ -2803,3 +2804,68 @@ def take_along_axis(arr, indices, axis):
...
+    Returns : 
+        Tensor: The indexed element, same dtype with arr
+    
+    Examples:
+        .. code-block:: python
+
+            import paddle
+            import numpy as np
+
+            x_np = np.array([[10, 30, 20], [60, 40, 50]])
```
iclementine, line 2828: A more representation example is better.

paddle/fluid/operators/put_along_axis_op.cu
```diff
@@ -0,0 +1,135 @@
...
+    } else if (reduce_op == "assign") {
+      if (index_type == framework::proto::VarType::INT32) {
+        gpu_scatter_assign_kernel<T, int32_t>(*input, axis, *index, *value,
+                                              device_ctx);
+      } else if (index_type == framework::proto::VarType::INT64) {
+        gpu_scatter_assign_kernel<T, int64_t>(*input, axis, *index, *value,
+                                              device_ctx);
+      }
+    } else {
+      platform::errors::InvalidArgument(
```
zhhsplendid: It seems I gave another comment that InvalidArgument only gives error message class. You have to enforce or throw it.

paddle/fluid/operators/gather_scatter_kernel.cu
zhhsplendid: It seems you didn't change what I said...

python/paddle/fluid/tests/unittests/test_put_along_axis_op.py
```diff
@@ -0,0 +1,151 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+import unittest
+import numpy as np
+import copy
+from op_test import OpTest, skip_check_grad_ci
```
zhhsplendid: Remove this useless import `skip_check_grad_ci` as you deleted the code of using it.

huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: Done.

paddle/fluid/operators/take_along_axis_op.cc
huangxu96: Not sure what you mean here.

paddle/fluid/operators/take_along_axis_op.cc
huangxu96: I don't think so, since the python API level won't support complex type.

paddle/fluid/operators/take_along_axis_op.h
huangxu96: Good suggestion, I just do the same thing as other operators did. Will do it in the next PR

paddle/fluid/operators/take_along_axis_op.cu
huangxu96: Same as above.

paddle/fluid/operators/put_along_axis_op.h
huangxu96: Modified into TensorCopy.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: I think it's not necessary in this PR, will do it in the next PR.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: I don't think split it is better since I think scatter and gather share a lot of logic. Making one kernel function can make the code more elegant. Don't want to make some redundant code.

paddle/fluid/operators/gather_scatter_kernel.cc
huangxu96: I just modified the input tensor in my code, is that allowed?

python/paddle/tensor/manipulation.py
huangxu96: Added

paddle/fluid/operators/put_along_axis_op.cu
huangxu96: Done.

paddle/fluid/operators/gather_scatter_kernel.h
huangxu96: Since I will modify the self tensor in my code, I declare using a const variable is not necessary here.

paddle/fluid/operators/take_along_axis_op.h
huangxu96: I don't know how to do in this way, maybe will do it in the next PR

paddle/fluid/operators/take_along_axis_op.cc
huangxu96: I don't think the scalar index is legal input here since this operator has to make sure the index tensor and the input tensor have the same dimensions.

huangxu96:
> 1. There are some suggesting on code style and sementic issues;
> 2. implementation of the gradient of put along aixs seems errorous(both the gradient of input & value should be computed, correctly). I think this is also the reason why it cannot pass grad check.
> 3. add docstring for APIs.
1. Code style fixed for some parts we share the same idea.
2. Gradient bug fixed.
3. Docstring added.

paddle/fluid/operators/gather_scatter_kernel.cc
iclementine: Mutating the input is generally not allowed in opertor kernels.
Operators assume the inputs const tensors, and write the results to the outputs.

The main rationale is that the input must retains to avoid errorous back-propagation, where the value of the intpu tensors may be used again.

Inplace operations are achived by passing the input also as an output parameter. But this happens with some restrictions: only when inplace operaiton at that case would not break back-propagation.

[P.S. It is about dynamic graph mode. For static graph mode, the machanism is different.]

iclementine: When you pass a tensor (say A), to `self`, A is assigned to self.

But `self` is a parameter using `pass by value` fashion, thus, `self` is destructed after the function call. The machanism that does not make it fail is that assignment of Tensor is a shallow copy, tensor A and tensor self shares their allocation, thus, mutating self's allocation also affects A's allocation. But if it is not the case, it would have no side effects.

If you want to use `self` as a output parameter, then `Tensor*` is an appropriate choice.



pass a pointer to tensor is much more common in the code base.

paddle/fluid/operators/put_along_axis_op.cu
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+      PADDLE_THROW(platform::errors::InvalidArgument(
+          "can not suppor reduce_op: (%s) for scatter kernel : %s, only "
+          "support reduce op: 'add‘, 'assign', 'mul' and 'multiply', the "
+          "defalut reduce "
+          "op is assign ",
+          reduce_op));
+      return;
+    }
+    framework::TensorCopy(*input, ctx.GetPlace(), result);
```
iclementine: You should copy the input -> result, then mutating result to avoid mutating input.

Mutating input then copy input -> result still mutates the input.

Check the different order in put_along_axis_op.h

zhhsplendid: Xu Huang and I changed the CPU kernel but not the GPU kernel

@huangxu96 , Please notice this comment and change it.

huangxu96: Done.

paddle/fluid/operators/put_along_axis_op.cc
```diff
@@ -0,0 +1,149 @@
...
+
+namespace paddle {
+namespace operators {
+
+class PutAlongAxisOp : public framework::OperatorWithKernel {
+ public:
+  using framework::OperatorWithKernel::OperatorWithKernel;
+
+  void InferShape(framework::InferShapeContext* ctx) const override {
+    PADDLE_ENFORCE_EQ(
```
chenwhql: 这里建议使用OP_INOUT_CHECK

paddle/fluid/operators/take_along_axis_op.cc
iclementine: If we pass a scalar tensor to `index`, then shape checking will have to be delayed to the execution of the kernel because its value is not known before execution.

But sometimes it is a real requirement. But we can just leave it now. There is a ongoing projects to deal with this kind of inpus(indices, axis, shape, rank, etc.) When it is ready, there may be a refactorization of related code.

paddle/fluid/operators/take_along_axis_op.cc
iclementine: I mean paddle may tend to name the input as "X", but it is not a compulsory requirement, just a guideline.

paddle/fluid/operators/take_along_axis_op.cc
iclementine: Is complex type not supported in python? 
There *are* some python APIs with complex dtype support.

PR #37974, opened 2021-12-08
Support Pylayer with different Dtype inputs

JiabinYang:
### PR types

Function optimization
### PR changes

OPs
### Describe

This PR override `GetExpectedKernelType` to skip check for `Duplicapable` Inputs dtype.


wanghuancoder (approved):
LGTM

PR #37978, opened 2021-12-08
Get base pointer from Allocation

From00:
### PR types
New features

### PR changes
Others

### Describe
  For performance considering, Paddle caches and manages the memory malloced from devices. The memory pointer given to user is not the base-pointer that directly returned from the device interface, as it may have undergone many transformations such as split and align by Allocator. 
  However, sometimes the base-pointer is required. For example, [cudaIpcGetMemHandle](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g8a37f7dfafaca652391d0758b3667539) needs a base-pointer of the device allocation created with _cudaMalloc_, so it can achieve memory sharing between different processes. 
  This PR maintains the base-pointer information in all allocators existing in Paddle, and implement a _base_ptr_ interface in Allocation class.  Thereby we can get the  base-pointer from an allocation like:
```C++
paddle::platform::CUDAPlace& place = paddle::platform::CUDAPlace(); 
size_t size = 1024; // bytes    
std::shared_ptr<Allocation> allocation = paddle::memory::AllocShared(place, size);
void* base_ptr = allocation->base_ptr();
```


zhiqiu (approved):
LGTM

PR #38007, opened 2021-12-09
Fix int32 overflow in cuda kernel loop

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix int32 overflow in cuda kernel loop

Given 

> Op(label_smooth): Inputs: X{auto_241_[LoDTensor<float, CUDAPlace(0), (40, 128, 250027)>]},   Outputs: Out{auto_242_[LoDTensor<NOT_INITED>]}

a `cuda error 700` error may be raised:

Srror Mostore Srmory


pangyoki (approved):
LGTM

PR #38012, opened 2021-12-09
fix fetch op rename_input bug in QAT export model

yghstill:
### PR types

Bug fixes
### PR changes

APIs
### Describe

If next_op is `fetch` and `out_var_name` in fetch_targets, fetch_targets must update, so fix fetch op rename_input bug in QAT export model.

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/imperative/qat.py
+++ b/python/paddle/fluid/contrib/slim/quantization/imperative/qat.py
@@ -484,7 +484,7 @@ class ImperativeQuantizeOutputs(object):
                 model_filename=model_filename,
                 params_filename=params_filename))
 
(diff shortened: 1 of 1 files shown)


wanghaoshuang (approved):

wanghaoshuang:
```
conv2d_1-->tensor_A-->moving_average_abs_max_scale-->tensor_B-->conv2d_2
                                                        |
                                                         --->fetch
```
To remove unused `moving_average_abs_max_scale` from quantization graph, the input of `conv2d_2` will be renamed to `tensor_A`. And hence `tensor_B` will be pruned with  `moving_average_abs_max_scale` when saving inference model.
So we should replace `tensor_B` with `tensor_A` in fetched list.

PR #38024, opened 2021-12-10
[fleet_executor] Fix overlap hang

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
For current, some forward ops may run simultaneously with some backward ops, which will cause some sync ops hang. This pr is aiming fix this prob.


wangxicoding (approved):
LGTM

PR #38047, opened 2021-12-10
move  distribution.py into  distribution package for better scalability

cxxly:
### PR types

Others
### PR changes

Others
### Describe

We plan to develop probability distribution package,  which will add 25 New APIs and corresponding operators.  Currently, paddle organizes all distributions in one file `python/paddle/distribution.py`.  It's difficultly to  extend this file for adding New API.  So we make new a distribution package `python/paddle/distribution`,  move `distribution.py` into  it,  and split it into more simple files.

This modification don't change the path of current api，so users and developer are not affected. 
For example, the follow sample code can normally run before and after this modification.
```python3
import paddle
from paddle.distribution import Normal

# Define a single scalar Normal distribution.
dist = Normal(loc=0., scale=3.)
# Define a batch of two scalar valued Normals.
# The first has mean 1 and standard deviation 11, the second 2 and 22.
dist = Normal(loc=[1., 2.], scale=[11., 22.])
# Get 3 samples, returning a 3 x 2 tensor.
print(dist.sample([3]))
```


iclementine (approved):
LGTM

jeff41404 (approved):
lgtm

raindrops2sea (approved):

XiaoguangHu01 (approved):
LGTM

PR #38056, opened 2021-12-10
[Paddle-TRT] Fix trt dynamic shape ernie unit test on V100

zlsh80826:
### PR types

Bug fixes

### PR changes

Others

### Describe

fix_trt_dynamic_shape_ernie unit test is failed on V100 + tensorrt8
This PR patches two things:
1. Only run ernie-varlen unit test on GPU architure after turing (SM75)
2. Fix-len ernie with fp16 will get errors on enqueue if TensorRT selects the fp32 precision. So this PR disables FP32 precision in supportsFormat after TensorRT8.


shangzhizhou (approved):
LGTM

PR #38057, opened 2021-12-11
[Paddle-TRT] Fix deformable_conv_op_plugin de/serialization and refine the data type selection

zlsh80826:
### PR types

Bug fixes

### PR changes

Others

### Describe

This PR fix some problem in deformable_conv_op_plugin.
1. Fix the problem that hostBuffer is not updated in `deserializeToDevice` and `serializeFromDevice`
2. This plugin doesn't support the running type is different with data_type_. (Many codes use data_type_ to determine the run type).
3. The best practice for `getOutputDataType` return type is input_type.


shangzhizhou (approved):
LGTM

PR #38112, opened 2021-12-14
add some dummy test for elementwise, test=OP-benchmark

iclementine:
### PR types

Others

### PR changes

Others

### Describe

To check if op benchmark fails.


iclementine:
This is a temporary PR to test the functionality of OP benchmark for elementwise ops. As the problem is solved, this PR is closed.

PR #38138, opened 2021-12-14
fix cmake find .h file by mistake

kuizhiqing:
### PR types

Bug fixes

### PR changes

Others

### Describe

FIND will find not only .h file but also the file contains .h in the path.


cmake/pten.cmake
```diff
@@ -20,8 +20,8 @@ string(FIND ${TARGET_PATH} "experimental" pos)
 if (pos GREATER 1)
     file(GLOB HEADERS "${TARGET_PATH}/*" "*.h")
     foreach(header ${HEADERS})
-        string(FIND ${header} ".h" hpos)
-        if (hpos GREATER 1)
+        string(FIND ${header} ".h" hpos REVERSE)
+        if (hpos EQUAL 1)
```
chenwhql, line 24: why use EQUAL?

cmake/pten.cmake
```diff
@@ -40,4 +40,4 @@ pten_header_path_compat(${PADDLE_INFERENCE_INSTALL_DIR}/paddle/include/experimen
 
 # In order to be compatible with the original behavior, the header file name needs to be changed
 file(RENAME ${PADDLE_INFERENCE_INSTALL_DIR}/paddle/include/experimental/extension.h
-            ${PADDLE_INFERENCE_INSTALL_DIR}/paddle/include/experimental/ext_all.h)
\ No newline at end of file
+            ${PADDLE_INFERENCE_INSTALL_DIR}/paddle/include/experimental/ext_all.h)
```
chenwhql, line 43: What changed here

cmake/pten.cmake
kuizhiqing: find file end with .h only

cmake/pten.cmake
kuizhiqing: nothing, pre commit did

PR #38154, opened 2021-12-15
fixed corner case in all shape function

Yelrose:
### PR types
Bug fixes

### PR changes
APIs

### Describe
Fix the corner case when the function has argument shape constructed by list with Tensor and int like the following case.
```
import paddle
import numpy as np
size_a = paddle.to_tensor(np.array(1))
paddle.zeros(shape=[size_a, 100], dtype="float32")

```
This case is caused by the following code
https://github.com/PaddlePaddle/Paddle/blob/086540ccb4e8cde02176f3134adf09e92ac9d3fc/python/paddle/fluid/layers/utils.py#L366.

It runs well in PyTorch
```
import torch
import numpy as np
size_a = torch.from_numpy(np.array(1))
torch.zeros(shape=[size_a, 100], dtype="float32")

```

### DIFF
--- a/python/paddle/fluid/layers/utils.py
+++ b/python/paddle/fluid/layers/utils.py
@@ -363,7 +363,7 @@ def convert_shape_to_list(shape):
     """
     if isinstance(shape, (list, tuple)):
         shape = list(
-            map(lambda x: x.numpy()[0] if isinstance(x, Variable) else x,
+            map(lambda x: x.numpy().flat[0] if isinstance(x, Variable) else x,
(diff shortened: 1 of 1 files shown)


luotao1:
Sorry for too late to notice this PR. The develop branch has already fixed this bug, see
https://github.com/PaddlePaddle/Paddle/blob/fa97e5bad5829a079749f3947bf93af0673c8d5d/python/paddle/fluid/layers/utils.py#L395-L405
Thanks for your contribution!

PR #38195, opened 2021-12-16
Improve the design of StreamSafeCUDAAllocator

From00:
### PR types
New features

### PR changes
Others

### Describe
This PR do the following things to improve the StreamSafeCUDAAllocator  impletemented in #37290 :
1. Rewrites the StreamSafeCUDAAllocator to directly create and record an event when the RecordStream is called, not when the allocation is freed
2. Changes the default stream of StreamSafeCUDAAllocator from NULL stream to the main stream in CUDADeviceContext
3. Adds a _GetStream_ interface for getting the owning stream of a  StreamSafeCUDAAllocation

**Why we no longer create and record events in a deferred manner?**
When the main process finish, Paddle may clear some remaining tensors in shared memory (see #22541), and thus free allocations that may be recorded some CUDA streams. However, the stream created in executor has already destroyed at that time, so we should not try to record events to it. Similar problems may occur in any scenarios where StreamSafeCUDAAllocator is used, so we cannot assume that the recorded streams are still valid when freeing CUDA allocations.

### DIFF
--- a/paddle/fluid/memory/allocation/CMakeLists.txt
+++ b/paddle/fluid/memory/allocation/CMakeLists.txt
@@ -42,7 +42,7 @@ endif()
 cc_library(retry_allocator SRCS retry_allocator.cc DEPS allocator)
 
 if (WITH_GPU OR WITH_ROCM)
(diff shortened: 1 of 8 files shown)


zhiqiu (approved):
LGTM

PR #38200, opened 2021-12-16
[Paddle-TRT] Use TRT inspector to show the information inside an engine to verbose log

leo0519:
### PR types

New features

### PR changes

Others

### Describe
Show the information from TRT inspector when

1. finish building an engine
2. execute with dynamic shape (the format of tensor cannot be confirmed before we bind dimensions)

There is no unit test because this feature only output information.


paddle/fluid/inference/tensorrt/engine.h
```diff
@@ -538,6 +540,17 @@ class TensorRTEngine {
     }
   }
 
+  void getEngineInfo() {
+#if IS_TRT_VERSION_GE(8200)
+    auto infer_inspector_ = infer_engine_->createEngineInspector();
+    infer_inspector_->setExecutionContext(context());
+    VLOG(3) << infer_inspector_->getEngineInformation(
+        nvinfer1::LayerInformationFormat::kJSON);
```
zlsh80826: A trailing underscore is usually used for class/struct member. Here we should not use it.

paddle/fluid/inference/tensorrt/engine.h
```diff
@@ -538,6 +540,17 @@ class TensorRTEngine {
     }
   }
 
+  void getEngineInfo() {
```
zlsh80826: Baidu's coding style follows [Google Style Guide](http://google.github.io/styleguide/cppguide.html). ([Reference](https://github.com/PaddlePaddle/Paddle/blob/develop/CONTRIBUTING.md#code-style))
So we should use `GetEngineInfo()` instead of `getEngineInfo()` ([Reference](
https://google.github.io/styleguide/cppguide.html#Function_Names))

paddle/fluid/inference/tensorrt/engine.h
```diff
@@ -538,6 +540,17 @@ class TensorRTEngine {
...
+#if IS_TRT_VERSION_GE(8200)
+    auto infer_inspector_ = infer_engine_->createEngineInspector();
```
zlsh80826: Use `std::unique_ptr` for  handling `IEngineInspector*` pointer.

paddle/fluid/inference/tensorrt/engine.h
leo0519: renamed variable infer_inspector_ to infer_inspector

paddle/fluid/inference/tensorrt/engine.h
leo0519: renamed getEngineInfo member function to GetEngineInfo

paddle/fluid/inference/tensorrt/engine.h
leo0519: Resolved.

jiweibo (approved):
LGTM

jeng1220:
@leo0519 
add your name at https://github.com/PaddlePaddle/Paddle/blob/develop/AUTHORS.md after Reese

PR #38206, opened 2021-12-16
[Paddle-TRT] Fix trt elementwise converter non-broadcast input shape

zlsh80826:
### PR types

Bug fixes

### PR changes

Others

### Describe

Make trt elementwise converter unit-test has same input shape (aligne to input_data2).
Elementwise operations accepts only same input shape when doing non-broadcast cases.
TensorRT8 will use opt shape for searching tactics, so setting two different opt_shape is invalid.


jiweibo (approved):
LGTM

PR #38264, opened 2021-12-19
Skip zero-size Allocation in RecordStream

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe
A zero-size CUDA Allocation is alloced by ZeroSizeAllocator, not by StreamSafeCUDAAllocator. We should not call _StreamSafeCUDAAllocation::RecordStream_ for a zero-size Allocation.


zhiqiu (approved):
LGTM

PR #38276, opened 2021-12-20
Add PassConflictChecker as a tool to check pass conflict

sneaxiy:
### PR types
Others

### PR changes
Others

### Describe
Add the pass conflict checking tool: `PassConflictChecker`. It is inherited from `DistPassTestBase` but it is different in:
- `PassConflictChecker` implements `DistPassTestBase.apply_passes` method. Developers should override the `PassConflictChecker.pass_config` method to return the passes to check.
- The order of `PassConflictChecker.pass_config()` makes sense. `PassConflictChecker` would check whether the pass order is valid, i.e., with no conflict.

Developers can see `check_pass_conflict_example.py` to learn how to use `PassConflictChecker` to check the pass conflict automatically.

TODO: add more models in the `model_zoo.py` to cover more cases.

### DIFF
--- a/python/paddle/fluid/tests/unittests/distributed_passes/dist_pass_test_base.py
+++ b/python/paddle/fluid/tests/unittests/distributed_passes/dist_pass_test_base.py
@@ -15,7 +15,6 @@
 import unittest
 import paddle
 import os
-import random
 import sys
 import pickle
 import shlex
@@ -24,6 +23,7 @@ import inspect
 import numpy as np
 from collections import OrderedDict
 from paddle.distributed.fleet.launch_utils import run_with_coverage
+from paddle.distributed.passes.pass_base import new_pass, PassBase, PassManager
 
 
 def prepare_python_path_and_return_module(path):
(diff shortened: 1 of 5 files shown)


python/paddle/fluid/tests/unittests/distributed_passes/dist_pass_test_base.py
```diff
@@ -208,23 +219,62 @@ def _distributed_launch(self, apply_pass, gpus=None, **kwargs):
...
+
+        auto_pass_manager = PassManager(passes, auto_solve_conflict=True)
+        new_passes = auto_pass_manager.passes
+        self.assertEqual(
+            len(passes),
+            len(new_passes),
+            "After solving conflicts, the left passes are: {}".format(
+                auto_pass_manager.names))
+
+        for i, (p1, p2) in enumerate(zip(passes, new_passes)):
```
aoyulong, line 273: In this case, the `passes ` and the `new_passes` produced by `PassManager` are the same. Is this just a demo?

aoyulong (approved):
For me, I think everything is OK.

zhiqiu (approved):
LGTM

python/paddle/fluid/tests/unittests/distributed_passes/dist_pass_test_base.py
sneaxiy: No. The `PassConflictChecker` is used to check both pass conflict and pass order. So the pass order of `passes` and `new_passes` should be the same.

PR #38284, opened 2021-12-20
Fixed corner case in all shape function

DesmonDay:
### PR types
Bug fixes 

### PR changes
APIs

### Describe
Fix the corner case when the function has argument shape constructed by list with Tensor and int like the following case.
```python
import paddle
import numpy as np
size_a = paddle.to_tensor(np.array(1))
paddle.zeros(shape=[size_a, 100], dtype="float32")
```
It runs well in PyTorch
```python
import torch
import numpy as np
size_a = torch.from_numpy(np.array(1))
torch.zeros(shape=[size_a, 100], dtype="float32")
```


ZHUI (approved):
LGTM

PR #38285, opened 2021-12-20
[pten] fix when out_dtype is same with x.dtype and still transform type error

MingMingShangTian:
### PR types

Bug fixes
### PR changes

OPs

### Describe

Fix when out_dtype is same with x.dtype and still transform type error.
This will cause error when  calculating  `tensor.mean(0)` and the type of  tensor is `bool` .


zyfncg (approved):
LGTM

PR #38293, opened 2021-12-20
fix dynamic recompute no grad warning

GuoxiaWang:
### PR types

Bug fixes 
### PR changes

APIs
### Describe

fix dynamic recompute no grad warning when using `model.eval()`


ForFishes (approved):
LGTM

Issue #37990, opened 2021-12-09
The fortmat processing of 6-dimensional input is incomplete, which will lead to MKLDNNMemoryFormat::undef error.


baoachun:
https://github.com/PaddlePaddle/Paddle/blob/1911b6f0da63dee423bc47754362da8fc45e1127/paddle/fluid/platform/mkldnn_helper.h#L307

<img width="306" alt="图片" src="https://user-images.githubusercontent.com/22114318/145324775-5447ae83-1eb0-49ac-8f3f-e51985dcc13e.png">

![图片](https://user-images.githubusercontent.com/22114318/145324814-b8487708-506f-4e60-8488-b3a66eb23a37.png)

baoachun:
@lidanqing-intel @jakpiase @jczaja

baoachun:
Export inference model:
```
git clone https://github.com/PaddlePaddle/PaddleClas.git

git checkout develop

bash test_tipc/prepare.sh test_tipc/config/RedNet/RedNet26_train_infer_python.txt whole_infer

python3 tools/export_model.py -c  ppcls/configs/ImageNet/RedNet/RedNet26.yaml
```

Running command:
```
python3 deploy/python/predict_cls.py -c deploy/configs/inference_cls.yaml -o Global.use_gpu=False -o Global.enable_mkldnn=True -o Global.cpu_num_threads=1 -o Global.inference_model_dir=inference/ -o Global.batch_size=16 -o Global.infer_imgs=dataset/ILSVRC2012/val -o Global.benchmark=True
```

jakpiase:
Hi @baoachun.
I have followed your steps and everything works on my setup. Could you please give me more information about system and CPU that you are using?

baoachun:
Hi @jakpiase, the basic information is as follows:
system version: ubuntu 18.04
python version: python3.7
paddle version: the newest develop branch
cpu: 6271

jakpiase:
Hi @baoachun, I have fixed this bug in a PR, but I have an additional question. Is this model pretrained? Even if I run it with `Global.enable_mkldnn=False` I am not getting any meaningful output from it. Is there any simple way to check the correctness of the model's output?

baoachun:
Hi @jakpiase, what do you mean by *meaningful output* ?

jakpiase:
[...]

PR #38299, opened 2021-12-20
Fix for undefined format for 6 dim tensor [Abandoned]

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Added support for acbde and acbdef format tags and disabled oneDNN eltwise kernels in reductions and eltwises if dims rank is higher than 5 due to extremely bad performance because of using reference implementation in that scenarios. Also deleted elementwise_mul's GetExpectedKernelType function,  because it was identical as it's base classes one, so that was just a code redundancy. It is a fix for #37990


lidanqing-vv:
@lidanqing-intel @baoachun Hi, this PR is ready for review and merging.

lidanqing-vv:
@baoachun Please review this PR.

jakpiase:
CIs are acting weirdly, so I will create a new PR with this fix

PR #38312, opened 2021-12-21
optimize buddy_allocator

m3ngyang:
### PR types
Function optimization

### PR changes
Others

### Describe
- rewrite RefillPool to avoid duplicated codes
- reduce Release algorithm complexity, from O(N^2) to O(NlogN)


zhhsplendid (approved):
LGTM, but memory allocator is very low level for Paddle, please have enough tests to make sure this PR doesn't break Paddle.

PR #38318, opened 2021-12-21
[WIP] [Paddle-TRT] Add an option to show TRT engine info by inspector

leo0519:
### PR types

New features

### PR changes

Others

### Describe

Add an option `use_inspector` in AnalysisConfig.EnableTensorRTEngine function.
If the option is True, we will show the information of layers in TensorRT engine when:
1. the engine is built or loaded
2. run the engine with dynamic shape (need to bind dimension first or the information will be incomplete)

This feature requires TensorRT 8.2.

### DIFF
--- a/paddle/fluid/inference/api/analysis_config.cc
+++ b/paddle/fluid/inference/api/analysis_config.cc
@@ -383,7 +383,7 @@ MkldnnQuantizerConfig *AnalysisConfig::mkldnn_quantizer_config() const {
 void AnalysisConfig::EnableTensorRtEngine(
     int workspace_size, int max_batch_size, int min_subgraph_size,
     AnalysisConfig::Precision precision_mode, bool use_static,
-    bool use_calib_mode) {
+    bool use_calib_mode, bool use_inspector, bool use_inspector_exec) {
 #if defined(PADDLE_WITH_CUDA) || defined(PADDLE_WITH_HIP)
   if (!use_gpu()) {
     LOG(ERROR) << "To use TensorRT engine, please call EnableGpu() first";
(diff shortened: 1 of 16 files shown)


leo0519:
The feature only supports TensorRT 8.2, so the unit test should be disabled if paddle was compiled with older TensorRT.
How can I get the TensorRT version with paddle python API so the unit test can run with different TensorRT version.

paddle/fluid/inference/api/analysis_config.cc
```diff
@@ -382,7 +383,7 @@ MkldnnQuantizerConfig *AnalysisConfig::mkldnn_quantizer_config() const {
 void AnalysisConfig::EnableTensorRtEngine(
     int workspace_size, int max_batch_size, int min_subgraph_size,
     AnalysisConfig::Precision precision_mode, bool use_static,
-    bool use_calib_mode) {
+    bool use_calib_mode, bool use_inspector) {
```
shangzhizhou, line 386: 这个debug开关建议单独增加配置项（比如EnableTensorRtInspector），enableTensorRtEngine接口过于复杂了。请@周平看下

python/paddle/fluid/tests/unittests/ir/inference/inference_pass_test.py
```diff
@@ -121,7 +121,8 @@ def _get_analysis_config(self,
                     self.trt_parameters.min_subgraph_size,
                     self.trt_parameters.precision,
                     self.trt_parameters.use_static,
-                    self.trt_parameters.use_calib_mode)
+                    self.trt_parameters.use_calib_mode,
```
shangzhizhou, line 124: 同上

PR #38323, opened 2021-12-21
fix clip extra when QAT export model

yghstill:
### PR types

Bug fixes
### PR changes

APIs
### Describe

When export model in QAT, `clip_extra=True`, which may cause some quantitative model fail to run normally on Paddle Lite. So reset `clip_extra=False`.

### DIFF
--- a/python/paddle/fluid/contrib/slim/tests/test_quantization_scale_pass.py
+++ b/python/paddle/fluid/contrib/slim/tests/test_quantization_scale_pass.py
@@ -173,7 +173,7 @@ class TestQuantizationScalePass(unittest.TestCase):
                 'quant_scale_model' + dev_name, ['image', 'label'], [loss],
                 exe,
                 server_program,
-                clip_extra=True)
+                clip_extra=False)
 
     def test_quant_scale_cuda(self):
         if fluid.core.is_compiled_with_cuda():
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/contrib/slim/tests/test_quantization_scale_pass.py
```diff
@@ -173,7 +173,7 @@ def build_program(main, startup, is_test):
                 'quant_scale_model' + dev_name, ['image', 'label'], [loss],
                 exe,
                 server_program,
-                clip_extra=True)
+                clip_extra=False)
```
winter-wang: 单测可以不改吗？

python/paddle/fluid/contrib/slim/tests/test_quantize_transpiler_v2.py
```diff
@@ -145,7 +145,7 @@ def build_program(main, startup, is_test):
                     './infer_model', ['image', 'label'], [loss],
                     exe,
                     test_program,
-                    clip_extra=True)
+                    clip_extra=False)
```
winter-wang: 单测可以不改吗

python/paddle/fluid/contrib/slim/tests/test_quantization_scale_pass.py
yghstill: done

python/paddle/fluid/contrib/slim/tests/test_quantize_transpiler_v2.py
yghstill: done

winter-wang (approved):
LGTM

PR #38337, opened 2021-12-22
Replaced core.ops with _C_ops

jim19930609:
### PR types
New features

### PR changes
Others

### Describe
Unified all python-level "core.ops" to call through "_C_ops", so that "_C_ops" become the only entrance to dygraph function calls. This way, we can easily switch between "core.ops" and "core.eager.ops" via control logics defined in "_C_ops.py"


JiabinYang (approved):
LGTM

PR #38374, opened 2021-12-22
[Paddle-TRT] Fix dynamic_shape_ernie_test resource life cycle

zlsh80826:
### PR types

Others

### PR changes

Others

### Describe

In the previous PR(https://github.com/PaddlePaddle/Paddle/pull/38056), we found the FP32 precision has problem on fp16 tests.
However, the root cause is the emb_layernorm dtor calls `terminate()`. But the `terminate()` should be called by TensorRT instead of the plugin. It caused two times cudaFree().


shangzhizhou (dismissed):
LGTM

shangzhizhou (approved):
LGTM

Issue #38126, opened 2021-12-14
The output format of mkldnn conv is wrong when data_format is NHWC


baoachun:
![图片](https://user-images.githubusercontent.com/22114318/145959037-c35b64e0-6f9e-47da-849b-fafd4e854d47.png)

The problem can be reproduced according to this pr：https://github.com/PaddlePaddle/Paddle/pull/38107

Runing the following command to reproduce.
```
git clone the code and cmake with -DWITH_TESTING=ON
then running the following command

ctest -R test_mkldnn_conv_gelu_fuse_pass -V
```

lidanqing-vv:
@jczaja Please comment on this issue. Otherwise that disabling PR will be merged.

jczaja:
@baoachun I implemented NHWC support for oneDNN kernels in PaddlePaddle so I can offer some insight.
If PaddlePaddle  input is of layout NCHW then shape is [N,C,H,W]. If PaddlePaddle input is of layout NHWC then shape is [N,H,W,C]. For oneDNN kernels input regardless if it is NCHW or NHWC then shape is always [N,C,H,W]. So internally for NHWC models when there is a CPU op followed by oneDNN op then there is shape rotation from [N,H,W,C] to [N,C,H,W].  And  when there is oneDNN op followed by CPU op then there is shape roatation from  [N,C, H, W] to [N,H,W,C]. The thing is that this assumes that after oneDNN op there is some CPU op. When We implemented NHWC support then there was fetch op always the last one so shape rotation happened there. But I noticed that later on PaddlePaddle by default is skipping fetch op and user get access to data that has still oneDNN shape. So proper fix would be to add this shape rotation  in a mechanics where you get access to oneDNN tensor. Here is line calling shape rotation: 
https://github.com/PaddlePaddle/Paddle/blob/21b7ed3e8d91ffb2fab3e28367c6eacfa292855b/paddle/fluid/framework/data_layout_transform.cc#L189-L191

Some more information (design doc on NHWC in oneDNN kernels):
https://github.com/PaddlePaddle/docs/blob/develop/docs/design/mkldnn/nhwc/nhwc.md

baoachun:
Hi @jczaja , at present, a user encounters the inference error problem, which is caused because conv2d does not support NHWC format.
![图片](https://user-images.githubusercontent.com/22114318/147805937-ab300300-a121-47a2-9055-ae8c7cb7e2d0.png)
[...]

PR #38387, opened 2021-12-23
fix conv2d and conv2d_transpose output format error when using mkldnn

baoachun:
### PR types
Bug fixes

### PR changes
OPs

### Describe

修复当开启mkldnn推理conv2d或conv2d_tranpose的data_format 属性为 NHWC时，输出format不对问题。
回退到CPU实现


baoachun:
@lidanqing-intel  Please review, thanks. When the issue https://github.com/PaddlePaddle/Paddle/issues/38126 is repaired, this change will be deleted.

lidanqing-vv:
@jczaja  Please review. Baoachun said that since it is will not be fixed in 15 days, so he need to disable NHWC first.

PR #38399, opened 2021-12-23
[Paddle-TRT] Replace GeLU plugin with TensorRT built-in layer for TensorRT 7.0.

leo0519:
### PR types

Function optimization
### PR changes

OPs
### Describe

Replace GeLU plugin with TensorRt built-in layer after TensorRT 7.0.
The benefit is that TensorRT can fuse the fc + gelu layer.
We implement both approximate and nonapproximate GeLU.


jeng1220:
@zhouping828 ,
Please help review. Thanks

zhouping828 (commented):
LGTM

Wangzheee (approved):
LGTM

PR #38439, opened 2021-12-24
Refine the float16 implementation by convert to float/double implicitly

sneaxiy:
### PR types
Function optimization

### PR changes
Others

### Describe
Add implicit conversion from `paddle::platform::float16` to `float/double`. This implicit conversion is accepted because there is no accuracy drop.

However, there are some obstacles when I do this implicit conversion. CUDA `half` has several constructors like `half::half(float)` and `half::half(double)`. After this PR, the following conversion may fail when compiling:

```C++
paddle::platform::float16 x;

// Compilation failure. The CUDA half has constructor like `half::half(float)` and `half::half(double)`
// The compiler may not find the suitable conversion function from these several options.
auto y = half(x);
```

To solve the problem above, I delete the explicit conversion function `paddle::platform::float16::operator half() const`. Users should use another function to do the conversion `half paddle::platform::float16::to_half() const`. I think this is OK, because we do not use the `half` explicitly in the framework codes. 

**We really need the implicit conversion from `float16` to `float`.** For example, if I use the `cub` library to implement `y=sum(x)`, where `x` is a fp16 tensor. The following codes would fail when compiling before this PR:

```C++
struct SquareFunctor { 
  HOSTDEVICE float operator ()(platform::float16 x) const {
    return static_cast<float>(x) * static_cast<float>(x);
  }
};

cub::TransformInputIterator<float, SquareFunctor, platform::float16*> input_iter(x_data);
float *output_iter = ...;
cub::DeviceReduce::Reduce(..., input_iter, ..., output_iter, ...);
```

The compilation error is something like `cannot convert paddle::platform::float16 to float implicitly`. Before this PR, many useful function in `cub` library (or many other libraries) cannot be used if the input data is fp16 and output data is float. I think that this is very terrible.

### DIFF
--- a/paddle/fluid/operators/clip_op.h
+++ b/paddle/fluid/operators/clip_op.h
@@ -95,7 +95,7 @@ class ClipKernel : public framework::OpKernel<T> {
                       platform::errors::InvalidArgument(
                           "max should be greater than or equal to min. "
                           "But received min = %f, max = %f",
-                          min, max));
+                          static_cast<float>(min), static_cast<float>(max)));
 
     auto* x_var = context.InputVar("X");
     if (x_var->IsType<framework::LoDTensor>()) {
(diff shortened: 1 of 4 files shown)


zhiqiu (approved):
LGTM

PR #38465, opened 2021-12-27
support new api : linalg.lstsq

haohongxiang:
### PR types

New features
### PR changes

APIs
### Describe

support new api : linalg.lstsq


jeff41404:
for paddle.linalg.lstsq API, lstsq should be add to list of \_\_all\_\_ and import in python/paddle/linalg.py

PR #38499, opened 2021-12-27
Fix scatter_op fp16 perf problem.

limin2021:
### PR types
Performance optimization

### PR changes
OPs

### Describe
The performance of scatter op of fp16 when override is false is not well because of the bad performance of CudaAtomicAdd on fp16. In this pr, we fix this performance issue by adding scatter op into _extra_unsupported_fp16_list for static graph and PURE_FP16_BLACK_LIST for dygraph.


sneaxiy (approved):
LGTM

PR #38577, opened 2021-12-29
【Auto Parallel】 Update dist tensor

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

This pr updates DistributedTensor, local tensor can be created by dist tensor and local tensor information can be obtained from DistributedTensor. In the future, a new pr will adapt other modules which use DistributedTensor.


Caozhou1995:
This pr is replaced by https://github.com/PaddlePaddle/Paddle/pull/38747

PR #38579, opened 2021-12-29
params file will not be a nessary file

JZZ-NOTE:
### PR types

Others
### PR changes

Others
### Describe

params file will not be a nessary file


jiweibo (approved):
LGTM

PR #38590, opened 2021-12-29
[Code Clean] Clean useless return in cudnn_helper.h

JamesLim-sy:
### PR types

Others
### PR changes

OPs
### Describe

- According to after-merged advice of [PR34409](https://github.com/PaddlePaddle/Paddle/pull/34409), clean useless `return` syntax.


Xreki (approved):
LGTM

PR #38647, opened 2021-12-31
Fix memcpyD2H sync behavior with other stream

Aurelius84:
### PR types

Performance optimization 
### PR changes

OPs
### Describe

Fix memcpyD2H sync behavior with other stream

在 transformer 上测试，开启GC策略后，速度从 23370 word/s 提升至 31377 word/s


zhiqiu (approved):
LGTM

PR #38672, opened 2022-01-04
[new-exec] avoid adding_feed_fetch in each run

zhiqiu:
### PR types

Performance optimization
### PR changes

Others
### Describe

[new-exec] avoid adding_feed_fetch in each run

- Before
Calling `_add_feed_fetch_ops` once per `exe.run` with `FLAGS_USE_STANDALONE_EXECUTOR=true`

- After
Calling `_add_feed_fetch_ops` only once in `exe.run`, and use the cached program afterwards.

yolov3 22.38 img/s->31.5 img/s, + 40.7 %


Aurelius84 (approved):
LGTM

PR #38720, opened 2022-01-05
Fix bug for UT GetAllocatorInterfaceTest

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe
When the _GetAllocatorInterfaceTest_ is the first executed test case in _stream_safe_cuda_alloc_test_, the _GetAllocator_ call will fail since the Allocator has been not created yet. It occurs in some test environments and cause a CI error:
image

This PR fix this bug, and by the way replace some shared_ptr with unique_ptr in this UT.


zhiqiu (approved):
LGTM

Issue #37877, opened 2021-12-06
Refactor save_quant_model.py to C++ pass


lidanqing-vv:
Refactor save_quant_model.py to C++ pass

lidanqing-vv:
Done by @baoachun

Issue #37422, opened 2021-11-22
[Feature] Adaptation of the new quantization method for mkldnn.


baoachun:
1. For output activation of ops, slim will insert QuantizeLinear and DequantizeLinear operation which named after ```quantize_linear``` and ```dequantize_linear```.
<img width="363" alt="图片" src="https://user-images.githubusercontent.com/22114318/142825558-ad164f56-50f0-4cd7-a8ad-509def8a548a.png">

2. For trainable parameters of ops, slim first quantizes and saves the weight in low bits, and then inserts an dequantize operation before entering op for calculation.
<img width="812" alt="图片" src="https://user-images.githubusercontent.com/22114318/142826303-23da22a2-3616-487e-acdc-fa626938ca69.png">

3. ```quantize_linear```
The quantization process requires two parameters, scale and zero_point, both of which are 1-dimensional tensors. 
<br>The quantitative formula is: ```y=saturate(round(x/scale)+zero_point)```.

- **Attributes**:
    - ```quant_axis```: INT32, optional. 
    In the ```per-axis``` quantification method, the axis on which the dimension is quantified. If this property is not set, the quantization method defaults to ```per-layer``` quantization. For convolution input ```[batch, channel, H, W]```, when the quantization method is ```channel-wise```, ```quant_axis``` is 1.
    - ```bit_length```: INT32, default is 8.
    The number of bits occupied by the quantized numerical precision, currently only supports quantization as a signed integer.

- **Inputs**:
    - ```X```: FP32.
    - ```Scale```: FP32 <br> When the quantization method is ```layer-wise```, the size of the ```Scale``` is 1. When the quantization method is ```axis-wise```, the size of the ```Scale``` is equal to the size of the input tensor in the axis dimension.
    - ZeroPoint: INT32, optional.  <br> The size is the same as the ```Scale```. The value range depends on ```bit_length```, when ```bit_length``` is 8, the value range of the Tensor is within the value range of Int8.
 
- **Outputs**:
    - ```Y```: INT32. <br> The shape of Y is the same as X. The value range depends on ```bit_length```, when ```bit_length``` is 8, the value range of the Tensor is within the value range of Int8.
 
4. ```dequantize_linear```
According to ```scale```, ```zero_point``` and ```quant_axis```, the low-precision value Tensor is inversely quantized into a high-precision value Tensor. <br> The de-quantitative formula is: ```y=(x-zero_point)*scale```.

- **Attributes**:
[...]

PR #38728, opened 2022-01-05
Add enabling FC MKLDNN passes

zuzg:
### PR types
New features
### PR changes
OPs

### Describe
* Since many models (especially int8) will benefit from enabling fc_mkldnn_pass and fc_act_mkldnn_fuse_pass, I've made enabling them easier
* Adding fc passes in ernie int8 model caused format error, so the fix is to change NHWC to NCHW in this narrow case


baoachun:
Hi @zuzg, I strongly do not recommend adding configuration items that are not universal, as this will cause a lot of confusion to users. I think this function needs to be set in the mkldnn quantization pass, when the adaptation of the new quantization method is completed, reference to https://github.com/PaddlePaddle/Paddle/issues/37422 and https://github.com/PaddlePaddle/Paddle/issues/37877.

@lidanqing-intel

PR #38732, opened 2022-01-05
Remove useless headers for some grad ops

JamesLim-sy:
### PR types

Others
### PR changes

OPs
### Describe

- 统一删除[PR37728](https://github.com/PaddlePaddle/Paddle/pull/37728), [PR38044](https://github.com/PaddlePaddle/Paddle/pull/38044), [PR37754](https://github.com/PaddlePaddle/Paddle/pull/37754) 中冗余的header信息

PR #38743, opened 2022-01-06
Revert "Remove useless headers for some grad ops"

JamesLim-sy:
- Reverts PaddlePaddle/Paddle#38732, for some useful advices has not been fully achieved


ZzSean (approved):
LGTM

PR #38757, opened 2022-01-06
[Cherry-pick] Add a check for multiplex op

aoyulong:
### PR types
Bug fixes

### PR changes
OPs

### Describe

The paddle.multiplex will break down when the size of candidate tensors is zero and doesn't give useful an error message. This pr will check the situation and gives user more information.


Xreki (approved):
LGTM

PR #38786, opened 2022-01-07
[NewExe]Fix blocking in lr_sheduler from Executor

Aurelius84:
### PR types

Performance optimization
### PR changes

Others
### Describe

Fix blocking in lr_sheduler from Executor, Peformace of Bert Base from 155 -> 158 seq/s


zhiqiu (approved):
LGTM

PR #38837, opened 2022-01-10
Fix mkldnn invalid infershape impl

chenwhql:
### PR types

Function optimization
### PR changes

Others
### Describe

Fix mkldnn invalid infershape impl

The external information to be used in the `InferShape` function of the op needs to be passed in through the `InferShapeContext`. Even the information in current op should not be used.

When registering the OP, the `InferShape` function of the OP is stored as a member pointer of `OpInfo`. This member pointer may be used independently of the op. If the `InferShape` function relies on the internal member information of the OP, the inference results independent of the OP and bound with the op will be different, which will be wrong in some scenarios.

Therefore, this PR uniformly modifies the behavior of OP members directly used in some mkldnn related `InferShape` functions.

related PR: https://github.com/PaddlePaddle/Paddle/pull/38776


zhiqiu (approved):
LGTM

PR #38840, opened 2022-01-10
Fix bug in elementwise_mul/div_grad when inplace strategy

ZzSean:
### PR types

Bug fixes 
### PR changes

OPs
### Describe

Fix bug in elementwise_mul/di _grad when inplace strategy：
1）For samedims cases，just do it directly
2）Because of Ternary compute，the inplace strategy doesn't effect the result

Example：
In some models，there is x=x*mask，where the shape of x and mask is the same and mask.stop_gradients is True.
In this case, the x_grad woule be wrong


JamesLim-sy (dismissed):
LGTM, this bug will be fixed later.

AshburnLee:
LGTM，验证了Slim模型，训练loss值符合预期

AshburnLee (approved):
已验证，Slim模型loss值符合预期

JamesLim-sy (approved):
LGTM

PR #38871, opened 2022-01-11
Support eager auto prune and share buffer to

JiabinYang:
### PR types

New features
### PR changes

Others
### Describe

This PR support `_share_buffer_to` api in eager mode which will work as sharing `TensorBase` of `EagerTensor`'s.
And Then, this pr enhance `AddEdges` in `GradNodeBase` to support some auto-prune case


wanghuancoder (approved):
LGTM

PR #38301, opened 2021-12-21
[Unify Tensors PR #1] Replaced pten::Allocation with shared_ptr<memory::Allocation> for Storage

jim19930609:
### PR types
New features

### PR changes
Others
### Describe
Replaced pten::Allocation with shared_ptr<memory::Allocation> for Storage.

PR #38946, opened 2022-01-14
[Refactoring Tensor PR #2] changes allocator type, test=develop 

Shixiaowei02:
### PR types

Others

### PR changes

APIs

### Describe

将 DenseTensor 的构造函数改为支持 Allocator 的形式，以与框架单例解耦。


chenwhql (approved):

paddle/pten/core/storage.h
```diff
@@ -91,12 +91,11 @@ class Storage : public intrusive_ref_counter<Storage> {
 class TensorStorage : public Storage {
  public:
   using Place = paddle::platform::Place;
-  using Allocator = deprecated::Allocator;
 
-  explicit TensorStorage(const std::shared_ptr<Allocator>& a) : alloc_(a) {}
+  explicit TensorStorage(Allocator* a) : alloc_(a) {}
 
-  TensorStorage(const std::shared_ptr<Allocator>& a, size_t size)
```
jim19930609, line 98: Better use "const Allocator*" as argument?

paddle/pten/core/storage.cc
```diff
@@ -18,7 +18,7 @@ namespace pten {
 
 void TensorStorage::Realloc(size_t size) {
   this->Clear();
-  data_ = paddle::memory::AllocShared(alloc_->place(), size);
+  data_ = alloc_->Allocate(size);
```
jim19930609, line 21: "Allocator* alloc_" could be nullptr by the time "Realloc()" is called, shall we add validation check here?

paddle/pten/core/storage.h
```diff
@@ -114,24 +113,18 @@ class TensorStorage : public Storage {
...
   }
 
   bool OwnsMemory() const noexcept override { return true; }
-  const std::shared_ptr<Allocator>& allocator() const noexcept {
-    return alloc_;
-  }
 
  private:
-  const std::shared_ptr<Allocator> alloc_;
+  Allocator* alloc_;
```
jim19930609, line 127: If I understand it correctly, the lifetime of Allocator is essentially controlled by the "Context" instead of the DenseTensor. If the Context happened to destruct its Allocator, then this Allocator* held by DenseTensor will be dangling.

paddle/pten/core/storage.h
Shixiaowei02: To allow the Tensor to modify the internal state of the Allocator, `const Allocator*` or `const shared_ptr<const Allocator>&` are not recommended here.

paddle/pten/core/storage.cc
Shixiaowei02: Thanks for the suggestion. A non-null check should be added here. At present, the data structure `Storage` is a bit messy. Such as redundant `Place` data members [here](https://github.com/PaddlePaddle/Paddle/pull/38301/files#diff-d02c2af16978fbd2c58f949c9e35b8840bb340f0e6e0169f6aca4448739f336aR115) . I will modify it uniformly in refactorings.

paddle/pten/core/storage.h
Shixiaowei02: Your suggestion is correct. I will remove this member pointer. Regarding the type of the constructor remaining the same, do you think it's ok?

jim19930609: yup, that should work.

paddle/pten/core/storage.cc
jim19930609: Agreed, I even thought you can leave some comments here for now, since we're at an intermediate state and these codes might experience another refactor in the near future.

paddle/pten/core/storage.h
jim19930609: I see, thx!

paddle/pten/core/storage.h
Shixiaowei02: Thanks!

jim19930609 (approved):
Approved.

In a consecutive PR, you may consider adding some more comments on how "Allocator* alloc_" will be modified/refactored together with the clean up for "Storage" in the future, or adding some validation checks to guarantee the robustness.

Shixiaowei02:
> Approved.
> 
> In a consecutive PR, you may consider adding some more comments on how "Allocator* alloc_" will be modified/refactored together with the clean up for "Storage" in the future, or adding some validation checks to guarantee the robustness.

Ok, I will remove the Storage member in Tensor as soon as possible and ask you to review the code.

XieYunshen (approved):

raindrops2sea (approved):

PR #38968, opened 2022-01-14
Add SquaredL2Norm functor using cub

sneaxiy:
### PR types
Function optimization

### PR changes
OPs

### Describe
The `Eigen.sum()` does not produce deterministic result on A100 GPU. Try to use the `cub` library to implement the squared L2-norm function to make producible results.


limin2021 (approved):
LGTM.

PR #39009, opened 2022-01-17
[hybrid] Fix out of memory bug

wuhuachaocoding:
### PR types

Bug fixes
### PR changes

Others
### Describe
fix a memory bug about optimizer in static graph,  the bug that does not update the op_device attribute before getting momentums leads to be out of memory.
2c48250e7ca3bbabeda001f2ef3ffff6


python/paddle/optimizer/optimizer.py
```diff
@@ -722,6 +722,13 @@ def _create_optimization_pass(self, parameters_and_grads):
                         self._append_optimize_multi_tensor_op(
                             target_block, parameters_and_grads)
         else:
+            if not framework.in_dygraph_mode():
+                params_grads_device_map = parameters_and_grads[
```
wangxicoding, line 726: 后面767-771的删掉

wuhuachaocoding: OK, done.

wangxicoding (approved):
LGTM

PR #39014, opened 2022-01-18
[Auto Parallel] Improve the dist op interface and the compatible computation

aoyulong:
### PR types
Others

### PR changes
Others

### Describe

Improve the dist op interface and the compatible computation for auto parallel


JZ-LIANG (approved):
LGTM

PR #39016, opened 2022-01-18
[Eager] Support test_imperative apply and Add a setter for EagerTensor

veyron95:
### PR types
New features 

### PR changes
APIs

### Describe
1. Support test_imperative_container_parameterlist with _test_eager_guard().
2. Add a _set_value() func for EagerTensor.


paddle/fluid/pybind/eager.h
```diff
@@ -19,6 +21,9 @@ namespace pybind {
 
 void BindEager(pybind11::module* m);
 void BindFunctions(PyObject* module);
+void InitEagerTensorWithNumpyValue(EagerTensorObject* self,
```
JiabinYang: why not declare this method in eager_method.cc or just implement this within wrapper, since it will only be used there.

JiabinYang (commented):
some comments

paddle/fluid/pybind/eager.h
veyron95: Done, declare this method in eager_method.cc now.

JiabinYang (approved):
LGTM

PR #39017, opened 2022-01-18
fix http gloo bug

kuizhiqing:
### PR types

Bug fixes

### PR changes

Others

### Describe

Can not and No need to set running false to http server when using it as gloo backend.


sandyhouse (approved):
LGTM

PR #39030, opened 2022-01-18
Changed default implementation of bfloat16 to eigen (copying files to paddle/utils)

ghost:
### PR types
Others

### PR changes
Others

### Describe
Changed the default implementation of the bfloat16 type to the one provided by the eigen library by copying files related to bfloat16 from the eigen library into paddle/utils.


ghost:
@chenwhql @Aganlengzi @lidanqing-intel
Copying only BFloat16.h from the eigen library into paddle/utils results in many compilation errors as shown here:
image
This PR tries to implement your suggested approach by copying all the required files from the Eigen library. As you can see, the amount of these files is very large and would be hard to maintain, which is why we don't recommend this approach.

PR #39035, opened 2022-01-19
[Dy2St]Replace comments with blank lines so that error messages are not misplaced

0x45f:
### PR types

Others
### PR changes

Others
### Describe

Replace comments with blank lines so that error messages are not misplaced


Aurelius84 (approved):
LGTM

PR #39062, opened 2022-01-20
Fix device_context place_.device print error

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
Fix the `place_.device` print format. It would be wrong previously because the type of `place_.device` has been changed to be `int8_t` in the PTen project.
image


chenwhql (approved):
LGTM

PR #39077, opened 2022-01-20
Make var of STRINGS dtype able to get lod_level property

LiuChiachi:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Make variables of  `core.VarDesc.VarType.STRINGS` dtype able to get `lod_level ` property, and set `lod_level = None`


zhiqiu (approved):
LGTM for framework.py

PR #38853, opened 2022-01-10
splits allocation for pten, test=develop

Shixiaowei02:
### PR types

Function optimization

### PR changes

APIs

### Describe

本提交是方案二的主框架修改部分，定义了计算库分配器的接口规格：

```
class pten::Allocator {
 public:
  using DeleterType = std::function<void(Allocation*)>;
  using AllocationPtr = std::unique_ptr<Allocation, DeleterType>;

  virtual ~Allocator() = default;
  virtual AllocationPtr Allocate(size_t bytes_size) = 0;

  virtual bool IsAllocThreadSafe() const { return false; }
};
```
框架修改后的继承关系：
```
     _______________ _______________________________________
   |               |                                       |
   |   Allocator <---  Allocator <---  Strategy Allocators |
   |    * pten *   |               * fluid *               |
    --------------- ---------------------------------------

    _______________ ____________________________________________________
   |               |                                                    |
   |   Allocation <---  Decorated Allocation <---  Strategy Allocations |
   |    * pten *   |                      * fluid *                     |
    --------------- ----------------------------------------------------
```
其中，pten 存放的是公共基类（接口类），供训推双方使用，并对原框架中用于装饰和策略的数据结构进行了更名。

PR #39145, opened 2022-01-22
Add GetBasePtr interface in paddle::memory

From00:
### PR types 
New features

### PR changes
Others

### Describe

After PR #38853 merged, the _memory::allocation::Allocation_ is a derived class of _pten::Allocation_, and the _Alloc_ interface in [malloc.h](https://github.com/PaddlePaddle/Paddle/blob/36d9a364d2822c8034cdd41c1f06cd9feefd2fdb/paddle/fluid/memory/malloc.h#L27) return a _pten::Allocation_ type. Since the _base_ptr_ function introduced in #37978 is implenmemted in _memoy::allocation::Allocation_, now we have to get the base-pointer from an allocation like:
```C++
std::shared_ptr<pten::Allocation> allocation = memory::AllocShared(place_, size);
void* base_ptr = static_cast<memory::allocation::Allocation*>(allocation.get())->base_ptr();
```
It is so inconvenient compared with the previous method:
```C++
void* base_ptr = allocation->base_ptr();
```

This PR adds a GetBasePtr interface in paddle::memory that hides the inelegant _static_cast_ in _AllocatorFacadePrivate_, therefore we can get the base-pointer by a simple way like:
```C++
void* base_ptr = memory::GetBasePtr(allocation);
```

### DIFF
--- a/paddle/fluid/memory/CMakeLists.txt
+++ b/paddle/fluid/memory/CMakeLists.txt
@@ -34,6 +34,13 @@ if (WITH_ROCM)
             DEPS device_context malloc)
 endif()
 
+if(WITH_GPU AND WITH_TESTING AND NOT "$ENV{CI_SKIP_CPP_TEST}" STREQUAL "ON")       
+  nv_test(get_base_ptr_test SRCS get_base_ptr_test.cu DEPS malloc gpu_info)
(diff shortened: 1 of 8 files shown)


zhiqiu (approved):
LGTM

Shixiaowei02 (approved):

PR #39150, opened 2022-01-23
Threadsafe program_desc_tracer

tongxin:
### PR types
Bug fixes

### PR changes
Others

### Describe
Current implementation of ProgramDescTracer::InsertOp is not threadsafe. As shown below, the `vars_` in  ProgramDescTracer points to a map which is vulnerable under multhreading environment. 

```
class ProgramDescTracer {
  DISABLE_COPY_AND_ASSIGN(ProgramDescTracer);

 public:
  ProgramDescTracer() = default;

 private:
  std::vector<std::unique_ptr<OpDescMeta>> ops_;
  VarDescMetaMap vars_;                       // <-------- a std::map maintaining var name to varbase pointer mapping
  VarBaseSet non_exist_input_vars_;
};
```

Potential mutation of the vars map happens inside `ProgramDescTracer::InsertVarIfNotExist`.  Since operator [] is a non-`const` member function of map, its returned reference is not thread safe according to the C++ language specification. Updating a map entry should be properly synchronized to avoid a race condition.

```
   vars_[new_var].reset(new_var_desc);
```


paddle/fluid/imperative/jit/program_desc_tracer.cc
```diff
@@ -221,40 +220,47 @@ void ProgramDescTracer::InsertVarIfNotExist(
     const std::shared_ptr<VarBase> &new_var, bool is_input) {
   PADDLE_ENFORCE_NOT_NULL(new_var, platform::errors::InvalidArgument(
                                        "The variable to insert is NULL."));
-  if (vars_.count(new_var) != 0) return;
 
-  auto new_var_desc = new framework::VarDesc("");
-  vars_[new_var].reset(new_var_desc);
+  for (; vars_.count(new_var) == 0;) {
+    if (!vars_mutex_.try_lock()) continue;
```
sneaxiy: Some questions:

- When would the `ProgramDescTracer` run in multiple threads? Users usually use imperative programming in a single thread.
- It is still not thread safe after this PR. There may be 2 threads running in the same time: thread 1 runs `vars_.count(new_var)` and thread 2 runs `vars_[new_var]`. There is still data race.

paddle/fluid/imperative/jit/program_desc_tracer.cc
```diff
@@ -221,40 +220,47 @@ void ProgramDescTracer::InsertVarIfNotExist(
     const std::shared_ptr<VarBase> &new_var, bool is_input) {
   PADDLE_ENFORCE_NOT_NULL(new_var, platform::errors::InvalidArgument(
                                        "The variable to insert is NULL."));
-  if (vars_.count(new_var) != 0) return;
 
-  auto new_var_desc = new framework::VarDesc("");
-  vars_[new_var].reset(new_var_desc);
+  for (; vars_.count(new_var) == 0;) {
```
sneaxiy: How about just write `while (...)`?

paddle/fluid/imperative/jit/program_desc_tracer.cc
tongxin: Thanks for your comments ;) 
On the second point first. The C++ standard stipulates that a const-member function of a container class should be thread-safe by default. I suppose std:: implementation follows the standard.  
Regarding the first question, there seem to be quite a few places in `imperative.cc` where the python GIL is released right before calling `tracer->traceop`. I'm not sure why multithreading happens in the imperative mode in the first place, it seems so...

paddle/fluid/imperative/jit/program_desc_tracer.cc
tongxin: Sure. it's just a personal preference. When working with low level C code I would prefer `for` loop .

paddle/fluid/imperative/jit/program_desc_tracer.cc
tongxin: Some background information. The issue was raised during an 8-GPU training and the triggering point is in ProgramDescTracer::InsertVarIfNotExist..

tongxin: Agree. Both `vars_.count()` and `vars_[]` should be protected..

PR #39157, opened 2022-01-23
Mish unittests

Silv3S:
### PR types
New features

### PR changes
OPs

### Describe
Add unittests to the new UT framework for:
- mish operator,
- conv + mish fuse pass
- fc + mish fuse pass


Silv3S:
Continued in https://github.com/PaddlePaddle/Paddle/pull/39340 due to CI bug (non refreshing checks)

PR #39165, opened 2022-01-24
[MLU]add mlu kernels for sum op

joeqiao12:
### PR types
New features 

### PR changes
OPs

### Describe
add mlu kernels for sum op


qili93 (approved):
LGTM

PR #39182, opened 2022-01-24
fix paddle.where broadcast bug

ronny1996:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix where broadcast bug

修复需要广播时，x中对应元素为inf，返回值为nan


qili93 (approved):
LGTM

PR #39221, opened 2022-01-25
Symbolic Hessian

tongxin:
### PR types
New features

### PR changes
APIs

### Describe

Computes the Hessian of second-order differentiable function. 

Code example:

```
from paddle.autograd.functional import Hessian
main = fluid.Program()
startup = fluid.Program()
with fluid.program_guard(main, startup):
    x = paddle.static.data('x', [2, 2, 2], dtype='float32')
    HH = paddle.autograd.functional.Hessian(f, xs, batch=True)
    full_hessian = HH[:]
exe = fluid.Executor(self.place)
exe.run(startup)
pd_hess = exe.run(main, feed={x: np.array([[[2., 2.], [2., 1.]], [[1., 2.], [2., 1.]]])}, fetch_list=[full_hessian])[0]

```

Another contribution of this PR is resolving an issue pertaining to building backward computation when higher-order gradients are present. The issue was due to incorrect input renaming when adding a `grad op` whose forward counterpart is also a `grad op`.  The scenario in question is shown as follows.  The forward op of XX_grad_grad is XX_grad. The forward op and backward op share some inputs. In case1, the share input's name X@GRAD conflicts with a gradient output's previous name and it should not be renamed. In case2, no conflict exists and the input variable should be renamed. However, the current implementation cannot tell the difference. The fix goes by checking variable names against the forward op's inputs first. If it's not the forward op's input then we apply the renaming check. 

```
case 1:

           X@GRAD --> Op(XX_grad)   
              |
              | 
              |                       grad/X@GRAD    (renamed from X@GRAD) <-- Op(Some_other_grad)
              |    
              |
              |--> Op(XX_grad_grad)

case 2:

           X@GRAD --> Op(XX_grad)   
              |
              | 
              |                       grad/Y@GRAD    (renamed from Y@GRAD) <-- Op(Some_other_grad)
              |                                |
              |                                |
              |--> Op(XX_grad_grad)<-----------|


```


levi131 (dismissed):
LGTM

zhwesky2010 (dismissed):
LGTM for change ut name

JiabinYang (approved):
LGTM

levi131 (approved):

chenwhql (approved):
LGTM  for backward.py

XieYunshen (approved):
LGTM
修改单测名称

PR #39349, opened 2022-02-05
Fix reduce_sum dtype dispatch bug on gpu

chenwhql:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix reduce_sum dtype dispatch bug on gpu

There is a bug with the implementation of `paddle.sum` on gpu, `paddle.sum` has a `dtype` argument, If specified, the input tensor is casted to dtype before the operation is performed.

`paddle.sum` corresponds to `reduce_sum` op. The CPU Kernel implementation of this op is consistent with the previous description, that is, when specifying dtype, cast first and then execute reduce sum; However, the implementation of the GPU kernel here is not complete, and the kernel actually only supports The api argument dtype is specified as `float32, float64, int32, int64, complex64, complex128`, a total of six types. When the specified dtype is not the above six types, api execution will fail, but for this implicit setting, the api document does not mention。

Therefore, this PR makes the behavior of gpu and cpu consistent, and solves the problem of incomplete types of GPU kernel support.

### DIFF
--- a/paddle/pten/api/ext/dispatch.h
+++ b/paddle/pten/api/ext/dispatch.h
@@ -302,4 +302,44 @@ namespace paddle {
     }                                                                         \
   }()
 
+#define PD_VISIT_BOOL_AND_FLOATING_AND_COMPLEX_AND_3_TYPES(                   \
+    SPECIFIED_TYPE1, SPECIFIED_TYPE2, SPECIFIED_TYPE3, TYPE, NAME, ...)       \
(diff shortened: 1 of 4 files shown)


YuanRisheng (approved):

MingMingShangTian (approved):

ZzSean (approved):
LGTM for ci op benchmark

sneaxiy (approved):
LGTM

PR #39370, opened 2022-02-07
[bf16] change bf16 c++ print behavior

zhiqiu:
### PR types

Breaking changes
### PR changes

Others
### Describe

[bf16] change bf16 print behavior

Original implementation prints `bfloat` number as uint16, this PR changes to `float`.
For example, bfloat number 1.0f is printed as `16256` before, and `1` now.


wozna (approved):
LGTM

zhangbo9674 (approved):
LGTM

PR #39381, opened 2022-02-08
Make Embedding layer support more int ids type

sneaxiy:
### PR types
Function optimization

### PR changes
OPs

### Describe
Make `Embedding` layer supports more int `Ids` type, such as `int16`, `int8`, etc.


chenwhql (approved):
LGTM

PR #39385, opened 2022-02-08
[PluggableDevice] custom kernel supports multi cpp_dtypes

Aganlengzi:
### PR types
New features

### PR changes
Others

### Describe
Custom kernel supports multi cpp_dtypes with upgrade of macro PD_REGISTER_KERNEL.
```
template <typename T, typename Context>
void FakeDot(const Context& dev_ctx, ...) {
  // custom kernel logic
}

PD_REGISTER_KERNEL(fake_dot,
                   CPU,
                   ALL_LAYOUT,
                   custom_kernel::FakeDot,
                   float,
                   double,
                   int, 
                   int64_t, 
                   int8_t, 
                   uint8_t) {}
```
More work will be done when pten core foundation can be used, but it will be transparent to outer users.

### DIFF
--- a/paddle/fluid/framework/custom_kernel_test.cc
+++ b/paddle/fluid/framework/custom_kernel_test.cc
@@ -35,13 +35,12 @@ limitations under the License. */
 // user kernel function
 namespace custom_kernel {
 
-// Here we use dot <CPU, ANY, UINT8> for test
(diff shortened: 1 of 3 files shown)


qili93 (approved):
LGTM

PR #39394, opened 2022-02-08
[MLU]fix compile and add cncl

qipengh:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix compile error and add cncl compile


qili93 (approved):
LGTM

PR #39403, opened 2022-02-08
[pten] fit pten for amp

zhiqiu:
### PR types

Function optimization
### PR changes

Others
### Describe

Amp module needs to get all operators that not support float16, while some pten kernels are not counted before, this PR enhance that.


zhangbo9674 (dismissed):
LGTM

zhangbo9674 (approved):
LGTM

PR #39406, opened 2022-02-08
Quantize elementwise mul operator

zuzg:
### PR types
New features

### PR changes
OPs

### Describe
Quantization of elementwise multiplication operator.


lidanqing-vv:
@wozna Could you please review this PR ?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass_tester.cc
```diff
@@ -588,6 +589,48 @@ TEST(CpuQuantizePass, elementwise_add_unsigned_and_signed_input) {
            expected_operators, added_nodes, 1.f, 1.f, "", "b");
 }
 
+static const std::initializer_list<std::string> variable_names_elementwise_mul =
```
wozna: Here code for elementwise_add and elementwise_mul should be the same. So you can parametrize `BuildProgramDescElementwiseAdd()` to `BuildProgramDescElementwise(const std::string elementwise_type)`. It will simplify test for next elementwise ops. You can reuse similar solution that was added for immutable ops in `TestImmutableOpWithManyOutputs`

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -877,6 +877,75 @@ void CPUQuantizePass::QuantizeElementwiseAdd(Graph* graph) const {
                   quantize_elementwise_add_count);
 }
 
+void CPUQuantizePass::QuantizeElementwiseMul(Graph* graph) const {
```
wozna: Please check if you can parametrize this function because code for `QuantizeElementwiseAdd` and `QuantizeElementwiseMul` should be the same, so you can just add type for `QuantizeElementwise(Graph* graph, const std::string elementwise_type)`

PR #39417, opened 2022-02-09
Make the result of lookup_v2_grad determined.

limin2021:
### PR types
Others

### PR changes
OPs

### Describe
Add the logic to make the result of mlperf bert determined.
1.make the atomic add serial in lookup_table_v2_grad.
2.make the scale and bias add in layer_norm op computed in fp16 to be consistent with nv mlperf bert.

We can use deterministic logic by export FLAGS_mlperf_bert_deterministic=1.

Note: when test performance, remember to set  FLAGS_mlperf_bert_deterministic=0(default).


paddle/fluid/operators/lookup_table_v2_op.cu
```diff
@@ -199,9 +199,18 @@ struct LookupTableV2GradCUDAFunctor {
       auto t = framework::EigenVector<T>::Flatten(*d_table_t);
       t.device(*dev_ctx.eigen_device()) = t.constant(static_cast<T>(0));
 
-      LookupTableV2Grad<T, IdT, 128, 8,
-                        8><<<grids, threads, 0, dev_ctx.stream()>>>(
-          d_table, d_output, ids, N, K, D);
+      // todo(@limin): need to set to false when test performance.
+      bool is_determine_result = true;
```
sneaxiy: How about add a `FLAGS` to control the branch here? For example, add the code in the global namespace:

```C++
PADDLE_DEFINE_EXPORTED_bool(lookup_table_deterministic, false, 
                            "Whether to run the deterministic lookup_table op.");
```

In this way, users can use `export FLAGS_lookup_table_deterministic=1` to control the code logic, instead of re-compiling the code.

You can also reuse the existing flags `FLAGS_cudnn_deterministic` instead of adding new flags, but maybe the semantic is not right.

limin2021: Done.

sneaxiy (commented):
Great work!

PR #39443, opened 2022-02-10
fix dtype bug in assign

Caozhou1995:
### PR types

Bug fixes
### PR changes

Others 
### Describe

If the x is list or tuple, np.array will make it be FP64, but paddle just supports FP32. In this scene, the input dtype is FP64, but dtype is FP32. This pr makes output dtype is FP32 when input dtype is FP64.


Caozhou1995:
This PR is replaced by PR:https://github.com/PaddlePaddle/Paddle/pull/41802

PR #39487, opened 2022-02-11
Ignore eof

liutiexing:
### PR types
Function optimization

### PR changes
Others

### Describe
EOF is not a fatal, ignore it when doing graceful exit.


Aurelius84 (approved):
LGTM

PR #39498, opened 2022-02-13
Disabled int_reorders profiling for non weights/bias reorders

jakpiase:
### PR types
Others

### PR changes
Others

### Describe
Disabled `int_reorders` profiling for non weights/bias reorders. Now if a reorder is a part of op's computation(such as quantization), then it will not be displayed as `int_reorder`. Only weights and bias reorders will be displayed as `int_reorder` for now.

### DIFF
--- a/paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
@@ -69,9 +69,6 @@ static Tensor FoldFirstAndLastDims(const MKLDNNDeviceContext& dev_ctx,
   auto reorder_p = reorder_handler.AcquireReorder(reorder_src_memory_p,
                                                   reorder_dst_memory_p);
 
-  paddle::platform::RecordEvent record_reorder(
-      "int_reorder", paddle::platform::EventRole::kUniqueOp);
-
   auto& astream = MKLDNNDeviceContext::tls().get_stream();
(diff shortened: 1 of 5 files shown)


jakpiase:
@Silv3S please review this PR

jczaja:
@jakpiase  Why this change?

Silv3S:
Shouldn't quantization also be profiled? If we delete it, we won't be able to track performance with new versions of oneDNN. Maybe just change event name from "int_reorder" to "op_int_reorder" if you want to distinguish them?

jakpiase:
@jczaja we have discussed it offline a long time ago. If a reorder is a part of op's computing(such as special case of reduce op when no reduction is applied, only data copying), then displaying it as `int_reorder` is misleading since it was designed to display weights reordering time compared to total op's computation time.
@Silv3S quantization will be profiled as well, but `int_reorder` won't be displayed in that scenario
Below you have the example how this change will work:

Before this PR:
image

After this PR:
image

`int_reorder` is deleted from `quantize` op profiling, because it's a part of main computation and gives very little or no meaning and just bloats this report. `conv2d`'s `int_reoder` will stay as it was, because it is associated with weights and bias reordering and gives us some valid information how this op behaves.

Silv3S:
LGTM

jczaja (approved):
LGTM

jakpiase:
@baoachun please review this PR

baoachun (approved):
LGTM

PR #39554, opened 2022-02-15
Modify the implementation of BlockXReduce to fit more scenes

ZzSean:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Modify the implementation of BlockXReduce to fit more scenes

After BlockXReduce, every thread with same threadIdx.y should get the correct result.


Xreki (approved):
LGTM

PR #39561, opened 2022-02-15
[Paddle-TRT] Test only trt group norm

zlsh80826:
### PR types

Others

### PR changes

Others

### Describe

1. Unit test should cover only the specific feature, so `test_trt_group_norm_op.py` should only test the `trt_group_norm` without other ops. If it is important to test the combination of `relu` + `relu6` + `tanh` + `conv2d` + `group_norm`, we could add another unit test.
2. The random input batch size is only 1, set the `max_batch_size` to 1


zlsh80826:
@Wangzheee please help review the PR

Wangzheee (approved):
LGTM

baoachun (approved):
LGTM

PR #39608, opened 2022-02-16
Shared selected rows

JiabinYang:
### PR types

Function optimization
### PR changes

Others
### Describe

This PR is trying to remove friend class in SelectedRows and make `framework::SelectedRows` a wrapper class with `shared_ptr` of origin `SelectedRows`. In that way, we can copy it in eager mode  instead of always move it which will cause lots problems.


Shixiaowei02 (approved):
Approve for inference use

chenwhql (approved):

PR #39622, opened 2022-02-16
save the name lists of variables of a cinn subgraph as its attributes

CtfGo:
### PR types

Others
### PR changes

Others
### Describe

save the name lists of the input, internal and output variables of a cinn subgraph as its attributes while building subgraphs in `build_cinn_pass`


thisjiang (approved):
LGTM

PR #39628, opened 2022-02-16
Support NoNeedBuffer for final state codegen

jim19930609:
### PR types
New features

### PR changes
Others

### Describe
Op developers may now write "no_need_buffer: X, Y" in Yaml file, so as to drop the underlying allocation for Tensor X and Tensor Y when creating TensorWrapper. 

77d95802797988a3134174a40fd05a83

AutoCoden script will detect "no_need_buffer" field and set a special boolean when constructing TensorWrapper. In that scenario, TensorWrapper will only copy everything except for the underlying allocation from Tensor X or Tensor.

993c7be9eefdef4727dec7c7adb274ec


paddle/fluid/eager/tensor_wrapper.h
```diff
@@ -48,7 +49,22 @@ class TensorWrapper {
     }
 
     // shallow copy tensor_impl here
-    intermidiate_tensor_.set_impl(tensor.impl());
+    if (no_need_buffer) {
+      if (pten::DenseTensor::classof(tensor.impl().get())) {
```
zyfncg: 这里的`pten`估计有代码冲突

zyfncg (approved):

PR #39673, opened 2022-02-17
fix bug: Test_range_xpu crashes when missing range XPU kernel in kunlun2

ShiningZhang:
### PR types

Bug fixes
### PR changes

OPs
### Describe

The Test_range_xpu will crash in kunlun2 with input data of type int32.  The input will not be transformed from xpu to cpu when missing xpu kernel and fallbacking to cpu kernel. Fix it.


shangzhizhou (approved):
LGTM

PR #39689, opened 2022-02-17
[Dy2St]Fix cond grad error when handle tensor array

0x45f:
### PR types

Others
### PR changes

Others
### Describe

fix cond grad error when handle tensor array


Aurelius84 (approved):
LGTM

PR #39700, opened 2022-02-17
[Paddle-TRT] Fix test_trt_dynamic_shape_transformer_prune wrong inputs

zlsh80826:
### PR types

Bug fixes

### PR changes

Others

### Describe

The input order of model used trt_dynamic_shape_transformer_prune is 
`token_ids`, `pos_ids`, `sent_ids`, `mask`.
But the original input order specify `token_ids`, **`sent_ids`**, **`pos_ids`**, `mask`, which results random illegal memory access(pos_id is much greater than sent_ids embedding).


zlsh80826:
image

input1 has shape `[513, 128]` which is for `pos_id`, input2 has shape `[2, 128]` which is for `sent_id`. The wrong order will cause illegal memory access when looking up table

Wangzheee (approved):
LGTM

Issue #39489, opened 2022-02-11
Huggingface Hub


AK391:
Hi, would you be interested in adding paddlepaddle to Huggingface Hub ([Hugging Face Hub](https://huggingface.co/models))? The Hub offers free hosting, and it would make your work more accessible and visible to the rest of the ML community. I setup a example paddlepaddle organization where models, datasets, and spaces(web demos) can be added similar to the paddlepaddle org on github.

PaddlePaddle: https://huggingface.co/PaddlePaddle

Example from other organizations:
Keras: https://huggingface.co/keras-io
Microsoft: https://huggingface.co/microsoft
Facebook: https://huggingface.co/facebook

<img width="354" alt="Screen Shot 2022-02-11 at 10 24 34 AM" src="https://user-images.githubusercontent.com/81195143/153619443-7791031a-b91b-45be-ade9-96438bf6c0ce.png">

Here is how to integrate a library in the Hub: https://huggingface.co/docs/hub/adding-a-library

Integrating the Hub with your library provides many benefits, including:

- Free model hosting for you and your users.
- Built-in file versioning - even for huge files - made possible by [Git-LFS](https://git-lfs.github.com/).
- All public models are powered by the [Inference API](https://api-inference.huggingface.co/docs/python/html/index.html).
- In-browser widgets allow users to interact with your hosted models directly.

Please let us know if you would be interested and if you have any questions, we can also help with the technical implementation.

iclementine:
Thank you for your invitation! I believe developers of PaddleNLP would be interested and I will get them inforrmed.

ZeyuChen:
@AK391 I am Zeyu, the tech lead of PaddleNLP and PaddleSpeech, and already sent a request to join the group. we are glad to participate hugging face hub project. Please let me know how can we move forward. Thanks.

AK391:
@iclementine thank you, @ZeyuChen great, accepted the invite, you can now add models, spaces, and datasets to the org

How to add a Space: https://huggingface.co/blog/gradio-spaces
how to add models: https://huggingface.co/docs/hub/adding-a-model
uploading a dataset: https://huggingface.co/docs/datasets/upload_dataset.html

for integrating a library to the Hub, please see: https://huggingface.co/docs/hub/adding-a-library. Let me know if you have any questions.

AK391:
[...]

PR #39705, opened 2022-02-17
add HF org badge

AK391:
added badge for organization on huggingface see https://github.com/PaddlePaddle/Paddle/issues/39489


AK391:
@rainyfly please review

PR #39736, opened 2022-02-18
add index initialization in the block loop for index_sample kernel when dealing with a input tensor whose shape is larger than block_dim * grid_dim

FlyingQianMM:
### PR types
Bug fixes

### PR changes
OPs

### Describe
When the length of input tensor is large than block_dim * grid_dim，the index_sample kernel would not deal with the exceeding part. So block and grid loop have been added in the kernel.  see https://github.com/PaddlePaddle/Paddle/pull/37816

But in the block loop, the index has not been initialize, which may cause that the exceeding part would not be computed, so we add the intialization for index of the block loop.


thisjiang (approved):
LGTM

PR #39738, opened 2022-02-18
Add mobilenetv3_large performance test for bf16 and int8 

wozna:
### PR types

New features
### PR changes

Others
### Describe

This PR:
- adds test for int8 and bf16 image classification model MobilenetV3_large_1x_0. Because this model doesn't have an accuracy layer I adjusted testing files and tester_helper.cc file to support preparing data and testing this kind of model. 
- corrects `paddle/fluid/inference/tests/api/CMakeLists.txt` because I discovered that defined downloading `quant_lstm_input_data.tar.gz` was happening every time the build was done because the path checked was wrong
- corrects BF16 test: disable the test if the device doesn't support BF16 computations because it takes too much time.

### DIFF
--- a/paddle/fluid/inference/tests/api/CMakeLists.txt
+++ b/paddle/fluid/inference/tests/api/CMakeLists.txt
@@ -763,6 +763,7 @@ if(WITH_MKLDNN)
     set_tests_properties(test_analyzer_quant_performance_benchmark PROPERTIES TIMEOUT 120)
     set_tests_properties(test_analyzer_int8_mobilenetv2 PROPERTIES TIMEOUT 120)
     set_tests_properties(test_analyzer_int8_mobilenetv1 PROPERTIES TIMEOUT 120)
+    set_tests_properties(test_analyzer_int8_mobilenetv3_large PROPERTIES TIMEOUT 120)
 endif()
 
 set_tests_properties(lite_resnet50_test PROPERTIES TIMEOUT 120)
--- a/paddle/fluid/inference/tests/api/CMakeLists.txt
+++ b/paddle/fluid/inference/tests/api/CMakeLists.txt
@@ -451,6 +451,23 @@ if(WITH_MKLDNN)
   download_int8_data_without_verify(${INT8_GOOGLENET_MODEL_DIR} "GoogleNet_int8_model.tar.gz" )
   inference_analysis_api_int8_test_run_custom_warmup_batch_size(test_analyzer_int8_googlenet ${INT8_IMG_CLASS_TEST_APP} ${INT8_GOOGLENET_MODEL_DIR} ${IMAGENET_DATA_PATH} 10)
 
+ # mobilenetv3_large_x1_0 int8
+ set(INT8_MOBILENETV3_LARGE_MODEL_DIR "${INT8_DATA_DIR}/mobilenetv3_large")
+ set(INT8_MOBILENETV3_FILE_NAME "MobileNetV3_large_x1_0_infer.tar")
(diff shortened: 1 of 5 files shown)


paddle/fluid/inference/tests/api/analyzer_int8_image_classification_tester.cc
```diff
@@ -22,7 +22,12 @@ namespace inference {
 namespace analysis {
 
 void SetConfig(AnalysisConfig *cfg) {
-  cfg->SetModel(FLAGS_infer_model);
+  std::ifstream model_file(FLAGS_infer_model + "/__model__");
```
lidanqing-vv, line 25: Where is this refering ? Baidu official UT ?

wozna:
@lidanqing-intel  @sfraczek @zuzg  Please add your review

zuzg:
LGTM

paddle/fluid/inference/tests/api/CMakeLists.txt
```diff
@@ -735,6 +763,7 @@ if(WITH_MKLDNN)
     set_tests_properties(test_analyzer_quant_performance_benchmark PROPERTIES TIMEOUT 120)
     set_tests_properties(test_analyzer_int8_mobilenetv2 PROPERTIES TIMEOUT 120)
     set_tests_properties(test_analyzer_int8_mobilenetv1 PROPERTIES TIMEOUT 120)
+    set_tests_properties(test_analyzer_int8_mobilenetv3_large PROPERTIES TIMEOUT 120)
```
lidanqing-vv, line 766: I think I read in somebody's PR that TIMEOUT is not necessary if it does not take too much time @sfraczek.
Could you please confirm in the team ? If this line does not exist, will the test pass ?

lidanqing-vv: To make the test fast you may change the bs to 50 maybe

lidanqing-vv: This test test_analyzer_int8_mobilenetv3_large is quite slow because it tests FP32, calculate scales for PTQ and test INT8 and it had timeout in Coverage-CI that is why extended timeout value.

paddle/fluid/inference/tests/api/analyzer_int8_image_classification_tester.cc
wozna: This is the test that tests PTQ int8 with OneDNN, therefore I adapted this test to the models we are testing.

lidanqing-vv (approved):
LGTM

wozna:
@XieYunshen or @chalsliu Could you review TIMEOUT setting?

paddle/fluid/inference/tests/api/CMakeLists.txt
sfraczek: I think we can extend timeout to 300 like in examples above this. Will it be enough?

baoachun (approved):
LGTM

paddle/fluid/inference/tests/api/CMakeLists.txt
wozna: This set of 120 is already enough.

chalsliu:
> 

Is it necessary to set the timeout to 120 seconds?

wozna:
@chalsliu Yes, this timeout is needed, because in this test we check the FP32 test, then we run warmup for the post-training quantization process and then we run the obtained model for int8. 120 seconds is the maximum time that similar models need to perform this test. Do you think this limit should be lower?

chalsliu:
> 

OK, got it. Thanks for the explanation.

chalsliu (approved):

PR #39747, opened 2022-02-19
Fix DistributedFusedLamb alignment bug

sneaxiy:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix the bug that the numel of the fused tensor inside `DistributedFusedLamb` op should be a factor of `alignment * nranks` instead of `lcm(alignment, nranks)`.


limin2021 (approved):
LGTM.

PR #39782, opened 2022-02-21
build_cinn_pass: fix bug because of output control var

CtfGo:
### PR types

Bug fixes
### PR changes

Others
### Describe

skip control variables in setting the output variables of a `cinn_launch` operator


thisjiang (approved):
LGTM

PR #39823, opened 2022-02-22
【Auto Parallel】update reshard for while op

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

Update reshard module for while op.


Caozhou1995:
This pr version is not the same with develop, close it and push again

PR #39842, opened 2022-02-23
fix 'invalid escape sequence'

zhiqiu:
### PR types

Function optimization
### PR changes

Others
### Describe

fix 'invalid escape sequence'
image


TCChenlong (approved):
LGTM

PR #39846, opened 2022-02-23
【Auto Parallel】update reshard to support while sub block

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

Update reshard module to support while sub block.


Caozhou1995:
This pr close by repr https://github.com/PaddlePaddle/Paddle/pull/40366

PR #39860, opened 2022-02-23
support ema optimizer in sharding optimizers

sljlp:
### PR types

New features
### PR changes

APIs
### Describe

When using EMA in sharding optimziers, the trainers whose ids are not 0 don't save EMA vars.
However, EMA parameters need to be saved in every trainer.
Now the feature is supported by add a member "_ema_0" in checks list in function is_opt_vars.


ForFishes (approved):
LGTM

PR #39874, opened 2022-02-23
Support slim eager

JiabinYang:
### PR types

Function optimization
### PR changes

Others
### Describe

This PR is trying to support slim and qat in eager mode:
1. Fix bug when we deal with input as outpus.
2. Support slice related Python C method and move it to eager_utils.h. Make it dependence of imperative.cc
3. Remove imperative.cc from `ON_INFER`, since in inference we don't need imperative
4. Fix save problem in `jit.save` and `paddle.save`, now we can support `jit.save` and `paddle.save` in eager mode


XiaoguangHu01 (dismissed):
LGTM

phlrain (approved):

XiaoguangHu01 (approved):

XieYunshen (approved):
LGTM

PR #39886, opened 2022-02-24
[Fix bug] fix fp16 atomicAdd compiler error on different cuda_arch.

limin2021:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix fp16 atomicAdd compiler error on different cuda_arch.
Note atomicAdd(half2* ) is only supported on  nvidia gpu with cuda_arch >= 700 and cuda_version >= 10000;


sneaxiy (approved):
LGTM.

Issue #39881, opened 2022-02-24
使用混合精度报错


jyjfjyjf:
代码：

    model, optimizer = paddle.amp.decorate(models=model, optimizers=optimizer, level=f16, master_weight=None, save_dtype=None)
    
    global_step = 0
    for epoch in range(1, epochs + 1):
        correct = 0
        data_num = 0
        batch_id = 0
    
        model.train()
    
        for data in tqdm(train_loader, desc=f'epoch {epoch} training'):
            input_ids = paddle.to_tensor(data[0])
            token_type_ids = paddle.to_tensor(data[1])
            labels = paddle.to_tensor(data[2])
    
            input_data = {'input_ids': input_ids,
                        'token_type_ids': token_type_ids,
                        'labels': labels}
    
            with paddle.amp.auto_cast(enable=True, custom_black_list=[
                "reduce_sum",
                "c_softmax_with_cross_entropy",
                "elementwise_div"
            ], level=f16):
                outputs = model(**input_data)
                logits = outputs[1]
                loss = outputs[0]
    
            probs = F.softmax(logits, axis=-1)
    
            correct = metric.compute(paddle.to_tensor(probs.cpu().detach().numpy()),
                                    paddle.to_tensor(labels.cpu().detach().numpy()))
            metric.update(correct)
            acc = metric.accumulate()
    
            global_step += 1
            if global_step % 100 == 0:
                logger.info("global step %d, epoch: %d, batch: %d, loss: %.5f, acc: %.5f" % (
                    global_step, epoch, batch_id, loss, acc))
    
            if f16 == 'O2':
                scaled = scaler.scale(loss)
                scaled.backward()
                scaler.step(optimizer)  
                scaler.update()

报错信息

              Traceback (most recent call last):
                File "C:\Users\woait\AppData\Roaming\JetBrains\IntelliJIdea2021.3\plugins\python\helpers\pydev\pydevd.py", line 1483, in _exec
                  pydev_imports.execfile(file, globals, locals)  # execute the script
                File "C:\Users\woait\AppData\Roaming\JetBrains\IntelliJIdea2021.3\plugins\python\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
                  exec(compile(contents+"\n", file, 'exec'), glob, loc)
                File "C:/JYN/workspace/DeBERTa_Decoding-enhanced_BERT_with_Disentangled_Attention/tools/train_eval.py", line 187, in <module>
[...]

PR #39950, opened 2022-02-25
fix pylayer problem with amp

zhiqiu:
### PR types

Bug fixes 
### PR changes

APIs
### Describe

When using PyLayer with amp, the `forward` part of  PyLayer may be invoked with `amp enabled`, while the `backward` part is not, thus resulting in errors.

related #39881

```
import paddle
from paddle.autograd import PyLayer

class MyMM(PyLayer):
    @staticmethod
    def forward(ctx, a, b):
        ctx.save_for_backward(a, b)
        return a.mm(b)
    
    @staticmethod
    def backward(ctx, grad):
        a, b = ctx.saved_tensor()  
        # NOTE(zhiqiu): a and b is float32 now, while grad is fp16 when forward runs with auto_cast()
        # thus, the mm operation raise errors because of the dtype of inputs are inconsistent
        return grad.mm(b.t()), a.t().mm(grad)

x = paddle.rand([10, 10])
y = paddle.rand([10, 10])
x.stop_gradient=False
y.stop_gradient=False

with paddle.amp.auto_cast():
   res = MyMM.apply(x, y)
   loss = paddle.mean(res)

loss.backward()
```

- before 
  run failed.
image
- after
 run ok

### DIFF
--- a/python/paddle/autograd/py_layer.py
+++ b/python/paddle/autograd/py_layer.py
@@ -14,6 +14,8 @@
 
 import paddle
 from paddle.fluid.framework import dygraph_only
+from paddle.fluid.dygraph.amp.auto_cast import amp_state
+from paddle.amp.auto_cast import auto_cast
 from paddle.fluid import core
 __all__ = []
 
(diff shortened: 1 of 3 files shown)


chenwhql (dismissed):
LGTM

zhangbo9674 (dismissed):
LGTM

zhangbo9674 (approved):
LGTM

PR #39963, opened 2022-02-27
Support initializing specific grad tensors to zero for selected operators

jim19930609:
### PR types
New features

### PR changes
Others

### Describe
Taking split_op as an example, the grad operation of which is to concat all the grad inputs into one. If, however, any of the grad inputs is uninitialized, we initialize it with zero value by default


JiabinYang (approved):
LGTM

PR #40005, opened 2022-02-28
refactoring proposal conv+elementwise_add (residual)

sfraczek:
### PR types
Others

### PR changes
OPs

### Describe
This is proposal of refactoring conv+elementwise_add (residual) to make it more readable. This way it is more aligned with design of other fuse passes. There is less code and less declarations in header files.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/conv_elementwise_add_mkldnn_fuse_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/conv_elementwise_add_mkldnn_fuse_pass.cc
@@ -22,6 +22,7 @@
 
 #include "paddle/fluid/framework/ir/graph_traits.h"
 #include "paddle/fluid/framework/op_version_registry.h"
+#include "paddle/fluid/string/pretty_log.h"
 
 namespace paddle {
 namespace framework {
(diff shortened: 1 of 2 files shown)


jakpiase (commented):
Amazing job, I really appreciate fixing this abomination. Do you think that merging `FuseConvAsX` and `FuseConvAsY` is easily doable? If yes then it would be nice to do it, but even in the current situation the code is also much more readable than before

paddle/fluid/framework/ir/mkldnn/conv_elementwise_add_mkldnn_fuse_pass.cc
```diff
@@ -374,39 +287,84 @@ GraphWithStats ResidualConnectionMKLDNNFusePass::FuseProjectionConv(
...
+           << " projection conv (as y) + elementwise_add patterns";
+    paddle::string::PrettyLogDetail(msg_ss.str().c_str());
+  }
+
+  return std::make_pair(graph_with_stats.first,
+                        found_projection_conv_count + graph_with_stats.second);
 }
 
-void ResidualConnectionMKLDNNFusePass::ApplyImpl(graph_ptr graph) const {
+void ResidualConnectionMKLDNNFusePass::ApplyImpl(ir::Graph* graph) const {
```
Silv3S, line 360: Could you unify naming convention of `ir::Graph* graph` and `Graph* g`?

Silv3S (approved):
LGTM 👍

paddle/fluid/framework/ir/mkldnn/conv_elementwise_add_mkldnn_fuse_pass.cc
sfraczek: I think this has been kind of a standard because in conv_activation and conv_bn there is Graph g in handler and ir::Graph graph otherwise so I made it the same. I think we could make it ir::Graph graph everywhere but we could make a unified change everywhere in other PR.

sfraczek:
> Amazing job, I really appreciate fixing this abomination. Do you think that merging `FuseConvAsX` and `FuseConvAsY` is easily doable? If yes then it would be nice to do it, but even in the current situation the code is also much more readable than before

Thank you. It crossed my mind to check if it is doable but I think it could be a separate task and I'm sick now so I can't work.

jakpiase (approved):
LGTM

lidanqing-vv (approved):
LGTM

lidanqing-vv:
@baoachun Please review this. Thanks.

baoachun (approved):
LGTM

PR #40041, opened 2022-03-01
add check for backward hook

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

Raise exception when the place or dtype changed in hook

Example:
```
    import paddle
    v = paddle.rand([3, 3])
    v.stop_gradient = False

    def foo(grad):
        print('grad', grad, grad.dtype)  # grad's dtype is float32
        res = paddle.mm(grad, grad)  # mm runs in fp16
        print('res', res, res.dtype)  # res's dtype is float16
        return res

    v.register_hook(foo)
    with paddle.amp.auto_cast():
        a = paddle.mm(v, v)
        loss = a.sum()
        loss.backward()

# PermissionDeniedError: The dtype of tensor(auto_0_@GRAD) before(float) and after(::paddle::platform::float16) hook are not consistent
   [Hint: Expected pre->DataType() == post->DataType(), but received pre->DataType():5 != post->DataType():4.] (at /Paddle/Paddle/paddle/fluid/imperative/gradient_accumulator.h:191)
```

### DIFF
--- a/paddle/fluid/imperative/basic_engine.cc
+++ b/paddle/fluid/imperative/basic_engine.cc
@@ -317,6 +317,7 @@ static std::shared_ptr<NameVarMap<VariableWrapper>> CallGradientHooks(
         auto tmp_var = var;
         for (const auto& hook_pair : var->GetVariableWrapperHooks()) {
           tmp_var = (*hook_pair.second)(tmp_var);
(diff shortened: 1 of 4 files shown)


JiabinYang (approved):
CHECK this in eager plz

PR #40042, opened 2022-03-01
run recompute's real backward with amp disabled

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

run recompute's real backward with amp disabled, otherwise, it may result in error when calling hook in backward during amp enabled guard.

Related #39950, #40041


ForFishes (approved):
LGTM

PR #40058, opened 2022-03-02
[Eager][YAML] Supported array-type parsing for output tensors

jim19930609:
### PR types
New features

### PR changes
Others

### Describe
Before this patch, AutoCodeGen wasn't able to recognize array-typed output tensors in Yaml, for example "... -> Tensor[](Out)". This feature will be enabled after this patch.

### DIFF
--- a/paddle/fluid/eager/auto_code_generator/final_state_generator/eager_gen.py
+++ b/paddle/fluid/eager/auto_code_generator/final_state_generator/eager_gen.py
@@ -213,8 +213,12 @@ def ParseYamlReturns(string):
 
     returns = [x.strip() for x in string.strip().split(",")]
     for i in range(len(returns)):
-        ret = returns[i]
-        returns_list.append(["", ret, i])
+        ret_type = returns[i]
+
+        assert ret_type in yaml_types_mapping.keys()
+        ret_type = yaml_types_mapping[ret_type]
+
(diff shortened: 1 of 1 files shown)


paddle/fluid/eager/auto_code_generator/final_state_generator/eager_gen.py
```diff
@@ -213,8 +213,12 @@ def ParseYamlReturns(string):
 
     returns = [x.strip() for x in string.strip().split(",")]
     for i in range(len(returns)):
-        ret = returns[i]
-        returns_list.append(["", ret, i])
+        ret_type = returns[i]
```
zyfncg, line 216: If `ret_type` is a string like `Tensor(out)`，can it be found in `yaml_types_mapping.keys()`?

jim19930609: okay this is interesting. So this function "ParseYamlReturns" is designed for parsing "forward_api" only, and it assumes there's no "name" associated with the outputs, in other words "Tensor[], Tensor" format outputs. 

"backward_api", on the other hand, will use "ParseYamlReturnsWithName" instead, which assumes "Tensor(out), Tensor[](out2)" format outputs.

zyfncg (approved):

PR #40078, opened 2022-03-02
[Phi] Change input vec tensor to pointer type

chenwhql:
### PR types

Function optimization
### PR changes

Others
### Describe

[Phi] Change input vec tensor to pointer type

将kernel的vector类型由 `const std::vector<DenseTensor>&` 改为 `const std::vector<const DenseTensor*>&`


MingMingShangTian (approved):
LGTM

zyfncg (approved):

PR #40079, opened 2022-03-02
[Phi]Fix var in output may be nullptr bug

MingMingShangTian:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix var in output may be nullptr , and will cause segment fault bug.


chenwhql (approved):
LGTM

PR #36541, opened 2021-10-19
Add an elementwise + activation fusion pass.

tsocha:
### PR types
Performance optimization

### PR changes
Others

### Describe
Add a pass which fuse element wise operation with an activation:
- "elementwise_add", "elementwise_sub", "elementwise_mul"
- "relu", "tanh", "leaky_relu", "swish", "hardswish", "sqrt", "abs", "clip", "gelu", "relu6", "sigmoid"
- required attributes are used in pass without registration in makers.

PR #40081, opened 2022-03-02
[Candidate solution] Add an elementwise + activation fusion pass, add mechanizm to add attributes outside of base op.

tsocha:
### PR types
Function optimization

### PR changes
Function optimization

### Describe
Add a pass which fuse elementwise operation with activations and add new attributes in the pass during runtime
- "elementwise_add", "elementwise_sub", "elementwise_mul"
- "relu", "tanh", "leaky_relu", "swish", "hardswish", "sqrt", "abs", "clip", "gelu", "relu6", "sigmoid"
This is another candidate solution of PR https://github.com/PaddlePaddle/Paddle/pull/36541, according to Achun and Hongyu's suggestion.


paddle/fluid/operators/elementwise/mkldnn/elementwise_mkldnn_op.h
```diff
@@ -32,6 +32,39 @@ using dnnl::stream;
...
+          algo_map.find(ctx.Attr<std::string>("activation_type"));
+
+      if (activation_type != algo_map.end()) {
+        post_operations.append_eltwise(scale, activation_type->second, alpha,
+                                       beta);
+      }
+    }
+    return post_operations;
+  }
+
```
lidanqing-vv, line 73: Hi, I only have one question, why these getting attributes and computation function are in `xx.h` file? It will not increase the lib size? Thanks !

tsocha: There is no .cc file and I didn't create it.

jakpiase (approved):
LGTM

tsocha:
This solution has been selected.

PR #40118, opened 2022-03-03
fix paddle.median torch diff

ronny1996:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix paddle.median torch diff


qili93 (approved):
LGTM

PR #40121, opened 2022-03-03
[Auto parallel] Redesign the tuner for auto parallel

aoyulong:
### PR types
New features

### PR changes
Others

### Describe

This pr redesigns the tuner and abstract the common modules to help users to implement the customized tuners. Based on these modules, the planner of our auto parallel can implement the automatic searcher or tuner for parallelization and optimization strategies.

- `TunableSpace` is a search space constructed by different `TunableVariable`.
- `Trail` is an instance of one tunable space and contains the current values for each tunable variable.
- `Tuner` is the base class which do the search process in the tunable space and evaluate the different trails.


JZ-LIANG (approved):
LGTM

PR #40153, opened 2022-03-04
[ROCm] fix hip test to update LD_LIBRARY_PATH, test=develop

qili93:
### PR types

Bug fixes

### PR changes

Others

### Describe

CPP unit test cases in HIP will fail due to "error while loading shared libraries: libiomp5.so: cannot open shared object file: No such file or directory". Fix hip_test to to add libraries into LD_LIBRARY_PATH.


Aganlengzi (approved):
LGTM

PR #40166, opened 2022-03-04
Add exception throw for norm_conv when platform is not supported

ZzSean:
### PR types

Others 
### PR changes

Others
### Describe

Add exception throw for norm_conv when platform is not supported


Xreki (approved):
LGTM

PR #40207, opened 2022-03-07
[Auto Parallel] Add distributed mul op for paddle.fluid.layers.fc

aoyulong:
### PR types

Bug fixes

### PR changes

Others

### Describe

This pr add a distributed mul op to make auto parallel to be compatible with the models which use `paddle.fluid.layers.fc`.


JZ-LIANG (approved):
LGTM

PR #40229, opened 2022-03-07
refine op_test and add support for concat op and variadic tensor list.

2742195759:
### PR types

Others
### PR changes

Others
### Describe

refine op_test and add support for concat op and variadic tensor list.


Aurelius84 (approved):
LGTM

PR #40232, opened 2022-03-07
fix the full_like with fill the value of inf

wawltor:
### PR types
Bug fixes 

### PR changes
OPs

### Describe
fix the full_like with fill the value of inf

### DIFF
--- a/paddle/phi/kernels/cpu/full_kernel.cc
+++ b/paddle/phi/kernels/cpu/full_kernel.cc
@@ -54,7 +54,7 @@ void FullLikeKernel(const Context& dev_ctx,
 
   auto common_type_value = static_cast<CommonType>(value);
 
-  // Judge the fill value is valid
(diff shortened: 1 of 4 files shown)


paddle/phi/kernels/cpu/full_kernel.cc
```diff
@@ -54,12 +54,22 @@ void FullLikeKernel(const Context& dev_ctx,
 
   auto common_type_value = static_cast<CommonType>(value);
 
-  PADDLE_ENFORCE_EQ(
-      (common_type_value >=
+  // Judge the fill value is valid
```
joey12300: Check whether the filled value is valid

wawltor: fixed it

joey12300 (approved):
LGTM

Issue #39473, opened 2022-02-11
AttributeError: module 'paddle' has no attribute 'uint16'


tangzhiyi11:
为使您的问题得到快速解决，在建立Issues前，请您先通过如下方式搜索是否有相似问题:【搜索issue关键字】【使用labels筛选】【官方文档】

如果您没有查询到相似问题，为快速解决您的提问，建立issue时请提供如下细节信息：
- 标题：简洁、精准概括您的问题
- 版本、环境信息：
    1）PaddlePaddle版本：develop分支

paddle c++接口返回 ::paddle::platform::bfloat16 这种类型，但是在python测返回的是 paddle.uint16，而 paddle.uint16 这种类型又报不存在，请问这里的显示的类型的转换是在哪里做的呢？

<img width="814" alt="image" src="https://user-images.githubusercontent.com/5955111/153549472-2ed87f40-a6b4-4027-ac36-045e8bf632ec.png">

iclementine:
python/paddle/fluid/dygraph/varbase_patch_methods.py
<img width="927" alt="图片" src="https://user-images.githubusercontent.com/16222986/153573207-adb13c51-7918-4189-b061-4ebdd32f3c5a.png">


里面有把 paddle::framework::proto::VarType::Type 打印出来的方法。里面把 bf16 对应到了 uint16. （至于这么做的原因是它一开始是用于和 Numpy 类型对应的，而 numpy 里没有 bfloat16）


而 paddle.float32 之类的是把 C++ 端的 paddle::framework::proto::VarType::FLOAT32 import 出来的 alias。参考 python/paddle/framework/dtype.py

tangzhiyi11:
> python/paddle/fluid/dygraph/varbase_patch_methods.py <img alt="图片" width="927" src="https://user-images.githubusercontent.com/16222986/153573207-adb13c51-7918-4189-b061-4ebdd32f3c5a.png">
> 
> 里面有把 paddle::framework::proto::VarType::Type 打印出来的方法。里面把 bf16 对应到了 uint16. （至于这么做的原因是它一开始是用于和 Numpy 类型对应的，而 numpy 里没有 bfloat16）
> 
> 而 paddle.float32 之类的是把 C++ 端的 paddle::framework::proto::VarType::FLOAT32 import 出来的 alias。参考 python/paddle/framework/dtype.py

多谢~

Aganlengzi:
[...]

Issue #34927, opened 2021-08-16
【paddle.cast】转换为 unit16 时的bug


wzzju:
关于bfloat16的转换问题，cast后的输出应该是1，但是这里的输出是16256；
环境 paddle 2.1.0，Mac
![image](https://user-images.githubusercontent.com/17102274/129527325-1178565c-9aeb-403a-98d4-caace3675707.png)

竞品都是1：
tf: 2.5.0 Tesla T4
![image](https://user-images.githubusercontent.com/17102274/129527372-6c424bce-5716-4f58-9558-7fac340cfc87.png)

pytorch: 1.8.1 Tesla T4
![image](https://user-images.githubusercontent.com/17102274/129527448-f510b876-d108-4883-a6f5-7f59104628b1.png)

============== 问题排查 ==============
BF16问题。见 python/paddle/fluid/framework.py 中 convert_np_dtype_to_dtype，由于numpy不支持bfloat16，代码中将np.uint16转为paddle.bfloat16
![image](https://user-images.githubusercontent.com/17102274/129527486-740157ae-d2e4-499a-867a-bf8edbc9e23a.png)

============== 解决方案 ===============

建议：可以参考tensorflow，add bfloat16 as numpy extensions, refer to：

tf bfloat16 numpy extension way： https://stackoverflow.com/questions/66354058/tensorflow-register-numpy-bfloat16-extension

tf bfloat16 numpy extension commit：   https://github.com/tensorflow/tensorflow/commit/24ffe9f729160a095a5cab8f5923920182803182

lidanqing-vv:
Temp comments

convert_bf16_to_fp32(a.cast(X, "bfloat16"))

str::cout<<a<<std::endl;

string()

lidanqing-vv:
This issue is pending by https://github.com/PaddlePaddle/Paddle/pull/35332

lidanqing-vv:
np.uint16 is used as bfloat16 because np.uint16 was reserved datatype, for nothing else, that is why np.uint16 is temporarily used to represent bfloat16.
With https://github.com/PaddlePaddle/Paddle/pull/35332/files merged, then in `python/paddle/fluid/framework.py`, we just need to `import paddle_bfloat`, and then do
```
 elif dtype == np.bfloat16:
        return core.VarDesc.VarType.BF16
 elseif dtype == np.uint16:
       print("Reserved datatype")
       return ...
```
[...]

PR #40323, opened 2022-03-08
[High priority][WIP] Added BF16 functionality inside paddle core

jakpiase:
### PR types
Others

### PR changes
Others

### Describe
Added BF16 functionality inside paddle core. After merging this PR uint16 won't be used as a placeholder for BF16 datatype anymore, so issues: #39473 and #34927 should be fixed


python/paddle/fluid/framework.py
```diff
@@ -992,7 +993,7 @@ def convert_np_dtype_to_dtype_(np_dtype):
         return core.VarDesc.VarType.INT64
     elif dtype == np.bool:
         return core.VarDesc.VarType.BOOL
-    elif dtype == np.uint16:
+    elif dtype == bfloat16:
         # since there is still no support for bfloat16 in NumPy,
         # uint16 is used for casting bfloat16
         return core.VarDesc.VarType.BF16
```
lidanqing-vv, line 1179: Hi, Do you think if it is necessary to map np.uint16 to core.VarDesc.VarType.UINT16?  Or maybe make it reserved datatype? Also please modify the comments ?
I think I saw more places that used uint16 in python than now, not in tests

jakpiase: Hi, this PR is not ready yet, I have made it just to see which tests and CIs will fail

lidanqing-vv: Hi, ok, thanks

lidanqing-vv:
Hi, @jakpiase there is file conflict, could you please rebase and solve the conflict? Thank you. 
After this PR is merged, please also cherry-pick it into release/2.3, this PR can catch up 2.3 formal release. Thank you very much for your work !

cmake/flags.cmake
```diff
@@ -133,7 +133,6 @@ if (NOT WIN32)
 set(COMMON_FLAGS
     -fPIC
     -fno-omit-frame-pointer
-    -Werror
```
jczaja: is this removing of -Werror intentional?

jakpiase: It's only temporary to silence paddle_bfloat compilation warnings, that won't be included into the final version of this PR

Aganlengzi:
@jakpiase hi, thanks for you work, you can attach https://github.com/PaddlePaddle/Paddle/issues/39473 and https://github.com/PaddlePaddle/Paddle/issues/34927 in Describe of this PR to synchronize states automatically.

jakpiase:
The work on that functionality was cancelled due to some issues on Windows-GPU evironment. Unfortunately I was not able to debug that and because of that I have aborted my work, but except for that case my current changes(after resurrecting this PR to the state of modern PaddlePaddle code) should enable numpy's bfloat16 datatype inside paddle core and enable passing that type of tensors both between python-c++ and c++-python.

PR #40414, opened 2022-03-10
[custom kernel] fix static object de-initialize bug

Aganlengzi:
### PR types
Bug fixes

### PR changes
Others

### Describe
1.Fix bug caused by static objects de-initialize order:
KernelFactory static instance is initialized BEFORE plugins are loaded for custom kernels, but de-initialized AFTER them are unloaded. We need manually clear symbols(may contain plugins' symbols) stored in this static instance to avoid illegal memory access.
2.Refine log info.


ronny1996 (dismissed):
LGTM

qili93 (approved):
LGTM

PR #40445, opened 2022-03-11
fix gpu callback

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix gpu callback that may lead to OOM

`std::async` with `std::launch::defered` is lazy-executed, and it will not be executed until `get` or `wait` method is called on the returen `future value`.

Ref: https://en.cppreference.com/w/cpp/thread/launch


jiweibo (dismissed):
LGTM

jiweibo (approved):

PR #40477, opened 2022-03-12
[hybrid fix] fix sharding save offload

wangxicoding:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix sharding `save_persistables` don't save `fp16 param` and `offload` var when use offload.


JZ-LIANG (approved):
LGTM

PR #40491, opened 2022-03-13
Improve the performance of fake quantize OP

leo0519:
### PR types

Performance optimization
### PR changes

OPs
### Describe

The modifications of fake quantize op are as follow.
1. (moving average) Move the computation of moving average scale to device (80us -> 3.5us)
2. (find max abs) Use register to save the local maximum in a thread


qingqing01 (approved):
LGTM

ceci3 (approved):
LGTM

PR #40500, opened 2022-03-14
[Phi] Move gather op kernel into phi

chenwhql:
### PR types

Function optimization
### PR changes

OPs
### Describe

[Phi] Move gather op kernel into phi


MingMingShangTian (approved):
LGTM

zyfncg (approved):

PR #40523, opened 2022-03-14
run python api in eager model and filter the out in argument list

2742195759:
### PR types

Bug fixes
### PR changes

Others

### Describe

run python api in eager model and filter the out in argument list

### DIFF
--- a/paddle/fluid/pybind/eager_utils.cc
+++ b/paddle/fluid/pybind/eager_utils.cc
@@ -34,6 +34,7 @@ limitations under the License. */
 namespace paddle {
 namespace pybind {
 
+using PyNameVarBaseMap = std::unordered_map<std::string, py::handle>;
 extern PyTypeObject* p_tensor_type;
 
 extern PyTypeObject* g_framework_scope_pytype;
--- a/paddle/fluid/pybind/eager_utils.cc
+++ b/paddle/fluid/pybind/eager_utils.cc
@@ -34,7 +34,6 @@ limitations under the License. */
 namespace paddle {
 namespace pybind {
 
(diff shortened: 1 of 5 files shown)


paddle/fluid/pybind/eager_utils.cc
```diff
@@ -34,6 +34,7 @@ limitations under the License. */
 namespace paddle {
 namespace pybind {
 
+using PyNameVarBaseMap = std::unordered_map<std::string, py::handle>;
```
JiabinYang: bad name, it seems not VarBaseMap

paddle/fluid/pybind/imperative.cc
```diff
@@ -2079,8 +2103,8 @@ void BindImperative(py::module *m_ptr) {
               const PyNameVarBaseMap &ins, const PyNameVarBaseMap &outs,
               framework::AttributeMap attrs) {
              // TODO(xiongkun): move this function outside of tracer.
-             auto ins_map = ConvertToNameVarBaseMap(ins);
-             auto outs_map = ConvertToNameVarBaseMap(outs);
+             auto ins_map = ConvertToNameTensorMap(ins);
```
JiabinYang, line 2106: what if its still VarBaseMap?

JiabinYang (commented):
some comments

paddle/fluid/pybind/eager_utils.cc
2742195759: This name is not used, will be removed.

chenwhql (approved):
LGTM for PADDLE_ENFORCE

PR #40577, opened 2022-03-15
[Ops] segment pool op support for int int64 kernel.

ZHUI:
### PR types

Others
### PR changes

OPs
### Describe

Segment pool op support for int int64 kernel.


TCChenlong (approved):
LGTM

PR #40609, opened 2022-03-16
fix double-free bug in variables of cinn subgraph

CtfGo:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix double-free bug in variables of cinn subgraph


thisjiang (approved):
LGTM

PR #40625, opened 2022-03-16
[Phi] Move grid sample op infershape into phi

chenwhql:
### PR types

Function optimization
### PR changes

OPs
### Describe

[Phi] Move grid sample op infershape into phi


MingMingShangTian (approved):
LGTM

Issue #39446, opened 2022-02-10
test_eigh_op, test_eigvalsh_op tolerance


zlsh80826:
- 标题：test_eigh_op, test_eigvalsh_op 單測問題 
- 版本、环境信息：
    1）PaddlePaddle版本：develop
    2）CPU/GPU：GPU, CUDA 11.6
    3）系统环境：Ubuntu20.04
    4）Python版本号: python3.8
- 复现信息:
```
ctest -R test_eigh_op
ctest -R test_eigvalsh_op
```
- 问题描述：请详细描述您的问题，同步贴出报错信息、日志/代码关键片段
1. 在 CUDA 11.6 後 cusolver 改進了 Ssyevd 的性能, 但同時和 numpy 的差異會略微放大 [test_eigh_op](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/test_eigh_op.py#L85-L86), [test_eigvalsh_op](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py#L81-L82) tolerance 需要修改至 2e-6 之上
2. Complex 單測需求 Hermitian matrix (對角元素 imag 的部分需要為 0), [參考文件](https://docs.nvidia.com/cuda/cusolver/index.html), [test_eigh_op]( https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/test_eigh_op.py#L91-92), [test_eigvalsh_op](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py#L87-L88)

Zjq9409:
正在修改复数随机数生成方式

Zjq9409:
eigh 算子修改PR链接为：https://github.com/PaddlePaddle/Paddle/pull/39568

zlsh80826:
Thank @Zjq9409. I am running the unit test on our environment to ensure the PR works well. 
Hi @huangjun12, `test_eigvalsh_op` has same the issue with `test_eigh_op`. Could you help to update the `test_eigval_sh_op.py`? Thanks

zlsh80826:
@Zjq9409 I have tested https://github.com/PaddlePaddle/Paddle/pull/39568, but it seems I had wrong conclusion in the past. The tolerance needs to be more higher. Following is the error message in our environment.
[...]

PR #40699, opened 2022-03-18
Enhance eigh, eigvalsh unit tests

zlsh80826:
### PR types

Others

### PR changes

Others

### Describe

This PR changes the validation mechanism to align cusolver and also fixes a batch test bug.

1. Validate `test_eigh_op` by [eigendecomposition](https://en.wikipedia.org/wiki/Eigendecomposition_of_a_matrix), which validates `|| A - QTQ' || / (||A|| * N)` and `|| I - QQ' || / M` to be less than the relative tolerance, where `A` is the input matrix, `Q` is  eigenvectors, `Q'` is the inverse of `Q`, `T` is the diagonal matrix whose diagonal elements are the corresponding eigenvalues.
2. Validate `test_eighvalsh_op` by comparing the Paddle results and numpy results and to ensure the relative tolerance is less than 5e-5. (Compare the norm vector)
3. The original `TestEighBatchAPI` is inherited from `TestEighAPI` and override [`init_input_shape`](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/tests/unittests/test_eigh_op.py#L178), but `TestEighAPI` didn't has `init_input_shape` so the original `TestEighBatchAPI` is duplicated with `TestEighAPI`.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
+++ b/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
@@ -41,7 +41,7 @@ def valid_eigenvalues(actual, expected):
 
     diff = np.abs(expected - actual)
     max_diff = np.max(diff)
-    max_ref = np.max(expected)
+    max_ref = np.max(np.abs(expected))
     relative_error = max_diff / max_ref
     np.testing.assert_array_less(relative_error, rtol)
 
--- a/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
+++ b/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
@@ -120,10 +120,6 @@ class TestEigvalshAPI(unittest.TestCase):
         self.complex_symm = np.divide(
             complex_data + np.conj(complex_data.transpose(self.trans_dims)), 2)
 
-    # def compare_result(self, actual_w, expected_w):
-    #     np.testing.assert_allclose(
-    #         actual_w, expected_w, rtol=self.rtol, atol=self.atol)
-
     def check_static_float_result(self):
         main_prog = paddle.static.Program()
         startup_prog = paddle.static.Program()
--- a/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
+++ b/python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
@@ -21,6 +21,31 @@ from op_test import OpTest
 from gradient_checker import grad_check
 
 
+def compare_result(actual, expected):
+    assert actual.ndim == 1 or actual.ndim == 2
+
+    if actual.ndim == 1:
+        valid_eigenvalues(actual, expected)
+        return
+
+    for batch_actual, batch_expected in zip(actual, expected):
+        valid_eigenvalues(batch_actual, batch_expected)
+
+
+def valid_eigenvalues(actual, expected):
+
+    FP32_MAX_RELATIVE_ERR = 5e-5
+    FP64_MAX_RELATIVE_ERR = 1e-14
+
+    rtol = FP32_MAX_RELATIVE_ERR if actual.dtype == np.single else FP64_MAX_RELATIVE_ERR
+
+    diff = np.abs(expected - actual)
+    max_diff = np.max(diff)
+    max_ref = np.max(expected)
+    relative_error = max_diff / max_ref
+    np.testing.assert_array_less(relative_error, rtol)
+
+
 class TestEigvalshOp(OpTest):
     def setUp(self):
         paddle.enable_static()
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/tests/unittests/test_gather_op.py
```diff
@@ -23,6 +23,21 @@
 from paddle.fluid.dygraph.base import switch_to_static_graph
 
 
+def check_eigh_results(A, eigh_vectors, eigh_values):
```
Zjq9409: Why do you need to modify the test_gather_op.py here? and why is this test method not placed in the test_eigh_op.py

zlsh80826: something went wrong. Fixed

jeng1220:
@Zjq9409 
I saw only PR-CI-Kunlun-KP-Build and PR-CI-Windows-Inference doesn't pass but this PR is not related to them.
Can you merge this PR now?

Zjq9409 (dismissed):
LGTM

huangjun12:
LGTM

huangjun12 (dismissed):

python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
```diff
@@ -21,6 +21,31 @@
...
+
+    FP32_MAX_RELATIVE_ERR = 5e-5
+    FP64_MAX_RELATIVE_ERR = 1e-14
+
+    rtol = FP32_MAX_RELATIVE_ERR if actual.dtype == np.single else FP64_MAX_RELATIVE_ERR
+
+    diff = np.abs(expected - actual)
+    max_diff = np.max(diff)
+    max_ref = np.max(expected)
+    relative_error = max_diff / max_ref
```
jeff41404, line 45: max_diff and max_ref may not be derived from the same element of eigenvalues, why not use np.allclose to check rtol and atol of all elements

zlsh80826: cusolver valid the eigenvalues by ensuring `|| d_ref - d_cusolver || / max|d_ref|` is less than relative error. The `norm` here is to get the max values of eigenvalues. All elements in `diff` are expected to be closed to 0 and also proportional to the input(output) scale, that's why we have to compare the `max(abs(expected))`.

So `max_ref = np.max(expected)` should be `np.max(np.abs(expected))`, let me update it.

zlsh80826: > proportional to the input(output) scale

It means the max residual becomes larger when the elements in input(output) get larger.

python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
```diff
@@ -103,9 +120,9 @@ def init_input_data(self):
         self.complex_symm = np.divide(
             complex_data + np.conj(complex_data.transpose(self.trans_dims)), 2)
 
-    def compare_result(self, actual_w, expected_w):
-        np.testing.assert_allclose(
-            actual_w, expected_w, rtol=self.rtol, atol=self.atol)
+    # def compare_result(self, actual_w, expected_w):
```
Zjq9409: 注释的代码可以删除

zlsh80826: DONE

python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
jeff41404: theoritically, np.allclose(expected, actual, rtol, atol) is more reasonable? 
if max_diff comes from eigenvalue e1 and max_ref from eigenvalue e2, I do not think it's reasonable

zlsh80826: As I said, the formula here is to implement `|| d_ref - d_cusolver || / max|d_ref| < tolerance` to align cusolver validation. If we want to choose a difference validation method, there is the risk to set a difference tolerance across different cuda version (https://github.com/PaddlePaddle/Paddle/issues/39446).

jeff41404: by reading issue, I understand backgroud, thank you.
but I have another question: does this increase in error mean that the calculation is more accurate or less?
this question do not block merge

zlsh80826: It is less accurate if we see the **third-party library as the ground truth**. But if we calculate the error through [eigendecomposition](https://en.wikipedia.org/wiki/Eigendecomposition_of_a_matrix), the error is in the same order and acceptable by cusolver team.

Zjq9409 (approved):
LGTM

python/paddle/fluid/tests/unittests/test_eigvalsh_op.py
jeff41404: ok, thank you

jeff41404 (approved):
LGTM

PR #40713, opened 2022-03-18
[new-exec] async prepare deps

zhiqiu:
### PR types

Performance optimization
### PR changes

Others
### Describe

Async prepare deps, since atomic is a little expensive and the time increases as the number of ops and variables in a program increase.

For example，for a program with about 110 operators, the time is 35us,
image

and can be reduced to 7.7us, which will not increase when the number of operators and variables increase.
image


chenwhql (approved):
LGTM  for scope.h

From00 (approved):
LGTM

PR #40714, opened 2022-03-18
fix lrn bug when export model

huangjun12:
### PR types
Bug fixes

### PR changes
APIs

### Describe
fix lrn bug when batch_size=-1 in static


chajchaj (approved):
LGTM

PR #40717, opened 2022-03-18
Correct MultipleQuantizeSquash

wozna:
### PR types

Bug fixes
### PR changes

Others
### Describe

This fixed the problem with running the yolov3_darknet model. 

This PR corrects MultipleQuantizeSquash in `cpu_quantize_squash_pass.cc`. 
It turned out that this mechanism didn't support the situation when under one input name are many inputs (eg. concat, sum). 
Before new inputs for the next operators were created, now we just updating the next operator input. 

Moreover, I added setting `mkldnn_data_type` to float32 if the operator is not quantised, because before even if the operator wasn't quantized this attribute was set to int8.

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
@@ -39,12 +39,12 @@ void UnlinkNodes(ir::Node* a, ir::Node* b) {
                   b->inputs.end());
 }
 
-void LogCannotQuantizeOp(Node* op, const char* details = nullptr) {
+void MarkAndLogCannotQuantizeOp(Node* op, const char* details = nullptr) {
   std::stringstream msg_ss;
   msg_ss << "Cannot quantize operator " << op->Name()
          << " (type: " << op->Op()->Type() << ", id: " << op->id() << ").";
   if (details) msg_ss << " " << details;
-  PrettyLogDetail(msg_ss.str().c_str());
+  VLOG(2) << msg_ss.str().c_str();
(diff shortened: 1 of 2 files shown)


lidanqing-vv:
@wozna Hi, Does all daily CI models could all pass? Or the coverage CI should test all the models?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -45,6 +45,7 @@ void LogCannotQuantizeOp(Node* op, const char* details = nullptr) {
          << " (type: " << op->Op()->Type() << ", id: " << op->id() << ").";
   if (details) msg_ss << " " << details;
   PrettyLogDetail(msg_ss.str().c_str());
+  op->Op()->SetAttr("mkldnn_data_type", std::string("float32"));
```
lidanqing-vv, line 48: Later on, could we refactor these log functions? It looks `VLOG(4)` is enough for these log functions?

lidanqing-vv: Does function below `void LogQuantizationDisabled(Node* op)`  maybe also need to set `"mkldnn_data_type", std::string("float32")` ?

sfraczek (approved):
LGTM!

lidanqing-vv (approved):
LGTM

lidanqing-vv:
Hi, @jczaja please review. This PR need one more approval to merge. Thanks

jczaja (approved):
LGTM

PR #40812, opened 2022-03-22
fix build_cinn_pass internal var may be control var problem

thisjiang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

In `BuildCinnPass`, an internal var may be a control var, that means its vardesc is empty. In previous code, we reported the error and existed when we check the vardesc failed, now we choice skip the vars because of cinn not support control var.


paddle/fluid/framework/paddle2cinn/build_cinn_pass.cc
```diff
@@ -300,10 +304,10 @@ std::unique_ptr<Graph> CreateNewSubGraph(const GraphNodeSet& cluster,
 
   GraphNodeMap old_var2new_var;
   for (auto* var : cluster_internals) {
-    PADDLE_ENFORCE_NOT_NULL(var->Var(),
-                            platform::errors::PreconditionNotMet(
-                                "The var desc of the node in cluster_internals "
-                                "shouldn't be null."));
+    if (!var->Var()) {
+      // skip control var
+      continue;
```
CtfGo: 添加TODO说明可能存在问题以及待改进；添加VLOG，便于debug

wzzju (approved):
LGTM

PR #40814, opened 2022-03-22
fix cinn graph may hasn't input problem

thisjiang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

The cinn-graph may hasn't input for CINN now support `fill_constant`, and it's all inputs may generated by fill_constant instead of by fetch.

So we remove the input check in InferShape, and rewrite the `GetExpectedKernelType` for the base function may error when the op hasn't input.

The new `GetExpectedKernelType` infer the kernel type from output, it's tensor has been initialized before (otherwise the default type is `DataType::UNDEFINED`).


CtfGo (approved):
LGTM

PR #40824, opened 2022-03-22
Polish optest: refine the optest parameter logic. support name, dtype, out, output in arbitrary position

2742195759:
### PR types

Others
### PR changes

Others
### Describe

Polish optest: refine the optest parameter logic. support name, dtype, out, output in arbitrary position


Aurelius84 (approved):
LGTM

PR #40856, opened 2022-03-23
upgrade predictor to support phi and models

Shixiaowei02:
### PR types

Others
### PR changes

Others
### Describe

upgrade predictor to support phi and models, test=develop


winter-wang (approved):
LGTM

PR #40859, opened 2022-03-23
[new-exec] skip var that is not tensor, selectedrows, tensorarray

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

skip var that is not tensor, selectedrows, tensorarray

fix the following problem
image


From00 (approved):
LGTM

PR #40865, opened 2022-03-23
[Auto Parallel] Update reshard

Caozhou1995:
### PR types

Others
### PR changes

Others
### Describe

This pr makes reshard module in the form of Class. 
In the past, the reshard module was composed of functions, which was not conducive to subsequent maintenance and expansion. 
Therefore, this time, the reshard module is divided into three main classes: Inserter, Remover and Resharder. 
Inserter is responsible for inserting the required OP and Var, Remover is responsible for removing the unnecessary OP and Var, and Resharder is responsible for the reshard process.


raindrops2sea (approved):

JZ-LIANG (approved):
LGTM

PR #40883, opened 2022-03-24
fix load bug and add distributed strategy from pslib

esythan:
### PR types
Others

### PR changes
Others 

### Describe
1、fix bug in load inference model (two stage)
2、add distributed strategy api to convert config from pslib


yaoxuefeng6 (approved):
LGTM

PR #40886, opened 2022-03-24
Make StreamSafeCUDAAllocator compatible with NaiveBestFit strategy

From00:
### PR types
Others

### PR changes
Others

### Describe

Since StreamSafeCUDAAllocator only supports allocator strategy of AutoGrowthBestFit, if set the _FLAGS_use_stream_safe_cuda_allocator=true_ for NaiveBestFit or ThreadLocal  strategy, an enforce error will be reported. 
This PR eliminates the error checking and make StreamSafeCUDAAllocator disabled implicitly in NaiveBestFit and ThreadLocal  strategies. The fast GC in new executor is also disabled for NaiveBestFit and ThreadLocal at the same time.


zhiqiu (dismissed):
LGTM

zhiqiu (approved):
LGTM

PR #40919, opened 2022-03-24
Fix Accuracy doc bug

LiuChiachi:
### PR types

Bug fixes
### PR changes

Docs
### Describe

Fix Accuracy doc bug, because `topk` shouldn't be int value.


TCChenlong (approved):
LGTM

PR #40940, opened 2022-03-25
Fix bug while specifying target grad in high order gradient

Aurelius84:
### PR types

 Bug fixes 
### PR changes

APIs
### Describe

Fix bug while specifying target grad in high order gradient

修复了在二阶以上微分时，`gradient()`接口传入自定义的初始化target_gradient时无法生效的问题


zhiqiu (approved):
LGTM

PR #40952, opened 2022-03-25
support gather when index is 0-zero shape

FlyingQianMM:
### PR types
Bug fixes | Performance optimization 

### PR changes

OPs
### Describe
- support gather when index is 0-zero shape
- remove synchronization in index_select and scatter_nd
- remove if cond in grad_clip to avoid device synchronization


paddle/phi/kernels/gpu/index_select_kernel.cu
```diff
@@ -82,7 +82,7 @@ void IndexSelectKernel(const Context& ctx,
         PADDLE_CUDA_NUM_THREADS,
         0,
         stream>>>(in_data, out_data, index_data, numel, stride, size, delta);
-    phi::backends::gpu::GpuStreamSync(stream);
+    //phi::backends::gpu::GpuStreamSync(stream);
```
ZzSean, line 85: 是不是可以直接删掉

PR #40966, opened 2022-03-25
[custom place] support tensor value operation

Aganlengzi:
### PR types
Others

### PR changes
Others

### Describe
1.Tensor _copy_to and set support CustomPlace type
2.phi kernel supports `std::vector<float>` parameter


Aganlengzi:
done by https://github.com/PaddlePaddle/Paddle/pull/41136

PR #41009, opened 2022-03-28
[new-exec] update the dependency of op that has inplace_back_map 

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

1. update the dependency of op that has inplace_back_map 
2. fix adam kernel to fit the case that betapow is cpu tensor in cuda kernel


From00 (approved):
LGTM

PR #41040, opened 2022-03-28
fix bug that some op has no op_role attr

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix bug that some op has no op_role attr


winter-wang (approved):
LGTM

PR #41078, opened 2022-03-29
[Op] Fix uncontrolled randomness of index_select op

haohongxiang:
### PR types

Bug fixes
### PR changes

OPs
### Describe

[Op] Fix uncontrolled randomness of index_select op


ForFishes (approved):
LGTM

PR #41084, opened 2022-03-29
Apply TransposeFolding & GemmRewriter passes when using CINN.

wzzju:
### PR types

Performance optimization
### PR changes

Others

### Describe

Apply TransposeFolding & GemmRewriter passes when using CINN.


thisjiang (approved):
LGTM

PR #41103, opened 2022-03-29
[AutoParallel] fix converter when sliced_shape is 1

zhaoyinglia:
### PR types

Bug fixes
### PR changes

Others
### Describe

- fix converter: the index is out of range when sliced_shape is 1.


JZ-LIANG (approved):
LGTM

PR #41105, opened 2022-03-29
add CUDA_TOOLKIT_ROOT_DIR option in cmake command

betterpig:
### PR types

Others
### PR changes

Others
### Describe

add CUDA_TOOLKIT_ROOT_DIR option in cmake command, make sure cuda and cudnn are consistent version.
CUDA_TOOLKIT_ROOT_DIR used in cmake must use / rather than \


zhwesky2010 (approved):
LGTM

PR #41106, opened 2022-03-29
[Auto Parallel] fix reshard bug due to the last update

Caozhou1995:
### PR types

Bug fixes
### PR changes

Others
### Describe

This pr fixes the reshard bug that var acquisition has not been updated.


JZ-LIANG (approved):
LGTM

PR #41130, opened 2022-03-30
Fix bug for UT test_calc_gradient

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

Scope records the mapping between variable names and variables, if **_scope_guard_** is not invoked, all variables and variable names are recorded in the default global scope, and the variables with the same name will overwrite each other. To avoid overwriting, the scope should be switched in test_calc_gradient.


zhiqiu (approved):
LGTM

PR #41200, opened 2022-03-31
[Dy2St]Use `self`as a parameter of _hash_with_id function to avoid error caused by hash_id reuse

0x45f:
### PR types

Others
### PR changes

Others
### Describe

Use `self`as a parameter of _hash_with_id function to avoid error caused by hash_id reuse


Aurelius84 (approved):
LGTM

PR #41262, opened 2022-04-01
make variable 'gradient_merge_cond' local

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

make variable 'gradient_merge_cond' local

The variable 'gradient_merge_cond' is created `global`, and it is set persistable in startup program but not persistable in main program, which may cause Error in executor.

startup program
38 persist trainable pars •0 K 6 LOO_ TENSOR  SMape (32, 128) etyseffloat32) step arodlenk (Fa 1se)

main program
Pasted Graphic 6

This PR make 'gradient_merge_cond' a local variable, since it is not needed to be initialized in startup program.


JZ-LIANG (approved):
LGTM

PR #41265, opened 2022-04-01
Add randperm and range yaml

zyfncg:
### PR types

Others

### PR changes

Others

### Describe

Add yaml config and unittest for randperm and range op.


MingMingShangTian (approved):
LGTM

PR #41290, opened 2022-04-01
Add UT for full_like after migration YAML

0x45f:
### PR types

Others
### PR changes

Others
### Describe

Add UT for full_like after migration YAML


Aurelius84 (approved):
LGTM

PR #41293, opened 2022-04-01
[ROCm] fix loss nan on ROCm 401, test=develop

qili93:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Loss Nan on ROCm 401 when running the following case:

```python
import paddle
import numpy as np

paddle.set_device("gpu")

if __name__ == "__main__":
  data = np.ones((1, 64, 128)).astype(np.float16)
  temp_v = paddle.to_tensor(data)
  print("temp_v = {}".format(temp_v))
  temp_y = paddle.sum(temp_v, axis=[-2, -1])
  print("temp_y = {}".format(temp_y))
```

### DIFF
--- a/paddle/phi/kernels/funcs/reduce_function.h
+++ b/paddle/phi/kernels/funcs/reduce_function.h
@@ -808,7 +808,7 @@ __global__ void ReduceHigherDimKernel(const Tx* x,
                                             1,
                                             1,
                                             left_num);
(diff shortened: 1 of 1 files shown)


qili93:
Close as duplicated with https://github.com/PaddlePaddle/Paddle/pull/41273

PR #41309, opened 2022-04-02
decrease the timeout of ut test_eager_dist_api from 300 to 100

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

decrease the timeout of ut test_eager_dist_api from 300 to 100


XieYunshen (approved):
LGTM

PR #41341, opened 2022-04-02
remove cudnn_deterministic=True

zhangjun:
### PR types

Others

### PR changes

Others

### Describe

remove remove cudnn_deterministic=True for inference


jiweibo (approved):
相关pr https://github.com/PaddlePaddle/Paddle/pull/37173

PR #41359, opened 2022-04-03
fix p_norm gpu nan bug while divide zero

zhiboniu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix p_norm gpu nan bug while divide zero


jeff41404 (approved):
LGTM

PR #41367, opened 2022-04-03
Add no need buffer

phlrain:
### PR types
Performance optimization

### PR changes
Others

### Describe

add no need buffer var suppport


JiabinYang (approved):
LGTM

PR #41374, opened 2022-04-03
[Eager] Fix empty tensor Initializer bug with shape=[]

Aurelius84:
### PR types

Bug fixes
### PR changes

Others
### Describe

+ [Eager] Fix empty tensor Initializer bug with shape=[].


JiabinYang (approved):
LGTM

PR #41383, opened 2022-04-03
fix index_select kernel configuration error where input numel is 0

FlyingQianMM:
### PR types
Bug fixes

### PR changes
 OPs

### Describe

For index_select, if the numel() of input x is 0, launching the kernel raises a error as bellow:
```
import paddle
data = paddle.empty((0, 7), dtype='float32')
index = paddle.to_tensor([0, 1], dtype='int32')
out = paddle.index_select(data, index, axis=-1)
```

image

After we fix this bug, the output can be computed correctly:

image


ZzSean (approved):
LGTM

PR #41403, opened 2022-04-04
Use densetensor instead of Tensor for ProcessGroup

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

Use densetensor instead of Tensor for ProcessGroup


FeixLiu (approved):
LGTM

PR #41423, opened 2022-04-05
[WIP] Reduce not needed HasAttr(), Attr() calls

jczaja:
### PR types
Performance optimization

### PR changes
OPs

### Describe
Reduce usage of HasAttr and Attr() calls in activation handler. Those functions are quite expensive and needed only for specific
activation algorithms.

### DIFF
--- a/paddle/fluid/platform/mkldnn_reuse.h
+++ b/paddle/fluid/platform/mkldnn_reuse.h
@@ -915,8 +915,8 @@ class ActivationMKLDNNHandler
       : platform::MKLDNNHandlerNoCachingT<T, dnnl::eltwise_forward,
                                           dnnl::eltwise_backward>(engine,
                                                                   cpu_place) {
-    float alpha = ctx.HasAttr("alpha") ? ctx.Attr<float>("alpha") : 0;
-    float beta = ctx.HasAttr("beta") ? ctx.Attr<float>("beta") : 0;
+    float alpha = 0.0;
+    float beta = 0.0;
 
     if (ctx.Type() == "scale") {
       bool bias_after_scale = ctx.Attr<bool>("bias_after_scale");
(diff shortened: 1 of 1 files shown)


paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -915,8 +915,8 @@ class ActivationMKLDNNHandler
       : platform::MKLDNNHandlerNoCachingT<T, dnnl::eltwise_forward,
                                           dnnl::eltwise_backward>(engine,
                                                                   cpu_place) {
-    float alpha = ctx.HasAttr("alpha") ? ctx.Attr<float>("alpha") : 0;
-    float beta = ctx.HasAttr("beta") ? ctx.Attr<float>("beta") : 0;
+    float alpha = 0.0;
```
jakpiase, line 918:
```suggestion
    float alpha = 0.0f;
```

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -915,8 +915,8 @@ class ActivationMKLDNNHandler
...
+    float beta = 0.0;
```
jakpiase, line 919:
```suggestion
    float beta = 0.0f;
```

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -938,11 +938,34 @@ class ActivationMKLDNNHandler
...
-      // paddle uses beta but mkldnn uses alpha for swish
-      if (algorithm == dnnl::algorithm::eltwise_swish) {
-        std::swap(alpha, beta);
-      } else if (algorithm == dnnl::algorithm::eltwise_bounded_relu) {
-        alpha = ctx.Attr<float>("threshold");
+      switch (algorithm) {
+        // alpha and beta does not exist for those
+        case dnnl::algorithm::eltwise_abs:
+        case dnnl::algorithm::eltwise_exp:
+        case dnnl::algorithm::eltwise_gelu:
```
jakpiase, line 945: Please add here `eltwise_round`, `eltwise_gelu_erf` and `eltwise_gelu_tanh`

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -938,11 +938,34 @@ class ActivationMKLDNNHandler
...
+          break;
+        case dnnl::algorithm::eltwise_swish:
+          // paddle uses beta but mkldnn uses alpha for swish
+          alpha = ctx.Attr<float>("beta");
+          break;
+        case dnnl::algorithm::eltwise_bounded_relu:
+          alpha = ctx.Attr<float>("threshold");
+          break;
+        default:
+          alpha = ctx.HasAttr("alpha") ? ctx.Attr<float>("alpha") : 0;
```
jakpiase, line 966: I have checked and after including 3 more eltwise algorithms which are mentioned above `default` branch of this switch-case won't ever be chosen, since all eltwises would be already covered(only `clip` and `linear` are using two arguments and are covered in lines 921-940).

jakpiase: So probably creating a guard - `PADDLE_THROW("Unsupported eltwise algorithm");` or something like that would be nice in this default branch

PR #41471, opened 2022-04-06
[new-exec] shrink downstream map

zhiqiu:
### PR types

Performance optimization
### PR changes

Others
### Describe

shrink downstream map


From00 (approved):
LGTM

PR #41481, opened 2022-04-06
[IPU] fix running error for ipu

gglin001:
### PR types

Others
### PR changes

Others
### Describe

add ipu backend for phi, or the UTs will fail


qili93 (approved):
LGTM

PR #41496, opened 2022-04-07
use group id to differentiate keys for tcp store

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

use group id to differentiate keys for tcp store


Baibaifan (approved):
LGTM

PR #41529, opened 2022-04-07
Fix multi-threads predict bug of RNN OP

joey12300:
### PR types

Bug fixes
### PR changes

OPs
### Describe

This PR fix the bug of RNN OP which modify the persistable variable `DropoutState` in inference mode.


wawltor (approved):
LGTM

PR #41577, opened 2022-04-09
Fix some issues on some ops

sneaxiy:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix bugs:
- `cumsum` uses default stream instead of the compute stream inside PaddlePaddle framework.
- register `int16_t` kernel for `size` op.
- fix `cross_entropy` API `ignore_index` issue. See https://github.com/PaddlePaddle/Paddle/pull/41541 .


limin2021 (approved):
LGTM.

PR #41596, opened 2022-04-10
Lml/lower2prim

levi131:
### PR types
New features

### PR changes
Others

### Describe
Add rules for lower original paddle ops into primitive ops and lower primitive ops into original paddle ops.


levi131:
代码在其他pr中打磨后合并

PR #41632, opened 2022-04-11
Increase the priority of CINN runtime_include_dir value setting by using env.

wzzju:
### PR types

Others
### PR changes

Others
### Describe

Increase the priority of CINN runtime_include_dir value setting by using env.


thisjiang (approved):
LGTM

PR #41676, opened 2022-04-12
[Eager]Fix clear_gradient bug in optimizer (#41658)

Aurelius84:
### PR types

Bug fixes 
### PR changes

APIs
### Describe

[Eager]Fix clear_gradient bug in optimizer (#41658)


JiabinYang (approved):
LGTM

PR #41701, opened 2022-04-12
Reduce trt convert unit test problem size

zlsh80826:
### PR types

Performance optimization
### PR changes

Others

### Describe

This PR is to reduce the trt convert unit tests running time through the following changes:

1. Reduce some inputs from H=W=64 to H=W=32.
2. Reduce too large max_input_shape, which affects trt engine build time.
3. Only test batch size = [1, 4] instead of [1, 2, 4].


JZZ-NOTE (approved):
LGTM

jiweibo (approved):

PR #41702, opened 2022-04-12
[CustomDevice] move member variable to dense_tensor.h

Aganlengzi:
### PR types
Others

### PR changes
Others

### Describe
Move member variable `format_` to `dense_tensor.h` to avoid runtime problem:
Currently, outer users can only use `dense_tensor.h` for plugins since `dense_tensor.inl`(to be deleted in near future) is not published. So, compiled Paddle and plugins probably have inconsistent DenseTensor class because `format_` and then lead to runtime problem.
04273eda296acd0b2177f6b59ae18a61


Shixiaowei02 (approved):

qili93 (approved):
LGTM

jim19930609 (approved):

PR #41805, opened 2022-04-14
[AutoParallel] dist p-norm op

zhaoyinglia:
### PR types

New features
### PR changes

Others
### Describe

add dist pnorm op: the 2-norm loss value is related to the number of input data. When the model is trained with dp strategy, the loss value is different in mathematics with serial strategy.


aoyulong (approved):
LGTM

PR #41821, opened 2022-04-14
[AutoParallel] dist reshape op

zhaoyinglia:
### PR types

New features
### PR changes

Others
### Describe

add dist reshape op(impl_idx=2) when input's dim length is equal to output's dim length.


JZ-LIANG (approved):
LGTM

PR #41824, opened 2022-04-14
fix a bug which will casue cuda address error when the input size is large

huangxu96:
### PR types

Bug fixes 
### PR changes

OPs
### Describe

This PR fixes a bug that will cause the Cuda address error. The reason for this bug is that the grid number of the Cuda Kernel had been wrongly set.


zhhsplendid (approved):
LGTM

PR #41878, opened 2022-04-16
Waste PR 

JamesLim-sy:
### PR types

Function optimization
### PR changes

OPs
### Describe

- Feature : Originally, the multiple input data was packed into a cpu std::vector and transfer to device memory by cudamemcpyAsyn api. However, this transfer can be with usage of structure type.


JamesLim-sy:
Because of failure in CLA license check, relocate to [PR46525](https://github.com/PaddlePaddle/Paddle/pull/46525)

PR #41947, opened 2022-04-18
[new-exec] clear the scope listener after run

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

clear the scope listener after run

For example, when loading parameters, there are several programs and thus several standalone_executors (one for each parameter), each standalone_executor contains a `variable scope` which will be set as a listener to `scope`.

Then, after `N` parameters loaded, the scope has `N` listener and has to sync `N` `variable scope` when create a variable, which cost much time.

image


2742195759 (dismissed):

From00 (approved):
LGTM

chenwhql (approved):

ZHUI (approved):
LGTM

PR #41985, opened 2022-04-19
Remove the .cpu() function call in roll API

0x45f:
### PR types

Others
### PR changes

Others
### Describe

Remove the .cpu() function call in roll API


Aurelius84 (approved):
LGTM

PR #42041, opened 2022-04-20
Downloading data for test_analyzer_vit_ocr

Silv3S:
### PR types
Bug fixes

### PR changes
APIs

### Describe
- VIT-OCR model and data file are now stored as single archive and they were moved to other server,
- This PR updates the configuration file.


Silv3S:
~~Hi @Aganlengzi, could you advise why `test_analyzer_vit_ocr`  in **PR-CI-Coverage** does not work? The logs indicate that the data file is missing, but it should be downloaded during building stage. This error occurs only in CI, locally run tests are passing.~~

I mistakenly assumed that tests API in CI are built from scratch. It turned out that `vit_ocr.tgz` file was already saved from the previous PR.  Therefore the test did not download any new files and the old directory structure made the paths to the files mismatched.

jczaja (approved):
LGTM

PR #42046, opened 2022-04-20
Support cinn_launch op in standalone executor

From00:
### PR types
New features

### PR changes
Others

### Describe

In standalone executor, we do not encourage directly using scope in OP kernel. However, some OPs do have such behavior (e.g., cinn_launch OP). To support to run cinn_launch OP in standalone executor, we specially pass a real scope for it in this PR.


zhiqiu (approved):
LGTM

PR #42098, opened 2022-04-21
[Eager] Fix CastPyArg2scalar for max value of int64

veyron95:
### PR types
Bug fixes

### PR changes
Others

### Describe
1. Fix CastPyArg2Scalar when input is the max value of int64.


JiabinYang (approved):
LGTM

PR #42128, opened 2022-04-22
fix recompute

sljlp:
### PR types

Bug fixes
### PR changes

APIs
### Describe

some time there occurs an error when using recompute in MoELayer
We fixed this bug in this PR


ForFishes (approved):
LGTM

PR #42149, opened 2022-04-23
Do not reset default stream for StreamSafeCUDAAllocator

From00:
### PR types
Bug fixes
### PR changes
Others

### Describe

The default stream in StreamSafeCUDAAllocator will be set when the CUDADeviceContext created.
```C++
CUDADeviceContext::CUDADeviceContext(CUDAPlace place) : phi::GPUContext(place) {
  phi::GPUContext::PartialInitWithoutAllocator();
  cuda_stream_.reset(new stream::CUDAStream(phi::GPUContext::stream(), place));
  auto& instance = memory::allocation::AllocatorFacade::Instance();
  instance.SetDefaultStream(place, phi::GPUContext::stream());
  workspace_.reset(new phi::DnnWorkspaceHandle(
      instance.GetAllocator(place).get(), stream()));
}
```

 Normally, the DeviceContextPool is a global singleton and **one Place only correspond to one DeviceContext**. However, to support multi-stream scheduling, standalone executor creates two extra DeviceContextPools for H2D and D2H stream in StreamAnalyzer, which make one Place correspond to multiple DeviceContext and unexpectedly reset the default stream in runtime. 
```C++
class StreamAnalyzer {
 public:
  explicit StreamAnalyzer(const platform::Place& place)
      : place_(place), d2h_ctx_pool_({place}), h2d_ctx_pool_({place}) {}
  
  ...

  platform::Place place_;
  platform::DeviceContextPool d2h_ctx_pool_;
  platform::DeviceContextPool h2d_ctx_pool_;
  std::map<size_t, std::shared_ptr<platform::DeviceEvent>> var_id2event_;
};
```

To avoid this behavior, this PR disables changing default stream after initially setting.

### DIFF
--- a/paddle/fluid/memory/allocation/allocator_facade.cc
+++ b/paddle/fluid/memory/allocation/allocator_facade.cc
@@ -415,6 +415,23 @@ class AllocatorFacadePrivate {
   void SetDefaultStream(const platform::CUDAPlace& place, gpuStream_t stream) {
     const std::shared_ptr<StreamSafeCUDAAllocator>& allocator =
         GetDefaultStreamSafeCUDAAllocator(place);
+
+    // NOTE(Ruibiao): The default stream will be set when the CUDADeviceContext
+    // created. Normally, the DeviceContextPool is a global singleton and one
+    // Place only correspond to one DeviceContext. However, to support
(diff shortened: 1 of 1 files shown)


zhiqiu (approved):
LGTM

PR #42205, opened 2022-04-24
Fix the race condition in cumsum operator

leo0519:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Followed by the issue https://github.com/PaddlePaddle/Paddle/issues/41932, the cumsum UT failed on P100.

The reason is that the `BlockReverse` function in `cumsum_op` has race condition.
This PR modifies part of codes to avoid hazard.


wawltor (approved):
LGTM

PR #42221, opened 2022-04-25
Adjust the max_relative_error of QR's grad checker

aoyulong:
### PR types
Bug fixes

### PR changes
OPs

### Describe

- Some of the unit tests of QR grad computation failed in CUDA 11.4 and can be fixed by changing the seed from 4 to 7. 
- After examining the diff values, we found the actual `max_relative_error` is between 1e-6 and 1e-7.  To avoid the failed tests caused by the random inputs, the `max_relative_error` is set to 1e-6 while keeping the `numeric_grad_delta` unchanged.


phlrain (approved):

juncaipeng (approved):
LGTM

PR #42292, opened 2022-04-26
[cherry-pick] range can not return shape when enable_static (#42275)

ShiningZhang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix bug: range can not return shape when enable_static


Aurelius84 (approved):
LGTM

PR #42377, opened 2022-04-28
fix lod_tensor_array gc

wangxicoding:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix lod_tensor_array gc.
Clear LodTensorArray in GC, otherwise LodTensorArray.size() is wrong, if size() decrease in next step, an error maybe occur.


FeixLiu (approved):
LGTM

PR #42385, opened 2022-04-28
【PaddlePaddle Hackathon 2】15 新增 API Nanmedian

thunder95:
### PR types
New features

### PR changes
APIs

### Describe
完成第二期第15项目开发任务: https://github.com/PaddlePaddle/Paddle/issues/40318
paddle.nanmedian 扩展了 paddle.median API 的功能，如果输入Tensor中有nan值， paddle.median 会将nan值考虑在内然后取中位数(有可能为nan值)，而 paddle.nanmedian 只考虑非nan值的中位数。

RFC设计文档: https://github.com/PaddlePaddle/community/pull/89
中文文档: https://github.com/PaddlePaddle/docs/pull/4820


TCChenlong:
根据提示，应该是API的示例代码错误，可以检查一下
image

thunder95:
> 根据提示，应该是API的示例代码错误，可以检查一下 image

@TCChenlong 多谢龙哥，已经修改好， 现在只剩approved error了

python/paddle/tensor/stat.py
```diff
@@ -253,6 +253,140 @@ def numel(x, name=None):
     return out
 
 
+def nanmedian(x, axis=None, ignore_nan=True, keepdim=True, name=None):
```
jeff41404: there is no parameter of `ignore_nan` in [nanmedian RFC](https://github.com/PaddlePaddle/community/pull/89). if we want to add this parameter, should change RFC first.

python/paddle/tensor/stat.py
```diff
@@ -253,6 +253,140 @@ def numel(x, name=None):
...
+
+    Args:
+        x (Tensor): The input Tensor, it's data type can be int32, int64, float16, float32, float64.
+        axis (None|int|list|tuple, optional):
+            The axis along which to perform median calculations ``axis`` should be int.
+            ``axis`` should be in range [-D, D), where D is the dimensions of ``x`` .
+            If ``axis`` is less than 0, it works the same way as :math:`axis + D`.
+            If ``axis`` is None, median is calculated over all elements of ``x``. Default is None.
+        ignore_nan (bool, optional): Whether to ignore nan values when median was calculated.
+            If `ignore_nan` is True, the calculation process is the same as `median` operator.
```
jeff41404: If `ignore_nan` is False, the calculation process is the same as `median` operator?

python/paddle/tensor/stat.py
thunder95: This RFC doc was updated  with a new PR:  https://github.com/PaddlePaddle/community/pull/124

python/paddle/tensor/stat.py
thunder95: modified.

python/paddle/tensor/stat.py
jeff41404: I discussed this case with some paddle experts today.
In Paddle v2 3(coming this month) C++ API will be officially introduced, a program (such as model Networking) can be composed by basic C++ APIs instead of Python APIs, which is very useful for some performance pursuit scenarios.
In order to reduce the threshold for users to use C++ API, it is necessary to ensure that the experience of using C++ API is the same with Python, that is, **the parameters of the two should be the same** (the different ones are being modified in plan), and the documents of Python API can be reused without writing another document about C++ API.
Therefore, in this case, the RFC doc does not need to be modified, and the parameter `ignore_nan` should be moved to a kernel function (e.g. BaseMedianLernel), and then both nanmediankernel and mediankernel call this function using different value of `ignore_nan`. So the parameters of Python API, C++ API and kernel are the same, and the paddle Median and padding Nanmedian is used in the same way.

python/paddle/tensor/stat.py
```diff
@@ -253,6 +253,140 @@ def numel(x, name=None):
...
+        x = paddle.transpose(x, perm=trans_axis)
+
+    if len(x.shape) == 1 and axis == 0:
+        axis = None
+    else:
+        axis = len(x.shape) - 1
+    res_shape = [x for x in out_shape if x > 0]
+
+    if _in_legacy_dygraph():
+        medians, out = _C_ops.nanmedian(x, 'ignore_nan', ignore_nan)
```
jeff41404: According to the comments above, the parameters of Python API and C++ API should be the same, and the preprocessing logic in Python above should also be put into C++.

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,105 @@
...
+        col_vec.begin(), x_ptr + i * stride, x_ptr + (i + 1) * stride);
+
+    int64_t num_nan =
+        std::count_if(col_vec.begin(), col_vec.end(), [&](const T& val) {
+          return std::isnan(static_cast<float>(val));
+        });
+
+    int64_t pos = (stride - num_nan - 1) / 2;
+    std::nth_element(col_vec.begin(),
+                     col_vec.begin() + pos,
```
jeff41404: use col_vec.begin() + pos + 1 is better? no need to use std::nth_element again in if statement below?

paddle/phi/kernels/cpu/nanmedian_grad_kernel.cc
```diff
@@ -0,0 +1,69 @@
...
+
+      int64_t row = static_cast<int64_t>(i / stride);
+      int64_t m_row = 2 * row;
+      if (std::isnan(static_cast<float>(m_ptr[m_row])) ||
+          (x_ptr[i] != m_ptr[m_row] && x_ptr[i] != m_ptr[m_row + 1])) {
+        x_grad_ptr[i] = zero;
+        continue;
+      }
+
+      x_grad_ptr[i] = out_grad_ptr[row];
```
jeff41404: this logic to get the gradient of nanmedian is not correct, consider this data: x = [float('nan'), 2., **3.**, **2.**, 3., float('nan')], after sort, only `2` in offset 3(start from 0) of x and `3` in offset 2 of x are used to compute median, so only this two will have gradient. `2` in offset 1 of x and `3` in offset 4 of x **do not** have gradient.

paddle/phi/kernels/gpu/nanmedian_grad_kernel.cu
```diff
@@ -0,0 +1,84 @@
...
+  CUDA_KERNEL_LOOP(index, numel) {
+    int64_t row = static_cast<int64_t>(index / stride);
+    int64_t m_row = 2 * row;
+    if (isnan(static_cast<float>(x_ptr[index])) ||
+        isnan(static_cast<float>(medians_ptr[m_row])) ||
+        (x_ptr[index] != medians_ptr[m_row] &&
+         x_ptr[index] != medians_ptr[m_row + 1])) {
+      x_grad_ptr[index] = zero;
+    } else {
+      x_grad_ptr[index] = out_grad_ptr[row];
```
jeff41404: this logic to get the gradient of nanmedian is not correct, discribed above

jeff41404:
please add link of RFC(API design documents) in description

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,105 @@
...
+                       col_vec.begin() + pos + 1,
+                       col_vec.end(),
+                       [](const T& l, const T& r) {
+                         return (!std::isnan(static_cast<float>(l)) &&
+                                 std::isnan(static_cast<float>(r))) ||
+                                (l < r);
+                       });
+      m_ptr[2 * i + 1] = col_vec[pos + 1];
+    }
+    o_ptr[i] = static_cast<T>((m_ptr[2 * i] + m_ptr[2 * i + 1]) / 2.0);
```
jeff41404: remind, how to calculate median in even numbers should be clearly written in the document.

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: Learned from https://en.cppreference.com/w/cpp/algorithm/nth_element, the order of elements before the nth element seems to be somehow unreliable, so I fetched the n-1th element again. If the index of elements are needed, this std:nth_element would not be the good choice.

python/paddle/tensor/stat.py
thunder95: Your proposal is quite reasonable. I will move the logic into c++ implementation. I will adopt this structure, BaseMedianKernel(with parameters `axis`, `keepdim`),  and then calculate the presence of nan values,  and futher determine which kernel to pick (nanmediankernel and mediankernel). This param `ignore_nan` wont be used any more.  If I understood incorrectly, pls let me know, thanks!

paddle/phi/kernels/gpu/nanmedian_grad_kernel.cu
thunder95: I will use the index of sorted elements, and determine which one should compute the gradient. A question here, before I designed this kernel , the sorting algorithm did not work in case of nan values. Since last sharing of nanquantile in PFCC,  according to my knowledge,  this sort api is already fixed by someone. So that means, I can feel free to use sort api with no need to replace nan value with minimum. am I correct?

python/paddle/tensor/stat.py
thunder95: A question: can c++ api support the python data type like tuple and None?

python/paddle/tensor/stat.py
thunder95: done

python/paddle/tensor/stat.py
thunder95: done

paddle/phi/kernels/cpu/nanmedian_grad_kernel.cc
thunder95: done

paddle/phi/kernels/gpu/nanmedian_grad_kernel.cu
thunder95: done

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: done

python/paddle/tensor/stat.py
thunder95: > please add link of RFC(API design documents) in description

already added.

paddle/fluid/operators/nanmedian_op.cc
```diff
@@ -0,0 +1,121 @@
...
+        .AsIntermediate()
+        .AsExtra();
+    AddOutput("Out",
+              "(Tensor),"
+              " the output of  NanmedianOp, whose dtype is the same as X");
+    AddAttr<bool>("keepdim",
+                  "(bool, default true) "
+                  "If true, retain the reduced axis with length 1.")
+        .SetDefault(true);
+    AddAttr<std::vector<int>>("axes",
```
jeff41404: as the parameter is `axis` in Python API, better to use the same name?

python/paddle/fluid/tests/unittests/test_nanmedian.py
```diff
@@ -0,0 +1,197 @@
...
+            pd_res = paddle.nanmedian(
+                paddle.to_tensor(data), axis=axis, keepdim=False)
+            axis = clean_axis_numpy(axis, len(data.shape))
+            np_res = np.nanmedian(data, axis=axis, keepdims=False)
+            self.assertTrue(np.allclose(np_res, pd_res.numpy(), equal_nan=True))
+
+        for name, data in self.fake_data.items():
+            test_data_case(data)
+            if "_normal" in name:
+                test_data_case(data)
```
jeff41404: why test single_axis_normal and multi_axis_normal twice, or want to test_axis_case?

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,207 @@
...
+                      DenseTensor* median_index,
+                      bool ignore_nan) {
+  DenseTensor x;
+  auto rank = input.dims().size();
+  if (axes.size() == 0) {
+    x = input;
+    x.Resize({input.numel()});
+  } else if (rank > 1) {
+    PreprocessMedianKernel<T, Context>(dev_ctx, input, axes, &x);
+  }
```
jeff41404: use 
`if ((axes.size() == 0) || rank <= 1){
 ...
} else {
PreprocessMedianKernel...}` 
is easier to understand?

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,207 @@
...
+  if (should_ignore_nan) {
+    for (i = 0; i < pre_dim; i++) {
+      if (nan_counts[i] == stride) {
+        m_ptr[i * 2] = -1;
+        m_ptr[i * 2 + 1] = -1;
+        o_ptr[i] = sort_out_ptr[pos];
+      } else {
+        int64_t nan_k = nan_counts[i] > 0
+                            ? static_cast<int64_t>(stride - nan_counts[i])
+                            : sort_k;
```
jeff41404, line 62: if `total_nan_num==0` in function `ProcessMedianKernel`, then `sort_k` will be ((stride >> 1) + 1) in L154, so here `nan_k = sort_k` is already at the position of median value because without any nan value, but in next line, `row_pos = static_cast<int64_t>(nan_k >> 1)`  is equivalent stride/4, am I right?

paddle/fluid/operators/nanmedian_op.cc
thunder95: yes, its a better way, I will do it.

python/paddle/fluid/tests/unittests/test_nanmedian.py
thunder95: That was my typo, I should remove it.

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: yes, thanks for this nice suggestion.

paddle/phi/kernels/cpu/nanmedian_grad_kernel.cc
```diff
@@ -0,0 +1,94 @@
...
+  int64_t rank = x_dim.size();
+  int64_t stride = x_dim[rank - 1];
+
+  int64_t pre_dim = numel / stride;
+  int64_t i = 0;
+  int64_t offset = 0;
+  for (i = 0; i < pre_dim; i++) {
+    if (m_ptr[2 * i] >= 0) {
+      x_grad_ptr[offset + m_ptr[2 * i]] = out_grad_ptr[i];
+      x_grad_ptr[offset + m_ptr[2 * i + 1]] = out_grad_ptr[i];
```
jeff41404: Here, we need to consider whether the median comes from 1 number(`stride - nan_counts` is odd), or 2 numbers(`stride - nan_counts` is even), If it's the former, this is right; if the latter, according to the law of derivation, each will be out_grad_ptr[i]/2

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,207 @@
...
+          m_ptr[2 * i] = row_pos > 0 ? sort_indices_ptr[off_set - 1]
+                                     : sort_indices_ptr[off_set];
+          m_ptr[2 * i + 1] = sort_indices_ptr[off_set];
+          T m_val_left =
+              row_pos > 0 ? sort_out_ptr[off_set - 1] : sort_out_ptr[off_set];
+          T m_val_right = sort_out_ptr[off_set];
+          o_ptr[i] = (m_val_left + m_val_right) / div_factor;
+        }
+      }
+      pos += sort_k;
```
jeff41404, line 79: `pos` is start position of each stride, should be `pos += stride`?

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,207 @@
...
+      if (nan_counts[i] == stride) {
+        m_ptr[i * 2] = -1;
+        m_ptr[i * 2 + 1] = -1;
+        o_ptr[i] = sort_out_ptr[pos];
+      } else {
+        int64_t nan_k = nan_counts[i] > 0
+                            ? static_cast<int64_t>(stride - nan_counts[i])
+                            : sort_k;
+        int64_t row_pos = static_cast<int64_t>(nan_k >> 1);
+        int64_t off_set = pos + row_pos;
```
jeff41404: offset is a single word, no need use underline

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,207 @@
...
+          o_ptr[i] = (m_val_left + m_val_right) / div_factor;
+        }
+      }
+      pos += sort_k;
+    }
+  } else {
+    pos = -1;
+    if (is_ori_odd) {
+      for (i = 0; i < pre_dim; i++) {
+        pos += sort_k;
```
jeff41404, line 85: `pos` here means offset of median value, need another variable to store start position of each stride in loop, this two variables can not be mixed.

paddle/phi/kernels/cpu/nanmedian_kernel.cc
```diff
@@ -0,0 +1,207 @@
...
+        o_ptr[i] = sort_out_ptr[pos];
+        m_ptr[2 * i] = sort_indices_ptr[pos];
+        m_ptr[2 * i + 1] = sort_indices_ptr[pos];
+      }
+    } else {
+      for (i = 0; i < pre_dim; i++) {
+        pos += sort_k;
```
jeff41404, line 92: `pos` here means offset of median value, need another variable to store start position of each stride in loop, this two variables can not be mixed.

paddle/phi/kernels/cpu/nanmedian_grad_kernel.cc
thunder95: done

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: `sort_k` will be the one what I think. After sorting, the dimention also changed with a new width `sort_k`, not `stride` any more. so `pos` indicates also the start of new dimention.

To better clarify my thoughts, I may raise two examples as belows, 
case 1: [[1, 2, 3], [4, 5, 6]], `stride=sort_k=3`, which goes into L82, and find the middle position directly.
case 2: [[1, 2, NAN], [4, NAN, NAN]], stride=3. And after sorting sort_k is equivalent to 2, new matrix would look so, [[1, 2], [4, NAN]]. All the following position calculation will go to L54 based on row dimention `sort_k=2`.

I hope this explaination can make sense with my limited english expression.

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: will correct in next submit.

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: exactly,  I wiil fix it.

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: exactly, I wiil fix it.

paddle/phi/kernels/nanmedian_grad_kernel.h
```diff
@@ -0,0 +1,73 @@
...
+
+  for (int64_t i = 0; i < rank; i++) {
+    if (std::find(axes.begin(), axes.end(), i) != axes.end()) {
+      trans_back[i] = offset;
+      reshape_back.push_back(input_dim[i]);
+      offset += 1;
+    }
+  }
+
+  input->Resize(make_ddim(reshape_back));
```
jeff41404, line 60: next line will transpose `input` to get `x`, why need input->Resize here ?
and I can't understand the meaning of `reshape_back`, shall we really need `reshape_back`? especially we have x->dims()

paddle/phi/kernels/gpu/nanmedian_grad_kernel.cu
```diff
@@ -0,0 +1,114 @@
...
+                                    const int64_t* medians_ptr,
+                                    const T* out_grad_ptr,
+                                    T* x_grad_ptr,
+                                    int64_t stride,
+                                    int64_t pre_dim) {
+  CUDA_KERNEL_LOOP(index, pre_dim) {
+    int64_t offset = index * stride;
+    if (medians_ptr[2 * index] >= 0) {
+      x_grad_ptr[offset + medians_ptr[2 * index]] = out_grad_ptr[index];
+      x_grad_ptr[offset + medians_ptr[2 * index + 1]] = out_grad_ptr[index];
```
jeff41404: Here, we need to consider whether the median comes from 1 number(stride - nan_counts is odd), or 2 numbers(stride - nan_counts is even), If it's the former, this is right; if the latter, according to the law of derivation, each will be out_grad_ptr[i]/2

paddle/phi/kernels/gpu/nanmedian_kernel.cu
```diff
@@ -0,0 +1,289 @@
...
+                      DenseTensor* out,
+                      DenseTensor* median_index) {
+  DenseTensor x;
+  auto rank = input.dims().size();
+  if (axes.size() == 0) {
+    x = input;
+    x.Resize({input.numel()});
+  } else if (rank > 1) {
+    PreprocessMedianKernel<T, Context>(dev_ctx, input, axes, &x);
+  }
```
jeff41404: use
if ((axes.size() == 0) || rank <= 1){ ... } else { PreprocessMedianKernel...}
is easier to understand?

paddle/phi/kernels/gpu/nanmedian_grad_kernel.cu
jeff41404: > Since last sharing of nanquantile in PFCC, according to my knowledge, this sort api is already fixed by someone. So that means, I can feel free to use sort api with no need to replace nan value with minimum. am I correct?

yes, sort api is updated recently, if have any use problem, you can issue or even start another pr to fix it:)

paddle/phi/kernels/nanmedian_grad_kernel.h
thunder95: Take the input with `dimention`: [2, 3, 4, 5] and target `axis` [1, 3] for example:
so the dimention of `original x_grad` should be [2, 3, 4, 5]
after transpose in the forward phase, the original input will get tranposed into [2, 4, 3, 5], and reshape into [2, 4, 15], 
so the output of `median index` would be in oder of [2, 4, 15], then I split all the elements in `reshap_back` [2, 4, 3, 5],  finally I transpose back to the origin x_grad [2, 3, 4, 5] with `perm` [0, 2, 1, 3].
So after that, the `axis` 2 is not the closest neightbor to the axis `3` any longer.
That was my consideration w.r.t this logic. I am also not quite sure weather the Resize is notwendig.
If I am wrong, pls tell me,  and also will test it what if I remove this `Resize`.

thunder95: I have tested, if I remove this, the result differs what I expect. So we possibly can keep it there firstly?

paddle/phi/kernels/gpu/nanmedian_grad_kernel.cu
thunder95: done

paddle/phi/kernels/gpu/nanmedian_kernel.cu
thunder95: done

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: done

paddle/phi/kernels/cpu/nanmedian_grad_kernel.cc
thunder95: done

paddle/phi/kernels/cpu/nanmedian_kernel.cc
thunder95: done

python/paddle/fluid/tests/unittests/test_nanmedian.py
thunder95: done

paddle/fluid/operators/nanmedian_op.cc
thunder95: done

thunder95:
@jeff41404 updated, please review again.

thunder95 (commented):
the previous comments are in pending status. sry~~~~

paddle/phi/kernels/nanmedian_grad_kernel.h
jeff41404: thank you, I understand

jeff41404 (dismissed):
LGTM

TCChenlong:
请补充中文文档，并将链接附在 Describe 中

thunder95:
> 请补充中文文档，并将链接附在 Describe 中

@TCChenlong 完成

python/paddle/tensor/stat.py
```diff
@@ -241,6 +241,103 @@ def numel(x, name=None):
     return out
 
 
+def nanmedian(x, axis=None, keepdim=True, name=None):
+    """
+    Compute the median along the specified axis, while ignoring NaNs.
```
Ligoml, line 246: 在文档前加`r` ，现在无法正常解析英文文档

thunder95: @Ligoml 已加

python/paddle/__init__.py
```diff
@@ -331,6 +331,7 @@
 from .tensor.stat import var  # noqa: F401
 from .tensor.stat import numel  # noqa: F401
 from .tensor.stat import median  # noqa: F401
+from .tensor.stat import nanmedian  # noqa: F401
```
Ligoml, line 334: 这个api需要加在 `__all__` 中，否则不会被正常展示在api列表中

thunder95: @Ligoml  已修改

Ligoml (commented):
LGTM for docs

paddle/fluid/operators/nanmedian_op.cc
```diff
@@ -0,0 +1,121 @@
+/*Copyright (c) 2022 PaddlePaddle Authors. All Rights Reserved.
```
chenwhql: license换行缩进有点问题，建议参考别的改一下

paddle/phi/infermeta/unary.cc
```diff
@@ -1246,6 +1246,76 @@ void MultinomialInferMeta(const MetaTensor& x,
...
+      } else if (keep_dim) {
+        out_dim.push_back(1);
+      }
+    }
+  }
+
+  out->set_dims(make_ddim(out_dim));
+}
+
+void NanmedianGradInferMeta(const MetaTensor& x,
```
chenwhql: 反向的麻烦放到backward.h/cc中

paddle/phi/kernels/gpu/nanmedian_kernel.cu
```diff
@@ -0,0 +1,289 @@
...
+template <typename T, typename Context>
+void ProcessMedianKernel(const Context& dev_ctx,
+                         const DenseTensor& x,
+                         bool ignore_nan,
+                         DenseTensor* out,
+                         int64_t* m_ptr) {
+  bool should_ignore_nan = ignore_nan;
+  auto stream = dev_ctx.stream();
+  auto* ctx =
+      reinterpret_cast<const paddle::platform::CUDADeviceContext*>(&dev_ctx);
```
chenwhql: 这里好像不需要转换？直接使用dev_ctx

paddle/phi/kernels/nanmedian_kernel.h
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+  reshape.push_back(post_numel);
+
+  DDim trans_dim(input_dim);
+  int ndims = perm.size();
+  for (int i = 0; i < ndims; i++) {
+    trans_dim[i] = input_dim[perm[i]];
+  }
+  x->mutable_data<T>(trans_dim, dev_ctx.GetPlace());
```
chenwhql: 麻烦换成dev_ctx.template Alloc<T>()？

paddle/fluid/operators/nanmedian_op.cc
thunder95: 老师 您好，已修改

paddle/phi/infermeta/unary.cc
thunder95: 老师 您好，已修改

paddle/phi/kernels/gpu/nanmedian_kernel.cu
thunder95: 老师 您好，已修改

paddle/phi/kernels/nanmedian_kernel.h
thunder95: 老师 您好，已修改 @chenwhql

chenwhql (approved):
LGTM

zhwesky2010 (approved):
LGTM for change `parallel_UT_rule`

Ligoml (approved):

jeff41404 (approved):

XiaoguangHu01 (approved):
LGTM

PR #42396, opened 2022-04-29
modify reshape to reshape2 in paddle.nn.initializer.dirac

zhwesky2010:
### PR types

Function optimization
### PR changes

APIs
### Describe

modify reshape to reshape2 in paddle.nn.initializer.dirac.


jeff41404 (approved):
LGTM

PR #42437, opened 2022-05-02
Hotfix Release 2.3 Bug for CUDA 11.2

zhhsplendid:
### PR types

Bug fixes
### PR changes

Others
### Describe

This PR hotfixed the `test_cond.py` in CUDA 11.2

The reason of the bug is that the `fill_constant` op returns wrong value in the modified test case `test_extremely_simple_net_with_op_in_condition`, SWEs can use `layers.Print(a)` and `layers.Print(b)` in the test case to reproduce it and they can see the `fill_constant` returns something `e-50` instead of `1.23` and `1.25`

This PR hotfixed the bug by comparing `b` value instead of actual number, which makes sure the `cond` logic is right. **However, the PR didn't fix `fill_constant`**. We would let the SWEs who are working here to find the op bug and fix it.


CtfGo (approved):
LGTM

PR #42506, opened 2022-05-05
oneDNN NHWC matmul & elementwise kernels fixes

jczaja:
### PR types
Bug fixes

### PR changes
 OPs

### Describe
Those are fixes to NHWC support in oneDNN kerenels (matmul and elementwise).  Without them upgrade to oneDNN 2.6 is not possible.

### DIFF
--- a/paddle/fluid/operators/elementwise/elementwise_op.h
+++ b/paddle/fluid/operators/elementwise/elementwise_op.h
@@ -103,11 +103,12 @@ class ElementwiseOp : public framework::OperatorWithKernel {
       std::vector<int> out_dims_array(max_dim);
 #ifdef PADDLE_WITH_MKLDNN
       // (jczaja): Broadcasting of dims has to be done on Paddle shapes (NHWC)
-      // if model is using NHWC.
+      // if model is using NHWC and any of shapes in at least 3D
       bool should_rotate =
           ctx->IsRunMKLDNNKernel() &&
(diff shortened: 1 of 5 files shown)


jczaja:
@piotrekobi Please help with review

wawltor:
Hi，please fix the test case of test_matmul_v2_mkldnn_op,  this test case execute in the gpu, not the cpu.  The code added that the test case will execute in the cpu.
image

piotrekobi (commented):
@jczaja Can you please explain why:
1. The rotation should only happen when the dims size is >= 3?
2. The rotation when matmul is the first op in the chain is only required for matmul, and not elementwise or other ops?
Also, some code is repeated in matmul and matmul_v2 kernels and if it ever needs to be added to more kernels it might be a good idea to move it into a common function.

jczaja:
@piotrekobi 
1.  Subject to rotation are formats like NWC -> NCW , NHWC -> NCHW , NDHWC -> NCDHW . For NC or N there is nothing rotate and if roating code is called there will be some undefined behaviour
2. It is required for all our ops. Currently many of them are having it but not all . I'm adding this code to new ops as I got to work on NHWC aspect of it. So it will be done.

piotrekobi (dismissed):
LGTM

lidanqing-vv (approved):
LGTM

Issue #40013, opened 2022-03-01
error: stray ‘@’ in programEagerTensor.


mingxu1067:
**System information**
-PaddlePaddle version: 4fbcf6f4c52adccbc6ea0786b302485f14e5a951
-CPU: including CPUMKL/OpenBlas/MKLDNN version
-GPU: including CUDA 11.6, cuDNN 8.2
-OS Platform and Distribution: Ubuntu 20.04
-Python version: 3.8
**Describe your current behavior**
When we add `@` as part of name in Ops, like
```python
class MyOpMaker  : public framework::OpProtoAndCheckerMaker {
public:
  void Make() override {
    AddInput("Out@Grad").
  }
}
```
, it would encounter into a compile error.
**Other info / logs**
```
paddle/fluid/eager/api/generated/fluid_generated/dygraph_forward_api.h:146:87: error: stray ‘@’ in programEagerTensor my_op_dygraph_function(const egr::EagerTensor& Out@GRAD,const egr::EagerTensor& X,const egr::EagerTensor& Y, const paddle::framework::AttributeMap& attr_map);
```

mingxu1067:
@sneaxiy

weisy11:
「@」is a special key word in paddle, please don't use it if you don't know the usage.

mingxu1067:
The background is that we are helping Paddle to integrate cuBlasLt's epilogue features (see this [PR](https://github.com/PaddlePaddle/Paddle/pull/39437) for details.)
For some reasons, the fwd and bwd Ops do not contain the same computing scope. Therefore, we made bwd op also as a normal op (it is with OpMaker). In this case, we would like to apply `framework::GradVarName` to names of input/output tensor in grad op, which would let naming style be consistent. But it seems not be allowed to have `@` in names of input/output tensor when Op is constructed by Maker. 
@sneaxiy  might have some comments.

sneaxiy:
@JiabinYang Please have a look.

JiabinYang:
Sorry this problem is caused by some auto code gene issue, and we are trying to use `@` to split forward var and grad var. However, this can be changed by 3.31

mingxu1067:
@JiabinYang Thanks for reply. When would 3.31 be released (merged into develop branch)?

mingxu1067:
Hi @JiabinYang, Would you kindly share any updates about this issue?

JiabinYang:
[...]

PR #42509, opened 2022-05-05
Fix stray error

JiabinYang:
### PR types

Others
### PR changes

Others
### Describe

This PR support using `@` as input or output name to fix issue #40013
Once OpMaker is written like this:
image
We will rename it in code-gene module to:
image


zhiqiu (approved):
LGTM for op_function_generator.cc

PR #42531, opened 2022-05-06
fix deformable_psroi_pooling bug where too many resources requested for launch

FlyingQianMM:
### PR types
Bug fixes

### PR changes
OPs

### Describe

In CUDA 11.2, deformable_psroi_pooling raises a bug as bellow:
```
Program hit cudaErrorLaunchOutOfResources (error 701) due to "too many resources requested for launch" on CUDA API call to cudaLaunchKernel.
```
we fix this bug by:
* Reduce the number of threads per block 
* move common local variable before loop


wawltor (approved):
LGTM

PR #42603, opened 2022-05-09
[EinsumOp] Polish forward logic and backward logic for optimize

2742195759:
### PR types

Others
### PR changes

Others
### Describe

1. Polish forward logic and backward logic for optimize
2. add unittest for static graph shape.


Aurelius84 (approved):
LGTM

PR #42623, opened 2022-05-09
Add cinn pass to program

sneaxiy:
### PR types
New features

### PR changes
Others

### Describe
Support program pass to invoke the CINN compiler.

### DIFF
--- a/paddle/fluid/framework/paddle2cinn/build_cinn_pass.cc
+++ b/paddle/fluid/framework/paddle2cinn/build_cinn_pass.cc
@@ -545,6 +545,15 @@ void ReplaceSubGraphWithCinnOpNode(
   RemoveSubGraphFromGraph(cluster, cluster_internals, graph);
 }
 
+static bool IsInplaceOp(const OpDesc& op_desc) {
+  auto inputs = op_desc.InputArgumentNames();
+  std::unordered_set<std::string> input_set(inputs.begin(), inputs.end());
+  for (auto& name : op_desc.OutputArgumentNames()) {
+    if (input_set.count(name) > 0) return true;
+  }
+  return false;
+}
+
 // Search all subgraphs which all op node supported by CINN,
 // Here we using SubgraphDetector to detecte the subgraph that
 // all of op node supported by CINN. We using OpMapperRegistry
@@ -565,9 +574,10 @@ void SearchAllSubgraphs(Graph* graph) {
     if (deny_ops.size()) {
       return registered && !deny_ops.count(node->Name());
     }
+
     // if the user doesn't set FLAGS_allow_cinn_ops and FLAGS_deny_cinn_ops,
     // return true only when it is registered in CINN
-    return registered;
+    return registered && (node->IsOp() && !IsInplaceOp(*node->Op()));
   };
   VLOG(4) << "The allowed Cinn Ops: " << FLAGS_allow_cinn_ops;
   VLOG(4) << "The denied Cinn Ops: " << FLAGS_deny_cinn_ops;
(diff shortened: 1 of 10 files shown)


paddle/fluid/framework/paddle2cinn/cinn_compiler.cc
```diff
@@ -61,8 +61,8 @@ using ::cinn::hlir::framework::BuildScope;
 using ::cinn::hlir::framework::GraphCompiler;
 
 CinnCompiler* CinnCompiler::GetInstance() {
-  static CinnCompiler instance;
-  return &instance;
+  static CinnCompiler* instance = new CinnCompiler();
```
sneaxiy, line 64: The reason why I do not destruct `CinnCompiler` because the process would hang when exiting.

image

paddle/fluid/framework/paddle2cinn/build_cinn_pass.cc
```diff
@@ -545,6 +545,15 @@ void ReplaceSubGraphWithCinnOpNode(
   RemoveSubGraphFromGraph(cluster, cluster_internals, graph);
 }
 
+static bool IsInplaceOp(const OpDesc& op_desc) {
```
sneaxiy, line 548: Inplace op is not supported in CINN because of the following codes.

https://github.com/PaddlePaddle/CINN/blob/d7a2c0187e59c1b98b17eece9c8e00ce42e767d9/cinn/frontend/op_mapper_registry.cc#L26

python/paddle/fluid/tests/unittests/distributed_passes/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ if ((NOT WITH_GPU) AND (NOT WITH_XPU) AND NOT (WITH_ASCEND OR WITH_ASCEND_CL))
     list(REMOVE_ITEM TEST_OPS "test_auto_parallel_recompute_pass")
     list(REMOVE_ITEM TEST_OPS "test_auto_parallel_sharding_pass")
     list(REMOVE_ITEM TEST_OPS "test_auto_parallel_fp16_pass")
+    list(REMOVE_ITEM TEST_OPS "test_build_cinn_pass_resnet")
```
sneaxiy: The `test_build_cinn_pass_resnet.py` is skipped because it run too slow (about ~7min).

zhhsplendid: Can we keep it? Since Paddle-CI-CINN CI time is not too long compared to Paddle-CI

paddle/fluid/framework/paddle2cinn/cinn_compiler.cc
zhhsplendid: Not sure whether it is a good fix, the best would be trying to fix underlying destruct bug. I would ask @wzzju to take a look at this comment

python/paddle/fluid/tests/unittests/distributed_passes/CMakeLists.txt
sneaxiy: Done.

paddle/fluid/framework/paddle2cinn/cinn_compiler.cc
sneaxiy: As we discuss offline, we can defer this discussion on this workaround fix. The reason why this bug occurs should be checked carefully after this PR. c.c. @zhhsplendid .

zhhsplendid (approved):
LGTM

wzzju (approved):
LGTM

PR #42625, opened 2022-05-09
[external reviewing] Params to int8 pass

sfraczek:
### PR types
Function optimization

### PR changes
Others

### Describe
Added a pass that converts fp32 params to int8. For now it support only mkldnn convolution. Filter is converted to int8 and Bias to int32.

**This PR depends on** https://github.com/PaddlePaddle/Paddle/pull/42241.


paddle/fluid/framework/ir/mkldnn/params_to_int8_pass_tester.cc
```diff
@@ -0,0 +1,188 @@
...
+    const std::string var_name = input + "_var";
+    op->SetInput(input, {var_name});
+    auto var = program.MutableBlock(0)->Var(var_name);
+    test_scope.CreateTensorInScope(var_name)[0] = 1.5f;
+    return var;
+  }
+
+  void SetOutput(OpDesc* op, std::string output) {
+    const std::string var_name = output + "_var";
+    op->SetOutput("Output", {var_name});
```
wozna: Here should be output var

paddle/fluid/framework/ir/mkldnn/params_to_int8_pass_tester.cc
```diff
@@ -0,0 +1,188 @@
...
+    op->SetAttr("Scale_in", 1.0f);
+
+    AddInputToOp(op, "Input");
+    AddInputToOp(op, "Filter")->SetPersistable(true);
+    SetOutput(op, "Output");
+    return op;
+  }
+
+ protected:
+  VarDesc* AddInputToOp(OpDesc* op, std::string input) {
```
wozna: It can be moved to ProgramStrategy class

wozna: Please change name same as below

paddle/fluid/framework/ir/mkldnn/params_to_int8_pass_tester.cc
```diff
@@ -0,0 +1,188 @@
...
+#include "paddle/fluid/platform/place.h"
+
+using LoDTensor = phi::DenseTensor;
+
+namespace paddle {
+namespace framework {
+namespace ir {
+
+struct TestScope {
+  float* CreateTensorInScope(const std::string& var_name) {
```
wozna: Please change names with the same convention

paddle/fluid/framework/ir/mkldnn/params_to_int8_pass_tester.cc
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  template <typename T>
+  const T* GetTensorPtr(const std::string& input) const {
+    Variable* var = scope.FindVar(input);
+    auto tensor = var->Get<LoDTensor>();
+    return tensor.data<T>();
+  }
+
+  std::unique_ptr<Graph> CreateGraphFromProgram(const ProgramDesc& program) {
```
wozna: Please try to move CreateGraphFromProgram to ProgramStrategy class to CreateGraph

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -656,14 +656,21 @@ class ConvMKLDNNHandlerT
     if (is_test && bias_mem_p) {
       return bias_mem_p;
     } else {
-      const K* bias_data = bias->data<K>();
+      // if K is int8 (weights are int8) then biases are int32
+      using K2 = typename std::conditional<std::is_same<K, int8_t>::value,
```
wozna: Maybe it could be `K_bias` or `B` instead of `K2`

paddle/fluid/framework/ir/mkldnn/params_to_int8_pass.cc
```diff
@@ -0,0 +1,246 @@
...
+    auto* int_weights_tensor =
+        scope->Var(int_weights_node->Name())->GetMutable<LoDTensor>();
+
+    // Initialize int8 weights
+    int_weights_tensor->Resize(weights_shape);
+    auto int_weights_data =
+        int_weights_tensor->mutable_data<int8_t>(platform::CPUPlace());
+    // Quantize weights to int8
+    auto len = weights->numel() / count;
+    if (is_multi_channel) {
```
wozna: You can move this to one function and if `is_multi_channel` is False set count to 1 and this code will work for both cases.

paddle/fluid/framework/ir/mkldnn/params_to_int8_pass.cc
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    // connect new int8 tensor to weights input
+    conv_op->Op()->SetInput("Filter", {int_weights_node->Name()});
+    IR_NODE_LINK_TO(int_weights_node, conv_op);
+    GraphSafeRemoveNodes(graph, {conv_filter});
+    conv_op->Op()->SetAttr("Scale_weights", std::vector<float>(1, 1));
+
+    // Get float biases
+    if (has_bias && conv_op->Op()->Input("Bias").size() > 0) {
```
wozna: I think you can merge data calculation and setting INT8 Bias/Weights and move it to separate function.

zuzg:
What do you think about naming it to be consistent with `int8_scale_calculation_mkldnn_pass`? Maybe something like `int8_parameters_converter_mkldnn_pass`?

paddle/fluid/framework/ir/mkldnn/params_to_int8_pass_tester.cc
```diff
@@ -0,0 +1,318 @@
...
+
+    program->CheckGraph(graph);
+  }
+  std::unique_ptr<ProgramStrategy> program;
+};
+
+Data GenericInput() { return Data({1, 4, 1, 1}, {1.5f, 1.5f, 1.5f, 1.5f}); }
+Data GenericOutput() { return GenericInput(); }
+
+TEST(ParamToInt8Pass, conv_without_bias_o1i1h1w1) {
```
zuzg: You could think of using parametrized tests (maybe for `Data`?), as there are some nice options and would probably make code shorter, but on the other hand it would get pretty messy with all the namings

zuzg (dismissed):
LGTM

paddle/fluid/framework/ir/mkldnn/params_to_int8_pass_tester.cc
sfraczek: I tried  to parametrize by `shared_ptr<ConvProgramStrategy>` but it got completely unreadable

sfraczek:
> What do you think about naming it to be consistent with `int8_scale_calculation_mkldnn_pass`? Maybe something like `int8_parameters_converter_mkldnn_pass`?

I will rename it to params_quantization_mkldnn_pass

sfraczek:
Hi @jiangjiajun. I am writing this pass to replace **fp32 weights** with **int8 weights**. I want to reduce memory consumption.
Do you know if I should call both `scope->EraseVars()` and `GraphSafeRemoveNodes()` to make sure that tensor holding weights gets memory deallocated? Previously we have been using only `GraphSafeRemoveNodes()` and I don't know if it's enough. If I have to call both methods, does order of the calls matter? According to my tests, it seems that the memory isn't released even if I call both methods.

jiangjiajun:
> Hi @jiangjiajun. I am writing this pass to replace **fp32 weights** with **int8 weights**. I want to reduce memory consumption. Do you know if I should call both `scope->EraseVars()` and `GraphSafeRemoveNodes()` to make sure that tensor holding weights gets memory deallocated? Previously we have been using only `GraphSafeRemoveNodes()` and I don't know if it's enough. If I have to call both methods, does order of the calls matter? According to my tests, it seems that the memory isn't released even if I call both methods.

Hi, @sfraczek 

I think `scope->EraseVars()` and `GraphSafeRemoveNodes()` just operate the describe information, no effect on memory. 
May be you can refer to this code https://github.com/Wangzheee/Paddle/blob/f45b1dc49095b18aa03aeb8b05966c01e32f766c/paddle/fluid/framework/ir/delete_weight_dequant_linear_op_pass.cc#L387

After generate new weight, it will call `clear` to remove the original weight

sfraczek:
It should work now without memory errors. I will run more tests to see if there is less memory in some scenarios

sfraczek:
For resnet50 batch_size 50 I got peak memory usage down by 2,45%

paddle/fluid/inference/tests/api/analyzer_lexical_analysis_gru_tester.cc
```diff
@@ -200,8 +200,10 @@ std::vector<double> Lexical_Test(
       reinterpret_cast<const PaddlePredictor::Config *>(config),
       input_slots_all, outputs, FLAGS_use_analysis);
   std::vector<double> acc_res(3);
+  PADDLE_ENFORCE_GT(outputs->size(), 0,
+                    platform::errors::InvalidArgument(
+                        "Size of output must be greater than 0"));
```
tsocha:
```suggestion
                        "Number of outputs must be greater than 0"));
```

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -664,14 +664,21 @@ class ConvMKLDNNHandlerT
     if (is_test && bias_mem_p) {
       return bias_mem_p;
     } else {
-      const K* bias_data = bias->data<K>();
+      // if K is int8 (weights are int8) then biases are int32
```
tsocha, line 663: obraz

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass_tester.cc
```diff
@@ -0,0 +1,305 @@
...
+#include "paddle/fluid/platform/place.h"
+
+using LoDTensor = phi::DenseTensor;
+
+namespace paddle {
+namespace framework {
+namespace ir {
+namespace {
+struct Data {
+  Data() {}
```
tsocha:
```suggestion
  Data() = default;
```

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass_tester.cc
```diff
@@ -0,0 +1,305 @@
...
+
+  Data(std::vector<int64_t>&& data_shape, std::vector<float>&& raw_data)
+      : shape(std::move(data_shape)), data(std::move(raw_data)) {
+    auto size_from_shape = std::accumulate(shape.begin(), shape.end(), 1,
+                                           std::multiplies<int64_t>());
+    PADDLE_ENFORCE_EQ(
+        size_from_shape, data.size(),
+        platform::errors::InvalidArgument("Shape and data don't match."));
```
tsocha:
```suggestion
        platform::errors::InvalidArgument("Shape size doesn't match data size."));
```

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass_tester.cc
```diff
@@ -0,0 +1,305 @@
...
+
+  framework::Scope* Scope() { return &scope; }
+
+ private:
+  framework::Scope scope;
+  CPUPlace place;
+};
+
+struct ProgramStrategy {
+  virtual ~ProgramStrategy() {}
```
tsocha, line 69: Do you need this destructor here?

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -0,0 +1,219 @@
...
+
+  auto filter_var = scope->GetVar(conv_filter->Name());
+  if (filter_var->Get<LoDTensor>().dtype() != phi::DataType::FLOAT32) {
+    VLOG(4) << "Skipping convolution (id: " << conv_op->id()
+            << ") because it's a bug that it is detected again.";
+    return true;
+  }
+
+  VLOG(4) << "Not skipping convolution (id: " << conv_op->id() << ")";
+  return false;
```
tsocha, line 69: You log when convolution is skipped and when it's not.
But you lack of consistency here, it will be confusing to debug.
Please add a VLOG to line 74.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -0,0 +1,219 @@
...
+
+  std::string conv_bias_name = conv_op->Op()->Input("Bias")[0];
+  auto* bias_tensor = scope->GetVar(conv_bias_name)->GetMutable<LoDTensor>();
+
+  QuantizeParams<int32_t>(bias_tensor, bias_scales);
+
+  conv_op->Op()->SetAttr("Bias_scales", std::vector<float>(1, 1));
+}
+
+}  // namespace
```
tsocha: These two functions are almost the same.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -0,0 +1,219 @@
...
+                                        bool with_residual_data) const {
+  GraphPatternDetector gpd;
+  patterns::ConvResidual conv_pattern(gpd.mutable_pattern(), name_scope_);
+  conv_pattern(with_residual_data);
+
+  int params_to_int8_conv_found = 0;
+
+  auto handler = [&](const GraphPatternDetector::subgraph_t& subgraph,
+                     Graph* g) {
+    if (!IsCompat(subgraph, g)) {
```
tsocha, line 137: this function is risky since we are modifying ops out of spec.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -0,0 +1,219 @@
...
+  paddle::string::PrettyLogDetail(msg_ss.str().c_str());
+}
+
+void ParamsQuantizationMkldnnPass::ApplyImpl(ir::Graph* graph) const {
+  PADDLE_ENFORCE_NOT_NULL(graph,
+                          platform::errors::InvalidArgument(
+                              "Pointer to graph argument should not be NULL."));
+  FusePassBase::Init(name_scope_, graph);
+  Conv(graph, true);
+  Conv(graph, false);
```
tsocha: create a helper variables for these flags, or enum.
It looks like magic right now.

tsocha (changes requested):

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
sfraczek: Because it's true :D

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass_tester.cc
sfraczek: Yes because I inherit from this base class in code below

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
sfraczek: Ok

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
sfraczek: ok

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
sfraczek: What do you mean?

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
sfraczek: What do you mean or suggest here?

sfraczek: Wouldn't it look weird with enum? In future there will be more function calls here. I got inspiration from here
https://github.com/PaddlePaddle/Paddle/blob/19a7524febaf732108d6faad26724b265b56c298/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc#L1181-L1197

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
sfraczek: I think there is no need to explain here specifically in this place why biases have to be int32 for quantization.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
tsocha: AFAIK This function compare an operator with the prototxt which define what is valid.
It can fail if someone add any custom attribute etc.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
tsocha: The easiest way to fix it for now is to create a variable `auto with_residual_data = true | false` and then pass it into this function.
Don't use bare bools, opaque it with variables.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
sfraczek: @lidanqing-intel can I get rid of this IsCompat?

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -0,0 +1,212 @@
...
+  paddle::string::PrettyLogDetail(msg_ss.str().c_str());
+}
+
+void ParamsQuantizationMkldnnPass::ApplyImpl(ir::Graph* graph) const {
+  PADDLE_ENFORCE_NOT_NULL(graph,
+                          platform::errors::InvalidArgument(
+                              "Pointer to graph argument should not be NULL."));
+  FusePassBase::Init(name_scope_, graph);
+  QuantizeConv(graph, true /*with_residual_data*/);
+  QuantizeConv(graph, false /*with_residual_data*/);
```
tsocha, line 181: @jczaja we need a judge here, please select prefered solution.
```suggestion
  auto with_residual_data = true;
  QuantizeConv(graph, with_residual_data );
  with_residual_data = false;
  QuantizeConv(graph, with_residual_data );
```

paddle/fluid/operators/conv_op.cc
```diff
@@ -187,7 +187,10 @@ framework::OpKernelType ConvOp::GetExpectedKernelType(
     customized_type_value =
         (input_data_type == framework::DataTypeTrait<int8_t>::DataType() ||
          input_data_type == framework::DataTypeTrait<uint8_t>::DataType())
-            ? kConvMKLDNNINT8
+            ? OperatorWithKernel::IndicateVarDataType(ctx, "Filter") ==
+                      framework::DataTypeTrait<int8_t>::DataType()
+                  ? kConvMKLDNNINT8WS8
+                  : kConvMKLDNNINT8
```
lidanqing-vv, line 193: Does `kConvMKLDNNINT8` need to be changed to `kConvMKLDNNINT8WU8`?

sfraczek: kConvMKLDNNINT8 specifies the regular conv as worked previously - it is int8 conv but the weights are initially in fp32  and converted in first iter. 
kConvMKLDNNINT8WS8 was added to differentiate a convolution after the pass which has weights stored in int8_t (and biases in int32_t). 
Therefore, if I were to change the name  kConvMKLDNNINT8 it would have to be kConvMKLDNNINT8WFP32 or kConvMKLDNNINT8WF32.
But then I should probably also touch other operators for consistency if I begin to change names. If we're going to add more operators in future maybe that's not a bad idea.

tsocha (dismissed):

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
jczaja: @tsocha I think that current proposal is fine

paddle/fluid/operators/conv_op.cc
sfraczek: @jczaja what do you think?

sfraczek:
Hi @jiangjiajun. This approach unfortunately didn't reduce memory allocated for tensors. It appears that the size is the same despite clearing and resizing. I see that there is a ResizeLazy method called that only resizes if `new_size` > `prev_size`.   I could add a method that would resize to smaller but I don't know if this is desirable approach.
Edit: there is a prbably a bug in my test results and I am investigating it

sfraczek:
So It actually works and can be merged. The memory reduction is not clear because of how PaddlePaddle pre-allocates memory. I will still do more investigation but if I produce any further work, it will be done in a separate PR. Please merge.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
wozna: I see not all of our mkldnn passes have this check, but maybe better make sure we can skip it.

wozna (approved):
LGTM

jczaja (approved):
LGTM

jczaja:
@jiangjiajun Please review

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -660,14 +660,21 @@ class ConvMKLDNNHandlerT
...
+      const K_Bias* bias_data = bias->data<K_Bias>();
       auto user_bias_md = platform::MKLDNNMemDesc(
-          phi::vectorize(bias->dims()), platform::MKLDNNGetDataType<K>(),
+          phi::vectorize(bias->dims()), platform::MKLDNNGetDataType<K_Bias>(),
           MKLDNNMemoryFormat::x);
 
       return this->AcquireMemoryWithReorder(
           user_bias_md, this->fwd_pd_->bias_desc(),
-          platform::to_void_cast<K>(bias_data), "@bias_mem_p", is_test, {},
+          platform::to_void_cast<K_Bias>(bias_data), "@bias_mem_p", is_test, {},
```
lidanqing-vv, line 677: Great! we should name the type with clear usage. It makes life reading code easier.

lidanqing-vv (approved):
LGTM.

PR #42626, opened 2022-05-10
Use tempfile to place the temporary files

zlsh80826:
### PR types

Bug fixes

### PR changes

Others

### Describe

1. Both `test_resnet` and `test_build_strategy` use `inference` directory and `resnet.dygraph.params` as temporary files, it may cause file race if we running both unit test simultaneously. This PR use `tempfile` to store those temporary files.
2. Warp helper functions in `test_resnet.py` for `test_build_strategy.py` to use.
3. Use tempfile for `test_bert` and `test_transformer`
4. Use tempfile for `test_dataset` and `test_dataset_consistency_inspection.py`, which may cause overwriten on `test_run_with_dump_a.txt`


Aurelius84 (approved):
LGTM

PR #42634, opened 2022-05-10
add int8 for cast op

sandyhouse:
### PR types

Others
### PR changes

OPs
### Describe

add int8 support for cast op (cpu kernel)


Baibaifan (approved):
LGTM

PR #42647, opened 2022-05-10
print start time, end time for each compilation command

zhiqiu:
### PR types

New features
### PR changes

Others
### Describe

print start time, end time，pid for each compilation command

image


lelelelelez (approved):
LGTM

PR #42669, opened 2022-05-11
[MLU] fix gather error when index is 2D

fwenguang:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix cnnl error when index is 2D


qili93 (approved):
LGTM

PR #42708, opened 2022-05-12
Fix hang problem when some workers exited unexcepted 

sandyhouse:
### PR types

Others
### PR changes

Others
### Describe

Fix hang problem when some workers exited unexcepted: the master worker wait for 15min (timeout) when some worker exited.
User can set timeout value by `export FLAGS_stop_check_timeout=xxx` in seconds.


Baibaifan (approved):
LGTM

PR #42732, opened 2022-05-12
matmul and matmul_v2 refactor

Silv3S:
### PR types
Others

### PR changes
OPs

### Describe
- move inferring "-1" shape to `reshape` function, so it can be reused,
- remove redundant `PADDLE_ENFORCE` which are checked both in fuse pass and operator, with same conditions,
- remove old unit tests in which `fuse_transpose_out` and `fuse_reshape_out` are intentionally set incorrectly,
- remove code which is commented out.


Silv3S:
@jakpiase @tsocha please review

paddle/phi/core/ddim.cc
```diff
@@ -171,11 +171,26 @@ DDim stride_numel(const DDim& ddim) {
...
+  if (it != shape.end()) {
+    int index = std::distance(shape.begin(), it);
+    auto ddim_out_vec = phi::vectorize(in_dims);
+    int ddim_out_product = std::accumulate(
+        ddim_out_vec.begin(), ddim_out_vec.end(), 1, std::multiplies<int>());
+    int reshape_out_product =
+        std::accumulate(shape.begin(), shape.end(), -1, std::multiplies<int>());
+
+    shape[index] = ddim_out_product / reshape_out_product;
+  }
```
tsocha: Lets move index calculation closer to usage.
I just moved line 184, github diff is crazy here :D
```suggestion
  // dim marked as "-1" must be infered
  auto it = std::find(shape.begin(), shape.end(), -1);
  if (it != shape.end()) {
    auto ddim_out_vec = phi::vectorize(in_dims);
    int ddim_out_product = std::accumulate(
        ddim_out_vec.begin(), ddim_out_vec.end(), 1, std::multiplies<int>());
    int reshape_out_product =
        std::accumulate(shape.begin(), shape.end(), -1, std::multiplies<int>());

    int index = std::distance(shape.begin(), it);
    shape[index] = ddim_out_product / reshape_out_product;
  }
```

tsocha (dismissed):
Great work

paddle/phi/core/ddim.cc
Silv3S: git diff looks really weird, but I get the point, thank you

paddle/fluid/operators/matmul_op.cc
```diff
@@ -686,76 +686,11 @@ class MatMulOp : public framework::OperatorWithKernel {
...
-      //               "should be less than output dim size, ",
-      //               "but the index is %d and output dim size is %d", i,
-      //               ddim_out.size()));
-      //       reshape_out[i] = ddim_out.at(i);
-      //     }
-      //   }
-      // }
-
-      // if "-1" is present then one of reshape dims must be infered
-      auto it = std::find(reshape_out.begin(), reshape_out.end(), -1);
```
jczaja, line 734: Why is this part with (-1) removed. E.g. is it no longer needed?

Silv3S: I moved inferring dim marked as "-1" to `ddim::reshape()`. Now we don't need to duplicate this block of code in each operator, because function can compute final dim.

python/paddle/fluid/tests/unittests/mkldnn/test_matmul_mkldnn_op.py
```diff
@@ -519,43 +519,6 @@ def init_data_type(self):
         self.data_type_ = np.int8
 
 
-class TestMatMulOpTransposeReshapeTransposeAxisNotSupportedException(
```
jakpiase, line 522: Why these tests are deleted?

Silv3S: These tests only checked if after setting incorrect values for `fuse_reshape_out` and `fuse_transpose_out` test would fail. I deleted it, because the only way of setting these parameters is via fuse_pass, which has all PADDLE_ENFORCES to prevent such case. Also, we have separate unit test dedicated for `matmul+transpose+reshape` fuse pass.

jakpiase: Really nice that you have spotted that! Thanks!

jakpiase (dismissed):
LGTM! It's good to see another really needed refactoring done

lidanqing-vv:
@Silv3S 
```
license/cla Expected — Waiting for status to be reported
```
Maybe you need to reassign license ?

jczaja (dismissed):

Silv3S:
> Maybe you need to reassign license ?

I checked the CLA assistant manually and everything looks fine. Unfortunately, despite the renewal of the license, the CI is still not updated. Can this be merged as it is, or should I trigger CI with empty commit?

Silv3S:
@baoachun please review

lidanqing-vv (dismissed):
LGTM

paddle/phi/core/ddim.cc
```diff
@@ -171,11 +171,25 @@ DDim stride_numel(const DDim& ddim) {
   return strides;
 }
 
-DDim DDim::reshape(const std::vector<int>& shape) const {
+DDim DDim::reshape(const std::vector<int>& new_shape) const {
+  std::vector<int> shape = new_shape;
```
lidanqing-vv: Creating a new vector here and do ifelse branch in core/ddim.cc will influence each cpu/gpu op I think ?

jczaja (approved):

PR #42761, opened 2022-05-13
Correct C++ QAT pass

wozna:
### PR types

 Bug fixes
### PR changes

Others 
### Describe

I have found that pass `compute_propagate_scales_mkldnn_pass.cc` doesn't work for VGG16 and VGG19 models, because it was trying to get access to unallocated memory. 
I fixed that and I added a test for vgg16. 

Moreover, I removed `fp32_cfg.EnableMKLDNN()` because it is already set here https://github.com/wozna/Paddle/blob/678f2bf329eefa8053d1bd27a56a8130bfbe3c52/paddle/fluid/inference/tests/api/analyzer_quant_image_classification_tester.cc#L31


lidanqing-vv (dismissed):
LGTM

sfraczek (dismissed):
LGTM

sfraczek (dismissed):
LGTM

lidanqing-vv:
https://github.com/PaddlePaddle/Paddle/issues/42972 slow performance is caused by fc quantization, removing fc from quantization list is a work-around to get perf recover. But will this PR finally fix Resnet50 slow performance issue?

lidanqing-vv (dismissed):
LGTM

lidanqing-vv:
Assuming customers always turn on EnableMKLDNN() after saving int8 model, because that is more intuitive. If this PR could not fix fc quantization slow perf issue, maybe we should just comment fc quant for some time. Leaving fc as fp32 is safer. We could think long term fix.
But please check if your PR already fix #42972, thanks

sfraczek (dismissed):
LGTM

lidanqing-vv:
@jiangjiajun  Hi, could you please approve this PR ?

wozna:
> Assuming customers always turn on EnableMKLDNN() after saving int8 model, because that is more intuitive. If this PR could not fix fc quantization slow perf issue, maybe we should just comment fc quant for some time. Leaving fc as fp32 is safer. We could think long term fix. But please check if your PR already fix #42972, thanks

This PR is only fixing VGG model like VGG16 and VGG19, so it doesn't fix #42972.

wozna:
Has been fixed in another PR https://github.com/PaddlePaddle/Paddle/pull/43184

PR #42766, opened 2022-05-13
OneDNN md-in-tensor refactoring part 3: Changes in quantize and dequantize

jakpiase:
### PR types
Others

### PR changes
OPs

### Describe
Added support for memory descriptor in quantize and dequantize ops. Also made an optimization: earlier both operators were having an additional pass over whole memory to set dst memory to quantization_shift value, because we were using `reorder` with `sum` post-op. Now we are achieving the same result, but with only `reorder` with added  `zero-point` shift. That change allows us to have only one pass over memory instead of two.

### DIFF
--- a/paddle/fluid/operators/mkldnn/dequantize_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/dequantize_mkldnn_op.cc
@@ -36,100 +36,59 @@ template <typename T>
 class DeQuantOpKernel : public framework::OpKernel<T> {
  public:
   void Compute(const framework::ExecutionContext& ctx) const override {
-    auto* input = ctx.Input<Tensor>("Input");
-    auto scale_data = ctx.Attr<float>("Scale");
-    auto scale_shift = ctx.Attr<float>("Shift");
-    bool with_shift = scale_shift != 0.0f;
-    auto* output = ctx.Output<Tensor>("Output");
+    auto* x = ctx.Input<Tensor>("Input");
+    const auto quantization_scale = ctx.Attr<float>("Scale");
+    const auto quantization_shift = ctx.Attr<float>("Shift");
+    const bool with_shift = quantization_shift != 0.0f;
+    auto* out = ctx.Output<Tensor>("Output");
 
-    PADDLE_ENFORCE_NE(scale_data, 0.0f,
+    PADDLE_ENFORCE_NE(quantization_scale, 0.0f,
                       platform::errors::InvalidArgument(
                           "Dequantization scale cannot be 0.0"));
-    PADDLE_ENFORCE_GE(scale_shift, 0,
+    PADDLE_ENFORCE_GE(quantization_shift, 0.0f,
                       platform::errors::Unimplemented(
                           "Dequantization shift must be nonnegative."));
     PADDLE_ENFORCE_LE(
-        scale_shift, 255,
+        quantization_shift, 255,
         platform::errors::Unimplemented(
             "Dequantization shift must be less than or equal to 255."));
 
     auto& dev_ctx =
         ctx.template device_context<platform::MKLDNNDeviceContext>();
-    const auto& engine = dev_ctx.GetEngine();
 
-    const T* input_data = input->data<T>();
-    float* output_data = output->mutable_data<float>(ctx.GetPlace());
+    auto x_tz = phi::vectorize<int64_t>(x->dims());
+    auto x_paddle_dtype = framework::TransToProtoVarType(x->dtype());
+    auto out_paddle_dtype = framework::TransToProtoVarType(out->dtype());
 
-    float reorder_shift = -scale_shift / scale_data;
+    dnnl::primitive_attr attrs;
+    static constexpr int32_t mask = 0;  // same shift and scale for whole tensor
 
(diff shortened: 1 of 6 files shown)


paddle/fluid/operators/mkldnn/quantize_mkldnn_op.cc
```diff
@@ -34,83 +35,73 @@ template <typename T>
...
+            "255 and greater or equal to 0, but got %f", quantization_shift));
 
     auto& dev_ctx =
         ctx.template device_context<platform::MKLDNNDeviceContext>();
-    const auto& engine = dev_ctx.GetEngine();
 
-    std::vector<primitive> pipeline;
-    auto src_tz = phi::vectorize<int64_t>(input->dims());
-    auto dst_tz = phi::vectorize<int64_t>(output->dims());
+    auto x_tz = phi::vectorize<int64_t>(x->dims());
```
jczaja, line 58: Why can't we just use x->mem_desc() ?  MD contains dims if really needed for anything. but maybe Reorder Handler can be modified not to recreate MD ?

paddle/fluid/operators/mkldnn/quantize_mkldnn_op.cc
```diff
@@ -34,83 +35,73 @@ template <typename T>
...
+    }
 
     if (with_shift) {
-      dnnl::post_ops post_operations;
-      post_operations.append_sum();
-      attri.set_post_ops(post_operations);
-      uint8_t* output_data = output->mutable_data<uint8_t>(ctx.GetPlace());
-      // memset casts scale_shift to unsigned char (uint8_t) internally
-      std::memset(output_data, scale_shift, output->numel());
+      attrs.set_zero_points(DNNL_ARG_DST, mask,
```
jczaja, line 71: very nice!

jczaja (approved):
LGTM

paddle/fluid/operators/mkldnn/quantize_mkldnn_op.cc
jakpiase: you're right, I'll make an internal JIRA ticket to refactor ReorderHandler after full memory descriptor enablement

paddle/fluid/operators/mkldnn/dequantize_mkldnn_op.cc
```diff
@@ -36,100 +36,58 @@ template <typename T>
...
+
+    dnnl::primitive_attr attrs;
+    static constexpr int32_t mask = 0;  // same shift and scale for whole tensor
+
+    const float reorder_scale = 1. / quantization_scale;
+    attrs.set_output_scales(mask, {reorder_scale});
+
+    if (with_shift) {
+      attrs.set_zero_points(DNNL_ARG_SRC, mask,
+                            {static_cast<int32_t>(quantization_shift)});
```
wozna, line 70: Have you check it if zero_point works with reorder primitive. On this page https://oneapi-src.github.io/oneDNN/dev_guide_reorder.html#doxid-dev-guide-reorder  I can see that only output_scale and sum post-op is available.

jakpiase: @wozna I think that there is an issue with their documentation since DNNL_VERBOSE=1 shows following when I am running quantize tests:
<pre>onednn_verbose,exec,cpu,reorder,jit:uni,undef,src_f32::blocked:ab:f0 dst_u8::blocked:ab:f0,attr-oscale:0:127.000000 <b>attr-zero-points</b>:dst:0:128 ,,2x3,0.000976562</pre>
so it seems like zero points are supported and additionally the results of this tests are correct, so it seems fine to me

wozna: If it works that's fine. I am wondering now about shift setting in dequantize. Because if we want to perform the equation in quantize `INT8 = output_scale * (FP32 + shift)`, everything has its order, first is zero_point and then output_scale. But in dequantize we have `FP32 = INT8 / output_scale - shift`. If we follow the order, zero_point first, then output_scale, the equation looks like this `FP32 = (1 / output_scale) * (INT8 - shift * output_scale)`. So I think shift should be `quantization_shift = - scale_shift * scale_data`. Please correct me if I am wrong.
Moreover, I think that we don't have any model that uses shift in dequantize, as all RNN int8 ops have force_fp32_output, therefore it could be never been tested in the production.

wozna: In unit test I can see that the order is `scale` first, then `zero_point`. It is important to confirm this.

wozna (approved):
LGTM

paddle/fluid/operators/mkldnn/dequantize_mkldnn_op.cc
wozna: Now everything is clear.

Correct equations: 
1) **Quantize** Using reorder primitive from FP32 to INT8 with set_output_scales and set_zero_points (DST) gives following equation:
`DST = (SRC * output_scales) + zero_points` 

2) **Dequantize** Using reorder primitive from INT8 to FP32 with set_output_scales and set_zero_points (SRC) gives following equation:
`DST = (SRC - zero_points) * output_scales`

PR #42793, opened 2022-05-16
fix quant aware scale name

yghstill:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix the problem of abnormal PACT quantization training accuracy caused by scale name.


qingqing01:
Please give the reason why the bug was introduced.

PR #42820, opened 2022-05-17
fix order of matmul's result and bias in final_state_linear

pangyoki:
### PR types

Performance optimization

### PR changes

OPs

### Describe

`final_state_linear`中，`final_state_add`操作中加数的顺序与老动态图对齐。


JiabinYang (approved):
LGTM

PR #42838, opened 2022-05-18
[AutoParallel] split data in dataloader

zhaoyinglia:
### PR types

Others
### PR changes

Others
### Describe

- remove `make_data_unshard`
- split data with numpy in dataloder
- add flag `gradient_scale` in `dist_context`. If True, gradients will take average value of data parallel size after allreduce, otherwise only allreduce gradients.


JZ-LIANG (approved):
LGTM

PR #42851, opened 2022-05-18
use fp32 compute type for cublasGemmStridedBatchedEx with fp16 input/output

zhiqiu:
### PR types

Function optimization
### PR changes

OPs
### Describe

use fp32 compute type for cublasGemmStridedBatchedEx with fp16 input/output for better accuracy

Add `FLAGS_gemm_use_half_precision_compute_type` to control this feature, default is True and uses fp16 compute type, which is consistent with the old behavior.

TODO: make `FLAGS_gemm_use_half_precision_compute_type`  default to `false`.

Note: 
1. Using fp16 compute type may result in some problems, for example, https://github.com/NVIDIA/FasterTransformer/issues/49. And, also see the case in unittest.
2. Using fp16 compute type may result in `nan/inf` in some `vit(vision transformer)` models.
3. Using fp32 compute type may decrease the training speed in some models. Users can set `FLAGS_gemm_use_half_precision_compute_type` to true if no `nan/inf` willl occur in these models.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_matmul_v2_op.py
+++ b/python/paddle/fluid/tests/unittests/test_matmul_v2_op.py
@@ -495,6 +495,26 @@ class TestMatMulV2API(unittest.TestCase):
                     y = paddle.to_tensor(input_y)
                     result = paddle.matmul(x, y)
 
+    def test_compute_type_fp32(self):
+        if core.is_compiled_with_cuda():
+            place = core.CUDAPlace(0)
+            if core.is_float16_supported(place):
+                with fluid.dygraph.guard(place):
+                    input_x = np.random.random([2, 8, 16]).astype("float16")
+                    input_y = np.random.random([2, 16, 8]).astype("float16")
+                    for i in range(0, 16, 2):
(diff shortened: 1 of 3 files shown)


gongweibao (dismissed):

python/paddle/fluid/tests/unittests/test_matmul_v2_op.py
```diff
@@ -495,6 +495,26 @@ def test_dygraph_fp16(self):
                     y = paddle.to_tensor(input_y)
                     result = paddle.matmul(x, y)
 
+    def test_compute_type_fp32(self):
+        if core.is_compiled_with_cuda():
+            place = core.CUDAPlace(0)
```
gongweibao, line 500: Does it need to add another test case that will be not finite if set `FLAGS_gemm_use_half_precision_compute_type=True`

zhiqiu: done

phlrain (approved):

ZzSean (approved):
LGTM for OP-Benchmark

PR #42852, opened 2022-05-18
Remove mkldnn attributes from base ops

Silv3S:
### PR types
Others

### PR changes
OPs

### Describe
This PR is proposition to remove obsolete mkldnn attributes.


wozna (approved):
LGTM

Silv3S:
Hi @Superjomn, we would like to delete obsolete attributes from base operators e.g. `fuse_brelu` and `fuse_brelu_threshold`. These attrs are just declared, and not used anywhere in code.

Unfortunately CI failed due to compatibility issue:
> The changes of OPs will cause that the new version inference fails to load model trained by the old version

Does it also affect attributes, which are marked as `.Extra()`?

jczaja (approved):
LGTM! And here is the new leader of delete/added  code lines contest e.g. ratio 48/1 is impressive!

cyj1986 (approved):
LGTM

PR #39635, opened 2022-02-16
[NPU] Fix fill_constant and set_value for CANN 5.0.4

windstamp:
### PR types

Bug fixes
### PR changes

OPs
### Describe

1. [NPU] Fix fill_constant int64 for CANN 5.0.4 (Ascend FillD does not support int64)
Ascend Issue: https://gitee.com/ascend/modelzoo/issues/I4TRBY

2. [NPU] Fix set_value
Ascend Issue: https://gitee.com/ascend/modelzoo/issues/I4WU48

PR #42856, opened 2022-05-19
[NPU] minor changes for version control to support code without suffix

Aganlengzi:
### PR types
Others

### PR changes
OPs

### Describe
For cann version 5.0.4(without any suffix), we can not get suffix like `.alpha001`, so it need be suffixed with 000 and get 504000 for numeric comparison, this PR make minor changes for known ops.

related: https://github.com/PaddlePaddle/Paddle/pull/39635


qili93 (approved):
LGTM

PR #42941, opened 2022-05-23
[launch] fix timeout reset

kuizhiqing:
### PR types

Bug fixes

### PR changes

Others

### Describe

This pr including the following minor change,
* elastic launch timeout should be reset after each change
* context info will print after some internal rewriting which is the true value
* default run_mode to run, default gloo enable for ps
* add PADDLE_NNODES for user threads


xymyeah (approved):
LGTM

ForFishes (approved):
LGTM

PR #42943, opened 2022-05-23
Add type() interface for paddle::variant

From00:
### PR types
New features

### PR changes
Others

### Describe

Add type() interface for paddle::variant, example:
```C++
#include <iostream>
#include "paddle/phi/core/enforce.h"
#include "paddle/utils/variant.h"

int main() {
  using phi::enforce::demangle;

  paddle::variant<bool, int, float> var;

  var = true;
  std::cout << demangle(var.type().name()) << std::endl; // print "bool"

  var = 0;
  std::cout << demangle(var.type().name()) << std::endl; // print "int"

  var = 0.f;
  std::cout << demangle(var.type().name()) << std::endl; // print "float"

  return 0;
}
```

### DIFF
--- a/paddle/utils/variant.h
+++ b/paddle/utils/variant.h
@@ -2199,6 +2199,18 @@ class impl : public copy_assignment<traits<Ts...>> {
     }
   }
 
+  inline const std::type_info &type() const {
+    return visitation::alt::visit_alt_at(
+        this->index(),
+#ifdef MPARK_GENERIC_LAMBDAS
+        [](auto &alt) -> const std::type_info & { return typeid(alt.value); }
+#else
+        typer {}
+#endif
+        ,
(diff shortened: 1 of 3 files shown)


MingMingShangTian (approved):
LGTM

paddle/utils/variant.h
```diff
@@ -2199,6 +2199,18 @@ class impl : public copy_assignment<traits<Ts...>> {
     }
```
chenwhql, line 2199: 在头部补充一下更改内容的注释

chenwhql (approved):

paddle/utils/variant.h
From00: Done in PR #43100

PR #43013, opened 2022-05-26
fix the bug of adamw which set the attribute in param group not working

GuoxiaWang:
### PR types

Bug fixes
### PR changes

APIs
### Describe

exclude `learning_rate`,  set the attributes in param group will overwrite the attributes set in outer.
The `learning_rate = 0.1 * 0.5 = 0.05` used to update the param for `embedding_m.weight`

```
    optimizer = paddle.optimizer.AdamW(
        learning_rate=0.1,
        parameters=[{
            'params': embedding_l.weight,
        }, {
            'params': embedding_m.weight,
            'weight_decay': 0.0001,
            'learning_rate': 0.5,
            'beta1': 0.8
        }],
        beta1=0.9,
        beta2=0.98,
        epsilon=1e-6,
        weight_decay=0.1)
```

### DIFF
--- a/python/paddle/optimizer/adamw.py
+++ b/python/paddle/optimizer/adamw.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import defaultdict
 from .optimizer import Optimizer
-from .adam import Adam
+from .lr import LRScheduler
 from ..fluid import core
 from ..fluid import framework
-from ..fluid.framework import Variable
+from ..fluid.framework import Variable, Parameter
+from ..fluid.clip import GradientClipBase
 from ..fluid.dygraph import base as imperative_base
 from collections.abc import Callable
 from .. import _C_ops
@@ -25,7 +27,7 @@ import paddle
 __all__ = []
 
 
-class AdamW(Adam):
+class AdamW(Optimizer):
     r"""
     The AdamW optimizer is implemented based on the AdamW Optimization
     in paper `DECOUPLED WEIGHT DECAY REGULARIZATION <https://arxiv.org/pdf/1711.05101.pdf>`_.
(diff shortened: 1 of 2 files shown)


python/paddle/optimizer/adamw.py
```diff
@@ -25,7 +31,7 @@
 __all__ = []
 
 
-class AdamW(Adam):
+class AdamW(Optimizer):
     r"""
     The AdamW optimizer is implemented based on the AdamW Optimization
```
minogame, line 36: Now it is based on the base Optimizer class. Should have cleaned the documents.

python/paddle/optimizer/adamw.py
```diff
@@ -160,37 +171,108 @@ def __init__(self,
...
             raise ValueError("Invaild value of epsilon, expect epsilon >= 0.")
-        coeff = weight_decay
-        if not isinstance(coeff, float) and \
-                not isinstance(coeff, framework.Variable):
-            raise TypeError("coeff should be float or Tensor.")
-        self._params_name = set()
-        self._apply_decay_param_fun = apply_decay_param_fun
-        self._coeff = coeff
-        self._lr_to_coeff = dict()
+        if not isinstance(weight_decay, float) and \
```
minogame, line 174: Suggest to check if the type is int, for some cases the user may use `weight_decay = 0`

python/paddle/optimizer/adamw.py
```diff
@@ -160,37 +171,108 @@ def __init__(self,
...
+            # list(paddle.Tensor) will be a error value
+            if isinstance(parameters, (paddle.Tensor, core.eager.Tensor)):
+                raise TypeError(
+                    "`parameters` argument given to the optimizer should be "
+                    "an iterable of paddle Tensors, but got argument type is `{}`.".
+                    format(type(parameters)))
+            if isinstance(parameters, dict):
+                raise TypeError(
+                    "`parameters` argument should not get dict type, "
+                    "if parameter groups is needed, please set `parameters`"
```
minogame, line 195: if parameter groups **are** needed

minogame (commented):
provided some trivial suggestions.

The fix now works in my codes.

python/paddle/optimizer/adamw.py
GuoxiaWang: It means parameters need a list of dict.

TCChenlong (approved):

XiaoguangHu01 (approved):
LGTM

PR #43032, opened 2022-05-26
[Eager] Support is empty

JiabinYang:
### PR types

Function optimization
### PR changes

Others
### Describe

When we call `clear_gradients`, we set it to zero as default instead of clear it, so we need to mark this to avoid additional accumulation for leaf tensor.
image


chenwhql (approved):
LGTM

PR #43037, opened 2022-05-26
experimental nvcc-lazy-module-loading

hp03:
### PR types
Others
### PR changes
Others

### Describe
cuda 11.6 has added module lazy loading, however many environments still use cuda before 11.6.
without lazy module loading, kenrels will be loaded into memory in eager mode when library is loaded, which waste a lot of memory. this pr enable module lazy loading before cuda 11.6, and keep compatible with cuda before 11.6.

if environment var CUDA_MODULE_LOADING=LAZY is set, kernels will be loaded as needed. just like after cuda 11.6.
to enable this feature, -DEXP_CUDA_MODULE_LOADING_LAZY=ON is needed.


XiaoguangHu01 (dismissed):
LGTM

XiaoguangHu01 (approved):
LGTM

PR #43063, opened 2022-05-28
enhance check for current block and docstring for prim2orig interface

levi131:
### PR types

Function optimization

### PR changes

APIs

### Describe

enhance check for current block and docstring for prim2orig interface


cxxly (approved):
LGTM

TCChenlong (approved):

PR #43088, opened 2022-05-30
Fix the underflow of fp16 fake quantize operators

leo0519:
### PR types

Bug fixes

### PR changes

OPs

### Describe

The float16 fake quantize operators may lead to underflow when dividing the scale.

1. Move the computation of fake quantize ops from float16 to float32 (dtype is still float16)
2. Simplify the code of fake quantize op unit test


qingqing01 (approved):
LGTM

PR #43093, opened 2022-05-30
[AutoParallel] engine.prepare only once

zhaoyinglia:
### PR types

Bug fixes
### PR changes

Others
### Describe

- prepare only once for an engine object. 
- add metrics method in engine api.


JZ-LIANG (approved):
LGTM

PR #43101, opened 2022-05-30
Allocate and use new memory for temp data in cumsum kernel

leo0519:
### PR types

Bug fixes

### PR changes

OPs

### Describe

The temporary data and output data share same memory space in cumsum kernel, which leads to wrong output.
This PR allocates a new space of GPU memory for temporary data to fix the bug.

Previous correct version:
https://github.com/PaddlePaddle/Paddle/blob/release/2.2/paddle/fluid/operators/cumsum_op.cu#L257-L259


wawltor (approved):
LGTM

PR #43110, opened 2022-05-30
fix slice plugin

jiweibo:
### PR types

Others

### PR changes

Others

### Describe

 Trt plugin should not use internal stream


shangzhizhou (approved):
LGTM

PR #43114, opened 2022-05-31
[Auto Parallel] Add comp cost of ops 

Caozhou1995:
### PR types

Others
### PR changes

Others
### Describe

This PR adds comp cost of ops used in transformer-based models.


JZ-LIANG (approved):
LGTM

PR #39167, opened 2022-01-24
Optimize layer norm forward when cols is 1024.

limin2021:
### PR types
Performance optimization

### PR changes
OPs 

### Describe
Optimize the performance of layer_norm forward kernel for layer_norm op and fused_dropout_reisdual_layer_norm op.

Performance results: (use nsys to collect kernel times)
(1) layer_norm op:

batch_size*seq_len | 28672 | 14336 | 7168 | 3584 | 1792 | 896 | 448 | 224 | 112 | 56
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
时间（ns）： |   |   |   |   |   |   |   |   |   |  
apex_fast_layer_norm | 74399.931 | 41035.3163 | 20555.8185 | 11111.16 | 8395.97375 | 6855.50875 | 6275.20375 | 5652.597 | 5469.5585 | 5408.00875
paddle | 367119.365 | 187225.566 | 101543.132 | 54855.2878 | 29603.3463 | 18901.4535 | 13006.04 | 9529.146 | 8568.81325 | 8007.138
paddle_opt | 74155.0323 | 40487.696 | 21343.357 | 11008.9173 | 8566.515 | 6866.36975 | 6272.02075 | 5826.07175 | 5712.10125 | 5598.60925
加速比： |   |   |   |   |   |   |   |   |   |  
apex/paddle | 0.20265869 | 0.21917582 | 0.20243436 | 0.20255404 | 0.2836157 | 0.36269744 | 0.48248381 | 0.5931903 | 0.63830992 | 0.67539847
apex/paddle_opt | 1.00330252 | 1.0135256 | 0.96310147 | 1.00928727 | 0.98009211 | 0.99841823 | 1.00050749 | 0.97022441 | 0.95753879 | 0.96595574

结论：优化后相比优化前获得2-5x加速；优化后基本打平竞品。在个别case相比竞品略差的原因是：在计算scale * (x-mean)/var + bias时，竞品均使用fp16进行计算，paddle采取转换为fp32，使用fp32进行计算的方法，后者相比前者有一些性能损耗。

(2) fused_dropout_reisudal_layer_norm op:

batch_size*seq_len | 28672 | 14336 | 7168 | 3584 | 1792 | 896 | 448 | 224 | 112 | 56
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
时间 (ns)： |   |   |   |   |   |   |   |   |   |  
[...]

PR #43118, opened 2022-05-31
Extend forward fast layer_norm kernel to support more dimensions.

limin2021:
### PR types
Function optimization

### PR changes
OPs

### Describe
Before, fast layer_norm kernel can only support cases when cols is 1024(#39167). To make full use of the fast kernel, , #42282 and this PR try to extend to more col values.

性能：
da0184eb19b84744314d0711f843a027

4f25d471d0cc172d103da1e9c2b565bf


sneaxiy (approved):
LGTM.

PR #43135, opened 2022-06-01
[fix] split nanmedian fluid deps

Aganlengzi:
### PR types
Others

### PR changes
Others

### Describe
nanmedian_kernel.h and nanmedian_grad_kernel.h bring fluid headers because math_function.h, which causes PaddleCustomDevice fail as fluid headers included by math_function.h can not be found in paddle wheel.

this PR splits math_function.h into _impl.h and they are used by .cc and .cu files.


qili93 (approved):
LGTM

chenwhql (approved):

PR #43141, opened 2022-06-01
[ Dy2static ] fix gpu memory leakage when multi-deepcopy-and-save.

2742195759:
### PR types

Others
### PR changes

Others
### Describe

Fix gpu memory leakage when multi-deepcopy-and-save.
Original FunctionCache is a strong ref to layers.forward, which makes the layers not be gc.


0x45f (approved):
LGTM

PR #43154, opened 2022-06-01
Enable fc on bfloat16

tsocha:
### PR types
Performance optimization

### PR changes
OPs

### Describe
This change have a huge impact on bfloat16 performance on VIT-OCR model.


jczaja (dismissed):
LGTM

paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -348,6 +348,9 @@ void CpuPassStrategy::EnableMkldnnQuantizer() {
 void CpuPassStrategy::EnableMkldnnBfloat16() {
 #ifdef PADDLE_WITH_MKLDNN
   if (!use_mkldnn_bfloat16_) {
+    passes_.push_back("fc_mkldnn_pass");
+    passes_.push_back("fc_act_mkldnn_fuse_pass");
+
```
Silv3S, line 354:
```suggestion
    passes_.push_back("fc_elementwise_add_mkldnn_fuse_pass");

```
I think we can get even better results by adding `fc_elementwise_add_mkldnn_fuse_pass` here. For now it's not working, because `fc_mkldnn_pass` is disabled by default in _paddle_pass_builder.cc_. Could you add this line and run benchmark one more time?

wozna (dismissed):
LGTM

paddle/fluid/inference/api/paddle_pass_builder.cc
tsocha: Test in progress...

tsocha: I tried to enable it but unfortunately it caused a segmentation fault in bfloat16 pass.
Disabled for now.

tsocha: I confirmed that this change speed-up VIT-OCR on bfloat16 significantly.

Silv3S (approved):
LGTM

jczaja (approved):

PR #42786, opened 2022-05-16
Support CUDA Graph for partial graph in dygraph mode

sneaxiy:
### PR types
New features

### PR changes
APIs

### Describe
Support CUDA Graph for partial graph in dygraph mode.

PR #43163, opened 2022-06-02
Fix bug of CUDAGraph kernel parameter comparation

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
Fix the bug when comparing the kernel parameter in CUDAGraph nodes. It should use `std::tuple_size<TupleT>::value` instead of `sizeof(std::tuple_size<TupleT>::value)` . Related to https://github.com/PaddlePaddle/Paddle/pull/42786 .


zhiqiu (approved):
LGTM

PR #43175, opened 2022-06-02
【code format check upgrade】 step3：cpplint

betterpig:
### PR types

Others
### PR changes

Others
### Describe

upgrade cpplint to 1.6.0, this will not affect the static_check CI result. Because this CI each time use a docker image not installing cpplint, which cause it always install cpplint of default newest version, and that is 1.6.0.
image


luotao1 (approved):
LGTM

PR #42106, opened 2022-04-21
New quantization model mkldnn adaption pass

lidanqing-vv:
### PR types
New features

### PR changes
APIs

### Describe
To support to new [Feature] Adaptation of the new quantization method for MKLDNN. #37422

PR #43184, opened 2022-06-02
Correct skip_quant condition

wozna:
### PR types

Bug fixes
### PR changes

Others 
### Describe

In quant-aware quantization when there is no quantization_type there is condition to check if before operator is `quantize-dequantize` op, if not op is not quantize. Some old QAT models have no quantization_type but has `fake_quantize` op before operator (picture below)  and this should be quantized. This PR changes this condition to work for all models. 
image


lidanqing-vv (approved):
LGTM. I also found this bug and fixed in https://github.com/PaddlePaddle/Paddle/pull/42106/files. It's great your PR could be merged and this bug fixed.

sfraczek (approved):
LGTM

PR #43198, opened 2022-06-04
Matmul post-ops for fuses

Silv3S:
### PR types
Performance optimization

### PR changes
OPs

### Describe
- create method for inserting oneDNN post-ops required for fuse passes (matmul+activation or matmul+elementwise),
- reduce number of method arguments.


Silv3S:
@tsocha, @jczaja please review

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -145,16 +145,10 @@ class MatMulMKLDNNHandler
...
+  MatMulMKLDNNHandler(const dnnl::engine engine, const ExecutionContext& ctx)
       : paddle::platform::MKLDNNHandlerNoCachingT<XT, dnnl::matmul>(
             engine, ctx.GetPlace()) {
-    dnnl::primitive_attr attr;
-    float scale_out = ComputeOutputScale(ctx);
-    if (scale_out != 1.0f) {
-      constexpr unsigned tensor_wide_scale = 0;
-      attr.set_output_scales(tensor_wide_scale, {scale_out});
-    }
+    const dnnl::primitive_attr matmul_attrs = CreatePostOps(ctx);
```
tsocha: You have to change `CreatePostOps` name.
It create matmul_attrs, not postops.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -427,6 +421,19 @@ class MatMulMKLDNNHandler
...
+    dnnl::post_ops post_operations;
+
+    float scale_out = ComputeOutputScale(ctx);
+    if (scale_out != 1.0f) {
+      matmul_attrs.set_output_scales(0, {scale_out});
+    }
+
+    matmul_attrs.set_post_ops(post_operations);
+    return matmul_attrs;
+  }
```
tsocha: You have to rename this function, its name lie to me.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -497,23 +504,19 @@ static void ExecuteMatMul(const ExecutionContext& ctx) {
   auto* x = ctx.Input<Tensor>("X");
   auto* y = ctx.Input<Tensor>("Y");
   auto* out = ctx.Output<Tensor>("Out");
-  float alpha = ctx.HasAttr("alpha") ? ctx.Attr<float>("alpha") : 1.0f;
   const auto& dev_ctx =
       ctx.template device_context<paddle::platform::MKLDNNDeviceContext>();
+  const auto& onednn_engine = dev_ctx.GetEngine();
```
tsocha, line 511: Do you need to get engine here?
Since ctx is passed to MatMulMKLDNNHandler as an attribute it can be accessed from there.

tsocha (changes requested):

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Silv3S: I agree, post-ops are part of matmul_attrs and it might be confusing, but that's the pattern. We have same `CreatePostOps` method in which we set attributes and post ops for FC and Conv

tsocha: If the pattern is bad, change it.
This method is private and visible only in this file.
There is no reason to preserve bad names IMHO.

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Silv3S: Yes, but engine is passed directly to constructor, so I think it's cleaner to create object and just pass it as argument

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Silv3S: I changed naming in all files with supporting post-ops. Now it should be more accurate

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -460,12 +436,13 @@ class ConvMKLDNNHandlerT
...
-    auto scale_out_data = force_fp32_output ? 1.0f
-                          : has_activation  ? 1.0f
-                                            : ctx.Attr<float>("Scale_out");
+    float activation_scale = (!force_fp32_output && has_activation)
+                                 ? ctx.Attr<float>("Scale_out")
+                                 : 1.0f;
+
+    float scale_out_data = (force_fp32_output || has_activation)
+                               ? 1.0f
+                               : ctx.Attr<float>("Scale_out");
```
tsocha, line 446: Nice change!

tsocha (approved):

jczaja (approved):
LGTM

jczaja:

PR #43206, opened 2022-06-05
[Paddle Inference] Add call once of handle init.

xiaoxiaohehe001:
### PR types

Function optimization
### PR changes

Others
### Describe

Add call once of handle init.


zhiqiu (approved):
LGTM, lazy initialization of handles.

PR #43057, opened 2022-05-27
【code format check upgrade】 step2：cmake-format

betterpig:
### PR types

Others
### PR changes

Others
### Describe

 add cmake-format to help format cmake files.
some points:

- wont format comment, because most comment are designed carefully and already have a good style. Only few of them has problem of line width over limit, which is easy to fix manully.
- use .cmake-format.py as config file, and add some pattens to tell cmake-format how to format the self-defining function such as cc_library, which can make the code more elegant and easy to read. For example:

before:
image

after:
 image

- exclude one file `paddle/fluid/operators/CMakeLists.txt`, because tools/remove_grad_op_and_kernel.py need to update it, and errors occured after format this file, so just let it keep original format.

PR #43222, opened 2022-06-06
【code format check upgrade】 step3： cmakelint

betterpig:
### PR types

Others
### PR changes

Others
### Describe

add [cmakelint](https://github.com/cmake-lint/cmake-lint) to lint cmake files
some points:

- format CMakeLists.txt which was left out by PR #43057
- add cmakelint hook and its config file, set some filter that are not needed
- temporarily exclude some files which need to be fixed, remove them when fix is done.
image
- yapf‘s hook config should change sha to rev, to avoid warning.


luotao1 (approved):
LGTM

PR #43228, opened 2022-06-06
[cuda graph] Add cuda graph attr to op desc

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
add cuda graph attr in python side op desc
pass this attr to backward ops during auto_grad part


python/paddle/fluid/backward.py
```diff
@@ -255,6 +261,9 @@ def _add_needed_descs_to_block(descs, block, main_block, in_memory_vars):
             if name not in in_memory_vars:
                 is_needed = True
         if is_needed:
+            if origin_is_operator:
+                global _grad_op_id_to_fwd_op
+                _grad_op_id_to_fwd_op[desc.original_id()] = origin_desc
```
sneaxiy: What if `origin_is_operator` is false?

python/paddle/fluid/backward.py
```diff
@@ -273,6 +282,9 @@ def _add_descs_to_block(descs, block):
     backward = core.op_proto_and_checker_maker.OpRole.Backward
     for desc in descs:
         if isinstance(desc, framework.Operator):
+            # for recompute, should record recompute ops
+            global _grad_op_id_to_fwd_op
+            _grad_op_id_to_fwd_op[desc.desc.original_id()] = desc
```
sneaxiy: Same above.

python/paddle/fluid/backward.py
```diff
@@ -614,6 +629,12 @@ def _addup_repetitive_outputs_(op_descs, block_idx, grad_var_to_var=None):
         # If not reverse, we first insert 'a' at idx 1, it becomes [0, 1, 'a', 2], and then insert 'b' at idx 2, it becomes [0, 1, 'a', 'b', 2].
         idx = key
         for i, op in enumerate(value):
+            # update the mapping between fwd and bwd
+            target_idx = idx - 1 if idx == op_descs_len else idx + i
```
sneaxiy, line 635: No need to update?

python/paddle/fluid/backward.py
```diff
@@ -42,6 +42,9 @@
                                 logging.INFO,
                                 fmt='%(asctime)s-%(levelname)s: %(message)s')
 
+_grad_op_id_to_fwd_op = {
```
sneaxiy: I prefer not using the global variable. Try to pass an argument to the method.

python/paddle/fluid/backward.py
FeixLiu: `origin_is_operator ` is false means that the desc is an instance of OpDesc on cpp side, it has been added to the map earlier when creating that cpp desc. Here we only handle if desc is an instance of Operation on python side, which is used for recompute's ff_ops.

python/paddle/fluid/backward.py
FeixLiu: same above

python/paddle/fluid/backward.py
FeixLiu: Should update? These op descs are created from `_create_op_desc_` func. I think these updates are necessary.

python/paddle/fluid/backward.py
FeixLiu: Any difference between these two implementations? Is this a best practice? I think global variable is somehow easy to use 😂. Besides, if pass an argument, it may have a very long using chain.

sneaxiy: Yes, the global variable is somehow easy to implement. The global variable should record the global status but not the local status. The `_grad_op_id_to_fwd_op` should be a map related to a certain program but not all programs.

FeixLiu: Make sense, just update

sneaxiy (approved):
LGTM.

zhiqiu (approved):
LGTM

PR #43237, opened 2022-06-06
Use tempfile to place the temporary files

shiyutang:
### PR types

Bug fixes

### PR changes

Others

### Describe

Use the tempfile package to replace the file paths that is not randomly generated, and clean up them afterwards.


XieYunshen (approved):
LGTM

PR #43248, opened 2022-06-06
【code format check upgrade】 step3：optimize installation and instruction of clang-format

betterpig:
### PR types

Others
### PR changes

Others
### Describe

Local installation of clang-format is not easy, optimize installation and instruction of it.

- add default python version limit
image
- upgrade pip before pip install since low version pip does not have the source of clang-format whl.
image
after upgrade:
image


luotao1 (approved):
LGTM

PR #43266, opened 2022-06-07
Add use_master_acc_grad for DistributedFusedLamb

sneaxiy:
### PR types
New features

### PR changes
OPs

### Describe
Add `use_master_acc_grad` argument for `DistributedFusedLamb`.  This argument is only valid when `gradient_accumulation_steps > 1`.

- If `use_master_acc_grad=True`, the gradient accumulation would use FP32 master gradient for accumulating FP16 gradient.
- If `use_master_acc_grad=False`, the gradient accumulation would use FP16 gradient for accumulating FP16 gradient.

This change would save about 600M GPU memory for MLPerf BERT Large model. After this PR, the MLPerf BERT Large model can run on 8 40G A100 GPUs when local batch size is 48 and gradient accumulation step is 8.


limin2021 (approved):
LGTM.

Shixiaowei02 (approved):

PR #43317, opened 2022-06-08
[Dy2Stat]Support deepcopy Net instance after @to_static

Aurelius84:
### PR types

New features
### PR changes

APIs
### Describe

[Dy2Stat]Support deepcopy Net instance after @to_static

```python
  import copy
  import paddle
  class Net(paddle.nn.Layer):
      def __init__(self):
          super(Net, self).__init__()
      def forward(self, x, flag=True):
          if flag:
              out = x + 1
          else:
              out = x - 1
          return out
  x = paddle.randn([10, 1], 'float32')
  net = paddle.jit.to_static(Net())  # convert into static mode
  copy_net = copy.deepcopy(net)      # deepcopy a new net without @to_static
```


0x45f (approved):
LGTM

PR #43330, opened 2022-06-08
Fix shrink downstream op bugs for standalone executor

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

In standalone executor, the OP dependency graph (i.e., op_downstream_map) will be shrunk to reduce some redundant dependencies. For a example served as the following graph pattern, **c** is indirectly dependent on **a** through **b**, so the directly dependence **a**->**c** is unnecessary and it can be deleted to improve performance.

image

This optimization is feasible when all dependences in the graph indicate data-dependences, i.e., **a**->**b** means there is a variable that is both the output of **a** and the input of **b**. However, in standalone executor, to finely control the execution order of OPs, some control-dependences are added.  Due to the existence of control-dependences, necessary data-dependences may be deleted . This result in the loss of data-dependence information in the built graph, and may cause some bugs in subsequent OP scheduling. For example, if the missing date-dependence relate to two cross-stream OPs, these two OPs will be considered as having no data-dependency in scheduling and not be synchronized,  then an asynchronous error occurrs. 

To fix this problem, this PR move the codes of dependence shrinking to the front of adding control-dependences.

### DIFF
--- a/paddle/fluid/framework/new_executor/interpretercore_util.cc
+++ b/paddle/fluid/framework/new_executor/interpretercore_util.cc
@@ -629,46 +629,75 @@ void update_var_min_rw_op(const std::map<int, std::set<int>>& op2dependences,
   var2min_rw_op->at(rw_var).push_back(cur_op);
 }
 
-std::map<int, std::list<int>> get_downstream_map(
-    const std::map<int, std::set<int>>& op2dependences,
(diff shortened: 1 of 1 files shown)


zhiqiu (approved):
LGTM

PR #43352, opened 2022-06-09
Sparse gradient add

zkh2016:
### PR types

Others
### PR changes

Others
### Describe

GradTensorHolder::add添加对SparseCooTensor梯度累加的支持

### DIFF
--- a/paddle/fluid/eager/grad_tensor_holder.cc
+++ b/paddle/fluid/eager/grad_tensor_holder.cc
@@ -18,6 +18,7 @@
 #include "paddle/fluid/framework/convert_utils.h"
 #include "paddle/fluid/framework/var_type.h"
 #include "paddle/fluid/imperative/gradient_accumulator.h"
+#include "paddle/phi/core/sparse_coo_tensor.h"
 #include "paddle/phi/kernels/funcs/math_function.h"
 
 namespace egr {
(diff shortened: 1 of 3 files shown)


paddle/fluid/eager/grad_tensor_holder.cc
```diff
@@ -130,6 +131,26 @@ void GradTensorHolder::add(size_t slot_id, size_t rank,
                                                   &new_buffer);
         buffer_tensor.set_impl(new_buffer.impl());
       }
+    } else if (t.is_sparse_coo_tensor()) {
```
JiabinYang, line 134: support this is accumulation_node.cc

JiabinYang: and how about dense add to sparse?

JiabinYang (commented):
some comment

paddle/fluid/eager/grad_tensor_holder.cc
zkh2016: > support this is accumulation_node.cc

done

> and how about dense add to sparse?

The sparse API requires that the gradient of SparseTensor is still a SparseTenosr, so dense+sparse is not required at present.

JiabinYang (approved):
LGTM

PR #43382, opened 2022-06-09
Generalize conv+activation fuse pass

Silv3S:
### PR types
Others

### PR changes
OPs

### Describe
- Generalize `conv+relu`, `conv+mish`, `conv+gelu` etc. as one `conv+activation` fuse pass,
- Adjust existing unit tests.


python/paddle/fluid/tests/unittests/ir/inference/test_conv_act_mkldnn_fuse_pass.py
```diff
@@ -170,33 +172,53 @@ def sample_program_config(self, draw):
...
                               inputs={"X": ["conv2d_out"]},
                               outputs={"Out": ["relu_out"]})
         if act_type == "swish":
-            self.passes = ["conv_swish_mkldnn_fuse_pass"]
             beta = draw(st.floats(min_value=0.1, max_value=1.0))
             act_op = OpConfig("swish",
                               inputs={"X": ["conv2d_out"]},
                               outputs={"Out": ["swish_out"]},
                               beta=beta)
+        if act_type == "gelu":
```
tsocha: Since there can be only one activation wouldn't be better to use elseif for this ladder instead of ifs?

paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -298,15 +298,8 @@ void CpuPassStrategy::EnableMKLDNN() {
...
              "conv_elementwise_add_mkldnn_fuse_pass",
              "conv_concat_relu_mkldnn_fuse_pass",
-             "conv_relu_mkldnn_fuse_pass",          //
-             "conv_leaky_relu_mkldnn_fuse_pass",    //
-             "conv_relu6_mkldnn_fuse_pass",         //
-             "conv_swish_mkldnn_fuse_pass",         //
-             "conv_hard_swish_mkldnn_fuse_pass",    //
-             "conv_mish_mkldnn_fuse_pass",          //
-             "conv_hard_sigmoid_mkldnn_fuse_pass",  //
              // TODO(baoachun) fix int8 accuracy
```
tsocha: I think we can remove this TODO from here. What do you think?

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
```diff
@@ -272,73 +163,114 @@ Conv2DGeluFusePass::Conv2DGeluFusePass() {
...
-}  // namespace framework
-}  // namespace paddle
+void ConvActivationMkldnnFusePass::ApplyImpl(Graph* graph) const {
+  std::vector<std::string> act_types = {
+      "relu", "mish",  "swish",   "hard_swish", "hard_sigmoid",
+      "gelu", "relu6", "sigmoid", "leaky_relu"};
 
-REGISTER_PASS(conv_activation_mkldnn_fuse_pass,
-              paddle::framework::ir::ConvActivationFusePass);
+  std::vector<std::string> conv_types = {"conv2d"};
```
tsocha: Do you know if we want to add more conv types for this pass?

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
```diff
@@ -272,73 +163,114 @@ Conv2DGeluFusePass::Conv2DGeluFusePass() {
...
+      conv_op->SetAttr("fuse_alpha",
+                       activation->Op()->GetAttrIfExists<float>("beta"));
+      conv_op->SetAttr("fuse_beta",
+                       activation->Op()->GetAttrIfExists<float>("alpha"));
+    } else {
+      conv_op->SetAttr("fuse_alpha",
+                       activation->Op()->GetAttrIfExists<float>("alpha"));
+      conv_op->SetAttr("fuse_beta",
+                       activation->Op()->GetAttrIfExists<float>("beta"));
+    }
```
tsocha: I don't like a solution with else if ladder.
Please use a map from: paddle/fluid/framework/ir/mkldnn/elt_act_mkldnn_fuse_pass.cc:37
This map is cleaner IMO.

tsocha (changes requested):

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
jakpiase: Maybe we should make a common pass for fusing op+activation or make it unified in some another way since the map that you have made in elt+act can be reused basically everywhere where activation is being fused into preceding op

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
Silv3S: I added this vector, so we can add `conv3d` or `conv_transpose` in the future

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
Silv3S: This is very good idea. I wonder if we should include `sqrt`, `abs` or `clip` because they use same mechanism, but they're not activations. Maybe just rename "operator_activation_pass" into something more inclusive?

tsocha: You can add it, why not.
I'm afraid that currently we don't have a better name for it.
If you will get it I think it can be renamed.

jakpiase: OneDNN is describing `sqrt`, `abs` and activations like `relu`, `gelu` etc. as `eltwise` operations, but in PaddlePaddle elementwise operations are totally different thing that in oneDNN, so I don't think that there exist a better option than just treating it as `activation`

paddle/fluid/inference/tests/api/analyzer_vit_ocr_tester.cc
```diff
@@ -108,7 +108,7 @@ TEST(Analyzer_vit_ocr, fuse_status) {
       static_cast<AnalysisPredictor *>(predictor.get()), &num_ops);
 
   CHECK_EQ(fuse_status.at("fc_mkldnn_pass"), 33);
-  CHECK_EQ(fuse_status.at("conv_activation_mkldnn_fuse"), 2);
+  CHECK_EQ(fuse_status.at("conv2d_gelu_mkldnn_fuse_pass"), 2);
```
lidanqing-vv: Hi @Silv3S Great job ! Thank you very much for the refactoring! 
Here I thought `conv2d_gelu_mkldnn_fuse_pass` has been deleted, why it is still here?

paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -399,14 +392,7 @@ void CpuPassStrategy::EnableMkldnnInt8() {
...
     passes_.push_back("conv_concat_relu_mkldnn_fuse_pass");
-    passes_.push_back("conv_relu_mkldnn_fuse_pass");
-    passes_.push_back("conv_leaky_relu_mkldnn_fuse_pass");
-    passes_.push_back("conv_relu6_mkldnn_fuse_pass");
-    passes_.push_back("conv_swish_mkldnn_fuse_pass");
-    passes_.push_back("conv_hard_swish_mkldnn_fuse_pass");
-    passes_.push_back("conv_mish_mkldnn_fuse_pass");
-    passes_.push_back("conv_hard_sigmoid_mkldnn_fuse_pass");
-    passes_.push_back("conv_gelu_mkldnn_fuse_pass");
+    passes_.push_back("conv_activation_mkldnn_fuse_pass");
```
lidanqing-vv, line 396: Could you also delete `conv_relu_mkldnn_fuse_pass.cc, conv_relu_mkldnn_fuse_pass.h`, `conv_leaky_relu_mkldnn_fuse_pass.cc, conv_leaky_relu_mkldnn_fuse_pass.h` and many these kind of files? Thank you !

lidanqing-vv:
@jczaja Could you please review, this PR removes a lot of redundancies

paddle/fluid/inference/api/paddle_pass_builder.cc
lidanqing-vv: Hi, I think this TODO can be removed. Accuracy issues will be found and fixed by models UT.

jakpiase (dismissed):
LGTM

paddle/fluid/inference/api/paddle_pass_builder.cc
Silv3S: These passes didn't have separate files. They were all declared inside `conv_activation_mkldnn_fuse_pass` by deriving base `conv_activation` and overwriting activation name.

paddle/fluid/inference/tests/api/analyzer_vit_ocr_tester.cc
Silv3S: This part needs further analysis and refactoring, because current approach can be misleading and it also affects other passes with activations.

`fuse_status.at()` is checking how many ops should be fused in optimized version of model. This information is added via `AddStatis()` method. Unfortunately, approach with bulk fusion of `conv+activation` returns number of fusion only for last pair of fused `conv+activation`.  For example only how many `conv+mish` were fused, instead of cumulated sum for all `conv+activations`. That's why for now I'm saving number of each `conv+activation` individually.

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
```diff
@@ -14,110 +14,104 @@
...
+}
+
+void ConvActivationMkldnnFusePass::FuseConvAct(
+    Graph* graph, const std::string& conv_type, std::string& act_type,
+    const std::unordered_map<std::string, std::string>& attrs_map) const {
   PADDLE_ENFORCE_NOT_NULL(
       graph, platform::errors::InvalidArgument("Graph cannot be nullptr."));
-  FusePassBase::Init("conv_activation_mkldnn_fuse", graph);
+  FusePassBase::Init(conv_type + "_" + act_type + "_mkldnn_fuse_pass", graph);
```
lidanqing-vv, line 60: Hi, ok so here the passes with different activations will have different names and AddStatis belongs to different passes. That's why CHECK_EQ(fuse_status.at("conv2d_gelu_mkldnn_fuse_pass"), 2); can find correctly.
But it is fuse_statis not fuse_status. But that is a small thing.

lidanqing-vv (dismissed):
LGTM! 

Could you please also merge `test_mkldnn_conv_hard_swish_fuse_pass.py`, `test_mkldnn_conv_mish_fuse_pass.py` ... into one `test_mkldnn_conv_activation_autoscanner_fuse_pass.py`?
Because the  `conv_activation_pass` detect different kind of patterns, I think the test should test out more complex case like `conv->gelu->relu->conv->mish` etc. This kind of test will also verify if conv_activation_pass after fusion conv->gelu, will continue fusing conv (gelu)->relu (Not sure if it will happen in real model), but it should be fused because conv accept several post-ops. You can consider it in the next PRs.

Also I saw for-loop is used, I just worry if the sequence of the activations will influence the fusions? It will be great if you could explain, thanks!

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
```diff
@@ -23,6 +23,94 @@ namespace ir {
...
+    IR_NODE_LINK_TO(conv, activation_out);
+    GraphSafeRemoveNodes(graph, {activation, conv_out});
+    found_conv_activation_count++;
+  };
+
+  gpd(graph, handler);
+  AddStatis(found_conv_activation_count);
+  if (!Has("disable_logs") || !Get<bool>("disable_logs")) {
+    PrettyLogDetail("---    fused %d conv with %s activation",
+                    found_conv_activation_count, act_type);
```
tsocha, line 115: Logic to disable logs should be placed inside PrettyLogDetail IMO.
Lets try to move it there.

python/paddle/fluid/tests/unittests/ir/inference/test_conv_act_mkldnn_fuse_pass.py
```diff
@@ -194,31 +193,6 @@ def sample_program_config(self, draw):
...
-                              outputs={"Out": ["mish_out"]})
-        if act_type == "hard_sigmoid":
-            slope = draw(st.floats(min_value=0, max_value=10))
-            offset = draw(st.floats(min_value=0, max_value=10))
-            act_op = OpConfig("hard_sigmoid",
-                              inputs={"X": ["conv2d_out"]},
-                              outputs={"Out": ["sigmoid_out"]},
-                              slope=slope,
-                              offset=offset)
-        if act_type == "hard_swish":
```
tsocha: Why did you remove it?

tsocha (changes requested):
obraz
Why some of the activations have its own files but some have not?

python/paddle/fluid/tests/unittests/ir/inference/test_conv_act_mkldnn_fuse_pass.py
```diff
@@ -170,28 +170,24 @@ def sample_program_config(self, draw):
...
             act_op = OpConfig("leaky_relu",
                               inputs={"X": ["conv2d_out"]},
                               outputs={"Out": ["relu_out"]},
                               alpha=alpha)
         if act_type == "relu":
-            self.passes = ["conv_relu_mkldnn_fuse_pass"]
             act_op = OpConfig("relu",
                               inputs={"X": ["conv2d_out"]},
                               outputs={"Out": ["relu_out"]})
         if act_type == "swish":
```
tsocha: Since you have modified this file already, could you replace ifs with elifs as well?

python/paddle/fluid/tests/unittests/ir/inference/test_conv_act_mkldnn_fuse_pass.py
Silv3S: I wanted to refactor `conv+activation` and also merge existing unit tests into one. But with all changes PR grew too large and required additional approvals. It's harder to maintain reviews and CI, so currently I canceled unit tests modifications. I will refactor/adjust unit tests in next PR.

Silv3S:
> obraz Why some of the activations have its own files but some have not?

That's actually very good question. It was probably due to introduction of new testing framework and old tests weren't rewritten. Current approach with multiple files has much redundancy, because only few lines of code are changed with respect to activation type.

As Danqing mentioned, we should delete existing unit tests and create generic `test_mkldnn_conv_activation_autoscanner_fuse_pass` instead. It will generate much more test cases, with different activations.

paddle/fluid/framework/ir/mkldnn/conv_activation_mkldnn_fuse_pass.cc
Silv3S: I didn't notice this typo, thank you!

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -505,41 +505,20 @@ class ConvMKLDNNHandlerT
...
-      post_operations.append_eltwise(activation_scale,
-                                     dnnl::algorithm::eltwise_hardswish,
-                                     fuse_alpha, fuse_beta);
-    } else if (fuse_activation == "mish") {
-      post_operations.append_eltwise(activation_scale,
-                                     dnnl::algorithm::eltwise_mish, fuse_alpha,
-                                     fuse_beta);
-    } else if (fuse_activation == "hard_sigmoid") {
+
+    if (fuse_activation == "hard_sigmoid") {
```
tsocha, line 509: Is it possible to add `hard_sigmoid` to `AcquireActivationAlgorithm` method?

tsocha (dismissed):

tsocha (approved):

jczaja (approved):
LGTM

jczaja:

lidanqing-vv (approved):
LGTM

PR #43412, opened 2022-06-10
[AutoParallel] fix fetch list

zhaoyinglia:
### PR types

Others
### PR changes

Others
### Describe

- update parameter`fetches` type from 'list' to 'list[str] | list[var] | dict{str: str} | dict{str: var}'
- example:
```
fetches = {'loss': model.loss}
engine.fit(data, batch_size, fetches=fetches)
# {'epoch': 0, 'step': 0, 'train_loss': ...}
```


JZ-LIANG (approved):
LGTM

PR #43440, opened 2022-06-13
[cuda graph] partial program with cuda graph under static mode

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
Support partial program with cuda graph under static mode.

Use `wrap_cuda_graph` to warp the function as cuda graph function. 
Then use `cuda_graph_transform` to transform the origin program to cuda graph progrm. 
NOTE that, the return value of `cuda_graph_transform` should be preserved in order to keep the cuda graph sub-program undeleted.


python/paddle/fluid/tests/unittests/test_cuda_graph_partial_graph_static_run.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return rst
+
+    def test_static_mode_cuda_graph(self):
+        if not is_cuda_graph_supported():
+            return
+        x_data = np.random.random((3, 10)).astype('float32')
+        cuda_graph_rst = self.run_with_cuda_graph(x_data)
+        normal_run_rst = self.normal_run(x_data)
+        assert np.allclose(cuda_graph_rst, normal_run_rst)
```
sneaxiy: Why allclose? It should be `np.array_equal`?

python/paddle/device/cuda/graphs.py
```diff
@@ -106,3 +109,323 @@ def wrap_cuda_graph(function, mode="thread_local", memory_pool="default"):
...
+    internal_section = []
+    internal_idx = []
+
+    current_section = []  # current recording cuda graph sections
+    current_idx = []  # current recording cuda graph ops' idx
+    current_cuda_graph_id = -1  # current recording cuda graph id
+    op_role_attr_name = core.op_proto_and_checker_maker.kOpRoleAttrName()
+
+    for idx, op in enumerate(block.ops):
+        assert op.type != 'conditional_block' and op.type != 'while', \
```
sneaxiy: If the program contains `conditional_block` or `while` ops but these ops are not marked with `wrap_cuda_graph`, should it be OK ?

python/paddle/device/cuda/graphs.py
```diff
@@ -106,3 +109,323 @@ def wrap_cuda_graph(function, mode="thread_local", memory_pool="default"):
...
+            local_cuda_graph_id = int(cuda_graph_attrs[2])
+            if local_cuda_graph_id == current_cuda_graph_id:
+                if len(internal_section) > 0:
+                    assert len(internal_section) == len(
+                        internal_idx
+                    ), "len of internal section should be equal with len of internal idx"
+                    for internal_op in internal_section:
+                        if internal_op.attr(
+                                op_role_attr_name) == 256 or internal_op.attr(
+                                    op_role_attr_name) == 257:
```
sneaxiy: Should we use enum value instead of `256` and `257`?

sneaxiy (commented):
Great work with little comments.

python/paddle/fluid/tests/unittests/test_cuda_graph_partial_graph_static_run.py
FeixLiu: In general, just leave some space for precision fluctuation. The default relative tolerance for `allclose` is 1e-5 and the absolute tolerance is 1e-8, which are sufficient for precision check under fp32.

python/paddle/device/cuda/graphs.py
FeixLiu: 256 has it's corresponding enum. Is there a enum for 257? It should be a bit operation between loss(256) and grad(1)?

python/paddle/device/cuda/graphs.py
FeixLiu: Great point, will be updated to: if the op is `conditional_block` or `while`, it's cuda graph attr must be none.

python/paddle/fluid/tests/unittests/test_cuda_graph_partial_graph_static_run.py
sneaxiy: Yes. But in fact, some operators are quite deterministic (such as matmul, and dropout with fixed seed). If it is not `np.array_equal`, this must be a bug instead of normal accuracy deviation.

sneaxiy (approved):
LGTM.

PR #43479, opened 2022-06-14
modify index dtype from int to int64_t of concat_and_split_functor

GuoxiaWang:
### PR types

Function optimization
### PR changes

OPs
### Describe

modify index dtype from int to int64_t of concat_and_split_functor


Xreki (approved):
LGTM

zhiqiu (approved):
LGTM，need to check if the performance will be affected.

PR #43510, opened 2022-06-14
[Bug Fix] set device_id=current_id when calling Tensor.cuda() without device_id

betterpig:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix issue #41819. When device_id not given, it should be set by current device id instead of 0.

- Note: use plat::form::GetCurrentDeviceId can not get the real current device_id, it has a latency, the device id will be set only when `x = paddle.to_tensor(1.0)` called.
image

Before:
image
After:
image

### DIFF
--- a/paddle/fluid/pybind/imperative.cc
+++ b/paddle/fluid/pybind/imperative.cc
@@ -1535,38 +1535,40 @@ void BindImperative(py::module *m_ptr) {
                 "Cannot copy this Tensor to GPU in CPU version Paddle, "
                 "Please recompile or reinstall Paddle with CUDA support."));
 #else
-             int device_count = platform::GetGPUDeviceCount();
-             int device_id = 0;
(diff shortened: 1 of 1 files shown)


paddle/fluid/pybind/imperative.cc
```diff
@@ -1535,40 +1535,40 @@ void BindImperative(py::module *m_ptr) {
...
-               return self;
-             } else {
-               auto new_var = self->NewVarBase(place, blocking);
-               new_var->SetOverridedStopGradient(self->OverridedStopGradient());
-               return new_var;
-             }
+            int device_count = platform::GetGPUDeviceCount();
+            int device_id = 0;
+            if (handle == py::none()) {
+              auto default_place =
```
zhwesky2010, line 1541: LGTM

zhwesky2010 (approved):

Issue #43232, opened 2022-06-06
Remove mkldnn related attributes in native ops and set them in passes.


lidanqing-vv:
### 需求描述 Feature Description

Remove mkldnn related attributes in native op and add and set them in passes.

lidanqing-vv:
@Silv3S Please link later PRs with this issue.
First proposition PR has been merged: https://github.com/PaddlePaddle/Paddle/pull/42852
Second PR waiting for review: https://github.com/PaddlePaddle/Paddle/pull/43160

lidanqing-intel:
@Silv3S Hi, please related to this issue when you have new PR that removing attributes. Thanks!

lidanqing-vv:
### 1. Baidu has concern with this solution forward and backward compatibility.
Please test follows:

Chinese
1. 旧版本save的，新版本能不能正确执行：a)是能否load起来, b)结果是否正确
2. 新版本save的，旧版本的能不能正确执行：a)是能否load起来, b)结果是否正确

English
1. models saved using old version
   - Could the new code load the old version model ?
   - Is the acc and perf results correct ?
2. models saved using new version
   - Could the old code load the new code saved model ?
   - Is the acc and perf results correct ?
   
Baidu has the concern because in Baidu other business apartments upgrade paddle very slowly so they would like to have two directions compatibility

### 2. If your solution could not pass above compatiblity tests, Baidu propose following solution
目前对于conv，matmul， elementwise系列，有一些部分attribute都是给mkl-dnn的kernel准备的
1） 新增一个mkldnn_xx_op，把对应的attribute放大这个op中
2） pass升级，原来pass里面直接set对应op的attribute，需要先将op名称改下，再set
3） 兼容性方面，需要converter来保证兼容性，以conv为例，旧的模型load起来，如果发现attribute比转移到了mkldnn-conv中，需要把op名称改了； 保存的时候，如果是给新版本用，先不要特殊处理，如果给旧版本用，需要加一个convert，将mkldnn-conv改成conv

### 3. If the current solution could pass the compatibility that's best, if not we can discuss if  Baidu's proposal is light enough.

Silv3S:
I ran multiple tests to check if any error occurs. Luckily all the tests worked.

1. Models saved using old version.
[...]

PR #43519, opened 2022-06-14
matmul+activation fuse pass

Silv3S:
### PR types
Performance optimization

### PR changes
OPs

### Describe
- fuse matmul_v1 with all supported activations,
- add required unit tests.


paddle/fluid/framework/ir/mkldnn/matmul_activation_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,301 @@
...
+
+  AddOpCompat(OpCompat("sqrt"))
+      .AddInput("X")
+      .IsTensor()
+      .End()
+      .AddOutput("Out")
+      .IsTensor()
+      .End();
+
+  AddOpCompat(OpCompat("abs"))
```
jakpiase: Why there is no AddOpCompat for `clip` activation function?

paddle/fluid/operators/activation_op.cc
```diff
@@ -492,26 +492,6 @@ class SoftplusOpMaker : public framework::OpProtoAndCheckerMaker {
         "(bool, default false) Only used in cudnn kernel, need install cudnn.")
         .SetDefault(false)
         .AsExtra();
-    AddAttr<std::string>(
```
jakpiase: Are we allowed to just delete deprecated parameters? I was thinking that due to backward compatibility they must stay as they were before

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
```diff
@@ -453,6 +454,30 @@ class MatMulMKLDNNHandler
...
+        const auto activation_algorithm =
+            paddle::platform::AcquireActivationAlgorithm(fuse_activation);
+        post_operations.append_eltwise(
+            activation_scale, activation_algorithm, fuse_alpha, fuse_beta);
+      }
+    }
+
     matmul_attrs.set_post_ops(post_operations);
     return matmul_attrs;
   }
```
jakpiase: What do you think about creating a common static function which would append eltwises into post_ops object? That behavior is common between all operators that are fused with activation, and all these functions look exactly the same. By using single function we will be able to add support for more activation with less redundancy

jakpiase (commented):
Awesome initiative! I have left some comments, but the code looks generally really nice

paddle/fluid/operators/activation_op.cc
Silv3S: Currently, there is a discussion and looking for a way to get rid of oneDNN-only attributes, which were used in fuse passes. Since we can get their values via `ExecutionContext`, we could slim base operators down https://github.com/PaddlePaddle/Paddle/issues/43232

paddle/fluid/framework/ir/mkldnn/matmul_activation_mkldnn_fuse_pass.cc
Silv3S: I missed it, thanks!

paddle/fluid/operators/mkldnn/matmul_mkldnn_op.cc
Silv3S: That's great idea. I will extend previously added `AcquireActivationAlgorithm` to append eltwises into post_ops, instead of just returning `dnnl::algorithm`

Silv3S:
@piotrekobi, @tsocha please review

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -587,18 +583,9 @@ class ConvMKLDNNHandlerT
...
-                                     fuse_alpha,
-                                     fuse_beta);
-      post_operations.append_eltwise(
-          activation_scale, dnnl::algorithm::eltwise_clip, 0.0f, 1.0f);
-    } else if (fuse_activation != "") {
-      const auto activation_algorithm =
-          platform::AcquireActivationAlgorithm(fuse_activation);
-      post_operations.append_eltwise(
-          activation_scale, activation_algorithm, fuse_alpha, fuse_beta);
+    if (!ctx.Attr<std::string>("fuse_activation").empty()) {
```
tsocha: `!` is easy to miss, lets make a helper variable
```suggestion
    auto is_fuse_activation_not_empty = !ctx.Attr<std::string>("fuse_activation").empty();
    if (is_fuse_activation_not_empty) {
```

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -587,18 +583,9 @@ class ConvMKLDNNHandlerT
...
+      paddle::platform::AppendActivation(
+          ctx, post_operations, activation_scale);
```
tsocha: Do you need to pass `activation_scale` as an argument?
I can see that in this context its value is 1.0 or read from attribute.

paddle/fluid/operators/elementwise/mkldnn/elementwise_mkldnn_op.h
```diff
@@ -50,21 +50,8 @@ class EltwiseMKLDNNKernel : public framework::OpKernel<T> {
...
-      const float beta = ctx.HasAttr("activation_beta")
-                             ? ctx.Attr<float>("activation_beta")
-                             : 0.0f;
-
-      const auto activation_algorithm = platform::AcquireActivationAlgorithm(
-          ctx.Attr<std::string>("activation_type"));
-
-      post_operations.append_eltwise(scale, activation_algorithm, alpha, beta);
+    if (ctx.HasAttr("fuse_activation")) {
+      platform::AppendActivation(ctx, post_operations);
```
tsocha: I think that `AppendActivation` can check if `ctx` have a valid `fuse_activation` attribute inside, it will simplify usage of this function(no need to write such `ifs` here and there.

paddle/fluid/framework/ir/mkldnn/softplus_activation_mkldnn_fuse_pass.h
```diff
@@ -36,7 +36,7 @@ class SoftplusActivationOneDNNPass : public FusePassBase {
 
   void FuseSoftplusActivation(
       ir::Graph *graph,
-      const std::string &fuse_activation_type,
```
tsocha: Since its only a variable name you can postpone this change to another PR to reduce number of modified files to 20(no approval needed).

paddle/fluid/framework/ir/mkldnn/matmul_activation_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,313 @@
...
+    found_matmul_activation_count++;
+  };
+
+  gpd(graph, handler);
+  AddStatis(found_matmul_activation_count);
+  if (!Has("disable_logs") || !Get<bool>("disable_logs")) {
+    PrettyLogDetail("---    fused %d matmul with %s activation",
+                    found_matmul_activation_count,
+                    act_type);
+  }
```
tsocha, line 93: It would be nice to move this if inside `PrettyLogDetail` but I won't enforce it.

paddle/fluid/framework/ir/mkldnn/matmul_activation_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,313 @@
...
+
+  int found_matmul_activation_count = 0;
+  auto handler = [&](const GraphPatternDetector::subgraph_t& subgraph,
+                     Graph* g) {
+    VLOG(4) << "handle " + matmul_type + "+" + act_type + " fuse";
+
+    if (!IsCompat(subgraph, g)) {
+      LOG(WARNING) << "matmul_activation_mkldnn_fuse_pass op compat failed.";
+      return;
+    }
```
tsocha, line 56: Do we need this if here?
It can disable this fuse if this method will complain about custom attrs etc.

paddle/fluid/framework/ir/mkldnn/matmul_activation_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,313 @@
...
+        matmul_op->SetAttr(attrs.second, act_op->GetAttr(attrs.first));
+      }
+    }
+
+    if (act_type == "gelu" && activation->Op()->HasAttr("approximate")) {
+      act_type = BOOST_GET_CONST(bool, activation->Op()->GetAttr("approximate"))
+                     ? "gelu_tanh"
+                     : "gelu_erf";
+      matmul_op->SetAttr("fuse_alpha", 0.0f);
+      matmul_op->SetAttr("fuse_beta", 0.0f);
```
tsocha: elt_act_mkldnn_fuse_pass don't need these attributes to be set explicitly so this pass don't need it either IMHO.
```suggestion
```

tsocha (changes requested):

paddle/fluid/framework/ir/mkldnn/elt_act_mkldnn_fuse_pass.cc
```diff
@@ -45,15 +45,15 @@ void ElementwiseActivationOneDNNPass::ApplyImpl(Graph *graph) const {
...
       else if (act_type == "clip") {
-        attr_map.emplace("min", "activation_alpha");
-        attr_map.emplace("max", "activation_beta");
+        attr_map.emplace("min", "fuse_alpha");
+        attr_map.emplace("max", "fuse_beta");
       } else {
-        attr_map.emplace("alpha", "activation_alpha");
-        attr_map.emplace("beta", "activation_beta");
+        attr_map.emplace("alpha", "fuse_alpha");
+        attr_map.emplace("beta", "fuse_beta");
```
piotrekobi: This code is repeated in softplus_activation_mkldnn_fuse_pass.cc and it's probably a good idea to minimize the use of nested if statements. Maybe attr_map can be a created somewhere as a static object instead of being created dynamically during every pass?

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
Silv3S: Yes, it's needed because conv operator is different than others. All other op+act fuse passes use default value 1.0, only `conv_mkldnn_op` computes it from attributes(`force_fp32_output`, `scale_out` and `activation_scale`).  If more ops compute `activation_scale` I will move this logic to `AppendActivation`.

paddle/fluid/framework/ir/mkldnn/softplus_activation_mkldnn_fuse_pass.h
Silv3S: That's good idea, but I need 2 other approvals (setting TIMEOUT and removing attrs) anyway, so exceeding 20 files shouldn't delay PR merge

paddle/fluid/framework/ir/mkldnn/matmul_activation_mkldnn_fuse_pass.cc
Silv3S: I think this topic needs further discussion, because some fuse passes have `OpCompat` for each fusion combination and others don't have any. We should decide if it gives us any benefit and then unify all passes

paddle/fluid/framework/ir/mkldnn/elt_act_mkldnn_fuse_pass.cc
Silv3S: Great idea 👍 I moved setting `attr_map` to helper function and also created method returning list of all supported activations, so it will be much easier to add new activations to all passes at once.

jakpiase (dismissed):
LGTM

paddle/fluid/framework/ir/mkldnn/elt_act_mkldnn_fuse_pass.cc
```diff
@@ -17,44 +17,25 @@
...
-        attr_map.emplace("threshold", "fuse_alpha");
-      else if (act_type == "clip") {
-        attr_map.emplace("min", "fuse_alpha");
-        attr_map.emplace("max", "fuse_beta");
-      } else {
-        attr_map.emplace("alpha", "fuse_alpha");
-        attr_map.emplace("beta", "fuse_beta");
-      }
+      auto attr_map = GetAttributeMap(act_type);
       FuseElementwiseAct(graph, elt_type, act_type, attr_map);
```
tsocha: You can reduce number of parameters by moving attr_map extraction inside FuseElementwiseAct function.
Let's do this here and in other passes.
```suggestion
      FuseElementwiseAct(graph, elt_type, act_type);
```

tsocha (changes requested):

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -1008,32 +1008,88 @@ class ActivationMKLDNNHandler
...
-  PADDLE_ENFORCE_NE(activation_type,
-                    activation_map.end(),
-                    platform::errors::InvalidArgument(
-                        "Activation '%s' not found in oneDNN algorithms mapper",
-                        activation_name));
-  return activation_type->second;
+static void AppendActivation(const framework::ExecutionContext& ctx,
+                             dnnl::post_ops& post_ops,
+                             float activation_scale = 1.0f) {
+  if (!ctx.HasAttr("fuse_activation")) return;
```
tsocha: Could you move this if to `AppendActivation` as well? part 2/2
```suggestion
  auto invalid_attribute = ctx.HasAttr("fuse_activation")
                           ? ctx.Attr<std::string>("fuse_activation").empty()
                           : true;
  if (invalid_attribute) return;
```

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -587,18 +583,10 @@ class ConvMKLDNNHandlerT
...
-    } else if (fuse_activation != "") {
-      const auto activation_algorithm =
-          platform::AcquireActivationAlgorithm(fuse_activation);
-      post_operations.append_eltwise(
-          activation_scale, activation_algorithm, fuse_alpha, fuse_beta);
+    bool has_activation = !ctx.Attr<std::string>("fuse_activation").empty();
+    if (has_activation) {
+      paddle::platform::AppendActivation(
+          ctx, post_operations, activation_scale);
     }
```
tsocha: Could you move this if to `AppendActivation` as well? part 1/2
```suggestion
    paddle::platform::AppendActivation(
        ctx, post_operations, activation_scale);
```

tsocha (changes requested):

Silv3S:
@jiangjiajun please review

Silv3S:
@XiaoguangHu01 could you please approve exceeding the limit of modified files?

Silv3S:
@chalsliu could you please approve setting TIMEOUT for newly added unit test?

tsocha (dismissed):

jczaja (dismissed):
LGTM

piotrekobi (dismissed):
LGTM

jczaja (approved):

PR #43567, opened 2022-06-16
Fix for FC for quant model

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix for FC for quant model. The error was associated with FC kernel not using infer shape and instead of that inferring output dims inside its kernel without updating oneDNN's memory descriptor, which caused softmax kernel to use previous batch size.


lidanqing-vv (approved):
LGTM

PR #43585, opened 2022-06-16
[Auto Parallel] Use a fast completion for data parallelism

aoyulong:
### PR types
Others

### PR changes

Others

### Describe

The dataflow analysis for the annotation completion seems to be a little slow when the number of nn layers is over ten thousands. However, we can use a fast completion for data parallelism without using the complicated dataflow analysis.


XieYunshen (approved):
LGTM for  
` set_tests_properties(test_engine_api_dp
                       PROPERTIES LABELS "RUN_TYPE=EXCLUSIVE" TIMEOUT 80)`

PR #43620, opened 2022-06-17
[Windows CI] make paddle build and unit test in python virtual environment

betterpig:
### PR types

Others
### PR changes

Others
### Describe

make paddle build and unit test in python virtual environment.
Benefits:

- Each CI task use a new virtual environment, avoid possible influence by previous task.
- CI use new environment without installed dependencies, can test whether the PR has set the python requirements properly.
- make sure CI python environment consistent and not use changed environment, e.g., the python env may be changed by one RD when he or she is debuging locally.

Influence:
- create new virtual environment and reinstall python dependencies cost a time about 1.5min average, if all dependencies can  use local cache, which is the general case.

### DIFF
--- a/paddle/scripts/paddle_build.bat
+++ b/paddle/scripts/paddle_build.bat
@@ -92,15 +92,18 @@ rem ------initialize set git config------
 git config --global core.longpaths true
 
 rem ------initialize the python environment------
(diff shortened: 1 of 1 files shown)


zhwesky2010 (approved):
LGTM

PR #43623, opened 2022-06-17
Fix test_distributed_fused_lamb_op_with_clip UT failure

Wong4j:
### PR types
Bug fixes

### PR changes
Others

### Describe
When testing NGC paddle-22.06, the `test_distributed_fused_lamb_op_with_clip` failed.

System info:
DGX V100
CUDA 11.7
cuDNN 8.4
paddle v2.3.0

Error info:
```
======================================================================
FAIL: test_main (__main__.TestDistributedFusedLamb)
----------------------------------------------------------------------
  File "/opt/paddle/paddle/build/python/paddle/fluid/tests/unittests/distributed_fused_lamb_test_base.py", line 303, in test_main
    self.run_main(use_fp16=False)
  File "/opt/paddle/paddle/build/python/paddle/fluid/tests/unittests/distributed_fused_lamb_test_base.py", line 299, in run_main
    self.assertTrue(max_diff < atol, msg)
AssertionError: False is not true : max_diff = 1.1920928955078125e-07 atol = 1e-07 when use_fp16 = False , use_master_param_norm = True
```
This test is to train `resnet` respectively with `lamb` and `DistributedFusedLamb` optimizer and then compare the value of parameters. I have checked all `max_diff`,  only 1 out of 62 exceeded the threshold 1e-7. And `1.1920928955078125e-07` is very close to `1e-7`, so I don't think it's an bug.

I can't reproduce it with `paddlepaddle/paddle:latest-dev-cuda11.4.1-cudnn8-gcc82` container and `paddle-v2.3.0` on V100. So I assume that this error is due to a numerical calculation difference caused by CUDA or cuDNN version difference. This failure may occur in the future paddle version. 

Resetting the random seed or increasing the tolerance can both make the test pass. So I choose to slightly increase the tolerance from 1e-7 to 2e-7 in this commit.

If you have better idea, we can discuss it here.


Wong4j:
This bug has been solved by @sneaxiy. Close this PR.

PR #43633, opened 2022-06-17
[code format] resume headers order of .cu file

betterpig:
### PR types

Others
### PR changes

Others
### Describe

Clang-format dest not consider .cu as main source files by default, so it sort the headers of .cu just by ASCIIbetical order.

- resume the orders of .cu files following Google C++ style such like below:
image
- Config Clang-format to consider .cu file as main source file, and not sort  headers cross different block, thant means you can seperate headers using block to avoid sorted


chenwhql (dismissed):
LGTM

luotao1 (dismissed):

XiaoguangHu01 (approved):
LGTM

PR #43708, opened 2022-06-21
Fix batch csr

zkh2016:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix batch csr when 0th batch is zero


zhwesky2010 (approved):
LGTM

PR #43722, opened 2022-06-21
fix the cumsum bug for large size

wawltor:
### PR types
Function optimization 

### PR changes
OPs 

### Describe
fix the cumsum bug for large size.
**problem**: In the cumsum kernel used the two dims block to cumsum the input tensor, but the max  of thread in the dim of `block.y` is  65535, in some case  the dim of `block.y`  exceed   65535, the computation of the kernel will be shutdown.

**solution**: Use the 1 dim of block to compute. 

```python
import numpy as np
import paddle

#  Small shape Tensor,  the result is correct 
x = np.random.randn(2, 128, 128)

x_paddle = paddle.to_tensor(x.copy(), stop_gradient=False)
print('paddle cumsum; ', x_paddle.cumsum(0).abs().sum().numpy())

# Big shape Tensor, the result is error 
x = np.random.randn(2, 1280, 1280)
x_paddle = paddle.to_tensor(x.copy(), stop_gradient=False)
print('paddle cumsum; ', x_paddle.cumsum(0).abs().sum().numpy())

```

### DIFF
--- a/paddle/phi/kernels/gpu/cum_kernel.cu
+++ b/paddle/phi/kernels/gpu/cum_kernel.cu
@@ -176,10 +176,8 @@ __global__ void BlockScanKernel(T* d_out,
   } temp_storage;
 
   int bx = blockIdx.x;
-  int by = blockIdx.y;
 
   BlockPrefixCallbackOp<T, Op> prefix_op(Identity<T, Op>::value, op);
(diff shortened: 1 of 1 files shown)


joey12300 (approved):
LGTM

Issue #42972, opened 2022-05-25
Could not create a primitive descriptor while running quantized model with MKLDNN


jiangjiajun:
### bug描述 Describe the Bug

We have test quantized mobilenet with PaddlePaddle develop version(not sure the commit id, but it's nightly build version 2022.05.24)

And find that while infer with different batch size, it will throw an error `Could not create a primitive descriptor for a reorder primitive`

The quantized model is running well with PaddlePaddle v2.3, but in 2.3 we need preprocess the quantized model by [this script](https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/contrib/slim/tests/save_quant_model.py)

In develop, we only need to use `enable_mkldnn_int8` instead of preprocessing the quantized model.
[quant_bug.tar.gz](https://github.com/PaddlePaddle/Paddle/files/8767459/quant_bug.tar.gz)

yeliang2258:
In our tests, we found that in Paddle's 2.3 and dev versions, the resnet50 quantization model did not get accelerated on the cpu, even slower than FP32, but running the same model and script in version Paddle 2.2 can get a significant acceleration, related The model, reproduction script and test data are in the network disk link.

My cpu is Intel(R) Xeon(R) Gold 6271 with avx512_vnni support.

Link: https://pan.baidu.com/s/1Cz10shHRKlw_x4f0rUxImQ 
Extraction code: mrvp

My test results are as follows:
Paddle2.3 + MKLDNN +  FP32：397.18ms
Paddle2.2 + MKLDNN + INT8 + save_quant_model ：111.59ms，
Paddle2.2 + MKLDNN + INT8 + save_quant_model ：529.34ms
Paddle develop + MKLDNN  + INT8 + mkldnnint8：539.26ms
![image](https://user-images.githubusercontent.com/30516196/173009415-37d8074f-e071-4108-9f11-bc727fbdd952.png)

wozna:
> In our tests, we found that in Paddle's 2.3 and dev versions, the resnet50 quantization model did not get accelerated on the cpu, even slower than FP32, but running the same model and script in version Paddle 2.2 can get a significant acceleration, related The model, reproduction script and test data are in the network disk link.
> 
> My cpu is Intel(R) Xeon(R) Gold 6271 with avx512_vnni support.
> 
> Link: https://pan.baidu.com/s/1Cz10shHRKlw_x4f0rUxImQ Extraction code: mrvp
> 
[...]

PR #43729, opened 2022-06-21
[WIP] Remove gpu_cpu_reshape2_matmul_fuse_pass in EnableMkldnn

wozna:
### PR types

Performance optimization
### PR changes

Others 
### Describe

This PR removes gpu_cpu_reshape2_matmul_fuse_pass from EnableMkldnn, EnableMkldnnInt8 and save_quant_model.py. When reshape2 is fused with matmul and then matmul and elementwise is fused to FC, Mkldnn FC for a batch size bigger than 1 slows down. This corrected Resnet50 performance, but I am still waiting for results for other models.

### DIFF
--- a/paddle/fluid/inference/api/paddle_pass_builder.cc
+++ b/paddle/fluid/inference/api/paddle_pass_builder.cc
@@ -286,7 +286,11 @@ void CpuPassStrategy::EnableMKLDNN() {
 #ifdef PADDLE_WITH_MKLDNN
   if (!use_mkldnn_) {
     passes_.insert(passes_.begin(), "mkldnn_placement_pass");
-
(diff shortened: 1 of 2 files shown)


wozna:
I have finished results and unfortunately removing pass `gpu_cpu_reshape2_matmul_fuse_pass` corrects performance only for Resnet50 QAT model mentioned in this issue https://github.com/PaddlePaddle/Paddle/issues/42972#issuecomment-1137015393. Which means that the best solution for now is to allow users to disable FC when necessary and then find a solution to use FC only for optimized kernels.

lidanqing-vv:
@wozna  So how is this PR's influence to other models ? Does other models slow down because of this PR or what?

PR #43803, opened 2022-06-23
Performance fix for recommender model

jakpiase:
### PR types
Performance optimization

### PR changes
OPs

### Describe
Three changes were made to improve the performance of the recommender model:

- elementwise_*: now we are always broadcasting Y into X, because it allows us to have optimal performance with the usage of binary primitive
- reduce_* : oneDNN's kernel is disabled if reduction is not being done throughout the outermost dims, because doing that was resulting in reference version being called which is terribly slow
- concat: if concat op is being used as a stack op(concatenating exactly same tensors which are having "1" dim on the concat dims, then plain format tag is enforced, because oneDNN were forcing the usage of non-optimal format tags, it is the same case as was done in stack op

Performance table measured on Intel(R) Core(TM) i9-9940X CPU @ 3.30GHz:
|OneDNN before this PR |Native Paddle  | OneDNN after this PR| 
|---|---|---|
| 1446ms| 786ms| 477ms| 

So the speed up compared too  oneDNN bfore this PR is 3.03x and the speed up for PaddlePaddle native is 1.65x.


lidanqing-vv:
@heliqi Hi please test on your target machine, I think your machine is faster than us.

paddle/fluid/operators/elementwise/mkldnn/elementwise_mkldnn_op.h
```diff
@@ -151,6 +157,13 @@ class EltwiseMKLDNNGradKernel : public ElemwiseGradKernel<T> {
     auto* dy = ctx.Output<Tensor>(framework::GradVarName("Y"));
     auto* dout = ctx.Input<Tensor>(framework::GradVarName("Out"));
 
+    // oneDNN's binary is optimized for broadcasting y into x, so in other case
```
jczaja, line 168: shouldn't we report to oneDNN request to optimize X broadcasting ? So we could could get rid of 
hacks at some point?

jakpiase: I think that they are aware of that and they expect users to always broadcast Y into X, but we can create an issue for them. Anyway, for now we will need that workaround and we can delete it later if they will fix it

heliqi:
Test results on an internal machinee(Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz):

|OneDNN before this PR |	Native Paddle |	OneDNN after this PR |
|:---|:---|:---|
|260ms	 |       130ms	| 73ms |

lidanqing-vv:
@yaomichael  The performance after this PR is 73ms. And Baidu's goal is to improved to lower than 50ms

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -657,8 +657,15 @@ class BinaryMKLDNNHandler
     auto attributes =
         CreateAttributes(algo, scale_x, scale_y, scale_out, post_ops);
 
-    this->AcquireForwardPrimitiveDescriptor(attributes, algo, src0_md, src1_md,
-                                            dst_md);
+    // oneDNN's binary is optimized for broadcasting y into x, so in other case
+    // we have to swap tensors to achieve optimal performance
+    if (x->numel() > y->numel()) {
```
lidanqing-vv: Hi, if x and y have been swapped, does it need to change the src0_md to src1_md? I think the md is from x or y, right ?

lidanqing-vv:
Please notice, there is a bug in the demo code. In the demo,
```
  restore(varNameTensorMap);
  for (int i = 0; i < 20; ++i) {
    predictor->Run();
  }
  auto t1 = time();
  for (int i = 0; i < 100; ++i) {
    predictor->Run();
  }
//  auto t2 = time();
//    auto t1 = time();
//    predictor->Run();
    auto t2 = time();

  std::cout << "predictor run takes " << time_diff(t1, t2)/10.0 << "ms" << std::endl;
```
Here it should be `time_diff(t1, t2)/100.0`. Then we can get similar results as @heliqi

paddle/fluid/platform/mkldnn_reuse.h
jakpiase: Hi, to not break the earlier handler behavior, x and y are swapped after creating BinaryMKLDNNHandler, so they must be swapped inside of it as well

lidanqing-vv (approved):
LGTM.

lidanqing-vv:
@jczaja Could you please review this PR and so that it could be merged soon?

jczaja (approved):

PR #43806, opened 2022-06-24
larger test_collective_sendrecv_api timeout

Aganlengzi:
### PR types
Others

### PR changes
Others

### Describe
test_collective_sendrecv_api may fail because of timeout in CI


Aganlengzi:
has been done because 2.3 CI treat test_collective_sendrecv_api as one card ut

PR #43862, opened 2022-06-27
[UPDATE FLUID API] PaddleRec-related only

wangzhen38:
### PR types
Others

### PR changes
APIs

### Describe
add some fluid api to paddle dir

### DIFF
--- a/python/paddle/__init__.py
+++ b/python/paddle/__init__.py
@@ -35,7 +35,6 @@ from .framework import set_flags  # noqa: F401
 from .framework import disable_static  # noqa: F401
 from .framework import enable_static  # noqa: F401
 from .framework import in_dynamic_mode  # noqa: F401
-from .fluid.layers import l2_normalize  # noqa: F401
 from .fluid.dataset import *  # noqa: F401
 
 from .framework.dtype import dtype as dtype  # noqa: F401
@@ -646,5 +645,4 @@ __all__ = [  # noqa
     'put_along_axis',
     'heaviside',
     'tril_indices',
-    'l2_normalize',
 ]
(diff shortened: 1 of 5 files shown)


python/paddle/static/__init__.py
```diff
@@ -64,11 +64,14 @@
...
 
 from ..fluid.layers import create_parameter  # noqa: F401
 from ..fluid.layers import create_global_var  # noqa: F401
 from ..fluid.layers.metric_op import auc  # noqa: F401
 from ..fluid.layers.metric_op import accuracy  # noqa: F401
+from ..fluid.contrib.layers import ctr_metric_bundle  # noqa: F401
+from ..fluid.layers import exponential_decay  # noqa: F401
 
 __all__ = [  #noqa
     'append_backward', 'gradients', 'Executor', 'global_scope', 'scope_guard',
```
jeff41404, line 77: shall we add ctr_metric_bundle  and exponential_decay in list of \_\_all\_\_ ?

python/paddle/__init__.py
```diff
@@ -35,6 +35,8 @@
 from .framework import disable_static  # noqa: F401
 from .framework import enable_static  # noqa: F401
 from .framework import in_dynamic_mode  # noqa: F401
+from .fluid.layers import l2_normalize  # noqa: F401
```
jeff41404: shall we use `paddle.divide` and `paddle.linalg.norm` instead of l2_normalize, so no need import l2_normalize.

python/paddle/static/__init__.py
wangzhen38: done

python/paddle/__init__.py
wangzhen38: done

jeff41404 (approved):
LGTM

frankwhzhang (approved):

PR #43957, opened 2022-06-29
[Dy2Stat]Enhance nonlocal machanism while returning single var

Aurelius84:
### PR types

Bug fixes
### PR changes

Others
### Describe

[Dy2Stat]Enhance nonlocal machanism while returning single var.


2742195759 (approved):
LGTM

PR #43962, opened 2022-06-29
[new-exec] support runing with different scope and the same program using scope_guard

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

[new-exec] support runing with different scope and the same program using scope_guard

for example,

```
for i in range(10):
    x_data = x_data + 1
    scope = paddle.fluid.Scope()
    with paddle.static.scope_guard(scope):
        exe.run(prog, feed={"x": x_data})
```


From00 (approved):
LGTM

PR #44023, opened 2022-07-01
Set FC input data format to ANY 

wozna:
### PR types

Bug fixes
### PR changes

OPs 
### Describe

This PR sets the input data format to `any`, so the OneDNN library can choose the most optimized format. This corrects the problem of performance drop for larger batches.


lidanqing-vv:
This is the fix PR that could finally solve the large batch size poor performance issue? That will be great! Thanks!

wozna:
> This is the fix PR that could finally solve the large batch size poor performance issue? That will be great! Thanks!

Yes, this PR resolves performance drop for bigger batches for int8 models, that was mentioned here https://github.com/PaddlePaddle/Paddle/issues/42972#issuecomment-1137015393. I checked all needed tests and everything works correctly.

lidanqing-vv (approved):
LGTM.
I have tested this works for me. Thank you for your fixes !

PR #44024, opened 2022-07-01
Remove boost::static_visitor

From00:
### PR types
Others

### PR changes
Others

### Describe

Remove some unnecessary _boost::static_visitor_ directly. For other cases related to _VisitPlace_ that require the _Visitor::result_type_, replace them with _std::unary_function_.


zhiqiu (approved):
LGTM, BTW, std::unary_function is deprecated in c++11 and will be removed from c++17.

PR #44095, opened 2022-07-05
【Auto Parallel】update base cost

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

This pr updates base cost of cost model, providing some base function for calling by dist op and other modules.


JZ-LIANG (approved):
LGTM

PR #44103, opened 2022-07-05
force single thread when CINN is on

Shixiaowei02:
### PR types

Others
### PR changes

Others
### Describe

At present, when CINN is on, each process can only contain one thread, so reset the number of threads to 1.

image


zhiqiu (approved):
LGTM

PR #44116, opened 2022-07-06
fix quant scale name

ZhangHandi:
### PR types

Bug fixes

### PR changes

APIs

### Describe

Fix the problem of quantization caused by scale name.


yghstill (dismissed):
LGTM，need fix CI.

yghstill (approved):

PR #44125, opened 2022-07-06
Fix nan in fast_ln_fwd_kernel when cols > 1024

ZzSean:
### PR types

Bug fixes
### PR changes

OPs
### Describe

- Fix nan in fast_ln_fwd_kernel when cols > 1024
When cols > 1024, the condition `tidx == 0` is correct only when cols is 4096. 
Otherwise, the wrong value in shared memory will be used during reduction, therefore the condition should be `tidx % THREADS_PER_ROW == 0`


limin2021 (dismissed):
LGTM for modifications of fused_ln.

qingqing01 (approved):

PR #41944, opened 2022-04-18
make inference_api_test compile with dynamic linking library

betterpig:
### PR types

Others
### PR changes

Others
### Describe

Make inference_api_test compile with shared lib in windows

- make unittests in paddle/inference/tests/api link paddle_inference_shared and paddle_inference_c_shared rather than 
a list of static libraries( paddle_inference_api, paddle_inference_io, ir_pass_manager, analysis_predictor, benchmark), and that is what will happen in linux.
- make paddle_inference_shared in the begining of all dependencies. Because, the order of linking libraries, depending on the order of dependencies, may influence the .obj included by the executable program. For example, if put paddle_inference_shared in the end, the compiler will find the symbols needed in the static lib such as libzero_copy_tensor.lib and then include corresponding zero_copy_tensor.cc.obj. But paddle_inference.dll also contains symbols from zero_copy_tensor.cc.obj, and then conflict happens. If put paddle_inference_shared in the begining, the zero_copy_tensor.cc.obj will not be included and problem gone. In short, .dll must be linked, but .obj may not, so we should put the static lib in the back of shared lib.
image
- remove PADDLE_API in place.h, it not, symbols redefinition happen and cant be fixed. According to the CI result, the modification seems not have influence on linux.
image
- make inference ut link paddle_inference_c_shared in windows and paddle_inference_c in linux. In windows, if still link paddle_inference_c in windows, symbols redefinition happen. While in linux, it change to paddle_inference_c_shared, error below happened.
image
- compile paddle_inference_c_shared even when ON_INFER=OFF, since inference ut need to link paddle_inference_c_shared in windows.
- 4 *_test.cc get changed. These tests will fail before due to .exe cant get device context pool initialized in .dll, so .exe must initialize its own pool before calling SetFakeImageInput which will call executor->run.

#### 动态库导出符号变化：
Windows:
- paddle_inference.dll由develop的922个符号变为PR的3085个符号，不再包含GetOrRegisterGlobalDeviceTypeId和GetGlobalDeviceType。
- core_avx.pyd由develop的1242个符号变为PR的1235符号，减少的额符号为GetOrRegisterGlobalDeviceTypeId，GetGlobalDeviceType，Exp_EnableUseGpuFp16@AnalysisConfig，Exp_EnableUseGpuFp16@GpuPassStrategy，Exp_EnableUseGpuFp16@PassStrategy，gpu_fp16_enabled@AnalysisConfig，use_gpu_fp16@PassStrategy.

Linux:
[...]

PR #44126, opened 2022-07-06
make inference_c test linking only paddle_inference_c

betterpig:
### PR types

Others
### PR changes

Others
### Describe

See the original PR #41944 #43936, after these 2 PR merged, the code becomes complicated and paddle_inference_c_shared build but not used in linux. So make inference_c test linking only paddle_inference_c.
- the paddle_inference_shared is not needed since paddle_inference_c already depend paddle_inference
- make 8 c_api test still compile with static library so that we needn's compile paddle_inference_c_shared when ON_INFER=OFF 
- Since only 8 test compiling method were restored, the optimization effect described in PR #41944 has not been influenced almost, the linking time of C++ test in paddle\fluid\inference\tests\api is still 10 second around.

### DIFF
--- a/paddle/fluid/inference/capi/CMakeLists.txt
+++ b/paddle/fluid/inference/capi/CMakeLists.txt
@@ -20,6 +20,10 @@ cc_library(
   SRCS ${C_API_SRCS}
   DEPS paddle_inference)
 
+if(NOT ON_INFER)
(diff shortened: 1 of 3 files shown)


zhwesky2010 (approved):
LGTM

PR #44135, opened 2022-07-06
Fix random seed for several unit tests

zlsh80826:
### PR types

Others

### PR changes

Others

### Describe

1. There are some unit tests are inherited from `unittest.Testcase` instead of `OpTest`. Those tests missed to fix the random seed which caused CI random failures.
2. Use `np.testing.assert_allclose` instead of `self.assertTrue(np.allclose(...))`. The former shows the detailed difference, while the later only shows `True is not False`.


qili93 (dismissed):
LGTM

zlsh80826:
Hi @qili93, the conflict was solved.

qili93 (approved):
LGTM

PR #44138, opened 2022-07-06
[AutoParallel] fix 'op_role' for gradient merge & recompute

zhaoyinglia:
### PR types

Function optimization
### PR changes

Others
### Describe

- fix 'op_role' attr to reduce 'allreduce' operate in micro batch parse
- fix 'op_role' attr to guarantee the recompute op's 'op_role' is 1


JZ-LIANG (approved):
LGTM

PR #44146, opened 2022-07-07
[Auto Parallel] Add dist op cost

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

This pr will add dist op cost of cost model in auto parallel. Distributed op cost can be got by comp and comm op cost.


JZ-LIANG (approved):
LGTM

PR #44148, opened 2022-07-07
[Auto Parallel] Add comm init control by socket

Caozhou1995:
### PR types

Others
### PR changes

Others
### Describe

This PR controls comm init rank in auto parallel to avoid hang when init process group


JZ-LIANG (approved):
LGTM

PR #44186, opened 2022-07-08
Add pool avg to quantization and concat scales correction

wozna:
### PR types

Performance optimization
### PR changes

Others 
### Describe

This PR:
- adds average pool2d to QAT OneDNN quantization process, because 3 years ago there was some problems with scale calculation for this, but it is not valid now.  
- corrects concat scale propagation, so when after concat we know output scale we can rewrite it to concat's inputs.  

Both changes will improve performance about 5% in INT8 Picodet model.


lidanqing-vv (approved):
LGTM

lidanqing-vv:
@yaomichael  This PR speedup picodet INT8 by 5% as Joanna commented.

sfraczek (approved):
LGTM

PR #44273, opened 2022-07-12
[Dy2St]Support zero dims input for eager run program OP

0x45f:
### PR types

Others
### PR changes

Others
### Describe

[Dy2St]Support zero dims input for eager run program OP


Aurelius84 (approved):
LGTM

PR #44277, opened 2022-07-12
add necessary dependency for fwd_func 

betterpig:
### PR types

Others
### PR changes

Others
### Describe

add necessary dependency for fwd_func to avoid compiling error of random parallel compiling.
Pasted Graphic 1

- fused_attention_fwd_func.cc's header amp_auto_cast.h depend cast_dygraph_function identifier, which is declared in dygraph_forward_api.h. 
- dygraph_forward_api.h will be generated actually only after copy_dygraph_forward_functions target built.
image

- So fused_attention_fwd_func should depend on copy_dygraph_forward_functions.


zhwesky2010 (approved):
LGTM

PR #44278, opened 2022-07-12
Fix var duplication bug for graph_to_program_pass

From00:
### PR types
Others

### PR changes
Others

### Describe

Fix var duplication bug for graph_to_program_pass


zhiqiu (approved):
LGTM

PR #44340, opened 2022-07-14
Fix run inference bug for standalone executor

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

Fix run inference bug for standalone executor.


pangyoki (approved):
LGTM

PR #44372, opened 2022-07-15
[API/OP] Set random seed for unittest of lstsq api/op

haohongxiang:
### PR types

Bug fixes
### PR changes

Others 
### Describe

[API/OP] Set random seed for unittest of lstsq api/op


ForFishes (approved):
LGTM

PR #44396, opened 2022-07-18
fix duplicate slice logic in _grad

cxxly:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix duplicate slice logic in _grad.


levi131 (approved):
LGTM

PR #44475, opened 2022-07-20
[MLU] transpose avg_pool2d to NHWC for better performance.

ShawnNew:
### PR types

Performance optimization

### PR changes

OPs

### Describe

transpose input of avg_pool2d for better cnnl kernel performance


ronny1996 (approved):
LGTM

PR #44485, opened 2022-07-20
[Dy2Sta]Fix Segment Fault while training multi-card if params have no grad

Aurelius84:
### PR types

Bug fixes
### PR changes

OPs
### Describe

[Dy2Sta]Fix Segment Fault while training multi-card if params have no grad


JiabinYang (approved):
LGTM

PR #44499, opened 2022-07-21
Add a feed op before each input parameter var.

wzzju:
### PR types

Bug fixes
### PR changes

Others
### Describe

Add a feed op before each input parameter var .


thisjiang (approved):
LGTM

PR #44525, opened 2022-07-21
Fix for breaking standalone PyInstaller builds

vethan:
### PR types
Bug fixes

### PR changes
APIs

### Describe
Currently when importing paddle in a standalone EXE built using pyinstaller, the build will crash as the interpreter path generated in this fix will not resolve to an existing file.  It should probably have been wrapped in a try/except in the first place.


luotao1:
@vethan Sorry for too late to review the PR, could you resolve the conflicts?

luotao1:
close due to the following PR is merged
- https://github.com/PaddlePaddle/Paddle/pull/51235

PR #44549, opened 2022-07-22
skip dist test cases if mlu card number only one, test=develop

qili93:
### PR types

Others

### PR changes

Others

### Describe

Update MLU parallel test CI script to avoid running multi-devices test cases with single MLU environment.


Aganlengzi (approved):
LGTM

PR #44640, opened 2022-07-26
[WIP] Matmul v1 & v2 unification -- part 1

jczaja:
### PR types
Performance optimization

### PR changes
OPs

### Describe
This unifies kernel for matmul and matmul_v2 fp32 and bf16 types (int8 will come in next PR). Major reason is that matmul mkldnn kernel is underperforming for bigger batches and more threads where matmul_v2 should reach better performance for multi-threaded scenario.


lidanqing-vv (approved):
LGTM

PR #44677, opened 2022-07-27
OneDNN md-in-tensor refactoring part 6: Memory descriptor enabled for conv and conv_transpose

jakpiase:
### PR types
New features

### PR changes
OPs

### Describe
Memory descriptor support enabled for conv and conv_transpose ops


paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -1078,7 +1041,6 @@ class ConvMKLDNNGradOpKernel : public framework::OpKernel<T> {
 
       auto conv_bwd_weights_p = handler.AcquireBackwardWeightsPrimitive();
 
-      // TODO(grygielski) why no bias_diff?
```
jakpiase, line 1081: Note: No bias_diff, because PaddlePaddle's convolution is not using Bias at all

paddle/fluid/operators/mkldnn/conv_mkldnn_op.cc
```diff
@@ -24,13 +24,13 @@ namespace paddle {
...
-                                           const int groups,
+inline MKLDNNMemoryFormat GetWeightsFormat(const int groups,
                                            const bool is_conv3d) {
   if (is_conv3d) {
-    return (groups == 1) ? format : MKLDNNMemoryFormat::goidhw;
+    return (groups == 1) ? MKLDNNMemoryFormat::oidhw
+                         : MKLDNNMemoryFormat::goidhw;
   } else {
-    return (groups == 1) ? format : MKLDNNMemoryFormat::goihw;
+    return (groups == 1) ? MKLDNNMemoryFormat::oihw : MKLDNNMemoryFormat::goihw;
```
jczaja, line 33: why don't we take mem_desc from tensor here?

jakpiase: Only "X" and "Out" tensor's memory descriptors are being set using our `set_mem_desc()` function. Weights, biases and all other tensors are "static"(no matter of the input data, its memory layout will always stay the same), and their memory layout is forced by the PaddlePaddle framework(it is always described in XXX_op.cc what is the real layout of the memory). If I will use `weights->mem_desc()` instead of creating memory descriptor and setting memory format tag manually, then its `mem_desc_` variable will be empty at the beginning and will be constructed in runtime using our deprecated `GetMKLDNNFormat()` function, which is not the desired behavior

jczaja (approved):

jczaja:

PR #44685, opened 2022-07-28
[AutoParallel] resnet with data parallel

zhaoyinglia:
### PR types

Function optimization
### PR changes

Others
### Describe

- tiny modify for resnet50 with data parallel


JZ-LIANG (approved):
LGTM

PR #44693, opened 2022-07-28
Skip CUDA Graph case for standalone executor

From00:
### PR types
Others

### PR changes
Others

### Describe

Skip CUDA Graph case for standalone executor


pangyoki (approved):
LGTM

PR #44705, opened 2022-07-28
[Eager] fix lerp grad kernel logic

veyron95:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix lerp grad kernel logic


JiabinYang (approved):
LGTM

PR #44718, opened 2022-07-29
Fix all_gather_object to support various length object

LiYuRio:
### PR types

Bug fixes

### PR changes

Others

### Describe

Fix all_gather_object for we don't consider the length of object may not equal, this will cause hang while doing all_gather. So make the ranks get all the object length first, find the max one, resize the tensor and do real all_gather communication.

支持all_gather_object时没有考虑不同进程间序列化后的object长度不一致的问题，因为通信原语要求每个进程发送缓冲区大小一致，否则会hang住，所以在实际进程all_gather之前，先同步序列化后object的长度，并对发送的tensor进行扩充。


gongweibao (approved):
LGTM

PR #44764, opened 2022-08-01
fix ut new_group_api

kuizhiqing:
### PR types

Bug fixes

### PR changes

Others

### Describe

The meaning of *use_calc_stream* is changed, user should *sync/wait* manually if using *use_calc_stream=False*.


ForFishes (approved):
LGTM

PR #44781, opened 2022-08-01
[Dy2St]Raise TypeError when call to_static to convert a method of a common class

0x45f:
### PR types

Others
### PR changes

Others
### Describe

[Dy2St]Fix to_static error when call to_static to convert a method of a common class


Aurelius84 (approved):
LGTM

PR #44795, opened 2022-08-01
Skip inplace for coalesce_tensor_op outputs

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

coalesce_tensor_op outputs a FusedOutput Tensor and a list of Output Tensors which is sliced from FusedOutput. These outputs should not be the outvar of the in-place var-pair since memory reuse between FusedOutput and Output Tensors is assumed. For the following example:
```
fused_var, var1, var2, var3 = coalesce_tensor(var1, var2, var3)
var1 = sum(var4, var5)
...

```
After running coalesce_tensor_op, var1 is assumed to share the buffer slices from fused_var. However, if sum_op is in-place, then the var1 would re-share the buffer with var4 instead of fused_var.

This PR skip in-place for the output Tensors of coalesce_tensor_op, and fix the released bugs in test_eager_deletion_padding_rnn for standalone executor.

### DIFF
--- a/paddle/fluid/framework/new_executor/interpretercore.cc
+++ b/paddle/fluid/framework/new_executor/interpretercore.cc
@@ -284,12 +284,11 @@ void InterpreterCore::BuildAndCacheInstructionCtx(Instruction* instr_node) {
 }
 
 void InterpreterCore::BuildInplace() {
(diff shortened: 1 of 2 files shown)


pangyoki (approved):
LGTM

PR #44805, opened 2022-08-02
when not compile with dist, abort all dist operations

wuhuachaocoding:
### PR types

Others
### PR changes

Others
### Describe

when not compile with dist, abort all dist operations for users.

### DIFF
--- a/python/paddle/distributed/parallel.py
+++ b/python/paddle/distributed/parallel.py
@@ -161,6 +161,10 @@ def init_parallel_env():
             "Currently not a parallel execution environment, `paddle.distributed.init_parallel_env` will not do anything."
         )
         return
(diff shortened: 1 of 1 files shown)


python/paddle/distributed/parallel.py
```diff
@@ -161,6 +161,10 @@ def train():
             "Currently not a parallel execution environment, `paddle.distributed.init_parallel_env` will not do anything."
         )
         return
+    #if not compiled with distributed, abort the following operation.
+    assert core.is_compiled_with_dist(
```
gongweibao, line 165: Put to line 152?

gongweibao (changes requested):

PR #44815, opened 2022-08-02
[Auto Parallel] Unify gradient synchronization procedure of data parallel

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Describe
before PR: every dist op has it own version gradient synchronization.
after PR: unified gradient synchronization function implemented in auto_parallel/operators/common, and any dist op should call this unified version.


zhiqiu (approved):
LGTM for core.ops

aoyulong (approved):

PR #44816, opened 2022-08-02
[DCU] fix hipDeviceAttributeManagedMemory not support on DTK, test=de…

qili93:
### PR types

Bug fixes

### PR changes

Others

### Describe

Save the following code as test.hip and run `hipcc test.hip` and then run `./a.out` to verify if the platform supports or not .

- ROCM 4.0.1 - work but return false
- DTK 22.04 - not support
- DTK 21.04 - not support
- DTK 22.04 - not support

```c++
#include <hip/hip_runtime.h>
#include <iostream>

int main(int argc, char* argv[])
{
  int ManagedMemoryAttr;
  hipError_t err = hipDeviceGetAttribute(&ManagedMemoryAttr, hipDeviceAttributeManagedMemory, 0);
  if (err != hipSuccess) {
    std::cout << "ERROR: HIP failed at line: " << __LINE__  <<  " with hip error " << hipGetErrorString(err) << std::endl;
  }
  std::cout << "ManagedMemoryAttr=<" << ManagedMemoryAttr << ">" << std::endl;
  if (!ManagedMemoryAttr) {
    std::cout << "Managed memory access not supported on the device" << std::endl;
  }

  return 0;
}
```

Fail on DTK 21.10
image

Fail on DTK 22.04
image


ronny1996 (approved):
LGTM

PR #44867, opened 2022-08-03
Fix reorder bug in Conv MKLDNN

yeliang2258:
### PR types

Bug fixes

### PR changes

OPs

### Describe

Fix reorder bug in conv MKLDNN


jczaja:
@yeliang2258 Please update your PR to latest codebase . We would like to evaluate it by our CI to check impact on all models we trace which will take around two days since conflicts are solved.

yeliang2258:
> @yeliang2258 Please update your PR to latest codebase . We would like to evaluate it by our CI to check impact on all models we trace which will take around two days since conflicts are solved.

Done

yeliang2258:
@jczaja Hello, I want to know, how can I print out the specific data in the onednnl memory

Silv3S:
Hi @yeliang2258,
currently fuses with `elementwise_add` require both elementwise_addends to have same dimensions. That's because `append_sum` combined with `ShareDataWith` accumulates both of these tensors in one memory buffer. So if they're not equal, probably some random data is treated as output from convolution.

I reproduced error offline, and it can be fixed by disabling fuse pass if dimensions don't match. Please see changes on my branch https://github.com/Silv3S/Paddle/tree/fix_elem_add_fuses.

Our new strategy for implementing fuse passes with elementwise_add is to use `append_binary` post-op, which supports broadcasting.

yeliang2258:
@Silv3S Thank you for your reply. The solution in your branch does work, but when the shape is different, what is the problem with the code I submitted?   I create a new memory object and return the calculated data, I don't understand why the result is wrong in this case

yeliang2258:
@Silv3S Hello, I had a user problem, same as the one here, I solved it using the solution you provided, and I made a PR to fix conv + elementwise_add merge pass first.
https://github.com/PaddlePaddle/Paddle/pull/45018

PR #44882, opened 2022-08-04
[Auto Parallel] Data Parallel Optimization Pass 1

JZ-LIANG:
### PR types
Performance optimization

### PR changes
Others

### Describe
In traditional Data parallel, system constraint all layers have the same data parallel group, therefore  all layers have the same data parallel degree.   Several optimization  for data parallel  are based on the constraint above, like:
* loss sacling: scaling the initial_backward_grad instead of each grad
* fuse allreduce 

BUT the Layer-wise Parallelism property in Auto Parallel break the **above constraint**, and therefore the above optimization could not be applied in Auto Parallel.  We try to generalize those optimization for Auto Parallel, and  make new optimization given the  Layer-wise Parallelism scenario. 

This is the first PR that generalize the  loss sacling optimization for data parallel.


aoyulong (approved):
LGTM

PR #44870, opened 2022-08-03
Merge matmul_v1 and matmul_v2 fuse passes

Silv3S:
### PR types
Others

### PR changes
OPs

### Describe
- merge two fuse passes into one as they share common code,
- fix op_compat,
- adjust unit tests.

PR #44941, opened 2022-08-05
Offload calculations from matmul op to fuse pass

Silv3S:
### PR types
Performance optimization

### PR changes
OPs

### Describe
- continuation of work started in https://github.com/PaddlePaddle/Paddle/pull/44870,
- add all necessary checks inside `matmul_transpose_reshape` and `reshape_transpose_matmul` passes,
- remove calculations from kernels.


Silv3S:
In this PR I made changes to matmul mkldnn kernel, so it uses dynamic attributes to reduce calculations. Unfortunately it can't be tested via existing tests from `test_matmul_mkldnn_op` as these unit tests are tailored to test operator itself, not fuse passes. By that I mean, it's not possible to set dynamic attributes in these tests. It was working before this PR, because we utilized only attributes set in OPMaker.

Please let me know if these cases from `test_matmul_mkldnn_op` may be removed as they are already covered in:
- `test_mkldnn_reshape_transpose_matmul_fuse_pass`,
- `test_mkldnn_reshape_transpose_matmul_v2_fuse_pass`,
- `test_mkldnn_matmul_transpose_reshape_fuse_pass`,
- `test_mkldnn_matmul_v2_transpose_reshape_fuse_pass`.

jczaja (approved):
LGTM

paddle/phi/core/ddim.cc
```diff
@@ -172,10 +172,13 @@ DDim stride_numel(const DDim& ddim) {
 }
 
 DDim DDim::reshape(std::vector<int>& shape) const {
```
jakpiase, line 174: I am not sure about these changes to this file. I see what you did here, but after your changes errors associated with incorrect dimensions will result in something like segfault while earlier we were getting framework-handled errors with nice description, what do you think about that?

Silv3S: Both `reshape()` and `transpose()` in ddim.cc were added for purpose of matmul fuse passes. As we check dimensions with PADDLE_ENFORCES during fusion, there is no possibility that values here could be incompatible.

To avoid possible misunderstandings, maybe we should move these functions outside phi/core to some mkldnn_helper function?

jakpiase: Moving it outside phi/core is a great idea! Except from that little issue, LGTM! Really good work(as always!)

Silv3S:
@jiangjiajun please review

PR #44971, opened 2022-08-08
Fix memory leak in processgroup

ForFishes:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix memory leak bug in processgroup.

- The input tensor during communication must be destructed and released after the communication is completed. only support auto_growth allocation.
- Avoid tasks from increasing the reference count of the input.

After the PR, for example, in the following code, the peak memory is 2400000000 * 2 bytes, not 2400000000 bytes. Although the resources of a are released immediately after the communication, because the communication is not completed asynchronously, the next allocation will re-allocate new  memory.

```python
import paddle
from paddle.distributed import init_parallel_env
import time
paddle.seed(1024)
init_parallel_env()
paddle.set_device('gpu:%d'%paddle.distributed.ParallelEnv().dev_id)
a = paddle.rand(shape=[20000, 30000], dtype="float32")

t = dist.all_reduce(a, use_calc_stream=False)
a._clear()
print(t.is_completed())  # False
b = paddle.rand(shape=[20000, 30000], dtype="float32")

time.sleep(5)
print(">> finish")
```

### DIFF
--- a/paddle/fluid/distributed/collective/ProcessGroupHCCL.cc
+++ b/paddle/fluid/distributed/collective/ProcessGroupHCCL.cc
@@ -197,17 +197,6 @@ std::shared_ptr<ProcessGroup::Task> ProcessGroupHCCL::Collective(
   SyncDefaultStream(places, places_to_events_[key], places_to_ctx_[key]);
 
   auto task = CreateTask(places, rank_, op_type, inputs);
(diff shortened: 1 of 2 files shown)


gongweibao (approved):
LGTM first.
We should fix the allocator method in the future.

PR #45001, opened 2022-08-09
[mag240m]Fix tensor shape bug, fix sample bug

DesmonDay:
### PR types
Bug fixes

### PR changes
Others

### Describe
1. Fix tensor shape bug, the dim's dtype should be int64_t.
2. Fix graph sample bug. When the input node is larger than the num_nodes of sample graphs, we should handle it.

### DIFF
--- a/paddle/phi/kernels/gpu/graph_sample_neighbors_kernel.cu
+++ b/paddle/phi/kernels/gpu/graph_sample_neighbors_kernel.cu
@@ -37,9 +37,13 @@ namespace phi {
 template <typename T>
 struct DegreeFunctor {
   const T* col_ptr;
-  HOSTDEVICE explicit inline DegreeFunctor(const T* x) { this->col_ptr = x; }
+  int64_t len_col_ptr;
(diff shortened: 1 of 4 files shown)


paddle/phi/kernels/gpu/graph_sample_neighbors_kernel.cu
```diff
@@ -88,6 +93,10 @@ __global__ void SampleKernel(const uint64_t rand_seed,
 
   while (out_row < last_row) {
     T node = nodes[out_row];
+    if (node > len_col_ptr - 1) {
+      out_row += BLOCK_WARPS;
+      continue;
+    }
```
ZHUI, line 99: `node > len_col_ptr - 1` 的逻辑是？

ZHUI (approved):
LGTM

paddle/phi/kernels/gpu/graph_sample_neighbors_kernel.cu
ZHUI: 采样的节点编号，超图节点

PR #45002, opened 2022-08-09
[Auto Parallel] Update reshard for auto search

Caozhou1995:
### PR types

Others
### PR changes

Others
### Describe

This PR updates reshard module for auto search, including bool data send/recv, lod tensor reset and so on


JZ-LIANG (approved):
LGTM

PR #45008, opened 2022-08-09
Fix mkldnn interpolate ops

yeliang2258:
### PR types

Bug fixes

### PR changes

Others

### Describe

When the interpolate OP only specifies the output dim but does not specify the scale, the value of the scale is illegally accessed, resulting in an incorrect output shape.
image

### DIFF
--- a/paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
@@ -105,7 +105,8 @@ class InterpolateMKLDNNKernel : public framework::OpKernel<T> {
           }
         }
       }
-      if (scale[0] > 0.0f && scale[1] > 0.0f && scale[2] > 0.0f) {
+      if (scale.size() == 3 && scale[0] > 0.0f && scale[1] > 0.0f &&
+          scale[2] > 0.0f) {
         int j = 0;
         std::vector<int64_t> in_dhw_vec = phi::vectorize(in_dhw_dims);
         std::transform(


yeliang2258:
@jczaja Please help me to review this PR, thanks.

paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
```diff
@@ -105,7 +105,8 @@ class InterpolateMKLDNNKernel : public framework::OpKernel<T> {
           }
         }
       }
-      if (scale[0] > 0.0f && scale[1] > 0.0f && scale[2] > 0.0f) {
+      if (scale.size() == 3 && scale[0] > 0.0f && scale[1] > 0.0f &&
+          scale[2] > 0.0f) {
```
jiangjiajun, line 109: scale最多只能为3是吗

yeliang2258: 是的

yeliang2258:
@Silv3S Please help me to review this PR, thanks.

paddle/fluid/operators/mkldnn/interpolate_mkldnn_op.cc
lidanqing-vv: Is there possibility that scale is less than 3? If its size is always 3, maybe we should set its size when we define the vector on line 86. Because I observe three places using `scale.resize(3,xx)`, `scale.reserve(3)`

yeliang2258: It is indeed possible that it is less than 3, and it may be 0, so I added a judgment here, and only make the subsequent judgment when it is equal to 3

lidanqing-vv (approved):
LGTM. Thank you for the fix : )

PR #45017, opened 2022-08-09
fix proto consistency bug

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

In Paddle, there are three copies of a Program/Block/Op/Var, i.e, Python side, C++ side, and Proto side.

However, there is a bug that proto side is not consistent with the C++ side. For example, updating C++ side OpDesc will not change the proto as below.

```
def build_program(self):
    main_program = paddle.static.Program()
    startuo_program = paddle.static.Program()
    with paddle.utils.unique_name.guard():
        with paddle.static.program_guard(main_program, startuo_program):
            x = paddle.static.data(name='x', shape=[3, 2, 1])
            out = paddle.static.nn.fc(x=x, size=1, num_flatten_dims=2)
    return main_program


class TestProgramProto(unittest.TestCase):

    def update_op(self):
        program = build_program()
        a = program.desc.serialize_to_string()
        program.current_block().ops[0]._set_attr('use_mkldnn', True)
        b = program.desc.serialize_to_string()
        self.assertFalse(a == b)  # before pr, it is failed

    def update_var(self):
        program = build_program()
        a = program.desc.serialize_to_string()
        program.current_block().vars("x").desc.set_stop_gradient(False)
        b = program.desc.serialize_to_string()  
        self.assertFalse(a == b)  # before pr, it is failed
```

### DIFF
--- a/paddle/fluid/framework/block_desc.cc
+++ b/paddle/fluid/framework/block_desc.cc
@@ -179,15 +179,18 @@ void BlockDesc::Flush() {
   for (auto &op_desc : ops_) {
     op_desc->Flush();
   }
+  // no flush for var_desc? or is op_desc flush really needed?
 
   if (NeedUpdate(true)) {
     this->desc_->mutable_ops()->Clear();
     for (auto &op_desc : ops_) {
       this->desc_->mutable_ops()->Add()->CopyFrom(*op_desc->Proto());
+      // op_desc's need_update is set to false in op_desc->Flush();
     }
     this->desc_->mutable_vars()->Clear();
     for (auto &var_desc : vars_) {
       this->desc_->mutable_vars()->Add()->CopyFrom(*var_desc.second->Proto());
+      var_desc.second->SetNeedUpdate(false);
     }
     need_update_ = false;
   }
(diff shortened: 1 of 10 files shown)


paddle/fluid/framework/block_desc.cc
```diff
@@ -176,19 +176,48 @@ std::vector<OpDesc *> BlockDesc::AllOps() const {
...
+
     for (auto &var_desc : vars_) {
-      this->desc_->mutable_vars()->Add()->CopyFrom(*var_desc.second->Proto());
+      if (var_names_set.count(var_desc.first) != 1) {
+        this->desc_->mutable_vars()->Add()->CopyFrom(*var_desc.second->Proto());
+        var_desc.second->SetNeedUpdate(false);
+      }
     }
+
+    // this->desc_->mutable_vars()->Clear();
```
chenwhql, line 216: 注释后续移除下？

chenwhql (approved):
LGTM

paddle/fluid/framework/block_desc.cc
zhiqiu: 嗯嗯

PR #45030, opened 2022-08-10
Fix problem that the shape of tensor is not inited correctly when backward in static graph

feifei-111:
### PR types

Others
### PR changes

Others

### Describe

Fix problem that the shape of tensor is not inited correctly when backward in static graph


Aurelius84 (approved):
LGTM

PR #45077, opened 2022-08-11
Add int8 support for matmul+elementwise_add fuse pass

wozna:
### PR types

Performance optimization
### PR changes

OPs 
### Describe

This PR add int8 support for `matmul_elementwise_add_mkldnn_fuse_pass`. 
Moreover I unified the order of passes in quantization with and without serialization, following the order of passes for use_mkldnn.


paddle/fluid/framework/ir/graph_pattern_detector.cc
```diff
@@ -1892,11 +1892,21 @@ PDNode *patterns::Reshape2Matmul::operator()() {
...
   auto prev_op_y = pattern->NewNode(prev_op_y_repr())->assert_is_op();
 
   auto matmul_op = pattern->NewNode(matmul_op_repr())->assert_is_op("matmul");
+
+  if (!with_residual) {
+    matmul_op->assert_more([&](Node *x) {
+      if (!HasInput(x, "ResidualData") ||
+          x->Op()->Input("ResidualData").size() == 0)
+        return true;
+      return false;
```
Silv3S:
```suggestion
      return (!HasInput(x, "ResidualData") ||
          x->Op()->Input("ResidualData").size() == 0)
```

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -780,6 +783,36 @@ void CPUQuantizePass::QuantizeMatmul(Graph* graph) const {
...
+                    "ResidualData",
+                    residual_scale,
+                    is_residual_unsigned,
+                    "Scale_in_eltwise");
+
+    } else {
+      if (!AreScalesPresentForNodes({matmul_in_x, matmul_in_y})) {
+        MarkAndLogCannotQuantizeOp(matmul_op,
+                                   "No scale available for the operator");
+        return;
```
Silv3S: This else block is common for both paths, so it could be checked before `if(with_residual)`. Then in if block you would check only if scales are present for `matmul_residual_data` node

Silv3S (dismissed):
LGTM :+1:

python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
```diff
@@ -403,11 +403,15 @@ def _remove_ctrl_vars(self, graph):
...
+                                 ['mkldnn_enabled_op_types'], [set()])
+        # remove dropout ops
+        graph = self._apply_pass(graph, 'simplify_with_basic_ops_pass')
         graph = self._apply_pass(graph, 'layer_norm_fuse_pass')
         graph = self._apply_pass(graph, 'attention_lstm_fuse_pass')
         graph = self._apply_pass(graph, 'seqconv_eltadd_relu_fuse_pass')
         #  graph = self._apply_pass(graph, 'seqpool_concat_fuse_pass')
-        graph = self._apply_pass(graph, 'seqpool_cvm_concat_fuse_pass')
+        #  graph = self._apply_pass(graph, 'seqpool_cvm_concat_fuse_pass')
         #  graph = self._apply_pass(graph, 'embedding_fc_lstm_fuse_pass')
```
tsocha: Could you remove this commented code?

tsocha (commented):
I can see that in many places there is `if(residual)` maybe it is worth to break this pass into `normal` version and `residual` version?

If it will be done properly it should reduce risk of bugs in the future development. Readability of the code can be higher as well.

Silv3S (approved):

jczaja (approved):
LGTM

wozna:
@jiangjiajun  Please start your review

PR #45091, opened 2022-08-11
Added concat workaround for vivo model

jakpiase:
### PR types
Others

### PR changes
OPs

### Describe
Added concat workaround for vivo model. Workaround is enforcing plain memory format tag if first src dimensions are having "1" inside of it, because oneDNN was chosing suboptimal format tag which was the cause of really bad concat performance.


yeliang2258 (approved):
LGTM

jczaja (approved):
LGTM

PR #45105, opened 2022-08-12
[ Dy2Static | Controlflow ]While + Cond support for python container.

2742195759:
### PR types

Others
### PR changes

Others
### Describe

While support for python container. It is convenient to convert more dynamic graph codes into static graphs.

support the following case:
```
for i in tensor:
    a = {'a': 1, 'b': 2, 'c': 3}
```
and 
```
  a = {"a": 3, "b": 4, "c": 9}                                                                                                                                                                                    
  for i in x:                                                                                           
      a = a  # make the a become loop vars. a will be convert into {'a', Tensor, 'b': Tensor, 'c': Tensor}                                                                            
      a['a'] = 12   
```


Aurelius84 (approved):
LGTM

PR #45107, opened 2022-08-12
Add strongly typed functions to set attributes to avoid unexpected type conversions.

iclementine:
### PR types

Performance optimization

### PR changes

Others

### Describe

Add strongly typed functions to set attributes to avoid unexpected type conversions.

Paddle's API to construct Operator has an effect of building an OpDesc. During the contruction of an OpDesc, various attributes are stored in the OpDesc via a function `SetAttr`, this function is bound to python via pybind11. However, it has an parameter with Attr, which is a variant representing sevral possible types in paddle as operator attributes.

https://github.com/PaddlePaddle/Paddle/blob/34234282a65eab864460a476c9380f1d22fa4126/paddle/fluid/framework/op_desc.cc#L620
https://github.com/PaddlePaddle/Paddle/blob/34234282a65eab864460a476c9380f1d22fa4126/paddle/fluid/pybind/protobuf.cc#L361

It is okay to do so inside a single language, but between python and c++, there may be some unexpected type conversions. According to pybind11, type conversions for builtin primitive types like int, float from python to c++ may be a bit strange, because python types and c++ types do not have a one-to-one mapping. There is no straight way to bring python variable as-is to c++.

https://pybind11.readthedocs.io/en/stable/advanced/functions.html#overload-resolution-order
https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#c-17-library-containers

Also, python does not natively handle overloaded function. Variant parameter in c++ functions bound to python has a similar effect as overload resolution. The argument is matched against each types in the variant from begining to end, once it is compatible with a type, the argument is converted to that type withou considering types after that. As a result, it adds extra semantics in the ordering of the types in a variant. It may not be a good idea to bind function with variant as parameters to python.

So we explicitly define several functions for different types to avoid unexpected type conversions. This will allow us to add double, complex, and vector of complex as possible attribute types.

### DIFF
--- a/paddle/fluid/pybind/protobuf.cc
+++ b/paddle/fluid/pybind/protobuf.cc
@@ -286,6 +286,8 @@ void BindOpDesc(pybind11::module *m) {
       .value("LONGS", pd::proto::AttrType::LONGS)
       .value("FLOAT", pd::proto::AttrType::FLOAT)
       .value("FLOATS", pd::proto::AttrType::FLOATS)
+      //  .value("FLOAT64", pd::proto::AttrType::FLOAT64)
+      .value("FLOAT64S", pd::proto::AttrType::FLOAT64S)
       .value("STRING", pd::proto::AttrType::STRING)
       .value("STRINGS", pd::proto::AttrType::STRINGS)
       .value("BOOL", pd::proto::AttrType::BOOLEAN)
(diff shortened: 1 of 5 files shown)


paddle/fluid/pybind/protobuf.cc
```diff
@@ -361,6 +363,21 @@ void BindOpDesc(pybind11::module *m) {
           py::arg("with_attr_var") = false)
       .def("_set_attr", &pd::OpDesc::SetAttr)
       .def("remove_attr", &pd::OpDesc::RemoveAttr)
+      .def("_set_bool_attr", &pd::OpDesc::SetPlainAttr<bool>)
+      .def("_set_int32_attr", &pd::OpDesc::SetPlainAttr<int>)
+      .def("_set_int64_attr", &pd::OpDesc::SetPlainAttr<int64_t>)
+      .def("_set_float32_attr", &pd::OpDesc::SetPlainAttr<float>)
+      //  .def("_set_float64_attr", &pd::OpDesc::SetPlainAttr<double>)
```
chenwhql, line 370: if we uncomment here, does it work correctly?

chenwhql (approved):
LGTM

paddle/fluid/pybind/protobuf.cc
iclementine: no, it marks the support for double as a TODO. Which will be done later.

PR #45159, opened 2022-08-15
[AutoParallel] Prune D2H memcpy for fp16 pass

JZ-LIANG:
### PR types
Performance optimization

### PR changes
Others

### Describe
when applying fp16 pass, the found_inf need be transfer from GPU to CPU to determine whether or not skipping the optimization (since nan or inf found in gradient). This PR prune the num of memcpy and improve the throughput of fp16 pass.

berfore PR, there are many H2D for each param:
image

after PR, there is only one H2D for all param:
image


aoyulong (approved):
LGTM

PR #45173, opened 2022-08-16
Data Parallel Comm & Calc Overlap Optimization

JZ-LIANG:
### PR types
Performance optimization

### PR changes
Others

### Describe
Support Comm & Calc Overlap in AutoParallel. 

There two difference compared to the existed Overlap in paddle static hybrid parallel:

1. Layer-wise Parallelism adaption: the existed Overlap  constraint all layers have the same data parallel comm,  and only synchronize calc with this comm for overlapping. In Auto Parallel there would be more than one data parallel comm for different param(grad), and this PR analyze each dp communication and add synchronization properly. 

2. The existed Overlap use sync_stream for synchronization, this PR replace it with wait event to mitigate the overhead of synchronization.


aoyulong (approved):
LGTM

PR #45188, opened 2022-08-16
Optimize performance of amp

ZzSean:
### PR types

Performance optimization
### PR changes

Others
### Describe

Optimize the performance of `unscale` function when using AMP.

Unit: samples/s
| model      | before opt | after opt          | speedup
| ---------- | ---------- | ------------------ |--|
| bert_large |   51.87563         | 56.37118           | 1.09x


zhangbo9674 (approved):
LGTM

PR #45191, opened 2022-08-16
fix bug of external_error.pb.h not found

betterpig:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix bug of external_error.pb.h not found. device_mesh depends on phi_enforce but not set dependency, so phi_enforce may not compiled when device_mesh is compiling, as a result paddle/fluid/platform/external_error.pb.h has not been generated.
image


zhwesky2010 (approved):
LGTM

PR #45216, opened 2022-08-17
【bugFix】added dtype 'float16' when check_variable_and_dtype the input of softmax_cross_entropy

feifei-111:
### PR types

Others

### PR changes

Others
### Describe

softmax_cross_entropy will not accept float16 input because of dtype check when it is in static mode.
however float16 is ok in dygraph mode


2742195759 (approved):
LGTM

PR #45219, opened 2022-08-17
[LAUNCH] fix ips offset

kuizhiqing:
### PR types

Bug fixes 

### PR changes

Others

### Describe

fIx offset calculation error when using ips.


ForFishes (approved):
LGTM

PR #45225, opened 2022-08-17
[Profiler] add views in summary API

kuizhiqing:
### PR types

New features

### PR changes

APIs

### Describe

* add enum type SummaryView
* add `views` in API `summary` to define which SummaryView to show in summary API

Usage
```
from paddle.profiler import SummaryView
...
p.summary(sorted_by=profiler.SortedKeys.GPUTotal, views=[SummaryView.MemoryView, SummaryView.KernelView])
```
Only views in the list will be shown.

**NOTE: the default value of `views`  is None, which make this change compatible.**

### DIFF
--- a/python/paddle/profiler/profiler.py
+++ b/python/paddle/profiler/profiler.py
@@ -46,7 +46,8 @@ class SummaryView(Enum):
     KernelView = 4
     OperatorView = 5
     MemoryView = 6
-    UDFView = 7
+    MemoryManipulationView = 7
+    UDFView = 8
 
 
 class ProfilerState(Enum):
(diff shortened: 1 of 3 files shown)


rainyfly (dismissed):
LGTM

python/paddle/profiler/profiler.py
```diff
@@ -734,6 +749,7 @@ def summary(self,
                 sorted_by=SortedKeys.CPUTotal,
                 op_detail=True,
                 thread_sep=False,
+                views=None,
```
jeff41404: In order to ensure compatibility, the new parameter is usually added at the end of API parameters  , but the `views` is not. do we have another considerations?

python/paddle/profiler/profiler_statistic.py
```diff
@@ -698,9 +698,12 @@ def _build_table(statistic_data,
                  sorted_by=SortedKeys.CPUTotal,
                  op_detail=True,
                  thread_sep=False,
+                 views=None,
```
jeff41404: same problem as above. 
But if this function is not public API and just used by internal calls, that's OK

jeff41404 (approved):
LGTM

Ligoml (approved):
LGTM for docs

sneaxiy (approved):
LGTM

PR #45237, opened 2022-08-18
[Auto Parallel] Add grad var shape infer of matmul and mul

Caozhou1995:
### PR types

Others
### PR changes

Others
### Describe

This PR adds grad var shape infer of matmul and mul.


JZ-LIANG (approved):
LGTM

PR #45249, opened 2022-08-18
Enable OMP multithreading in lookup_table_v2

piotrekobi:
### PR types

Performance optimization
### PR changes

OPs
### Describe

This PR enables OMP multithreading on the lookup_table_v2 operator which makes the operator take less time when running the ernie 3.0 model with more than 1 thread,

The results of the speedup on SPR with FP32 datatype are as follows (Total execution time measured by the Paddle profiler. Before all the times were similar to the 1 thread time):

1 thread: 727.331
2 threads: 483.426
5 threads: 374.056

The change allows for a 2x speedup of the operator after increasing the number of threads.

### DIFF
--- a/paddle/phi/kernels/cpu/embedding_kernel.cc
+++ b/paddle/phi/kernels/cpu/embedding_kernel.cc
@@ -48,8 +48,10 @@ struct EmbeddingCPUFunctor {
     dev_ctx_.template Alloc<T>(out_);
     auto* output = out_->data<T>();
 
+#if defined _OPENMP
 #ifndef PADDLE_WITH_GPU
 #pragma omp parallel for
+#endif
 #endif
     for (int64_t i = 0; i < ids_numel; ++i) {
       if (padding_idx_ != kNoPadding && ids[i] == padding_idx_) {
--- a/paddle/phi/kernels/cpu/embedding_kernel.cc
+++ b/paddle/phi/kernels/cpu/embedding_kernel.cc
@@ -48,6 +48,7 @@ struct EmbeddingCPUFunctor {
     dev_ctx_.template Alloc<T>(out_);
     auto* output = out_->data<T>();
 
+#pragma omp parallel for
     for (int64_t i = 0; i < ids_numel; ++i) {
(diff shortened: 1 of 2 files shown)


piotrekobi:
@jczaja @tsocha Please review

paddle/phi/kernels/cpu/embedding_kernel.cc
```diff
@@ -48,6 +48,11 @@ struct EmbeddingCPUFunctor {
     dev_ctx_.template Alloc<T>(out_);
     auto* output = out_->data<T>();
 
+#if defined _OPENMP
+#ifndef PADDLE_WITH_CUDA
```
jczaja: Maybe you could reformulate this conditions like: #if defined(_OPENMP) && !defined(PADDLE_WITH_CUDA).....  perhaps it would look more readable?

paddle/phi/kernels/cpu/embedding_kernel.cc
```diff
@@ -48,6 +48,11 @@ struct EmbeddingCPUFunctor {
...
+#pragma omp parallel for
+#endif
+#endif
```
tsocha: Why do you need these ifdefs here?
```suggestion
#pragma omp parallel for
```

tsocha (changes requested):

paddle/phi/kernels/cpu/embedding_kernel.cc
piotrekobi: One CI was complaining about not being able to include the OMP library, so I'm checking if it's available. Also, the ops tests were failing on CI Coverage so I decided to disable GPU support.

jczaja (approved):

paddle/phi/kernels/cpu/embedding_kernel.cc
tsocha: sounds like CMake mess

tsocha (approved):

piotrekobi:
@jiangjiajun Please review

PR #45254, opened 2022-08-18
Simplify for generate_square_subsequent_mask

JamesLim-sy:
### PR types

Function optimization
### PR changes

APIs
### Describe

- Replace `paddle.ones()  *  -np.inf` with one api `paddle.full`


FrostML (approved):
LGTM

PR #45256, opened 2022-08-19
make tensor_util contains no cuda code

zhiqiu:
### PR types

Others
### PR changes

Others
### Describe

When compiled with CUDA/ROCM, `tensor_util.cu` is a symbolic link of `tensor_util.cc`, and it is compiled with NVCC since it contains device code. This PR remove the device code from `tensor_util.cc` so it can be compiled with GCC and be compiled with other source files(.cc) parallelly.

image

compilation time: tensor_util.cc 36s -> tensor_util.cc 20s


chenwhql (approved):
LGTM

raindrops2sea (approved):

PR #45259, opened 2022-08-19
[Eager] polish scale and pow python interface logic

veyron95:
### PR types

Function optimization
### PR changes

APIs
### Describe

polish scale and pow python interface logic


JiabinYang (approved):
LGTM

PR #45265, opened 2022-08-19
Fix random op dependency and lr_shedule bugs for standalone executor

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

Fix two bugs for standalone executor:
1. Fix bug for building random-op dependency where the sampling_id and dropout ops are not identified as random because of a typo
2. Add lr_shedule to the converted Program from the Compiled Program, the missing of lr_shedule make Transformer not converge


zhiqiu (approved):
LGTM

PR #45269, opened 2022-08-19
add variable to decide whether test inference ut and demo

betterpig:
### PR types

Others
### PR changes

Others
### Describe

add variable to decide whether test inference ut and demo.


XieYunshen (approved):
LGTM

PR #45310, opened 2022-08-22
fix roi_align_op_npu to pass the unittest

USTCKAY:
### PR types

Bug fixes
### PR changes

OPs
### Describe

the original roi_align_grad_op for npu made some changes to tensor ROIs_N5 to match the cpu grad version, but it is unnecessary and would make the unittest fail with CANN of version 504,  so I fix it.


ronny1996 (approved):
LGTM

PR #44985, opened 2022-08-08
[new-exec] set cuda device before run

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

set cuda device before run

PR #45314, opened 2022-08-22
Enable startup program for standalone executor

From00:
### PR types
Others

### PR changes
Others

### Describe

This PR enables startup program for standalone executor. Using standalone executor for startup program causes the following UTs failed, this PR fixes them by the way:
1. test_py_reader_using_executor: It manually generates var name in py_reader, says py_reader_name. The var is created in Scope while not existed in program. Standalone executor has hot fixed for auto-generated variable names like lod_tensor_blocking_queue_xxx (see interpretercore_util.cc), however, manual-generated py_reader name is not allowed. This PR updates the UT to avoid such a scenario. 
2. test_parallel_executor_mnist and test_sync_batch_norm_op: They use PE to run a program and then use standalone executor to run another program (i.e., the startup program for the next testing case). As the CUDA device ID is changed after PE ran, it is inconsistent with that in standalone executor, and therefore causes a wrong result.  PR #44985 has tried to set device before standalone executor runs, however, the setting is invalid due to the typos of macro. This PR fixes it.


pangyoki (dismissed):
LGTM

pangyoki (approved):
LGTM

zhiqiu (approved):
LGTM

PR #45335, opened 2022-08-23
print log while grad ops are used in new executor

pangyoki:
### PR types

Others

### PR changes

Others

### Describe

- 如果新执行器中执行了反向（grad）op，开GLOG_v=0时会打印`Standalone Executor is Used.`
- 新执行器执行的program中存在多个block时，会打印warning：`There are more than 1 block in program.`


From00 (approved):
LGTM

zhiqiu (approved):
LGTM

PR #45385, opened 2022-08-24
make full_like support double_max in dygraph

betterpig:
### PR types

Bug fixes
### PR changes

OPs
### Describe

use double rather than float to convert fill_value, to make full_like support double_max in dygraph

original issue: 
```
>>> x=torch.empty(2, dtype=torch.float64)
>>> torch.full_like(x, np.finfo(np.float64).max)
tensor([1.7977e+308, 1.7977e+308], dtype=torch.float64)
```

paddle.full_like return wrong result.
```
>>> x=paddle.to_tensor([1,1], dtype=paddle.float64)
>>> paddle.full_like(x, np.finfo(np.float64).max)
Tensor(shape=[2], dtype=float64, place=Place(gpu:0), stop_gradient=True,
[inf., inf.])
```

result:
image

### DIFF
--- a/paddle/fluid/pybind/eager_utils.cc
+++ b/paddle/fluid/pybind/eager_utils.cc
@@ -1289,7 +1289,7 @@ paddle::experimental::Scalar CastPyArg2Scalar(PyObject* obj,
     int64_t value = CastPyArg2Long(obj, op_type, arg_pos);
     return paddle::experimental::Scalar(value);
   } else if (PyFloat_Check(obj)) {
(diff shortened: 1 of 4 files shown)


iclementine (dismissed):
LGTM

To support double in static mode and legacy dygraph, the operator's description proto must be changed, which is documented as "incompatible change". We are working on to support double and attributes whose type is not statically known.

zhwesky2010 (approved):

PR #45390, opened 2022-08-24
Add support for double attribute

iclementine:
### PR types

New features

### PR changes

Others

### Describe

Add support for double attribute.

Double attributes are required by various operators that need finer precision or wider range of floating point numbers, especially for those operators that would somehow copy the value of its attributes to, or perform arithemetic operations with elements in the Tensor of its operands, including `fill, maked_fill, set_value, assign_value, pad, ...`.

This PR extents the type system of paddle's attribute, by adding double as a possible plain attribute type.


chenwhql (approved):
LGTM

jeff41404 (approved):
LGTM

PR #45395, opened 2022-08-24
[Auto Parallel] Dist embedding Supports Ernie BigTable with 3d input

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Describe
this PR support dist embedding IDs input to be 3 dimension which is need by Ernie BigTable model.


aoyulong (approved):
LGTM

PR #45402, opened 2022-08-25
[dygraph hybrid pp for interleave] Virtual pp stage layer split

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
Support virtual pp stage layer split.
This is one of the pre-work for interleave scheduler for pipeline.
Add a new param to pp_layer, `num_virtual_pipeline_stages`, which means virtual pipeline stage on each real pipeline stage.
As we known, for a model with 8 layers, if real pipeline stage is 2, layers `[0, 1, 2, 3]` will be assigned to the first pipeline stage and layers `[4, 5, 6, 7]` will be assigned to the second pipeline stage.
Take the same model with 8 layers as an example, if the real pipeline stage is 2 and the `num_virtual_pipeline_stages` is also set to 2, layers `[0, 1]` and layers `[4, 5]` will be assigned to the first pipeline stage, while layers `[2, 3]` and layers `[6, 7]` will be assigned to the second pipeline stage. Each real pipeline stage has two separate model chunks, which are called virtual pipeline stage.

All works for supporting interleave scheduler for pipeline:
- [x] Support virtual pipeline layer spit (https://github.com/PaddlePaddle/Paddle/pull/45402)
- [ ] Support virtual pipeline layer forward function (compatible with recompute)
- [ ] Implement interleave scheduler in pipeline_parallel
- [ ] Implement save/load function for pipeline_parallel with virtual stage
- [ ] All necessary UT relating with recompute, AMP and etc.
- [ ] Some other functions but not sure yet

### DIFF
--- a/python/paddle/distributed/fleet/meta_parallel/parallel_layers/pp_layers.py
+++ b/python/paddle/distributed/fleet/meta_parallel/parallel_layers/pp_layers.py
@@ -298,6 +298,9 @@ class PipelineLayer(Layer):
                                           start_idx + stage + 1]:
                     return stage
 
+    def get_model_chunks(self):
+        return None if self._num_virtual_pipeline_stages == 1 else self._model_chunks
+
     def _construct_shared_comm(self):
         shared_comm = {}
         if self._topo.get_dim("pipe") == 1:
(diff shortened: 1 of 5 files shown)


sneaxiy (approved):

XieYunshen (approved):
LGTM for `  set_tests_properties(
    test_parallel_dygraph_pipeline_parallel_with_virtual_stage
    PROPERTIES TIMEOUT 500)`

PR #45418, opened 2022-08-25
[API] fix bug in to_tensor which returns err dtype

feifei-111:
### PR types

others

### PR changes

others

### Describe

fixed the bug : to_tensor will return wrong dtype in static mode (if input is transformed to numpy array, its dtype should be used)


Aurelius84 (approved):
LGTM

PR #45443, opened 2022-08-26
[Auto Parallel] DP Calc-Comm Overlapping Support Weight Sharing

JZ-LIANG:
### PR types
Function optimization
### PR changes
Others

### Describe
The naive overlap implement in static hybird parallel only sync comm stream at the end of Backward phase, based on a strong assumption that all communicating gradient would NOT be used after communication in Backward phase.
        
BUT this assumption will fail for scenario like Weight-Sharing and Higher-Order Differentiation, where gradient will be involved in other calculation between data-parallel allreduce kernel submmited into comm streams and the synchronization of comm stream at the end of Backward phase.

Synchronization of  comm stream should be added according to the usage of communicating gradients to support Overlapping for Weight-Sharing and Higher-Order Differentiation.


aoyulong (approved):
LGTM

PR #45455, opened 2022-08-26
refine cmake of framework

zhiqiu:
### PR types

Others
### PR changes

Others
### Describe

There are circular dependency between `alocator->device_context->malloc->allocator`, this PR fixes that.

Delete `AllocationPtr Alloc(const platform::DeviceContext& dev_ctx, size_t size)`, use `AllocationPtr Alloc(const platform::CUDAPlace& place, size_t size, const phi::Stream& stream)` instead, so device_context is not depended on allocator.


chenwhql (approved):
LGTM

XiaoguangHu01 (approved):
LGTM

PR #45537, opened 2022-08-29
[PHI] Migrate scale kernel

Silv3S:
### PR types
Others

### PR changes
Others

### Describe
- migrate OneDNN scale kernel to PHI


Silv3S:
Hi @YuanRisheng, @chenwhql
Could you please advise on what can cause problem with unregistered kernel?

After migrating `scale` to PHI, it passes `test_scale_mkldnn` unit test, but PR-CE-Framework fails with error. _"UnavailableError: There are no kernels which are registered in the scale operator"_.
@piotrekobi has same issue with migrating reduce kernels https://github.com/PaddlePaddle/Paddle/pull/45536.

Activation and gaussian_random kernels passed all required CI's with same registration pattern.

chenwhql (approved):

PR #45577, opened 2022-08-30
fix bug that Op with id 0 can not be lauched

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix bug that Op with id 0 can not be lauched


pangyoki (approved):
LGTM

PR #45598, opened 2022-08-31
[Eager, Performance optimization] use Python-C interface directly

veyron95:
### PR types

Function optimization
### PR changes

APIs
### Describe

Function optimization


wanghuancoder (approved):
LGTM

PR #45616, opened 2022-08-31
fix bug in PostTrainingProgram for certain cases

ZhangHandi:
### PR types
Bug fixes

### PR changes
APIs

### Describe
Fix bug for dataloader type is fluid.reader.GeneratorLoader and not full quantization case

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/post_training_quantization.py
+++ b/python/paddle/fluid/contrib/slim/quantization/post_training_quantization.py
@@ -15,6 +15,7 @@
 import os
 import re
 import math
+import fluid
 import shutil
 import logging
 import numpy as np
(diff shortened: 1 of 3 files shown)


python/paddle/fluid/contrib/slim/quantization/post_training_quantization.py
```diff
@@ -18,6 +18,7 @@
 import shutil
 import logging
 import numpy as np
+import paddle.fluid as fluid
```
ceci3: 用相对路径import fluid

ZhangHandi: 已修改

ceci3 (approved):

PR #45628, opened 2022-09-01
update some input for pp and moe about recompute.

wuhuachaocoding:
### PR types

Others
### PR changes

APIs
### Describe

update some input for pp and moe about recompute.


gongweibao (approved):
LGTM

PR #45643, opened 2022-09-01
Gradient Fuse Allreduce

JZ-LIANG:
### PR types
Performance optimization

### PR changes
Others

### Describe
Inplace Gradient Coalesce for fuse allreduce.
Avoid  memory copy and temporary gradient buffer.

Bugfix for FP16 Pass initialization for param.


aoyulong (approved):
LGTM

PR #45671, opened 2022-09-01
Fc residual quantization

sfraczek:
### PR types
Performance optimization

### PR changes
OPs

### Describe
Add quantization of FC+residual fused op


sfraczek:
It's become impossible to merge after recent change. I need to rewrite it.

PR #45696, opened 2022-09-02
[ControlFlow] replace executor in run method of control flow ops with standalone_executor

kangguangli:
### PR types

Others
### PR changes

OPs
### Describe

The main content of this pr is to replace executor in run method of conditional block op with standalone_executor. The changes are mainly about two parts: `ConditionalBlockOp` and `InterpreterCore`.

1. Changes about `ConditionalBlockOp`
    1.  replace `Executor` in `RunImpl` with `InterpreterCore`
    2. add class member `std::shared_ptr<InterpreterCore> core` to cache 
    4. use flag `FLAGS_control_flow_use_new_executor` to decide whether use `InterpreterCore`
    6. no longer create new `scope` every times executing `ConditionalBlockOp::RunImpl`, only when the cached `scope` is invalid
    7.  by default, `InterpreterCore` will rebuild and reset scope. The scope that was used to initialized it may be invalid, since it may be dropped by outer executor.
8. Changes about `InterpreterCore`
    1. add parameter `used_for_control_flow_op` for the initialization of `InterpreterCore`
    1. The executor in control flow does not need return value, so add a parameter `need_fetch` in `InterpreterCore::Run`
    2. `interpretercore_utils.cc::BuildVariableScope` should put persistent variables in global scope, which is the root of the scope tree.
    3. control flow block won't contains variables which occur in outer block, so these variables should be skipped in gc.
    4. for `interpretercore_util.cc::BuildVariableMap`, some inputs or outputs of op in the control flow block is recorded in outer block, so the corresponding variables should be queried recursively. 
10. Other changes
    1. sometimes the layout of variable will be changed after first execution of executor, so there's no need to execute `transfer_layout` really. We simply call `memcpy` instead.


paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
```diff
@@ -227,7 +232,7 @@ void BuildVariableScope(const framework::BlockDesc& block,
     }
 
     if (var_desc->Persistable()) {
-      auto* ptr = inner_scope->Var(var_name);
+      auto* ptr = const_cast<Scope*>(ancestor_scope)->Var(var_name);
```
From00, line 238: Better to move the declaration and assignment of ancestor_scope into this `if` statement and add a NOTE for why we should get vars from the ancestor_scope.

zhangbo9674 (dismissed):
LGTM for interpretercore

0x45f (dismissed):

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -71,9 +112,19 @@ class ConditionalBlockOp : public ConditionalOp {
           platform::errors::PreconditionNotMet(
               "Expect Scope variable to be set in conditional_block_op, but "
               "got a null Scope variable. Please set the Scope variable."));
+      // PE call executor, cannot know which call this
```
From00: Don't understand what this NOTE means.

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -82,27 +133,62 @@ class ConditionalBlockOp : public ConditionalOp {
       platform::DontClearMKLDNNCache(dev_place);
 #endif
       auto *block = Attr<framework::BlockDesc *>("sub_block");
-      VLOG(3) << "Conditional block.idx = " << block->ID()
-              << ", scope = " << &cur_scope;
+      // framework::StandaloneExecutor exec(dev_place, *block->Program());
```
From00: Seems it is no need for this code.

paddle/fluid/framework/new_executor/interpretercore.cc
```diff
@@ -224,7 +232,7 @@ paddle::framework::FetchList InterpreterCore::Run(
 }
 
 paddle::framework::FetchList InterpreterCore::Run(
-    const std::vector<std::string>& feed_names) {
+    const std::vector<std::string>& feed_names, bool need_fetch) {
```
From00, line 235: Since there is no fetch var in the running Program of conditional_block_op, whether the need_fetch parameter may not be required?

paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
```diff
@@ -292,7 +297,8 @@ std::tuple<VariableValueMap, VariableIdMap> BuildVariableMap(
     VariableScope* var_scope,
     Scope* local_scope,
     bool allow_var_not_in_program = false,
-    bool allow_var_not_in_scope = false) {
+    bool allow_var_not_in_scope = false,
+    bool used_for_control_flow_op = false) {
```
From00: It seems that used_for_control_flow_op has the same behavior as allow_var_not_in_scope. Maybe you can just simply assign the value of used_for_control_flow_op to allow_var_not_in_scope, and It is unnecessary to add another control parameter.

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: It's a comment forgot to delete after debug, there was a vlog following it.  Will remove it

paddle/fluid/framework/new_executor/interpretercore.cc
kangguangli: For `ConditionalBlockOP` , it should return values when it does not need. This will make the variable fetched  deconstructed unexpectedly. So we need this parameter to tell executor whether it should return values.

paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
kangguangli: `used_for_control_flow_op` and `allow_var_not_in_scope ` work differently. 
`used_for_control_flow_op` work for all ops in control flow block.
`allow_var_not_in_scope` only work for the 'registered' op.
So if we registered `conditional_block_op` as `allow_var_not_in_scope`, mostly it only works for `conditional_block_op` in global block. The normal op in control flow block may still have inputs or outputs not occurred in `block.AllVars()`.  This doesn't solve the problem.

paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
kangguangli: Ok.The reason is that in principle, we should put all trainable parameters in global scope, which means the root of the scope tree.  Some cases like quantization will look up these parameters in global scope. The original implementation does not guarantee this.

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: Will remove it.

paddle/fluid/framework/new_executor/interpretercore.cc
kangguangli: Maybe we can decide by whether the block has the `fetch op`? @zhiqiu

paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
From00: It is just a difference about applicable OP, the code behavior is still all the same. You can simply change the assignment code of allow_var_not_in_scope into 
```C++
bool allow_var_not_in_scope = used_for_control_flow_op || ops_with_var_not_in_scope.count(op_type);
```
and the interface of BuildVariableMap does not need to be changed.

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -204,7 +325,8 @@ class ConditionalBlockGradOp : public ConditionalOp {
     for (size_t i = 0; i < outside_grads.size(); ++i) {
       const std::string &outside_grad_name = outside_grads[i];
       const std::string &inside_grad_name = inside_grads[i];
-      VLOG(4) << "inside_grad_name = " << inside_grad_name
+      VLOG(4) << "[assign local]"
```
zhiqiu, line 319: what's it mean?

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -161,23 +248,56 @@ class ConditionalBlockGradOp : public ConditionalOp {
...
-        exec.reset(new Executor(dev_place));
-        if (FLAGS_use_mkldnn) exec->EnableMKLDNN(pdesc);
-        ctx = exec->Prepare(pdesc, block->ID(), inside_grads, false);
+
+      if (FLAGS_control_flow_use_new_executor) {
+        std::set<std::string> skip_gc_vars(inside_grads.begin(),
+                                           inside_grads.end());
+        bool old_flag_new_executor_use_loacl_scope =
+            FLAGS_new_executor_use_local_scope;
+        FLAGS_new_executor_use_local_scope = false;
```
zhiqiu: why change the flag to False?

paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
```diff
@@ -495,7 +509,7 @@ void BuildOpFuncList(const platform::Place& place,
         BuildVariableMap(output_name_map,
                          var_scope,
                          local_scope,
-                         /*allow_var_not_in_program=*/false,
+                         allow_var_not_in_program,
```
From00: Here `allow_var_not_in_program` should be `used_for_control_flow_op`

kangguangli: You're right. And I think then I don't need to change`allow_var_not_in_program`. Just pass `used_for_control_flow_op || allow_var_not_in_program` when calling `BuildVariableMap` for inputs.

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: For `ConditionalGradBlockOp`, only if the condition meets, its block will be executed and we can get the corresponding gradient. In principle, we do not need to do anything special. However, when appending grad block, unlike forward block, paddle may add duplicate variable. So only the variables in the local scope of grad block will be evaluated and we need to copy their values to outer scope. This is what `AssignLocalGradientToParentScope` does.

I changed the log here to make it more clear because it's similar with the log in `AssignZeroToParentScope`.

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: Because there's no need to create a new scope when executing `ConditionalBlockOp`. I will change the use of flag to initialization parameters of `InterpreterCore`.

paddle/fluid/framework/new_executor/interpretercore.cc
```diff
@@ -244,11 +252,12 @@ paddle::framework::FetchList InterpreterCore::Run(
         &op_func_nodes,
         &var_scope_,
         HasLocalScope(),
-        execution_config_.used_for_jit);
-    is_build_ = true;
+        execution_config_.used_for_jit,
+        execution_config_.used_for_control_flow_op);
```
Aurelius84: 后续优化PR可以考虑这里直接传入 execution_config_（其为成员变量，似乎也不需要传递）。因为这些used_xx在函数之间传递不太好看，函数也只需要知道config即可，应该屏蔽内部配置细节

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -35,6 +37,45 @@ const char ConditionalOp::kSkipEagerDeletionVars[] = "skip_eager_deletion_vars";
...
+static void BuildScopeForConditionalBlockOp(
+    const paddle::framework::InterpreterCore &interpreter_core,
+    const paddle::framework::BlockDesc &block,
+    paddle::framework::Scope *scope) {
+  for (auto &var_desc : block.AllVars()) {
+    auto var_name = var_desc->Name();
+    if (var_name == framework::kEmptyVarName) {
+      continue;
+    }
+    VLOG(10) << "[BuildScopeForConditionalBlockOp]"
```
Aurelius84: 这个VLOG的级别似乎非常大，如果要调试的话，10以下的其他所有不关心的LOG都会输出。可以调整为3~6之间？

paddle/fluid/operators/controlflow/conditional_block_op.cc
```diff
@@ -84,25 +135,61 @@ class ConditionalBlockOp : public ConditionalOp {
...
-        auto &pdesc = *block->Program();
-        exec.reset(new Executor(dev_place));
-        if (FLAGS_use_mkldnn) exec->EnableMKLDNN(pdesc);
-        ctx = exec->Prepare(pdesc, block->ID(), skip_vars, false);
+
+      if (FLAGS_control_flow_use_new_executor) {
+        std::set<std::string> skip_gc_vars(skip_vars.begin(), skip_vars.end());
+        bool old_flag_new_executor_use_loacl_scope =
+            FLAGS_new_executor_use_local_scope;
+        FLAGS_new_executor_use_local_scope = false;
```
Aurelius84: 我不确定gflag变量读写是否是线程安全的。比如你在cond_op动态地修改了这个gflag，其他线程的其他模块（如新执行器内部）依赖了这个flag，可能会在修改后的值语义下运行。这里不一定有问题，可能需要考虑下是否存在潜在风险，框架里对flag的操作几乎都只是「读」而已

paddle/fluid/framework/new_executor/interpretercore.cc
kangguangli: 这些参数在interpreter_core内部是不需要传递的，但是调用`interpreter_util.cc`里的函数需要作为参数传递，因为它们是interpreter_core的外部函数。我稍后改一下这部分。

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: 根据和 @zhiqiu 的讨论，这里对flag的修改会被去掉，改为在interpreterCore初始化时传入参数指定。

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: Done

paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
kangguangli: Done

paddle/fluid/framework/new_executor/interpretercore.cc
kangguangli: Done

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: Done

paddle/fluid/operators/controlflow/conditional_block_op.cc
kangguangli: Done

paddle/fluid/framework/new_executor/interpreter/interpreter_util.h
```diff
@@ -75,9 +76,8 @@ void BuildOpFuncList(const platform::Place& place,
                      const std::set<std::string>& skip_gc_vars,
                      std::vector<OpFuncNode>* vec_func_list,
                      VariableScope* scope,
-                     bool use_local_scope = true,
-                     bool used_for_jit = false,
-                     bool used_for_control_flow_op = false);
+                     ExecutionConfig execution_config,
```
From00: ExecutinoConfig contains a member of skip_gc_vars set that value copying may have high overhead. Therefore, better to use const reference for parameter ExecutionConfig here.

kangguangli: Done.

zhiqiu (dismissed):
LGTM

From00 (dismissed):

Aurelius84 (approved):
LGTM

zhiqiu (approved):

lanxianghit (approved):

PR #45699, opened 2022-09-02
update for runtype and serial

sljlp:
### PR types

Bug fixes
### PR changes

Others
### Describe

1. When using py_test_module,  it should be still set that the run_type=dist when using dist tests
2.  To unset the run_serial option to speed up the tests


gongweibao (approved):
LGTM

XieYunshen (approved):
LGTM

PR #45711, opened 2022-09-02
Fix some ops which exceed int32 range when the input shape is large

sneaxiy:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Fix some ops which exceed int32 range when the input shape is large.


Xreki (approved):
LGTM

PR #45720, opened 2022-09-04
[just for my self-learning]  take some notes about sparse API 

OccupyMars2030:
### PR types

Performance optimization 
### PR changes

Others 
### Describe

change some Python C APIs to no-error-checking versions in hope of improving performance

### DIFF
--- a/paddle/fluid/pybind/eager_functions.cc
+++ b/paddle/fluid/pybind/eager_functions.cc
@@ -523,8 +523,8 @@ static PyObject* eager_api_sparse_coo_tensor(PyObject* self,
       std::dynamic_pointer_cast<phi::DenseTensor>(non_zero_indices.impl());
   auto dense_elements =
       std::dynamic_pointer_cast<phi::DenseTensor>(non_zero_elements.impl());
-  // TODO(zhangkaihuo): After creating SparseTensor, call coalesced() to sort and
-  // merge duplicate indices
+  // TODO(zhangkaihuo): After creating SparseTensor, call coalesced() to sort
+  // and merge duplicate indices
   std::shared_ptr<phi::SparseCooTensor> coo_tensor =
       std::make_shared<phi::SparseCooTensor>(
           *dense_indices, *dense_elements, phi::make_ddim(dense_shape));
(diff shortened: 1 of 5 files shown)


paddle/fluid/pybind/eager_utils.cc
```diff
@@ -305,7 +305,7 @@ std::vector<int> CastPyArg2VectorOfInt(PyObject* obj, size_t arg_pos) {
     Py_ssize_t len = PyList_Size(obj);
     PyObject* item = nullptr;
     for (Py_ssize_t i = 0; i < len; i++) {
-      item = PyList_GetItem(obj, i);
+      item = PyList_GET_ITEM(obj, i);
```
OccupyMars2030, line 308: In this place, actually, you don't have to do any error checking on "obj" and "i",  so PyList_GetItem  can be changed to  PyList_GET_ITEM  to reduce redundant error checking. You can refer to the following materials:

https://docs.python.org/3/c-api/list.html?highlight=pylist_getitem#c.PyList_GetItem
image
image
image

paddle/fluid/pybind/eager_functions.cc
```diff
@@ -523,8 +523,8 @@ static PyObject* eager_api_sparse_coo_tensor(PyObject* self,
       std::dynamic_pointer_cast<phi::DenseTensor>(non_zero_indices.impl());
   auto dense_elements =
       std::dynamic_pointer_cast<phi::DenseTensor>(non_zero_elements.impl());
-  // TODO(zhangkaihuo): After create SparseTensor, call coalesced() to sort and
-  // merge duplicate indices
+  // TODO(zhangkaihuo): After creating SparseCooTensor, call coalesced() to sort
+  // and merge duplicate indices
```
OccupyMars2030, line 527: For now, coalesced() is just used to determine if  the sparse coo tensor has been coalesced.  But  there is no corresponding implementation of how to coalesce.  Maybe we can  add  two methods:  coalesce() to actually  coalesce,  is_coalesced()  to  determine if the tensor has been coalesced.

JiabinYang (approved):
LGTM for PyList_GET_ITEM

luotao1:
LGTM，如果有时间和兴趣，可以看下其他地方的`PyList_GetItem`，是否可以替换成`PyList_GET_ITEM`

OccupyMars2030:
好，我看到会提PR

PR #45741, opened 2022-09-05
Support broadcast elementwise operators with int64 index type

sneaxiy:
### PR types
Bug fixes

### PR changes
OPs

### Describe
Support broadcast elementwise operators with int64 index type when the numel of the tensor is larger than `INT32_MAX`.

**<font color="red">NOTE: This PR makes the PaddlePaddle wheel and libpaddle_inference.so increase about 70M. The reason is that the `BroadcastKernel` is a commonly used C++ template function, and it is instantiated many times (for different data types, different operators, etc). It is inevitable so far.</font>**


FeixLiu (approved):

qingqing01 (approved):

Shixiaowei02 (approved):

tianshuo78520a (approved):

JiabinYang (approved):
LGTM

PR #45783, opened 2022-09-06
[Eager] Optimize log

JiabinYang:
### PR types

 New features
### PR changes

Others
### Describe

This PR polished loginfo：
 1. We can print each ad function's input, output info ( name, ptr, initialized) with log level 4, TensorInfo(type, dtype, place) with log level 5,  AD_info(grad, grad_node) with level 6
 2. We can print Grad node info(backward output meta size, backward input meta size) with log level 5, more info (edge(slotid, rankid)(target node, ptr)) with level 6
 3. All check log moved to level 7

Normally, we just need log level 3 to show which func run, level 4 to show the input and output info, 5 to show some backward info, 6 to show details of execution, 7 to show all info


XiaoguangHu01 (approved):
LGTM

PR #45790, opened 2022-09-06
Expose device_context from ProcessGroup

LiYuRio:
### PR types

New features

### PR changes

Others

### Describe

In some scenarios, we want to put calculation on comm stream. This will save the waiting time of cuda event. This pr expose the device_context from process group for users can get stream from device_context then.


gongweibao (approved):
LGTM

PR #45794, opened 2022-09-06
Construct exec and ctx only once in cond op to speed up

0x45f:
### PR types

Performance optimization
### PR changes

OPs
### Describe

Construct exec and ctx only once in cond op to speed up


zh794390558 (approved):
LGTM

Aurelius84 (approved):
LGTM

PR #45799, opened 2022-09-06
Use logging to print log of FLAGS_FORCE_USE_PROGRAM_CACHE

From00:
### PR types
Others

### PR changes
Others

### Describe

warnings.warn may not print log in some model (maybe because _**warnings.filterwarnings("ignore")**_ is set), this PR replaces it with logging.warning.


pangyoki (approved):
LGTM

PR #45820, opened 2022-09-07
optimization of max_pool3d forward

s5u13b:
### PR types

Performance optimization
### PR changes

OPs
### Describe

* Environment:
    * V100-32G, CUDA 11.2, cuDNN 8
* Feature：
    * replace the div and mod operation with fast_divmod operation
    * replace 1d gpu launch with 3d gpu launch
* Performance (OP Benchmark):

| Paddle Kernel | Config ID | Performance Before | Performance After | Improvement |
| :-----| :---- | :---- | :---- | :---- |
| KernelMaxPool3DWithIdx | 1 | 1103.2us | 966.28us | 14.16% |

### DIFF
--- a/paddle/phi/kernels/funcs/pooling.cu
+++ b/paddle/phi/kernels/funcs/pooling.cu
@@ -38,6 +38,24 @@ struct FastDivModForPooling {
   }
 };
 
+struct FastDivModForPooling3D {
+ public:
+  paddle::platform::FastDivMod channel;
+  paddle::platform::FastDivMod width;
+  paddle::platform::FastDivMod height;
+  paddle::platform::FastDivMod depth;
+
+  explicit HOSTDEVICE FastDivModForPooling3D(const int channels,
+                                             const int output_width,
+                                             const int output_height,
+                                             const int output_depth) {
+    channel = paddle::platform::FastDivMod(channels);
+    width = paddle::platform::FastDivMod(output_width);
+    height = paddle::platform::FastDivMod(output_height);
+    depth = paddle::platform::FastDivMod(output_depth);
+  }
+};
+
 struct FastDivModForPoolingWithMoreStaff {
  public:
   paddle::platform::FastDivMod channel;
@@ -2003,7 +2021,7 @@ template class MaxPool2dWithIndexFunctor<phi::GPUContext, double, int>;
 template class MaxPool2dWithIndexGradFunctor<phi::GPUContext, double, int>;
 
 template <typename T1, typename T2>
-__global__ void KernelMaxPool3DWithIdx(const int nthreads,
+__global__ void KernelMaxPool3DWithIdx(const int ncd,
                                        const T1* input_data,
                                        const int channels,
                                        const int input_depth,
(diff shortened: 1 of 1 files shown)


JamesLim-sy:
I think it is better to list the perf value with a table.

paddle/phi/kernels/funcs/pooling.cu
```diff
@@ -2023,57 +2041,65 @@ __global__ void KernelMaxPool3DWithIdx(const int nthreads,
...
-            ele = input_data[max_index];
+      T1 ele = -FLT_MAX;
+      int max_index = -1;
+      for (int d = dstart; d < dend; ++d) {
+        for (int h = hstart; h < hend; ++h) {
+          for (int w = wstart; w < wend; ++w) {
+            if (ele <
+                input_data_cur[(d * input_height + h) * input_width + w]) {
+              max_index = (d * input_height + h) * input_width + w;
+              ele = input_data_cur[max_index];
```
JamesLim-sy, line 2095: Formula `(d * input_height + h) * input_width + w`  has appeard twice, would it be better to  assign this formula to a local data, example below:
```
T1  cur_data = input_data_cur[(d * input_height + h) * input_width + w];
ele = ele < cur_data] ?  cur_data: ele;
```

s5u13b:
> I think it is better to list the perf value with a table.

Thanks for your advice, I have modified description to present the perf value in the form of table.

paddle/phi/kernels/funcs/pooling.cu
s5u13b: Because max_index should be assigned only when the condition `ele < input_data_cur[(d * input_height + h) * input_width + w]` is true and it would be used at the end of funciton for return mask_data, so I should assign the value to max_index inside the if statement and therefore formula `(d * input_height + h) * input_width + w` appears twice.

JamesLim-sy: Will time cost of formula  `(d * input_height + h) * input_width + w` shrink with code below ?
```
        for (int w = wstart; w < wend; ++w) {
          max_index = (d * input_height + h) * input_width + w;
          if (ele < input_data[max_index]) {
            ele = input_data[max_index];
          }
        }
```

s5u13b: The max_index cannot be calculated correctly with this codes, and consequently the mask_data return in the function would be incorrect.

JamesLim-sy: okay

paddle/phi/kernels/funcs/pooling.cu
```diff
@@ -2201,19 +2227,25 @@ class MaxPool3dWithIndexFunctor<phi::GPUContext, T1, T2> {
...
-    int thread_num = 1024;
-#ifdef WITH_NV_JETSON
-    backends::gpu::ChangeThreadNum(context, &thread_num);
-#endif
+    int ncd = batch_size * input_channels * output_depth;
 
-    int blocks = (nthreads + thread_num - 1) / thread_num;
-    dim3 threads(thread_num, 1);
-    dim3 grid(blocks, 1);
+    int thread_x = 32;
```
JamesLim-sy, line 2232: @fengxiaoshuai  hi, this PR deletes code for NV_JETSON about threads config, however the thread_num is 256 in all which is under the limitation for NV_JETSON. 
Can this PR be merged?

fengxiaoshuai: jetson系列中比较低端的GPU寄存器很少，有时候会出现资源不足，kernel launch失败，这个QA后台有CE监控

b3602sss (approved):

JamesLim-sy (approved):
LGTM

PR #45835, opened 2022-09-07
cinn_launch op: fix dtype of tensor is always mutable_data<float>

CtfGo:
### PR types

Bug fixes
### PR changes

OPs
### Describe

currently when `FLAGS_enable_pe_launch_cinn=false`, cinn_launch will set cinn_buffers used in CINN with mutable_data<float>, which maybe modify data type of  corresponding tensor, and cause error


thisjiang (approved):
LGTM

PR #45816, opened 2022-09-06
modify some typos when I  study sparse tensor

OccupyMars2030:
### PR types

Others 
### PR changes

Others 
### Describe

modify some typos

PR #45875, opened 2022-09-08
Modify typos  about sparse tensors

OccupyMars2030:
### PR types

Others 
### PR changes

Others 
### Describe

Modify typos  about sparse tensors

You can refer to https://github.com/PaddlePaddle/Paddle/pull/45816

暂时不用管这个PR，在2022年9月20日之前，我还会不间断地向这个PR提交我的各种修改


paddle/phi/common/backend.h
```diff
@@ -130,6 +130,9 @@ inline std::ostream& operator<<(std::ostream& os, Backend backend) {
       os << "IPU";
       break;
     default: {
+      // As backend is of enum class type, backend can only takes
+      // the enumerated values, so what's the point of the default
+      // clause ???
       size_t device_type_id_ = static_cast<size_t>(backend) -
                                static_cast<size_t>(Backend::NUM_BACKENDS);
       std::string device_type = phi::GetGlobalDeviceType(device_type_id_);
```
OccupyMars2030: 1. As backend is of enum class type, backend can only takes the enumerated values, so what's the point of the default
 clause ??? 
2. What's the point of `static_cast<size_t>(backend) -
                               static_cast<size_t>(Backend::NUM_BACKENDS);`  ?  And how can you ensure that the result is a  non-negative integer ?  The similar computation occurs at  line 178.

paddle/phi/common/data_type.h
```diff
@@ -153,6 +153,8 @@ PD_FOR_EACH_DATA_TYPE(PD_SPECIALIZE_CppTypeToDataType)
 
 #undef PD_SPECIALIZE_CppTypeToDataType
 
+#undef PD_FOR_EACH_DATA_TYPE
+
```
OccupyMars2030: After the macro PD_FOR_EACH_DATA_TYPE has accomplished its mission,  it seems  ok  to  #undef  it.

paddle/phi/api/yaml/generator/sparse_api_gen.py
```diff
@@ -126,12 +125,12 @@ def gen_sparse_kernel_context(self, kernel_output_names):
             elif isinstance(param, bool):
                 param = str(param).lower()
             else:
-                param + str(param) + ", "
-            kernel_context_code = kernel_context_code + f"""
+                param = str(param) + ", "
```
OccupyMars2030: It seems that "param +" should be changed to "param ="

PR #45878, opened 2022-09-08
[CustomDevice] add dy2static support

ronny1996:
### PR types

Others
### PR changes

Others
### Describe

1. add custom device dy2static support
2. add a custom device common kernel registrar and register all common kernels after init_device
3. only work in the old exeuctor, affter custom device supports new executor, it can work in the new executor 

image


qili93 (approved):
LGTM

PR #45898, opened 2022-09-08
Allow manaully set py_reader name in standalone executor

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

Some input vars of create_py_reader OP is created in Scope while not existed in program. Standalone executor has hot fixed for auto-generated variable names like lod_tensor_blocking_queue_xxx (see interpretercore_util.cc), however, manual-generated py_reader name is not allowed. This PR supports this case.


zhiqiu (approved):
LGTM

PR #41179, opened 2022-03-30
[new-exec] enable the new standalone executor by default

zhiqiu:
### PR types

New features
### PR changes

Others
### Describe

- enable standalone executor by default 

set FLAGS_USE_STANDALONE_EXECUTOR=false to use the legacy executor

PR #45913, opened 2022-09-09
Enable standalone executor for single-GPU training

From00:
### PR types
New features

### PR changes
Others

### Describe

PR #41179 has enabled Standalone Executor for Program running, this PR enabled it for Compiled Program. After this PR, all single-GPU training using old Executor or Parallel Executor (PE) are set to use Standalone Executor by default.
The supporting for distributed running is ongoing.

### DIFF
--- a/python/paddle/fluid/executor.py
+++ b/python/paddle/fluid/executor.py
@@ -1566,9 +1566,19 @@ class Executor(object):
                 compiled_program = program if isinstance(
                     program, compiler.CompiledProgram) else program._graph
 
+                # delete this code after supporting compiled_program._graph
+                if compiled_program._program is None:
+                    warnings.warn("Standalone executor is not used for Graph",
+                                  UserWarning)
+                    return use_standalone_executor_for_distribution
+
+                # delete this code after supporting distribution
                 if program._build_strategy is not None and (
                         program._build_strategy.is_distribution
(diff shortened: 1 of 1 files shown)


pangyoki (approved):
LGTM

python/paddle/fluid/executor.py
```diff
@@ -1555,23 +1555,32 @@ def _can_use_interpreter_core(program, place):
                     place, core.CustomPlace):
                 return False
 
-            use_standalone_executor_for_compiled_program = os.environ.get(
+            use_standalone_executor_for_distribution = os.environ.get(
```
zhiqiu, line 1558: NIT, why change to for distribution?

zhiqiu (approved):
LGTM

python/paddle/fluid/executor.py
From00: After this PR, the FLAGS_CONVERT_GRAPH_TO_PROGRAM is no longer control whether to run Compiled Program with Standalone Executor, but control whether to run distribution. 
It is better to change the FLAGS name at the same time. However, considering that it is just a temporary FLAGS and QAs have been set it in their testing environment, keeping the same FLAGS name can avoid the requirement for QA to reset FLAGS when they try to test the distributed scenario of Standalone Executor.

PR #45920, opened 2022-09-09
Fix quantize model deploy bugs when using MKLDNN

yeliang2258:
### PR types

Bug fixes

### PR changes

Others

### Describe

Fix ppyoloe、MobileNetV3_large、Ernie 3.0-Medium and bert acc drop bugs.

1. Avoid the problem that the weights are quantize and dequantized multiple times when multiple conv2d share the same weight. Weight names in the picodet model: conv2d_160.w_0.quantized.dequantized

2. Fix immutable op quantize bugs in MKLDNN.
Similar to the following combination, the output of shape is int32 type, if the slice is quantized, an error will be reported.
Therefore, I add a judgment,  If the input of the OP to be quantized is not float32 type, the OP will not be quantized.

image

3. Fix the bug of OP's weight dequantization such as matmul and fc in MobileNetV3_large and Ernie 3.0-Medium.

The model is: https://bj.bcebos.com/v1/paddle-slim-models/act/ppyoloe_crn_l_300e_coco_quant.tar


paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -1170,7 +1177,6 @@ void CPUQuantizePass::ApplyImpl(ir::Graph* graph) const {
   QuantizeImmutable(graph, "reshape2", "X");
   QuantizeImmutable(graph, "transpose2", "X");
   QuantizeImmutable(graph, "slice", "Input");
```
wozna, line 1184: If you added a check above, it seems to me that this shape deletion is not needed

yeliang2258: shape OP will never be quantized, so I think this line of code is redundant.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -151,16 +134,34 @@ void ParamsQuantizationMkldnnPass::QuantizeConv(ir::Graph* graph,
...
+    // If not a quantized OP
+    if (!platform::HasOpINT8DataType(conv_op->Op())) {
       return;
     }
 
-    QuantizeConvInput<int8_t>(
-        scope, g, conv_op, conv_filter->Name(), "Scale_weights");
+    auto filter_var = scope->GetVar(conv_filter->Name());
+    if (filter_var->Get<LoDTensor>().dtype() != phi::DataType::FLOAT32) {
+      VLOG(0) << "Skipping convolution filter: " << conv_filter->Name()
```
wozna: I think you can easily move these new checks to the function above QuantizeConvInput to avoid duplicating the code

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -151,16 +134,34 @@ void ParamsQuantizationMkldnnPass::QuantizeConv(ir::Graph* graph,
...
+      QuantizeConvInput<int8_t>(
+          scope, g, conv_op, conv_filter->Name(), "Scale_weights");
+    }
 
     if (HasBias(conv_op)) {
-      QuantizeConvInput<int32_t>(
-          scope, g, conv_op, conv_op->Op()->Input("Bias")[0], "Bias_scales");
+      auto bias_var = scope->GetVar(conv_op->Op()->Input("Bias")[0]);
+      if (bias_var->Get<LoDTensor>().dtype() != phi::DataType::FLOAT32) {
+        VLOG(0) << "Skipping convolution bias: "
```
wozna: Ditto

yeliang2258: Thanks, Done

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -695,6 +704,14 @@ void CPUQuantizePass::QuantizeImmutable(Graph* graph,
       return;
     }
 
+    // skip if the dtype of immutable_in is not float32
+    auto dtype = immutable_in->Var()->GetDataType();
+    if (dtype != proto::VarType::FP32) {
+      VLOG(0) << "dytpe: " << dtype;
```
jiangjiajun: 删除

jiangjiajun (changes requested):

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
yeliang2258: Done

jiangjiajun (approved):

PR #45934, opened 2022-09-09
optimization of max_pool3d grad

s5u13b:
### PR types

Performance optimization
### PR changes

OPs
### Describe

* Environment:
    * V100-32G, CUDA 11.2, cuDNN 8
* Feature：
    * Replace the div and mod operation with fast_divmod operation.
    * Replace 1d gpu launch with 3d gpu launch.
    * Optimize the computation logic of input grad accumulation. Before optimization, the gpu launch config is based on the input data, and the input grad is accumulated through traverse the coresponding index of output mask data,  which introduces the overhead of much output index computation.  After optimization, the gpu launch config is based on the output data, the input grad is directly accumulated through  looking up the max index of each output data, which saves the overhead of output index computation but requires atomic add operation.
    * (Config 0 is not optimized yet because paddle calls cudnn kernel in the config 0 of max_pool3d benchmarking.) 
* Performance (OP Benchmark):

| Paddle Kernel | Config ID | Perf Before | Perf After | Improvement | Perf of Pytorch | 
| :-----| :---- | :---- | :---- | :---- | :---- |
| cudnn::pooling_bw_5d_kernel_max | 0 | 1779.7us | - | - | 725.72us | 
| KernelMaxPool3DWithIdxGrad | 1 | 6128.1us | 677.62us | 804.3% | 725.83us |

### DIFF
--- a/paddle/phi/kernels/funcs/pooling.cu
+++ b/paddle/phi/kernels/funcs/pooling.cu
@@ -2104,87 +2104,52 @@ __global__ void KernelMaxPool3DWithIdx(const int ncd,
 }
 
 template <typename T1, typename T2>
-__global__ void KernelMaxPool3DWithIdxGrad(const int nthreads,
-                                           const T1* output_grad,
-                                           const T2* mask,
-                                           const int channels,
-                                           const int input_depth,
-                                           const int input_height,
-                                           const int input_width,
-                                           const int output_depth,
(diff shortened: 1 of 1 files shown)


JamesLim-sy (approved):
LGTM，补充下其他走相同Kernel的性能数据吧

PR #45993, opened 2022-09-13
set device id before op run

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

set device id before op run


zhangbo9674 (approved):
LGTM

PR #45991, opened 2022-09-13
Fix arm fp16 compile error

chenwhql:
### PR types

Bug fixes
### PR changes

Others 
### Describe

Fix arm fp16 compile error

PR #45997, opened 2022-09-13
fix arm compile error due to immintrin.h

YanhuiDua:
### PR types
Bug fixes

### PR changes
 Others

### Describe
Fix ARM compile error: https://github.com/PaddlePaddle/Paddle/pull/45873 adds `#include <immintrin.h>` in `paddle/phi/common/float16.h`, but ARM doesn't support `immintrin`. This PR adds the conditional compilation.

### DIFF
--- a/paddle/phi/common/float16.h
+++ b/paddle/phi/common/float16.h
@@ -14,7 +14,9 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


YanhuiDua:
https://github.com/PaddlePaddle/Paddle/pull/45991 has solved this problem

PR #46020, opened 2022-09-14
[ CherryPick ] Fix bugs when select inputs meeting different shape

2742195759:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix select_input with different shape errors:
1. select_input_with_buildin_type directly return non-undefinedvar branch when meeting undefined var
2. the output shape of select_input is inferred from inputs.


Aurelius84 (approved):
apprvoed

PR #46033, opened 2022-09-14
remove fp32 tmp var for fp16 gaussian_random

GuoxiaWang:
### PR types

Function optimization
### PR changes

APIs
### Describe

参考 PR https://github.com/PaddlePaddle/Paddle/pull/38818


wanghuancoder (approved):
LGTM

PR #46086, opened 2022-09-15
[Auto Parallel] Bugfix allreduce fuse for MP

JZ-LIANG:
### PR types
Bug fixes

### PR changes
Others

### Describe
temporary fixed for allreduce fuse for MP-DP 2D parallel.


aoyulong (approved):
LGTM

PR #46089, opened 2022-09-15
optimize device synchronization in profiler

rainyfly:
### PR types

Others
### PR changes

Others
### Describe

Currently, when start or stop profiler, it will perform synchronization for all devices, which is unnecessary and sometimes causes out-of-memory problem. Actually, we only need to synchronize current device to gaurantee all profiler data collected.


ronny1996 (approved):
LGTM

From00 (approved):
LGTM

PR #46180, opened 2022-09-19
[Auto Parallel] performance improvement for Sharding-DP hybrid parallelism

JZ-LIANG:
### PR types
Performance optimization

### PR changes
Others

### Describe
* ensure the embedding to be communicated in FP16 when AMP O2-3
* reduce the dp allreduce communication when Sharding-DP 2D paramllelism


aoyulong (approved):
LGTM

PR #46194, opened 2022-09-19
[cherry-pick] [dy2static] support user to use decorator in their program

feifei-111:
### PR types

others
### PR changes

others
### Describe

1. user should only use 'to_static' to function 'forward'
2. call decorated functions in 'forward' is allowed
3. decorator will be transformed into function call
4. support decos with args
5. fixed a bug in return_transformer

- Fixed a bug in decorator transformer， it can not analyze decorator with params correctly
error happens when the param is empty


Aurelius84 (approved):
LGTM

PR #46254, opened 2022-09-19
Reverted changes for concat workaround

jakpiase:
### PR types
Others

### PR changes
OPs

### Describe
Reverted changes for concat workaround since it caused significant drops in other models. The issue is on oneDNN side and I have made an issue on their internal repo which hopefully will fix the performance of this op without the need of this particular workaround.


jczaja (approved):
LGTM

Issue #45954, opened 2022-09-12
Adding is_test argument to PHI kernels that don't use it


piotrekobi:
During migration of some oneDNN kernels to PHI, it turned out that some kernels don't use the is_test variable as an argument. This variable is used in some kernels to make oneDNN perform less computation during inference. I migrated the pool2d kernel without using this variable https://github.com/PaddlePaddle/Paddle/pull/45775 but this resulted in a performance decrease. For now the oneDNN kernels we've found to require is_test, but are not provided its value in PHI kernels are pool2d and prelu. 
We would like to add this variable to those PHI kernels. This would change the kernel function headers as follows:
From:
```c++
template <typename T, typename Context>
void PReluKernel(const Context& dev_ctx,
                 const DenseTensor& x,
                 const DenseTensor& alpha,
                 const std::string& data_format,
                 const std::string& mode,
                 DenseTensor* out);
```
To:
```c++
template <typename T, typename Context>
void PReluKernel(const Context& dev_ctx,
                 const DenseTensor& x,
                 const DenseTensor& alpha,
                 const std::string& data_format,
                 const std::string& mode,
                 bool is_test,
                 DenseTensor* out);
```
Are we allowed to make this change in required kernels?
Is there anything else that needs to be done apart from adding the variable to the kernel function declarations and implementations?

jczaja:
@chenwhql  Please help with this one

chenwhql:
This is why we have `[Migrate later] OpKernel list`

- `Migrate later: 23 in total. This kind of Kernel has the problem of Extra parameters in OpMaker that we mentioned before. It requires more parameters than CPU&GPU Kernel, so this part of Kernel is recommended to be migrated after we complete the extra parameter cleaning.`

`is_test` is not a argument of mathematical meaning in pool2d, it only used for mkldnn, if we add it to the public kernel declaration, it will affects the kernel of all other devices,  it is not a good design.

let's also count pool2d and pool2d_grad into `[Migrate later] OpKernel list`, don't migrate them now.

Now we are building the library of custom device, if we change this declaration in framwork, we also need to change kernel implemention of npu and mlu in `PaddleCustomDeviceDevice`, like

- https://github.com/PaddlePaddle/PaddleCustomDevice/blob/develop/backends/npu/kernels/pool2d_kernel.cc

Such custom device will continue to increase in the future, for other hardware manufacturers, these extra parameters only increase the cost of understanding.

[...]

PR #46257, opened 2022-09-19
[PHI] migrate softmax_grad kernel

Silv3S:
### PR types
Others

### PR changes
Others

### Describe
- migrate OneDNN softmax_grad kernel to PHI


Silv3S:
Forward softmax kernel is using `is_test` argument, which is currently unavailable in PHI. We mark this kernel as "Migrate later" and migrate only grad kernel for now.

It's same case as in https://github.com/PaddlePaddle/Paddle/issues/45954

paulinagacek (approved):
LGTM

chenwhql (approved):

PR #46279, opened 2022-09-20
[JitLayer]Fix Load error when load path like 'export.jit'

0x45f:
### PR types

Bug fixes
### PR changes

Others
### Describe

[JitLayer]Fix Load error when load path like 'export.jit'


zh794390558 (approved):
LGTM

PR #46281, opened 2022-09-20
share threadpool of executor in dy2static

zhiqiu:
### PR types

Function optimization
### PR changes

Others
### Describe

share threadpool of executor in dy2static


zhangbo9674 (approved):
LGTM

PR #46312, opened 2022-09-20
fix error: control reaches end of non-void function

Ningsir:
### PR types

Bug fixes
### PR changes

Others
### Describe

当使用`Debug`模式编译`paddle`时，会出现如下错误:
```
error: control reaches end of non-void function [-Werror=return-type]
```
原因是因为[backward.cc](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/eager/backward.cc#L390-L391)中return后调用VLOG


luotao1 (approved):
LGTM

PR #46315, opened 2022-09-20
Limit python version to >=3.7

### LINKED ISSUE #46314: Limit Python version >= 3.7
### 问题描述 Please describe your issue

Python supports type hint from version 3.5 https://docs.python.org/3/library/typing.html. And we can write codes like this
```
def greeting(name: str) -> str:
    return 'Hello ' + name
```
which is more readable and easier to use than we use now. And Python 3.6 has an [surprisingly bug](https://stackoverflow.com/questions/56394350/how-to-resolve-this-import-error-in-python-3-6)
So will we limit the Python version to greater than 3.7?

gongweibao:
### PR types
Breaking changes

### PR changes
Others

### Describe
Fix https://github.com/PaddlePaddle/Paddle/issues/46314


luotao1:
close due to https://github.com/PaddlePaddle/Paddle/pull/46477 merged

PR #35781, opened 2021-09-15
Disabled oneDNN reshape1/2 and squeeze1/2 kernels

jakpiase:
### PR types
Bug fixes

### PR changes
OPs

### Describe
This PR disables reshape and squeeze kernels. It must be done, because I am not able to easily fix bugs with these ops.

PR #46331, opened 2022-09-20
[PHI] migrating Squeeze2 and reshape2

sfraczek:
### PR types
 Function optimization

### PR changes
OPs

### Describe
Migration of reshape2 and squeeze2 from fluid to phi.


sfraczek:
Hi @chenwhql  can you help me fix the PR-CI-Windows error?
image
I think it might be a problem with registration macro? But I don't know what is wrong.

paulinagacek (dismissed):
LGTM

paddle/fluid/operators/mkldnn/reshape_mkldnn_op.cc
```diff
@@ -104,21 +104,16 @@ class ReshapeMKLDNNKernel : public framework::OpKernel<T> {
   }
 
   void InferInOutShape(const framework::ExecutionContext& ctx,
+
```
Silv3S: remove empty line

paddle/phi/kernels/onednn/reshape_kernel.cc
```diff
@@ -0,0 +1,171 @@
...
+              "'shape'. "
+              "But received X's shape = [%s], X's size = %d, 'shape' is "
+              "[%s], the capacity of 'shape' is %d.",
+              in_dims,
+              in_size,
+              make_ddim(shape),
+              capacity));
+    }
+  }
+  return make_ddim(output_shape);
```
Silv3S, line 114: This method is already implemented in **paddle/phi/infermeta/unary.cc**. Maybe it's better to import header and call that implementation instead of copying it here?

Silv3S: If it's not possible to import that file, I'd move this method to **onednn_reuse.h**

YuanRisheng (commented):
The CI-coverage is not successful. Please check if the `ReshapeKernel` and `SqueezeKernel` are called. Check these unittests:`test_reshape_mkldnn_op` and `test_squeeze2_mkldnn_op` with Paddle that is not compiled with cuda. If they are called, I will mark this ci as successful. Otherwise, I will help you to find the reason why they are not called.

paddle/phi/kernels/onednn/reshape_kernel.cc
sfraczek: This method is not exported in header (unary.h) so it's invisible.

sfraczek:
> The CI-coverage is not successful. Please check if the `ReshapeKernel` and `SqueezeKernel` are called. Check these unittests:`test_reshape_mkldnn_op` and `test_squeeze2_mkldnn_op` with Paddle that is not compiled with cuda. If they are called, I will mark this ci as successful. Otherwise, I will help you to find the reason why they are not called.

When I run the test test_reshape_mkldnn_op I have 4 errors on my branch and on develop the same. The errors regard bfloat on plain CPU library. They say something about missing shape variable.
I see only the default cpu reshape kernel executed. Old and new mkldnn kernel is not executed. Please help me find out why. I haven't checked squeeze yet.

sfraczek:
@YuanRisheng I would also like to register this kernel only for float and bfloat but I didn't know how, this macro seems to ignore dtype.

YuanRisheng:
> @YuanRisheng I would also like to register this kernel only for float and bfloat but I didn't know how, this macro seems to ignore dtype.

Well,  I find out the truth of the bug:https://github.com/PaddlePaddle/Paddle/pull/35781. The mkldnn kernel of `Reshape2` and `Squeeze2` have bugs and have been disabled last year. So, there are 2 options you can choose:
1, Delete the code of `Reshape2` and `Squeeze2` and they will not  migrated to PHI.
2, Fix these two kernels' bugs and migrate to PHI

I suggest the first option, because these two kernels are not used at all now and it is simple. @sfraczek

sfraczek:
> > @YuanRisheng I would also like to register this kernel only for float and bfloat but I didn't know how, this macro seems to ignore dtype.
> 
> Well, I find out the truth of the bug:#35781. The mkldnn kernel of `Reshape2` and `Squeeze2` have bugs and have been disabled last year. So, there are 2 options you can choose: 1, Delete the code of `Reshape2` and `Squeeze2` and they will not migrated to PHI. 2, Fix these two kernels' bugs and migrate to PHI
> 
> I suggest the first option, because these two kernels are not used at all now and it is simple. @sfraczek

Thank you. We will try to fix it first and delay the migration.

sfraczek:
waiting for https://github.com/PaddlePaddle/Paddle/pull/48359

PR #46342, opened 2022-09-21
Adapting device-specific Extra Attributes for the PHI kernel

chenwhql:
### PR types

New features
### PR changes

Others 
### Describe

Adapting device-specific Extra Attributes for the PHI kernel

After cleaned up most of the Extra parameters in OpMaker, it is necessary to remove the parameters belonging to the Extra range in the PHI Kernel parameter list. This is a necessary process to standardize the definition of the Kernel, and can also reduce the understanding cost of other hardware manufacturers accessing the Paddle Kernel. For example:

The function signature of the current `conv2d` kernel is as follows:

```c++
template <typename T, typename Context>
void ConvKernel(const Context& dev_ctx,
                const DenseTensor& input,
                const DenseTensor& filter,
                const std::vector<int>& strides,
                const std::vector<int>& paddings,
                const std::string& padding_algorithm,
                int groups,
                const std::vector<int>& dilations,
                const std::string& data_format,
                bool use_addto,
                int workspace_size_MB,
                bool exhaustive_search,
                DenseTensor* out);
```

And the Python API signature of conv2d is as follows:

```python
def conv2d(x,
           weight,
           bias=None,
           stride=1,
           padding=0,
           dilation=1,
           groups=1,
           data_format="NCHW",
           name=None):
```

We can found that the following three parameters in the Kernel signature have nothing to do with the python api

```
                bool use_addto,
                int workspace_size_MB,
                bool exhaustive_search,
```

In fact, these three parameters are dedicated to the cudnn kernel of conv2d, but since there can only be one signature of the Kernel, CPU, GPU, XPU in paddle, and NPU. MLU in the [CustomDevice](https://github.com/PaddlePaddle/PaddleCustomDevice/tree/develop/backends) need to have these three parameters for conv2d kernel. In fact, for these hardware, these three parameters are meaningless, which increases the understanding cost of heterogeneous Kernel development, so we need to remove such parameters. If you add MKLDNN, conv2d has more such parameters like follows:

```
- op : conv2d
  backward : conv2d_grad
  extra :
    attrs : [bool is_test = false, bool use_cudnn = true, bool fuse_relu_before_depthwise_conv = false, bool use_mkldnn = false,
             bool use_quantizer = false, str mkldnn_data_type = "float32", bool fuse_relu = false,
             str fuse_activation = "", float fuse_alpha = 0.0f, float fuse_beta = 0.0f, bool use_addto = false,
             bool fuse_residual_connection = false, float Scale_in = 1.0f, float Scale_out = 1.0f,
             float Scale_in_eltwise = 1.0f, 'float[] Scale_weights = {1.0f}', bool force_fp32_output = false,
             int workspace_size_MB = platform::GetDefaultConvWorkspaceSizeLimitMB(), bool exhaustive_search = false]
```

We cannot add such parameters to the public Kernel signature to affect the development experience of many hardware kernels. Since the parameter dedicated to a certain hardware or acceleration library, it should be passed in in a dedicated way.

However, this removal action needs to ensure that the current MKLDNN and CUDNN Kernel functions are compatible. For the Kernels of MKLDNN and CUDNN, these removed parameters are still needed at present, so they are removed from the parameter list, but need to passed into the kernel by other ways, this passing process has the following limitations:
1. It cannot be passed in through global variables (this goes against the design principles of the phi kernel function)
2. It cannot be passed in by any method that changes the declaration of the kernel function

Under the premise of the existence of the above restrictions, we can only pass it in through the existing parameters of the Kernel. At present, it seems more appropriate to pass in the Context dedicated to the Kernel:
- MKLDNN-specific parameters are passed in through OneDNNContext
- CUDNN-specific parameters are passed in through GPUContext (there is no GPUDNNContext for the time being)

This PR adopts this scheme to adapt the existing MKLDNN and CUDNN dedicated parameters.

Specifically:
1. Add thread_local `dnn_attrs_` member to store special parameters in OneDNNContext and GPUContext
2. Normalize the parameter list of the current `conv2d` kernel, obtain the cudnn-specific parameters through dev_ctx in the CUDNN Kernel, and verify it through unittests
3. According to the normalized `conv2d` kernel signature, migrate mkldnn's `conv2d` kernel to phi, and also obtain mkldnn-specific parameters through dev_ctx, and verify it through unittests

Other Changing:
1. Add TypeInfo into all DeviceContext for judging Context type by enum value
2. Polish the marco in enforce.h, move to suitable location
3. Fix conv argument name typo
4. The NPU and MLU conv2d kernel signature also need to be updated, https://github.com/PaddlePaddle/PaddleCustomDevice/pull/151

TODO:
1. Do not remove the original ConvMKLDNN(Grad)Kernel for the time being in this PR, and remove it after both depthwise_conv2d and conv3d are migrated


paddle/fluid/framework/operator.cc
```diff
@@ -2369,7 +2371,10 @@ Scope* OperatorWithKernel::PrepareData(
   };
 
   if (run_phi_kernel_) {
-    const auto& input_names = kernel_signature_->input_names;
+    std::vector<std::string> input_names;
```
zyfncg: why need a new vector for input_names?

qili93 (dismissed):
LGTM

chenwhql:
Could you help to review this PR and evaluate these changes on your internal system? After this PR is merged, we can start the migration of remaining kernels @jczaja @Silv3S

YuanRisheng (dismissed):

ronny1996 (dismissed):
LGTM

jiahy0825 (dismissed):
LGTM

paddle/phi/kernels/cpu/conv_grad_kernel.cc
```diff
@@ -101,3 +131,14 @@ PD_REGISTER_KERNEL(depthwise_conv2d_grad,
 
 PD_REGISTER_KERNEL(
     conv3d_grad, CPU, ALL_LAYOUT, phi::Conv3DGradKernel, float, double) {}
+
+PD_REGISTER_KERNEL(
+    conv2d_grad_grad, CPU, ALL_LAYOUT, phi::ConvGradGradKernel, float, double) {
```
zyfncg, line 136: conv2d_grad_grad -> cond2d_double_grad?

paddle/fluid/framework/operator.cc
```diff
@@ -3066,6 +3153,63 @@ void OperatorWithKernel::BuildPhiKernelContext(
...
+  // which increases the The cost of development and understanding, so we
+  // still use Attrs to get and the attributes set by these passes from Attrs
+  // for the time being. In the future, it is necessary to clarify the
+  // positioning of RuntimeAttrs and expand related functions.
+  auto& attrs = Attrs();
+  for (const auto& attr_iter : attrs) {
+    auto& attr_name = attr_iter.first;
+    auto& attr = attr_iter.second;
+    auto attr_propertys = paddle::operators::GetExtraAttrPropertys(attr_name);
+    SetDnnAttrIntoDeviceContext(dev_ctx, attr, attr_name, attr_propertys);
```
zyfncg, line 3187: Is there some cases that the dnn_attr of one op is misused by following ops?

paddle/phi/kernels/cpu/conv_grad_kernel.cc
chenwhql: All naming of conv double grad kernel is invalid, include conv3d, normlize them in another PR

paddle/fluid/framework/operator.cc
chenwhql: No such case currently exists

paddle/fluid/framework/operator.cc
chenwhql: Because the member of input_names is `char*`, and we need to ensure a input name whether in the `input_names `, the `find` operation need to trans input_names from `vector<char*>` to `vector<string>`

chenwhql: we can use the recored extra input names to avoid create this vector

Silv3S:
> Could you help to review this PR and evaluate these changes on your internal system? After this PR is merged, we can start the migration of remaining kernels @jczaja @Silv3S

Yes, we will review changes and check performance/accuracy in our internal validation

paddle/fluid/framework/operator.cc
chenwhql: done, thx

zyfncg (dismissed):
LGTM

paddle/fluid/operators/ops_extra_info.h
```diff
@@ -14,11 +14,139 @@
...
+        {"force_fp32_output", ExtraAttrProperty::ONEDNN},
+        {"fuse_activation", ExtraAttrProperty::ONEDNN},
+        {"fuse_activation_type", ExtraAttrProperty::ONEDNN},
+        {"fuse_activation_alpha", ExtraAttrProperty::ONEDNN},
+        {"fuse_activation_beta", ExtraAttrProperty::ONEDNN},
+        {"fuse_activation_scale", ExtraAttrProperty::ONEDNN},
+        {"fuse_alpha", ExtraAttrProperty::ONEDNN},
+        {"fuse_beta", ExtraAttrProperty::ONEDNN},
+        {"fuse_brelu", ExtraAttrProperty::ONEDNN},
+        {"fuse_brelu_threshold", ExtraAttrProperty::ONEDNN},
```
Silv3S: `fuse_brelu` or `fuse_brelu_threshold` are no longer used by any oneDNN kernel. Is there any compatibility issue preventing us from deleting these attrs?

paddle/fluid/operators/ops_extra_info.h
```diff
@@ -14,11 +14,139 @@
...
+
+const std::unordered_map<std::string, ExtraAttrPropertySet>
+    extra_attr_properties = {
+        // DEPRECATED attributes
+        {"use_quantizer", ExtraAttrProperty::DEPRECATED},
+        // SCHEDULE attributes
+        {"use_cudnn", ExtraAttrProperty::SCHEDULE},
+        {"use_mkldnn", ExtraAttrProperty::SCHEDULE},
+        // ONEDNN dedicated attributes
+        {"Bias", ExtraAttrProperty::ONEDNN},
```
Silv3S, line 91: Can new extra oneDNN-specific attributes be added in future?

paddle/phi/kernels/onednn/conv_grad_kernel.cc
```diff
@@ -0,0 +1,193 @@
...
+                    int groups,
+                    const std::string& data_format,
+                    DenseTensor* input_grad,
+                    DenseTensor* filter_grad) {
+  PADDLE_ENFORCE_EQ(dev_ctx.GetPlace().GetType(),
+                    AllocationType::CPU,
+                    phi::errors::PreconditionNotMet(
+                        "Operator DNNL ConvGrad must use CPUPlace"));
+  const auto& mkldnn_engine = dev_ctx.GetEngine();
+  VLOG(1) << "ConvGrad Has is_test " << dev_ctx.HasDnnAttr("is_test");
```
Silv3S:
```suggestion
  const auto& onednn_engine = dev_ctx.GetEngine();
```

Silv3S:
We've run first performance and functional tests and this PR passed everything with expected results. We will run one more test on other platform, to make sure that everything is ok. Tests will be finished by the end of this week.
Meanwhile we tried to migrate `softmax` kernel with changes from this branch and everything is working as expected 👍 https://github.com/PaddlePaddle/Paddle/pull/47339

paddle/phi/kernels/onednn/conv_grad_kernel.cc
chenwhql: thx

paddle/fluid/operators/ops_extra_info.h
chenwhql: thx, I deleted these two attributes, if there are other attributes that are deprecated, they can also be removed

Silv3S: We will update this list along with second phase of kernel migration

paddle/fluid/operators/ops_extra_info.h
chenwhql: In the process of kernel migration, if there are missing attributes, you can add them temporarily . However, in the long run, it is recommended to add fusion op and kernel, and replace the standard op with fusion op in the pass processing stage to support additional functions, so as to avoid the increasingly complex implementation of the basic Op kernel.

chenwhql:
@Silv3S Hello, are the tests finished?

Silv3S:
LGTM
There are small differences in accuracy (some positive, some negative). We will run more test after this PR is merged, as they are more stable.

YuanRisheng (approved):

qili93 (approved):
LGTM

XiaoguangHu01 (approved):
LGTM

zyfncg (approved):

paddle/phi/backends/onednn/onednn_context.cc
```diff
@@ -284,15 +285,105 @@ struct OneDNNContext::Impl {
...
+
+  const std::vector<std::string>& GetInputsName(
+      const std::string& input) const {
+    auto it = inputs_name_.find(input);
+    PADDLE_ENFORCE_NE(it,
+                      inputs_name_.end(),
+                      phi::errors::NotFound(
+                          "OneDnnContext does not have the input %s.", input));
+    return it->second;
+  }
```
sfraczek, line 338: I think name of this function is similar to `SetInputsName` but _getter_ doesn't return what _setter_ is setting which might be confusing at first sight. This could have a more distinct or precise name.
The same goes for `SetOutputsName` and `GetOutputsName`.

paddle/phi/backends/onednn/onednn_context.cc
```diff
@@ -284,15 +285,105 @@ struct OneDNNContext::Impl {
...
+
+  // Holds some attributes only used by the onednn kernel calculation
+  // Since original mkldnn op kernel directly adds the operations that require
+  // fusion to the native kernel operations, and uses the attribute `fuse_xxx`
+  // to control, for onednn, there will be some attributes that seem to be
+  // independent of the device are also saved here.
+  // Here, the operation of fusion needs to be implemented separately as
+  // a fusion op and kernel, instead of patching it to a basic operation.
+  // Because DeviceContext is a global singleton, you need to ensure thread
+  // safety, use the thread_local variable
```
sfraczek, line 367:
```suggestion
  // Holds some attributes only used by the onednn kernel calculation.
  // Since original onednn op kernel directly adds the operations that require
  // fusion to the native kernel operations, and uses the attribute `fuse_xxx`
  // to control, for onednn, there will be some attributes that seem to be
  // independent of the device are also saved here.
  // Here, the operation of fusion needs to be implemented separately as
  // a fusion op and kernel, instead of patching it to a basic operation.
  // Because DeviceContext is a global singleton, you need to ensure thread
  // safety, use the thread_local variable
```

PR #46365, opened 2022-09-21
[Cherry-Pick]Fix bug in convert call: tranform the static func. (#46278)

2742195759:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix bug in convert call: tranform the static func. (#46278)


Aurelius84 (approved):
LGTM

PR #46470, opened 2022-09-24
Move some singleton to cc file

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
Move some singleton implementation method from header file to source file. 

If some external C++ source codes are built upon the pre-built `libpaddle.so` and include the header file containing the singleton implementation method, multiple singleton would be generated in the built so file. This PR is used to avoid this error.


gongweibao (approved):
LGTM

PR #46486, opened 2022-09-26
Fix op_happen_before_ update bug for AddDownstreamOp

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

In DependencyBuilder, after we add a OP dependency _a_->_b_, the _op_happen_before_ in the following two cases need to be updated:
1. For all _c_ happens before _a_, it also happens before _b_, so _op_happen_before[c][b]_ should be set to True.
2. For all _c_ happens after _b_, it also happens after _a_, so _op_happen_before[a][c]_ should be set to True.

The previous code only deals with the latter situation, this PR adds consideration to the former one.


zhiqiu (approved):
LGTM

PR #46519, opened 2022-09-26
[Auto Parallel] Generalize Amp Pass

JZ-LIANG:
### PR types
Function optimization
### PR changes
Others

### Describe
generalize amp pass to support backward operators that no need forward output
and ERNIE training strategies like mask.


aoyulong (approved):
LGTM

PR #46572, opened 2022-09-27
[AutoParallel] fix sharding

zhaoyinglia:
### PR types

Bug fixes
### PR changes

Others
### Describe

- fix sharding in pp scene.
image


JZ-LIANG (approved):
the actual problem for that bug  is weight-sharing + pipeline + sharding。 
weight-sharing + pipeline： need an extra allreduce for gradient synch, 
sharding should be adopted to consider above situation.

JZ-LIANG (approved):
LGTM

Will has a more formal implement for this case in future.

PR #46636, opened 2022-09-29
[inference] CPU-> GPU async io copy for TensorRT using ShareExternalData API

zhangjun:
### PR types

Others
### PR changes

Others
### Describe

- CPU -> GPU async io for  TensorRT 
- fix copy core dump using `ShareExternalData` if GPU id is not 0


jiweibo (approved):
LGTM

PR #46670, opened 2022-09-30
Release memory cache after build_op_func_list in interpretercore

From00:
### PR types
Performance optimization

### PR changes
Others

### Describe

The completely different OP running order between prepare (i.e., build_op_func_list) and real running (i.e., RunInstructionAsync) may lead to serious memory fragments in Allocator. Therefore, in this PR, we release memory cache after build_op_func_list to avoid memory fragments. It reduces about 10% memory usage for V100 8-GPU training of transformer_base_bs4096_amp_fp16 and transformer_base_bs4096_pure_fp16 models.


zhiqiu (approved):
LGTM

PR #46674, opened 2022-09-30
summer-ospp 2022: 飞桨PaddlePaddle Sparse Conv开发和优化——hash table

umiswing:
### PR types
Performance optimization

### PR changes

OPs

### Describe

1. Add a custom cudf (based on v0.6.1) to implement hash table on GPU (refer to [existing code in paddle](https://github.com/PaddlePaddle/Paddle/tree/798670bba6d720bebd3f07bf98b06622b2d864ea/paddle/fluid/framework/fleet/heter_ps/cudf)).
2. Optimize Conv3dCooGPUKernel with hash table when performing subm conv. Using a hash table for input with a large dense shape to avoid initializing a large index table. 
3. Dynamic choose hash table or normal index table for different input shapes (because hash computation is slow on GPU, hash table may hurt the performance for some input shapes). The condition needs to be tuned on different hardware. I choose the best one for GeForce RTX 3060 and the current test shape now.

Speed up about 15% on GeForce RTX 3060 Mobile GPU for some input when performing subm conv ( I perform the experiment by running 100 times warm up first, and testing the time of 500 times forward):
|(x, y, z, in_channel, out_channel, sparsity)|(41,1600,1408, 4, 16, 0.9985)|(41,1600,1408, 16, 16, 0.9985)|
|---|---|---|
|without hash table(sec)|1.01882895099993|1.05665245900002
|with hash table(sec)|0.86128283599879|0.9077350210000077

I overload some kernel to avoid performing if-else conditions on GPU. We could use Marco to restructure these overloaded functions later.

### DIFF
--- a/paddle/phi/kernels/sparse/gpu/conv.cu.h
+++ b/paddle/phi/kernels/sparse/gpu/conv.cu.h
@@ -674,19 +674,23 @@ int ProductRuleBook(const Context& dev_ctx,
   DenseTensorMeta rulebook_meta(
       indices_dtype, {rulebook_rows, rulebook_cols}, DataLayout::NCHW);
 
-  int64_t table_size = 1;
+  int64_t dense_size = 1;
   for (int i = 0; i < out_dims.size() - 1; i++) {
-    table_size *= out_dims[i];
+    dense_size *= out_dims[i];
   }
 
   if (subm) {
+    int64_t dense_size = 1;
+    for (int i = 0; i < out_dims.size() - 1; i++) {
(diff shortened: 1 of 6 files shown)


paddle/phi/kernels/sparse/gpu/conv.cu.h
```diff
@@ -1,11 +1,8 @@
 /* Copyright (c) 2022 PaddlePaddle Authors. All Rights Reserved.
-
```
zkh2016, line 2: license的修改这里恢复下。

PR #46675, opened 2022-09-30
[MKLDNN] Delete #ifdef MKLDNN in prelu GetKernelTypeForVar

jiahy0825:
### PR types
Others

### PR changes
OPs

### Describe
In `PADDLE_WITH_MKLDNN` mode, the layout result of `GetExpectedKernelType` is **kMKLDNN**. Either the layout result of `GetKernelTypeForVar` is **kNHWC** or **tensor.layout()**, tensor needs to transform to **kMKLDNN**. Therefore, the hard code within `#ifdef PADDLE_WITH_MKLDNN` of `GetKernelTypeForVar` can be deleted directly.

在`PADDLE_WITH_MKLDNN`模式下，`GetExpectedKernelType`的layout结果是**kMKLDNN**。不管`GetKernelTypeForVar`的返回值是**kNHWC**还是**tensor.layout()**，tensor都需要转换成**kMKLDNN**，因此可以直接删除`GetKernelTypeForVar`中的硬编码

### DIFF
--- a/paddle/fluid/operators/prelu_op.cc
+++ b/paddle/fluid/operators/prelu_op.cc
@@ -52,17 +52,6 @@ class PReluOp : public framework::OperatorWithKernel {
       const std::string &var_name,
       const Tensor &tensor,
       const framework::OpKernelType &expected_kernel_type) const override {
-#ifdef PADDLE_WITH_MKLDNN
-    // All inputs (including alpha) need shape rotating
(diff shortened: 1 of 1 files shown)


jiahy0825:
During the kernel selection process, DataTransform has a specific implementation for MKLDNN layout, which only records data format instead of doing the real transformation. Thus hardcode within `GetKernelTypeForVar` cannot be deleted directly, we need to find another way to deal with it.

PR #46818, opened 2022-10-09
[inference][trt] reshape do not support ShapeTesnor Input using dynamic shape mode

zhangjun:
### PR types

Others
### PR changes

Others
### Describe

- reshape do not support ShapeTesnor Input using dynamic shape mode
image


zhangjun:
#46482  will fix this.

PR #46851, opened 2022-10-10
[Dy2St]Remove usless cast operation to speed up FP16 training

Aurelius84:
### PR types

Performance optimization
### PR changes

Others
### Describe

[Dy2St]Remove usless cast operation to speed up FP16 training


zhangbo9674 (approved):
LGTM

PR #46861, opened 2022-10-10
[inference][trt] fix reshape2 opteller and elementwise min/max trt registration

zhangjun:
### PR types

Others
### PR changes

Others
### Describe

- fix reshape2 opteller
- add elementwise min/max trt registration


jiweibo (approved):
LGTM

PR #46757, opened 2022-10-05
Add fc residual pattern

sfraczek:
### PR types
New features

### PR changes
Others

### Describe
fc mkldnn pattern didn't use `bias` so I replaced it with `residual_data`.

PR #41776, opened 2022-04-13
FC+elementwise_add (residual connection)

Silv3S:
### PR types
Performance optimization

### PR changes
OPs

### Describe
- Fuse FC with elementwise_add so residual data is added inside FC operator,
- add unit tests with multiple cases for this fuse pass,
- download VIT-OCR model and perform accuracy test.

### Benchmarks
Test were conducted with oneDNN 2.5.4 , `fc_mkldnn_pass` and `fc_act_mkldnn_pass` were enabled:
- VIT-OCR is performing 4% faster (12 fc+residual fusions),
- BERT is performing 8% faster (24 fc+residual fusions).

PR #40834, opened 2022-03-22
FC + elementwise_add (Residual connection)

Silv3S:
### PR types
Performance optimization

### PR changes
OPs

### Describe
- Fuse FC with elementwise_add so residual data is added inside FC operator,
- add unit tests with multiple cases for this fuse pass,

### Benchmarks
Test were conducted with oneDNN 2.5.4 , `fc_mkldnn_pass` and `fc_act_mkldnn_pass` were enabled:
- VIT-OCR is performing 4% faster (12 fc+residual fusions),
- BERT is performing 8% faster (24 fc+residual fusions).

PR #46917, opened 2022-10-11
add fc-residual quantization

sfraczek:
### PR types
Performance optimization

### PR changes
Others

### Describe
Add quantization of FC+residual fused op


paddle/fluid/framework/ir/mkldnn/cpu_quantize_squash_pass.cc
```diff
@@ -426,9 +427,25 @@ void CPUQuantizeSquashPass::MultipleQuantizeSquash(Graph* graph) const {
...
+                            platform::errors::InvalidArgument(
+                                "Only the ResidualData output is allowed to be "
+                                "linked to earlier op."));
+          PADDLE_ENFORCE_EQ(last_op_op->Type(),
+                            "fc",
+                            platform::errors::InvalidArgument(
+                                "Only fc operator supports ResidualData output "
+                                "linked as input."));
+          residual_fc = true;
+        }
```
tsocha: This logic seems a little bit odd for me.
Does residual connection must have an empty name?
Are there any other scenarios where the name can be empty?
What if op will have empty name in graph, will this pass throw an exception?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -85,14 +85,16 @@ void CPUQuantizePass::QuantizeInput(Graph* g,
...
+  auto inputs = residual_fc ? op->Op()->OutputNames() : op->Op()->InputNames();
   bool name_found =
       std::find(inputs.begin(), inputs.end(), input_name) != inputs.end();
   PADDLE_ENFORCE_EQ(name_found,
                     true,
                     platform::errors::InvalidArgument(
-                        "Var(%s) isn't the input of the %s operator.",
+                        "Var(%s) isn't the %s of the %s operator.",
                         input_name,
+                        residual_fc ? "output" : "input",
```
tsocha: It's stinky for me, what if we will add more ops with residual connection in the future ?

tsocha (changes requested):

paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
```diff
@@ -32,7 +32,7 @@ class Graph;
 namespace {
 void LogEnabledOps(const int counter, const std::string& details) {
```
Silv3S: This method is redefined in each pass. Some log only if count is > 0, others log all. Do you think we could extract this fuse pass logger to `mkldnn_reuse.h` and only call it?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_squash_pass.cc
sfraczek: it's empty because FindInputNameByVarName returns empty string if it's not found. In that case we search for name in the outputs instead of inputs because fc is special case when residualData is output instead of input.
If this if is entered and those two checks pass, then we know we are dealing with fc with residual data (in output). And so we set residual_fc= true which later informs decisions whether we should call SetInput or SetOutput

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
sfraczek: It depends on what we will agree on with baidu. Previously it was easy because we assumed that ResidualData will be input because on the graph it's linked as input and in OpDesc it was set as input. But recently when FC with residual was added, we were instructed to use Output in OpDesc but it's impossible to set it as output in graph so it's still linked as input. From this comes this confusion in code both in graph pattern detector I had to write custom asserts and here I had to do some hacks that I don't have a good idea for design yet especially that it might be subject to change based on what baidu will say about this. I don't know if there is a great way to make this clear because it is inherently confusing so let's say it's a POC.

paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
sfraczek: I saw one in cpu_quantize_pass.cc from which I copied and modified it but this is only two examples so please give me a third as per the rule of three :D
https://understandlegacycode.com/blog/refactoring-rule-of-three/#use-the-rule-of-three

sfraczek:
Hi @zhiqiu,
This PR https://github.com/PaddlePaddle/Paddle/pull/41776 has created some additional confusion with regards to writing asserts for residual connection in graph_pattern_detector.cc https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/ir/graph_pattern_detector.cc#L1081-L1094. Also in quantization (this PR) now we have to look for residual data not just in input but also in output. Can you please share your thoughts on how we should proceed with residual connections?

paddle/fluid/framework/ir/mkldnn/fc_mkldnn_pass.cc
Silv3S: I mean we have multiple fuse pass counters with logs e.g. elt_act_mkldnn_fuse_pass.cc:102, fc_elementwise_add_mkldnn_fuse_pass.cc:119, matmul_transpose_reshape_mkldnn_fuse_pass.cc:108 etc.

I think that same code is repeated each time, with only hardcoded op name being changed. It is not particularly related to your PR, I just wanted to gather opinions if you find it good idea as future improvement.

sfraczek: Ok there is some repetition even in non mkldnn passes so if we were to add such method in future we could put it in FusePassBase or some other common parent class.

tsocha (dismissed):

jakpiase (dismissed):
LGTM, but this logic with residualdata as an output looks weird to me

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -493,9 +502,29 @@ void CPUQuantizePass::QuantizeFc(Graph* graph) const {
...
+
+      QuantizeInput(g,
+                    fc,
+                    residual_data,
+                    "ResidualData",
+                    residual_scale,
+                    is_residual_unsigned,
+                    "Scale_in_eltwise");
+    } else {
+      if (!AreScalesPresentForNodes({input, weights})) {
```
Silv3S: You can check if scales are present for input and weights first, because it's common for both conditions. Then if `with_residual_data` check only if scales are present for node `residual_data`

Silv3S (dismissed):
LGTM

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
sfraczek: I didn't want to do that only because this method is traversing the graph which I think it's a waste of computational power ;D

sfraczek: Ok

zhiqiu:
> Hi @zhiqiu, This PR #41776 has created some additional confusion with regards to writing asserts for residual connection in graph_pattern_detector.cc https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/ir/graph_pattern_detector.cc#L1081-L1094. Also in quantization (this PR) now we have to look for residual data not just in input but also in output. Can you please share your thoughts on how we should proceed with residual connections?

Hi, I don't really know the background of this work. But I think it is not a good design that `ResidualData` can be both input and output. Can you explain more about your coments above, i.e.,`instructed to use Output in OpDesc`?

sfraczek:
> Hi, I don't really know the background of this work. But I think it is not a good design that `ResidualData` can be both input and output. Can you explain more about your coments above, i.e.,`instructed to use Output in OpDesc`?

I was referring to your suggestion that ResidualData should be made output instead of input https://github.com/PaddlePaddle/Paddle/pull/40834#issuecomment-1091138330. This PR was closed and reopened as new one PR #41776 and merged. Previously, we had already been using ResidualData as input for convolution and now we have ResidualData for FC as output. In Paddle we have quantization code for convolution op with ResidualData as input. Here I am adding quantization of fc with ResidualData as Output. 
As you say, it seems wrong to have both input and output so we want to unify it.
We can
* change CONV's ResidualData to Output,
* change FC's ResidualData to Input.

To me it seems that Input was better for the following reasons:
1. In graph_pattern_detector.cc it was easy to write a pattern using existing assert functions but for residual_data as output the pattern got this complicated because it checks if it is **input** and **output** at the same time so I added a comment https://github.com/PaddlePaddle/Paddle/pull/46757/files#diff-3363ff47a1f22a111386eeff77a8572f6618d7d0602ffe71e92ecf8682faa84eR1081-R1094,
2. Netron shows an arrow for ResidualData connection only when it is **input** but when it's **output** it doesn't show it so at the first glance there is no connection visible when analyzing graph of model and it requires clicking on a node to see if it has residualData output. Using netron to inspect models is what we are used to do routinely to look for optimizations.

It is logically a little confusing to me that we have to SetOutput and ir_node_link_to residual_data -> fc at the same time https://github.com/PaddlePaddle/Paddle/pull/41776/files#diff-9bebe579d9782bcfbf4b11872065b743461e89a36c3924972e98ee7df8824d16R102-R109

Beside the above a small extra argument for me is that In quantization code (cpu_quantize_pass.cc) we will not be able to reuse the QuantizeInput method like it is used for all inputs but we will have to add another method that will be doing mostly the same thing but replace input with output - perhaps called QuantizeResidualData.

zhiqiu:
> > Hi, I don't really know the background of this work. But I think it is not a good design that `ResidualData` can be both input and output. Can you explain more about your coments above, i.e.,`instructed to use Output in OpDesc`?
> 
> I was referring to your suggestion that ResidualData should be made output instead of input [#40834 (comment)](https://github.com/PaddlePaddle/Paddle/pull/40834#issuecomment-1091138330). This PR was closed and reopened as new one PR #41776 and merged. Previously, we had already been using ResidualData as input for convolution and now we have ResidualData for FC as output. In Paddle we have quantization code for convolution op with ResidualData as input. Here I am adding quantization of fc with ResidualData as Output. As you say, it seems wrong to have both input and output so we want to unify it. We can
> 
> * change CONV's ResidualData to Output,
> * change FC's ResidualData to Input.
> 
> To me it seems that Input was better for the following reasons:
> 
> 1. In graph_pattern_detector.cc it was easy to write a pattern using existing assert functions but for residual_data as output the pattern got this complicated because it checks if it is **input** and **output** at the same time so I added a comment https://github.com/PaddlePaddle/Paddle/pull/46757/files#diff-3363ff47a1f22a111386eeff77a8572f6618d7d0602ffe71e92ecf8682faa84eR1081-R1094,
> 2. Netron shows an arrow for ResidualData connection only when it is **input** but when it's **output** it doesn't show it so at the first glance there is no connection visible when analyzing graph of model and it requires clicking on a node to see if it has residualData output. Using netron to inspect models is what we are used to do routinely to look for optimizations.
> 
> It is logically a little confusing to me that we have to SetOutput and ir_node_link_to residual_data -> fc at the same time https://github.com/PaddlePaddle/Paddle/pull/41776/files#diff-9bebe579d9782bcfbf4b11872065b743461e89a36c3924972e98ee7df8824d16R102-R109
> 
> Beside the above a small extra argument for me is that In quantization code (cpu_quantize_pass.cc) we will not be able to reuse the QuantizeInput method like it is used for all inputs but we will have to add another method that will be doing mostly the same thing but replace input with output - perhaps called QuantizeResidualData.

Thanks, I got your views. If Input s already used and it is BETTER, I agree that you can change `output` to `input`.

sfraczek:
Something is broken about the CI.

yeliang2258:
@sfraczek  Please resolve the conflict of the code, thank you

yeliang2258:
@sfraczek Please resolve the conflict of the code, thanks

sfraczek:
paddle is broken again

sfraczek:
@chenwhql Please approve because the test that failed is outdated. It uses old Quant2Int8MkldnnPass which was replaced by C++ passes. We are planning to rewrite the test in Q1.

chenwhql:
> @chenwhql Please approve because the test that failed is outdated. It uses old Quant2Int8MkldnnPass which was replaced by C++ passes. We are planning to rewrite the test in Q1.

@sfraczek I don't understand, approving cannot resolve the test failed problem

python/paddle/fluid/contrib/slim/quantization/quant2_int8_mkldnn_pass.py
```diff
@@ -482,6 +482,9 @@ def _optimize_fp32_graph(self, graph):
             # Disabled due to topology-dependent speed-up
             graph = self._apply_pass(graph, 'fc_mkldnn_pass')
             graph = self._apply_pass(graph, 'fc_act_mkldnn_fuse_pass')
+            graph = self._apply_pass(
+                graph, 'fc_elementwise_add_mkldnn_fuse_pass'
+            )
```
sfraczek:
```suggestion
```
I will revert changes to no longer supported python script

tsocha (approved):

yeliang2258 (approved):
LGTM

jczaja (approved):
LGTM

PR #46934, opened 2022-10-12
[Eager] Add shape info into eager log

JiabinYang:
### PR types

Others
### PR changes

Others
### Describe

This PR Add Tensor shape info in eager log


wanghuancoder (approved):
LGTM

PR #47055, opened 2022-10-17
[IPU] fix runtime error

gglin001:
### PR types

Others
### PR changes

Others
### Describe

fix runtime error when enable building with ipu


gglin001:
moved to https://github.com/PaddlePaddle/Paddle/pull/47133

PR #47089, opened 2022-10-18
CudnnNormConvolution is no longer supported on NVIDIA Hopper GPUs

Tom-Zheng:
### PR types

Bug fixes
### PR changes

OPs
### Describe

CudnnNormConvolution relies on CUDNN fused_ops, which is deprecated and is no longer supported on Hopper GPUs (H100 and later). We add necessary prompts when the user tries to use this OP on Hopper devices. 

Meanwhile, we fix the related unit tests on the new hardware.


Xreki:
conflict了，请先更新一下develop代码吧

PR #44124, opened 2022-07-06
Quantize shape operator

zuzg:
### PR types
New features

### PR changes
OPs

### Describe
Quantize shape operator, this could speed up faster_rcnn_r50_fpn_1x_coco.

PR #47119, opened 2022-10-18
[cherry-pick] Fix quantize model deploy bug in MKLDNN

yeliang2258:
### PR types

Performance optimization

### PR changes

OPs 
### Describe

refs https://github.com/PaddlePaddle/Paddle/pull/45920

### DIFF
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
@@ -348,8 +348,10 @@ void CPUQuantizePass::GetQuantInfo(Graph* graph) const {
   }
 }
 
-void CPUQuantizePass::QuantizeConv(Graph* graph,
-                                   bool with_residual_data) const {
+void CPUQuantizePass::QuantizeConv(
+    Graph* graph,
+    bool with_residual_data,
+    std::vector<std::string>* changed_weight) const {
   GraphPatternDetector gpd;
   auto pattern = gpd.mutable_pattern();
   patterns::ConvResidual conv_pattern{pattern, name_scope_};
@@ -425,11 +427,11 @@ void CPUQuantizePass::QuantizeConv(Graph* graph,
 
     // If the scale value of a weight is already multiplied by S8_MAX, it does
     // not need to be multiplied again
-    if (std::find(change_weight_->begin(),
-                  change_weight_->end(),
-                  conv_filter->Name()) == change_weight_->end()) {
+    if (std::find(changed_weight->begin(),
+                  changed_weight->end(),
+                  conv_filter->Name()) == changed_weight->end()) {
       eigen_tensor *= static_cast<double>(S8_MAX);
-      change_weight_->push_back(conv_filter->Name());
+      changed_weight->push_back(conv_filter->Name());
     }
 
     std::vector<float> filter_scale{
@@ -1174,9 +1176,12 @@ void CPUQuantizePass::ApplyImpl(ir::Graph* graph) const {
       param_scope(),
       platform::errors::InvalidArgument("Scope cannot be nullptr."));
 
+  // Save the scale values of which weights have been processed to avoid
+  // secondary processing
+  std::vector<std::string> changed_weight = {};
   GetQuantInfo(graph);
-  QuantizeConv(graph, false /* with_residual_data */);
-  QuantizeConv(graph, true /* with_residual_data */);
+  QuantizeConv(graph, false /* with_residual_data */, &changed_weight);
+  QuantizeConv(graph, true /* with_residual_data */, &changed_weight);
   QuantizePool(graph);
   QuantizeConcat(graph);
   QuantizePriorBox(graph);
--- a/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
+++ b/paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
@@ -422,7 +422,16 @@ void CPUQuantizePass::QuantizeConv(Graph* graph,
     auto filter_scale_tensor = GetScaleTensorForNode(conv_filter);
     EigenVectorArrayMap eigen_tensor{filter_scale_tensor.data<double>(),
                                      filter_scale_tensor.numel()};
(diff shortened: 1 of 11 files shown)


paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -52,36 +52,25 @@ bool HasBias(ir::Node* conv_op) {
...
-      conv_op->Op()->GetAttrIfExists<std::vector<float>>(scales_attr_name);
-
-  auto* tensor = scope->GetVar(input_name)->GetMutable<LoDTensor>();
-  QuantizeParams<T>(tensor, scales);
-
-  conv_op->Op()->SetAttr(scales_attr_name, std::vector<float>(1, 1));
+  auto var = scope->GetVar(input_name);
+  if (var->Get<LoDTensor>().dtype() != phi::DataType::FLOAT32) {
+    VLOG(0) << "Skipping convolution filter: " << input_name
+            << " because it is detected again.";
```
sfraczek: QuantizeConvInput is called for both filters and biases so I think this message might be printed twice now and will say filter even when it may be bias. The reason for this if check was to quit early if the same conv pattern was detected twice. Now the handler code might be executed twice and it will just skip only quantization of weights and biases but in the end the counter will show it as quantized so the LOG will say that more conv ops were quantized than actually exist in the model.

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -52,36 +52,25 @@ bool HasBias(ir::Node* conv_op) {
...
+    conv_op->Op()->SetAttr(scales_attr_name, std::vector<float>(1, 1));
```
sfraczek: this line is duplicated in the else branch

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -1170,7 +1186,6 @@ void CPUQuantizePass::ApplyImpl(ir::Graph* graph) const {
   QuantizeImmutable(graph, "reshape2", "X");
   QuantizeImmutable(graph, "transpose2", "X");
   QuantizeImmutable(graph, "slice", "Input");
-  QuantizeImmutable(graph, "shape", "Input");
```
sfraczek, line 1171: If shape is to be never quantized, you could remove it from all other places where it was added here
https://github.com/PaddlePaddle/Paddle/pull/44124/files

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.h
```diff
@@ -110,6 +110,11 @@ class CPUQuantizePass : public FusePassBase {
   VarQuantScale string_pair_map = {};
   VarQuantScale* const var_quant_scales_ = &string_pair_map;
 
+  // Save the scale values of which weights have been processed to avoid
+  // secondary processing
+  std::vector<std::string> change_weight = {};
+  std::vector<std::string>* const change_weight_ = &change_weight;
```
sfraczek: Does it need both value and pointer?

yeliang2258: Because all modifications must be done through pointers

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
yeliang2258: Thanks, Done

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
yeliang2258: Done

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.h
sfraczek: But why here?

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
sfraczek: Maybe I was wrong about the second part. Maybe it's not detected twice. The ConvResidual pattern used might be working correctly as opposed to Conv pattern in int8_scale_calculation_mkldnn_pass. If so, it will display correct number of detected ops and just avoid quantizing them again. So it seems good to me.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.h
yeliang2258: thanks, i updated the code

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
```diff
@@ -52,35 +52,23 @@ bool HasBias(ir::Node* conv_op) {
...
                        const std::string& scales_attr_name) {
-  const auto scales =
-      conv_op->Op()->GetAttrIfExists<std::vector<float>>(scales_attr_name);
-
-  auto* tensor = scope->GetVar(input_name)->GetMutable<LoDTensor>();
-  QuantizeParams<T>(tensor, scales);
-
+  auto var = scope->GetVar(input_name);
+  if (var->Get<LoDTensor>().dtype() != phi::DataType::FLOAT32) {
+    VLOG(0) << "Skipping quantize the input: " << input_name
```
jiangjiajun: 删除

jiangjiajun (changes requested):

paddle/fluid/framework/ir/mkldnn/params_quantization_mkldnn_pass.cc
yeliang2258: Done

jiangjiajun (approved):
LGTM

ZeyuChen (approved):

PR #47127, opened 2022-10-18
FC/matmul(v2) + scale fuse pass

Silv3S:
### PR types
Performance optimization

### PR changes
OPs

### Describe
- fuse FC with scale,
- fuse matmul and matmul_v2 with scale,
- add unit tests.


paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -331,6 +331,7 @@ void CpuPassStrategy::EnableMKLDNN() {
              // "mkldnn_inplace_pass",  // This pass should be activated after
              // fuses. Disabled by default due to
              // little gain and lots of problems
+             "operator_scale_onednn_fuse_pass",  //
```
tsocha: move it before the comment, this commented out `mkldnn_inplace_pass` can cause trouble if someone will uncomment it.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -123,6 +123,12 @@ class FCMKLDNNHandler
     }
     platform::AppendActivation(ctx, post_operations, scale);
 
+    if (ctx.HasAttr("fused_output_scale")) {
+      float scale_alpha = ctx.Attr<float>("fused_output_scale");
+      post_operations.append_eltwise(
+          1.0, dnnl::algorithm::eltwise_linear, scale_alpha, 0.0f);
+    }
+
     attributes.set_post_ops(post_operations);
```
tsocha, line 132: Can we use mechanism for post-ops from line 132 here?

paddle/fluid/platform/mkldnn_reuse.h
```diff
@@ -249,6 +249,12 @@ class MatMulV2MKLDNNHandler
 
     AppendActivation(ctx, post_operations);
 
+    if (ctx.HasAttr("fused_output_scale")) {
+      float scale_alpha = ctx.Attr<float>("fused_output_scale");
+      post_operations.append_eltwise(
+          1.0, dnnl::algorithm::eltwise_linear, scale_alpha, 0.0f);
+    }
+
```
tsocha, line 257: as above

paddle/fluid/framework/ir/mkldnn/operator_scale_onednn_fuse_pass.cc
```diff
@@ -0,0 +1,107 @@
...
+    if (operator_op->Op()->HasAttr("use_mkldnn") &&
+        !(PADDLE_GET_CONST(bool, operator_op->Op()->GetAttr("use_mkldnn")))) {
+      VLOG(4) << "Only oneDNN operator can be fused.";
+      return;
+    }
+
+    if (scale_op->Op()->GetAttrIfExists<float>("bias") != 0.0) {
+      VLOG(4) << "Only unbiased Scale can be fused.";
+      return;
+    }
```
tsocha: It would be nice if you could add an op name here. It will make it more useful for debug.

tsocha (changes requested):

paddle/fluid/inference/api/paddle_pass_builder.cc
Silv3S: Do you mind if I remove `mkldnn_inplace_pass` as we're probably not going to maintain it?

sfraczek (dismissed):
LGTM

paddle/fluid/framework/ir/mkldnn/operator_scale_onednn_fuse_pass.cc
```diff
@@ -0,0 +1,107 @@
...
+
+namespace paddle {
+namespace framework {
+namespace ir {
+
+using string::PrettyLogDetail;
+
+void FuseOperatorScaleOneDNNPass::ApplyImpl(Graph *graph) const {
+  const std::vector<std::string> fusable_ops{"fc", "matmul", "matmul_v2"};
+  for (auto op : fusable_ops) FuseScale(graph, op);
```
jakpiase:
```suggestion
  for (const auto& op : fusable_ops) FuseScale(graph, op);
```
It is a really minor thing, but just to maintain the true beauty of your code :)

paddle/fluid/framework/ir/mkldnn/operator_scale_onednn_fuse_pass.cc
```diff
@@ -0,0 +1,107 @@
...
+  GraphPatternDetector gpd;
+  patterns::OperatorActivation op_scale_pattern(
+      gpd.mutable_pattern(), op_type + "_scale_onednn_fuse_pass");
+  op_scale_pattern(op_type, "scale");
+
+  int found_operator_scale_count = 0;
+
+  auto handler = [&](const GraphPatternDetector::subgraph_t &subgraph,
+                     Graph *g) {
+    GET_IR_NODE_FROM_SUBGRAPH(operator_op, preceding_op, op_scale_pattern);
```
jakpiase, line 47: Since you are an expert in passes, what is the difference between these two first arguments(`operator_op` and `preceding_op`)? I think that in most of the passes they are equal, but in your case they are different.

Silv3S: First argument is name given/used in `_fuse_pass.cc`, second argument is name from `graph_pattern_detector.h`

tsocha: sounds like variables have bad names.

Silv3S: So for example if we have generic `OperatorActivation` graph pattern detector with very general names, we can use specific node name inside fuse pass. Like instead of `preceding_op` we can use `fc_op`.

Silv3S: > sounds like variables have bad names.

Yes I agree. Moreover when we set attributes, we need to get pointer to node->Op() which currently leads to `operator_op->Op()`. If you have any propositions I'm happy to change naming.

tsocha: > First argument is name given/used in `_fuse_pass.cc`, second argument is name from `graph_pattern_detector.h`

this comment suggests names like `pass_name` and `pattern_name`

> So for example if we have generic `OperatorActivation` graph pattern detector with very general names, we can use specific node name inside fuse pass. Like instead of `preceding_op` we can use `fc_op`.

but this one suggests that operator_op and preceding_op are fine.

I'm a little bit confused now.

sfraczek (dismissed):

tsocha (dismissed):

paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -326,6 +326,7 @@ void CpuPassStrategy::EnableMKLDNN() {
              "softplus_activation_mkldnn_fuse_pass",  //
              "shuffle_channel_mkldnn_detect_pass",    //
              "elt_act_mkldnn_fuse_pass",              //
+             "operator_scale_onednn_fuse_pass",       //
```
jakpiase, line 335: One more question - are we gonna use this pass only in FP32 models or in INT8 as well?

yeliang2258 (dismissed):
LGTM

zh794390558 (dismissed):
LGTM

paddle/fluid/inference/api/paddle_pass_builder.cc
Silv3S: We can use it in INT8 as well. For now, we will reuse approach with post-op like in FP32.

Silv3S:
@jczaja please merge

jczaja (approved):

paddle/fluid/framework/ir/mkldnn/operator_scale_onednn_fuse_pass.cc
```diff
@@ -0,0 +1,108 @@
...
+    auto const &names = scale_op->Op()->InputNames();
+    bool has_scale_tensor =
+        std::find(names.begin(), names.end(), "ScaleTensor") != names.end();
+
+    if (has_scale_tensor && scale_op->Op()->Input("ScaleTensor").size() > 0) {
+      std::string scale_var_name = scale_op->Op()->Input("ScaleTensor").front();
+      auto *scale_var = scope->FindVar(scale_var_name);
+      // ScaleTensor must be weight
+      if (scale_var == nullptr) return;
+      auto *scale_tensor = scale_var->GetMutable<LoDTensor>();
```
sfraczek, line 76: did you mean phi::DenseTensor?

Silv3S: Few hours before this PR was merged, `LoDTensor` was replaced by `phi::DenseTensor` and it caused conflict. It's already fixed in https://github.com/PaddlePaddle/Paddle/pull/47388

PR #47181, opened 2022-10-19
log only if > 0

sfraczek:
### PR types
Others

### PR changes
Others

### Describe
Quick fix to add condition to print logs in passes only if actually found any patterns to reduce clutter.

### DIFF
--- a/paddle/fluid/framework/ir/fc_gru_fuse_pass.cc
+++ b/paddle/fluid/framework/ir/fc_gru_fuse_pass.cc
@@ -350,7 +350,7 @@ void FCGRUFusePass::ApplyImpl(ir::Graph* graph) const {
       graph, name_scope_, param_scope(), true /*with_fc_bias*/);
 
   AddStatis(fusion_count);
-  if (!Has("disable_logs") || !Get<bool>("disable_logs"))
+  if ((!Has("disable_logs") || !Get<bool>("disable_logs")) && fusion_count > 0)
     string::PrettyLogDetail("---    fused %d pairs of fc gru patterns",
                             fusion_count);
 }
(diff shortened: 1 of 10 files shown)


Silv3S (approved):
LGTM

paddle/fluid/framework/ir/fc_gru_fuse_pass.cc
```diff
@@ -350,7 +350,7 @@ void FCGRUFusePass::ApplyImpl(ir::Graph* graph) const {
       graph, name_scope_, param_scope(), true /*with_fc_bias*/);
 
   AddStatis(fusion_count);
-  if (!Has("disable_logs") || !Get<bool>("disable_logs"))
+  if ((!Has("disable_logs") || !Get<bool>("disable_logs")) && fusion_count > 0)
```
tsocha, line 353: What do you think about it?
```suggestion
  const auto enabled_logs = !Has("disable_logs") || !Get<bool>("disable_logs");
  if (enabled_logs && fusion_count > 0)
```

tsocha (changes requested):
I have one suggestion which can be used for all of these cases.

paddle/fluid/framework/ir/fc_gru_fuse_pass.cc
sfraczek: I think we might be extracting this to a separate function in future as @Silv3S proposed and there it could be good idea to do it as you suggest. For now, I think it's fine both ways. I just made this quickly while working on something else and I decided to share it. What do you think about leaving it as it is for now and rewriting this later?
Here is the discussion:
https://github.com/PaddlePaddle/Paddle/pull/46917#discussion_r993268723

tsocha: @Silv3S liked your suggestion so I'm fine with it as well.
We just need to remember about it.

tsocha (approved):

jczaja (approved):
LGTM

PR #47184, opened 2022-10-19
remove tcp store barrier

LiYuRio:
### PR types

Performance optimization

### PR changes

Others

### Describe

remove tcp store barrier using all_reduce to barrier


FeixLiu (approved):
LGTM

PR #47189, opened 2022-10-19
[AutoParallel] fix fp16 for subblock

zhaoyinglia:
### PR types

Bug fixes
### PR changes

Others
### Describe

- get var with recursive method


aoyulong (approved):
LGTM

PR #47299, opened 2022-10-24
[BugFix] cond receives dict as input

2742195759:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fix bugs while cond receives dict as input


0x45f (approved):
LGTM

PR #47313, opened 2022-10-25
Added operator+unsqueeze2 oneDNN fuse pass 

jakpiase:
### PR types
New features

### PR changes
OPs

### Describe
Added operator+unsqueeze2 oneDNN fuse pass. Relative performance gain on U2++ model should be around ~1%. Currently only transpose2 and elementwise_mul are supported in this pass. but the support may be extended in the future.


jczaja (commented):
It seems there is no restriction on preceeding op, so why SetOutMemDescWithUnaqueeze2FuseSupport is only called from Transpose and Elementwise_xxx. How about FC+unsqueeze ?

Silv3S (dismissed):
LGTM

paddle/fluid/framework/ir/mkldnn/operator_unsqueeze2_onednn_fuse_pass.cc
```diff
@@ -0,0 +1,108 @@
...
+
+    bool has_axes_tensor =
+        std::find(names.begin(), names.end(), "AxesTensor") != names.end();
+    bool has_axes_tensor_list =
+        std::find(names.begin(), names.end(), "AxesTensorList") != names.end();
+
+    if (has_axes_tensor || has_axes_tensor_list) {
+      VLOG(3) << "Cannot fuse op with unsqueeze2, because its dims are "
+                 "specified either by AxesTensor or AxesTensorList and can be "
+                 "changed at runtime!";
```
Silv3S, line 78: I think here should be `return` to cancel fuse under this condition. Currently we display warning and fuse anyway

jakpiase: You're right, I have forgotten to add that

jakpiase:
@jczaja as we have discussed offline, the restriction on preceding ops is listed in fuse pass file, so it will work properly

jakpiase:
Restarting due to CheckPRTemplate not responding

zh794390558 (approved):
LGTM

PR #47314, opened 2022-10-25
Minor oneDNN split optimization

jakpiase:
### PR types
Function optimization

### PR changes
OPs

### Describe
Minor oneDNN split optimization. It should remove some unnecessary vector copying


jczaja (approved):
LGTM

PR #47315, opened 2022-10-25
[Eager, Performance optimization] support equal under cpp

veyron95:
### PR types

Performance optimization
### PR changes

Others
### Describe

Performance optimization


wanghuancoder (approved):
LGTM

PR #47328, opened 2022-10-25
[audio]fix split fold in tess dataset

SmileGoat:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix split fold in tess dataset. the old split way will put all wav data of a specific class into the dev set , which will cause training error.


python/paddle/audio/datasets/tess.py
```diff
@@ -132,11 +132,10 @@ def _get_data(self, mode, n_folds, split) -> Tuple[List[str], List[int]]:
 
         files = []
         labels = []
-        n_samples_per_fold = len(meta_info) // n_folds
         for idx, sample in enumerate(meta_info):
             _, _, emotion = sample
             target = self.label_list.index(emotion)
-            fold = idx // n_samples_per_fold + 1
+            fold = idx % n_folds + 1
```
ZeyuChen, line 145: 这种注意极端异常的模0情况，框架API需要主动防御这些边界错误。

ZeyuChen (approved):

python/paddle/audio/datasets/esc50.py
```diff
@@ -143,8 +143,11 @@ def __init__(
         split: int = 1,
         feat_type: str = 'raw',
         archive=None,
-        **kwargs
+        **kwargs,
     ):
+        assert split in range(
+            1, 6
```
ZeyuChen, line 149: 限制1～6是个合理的处理方式吗？

PR #47336, opened 2022-10-25
Fix a bug that print log twice

sljlp:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix a bug that print log twice


ForFishes (approved):
LGTM

PR #47350, opened 2022-10-26
[Cherry-pick][audio] fix tess split fold

SmileGoat:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix split fold in tess dataset. The old split way will put all wav data of a specific class into the dev set , which will cause training error.


ZeyuChen (approved):
LGTM

PR #47352, opened 2022-10-26
Fix multi_tensor adam/momentum bug when the parameter is list of dict

sneaxiy:
### PR types
Bug fixes

### PR changes
APIs

### Describe
When `use_multi_tensor = True` in `Adam` or `Momentum` optimizer and the `parameters` is a list of dict, the original codes were wrong because `self._param_dict` can only handle the first parameter group. The other parameter groups would reuse the `self._param_dict` from the first group, and the bug would occur.

This PR fixes this bug by adding `param_group_idx` to some method in `Optimizer`.


zhangbo9674 (approved):
LGTM

PR #47441, opened 2022-10-28
[Fluid Clean] remove paddle.fluid.dygraph.nn.conv2D

zoooo0820:
### PR types

Others
### PR changes

APIs
### Describe

To clean the code in paddle.fluid,  the useless codes should be removed and code still be used should move to a new position. 

For function `paddle.fluid.dygraph.nn.Conv2D`, since it is not used in version 2.x, should be removed. 

For unittest code where this function is used,  if it is aimed to test the old API, it can be removed directly. Otherwise, it should be replaced with `paddle.nn.Conv2D`. And please note names of parameter are slightly different.


python/paddle/fluid/tests/unittests/test_layers.py
```diff
@@ -456,182 +456,6 @@ def test_matmul(self):
         np.testing.assert_allclose(static_ret, dy_ret_value, rtol=1e-05)
         np.testing.assert_allclose(static_ret, dy_eager_ret_value, rtol=1e-05)
 
-    def test_conv2d(self):
-        with self.static_graph():
-            images = layers.data(name='pixel', shape=[3, 5, 5], dtype='float32')
-            ret = layers.conv2d(input=images, num_filters=3, filter_size=[2, 2])
```
jeff41404, line 462: shall we delete test of `layers.conv2d` this time, or with `layers.conv2d` delete ?

zoooo0820: This unittest is to check whether the results of  `layers.conv2d` and `dygraph.Conv2D` are same.  There is no need to keep this unittest here since the latter one will be removed.  


And for api `layers.conv2d`, there are `test_functional_conv2d.py`, `test_conv2d_layer.py` and `test_conv2d_op.py` to check its correctness.

jeff41404 (approved):
LGTM

python/paddle/fluid/tests/unittests/test_layers.py
jeff41404: ok, thanks

Ligoml (approved):
LGTM for docs

XiaoguangHu01 (approved):
LGTM

PR #47502, opened 2022-10-31
remove functions not belong to public-api from __all__

zoooo0820:
### PR types

Bug fixes
### PR changes

APIs
### Describe

From version 2.0，the public APIs will be add to `__all__` in each module. So the **definition of public api** in Paddle is the Class / function in `__all__`  (exclude modules belongs to `fluid`), and each of them should be maintained continuously. However, there are some unexpected functions are wrongly regarded as public apis, which we do not recommend  to use them by the path.  This PR is to remove them. 

**Note: the definition of these functions are not be removed, so the developers can still use these functions.**

The following will be removed from the public api：
```
paddle.tensor.ops.*    (should be used by paddle.xxx or paddle.Tensor.xxx)
paddle.audio.datasets,esc50.ESC50 (there is already paddle.audio.datasets.ESC50)
paddle.audio.datasets,tess.TESS (there is already paddle.audio.datasets.TESS)
paddle.distributed.commuication.stream (stream.py is a module, only callable function or Class can be API)
paddle.geometric.message_passing.* (should be used by paddle.geometric.xxx)
paddle.geometric.sampling.* (should be used by paddle.geometric.xxx)
paddle.incubate.multiprocessing.* (muiltiprocessing is a other python packages, not a function provided by Paddle)
```

### DIFF
--- a/python/paddle/audio/datasets/esc50.py
+++ b/python/paddle/audio/datasets/esc50.py
@@ -20,7 +20,7 @@ from paddle.utils import download
 from paddle.dataset.common import DATA_HOME
 from .dataset import AudioClassificationDataset
 
-__all__ = ['ESC50']
+__all__ = []
 
 
 class ESC50(AudioClassificationDataset):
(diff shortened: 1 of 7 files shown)


SmileGoat (approved):
LGTM，paddle audio

DesmonDay (approved):
LGTM

jeff41404 (approved):
LGTM

Ligoml (approved):

dingjiaweiww (approved):

PR #47497, opened 2022-10-31
fix p2p memory release logic

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
fix p2p memory release logic

PR #47517, opened 2022-11-01
fix p2p comm memory release logic (#47497)

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
fix p2p comm memory release logic


sneaxiy (approved):
LGTM.

PR #47525, opened 2022-11-01
remove paddle.fluid.layers.reduce_prod

zoooo0820:
### PR types

Others
### PR changes

APIs
### Describe

To clean the code in paddle.fluid, the useless codes should be removed and code still be used should move to a new position.

For function `paddle.fluid.layers.reduce_prod`, since it is not used in version 2.x, should be removed.

For unittest code where this function is used, if it is aimed to test the old API, it can be removed directly. Otherwise, it should be replaced with `paddle.prod`.


jeff41404 (dismissed):
LGTM

Ligoml (dismissed):

Ligoml (approved):

jeff41404 (approved):

PR #47573, opened 2022-11-02
[Control Flow] replace executor in while op with InterpreterCore

kangguangli:
### PR types

Others
### PR changes

OPs
### Describe

The main content of this pr is to replace executor in run method of while_op with standalone_executor. 
The changes include:
  1. Replace Executor in RunImpl with InterpreterCore
  2. Add class member to cache for both Executor and InterpreterCore
  3. Use flag FLAGS_control_flow_use_new_executor to decide whether use InterpreterCore
  4. By default, InterpreterCore will rebuild and reset its scope. 
  5. The assign op in while loop may change inputs' places, which violates the hypothesis that all kernels can be fixed in InterpreterCore.  So we simply record inputs' places at beginning and restore after execution of InterpreterCore.

#### Other Changes
 1. extract the `BuildScopeForControlFlowOp` function in an individual header file.

#### Relevant Pr
#45696 replace executor in `conditional_block_op`


2742195759 (dismissed):
LGTM

paddle/fluid/operators/controlflow/while_op.cc
```diff
@@ -445,6 +582,11 @@ class WhileGradOp : public framework::OperatorBase {
     }
     step_scopes->clear();
   }
+
+ private:
+  mutable std::shared_ptr<framework::Executor> executor{nullptr};
```
zhiqiu:
```suggestion
  mutable std::shared_ptr<framework::Executor> executor_{nullptr};
```
same for others.

kangguangli: done

zhiqiu (dismissed):
LGTM

zhiqiu (approved):
LGTM

PR #47576, opened 2022-11-02
Define ConvRunner to wrapper the call of cudnn conv functions.

Xreki:
### PR types

Function optimization

### PR changes

OPs

### Describe

定义`ConvRunner`，对cudnn conv计算函数进行封装，方便后续依据cache统一设置math type、支持try-catch和fallback。


JamesLim-sy (approved):
LGTM

PR #47584, opened 2022-11-02
fix link order

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix link order

`libpaddle_gtest_main_new.a` uses symbols in `libpaddle.so`, so `libpaddle.so` should be added after `libpaddle_gtest_main_new.a`, otherwise, `undefined reference` may occur.


zhangbo9674 (approved):
LGTM

PR #47613, opened 2022-11-02
Fix cublasGemmStridedBatchedEx compute_type bug in CUDA >= 11.0

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Describe
Since CUDA 11.0+, `cublasGemmStridedBatchedEx` has 2 versions as follows (see https://docs.nvidia.com/cuda/archive/11.0/cublas/index.html#cublas-GemmStridedBatchedEx), but in CUDA < 11.0, it has only the second version one (see https://docs.nvidia.com/cuda/archive/10.2/cublas/index.html#cublas-GemmStridedBatchedEx). This is an incompatible issue caused by CUDA itself. 

```
cublasStatus_t cublasGemmStridedBatchedEx(cublasHandle_t handle,
                            cublasOperation_t transa,
                            cublasOperation_t transb,
                            int m,
                            int n,
                            int k,
                            const void    *alpha,
                            const void     *A,
                            cudaDataType_t Atype,
                            int lda,
                            long long int strideA,
                            const void     *B,
                            cudaDataType_t Btype,
                            int ldb,
                            long long int strideB,
                            const void    *beta,
                            void           *C,
                            cudaDataType_t Ctype,
                            int ldc,
                            long long int strideC,
                            int batchCount,
                            cublasComputeType_t computeType,
                            cublasGemmAlgo_t algo)

#if defined(__cplusplus)
cublasStatus_t cublasGemmStridedBatchedEx(cublasHandle_t handle,
                            cublasOperation_t transa,
                            cublasOperation_t transb,
                            int m,
                            int n,
                            int k,
                            const void    *alpha,
                            const void     *A,
                            cudaDataType Atype,
                            int lda,
                            long long int strideA,
                            const void     *B,
                            cudaDataType Btype,
                            int ldb,
                            long long int strideB,
                            const void    *beta,
                            void           *C,
                            cudaDataType Ctype,
                            int ldc,
                            long long int strideC,
                            int batchCount,
                            cudaDataType computeType,
                            cublasGemmAlgo_t algo)
#endif
```

Regarding the PaddlePaddle framework, we use `dlopen + dlsym` to query the `cublasGemmStridedBatchedEx` API from the dynamic library (i.e., `libcublas.so`). We get the first version one but not the second version one. However, we use the parameter list as the second version one when calling `cublasGemmStridedBatchedEx`. This is a bug that needs to be fixed.

This bug does not occur in PyTorch, because it does not use the `dlopen + dlsym` way to load the `cublasGemmStridedBatchedEx` API.

### DIFF
--- a/paddle/phi/kernels/funcs/blas/blas_impl.cu.h
+++ b/paddle/phi/kernels/funcs/blas/blas_impl.cu.h
@@ -1452,7 +1452,11 @@ void Blas<phi::GPUContext>::BatchedGEMM(CBLAS_TRANSPOSE transA,
             << FLAGS_gemm_use_half_precision_compute_type;
 
     auto fp = std::is_same<T, float>::value ? CUDA_R_32F : CUDA_R_16F;
-    cudaDataType_t compute_type = CUDA_R_32F;
+#if CUDA_VERSION >= 11000
+    auto compute_type = CUBLAS_COMPUTE_32F;
+#else
+    auto compute_type = CUDA_R_32F;
+#endif
 
     float h_alpha = static_cast<float>(alpha);
     float h_beta = static_cast<float>(beta);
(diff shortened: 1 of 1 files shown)


FeixLiu (approved):
LGTM

ZzSean (approved):
LGTM for CI-OP-Benchmark

PR #47616, opened 2022-11-02
Fix matmul_elementwise_add_mkldnn_fuse_pass for MKLDNN

yeliang2258:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix matmul_elementwise_add_mkldnn_fuse_pass for MKLDNN


yeliang2258:
```
git clone https://github.com/PaddlePaddle/PaddleOCR.git
cd PaddleOCR
git checkout dygraph
mkdir models_inference # then download model at https://ecloud.baidu.com?t=96715119ba2b879fa4408e999b07d58f
python tools/infer/predict_rec.py --image_dir="./doc/imgs_words_en/word_336.png" --rec_model_dir="./models_inference/"rec_resnet_rfl_att --rec_image_shape=1,32,100 --rec_algorithm=RFL --rec_char_dict_path=./ppocr/utils/ic15_dict.txt --use_gpu=False --use_tensorrt=0 --enable_mkldnn=True --use_space_char=False
```
Model is here：https://ecloud.baidu.com?t=96715119ba2b879fa4408e999b07d58f

image

yeliang2258:
Hi @wozna @jczaja @Silv3S     Here I found that when ResidualData is a constant, the result is right, and when it is non-const, the result is wrong, I am not sure if matmul_mkldnn_kernel is implemented with a bug

Silv3S:
Hi @yeliang2258, could you please explain what does it mean that ResidualData is constant? 

If there is bug in fuse pass, we have to adjust unit tests to skip these edge cases.

Silv3S (dismissed):

wozna (dismissed):
LGTM

yeliang2258:
Hi @Silv3S @wozna  I mean when the ResidualData tensor can be found in param_scope(), the result is right, but when it is not found, the result is wrong.  

I'm not sure if my fix is correct. It may be a problem with the mutmul kernel, or a problem with the onednn library, please take a look with me, thanks.

The result of the node below is correct
image

The result of the node below is wrong
image

yeliang2258:
@Silv3S @wozna  In addition, I found that we use handler.AcquireSrcMemory(residual_data) to get the memory of ResidualData in the kenel of matmul, but the memory::desc of ResidualData and X are different, wouldn't this lead to wrong results?
image

PR #47652, opened 2022-11-03
[PHI] Migrate batch_norm

Silv3S:
### PR types
Others

### PR changes
Others

### Describe
- migrate FWD `batch_norm` kernel to PHI


Silv3S:
Hi @chenwhql
I have two questions regarding `batch_norm` kernel:
1. Should `batch_norm_grad` also be migrated to PHI?
2. I have problem with kernel registration. As stated in *batch_norm_sig.cc* either `batch_norm_infer` or `batch_norm` kernel is chosen depending on available arguments. I implemented and migrated `batch_norm`, but when trying to register oneDNN version of `batch_norm_infer` in */phi/kernels/batch_norm_kernel.cc* code doesn't build. There is problem with undefined reference to `void phi::EmptyLikeKernel<float, phi::OneDNNContext>`. Could you please advice what could be problem here?

YuanRisheng (approved):

chenwhql (approved):

PR #47761, opened 2022-11-08
Replace built-in print with logger in distributed_strategy.py

sljlp:
### PR types

Others
### PR changes

Others
### Describe

Replace built-in print with logger in distributed_strategy.py


OccupyMars2025:
image
The fatal message seems to show that this CI failed because of a network error. Maybe it can be fixed by retriggering this CI. Just my personal understanding.

FeixLiu (approved):

PR #47764, opened 2022-11-08
[Dygraph] Support grad division to nranks before reduce in sharding stage2

haohongxiang:
### PR types

Bug fixes
### PR changes

Others
### Describe

[Dygraph] Support grad division to nranks before reduce in sharding stage2

GPT模型loss曲线对比，梯度累加 + FP16 + (DP vs 修改后的sharding stage2)：
image


ForFishes (approved):
LGTM

PR #47767, opened 2022-11-08
[Phi MKLDNN] Support mkldnn kernel in phi

jiahy0825:
### PR types
New features

### PR changes
OPs

### Describe
Support MKLDNN kernels in phi, this PR supports activation kernels for example.


jiahy0825:
Information update: There is no need to support mkldnn kernels in phi, thus close this PR.

PR #47780, opened 2022-11-08
Move dropout removing pass at the beginning of the INT8 process

wozna:
### PR types

Bug fixes
### PR changes

Others 
### Describe

It turn out that pass `simplify_with_basic_ops_pass` that is removing dropout was placed after `quant_dequant_mkldnn_pass`. 
So if the dropout was before some quantizable operator, the scale to it [tensor_name, scale] was lost because the tensor_name was changing. The simplest solution is to move this pass before quant_dequant_mkldnn_pass. 

It improves U2++ int8 model: 
- in encoder model before only 61 fc was quantized and right now 85 fc is quantized. 
- in decoder model before only 39 was quantized and right now 44 fc is quantized. 
This should give us almost a 12% performance improvement
I am still checking whether it will affect other models.

### DIFF
--- a/paddle/fluid/inference/api/paddle_pass_builder.cc
+++ b/paddle/fluid/inference/api/paddle_pass_builder.cc
@@ -390,9 +390,9 @@ void CpuPassStrategy::EnableMkldnnInt8() {
 #ifdef PADDLE_WITH_MKLDNN
   if (!use_mkldnn_int8_) {
     passes_.clear();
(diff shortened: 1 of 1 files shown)


jczaja (approved):
LGTM

jakpiase (approved):
LGTM, awesome job!

yeliang2258 (approved):
LGTM

zh794390558 (approved):
LGTM

PR #47844, opened 2022-11-10
Modified mem_desc() to return reference to Tensor::memory::desc to avoid copying

jczaja:
### PR types
 Others 

### PR changes
Others 

### Describe
Previously value of memory descriptor was returned from Tensor::mem_desc() . Which could result in unnecessery copying of 400 bytes. So this PR is to make mem_desc() return const reference. This should be safe as tensors live longer than operators where mem_desc() returned value is used.


onecatcn:
@chenwhql please take a look at the pr

chenwhql (approved):

PR #47846, opened 2022-11-10
[Auto Parallel] Recompute Support New Graph Executor

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Describe
The origin recompute strategy has a strong assumption that all operators will be issued by the order in ProgramIR where it takes the advantage of implicitly dependencies in ProgramIR .
This assumption is failed when switch to Standalone Executor (Graph based), we need to add (and ONLY add) necessary  dependencies for it.

Model:GPT3-1.3B | Mem mb | speed step/s
-- | -- | --
old exe | 23062| 1.463 
new exe + Naive Redundant Sequential Deps  |22916|1.503 
new exe + PR |23278 | 1.526


aoyulong (approved):
LGTM

PR #47882, opened 2022-11-11
correct sync behavior for XPU distributed training

XiaociZhang:
### PR types

Others

### PR changes

Others

### Describe

XPU support event mechanism similar to cuda event, so it is advisable to use an event to sync compute/comm streams for performance. However this mechanism is never fully tested, and inconsistent loss/ending_epochs are reported. Therefore, this PR replaces event sync with stream waiting as a temporary solution.


ykkk2333 (approved):
LGTM

paddle/phi/backends/xpu/xpu_context.cc
```diff
@@ -64,7 +64,7 @@ struct XPUContext::Impl {
       // manually destroy XPUStream here until xpu::api integrates this work
       // into Context dtor
       xpu_wait(context_->xpu_stream);
-      PADDLE_ENFORCE_XPU_SUCCESS(xpu_stream_destroy(context_->xpu_stream));
+      xpu_stream_destroy(context_->xpu_stream);
```
QingshuChen, line 67: 为什么去掉返回值判断？

QingshuChen (approved):

paddle/phi/backends/xpu/xpu_context.cc
XiaociZhang: 编译器会complain在析构函数里抛异常，（这里其实有个try catch所以运行时倒不会出现程序异常终止，但还是可能会导致内存泄露）
最佳方法还是直接不去判断，如果xpu_stream_destroy出错可以依赖runtime的log

PR #47897, opened 2022-11-11
support fp16 index sample

wangxn12138:
### PR types

New features
### PR changes

OPs
### Describe

support index_sample OP fp16 input.


AndPuQing:
Hello, I also have a pr https://github.com/PaddlePaddle/Paddle/pull/47646 trying to support flout16 type input, but after benchmark, I found that the flout16 data type will seriously affect the reverse performance of index_sample. Do you plan to do any performance optimization?

Thunderbrook (approved):

PR #47910, opened 2022-11-13
Do not release memory cache after build_op_func_list in interpretercore

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

PR https://github.com/PaddlePaddle/Paddle/pull/46670 releases memory cache after build_op_func_list to avoid memory fragments in V100 8-GPU training of transformer_base_bs4096_amp_fp16 and transformer_base_bs4096_pure_fp16 models. However, this optimization method is not general that leads to a memory occupancy growth in other models such as A100 single-GPU training of deeplabv3 and SE_ResNeXt50_32x4d_bs32.
This PR tries to revert https://github.com/PaddlePaddle/Paddle/pull/46670. A more general fix-up method that avoid running kernel in build_op_func_list is ongoing.


zhiqiu (approved):
LGTM

PR #47975, opened 2022-11-14
fix: multihead matmul biasqk broadcast support for [1,1,seq,seq] shape

wwbitejotunn:
### PR types
Function optimization

### PR changes
OPs

### Describe
Add multihead_matmul biasqk broadcast support from (1,1,seq,seq) shape to fix the accuracy problem in paddle nlp stable diffusion.


heavengate (approved):

kolinwei (approved):

Xreki (approved):
LGTM

PR #48061, opened 2022-11-16
[PHI] Migrate mul_grad kernel

Silv3S:
### PR types
Others

### PR changes
Others

### Describe
- migrate `mul_grad` kernel to PHI


Silv3S:
@PaulinaGacek, @sfraczek please review

sfraczek (approved):
LGTM

paddle/phi/backends/onednn/onednn_reuse.h
```diff
@@ -1912,6 +1912,47 @@ class MatmulOneDNNHandler
   }
 };
 
+template <typename T>
+static void ExecuteMul(const OneDNNContext& dev_ctx,
```
paulinagacek, line 1916: I know it's nitpicking but shouldn't it be named `ExecuteMatMul` as it is in fluid version instead of `ExecuteMul` ?

paulinagacek (approved):

paddle/phi/backends/onednn/onednn_reuse.h
Silv3S: In fluid we had `mul`, `matmul` and `matmul_v2`. Now both `matmul_v2` and `mul` are migrated to PHI as matmul kernels and `matmul(v1)` stays in fluid.  In PHI FWD and BWD kernels are separated, so it raises conflicts in methods naming, which could be declared locally in fluid. `ExecuteMatmul` is reserved for handling `matmul_v2` so I think `ExecuteMul` is right name for this kernel and we can do further refactors after all kernels are migrated.

paulinagacek: That makes sense, thanks!

YuanRisheng (approved):

PR #48090, opened 2022-11-17
[CustomPlace] fix amp

Aganlengzi:
### PR types
Others

### PR changes
Others

### Describe
using amp on custom device, conv2d input should be casted but not, which should be fixed.
image
matmul float16 kernel is too slow, so we put it into amp blacklist.


qili93 (dismissed):
LGTM

zhangbo9674 (dismissed):
LGTM

ronny1996 (approved):

PR #48150, opened 2022-11-18
Adjust mkldnn_placement_pass to check library type and data type

wozna:
### PR types

Others 
### PR changes

Others 
### Describe

Previously in mkldnn_placement_pass use_mkldnn was only set based on whether there is a `use_mkldnn` attribute. I changed it to check if there is an OneDNN kernel registered with the appropriate datatype and only, in this case, `use_mkldnn` is set to True.   

In addition, I added an option to exclude some operators from using MKLDNN `SetMKLDNNExcludedOps`, this can be very useful in some cases for not very well known models.

### DIFF
--- a/paddle/fluid/pybind/inference_api.cc
+++ b/paddle/fluid/pybind/inference_api.cc
@@ -868,7 +868,7 @@ void BindAnalysisConfig(py::module *m) {
            )DOC")
 #endif
       .def("set_mkldnn_op", &AnalysisConfig::SetMKLDNNOp)
-      .def("set_mkldnn_excluded_op", &AnalysisConfig::SetMKLDNNExcludedOp)
+      .def("set_mkldnn_excluded_ops", &AnalysisConfig::SetMKLDNNExcludedOps)
       .def("set_model_buffer", &AnalysisConfig::SetModelBuffer)
       .def("model_from_memory", &AnalysisConfig::model_from_memory)
       .def("delete_pass",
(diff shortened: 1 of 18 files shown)


jczaja (dismissed):
LGTM

paddle/fluid/pybind/inference_api.cc
```diff
@@ -830,6 +830,7 @@ void BindAnalysisConfig(py::module *m) {
            )DOC")
 #endif
       .def("set_mkldnn_op", &AnalysisConfig::SetMKLDNNOp)
+      .def("set_mkldnn_excluded_op", &AnalysisConfig::SetMKLDNNExcludedOp)
```
sfraczek: Maybe SetMKLDNNExcludedOpTypes the same like in AnalysisPredictor would be good? Or a plural ExcludedOps?

wozna:
@qingqing01 @yaomichael  I added API option to exclude ops from mkldnn `SetMKLDNNExcludedOps` this can be very useful in some cases for not very well known models. Could you check it?

yeliang2258 (dismissed):
LGTM

sfraczek (dismissed):
LGTM

jczaja (dismissed):
LGTM

jczaja (approved):

wozna:
I am going to split this PR into two separate PRs: 
1. https://github.com/PaddlePaddle/Paddle/pull/49899
2. TODO

PR #48184, opened 2022-11-21
Fix Ctx Dev pointer for KUNLUN

sljlp:
### PR types

Bug fixes
### PR changes

Others
### Describe

return pointer rather than refernce


haohongxiang (approved):
LGTM

PR #48226, opened 2022-11-22
remove paddle.fluid.layers.nn.shuffle_channel

zhengqiwen1997:
### PR types

Others
### PR changes

APIs
### Describe

To clean the code in paddle.fluid, the useless codes should be removed and code still be used should move to a new position.
For function paddle.fluid.layers.nn.shuffle_channel, since it is not used in version 2.x, should be removed.
For unittest code where this function is used, if it is aimed to test the old API, it can be removed directly. Otherwise, it should be replaced with paddle.nn.functional.channel_shuffle.


zoooo0820 (approved):
LGTM

jeff41404 (approved):
LGTM

Ligoml (approved):

PR #48238, opened 2022-11-22
Make bfloat16 implicitly convert to float/double

sneaxiy:
### PR types
Function optimization

### PR changes
Others

### Describe
Make `bfloat16` convert to `float` and `double` implicitly. This conversion is safe without any accuracy drop.


FeixLiu (approved):
LGTM

PR #48255, opened 2022-11-22
fix vector out of range error

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix vector out of range error

When ins.size() == 1, the dims_simplifier.in_dims.size() is 2.

in dims_simplifier.h
```
    N = std::max(static_cast<int>(ins.size()), 2);
    in_dims.resize(N);
```


Xreki (approved):
LGTM

PR #48271, opened 2022-11-23
[FRAMEWORK] add id for OperatorBase

kuizhiqing:
### PR types

Others

### PR changes

OPs

### Describe

add id attribute for OperatorBase.

### DIFF
--- a/paddle/fluid/framework/operator.h
+++ b/paddle/fluid/framework/operator.h
@@ -278,7 +278,7 @@ class OperatorBase {
   const OpInfo* info_;
 
   // OpDesc Id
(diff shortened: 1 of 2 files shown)


paddle/fluid/framework/operator.h
```diff
@@ -273,6 +277,9 @@ class OperatorBase {
   // OpInfo
   const OpInfo* info_;
 
+  // OpDesc Id
+  uint64_t id_;
```
sneaxiy: How about give a default value, like `-1` (also change `uint16_t` to `int64_t`)? In this way, we can know this is not initialized.

PR #48276, opened 2022-11-23
Use phi layernorm in fused_dropout_helper

MARD1NO:
### PR types
Performance optimization
### PR changes
OPs
### Describe
Use phi layernorm Kernel in fused_dropout_helper.h, it's performance is better


qingqing01 (approved):
LGTM

PR #48295, opened 2022-11-23
[AutoParallel] dist_scale

zhaoyinglia:
### PR types

New features
### PR changes

Others
### Describe

- add dist_scale op


JZ-LIANG (approved):
LGTM for emergency case in fleetx, should revise this logic to auto completion in future.

PR #48348, opened 2022-11-24
[API] Add Reshape layer API

minghaoBD:
### PR types
Others

### PR changes
APIs

### Describe
Add Reshape Layer:

paddle.nn.functional.reshape(x, shape=[])
paddle.nn.Reshape(shape=[], name='None')

### DIFF
--- a/python/paddle/nn/functional/common.py
+++ b/python/paddle/nn/functional/common.py
@@ -223,15 +223,6 @@ def reshape(x, shape, name=None):
     act = None
     inplace = False
 
-    def _contain_var(list_or_tuple):
-        """
-        Check whether list or tuple contains variable.
-        """
-        for item in list_or_tuple:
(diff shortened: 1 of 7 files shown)


python/paddle/nn/functional/common.py
```diff
@@ -175,6 +177,163 @@ def unfold(x, kernel_sizes, strides=1, paddings=0, dilations=1, name=None):
     return out
 
 
+def reshape(x, shape, name=None):
```
qingqing01, line 180: 这个和 paddle.tensor.reshape重复  https://github.com/PaddlePaddle/Paddle/blob/505f4100a02ebbc410136194e0816a679adcd8d5/python/paddle/tensor/manipulation.py#L3507

minghaoBD: 好的，关闭了

PR #48379, opened 2022-11-25
[Paddle Inference]fix loopup_table plugin deserialize size error

zhangxin81:
### PR types
Bug fixes

### PR changes
Others

### Describe
use getElementSize to get type size instead of sizeof which may case diff in size between serialize and deserialize in LookupTablePluginDynamic


Wangzheee (approved):
LGTM

PR #48386, opened 2022-11-25
fix mma_tensorcore 

carryyu:
### PR types

Bug fixes
### PR changes

OPs
### Describe

disable tensorcore by default, because the judgment of __CUDA_ARCH__ will cause undefined behavior in some environments, can manually enable it on a machine that supports tensorcore.


Aganlengzi (approved):
LGTM

PR #48409, opened 2022-11-26
[Opt Kernel Selection] Regist kernel_bs_ in KernelFactory, opt kernel selection

jiahy0825:
### PR types
Performance optimization

### PR changes
Others

### Describe
Regist `kernel_bs_` in KernelFactory, opt kernel selection


jiahy0825:
Cannot search backend once, a fallback mechanism is necessary. For example, different backends have various datatypes registered in paddle, as shown in the following error message, this scenario needs fallback mechanism: 

> RuntimeError: (NotFound) The kernel with key (GPU, NCHW, uint8) of kernel `numel` is not registered. Selected wrong DataType `uint8`. Paddle support following DataTypes: float16, float64, int16, int32, int64, bfloat16, bool, float32. (at /work/dev_op/Paddle/paddle/phi/core/kernel_factory.cc:198)

PR #48450, opened 2022-11-28
 remove paddle.static.sparsity and move out sparsity from fluid

GGBond8488:
### PR types
 Others 

### PR changes
APIs

### Describe
In the current paddle version, there are two paths pointing to the same api (sparsity), one of which is paddle.incubate.asp.sparsity, and the other is paddle.static.sparsity, which is obviously unreasonable, so we carefully After discussion and consideration, it is considered that sparsity is not suitable for regularization at present.

So we removed paddle.static.sparsity,kept paddle.incubate.asp.sparsity.  And moved the real implementation of sparsity to paddle.incubate.asp.sparsity


zoooo0820 (approved):
LGTM

zkh2016 (approved):

Ligoml (approved):
LGTM for docs

jeff41404 (approved):
LGTM

zyfncg (approved):

XiaoguangHu01 (approved):
LGTM

PR #48471, opened 2022-11-28
[Eager] fix recompute for stop_gradient and inpalce

sljlp:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix recompute for stop_gradient and inpalce.

In some cases a subset of outputs is identity to the subset of inputs, which is inplace operating. When the inputs' stop_gradient is True, an error will occurs because the stop_gradient=True and inpalce-op are not supported at the same time. The solution is to mark the inputs non_differentiable if its stop_gradient is True.
- Note: if not marked non_differentiable, all output tensors'
attribute `stop gradient` will be reset to `False` in c++ backend . code: https://github.com/PaddlePaddle/Paddle/blob/9d62efb0e6e5373823039d9eda96cd5905426c0a/paddle/fluid/pybind/eager_py_layer.cc#L388

### DIFF
--- a/python/paddle/distributed/fleet/recompute/recompute_hybrid.py
+++ b/python/paddle/distributed/fleet/recompute/recompute_hybrid.py
@@ -151,7 +151,7 @@ class _HPRecomputeFunction(PyLayer):
                 ctx.tensor_indices.append(i)
                 ctx.inputs.append(None)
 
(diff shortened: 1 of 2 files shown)


LiYuRio (approved):
LGTM

PR #48513, opened 2022-11-29
[Fluid clean]Migrate base/call/print et.al transformer into paddle.jit

Aurelius84:
### PR types

Others
### PR changes

Others
### Describe

[Fluid clean]Migrate base/call/print et.al transformer into paddle.jit


wanghuancoder (approved):
LGTM

0x45f (approved):
LGTM

PR #48537, opened 2022-11-29
onednn reshape fix to match CPU reshape kernel behaviour

jczaja:
### PR types
Others

### PR changes
OPs

### Describe
ShapeTensor vector should contain Tensors holding int32 data, but it happens that int64 is there. So CPU kernel is converting 
all Tensor data to int32 with IntArray so the same behaviour was introduced to reshape onednn kernel..


chenwhql (dismissed):

jczaja:
This PR is no longer needed as conversion to PHI sorted out problemes adressed in this PR

PR #48652, opened 2022-12-02
Support elementwise_add operator for QAT in dynamic mode

LiuChiachi:
### PR types

New features
### PR changes

APIs
### Describe

Supports `elementwise_add` operator for QAT in dynamic mode

usage:
Step 1. For layer definition
```python
class MyLayer(Layer):
    def __init__(self):
        self.add = paddle.nn.quant.add()
    def forward(self, x, y):
        # Before: sum = x + y
        sum = self.add(x, y)
        return sum
```
Step2. For QAT

```python

from paddleslim import QAT

quant_config = {
      'activation_preprocess_type':
      args.activation_preprocess_type,  # PACT or None
      'weight_preprocess_type':
      args.weight_preprocess_type,  # PACT or None
      'weight_quantize_type':
      args.weight_quantize_type,
      'activation_quantize_type':
      'moving_average_abs_max' if args.activation_quantize_type is None else
      args.activation_quantize_type,
      'weight_bits':
      8,
      'activation_bits':
      8,
      'dtype':
      'int8',
      'window_size':
      10000,
      'quantizable_layer_type': ['Linear', 'Matmul', 'Add'], # 'Add' should be added here
      'moving_rate':
      args.moving_rate,
      'onnx_format': False, # QAT+PTQ only supports False now.
  }

quanter = QAT(config=quant_config)

quanter.quantize(model)
```

### DIFF
--- a/python/paddle/fluid/contrib/slim/quantization/imperative/utils.py
+++ b/python/paddle/fluid/contrib/slim/quantization/imperative/utils.py
@@ -46,6 +46,8 @@ layer_name_map = {
     'BatchNorm': paddle.nn.BatchNorm,
     'GroupNorm': paddle.nn.GroupNorm,
     'LayerNorm': paddle.nn.LayerNorm,
+    'Matmul': paddle.nn.quant.matmul,
+    'Add': paddle.nn.quant.add,
 }
 
 # Apply fake quant for the inputs of these layers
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/contrib/slim/quantization/imperative/utils.py
```diff
@@ -46,13 +46,17 @@
...
+    'Add': paddle.nn.quant.add,
 }
 
 # Apply fake quant for the inputs of these layers
 fake_quant_input_layers = [
     paddle.nn.Conv2D,
     paddle.nn.Linear,
     paddle.nn.Conv2DTranspose,
+    paddle.nn.quant.add,
+    paddle.nn.quant.matmul,
```
yghstill, line 59: fake_quant_input_layers里可不加add和matmul，开发者如果想给add输入插fake quant，可直接调用`paddle.nn.quant.QuantizedAdd()` 。无需配置其他参数

PR #48668, opened 2022-12-02
[DataParallel] Fix bug of DataParallel in best_fit allocator strategy

ForFishes:
### PR types

Bug fixes
### PR changes

Others
### Describe

The best_fit allocator strategy is multi-stream insecure. In the Split operator, additional memory will be applied for calculation, and if it is asynchronous, an illegal memory access may be encountered.

image


haohongxiang (approved):
LGTM

PR #48702, opened 2022-12-04
Register exp/expm1/logit bf16 activation op kernels

sneaxiy:
### PR types
Function optimization

### PR changes
OPs

### Describe
Register `bfloat16` kernel for `exp`, `expm1` and `logit`.


FeixLiu (approved):
LGTM

PR #48730, opened 2022-12-05
fix static check ci

zhwesky2010:
### PR types

Others
### PR changes

Others
### Describe

fix static check error, avoid git fetch upstream develop because it may result in api diff is error.


luotao1 (approved):
LGTM

Issue #45308, opened 2022-08-22
Migrate oneDNN operators to under PHI


yaomichael:
### 需求描述 Feature Description

design doc: https://github.com/PaddlePaddle/docs/blob/develop/docs/design/phi/design_en.md

with Op name (op_type) as the unit, MKLDNN currently has a total of 100 OpKernels, which we divide into three categories:
1. Migrate first: a total of 53 (42 to be migrated, 11 have been migrated by YuanRisheng). The parameters required by this type of Kernel are consistent with the CPU&GPU Kernel. Currently, the migration is non-blocking. In order to verify the mechanism, we have migrated 11 of them;
2. Migrate later: 23 in total. This kind of Kernel has the problem of Extra parameters in OpMaker that we mentioned before. It requires more parameters than CPU&GPU Kernel, so this part of Kernel is recommended to be migrated after we complete the extra parameter cleaning. The specific time will be tell you later (expectedly in mid to late September);
3. No need to migrate: 24 in total. This type of Kernel mainly includes two types of operators, one is the historical operators that have been abandoned, such as reshape, flatten, etc, these abandoned operators will be gradually removed from the framework, so they will no need to be migrated; the other is framework function related  operators. For example, related to fusion or quantize, such operators may not be migrated to PHI in the future, and may continue to be maintained in the main framework.

yaomichael:
@chenwhql  @jiangjiajun @jczaja  @wozna

Silv3S:
Merged
- [x] `abs`, `abs_grad`, `relu6` https://github.com/PaddlePaddle/Paddle/pull/45397
- [x] `gaussian_random` https://github.com/PaddlePaddle/Paddle/pull/45481
- [x] Moving `fluid/mkldnn_reuse.h` to `phi/onednn_reuse.h` https://github.com/PaddlePaddle/Paddle/pull/45626
- [x] `scale` https://github.com/PaddlePaddle/Paddle/pull/45537
- [x] `reduce_sum`, `reduce_sum_grad`, `reduce_mean`, `reduce_mean_grad`, `reduce_min`, `reduce_max` https://github.com/PaddlePaddle/Paddle/pull/45536
- [x] `cast`, `clip`, `clip_grad` https://github.com/PaddlePaddle/Paddle/pull/45775
- [x] `shape` https://github.com/PaddlePaddle/Paddle/pull/46051
- [x] `softmax_grad` https://github.com/PaddlePaddle/Paddle/pull/46257
[...]

PR #48749, opened 2022-12-05
[PHI] Migrate reshape kernel

Silv3S:
### PR types
Others

### PR changes
Others

### Describe
- migrate reshape kernels (with and without XShape) to PHI.

### DIFF
--- a/paddle/fluid/operators/mkldnn/reshape_mkldnn_op.cc
+++ b/paddle/fluid/operators/mkldnn/reshape_mkldnn_op.cc
@@ -100,9 +100,6 @@ class ReshapeMKLDNNKernel : public framework::OpKernel<T> {
       case ReshapeKernelOpName::reshape:
         InferShapeReshapeOp(ctx, x_dims, out_dims);
         break;
-      case ReshapeKernelOpName::reshape2:
-        InferShapeReshape2Op(ctx, x_dims, out_dims);
-        break;
       case ReshapeKernelOpName::squeeze:
         InferShapeSqueezeOp(ctx, x_dims, out_dims);
         break;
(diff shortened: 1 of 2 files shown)


Silv3S:
Hi @YuanRisheng, @chenwhql
Could you please advise on these two issues?
1. There are [Windows errors](https://xly.bce.baidu.com/paddlepaddle/paddle/newipipe/detail/7344401/stage/12672514/0?jump=latestJob) which I can't reproduce locally. Probably it's related to adding `typename T`, because *reshape_kernel.h* has only `template <typename Context>` and it's registered for all data types. Is there any other solution to register oneDNN kernel only for fp32 and bf16, without introducing this extra argument? 
2. `reshape2_grad` would be migrated to PHI, but *reshape_grad_kernel.h* doesn't have declaration of kernel with XShape. Should we migrate only `reshape(1)_grad` kernel instead?

YuanRisheng:
> Hi @YuanRisheng, @chenwhql Could you please advise on these two issues?
> 
> 1. There are [Windows errors](https://xly.bce.baidu.com/paddlepaddle/paddle/newipipe/detail/7344401/stage/12672514/0?jump=latestJob) which I can't reproduce locally. Probably it's related to adding `typename T`, because _reshape_kernel.h_ has only `template <typename Context>` and it's registered for all data types. Is there any other solution to register oneDNN kernel only for fp32 and bf16, without introducing this extra argument?
> 2. `reshape2_grad` would be migrated to PHI, but _reshape_grad_kernel.h_ doesn't have declaration of kernel with XShape. Should we migrate only `reshape(1)_grad` kernel instead?

For Point1, you can try delete header `reshape_kernel.h` in `reshape_kernel.cc`. Maybe the difference between function declaration in `reshape_kernel.h` and the function definition in `reshape_kernel.cc` cause confusion.
For Point2, you don't have to migrate `reshape(1)_grad`. It is right that _reshape_grad_kernel.h_ doesn't have declaration of kernel with XShape. The XShape is used when run InferMeta and it is not used in _reshape_grad_kernel.h_

Silv3S:
> > Hi @YuanRisheng, @chenwhql Could you please advise on these two issues?
> > 
> > 1. There are [Windows errors](https://xly.bce.baidu.com/paddlepaddle/paddle/newipipe/detail/7344401/stage/12672514/0?jump=latestJob) which I can't reproduce locally. Probably it's related to adding `typename T`, because _reshape_kernel.h_ has only `template <typename Context>` and it's registered for all data types. Is there any other solution to register oneDNN kernel only for fp32 and bf16, without introducing this extra argument?
> > 2. `reshape2_grad` would be migrated to PHI, but _reshape_grad_kernel.h_ doesn't have declaration of kernel with XShape. Should we migrate only `reshape(1)_grad` kernel instead?
> 
> For Point1, you can try delete header `reshape_kernel.h` in `reshape_kernel.cc`. Maybe the difference between function declaration in `reshape_kernel.h` and the function definition in `reshape_kernel.cc` cause confusion. For Point2, you don't have to migrate `reshape(1)_grad`. It is right that _reshape_grad_kernel.h_ doesn't have declaration of kernel with XShape. The XShape is used when run InferMeta and it is not used in _reshape_grad_kernel.h_

1. Thank you very much for another great idea 👍 I checked unit test manually and `reshape` kernel is executed correctly in _test_reshape_mkldnn_op.py_
2. Ok, so I will remove `reshape2_grad` from TODO list https://github.com/PaddlePaddle/Paddle/issues/45308

YuanRisheng (approved):

PR #48803, opened 2022-12-06
Fix accuracy fp16 kernel return fp32 tensor error

0x45f:
### PR types

Bug fixes
### PR changes

OPs
### Describe

Fix accuracy fp16 kernel return fp32 tensor error


Aurelius84 (approved):
LGTM

PR #48819, opened 2022-12-07
test_convert_to_mixed_precision.py use tempfile for temporary models/params

jiweibo:
### PR types

Others
### PR changes

Others
### Describe

fix ut.

### DIFF
--- a/python/paddle/fluid/tests/unittests/ir/test_convert_to_mixed_precision.py
+++ b/python/paddle/fluid/tests/unittests/ir/test_convert_to_mixed_precision.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+import tempfile
 import unittest
 
 import paddle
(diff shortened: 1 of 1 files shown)


zlsh80826:
I am wondering if this unit test can be benefit by [unittest subtest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests)? `setUp` and `tearDown` will be called for each `test_xxx`, it means this test will repeat `setUp/tearDown` four times (for `test_convert_to_fp16`, `test_convert_to_fp16_with_fp16_input`, `test_convert_to_fp16_with_blacklist`, `test_convert_to_bf16`)

jiweibo:
> I am wondering if this unit test can be benefit by [unittest subtest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests)? `setUp` and `tearDown` will be called for each `test_xxx`, it means this test will repeat `setUp/tearDown` four times (for `test_convert_to_fp16`, `test_convert_to_fp16_with_fp16_input`, `test_convert_to_fp16_with_blacklist`, `test_convert_to_bf16`)

Yes, the single test is written separately according to the case, and the initialization `setUp` is inevitable.

Put 4 tests into one test function to avoid multiple initializations

luotao1 (dismissed):

zlsh80826:
How about the following code? `subTest` context can report which configuration is wrong and we can save the number of setUp time. (My experiments show this approach can save ~10s)
```
class TestConvertToMixedPrecision(unittest.TestCase):
    def setUp(self):
        self.temp_dir = tempfile.TemporaryDirectory()
        model = resnet50(True)
        net = to_static(
            model, input_spec=[InputSpec(shape=[None, 3, 224, 224], name='x')]
        )
        paddle.jit.save(
            net, os.path.join(self.temp_dir.name, 'resnet50/inference')
        )

    def tearDown(self):
        self.temp_dir.cleanup()

    def test_convert_to_mixed_precision(self):
        mixed_precision_options = [PrecisionType.Half, PrecisionType.Half, PrecisionType.Half, PrecisionType.Bfloat16]
        keep_io_types_options = [True, False, False, True]
        black_list_options = [set(), set(), set(['conv2d']), set()]

        test_configs = zip(mixed_precision_options, keep_io_types_options, black_list_options)

        for mixed_precision, keep_io_types, black_list in test_configs:
            config = f'{mixed_precision=}-{keep_io_types=}-{black_list=}'
            with self.subTest(mixed_precision=mixed_precision, keep_io_types=keep_io_types, black_list=black_list):
                convert_to_mixed_precision(
                    os.path.join(self.temp_dir.name, 'resnet50/inference.pdmodel'),
                    os.path.join(self.temp_dir.name, 'resnet50/inference.pdiparams'),
                    os.path.join(self.temp_dir.name, f'{config}/inference.pdmodel'),
                    os.path.join(self.temp_dir.name, f'{config}/inference.pdiparams'),
                    backend=PlaceType.GPU,
                    mixed_precision=mixed_precision,
                    keep_io_types=keep_io_types,
                    black_list=black_list,
                )
```

python/paddle/fluid/tests/unittests/ir/test_convert_to_mixed_precision.py
```diff
@@ -30,71 +32,70 @@
...
-            'mixed2/inference.pdmodel',
-            'mixed2/inference.pdiparams',
+            os.path.join(self.temp_dir.name, 'resnet50/inference.pdmodel'),
+            os.path.join(self.temp_dir.name, 'resnet50/inference.pdiparams'),
+            os.path.join(self.temp_dir.name, 'mixed2/inference.pdmodel'),
+            os.path.join(self.temp_dir.name, 'mixed2/inference.pdiparams'),
             PrecisionType.Half,
             PlaceType.GPU,
             False,
             set('conv2d'),
```
zlsh80826: Is this expected? I bet you want `set(['conv2d'])`

```
Python 3.8.10 (default, Jun 22 2022, 20:18:18)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> set('conv2d')
{'n', 'v', 'd', 'o', '2', 'c'}
>>> set(['conv2d'])
{'conv2d'}
```

jiweibo: Thanks very much!

zlsh80826 (dismissed):

luotao1 (approved):

PR #48897, opened 2022-12-08
support sharding in fp16 on xpu, 

sljlp:
### PR types

New features
### PR changes

Others
### Describe

Support sharding + fp16 on XPU 
Change allreduce_max to allreduce_sum for found nan or inf becasue allredeuce_max is not supprted on XPU


LiYuRio (approved):
LGTM

PR #48902, opened 2022-12-08
Save fused_attention op memory when dropout_rate = 0.0

sneaxiy:
### PR types
Performance optimization

### PR changes
OPs

### Describe
When `dropout_rate` is 0.0, some tensors are not necessary when computing backward in the `fused_attention` op. This PR refines the memory in this case.


wanghuancoder (approved):
LGTM

PR #48934, opened 2022-12-08
Move GeneralDivMod from segmented_array.h to fast_divmod.h

JamesLim-sy:
### PR types

Function optimization
### PR changes

OPs
### Describe

- I think it\`s better for `GeneralDivMod`  being adjacent with `FastDivMod` in the same header file.


CtfGo (approved):
LGTM

PR #48960, opened 2022-12-09
Fix bug when invoking amp.decorate with non-None optimizer when using Sharding Stage 1 + AMP O2

sneaxiy:
### PR types
Bug fixes

### PR changes
APIs

### Describe
When invoking `amp.decorate` with non-None `optimizer` when using Sharding Stage 1 + AMP O2, the `optimizer` is an instance of `DygraphShardingOptimizer`. It is not supported before this PR.


zhangbo9674 (approved):
LGTM

PR #49010, opened 2022-12-12
[Bugfix] recompute dep filter param

JZ-LIANG:
### PR types
Bug fixes

### PR changes
Others

### Describe
Sharding would partition param to different rank, so param should not be choose as dep var for recompute when sharding on.
support recompute depend op for pipline reshard


aoyulong (approved):
LGTM

PR #49056, opened 2022-12-13
quantize softmax

sfraczek:
### PR types
Performance optimization

### PR changes
OPs

### Describe
Enable softmax int8


sfraczek:
Taken over by Paulina in https://github.com/PaddlePaddle/Paddle/pull/50467

PR #49068, opened 2022-12-14
Keep double-buffer reader for static mode

sljlp:
### PR types

Performance optimization
### PR changes

Others
### Describe

double_buffer is removed from fluid, which caused the reduction of performance of loading data.
This PR keeps the double buffer function in GeneratorLoader, not using it as an open api.


LiYuRio (approved):
LGTM

PR #49080, opened 2022-12-14
SetDeviceId in StreamSafeCUDAAllocation

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

`StreamSafeCUDAAllocation` utilizes `cudaEvent` to achieve cross-stream synchronization. If the `SetDeviceId` is not called, the default `cudaDeviceId` will be 0, and all event-releated API calls are launched to gpu-0. 
This leads to an error in ResNet50_bs256_fp32 N1C8 static-graph training.  In this case, a cross-stream Allocation records events in a thread (the `DeviceThread` of the standalone executor) and queries events in another (the data loader thread).  Since data loader thread never call `SetDeviceId`, it launches event queries to gpu-0 even though the process is not trained in gpu-0 (and thus the events are not created and recorded for gpu-0). Then a CUDA error occurs: 

ac2ac4b70a76479774a0e983517d0e1b

This PR fixes it by calling `SetDeviceID` before event-releated API call in `StreamSafeCUDAAllocation`.

### DIFF
--- a/paddle/fluid/memory/allocation/stream_safe_cuda_allocator.cc
+++ b/paddle/fluid/memory/allocation/stream_safe_cuda_allocator.cc
@@ -16,6 +16,7 @@
 #include <thread>
 
 #include "paddle/fluid/platform/profiler/event_tracing.h"
+#include "paddle/phi/backends/gpu/gpu_info.h"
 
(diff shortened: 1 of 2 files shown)


zhiqiu (approved):
LGTM

PR #49091, opened 2022-12-15
[Paddle Inference] repalce log error with paddle throw

yuanlehome:
### PR types

Others
### PR changes

Others
### Describe

1. repalce log error with paddle throw
2. PADDLE_WITH_TENSORRT


jiweibo (approved):
LGTM

PR #49136, opened 2022-12-17
Develop sequence lod all

mhy-666:
### PR types

Others
### PR changes

APIs

### Describe

According to clean the code in paddle.fluid. The useless codes should be removed and code still be used should move to a new position.

For function paddle.fluid.layers.sequence_lod, since it is still be used by paddle.static.nn.sequence_lod, it should be moved to the module static.nn.

### DIFF
--- a/paddle/phi/backends/stream.h
+++ b/paddle/phi/backends/stream.h
@@ -44,7 +44,7 @@ class Stream {
   using Callback = std::function<void()>;
 
   Stream() = default;
-  // For compatiable
+  // For compatible
   Stream(const Place& place, stream_t stream);
   ~Stream();
   const stream_t& raw_stream() const;


python/paddle/static/nn/sequence_lod.py
```diff
@@ -0,0 +1,1496 @@
...
+                output.data = [[ 0.3234, -0.2334,  0.7433],
+                               [ 0.5646,  0.9464, -0.1223],
+                               [-0.1343,  0.5653,  0.4555],
+                               [ 0.9954, -0.1234, -0.1234]]
+                output.shape = [4, 3]     # 3 = num_filters
+                output.lod = [[0, 3, 4]]  # Remain the same
+
+
+    Args:
+        input (Variable): LoDTensor with shape :math:`(M, K)`, where M is the total time-step of mini-batch
```
zoooo0820: 1. Please remove `python/fluid/layers/sequence_lod.py`
2. Please modify doc-string of these APIs, making it comply with the document specification of 2.0. You can refer this [Link](https://www.paddlepaddle.org.cn/documentation/docs/zh/dev_guides/api_contributing_guides/api_docs_guidelines_cn.html)

PR #49138, opened 2022-12-17
[Unify KernelKey] change OpKernelType->KernelKey

jiahy0825:
### PR types
Function optimization

### PR changes
OPs

#### Background
fluid and phi have different kernel selection methods, fluid selects kernel according to OpKernelType tuple <place, layout, dtype, library_type, custom_value>; phi selects kernel according to KernelKey tuple <backend, layout, dtype>. With the continuous migration of kernel to phi and the improvement of operator standardization, the selection and distribution methods of the framework need to be unified. This PR changes the return value of the entry function (GetExpectedKernelType and KernelTypeForVar) from OpKernelType to KernelKey, which means fluid and phi both use KernelKey as the first choice during kernel selection.

#### Main problem
- OpKernelType should only be used for kernel selection. However fluid uses the place of OpKernelType to determine the place of the input tensor.
- The place of OpKernelType contains information of backend and place, but the backend of KernelKey doesn't hold place index. We need to select kernels with less information.
- Static mode, new interpreter and old dygraph have three different kernel selection logics, we need to assure the accuracy of these three modes at the same time.

#### The principle of this PR
- The place of the input tensor should be determined by the place of ExecutionContext, as we have observed that the output place of GetExpectedKernelType is acquired from ExecutionContext.
- KernelTypeForVar returns only two results: KernelType of tensor itself (which means need DataTransform) and KernelType of expected_kernel_key (do not need DataTransform, we should return [ALL_BACKEND, expected_kernel_key.layout(), expected_kernel_key.dtype()] )

#### 背景
fluid 和 phi 下的 kernel 选择体系是两套，fluid 根据 OpKernelType 五元组 <place, layout, dtype, library_type, custom_value> 选择 kernel；phi 根据 KernelKey 三元组 <backend, layout, dtype> 选择 kernel。随着 kernel 不断向 phi 迁移以及算子规范化程度的提高，框架的选择分发体系也亟待统一。本 PR 将选择分发入口函数（GetExpectedKernelType 和 KernelTypeForVar）的返回值将 OpKernelType 更改为 KernelKey，相当于后续 fluid 和 phi 都优先使用 KernelKey 来进行选择分发。

#### 存在问题
- OpKernelType 本来只应该用来寻找 kernel，然而 fluid 体系下 kernel 输入 tensor 的 place 也由 OpKernelType 决定。
- OpKernelType 五元组中的 Place 包括了 backend 和 place 两方面的信息，然而 phi 下的 KernelKey 只有 backend 信息，需要在缩减信息量的前提下，完成 Kernel 选择
- 静态图、新执行器、老动态图有三套不同的选择分发逻辑，统一 KernelKey 要同时保证这三套逻辑的正确性。

#### 选择分发体系统一的原则
- 观察到 GetExpectedKernelType 返回的 Place 信息主要是从 ExecutionContext 中拿到的，因此输入 tensor 的 place 应该由 ExecutionContext 的 place 决定。
- KernelTypeForVar 返回的结果只有两种：一种是 tensor 自己的 KernelType （表示需要 DataTransform），另外一种是 expected_kernel_key 的 KernelType（不需要 DataTransform，应该返回 [ALL_BACKEND, expected_kernel_key.layout(), expected_kernel_key.dtype()] ）


YuanRisheng (dismissed):

chenwhql (dismissed):
LGTM

XiaoguangHu01 (dismissed):
LGTM

YuanRisheng (approved):

paddle/fluid/framework/operator.cc
```diff
@@ -1960,7 +1971,9 @@ void OperatorWithKernel::RunImpl(const Scope& scope,
 
 OpKernelType OperatorWithKernel::InnerGetExpectedKernelType(
     const ExecutionContext& ctx) const {
```
zyfncg, line 1973: 这里的返回类型后面也可以换为`KernelKey`吗？

zyfncg (approved):

paddle/fluid/framework/operator.cc
jiahy0825: 此处的返回值没有必要换成 `KernelKey`，对于 `InnerGetExpectedKernelType` 函数：
1. 函数内：需要处理 ctx 的 place 信息，`KernelKey` 不包含这个信息
2. 函数外：fluid 选择 kernel 时会调用此函数，需要 `OpKernelType` 类型来做 kernel 选择

It's not necessary to change the return value of the `InnerGetExpectedKernelType` function to `KernelKey` for two reasons:
1. In the function: it needs to handle the place information of ctx, `KernelKey` doesn't contains this information.
2. Outside the function: fluid will call this function during kernel selection, which needs `OpKernelType` to choose kernel.

XiaoguangHu01 (approved):
LGTM

PR #49156, opened 2022-12-19
[fluid clean]replace fliud.io.load_inference_model from util_factory.py

wangxn12138:
### PR types

Others
### PR changes

APIs
### Describe

There are some deferences between the 2.0 API static.load_inference_model and old API fluid.io.load_inference_model, to clean the fluid APIs in util_factory.py, make it a distributed private function temporary, put it in distributed directory.


yaoxuefeng6 (approved):
LGTM

PR #49166, opened 2022-12-19
[Dy2St]Add extra_params to to_static API

0x45f:
### PR types

Others
### PR changes

Others
### Describe

[Dy2St]Add extra_param to to_static API
形如下面的动转静代码能够正确运行
```python
import paddle
paddle.set_device('cpu')

class Obj():
    def __init__(self):
        pass

    def func(self, x):
        return x + 1

obj = Obj()

class Net():
    def __init__(self):
        super(Net, self).__init__()
        self.layer1 = paddle.nn.Linear(10, 10)
    
    def forward(self, data):
        
        @paddle.jit.to_static(extra_params=self.layer1.parameters())
        def func(ins, x, loss_fn):
            x = ins.layer1(x)
            return loss_fn(x)

        def func1(x):
            # import pdb; pdb.set_trace()
            return func(self, x, obj.func)
        
        return func1(data)
            


net = Net()
x = paddle.randn([5, 10])
out = net.forward(x)
print(out)
```


0x45f:
see: https://github.com/PaddlePaddle/Paddle/pull/49360

PR #49219, opened 2022-12-21
[AutoParallel-Performance] AMP Flag Memcpy support newexe Overlap

JZ-LIANG:
### PR types
Performance optimization

### PR changes
 Others

### Describe
newexe will identify memcpy_d2h and schedule it to memcpy stream for overlap.
but original memcpy could not enjoy this facility.


From00 (approved):
LGTM

PR #49228, opened 2022-12-21
Fix typos of managed_memory_msg when OOM

From00:
### PR types
Bug fixes

### PR changes
Others

### Describe

When OOM, the threw message tell users to try to use CUDA managed memory by command `FLAGS_use_cuda_managed_memory=false`. However, the corrected command to enable managed memory should be `FLAGS_use_cuda_managed_memory=true`. The incorrect prompts mislead users (see issue https://github.com/PaddlePaddle/Paddle/issues/49225). This PR fixes it.


zhiqiu (approved):
LGTM

PR #49305, opened 2022-12-23
[Dy2st]Get out grad names from program

0x45f:
### PR types

Bug fixes
### PR changes

Others
### Describe

[Dy2st]Get out grad names from program


0x45f:
see: https://github.com/PaddlePaddle/Paddle/pull/49461

PR #49360, opened 2022-12-27
[Dy2St]Add PARAMS_CTX to support call layer.forward in funciont under @to_static

0x45f:
### PR types

New features
### PR changes

Others
### Describe

[Dy2St]Add PARAMS_CTX to support call layer.forward in funciont under @to_static
```python
import paddle
paddle.set_device('cpu')

class Obj():
    def __init__(self):
        pass

    def func(self, x):
        return x + 1

obj = Obj()

class Net():
    def __init__(self):
        super(Net, self).__init__()
        self.layer1 = paddle.nn.Linear(10, 10)
    
    def forward(self, data):
        
        @paddle.jit.to_static
        def func(ins, x, loss_fn):
            x = ins.layer1(x)
            return loss_fn(x)

        def func1(x):
            # import pdb; pdb.set_trace()
            return func(self, x, obj.func)
        
        return func1(data)
            


net = Net()
x = paddle.randn([5, 10])
out = net.forward(x)
print(out)
```

修复之前的报错为：
b575f5ad2a9a37400ae5c7bb92eeee65

### DIFF
--- a/python/paddle/fluid/dygraph/base.py
+++ b/python/paddle/fluid/dygraph/base.py
@@ -131,6 +131,9 @@ def param_guard(parameters):
         yield
 
 
+PARAMS_CTX = set()
+
+
 def _convert_into_variable(tensor):
     """
     Convert Varbase into Variable.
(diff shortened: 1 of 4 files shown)


python/paddle/fluid/dygraph/base.py
```diff
@@ -131,6 +131,9 @@ def param_guard(parameters):
         yield
 
 
+PARAMS_CTX = set()
```
Aurelius84, line 134: 这里只是一层结构，能处理多个子图转写的问题么？我看下面会在from_func_spec会clear下

PR #49373, opened 2022-12-27
Fix CUDA11.8 Unittest Accuracy

zhhsplendid:
### PR types

Bug fixes
### PR changes

Others
### Describe

This PR increased the `delta` in unit test for CUDA 11.8. The reason of this fix:
(1) It seems CUDA 11.8 has higher delta in accuracy result. Our other targets for seresnext under parallel executor have already added delta such as CPU, all reduce test cases, so we did same for GPU base case with CUDA 11.8
(2) A new executor is under developing in PaddlePaddle team, so the unit test for old executor can be relaxed.


zhangbo9674 (approved):
LGTM

PR #49525, opened 2023-01-04
sequence_mask fix: when the bug when the input length is an empty tensor

iclementine:
### PR types

Bug fixes

### PR changes

OPs

### Describe

sequence_mask fix: when the input length is an empty tensor, the kernel tries to dereference illegal sentinel iterator


jeff41404 (approved):
LGTM

PR #49521, opened 2023-01-03
Add for-else

Liyulingyue:
### PR types

Performance optimization

### PR changes

Others 

### Describe

an expansion of https://github.com/PaddlePaddle/Paddle/pull/39119/files

PR #49526, opened 2023-01-04
refix 39089

enkilee:
### PR types

Others
### PR changes

Others
### Describe

Others


luotao1:
- #49521 has already refix it.

PR #49544, opened 2023-01-04
Add 0d Tensor Test Cases for cond, case, switch_case

zhhsplendid:
### PR types

Others
### PR changes

Others
### Describe

Add 0d Tensor Test Cases for `cond, case, switch_case`. Since the 3 APIs are control flow APIs, their support for 0d tensor relies on the underneath APIs. This PR just added test cases to prove that the  3 APIs have already handled 0d tensor well.


Aurelius84 (approved):
LGTM

PR #49548, opened 2023-01-04
fix trace heap overflow

jeff41404:
### PR types
Bug fixes

### PR changes
APIs

### Describe
fix bug of heap overflow in`paddle.trace` when  `axis1` and `axis2` parameters exceed the range limit


wanghuancoder (approved):
LGTM

PR #49584, opened 2023-01-05
fix trt engine memory sharing

yuanlehome:
### PR types

Others
### PR changes

Others
### Describe

fix trt engine memory sharing


jiweibo (approved):
LGTM

PR #49611, opened 2023-01-06
[Don't merge]Revert "move fc api to paddle2.0"

Ligoml:
Reverts PaddlePaddle/Paddle#49379


Ligoml:
非bug相关pr，不需要revert，可以关闭

PR #49631, opened 2023-01-08
[Bug fixes] remove duplicated element_size

wj-Mcat:
### PR types

Bug fixes

### PR changes

APIs

### Describe

remove duplicated element_size binding.

there are two [element_size](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/pybind/eager_method.cc#L2118-L4142) binds in `eager_method.cc` file. so I think the `tensor_methon_element_size` should be removed.


wanghuancoder (approved):
LGTM

PR #49695, opened 2023-01-10
fix save_combine_op

YanhuiDua:
### PR types
Bug fixes

### PR changes
OPs

### Describe
`std::vector<const phi::DenseTensor*> x(inp_vars.size())` constructs a vector of size inp_vars.size(). But `x.push_back(&(inp_var->Get<phi::DenseTensor>()))` add DenseTensor* at the end of the vector. If inp_vars.size()=_n_, so the vector will contain _n_ zero elements and _n_ phi::DenseTensor*, which will leads to segment fault when extracting tensor in line 94. The code is `auto& tensor = *(x[i]);`


shentanyue (approved):
LGTM

qili93 (approved):
LGTM

PR #49767, opened 2023-01-12
fix a bug of stage2 offload.

wuhuachaocoding:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix a bug of stage2 offload：

-  when use offload with stage2, error will occur if use HybridParallelClipGrad, because there is no 'is_distributed' attribute in offload tensors.

```
if not_shared_enable:
      if p.is_distributed:
          if p.dtype == paddle.float16:
              sum_square_dist_fp16.append(sum_square)
          elif p.dtype == paddle.float32:
              sum_square_dist_fp32.append(sum_square)
      else:
          if p.dtype == paddle.float16:
              sum_square_not_dist_fp16.append(sum_square)
          elif p.dtype == paddle.float32:
              sum_square_not_dist_fp32.append(sum_square)
```

-  when using offload with sharding stage2, please use pure sharding stage2, exclude data parallel.


haohongxiang (approved):
LGTM

PR #49812, opened 2023-01-13
[Dy2St]Support call backward() without params in dy2st

0x45f:
### PR types

Bug fixes 
### PR changes

Others
### Describe

[Dy2St]Support call backward() without params in dy2st


Aurelius84 (approved):
LGTM

PR #49838, opened 2023-01-15
fix ninja compilation for cinn

zhiqiu:
### PR types

Others
### PR changes

Others
### Describe

fix ninja compilation for cinn

- `CMAKE_GENERAROR`  is a built-in enviroment of CMAKE, and it affects all projects, including the third_party of paddle and the third_party of third_pary. Some third_parties are not supported ninja yet, so we can not use `CMAKE_GENERAROR` by to enable all.
- Update `cinn.cmake` to support ninja.


risemeup1 (approved):
LGTM for modfiy setup.py

PR #49854, opened 2023-01-16
add add_n for the 0d tensor

wawltor:
### PR types
Function optimization

### PR changes
OPs 

### Describe
add add_n for the 0d tensor


zhwesky2010 (approved):
LGTM

PR #49899, opened 2023-01-17
Adjust mkldnn_placement_pass to check library type and data type

wozna:
### PR types

Function optimization
### PR changes

Others 
### Describe

Previously in mkldnn_placement_pass `use_mkldnn` was only set based on whether there is a `use_mkldnn` attribute. I changed it to check if there is an OneDNN kernel registered with the appropriate datatype and only, in this case, use_mkldnn is set to True.

### DIFF
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -671,8 +671,22 @@ void Executor::EnableMKLDNN(const ProgramDesc& program) {
   for (size_t bid = 0; bid < program.Size(); ++bid) {
     auto* block = const_cast<ProgramDesc&>(program).MutableBlock(bid);
     for (auto* op : block->AllOps()) {
-      if (op->HasAttr("use_mkldnn")) {
-        op->SetAttr("use_mkldnn", true);
+      auto op_type = op->Type();
+
+      auto& all_kernels = OperatorWithKernel::AllOpKernels();
+      auto it = all_kernels.find(op_type);
+      if (it != all_kernels.end())
+        for (auto& kernel_pair : it->second)
+          if (platform::is_cpu_place(kernel_pair.first.place_) &&
+              (kernel_pair.first.library_type_ == LibraryType::kMKLDNN))
+            op->SetAttr("use_mkldnn", true);
+
+      auto phi_kernels = phi::KernelFactory::Instance().SelectKernelMap(
+          phi::TransToPhiKernelName(op_type));
+
+      for (auto& kernel_pair : phi_kernels) {
+        if (kernel_pair.first.backend() == phi::Backend::ONEDNN)
+          op->SetAttr("use_mkldnn", true);
       }
     }
   }
--- a/paddle/fluid/framework/executor.cc
+++ b/paddle/fluid/framework/executor.cc
@@ -671,23 +671,8 @@ void Executor::EnableMKLDNN(const ProgramDesc& program) {
   for (size_t bid = 0; bid < program.Size(); ++bid) {
     auto* block = const_cast<ProgramDesc&>(program).MutableBlock(bid);
     for (auto* op : block->AllOps()) {
-      auto op_type = op->Type();
-
-      auto& all_kernels = OperatorWithKernel::AllOpKernels();
-      auto it = all_kernels.find(op_type);
-      if (it != all_kernels.end())
-        for (auto& kernel_pair : it->second)
-          if (platform::is_cpu_place(kernel_pair.first.place_) &&
-              (kernel_pair.first.library_type_ == LibraryType::kMKLDNN))
-            op->SetAttr("use_mkldnn", true);
-
-      auto phi_kernels = phi::KernelFactory::Instance().SelectKernelMap(
-          phi::TransToPhiKernelName(op_type));
-
-      for (auto& kernel_pair : phi_kernels) {
-        if (kernel_pair.first.backend() == phi::Backend::ONEDNN)
-          op->SetAttr("use_mkldnn", true);
-      }
+      if (FoundOneDNNKernel(op) || FoundPhiOneDNNKernel(op))
+        op->SetAttr("use_mkldnn", true);
     }
   }
 #else
--- a/paddle/fluid/framework/ir/mkldnn/mkldnn_fc_rnn_fuse_pass_tester.cc
+++ b/paddle/fluid/framework/ir/mkldnn/mkldnn_fc_rnn_fuse_pass_tester.cc
@@ -23,6 +23,20 @@ namespace paddle {
 namespace framework {
 namespace ir {
 
+void RegisterOpKernels() {
+  auto& all_kernels = OperatorWithKernel::AllOpKernels();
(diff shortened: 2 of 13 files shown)


paddle/fluid/framework/executor.cc
```diff
@@ -671,8 +671,22 @@ void Executor::EnableMKLDNN(const ProgramDesc& program) {
...
-        op->SetAttr("use_mkldnn", true);
+      auto op_type = op->Type();
+
+      auto& all_kernels = OperatorWithKernel::AllOpKernels();
+      auto it = all_kernels.find(op_type);
+      if (it != all_kernels.end())
+        for (auto& kernel_pair : it->second)
+          if (platform::is_cpu_place(kernel_pair.first.place_) &&
+              (kernel_pair.first.library_type_ == LibraryType::kMKLDNN))
+            op->SetAttr("use_mkldnn", true);
```
sfraczek: Could you add a `break` after this line to exit the for loop after it is already set to true?
Or even it would be great for readability to put this block in a function that returns a boolean. Something like 
```c++
if (FoundOneDNNKernel(op_type)) op->SetAttr("use_mkldnn", true);
```

paddle/fluid/framework/executor.cc
```diff
@@ -671,8 +671,22 @@ void Executor::EnableMKLDNN(const ProgramDesc& program) {
...
+
+      auto phi_kernels = phi::KernelFactory::Instance().SelectKernelMap(
+          phi::TransToPhiKernelName(op_type));
+
+      for (auto& kernel_pair : phi_kernels) {
+        if (kernel_pair.first.backend() == phi::Backend::ONEDNN)
+          op->SetAttr("use_mkldnn", true);
       }
```
sfraczek: Same here, this would be great as 
```c++
if (FoundPhiOneDNNKernel(op_type)) op->SetAttr("use_mkldnn", true);
```
And then both could land in a single if (A or B)

paddle/fluid/framework/ir/mkldnn/mkldnn_fc_rnn_fuse_pass_tester.cc
```diff
@@ -23,6 +23,20 @@ namespace paddle {
 namespace framework {
 namespace ir {
 
+void RegisterOpKernels() {
+  auto& all_kernels = OperatorWithKernel::AllOpKernels();
+  platform::CPUPlace place = platform::CPUPlace();
+  OpKernelType mkldnn_kernel_type = OpKernelType(proto::VarType::FP32,
```
sfraczek: Consider replacing mkldnn with onednn

paddle/fluid/framework/ir/mkldnn/mkldnn_placement_pass.cc
```diff
@@ -13,6 +13,47 @@ See the License for the specific language governing permissions and
...
+  auto& all_kernels = OperatorWithKernel::AllOpKernels();
+  auto it = all_kernels.find(op_type);
+  if (it != all_kernels.end()) {
+    for (auto& kernel_pair : it->second) {
+      if (platform::is_cpu_place(kernel_pair.first.place_) &&
+          (kernel_pair.first.library_type_ == LibraryType::kMKLDNN)) {
+        if (op->inputs.size() > 0 && op->inputs[0]->IsVar() &&
+            kernel_pair.first.data_type_ == op->inputs[0]->Var()->GetDataType())
+          return true;
+        return true;
```
sfraczek: whether if is true or false it returns true

paddle/fluid/framework/ir/mkldnn/mkldnn_placement_pass.cc
```diff
@@ -13,6 +13,47 @@ See the License for the specific language governing permissions and
...
+  auto phi_kernels = phi::KernelFactory::Instance().SelectKernelMap(
+      phi::TransToPhiKernelName(op_type));
+
+  for (auto& kernel_pair : phi_kernels) {
+    if (kernel_pair.first.backend() == phi::Backend::ONEDNN) {
+      if (op->inputs.size() > 0 && op->inputs[0]->IsVar() &&
+          kernel_pair.first.dtype() == framework::TransToPhiDataType(
+                                           op->inputs[0]->Var()->GetDataType()))
+        return true;
+      return true;
```
sfraczek: same here

paddle/fluid/framework/ir/mkldnn/mkldnn_placement_pass.cc
```diff
@@ -13,6 +13,56 @@ See the License for the specific language governing permissions and
...
+          (kernel_pair.first.library_type_ == LibraryType::kMKLDNN)) {
+        if (op->inputs.size() > 0) {
+          if (op->inputs[0]->IsVar() &&
+              op->inputs[0]->Var()->Name() != "feed" &&
+              kernel_pair.first.data_type_ ==
+                  op->inputs[0]->Var()->GetDataType())
+            return true;
+        } else {
+          return true;
+        }
```
sfraczek, line 40: Since this is repeated twice in the same function, could you put it into a separate function  with a descriptive name for easier readability?

paddle/fluid/framework/ir/mkldnn/mkldnn_fc_rnn_fuse_pass_tester.cc
```diff
@@ -23,6 +23,20 @@ namespace paddle {
 namespace framework {
 namespace ir {
 
+void RegisterOpKernels() {
```
sfraczek: Maybe this could be put in pass_tester_helper.h and used in both test files. It could accept something like const std::vector<string>& with names of ops that need to be registered.

sfraczek: Why does this function not use static variable and this one does?
https://github.com/PaddlePaddle/Paddle/pull/49899/files#diff-e89175d68effdff4937805b6a1756de1ccb7592a23c69d25d00d4780e6e545ecR28

paddle/fluid/framework/executor.cc
wozna: This is a nice suggestion, the code will be more readable

paddle/fluid/framework/ir/mkldnn/mkldnn_fc_rnn_fuse_pass_tester.cc
wozna: Good idea, I threw it into pass_tester_helper.h

paddle/fluid/framework/executor.cc
wozna: I corrected it a bit, but due to the fact that in the executor I can't get to the data type because we only have OpDesc. In mkldnn_placement_pass we have access to Node, which allows us to check the data type.

wozna:
@sfraczek @jczaja Could you continue your review?

sfraczek (approved):
LGTM!

jczaja (approved):
LGTM

Issue #49927, opened 2023-01-18
good first issue lists


Ligoml:
大家好，我们通过代码扫描工具，在 Paddle 框架中检测出若干适合入门框架开发的 issue，这些 issue 可以通过简单的 case 复现。欢迎大家一起提 PR 解决，你可以通过下表认领感兴趣的任务。为表感谢，我们为你准备了新春礼品🎁

对每个 case，我们都提供了详细的错误调用堆栈，能够看到代码在 Paddle 的哪个源码文件中出现这种错误（所以很容易锁定错误代码位置）。这些问题大概率是对输入参数检查不全面导致的，可以通过添加 PADDLE_ENFORCE_相关的参数检查和报错来解决。当然，也有可能是 Paddle 框架/算子内有更深层次的问题。我们希望你做的是：
1. 安装 [Paddle官网develop(Nightly Build）包](https://www.paddlepaddle.org.cn/install/quick?docurl=/documentation/docs/zh/develop/install/pip/linux-pip.html) 复现 case。
2. 参考[《Paddle贡献指南》](https://www.paddlepaddle.org.cn/documentation/docs/zh/develop/dev_guides/index_cn.html) 和 [《Paddle报错信息文案书写规范》](https://github.com/PaddlePaddle/Paddle/wiki/Paddle-Error-Message-Writing-Specification)，修复 case。
3. 提交 PR，1个 PR 修复1个 case。
  a. PR 描述中请介绍你发现的原因及你的解决办法。
  b. 请同时将复现 case 添加为1个单测。不要新建单测文件，将单测放到框架(模块)/算子已有的单测文件中。

注：
1. 你可以参考这个修复 PR 实例：https://github.com/PaddlePaddle/Paddle/pull/49656
2. 这些 case 修复难度不大，还能借此了解Paddle的机制，欢迎参与~
3. [历史上的 good first issue 列表](https://github.com/PaddlePaddle/community/tree/master/pfcc#good-first-issue)，也欢迎来提 PR 解决~
4. 春节期间（2023.1.21-2023.1.27）可能会存在回复不及时的情况，会在年后逐渐恢复，礼品也会在年后统一寄出，敬请关注！
5. 欢迎联系花花加入社区，和我们一起快乐开源！
<img width="260" alt="image" src="https://user-images.githubusercontent.com/39876205/213131273-361d2bb8-ea48-4f62-9603-0b0401f50337.png">

[...]

PR #50007, opened 2023-01-22
Fix Python IndexError of case11: paddle.static.nn.data_norm

longranger2:
### PR types
Bug fixes 

### PR changes
APIs 

### Describe
- https://github.com/PaddlePaddle/Paddle/issues/49923
- https://github.com/PaddlePaddle/Paddle/issues/49927

### Solution
判断input的维度是否为4


wanghuancoder (commented):
为这个case加个单测吧

PR #50016, opened 2023-01-23
Fix Python IndexError of Case14: paddle.nn.functional.glu

longranger2:
### PR types
Bug fixes 

### PR changes
APIs 

### Describe
- https://github.com/PaddlePaddle/Paddle/issues/49923
- https://github.com/PaddlePaddle/Paddle/issues/49927

### Solution
原因：glu函数的axis取值要求小于x.shape
解决方法：添加axis取值判断条件


python/paddle/tensor/manipulation.py
```diff
@@ -1918,6 +1918,7 @@ def split(x, num_or_sections, axis=0, name=None):
             dim = dim.numpy()
             dim = dim.item(0)
         assert len(input.shape) + dim >= 0, "(rank(x) + axis) must >= 0"
+        assert dim < len(input.shape), "axis must < (rank(x)"
```
wanghuancoder: (rank(x)是不是应该改成rank(x)

python/paddle/tensor/manipulation.py
```diff
@@ -1997,6 +1998,7 @@ def _get_SectionsTensorList(one_list):
             inputs['AxisTensor'] = dim
         else:
             assert len(input.shape) + dim >= 0, "(rank(x) + axis) must >= 0"
+            assert dim < len(input.shape), "axis must < (rank(x)"
```
wanghuancoder: 同上

wanghuancoder (commented):
为这个case加个单测吧

python/paddle/tensor/manipulation.py
longranger2: 对的，我的疏忽

longranger2:
单测我今天搞定

longranger2:
> 为这个case加个单测吧

单测添加好了

wanghuancoder (dismissed):
LGTM

python/paddle/nn/functional/activation.py
```diff
@@ -1622,6 +1622,7 @@ def glu(x, axis=-1, name=None):
     check_variable_and_dtype(
         x, 'input', ['float16', 'float32', 'float64'], "glu"
     )
+    assert axis < len(x.shape), "axis must < rank(x)"
```
iclementine: I think it's better to check `-R <= axis < R` because this is more precise.
Also, to report the expected behavior and what actually happened is more informative.

For example
```python
rank = x.shape
if not (-rank <= axis < axis):
    raise ValueError("Expected value range of `axis` is [{}, {}), but received axis: {}".format(-rank, rank, axis))
```

iclementine (changes requested):

python/paddle/nn/functional/activation.py
longranger2: Thank you for your advice. I will fix it later

iclementine (approved):
LGTM

PR #50040, opened 2023-01-28
Add the missing ps.proto and remove ps_pb2.py

sneaxiy:
### PR types
Others

### PR changes
Others

### Describe
This PR removes `python/paddle/fluid/incubate/fleet/parameter_server/pslib/ps_pb2.py` and add the corresponding `paddle/fluid/distributed/ps.proto` file. 

Note that since I do not know where the exact `ps.proto` is (the exact commit id), I re-generated the `ps.proto` from the old `ps_pb2.py` file. The re-generation process is referred from https://stackoverflow.com/questions/70479912/reverse-engineering-proto-files-from-pb2-py-generated-with-protoc .


zmxdream (approved):
LGTM

PR #50050, opened 2023-01-29
make FLAGS_gemm_use_half_precision_compute_type=false by default

sneaxiy:
### PR types
Others

### PR changes
Others

### Describe
Make `FLAGS_gemm_use_half_precision_compute_type=false` by default. It would improve the performance of the ViT models and make FP16 models convergence better.


zhiqiu (approved):
LGTM

PR #50085, opened 2023-01-30
Fix send start msg

sljlp:
### PR types

Bug fixes
### PR changes

Others
### Describe

set src id to SOURCEID_ID when sending start msg


LiYuRio (approved):
LGTM

PR #50086, opened 2023-01-30
[BugFix][ConditionalBlock] fix judgement about scope validation

kangguangli:
### PR types

Bug fixes
### PR changes

OPs
### Describe

This PR is to fix bug about scope validation in `ConditionalBlock`.

In `ConditionalBlockOp`, we got scope that sub block needs from the parameter `scope` of method `run`, and we cache it in `InterpreterCore`. This scope may be invalid cause the outer executor may destruct it, so we need to check whether it's valid. 
However, we don't have a solid method to check it currently. So simply let the logic about scope validation revert to that before #45696.


zhiqiu (approved):
LGTM

PR #50101, opened 2023-01-31
support fp16 index_select

wangxn12138:
### PR types

Others
### PR changes

OPs
### Describe

add fp16 dtype check of index_select


zmxdream (approved):
LGTM

PR #50133, opened 2023-02-01
[Deprecate API] add deprecated decorator for paddle.utils.profiler

zoooo0820:
### PR types

Others
### PR changes

APIs
#### Motivation：
- The APIs `paddle.utils.profiler` is `paddle.fluid.profiler` essentially, if we delete the code of `paddle.fluid.profiler`, `paddle.utils.profiler` will not work.  So the `paddle.utils.profiler` need to be deprecated, then we can remove the code of `paddle.fluid.profiler`.
- From version 2.3, there is a new profiler `paddle.profiler`, which is more comprehensive.  As outdated APIs, `paddle.utils.profiler` can be removed. 
- There are **8** public APIs need to be deprecated，**5** of them have already been marked at `2.3.0`, we need to mark the remaining **3** APIs. Then these APIs can be removed together in the next release version of PaddlePaddle.

#### What Is This PR Doing:
This PR adds deprecated decorator for following 3 public APIs:
```
paddle.utils.profiler.ProfilerOptions
paddle.utils.profiler.Profiler
paddle.utils.profiler.get_profiler
```
#### What to Do Next
- From version `2.4.2`, these APIs will be marked as `deprecated`. Users can still use them, however, we strongly recommend using the new API `paddle.profiler`
- From version `2.5.0`, these APIs will be removed.

### DIFF
--- a/python/paddle/utils/profiler.py
+++ b/python/paddle/utils/profiler.py
@@ -19,6 +19,7 @@ from ..fluid import core
 from ..fluid.profiler import cuda_profiler  # noqa: F401
 from ..fluid.profiler import profiler  # noqa: F401
 from ..fluid.profiler import reset_profiler, start_profiler, stop_profiler
+from .deprecated import deprecated
 
 __all__ = [  # noqa
     'Profiler',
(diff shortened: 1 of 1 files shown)


rainyfly:
LGTM.

jeff41404 (approved):

jzhang533 (commented):
我觉得做直接做删除就可以，有以下原因：

- 这三个API，如果不配合已经在2.3版本已经deprecate的API，应该也无法单独使用吧？
- 2.4.2版本是patch版本，应该只做bugfix。
- 已经有了可以替代的API存在了。

PR #50340, opened 2023-02-08
fix set_value when source tensor is fp16 dtype and index is scalar tensor

Courtesy-Xs:
### PR types

Bug fixes
### PR changes

Others
### Describe

when code set like the following runs in paddle, it will failed because paddle doesn't support assign when source tensor is fp16 type and index is scalar tensor,.
it's restricted by set_value mechanism and its component will be reconstructed soon, so here we just make minor changes for this program runs correctly 

```
import paddle
cls_targets = paddle.full([3,3],0,dtype="float16")
ix1 = paddle.full([1],1,dtype="int64")
ix2 = paddle.full([1],2,dtype="int64")
cls_targets[ix1,ix2]= 1
print(cls_targets)
```


zhangting2020 (approved):
LGTM

PR #50348, opened 2023-02-08
Remove redundant cpu computation in slice compute

zhangboSJTU:
### PR types

Performance optimization
### PR changes

OPs 
### Describe

Reduce redundant cpu computation in SliceCompute
Since InferShape have made checks thus here only UpdateSliceAttrs


JamesLim-sy (approved):
LGTM

PR #50354, opened 2023-02-08
Add matmul_v2 and fused_matmul to the quantization process and adjust Ernie model test

wozna:
### PR types

New features
### PR changes

OPs 
### Describe

This PR: 
- adds matmul_v2 and fused_matmul to quantization process 
- changes matmul and matmul_v2 to fused_matmul in cpu_quantize_placement_pass
- remove duplicated list of of in `paddle/fluid/inference/api/analysis_config.cc`. It is only needed to check if op_types from users are supported, but this check is already implemented in `cpu_quantize_placement_pass.cc`. 
- adds a check not to store information in the `fill_constant` op, as this type of operator may be fused by some passes.

### DIFF
--- a/paddle/fluid/framework/ir/graph_pattern_detector.cc
+++ b/paddle/fluid/framework/ir/graph_pattern_detector.cc
@@ -2047,10 +2047,11 @@ PDNode *patterns::Reshape2Matmul::operator()() {
 }
 
 PDNode *patterns::MatmulWithInputOps::operator()(bool with_residual) {
-  auto prev_op_x = pattern->NewNode(prev_op_x_repr())->assert_is_op();
-  auto prev_op_y = pattern->NewNode(prev_op_y_repr())->assert_is_op();
+  std::unordered_set<std::string> matmul_ops{
+      "matmul", "matmul_v2", "fused_matmul"};
 
-  auto matmul_op = pattern->NewNode(matmul_op_repr())->assert_is_op("matmul");
+  auto matmul_op =
+      pattern->NewNode(matmul_op_repr())->assert_is_ops(matmul_ops);
 
   if (!with_residual) {
     matmul_op->assert_more([&](Node *x) {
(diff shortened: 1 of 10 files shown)


paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
```diff
@@ -934,6 +926,11 @@ void CPUQuantizePass::QuantizeMatmul(Graph* graph, bool with_residual) const {
                           is_x_unsigned,
                           is_y_unsigned));
 
+    if (matmul_op->Op()->Type() == "matmul" ||
+        matmul_op->Op()->Type() == "matmul_v2") {
+      matmul_op->Op()->SetType("fused_matmul");
+    }
```
Silv3S: I think we could move this code to _cpu_quantize_placement_pass_, because `QuantizeMatmul` has extra checks which allow quantization only if operator has `mkldnn_data_type`=="int8". We are going to remove all oneDNN-specific attributes from base matmul operator, so they won't be accessible.

paddle/fluid/framework/ir/graph_pattern_detector.cc
```diff
@@ -2047,10 +2047,11 @@ PDNode *patterns::Reshape2Matmul::operator()() {
 }
 
 PDNode *patterns::MatmulWithInputOps::operator()(bool with_residual) {
```
Silv3S: As inputs are removed from this pattern, maybe base `patterns::Matmul` could be adjusted and used?

paddle/fluid/framework/ir/mkldnn/cpu_quantize_pass.cc
wozna: Okay, I'll move it to cpu_quantize_placement_pass. And since cpu_quantize_placement is always before cpu_quantize_pass, it can be inferred that only `fused_matmul` should be searched in cpu_quantize_pass.

paddle/fluid/framework/ir/graph_pattern_detector.cc
wozna: All in all, if I move the change to fused_matmul to the placement pass, onyl fused_matmul will be needed here, and it will be the same with other operators.

paddle/fluid/framework/ir/mkldnn/cpu_quantize_placement_pass.cc
```diff
@@ -69,6 +96,8 @@ void CPUQuantizePlacementPass::ApplyImpl(ir::Graph* graph) const {
     if (op->Op()->GetAttrIfExists<int>("skip_quant") == 1) {
       return;
     }
+
+    ReplaceWithFusedOp(op);
```
Silv3S, line 100: I like this idea, as it can be reused by all oneDNN fuse passes, not only quantization. Could you please move this method to `/paddle/fluid/framework/ir/mkldnn/mkldnn_pass_util.h`?

Silv3S (approved):
LGTM

Silv3S (approved):

jczaja (approved):
LGTM

PR #50351, opened 2023-02-08
[with_data_parallel][part1] remove with_data_parallel in unit test

kangguangli:
### PR types

Others
### PR changes

APIs
### Describe

remove with_data_parallel in unit test

#### 1. Why delete these unit case? ##
The `ParallelExecutor` will be removed in the later of this year and are not maintained right now. In this stage, we just delete the unit cases that are related to multi-gpus.
#### 2. What else does this PR do?
The unit tests that contains `with_data_parallel` but are not about `ParallelExecutor` should leave it as it is except removing `with_data_parallel`.

PR #50501, opened 2023-02-14
[with_data_parallel][part2] remove with_data_parallel in unit test

kangguangli:
### PR types

Others
### PR changes

APIs
### Describe

remove with_data_parallel in unit test

#### 1. Why delete these unit case? ##
The `ParallelExecutor` will be removed in the later of this year and are not maintained right now. In this stage, we just delete the unit cases that are related to multi-gpus.
#### 2. What else does this PR do?
1. The unit tests that contains `with_data_parallel` but are not about `ParallelExecutor` should leave it as it is except removing `with_data_parallel`.
2. add `assert` in `with_data_parallel` to prevent new multi-gpu case before deleting api. 

#### Related PRs
1. #50351


zhiqiu (approved):
LGTM

PR #50517, opened 2023-02-15
fix cross step sync problem on npu

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Describe

fix cross step sync problem on npu, see details in comments


From00 (approved):
LGTM

PR #50578, opened 2023-02-16
buf fix: add dtype arg in func:full of sqrt_grad

zhengqiwen1997:
### PR types

Bug fixes
### PR changes

Others
### Describe

Add dtype arg in func:full of sqrt_grad.


Charles-hit (approved):
LGTM

PR #50635, opened 2023-02-18
[Zero-dim] zero dim support for paddle.static.nn.prelu

juncaipeng:
### PR types

Others
### PR changes

Others
### Describe

[Zero-dim] zero dim support for paddle.static.nn.prelu


zhwesky2010 (approved):
LGTM

PR #50676, opened 2023-02-20
[Hybrid parallelism] Tensor Parallel Extra Sync

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Describe
Add extra synchronization for Tensor Parallelism.

In TP training, Parameters like pos embedding, LayerNorm, bias of row parallel are not distributed and are supposed to be the SAME among TP group.

BUT after a long-term training, there would be difference of those parameter among TP group, and might hurt model performance (especially for big NLP generation model where pos embed plays an important role in generation).

the diff of params comes from two reasons:
1. Tensor Parallelism only synchronize activation (activation_grad) among group, leave param_grad unsynchronized.
2. some Layers are NOT numerical stable(LayerNorm_grad, Embedding_grad): given the synchronized activation_grad, each rank conducts layer computation independently and might get different param_grads. 

This PR add Extra synchronization to compromise this problem in long-term Tensor Parallelism training.
A Final solution should be fix the numerical stable problem of those layers.

image

### DIFF
--- a/python/paddle/distributed/fleet/utils/__init__.py
+++ b/python/paddle/distributed/fleet/utils/__init__.py
@@ -21,6 +21,8 @@ from paddle.distributed import fleet
 import paddle
 from . import log_util  # noqa: F401
 from . import hybrid_parallel_util  # noqa: F401
+from . import tensor_parallel_utils  # noqa: F401
+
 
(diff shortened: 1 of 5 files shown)


aoyulong (approved):
LGTM

PR #50692, opened 2023-02-21
Flatten input_spec when check neg shape

0x45f:
### PR types

Bug fixes
### PR changes

Others
### Describe

Flatten input_spec when check neg shape


2742195759 (approved):
LGTM

PR #50753, opened 2023-02-21
update device value error formatting

Infrared1029:
### PR types
Others

### PR changes
Others 

### Describe
instead of printing the match object, it prints the actual matched string (e.g. gpu:0) by applying the `group()` method to the match object.

`ValueError: The device should not be gpu:0, since PaddlePaddle is not compiled with CUDA`

instead of

`ValueError: The device should not be <re.Match object; span=(0, 5), match='gpu:0'>, since PaddlePaddle is not compiled with CUDA`


zhiqiu (approved):
LGTM

PR #50780, opened 2023-02-22
[Prim]fix attrs loss in creating op

cyber-pioneer:
### PR types
Bug fixes

### PR changes
Others

### Describe
When setting composite forward ops, during creating operator, there may be some attrs loss which should be added.

For example, after constructing program of op dropout, attrs seed and fix_seed will not appear in program. In this case, when we need to copy the same op from origin program, attrs seed and fix_seed cannot be got by op.attr_names. So we use special member property defined during origin op initialization.


python/paddle/fluid/framework.py
```diff
@@ -3080,13 +3082,22 @@ def find_name(var_list, name):
...
+                self.legacy_attrs[item] = self.desc.attr(item)
+
             if self._has_kernel(type):
                 self.desc.infer_var_type(self.block.desc)
                 self.desc.infer_shape(self.block.desc)
 
     def _has_kernel(self, op_type):
         return op_type not in self.OP_WITHOUT_KERNEL_SET
 
+    def _get_runtime_attrs(self):
```
JiabinYang, line 3097: Add comments to avoid other one using this.

python/paddle/incubate/autograd/primx.py
```diff
@@ -656,8 +656,10 @@ def expand_nested_list(xs):
                     outputs[op.output_names[i]] = op.output(op.output_names[i])
 
                 attrs = {}
-                for name in sorted(op.attr_names):
-                    attrs[name] = op.attr(name)
+                runtime_attrs = op._get_runtime_attrs()
```
JiabinYang, line 661: Add comments to explain why we need this.

JiabinYang (commented):
some comments

python/paddle/fluid/framework.py
cyber-pioneer: done

python/paddle/incubate/autograd/primx.py
cyber-pioneer: done

XiaoguangHu01 (approved):
LGTM

Issue #47618, opened 2022-11-02
Error FileNotFoundError: [WinError 2] The system cannot find the file specified


MikeLud:
### bug描述 Describe the Bug

I am working on making License Plate Module using paddleOCR for the below project and I am getting the below errors when importing paddle. It imports fine using an interactive window.

System Environment：Python 3.7
Paddle Version：paddlepaddle==2.3.2
PaddleOCR Version: paddleocr==2.6.1.0

Project Link:
https://github.com/codeproject/CodeProject.AI-Server

License Plate Module code Link:
https://github.com/MikeLud/CodeProject.AI-Server/tree/main/src/modules/ALPR

![image](https://user-images.githubusercontent.com/36526272/199621369-5d8a483b-1824-447f-9c14-376d3cf2d847.png)
![image](https://user-images.githubusercontent.com/36526272/199621642-5d0179fd-ac64-4729-80c6-968ae7b166c9.png)

MikeLud:
@luotao1 thanks for looking into this issue, we were able to get it to work, see the below link that has all of the details
https://www.codeproject.com/Tips/5347636/Getting-PaddleOCR-and-PaddlePaddle-to-work-in-Wind

luotao1:
> 
https://www.codeproject.com/Tips/5347636/Getting-PaddleOCR-and-PaddlePaddle-to-work-in-Wind

cc to @dyning

luotao1:
@MikeLud I see you mentioned：
<img width="947" alt="image" src="https://user-images.githubusercontent.com/6836917/212796958-85d25295-22cb-4f0a-9a4a-bcf17b1e64c9.png">
<img width="924" alt="image" src="https://user-images.githubusercontent.com/6836917/212797018-f9d99b91-30e0-46b9-8e41-ac4b6b74d0e5.png">
How about creating a PR to fix it?

luotao1:
> 
https://www.codeproject.com/Tips/5347636/Getting-PaddleOCR-and-PaddlePaddle-to-work-in-Wind

cc to @dyning @Evezerest @D-DanielYang

zhwesky2010:
@MikeLud hello，I tried on Windows, when `shell=False` , it always works.  How to trigger this problem?

<img width="967" alt="infoflow 2023-01-17 11-15-14" src="https://user-images.githubusercontent.com/52485244/212802539-8130b866-f0a3-40ce-81b7-95c245d9e39f.png">

[...]

PR #50851, opened 2023-02-23
Fix for breaking standalone PyInstaller builds

ShiYue-oo:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Currently when importing paddle in a standalone EXE built using pyinstaller, the build will crash as the interpreter path generated in this fix will not resolve to an existing file. It should probably have been wrapped in a try/except in the first place.

related:
-  https://github.com/PaddlePaddle/Paddle/pull/44525
- https://github.com/PaddlePaddle/Paddle/issues/47618#issuecomment-1384761633


luotao1:
CodeStyle流水线，请修复下代码格式
image
Coverage流水线，需要添加python单测
image

luotao1:
close due to the following PR is merged
- https://github.com/PaddlePaddle/Paddle/pull/51235

PR #50854, opened 2023-02-24
fix composite grad maker code gen

Charles-hit:
### PR types

Bug fixes
### PR changes

Others
### Describe

Fix composite grad maker code gen, when attribute is Scalar.


zyfncg (approved):

JiabinYang (approved):
LGTM

PR #50908, opened 2023-02-25
[Bfloat16]register bfloat16 datatype for squared l2 norm

shaojiewang:
### PR types

Others
### PR changes

OPs
### Describe

register bfloat16 datatype for squared l2 norm kernel and fused softmax with upper triangular mask.


lanxianghit (approved):

Xreki (approved):
LGTM

PR #51029, opened 2023-02-28
[XPU] delete op_device attr

zhupengyang:
### PR types

Bug fixes

### PR changes

Others

### Describe

"op_device" attr is only used in model training. "op_device" attr will change place of op kernel, so we use "delete_op_device_pass" to remove it.


hong19860320 (approved):
LGTM

PR #51035, opened 2023-02-28
[Auto Parallel] Speedup the completion process

aoyulong:
### PR types

Performance optimization

### PR changes

Others

### Describe

This pr can  speedup the completion process and gives about 2.45 times faster than the original one.


JZ-LIANG (approved):
LGTM

PR #51052, opened 2023-03-01
[xpu] fix bugs of split/embedding_with_wltwise_add/beam_search_decode kernel

mayang002:
### PR types
Bug fixes

### PR changes
Others

### Describe
1. early return for xpu split kernel when x.numel is zero
2. fix missed lod info for beam_search_decode xpu kernel
3. Embedding_with_wltwise_add xpu kernel should support int32 / int64 ids


zhupengyang (approved):
LGTM

PR #51071, opened 2023-03-01
adjust defalut tolerance of output and grad

Vvsmile:
### PR types

Others
### PR changes

Others
### Describe

adjust defalut tolerance of output and grad


python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -2392,11 +2392,14 @@ def check_grad_with_place(
             if grad.dtype == np.uint16:
                 grad = convert_uint16_to_float(grad)
                 max_relative_error = (
-                    0.04 if max_relative_error < 0.04 else max_relative_error
+                    0.01 if max_relative_error < 0.04 else max_relative_error
```
ZzSean: 这个判断是不是得改成<0.01

python/paddle/fluid/tests/unittests/op_test.py
```diff
@@ -2382,7 +2382,7 @@ def check_grad_with_place(
...
```
ZzSean: 同

PR #51100, opened 2023-03-02
[dy2static] fix memory leakage problem in T5 model

2742195759:
### PR types

Bug fixes
### PR changes

OPs
### Describe

fix memory leakage problem in T5 model:
There is memory leakage problem in conditional block.


0x45f (approved):
LGTM

PR #51166, opened 2023-03-03
fluid clean: remove print/switch from fluid to static

qizhaoaoe:
### PR types
 Others

### PR changes
 Others 

### Describe
1. remove print/switch from fluid to static.


python/paddle/static/__init__.py
```diff
@@ -80,6 +79,7 @@
 
 from .nn.metric import auc  # noqa: F401
 from .nn.metric import accuracy  # noqa: F401
+from .nn.control_flow import Print, Switch
```
GuoxiaWang: `Switch` 不需要导入

qizhaoaoe: fix yet.

PR #51208, opened 2023-03-05
[Prim] Filter tensor type for int_array and scalar input in composite rule

cyber-pioneer:
### PR types
Bug fixes

### PR changes
Others

### Describe

Tensor type for int_array and scalar input will cause dynamic shape in static mode, but composite rule does not support currently. 

As for arg "shape" of int_array type. It may be one of [list/tuple of int, tensor]. The former is widely used actually, while the latter is rarely used and will cause dynamic shape in static mode which needs extra unnecessary process. Thus, it would be better to filter it in composite rules.

### DIFF
--- a/python/paddle/incubate/autograd/utils.py
+++ b/python/paddle/incubate/autograd/utils.py
@@ -14,6 +14,7 @@
 import typing
 
 import paddle
+import paddle.framework.dtype as dtypes
 from paddle.fluid import framework as framework
 
 from .phi_ops_map import op_info, op_map
@@ -159,7 +160,7 @@ def _solve_arg(item):
     return arg_type.strip(), arg_name.strip()
 
 
-def _get_attr_value(op, arg_name):
+def _get_attr_value(op, arg_type, arg_name):
     op_content = op_map[op.type]
     if "attrs" in op_content.keys() and arg_name in op_content["attrs"].keys():
         arg_name = op_content["attrs"][arg_name]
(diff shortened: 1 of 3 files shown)


JiabinYang:
> ### PR types
> Bug fixes
> 
> ### PR changes
> Others
> 
> ### Describe
> Tensor type for int_array and scalar input is legacy type. And it will be removed in the future.
> 
> As for arg "shape" of int_array type. It may be one of [list/tuple of int, tensor]. The former is widely used actually, while the latter is rarely used and will cause dynamic shape in static mode which needs extra unnecessary process. Thus, it would be better to filter it in composite rules.

Actually, tensor type for int_arrary and scalar is not legacy type, it just not suitable for composite for now. We may find a way to limit this during dy2st process

python/paddle/incubate/autograd/utils.py
```diff
@@ -159,15 +160,52 @@ def _solve_arg(item):
...
+        # IntArray and Scalar are legacy type. In these case, tensor-relative types are removed in composite op.
+        if arg_type in ("IntArray", "Scalar"):
+            tensor_key = "int_array" if arg_type == "IntArray" else "scalar"
+            if op_content.get(tensor_key):
+                tensor_content = op_content[tensor_key].get(arg_name)
+                if not tensor_content:
+                    raise ValueError(
+                        f'No value found for {arg_name} of {arg_type} type for operator {op.type}.'
+                    )
+                for item in ("tensor_name", "tensor_name"):
```
JiabinYang: why?

JiabinYang (commented):
Add a unit test to check if it can raise error correctly

python/paddle/incubate/autograd/utils.py
cyber-pioneer: it's a typo. Fixed.

JiabinYang (approved):
LGTM

PR #51215, opened 2023-03-06
Comp index select

sljlp:
### PR types

Others
### PR changes

Others
### Describe

add composite index_select op


python/paddle/incubate/autograd/primitives.py
```diff
@@ -124,5 +127,7 @@
     'full',
     'uniform',
     'greater_equal',
+    'stack'
```
cyber-pioneer: why import these ops? they are not used in this composite rule.

cyber-pioneer (dismissed):
ensure code coverage

paddle/phi/kernels/gpu/index_select_grad_kernel.cu
```diff
@@ -31,7 +31,6 @@ template <typename T, typename IndexT>
 __global__ void index_select_grad_cuda_kernel(const T* output_grad,
                                               T* input_grad,
                                               const IndexT* index,
-                                              int64_t nums,
```
cyber-pioneer: why to delete this?

python/paddle/incubate/autograd/composite_rules.py
```diff
@@ -300,6 +300,15 @@ def hard_swish_composite(x):
     return res
 
 
+@REGISTER_COMPOSITE('index_select')
+def index_select_composite(x, index, axis=0):
+    """define composite rule of op index_select."""
```
cyber-pioneer, line 394: composite args will be overwritten. Thus  do not set default here.

paddle/phi/kernels/gpu/index_select_grad_kernel.cu
sljlp: not used in this kernel function.

python/paddle/incubate/autograd/composite_rules.py
sljlp: updated

Charles-hit (approved):

PR #51237, opened 2023-03-06
[with_data_parallel][part8] remove with_data_parallel in OpTest

kangguangli:
### PR types

Others
### PR changes

Others
### Describe

#### Others
Pcard-67004


zhiqiu (approved):
LGTM

Issue #46442, opened 2022-09-23
Collecting Paddle Codebase Authors


jzhang533:
To Paddle codebase authors

Please feel free to create a PullRequest to include your information into [AUTHORS.md](https://github.com/PaddlePaddle/Paddle/blob/develop/AUTHORS.md), and assign that PR to @jzhang533 asking for merge.

There is a incomplete list of [Paddle codebase](https://github.com/PaddlePaddle/Paddle) authors in [AUTHORS.md](https://github.com/PaddlePaddle/Paddle/blob/develop/AUTHORS.md), 85 authors are included, but according to GitHub, there are [622 contributors](https://github.com/PaddlePaddle/Paddle/graphs/contributors).

To create additional visibility for contributor acknowledgment, PaddlePaddle community encourges every Paddle codebase author include his/her GitHub account and fullname into [AUTHORS.md](https://github.com/PaddlePaddle/Paddle/blob/develop/AUTHORS.md).

Thank you for contributing to PaddlePaddle.

ziwang-com:
百度paddle系列模块库架构图map
[https://github.com/ziwang-com/paddle-map](https://github.com/ziwang-com/paddle-map)

# paddle-map
百度paddle系列架构图，全家桶主要模块库都有了
MF团队，全球首家发布百度paddle系列架构图，包括以下主流模块库：
paddle3d,paddle,paddlegan,paddlehub,paddle-lite,paddleocr,paddlespeech,


![paddle-develop_paddle_fluid_eager_auto_code_generator_generator_eager_gen](https://user-images.githubusercontent.com/11691791/227384065-8470ca8d-86b2-4b80-8d04-d2b44a9c3256.png)

俗话说：千言万语不如一张图。zwMAP系列模块库构架图，为国内众多AI初创企业，快速掌握底层核心算法，提供了一个超级神器。

PR #51245, opened 2023-03-06
register as author

will-jl944:
### PR types

Others
### PR changes

Others
### Describe

Register as author, according to https://github.com/PaddlePaddle/Paddle/issues/46442


jzhang533 (approved):
thanks for your contribution.

PR #51369, opened 2023-03-08
[with_data_parallel][part10] remove with_data_parallel in parallel_executor.py

kangguangli:
### PR types

Others
### PR changes

Others
### Describe

`ParallelExecutor` will be deleted in the end but for now we just remove `with_data_parallel`. This PR also deletes several
unit test cases using `ParallelExecutor` directly, they seems to run with single gpu since no cmake test property set. So this should be fine.  

#### Other
Pcard-67004

### DIFF
--- a/python/paddle/fluid/compiler.py
+++ b/python/paddle/fluid/compiler.py
@@ -531,16 +531,6 @@ class CompiledProgram:
             else:
                 self._places = [self._place]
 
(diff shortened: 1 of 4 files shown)


zhiqiu (approved):
LGTM

luotao1 (approved):

PR #51382, opened 2023-03-08
[with_data_parallel][part12] remove with_data_parallel in test_sync_batch_norm_op

kangguangli:
### PR types

Others
### PR changes

Others
### Describe

So we have to replace `with_data_parallel` with `paddle.distributed.launch` to do multi-device training. 
The biggest difference here is for the sake of ease, we only checked the output of device with id `0`.

#### Others
Pcard-67004


zhiqiu (dismissed):
LGTM

zhiqiu (approved):

PR #51383, opened 2023-03-08
[FlashAttention] add raw kernel for flashattn

kuizhiqing:
### PR types

Function optimization

### PR changes

OPs

### Describe

Add raw kernel for flash-attn, which makes it possible to support mask.

UT results,
image


sneaxiy (approved):
LGTM

PR #51390, opened 2023-03-08
Fix nccl_test_op failure on hopper

Wong4j:
### PR types
Bug fixes

### PR changes
Others

### Describe
In our CI tests, nccl_op_test failed on multi-H100. 
For example, suppose I use 2 GPUs and run allreduce test for 4 times, the results are as follows (failure rate 3/4):
```
Input tensor:
    Thread 0: [42, 42, 42, 42, 42, 42, 42, 42]
    Thread 1: [43, 43, 43, 43, 43, 43, 43, 43]
Expected results after allreduce:  
    [85, 85, 85, 85, 85, 85, 85, 85]
Actual results after allreduce:  
    Run1: [0, 0, 0, 0, 0, 0, 0, 0]             Fail
    Run2: [85, 85, 85, 85, 0, 0, 0, 0]         Fail
    Run3: [85, 85, 0, 0, 0, 0, 85, 85]         Fail
    Run4: [85, 85, 85, 85, 85, 85, 85, 85]     Pass
```

- Reason
The stream that the nccl op (such as allreduce) runs on is different from the stream for ctx-Wait() at the end. The nccl op is not synchronized after it runs, so the result copied back to the CPU is wrong. 
Print pointer address of the two CUDA streams:
image

- Solution
I use `paddle::framework::InitDevices()` to create a (singleton) `DeviceContextPool` instance in `SetUp()`. All subsequent `nccl op` and `memcpyAsync` op run on the same stream got from `DeviceContextPool`, and finally use `ctx->wait()` to synchronize.


ForFishes (approved):
LGTM

PR #51438, opened 2023-03-09
Fix custom plugin

heavengate:
### PR types
Bug fixes

### PR changes
Others

### Describe
**fix custom plugin**
- fix custom plugin attributes format compatible custom op
- fix custom plugin attributes setting be rewritten by following attributes


weishengying:
LGTM

PR #51439, opened 2023-03-09
[Paddle Inference] Support dims 1 input of activation and unary op for dynamic shape.

xiaoxiaohehe001:
### PR types

Others
### PR changes

Others
### Describe

Support dims 1 input of activation and unary op for dynamic shape.


zhangjun (approved):
LGTM

PR #51466, opened 2023-03-10
fix UT when np >= 1.23

zoooo0820:
### PR types

Others
### PR changes

Others
### Describe

Pcard-66985
#### Motivation
In Numpy, the behavior of non-tuple sequence for multidimensional indexing has been changed since 1.23.  This incompatible upgrade caused our unittests of indexing to fail, which preventing our CI environment from upgrading to new numpy versions. (see PR https://github.com/PaddlePaddle/Paddle/pull/49556)

In a word, currently, the numpy versions are different between CI and user's environment.
This PR makes indexing unit test works well at numpy version >=1.23 and < 1.23, both.

#### What Numpy changed
  In fact, numpy already gave warnings in version <1.23:
```
__main__:1: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
```
For List case, the outermost `[]` will be treated as tuple `()` in version less than 1.23, which is used to wrap index elements for multiple axes. And from 1.23, this will be treat as a whole and only works on one axis.

#### For future Paddle
Whether the paddle behavior in this case will change is still up for debate. Currently, we still keep the same behavior.

### DIFF
--- a/python/paddle/fluid/tests/unittests/test_var_base.py
+++ b/python/paddle/fluid/tests/unittests/test_var_base.py
@@ -954,16 +954,17 @@ class TestVarBase(unittest.TestCase):
         py_idx = [[0, 2, 0, 1, 3], [0, 0, 1, 2, 0]]
 
         # note(chenjianye):
-        # Non-tuple sequence for multidimensional indexing is supported in numpy < 1.24.
-        # For List case, the outermost `[]` will be treated as tuple `()` in version less than 1.24,
+        # Non-tuple sequence for multidimensional indexing is supported in numpy < 1.23.
+        # For List case, the outermost `[]` will be treated as tuple `()` in version less than 1.23,
         # which is used to wrap index elements for multiple axes.
-        # And from 1.24, this will be treat as a whole and only works on one axis.
+        # And from 1.23, this will be treat as a whole and only works on one axis.
         #
-        # e.g. x[[[0],[1]]] = x[([0],[1])] = x[[0],[1]] (in version < 1.24)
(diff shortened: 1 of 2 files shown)


python/paddle/fluid/tests/unittests/test_var_base.py
```diff
@@ -952,9 +952,21 @@ def _test_list_index(self):
...
         py_idx = [[0, 2, 0, 1, 3], [0, 0, 1, 2, 0]]
+
+        # note(chenjianye):
+        # Non-tuple sequence for multidimensional indexing is supported in numpy < 1.24.
+        # For List case, the outermost `[]` will be treated as tuple `()` in version less than 1.24,
+        # which is used to wrap index elements for multiple axes.
+        # And from 1.24, this will be treat as a whole and only works on one axis.
+        #
+        # e.g. x[[[0],[1]]] = x[([0],[1])] = x[[0],[1]] (in version < 1.24)
+        #      x[[[0],[1]]] = x[array([[0],[1]])] (in version >= 1.24)
```
jeff41404: it is better to understand that using `==` instead of `=`

jeff41404: in numpy version >= 1.24, adding the description of behavior of x[tuple([[0],[1]])]), It is easier to see that such modification is working.

jeff41404 (approved):
LGTM

PR #51482, opened 2023-03-10
[BugFix] fix install_check failure when multi cpu but no gloo

kangguangli:
### PR types

Bug fixes
### PR changes

APIs
### Describe

Fix Bug: install_check failed when paddle is compiled with cpu only.  
We removed the multi-cpu checking since this case has little meaning.

#### Related Issues

#51437 

#### Other
Pcard-67004


pangyoki (approved):
LGTM

PR #51486, opened 2023-03-10
fix conv axis is not the default value -1

veyron95:
### PR types

Others
### PR changes

Others
### Describe

Pcard-66975
fix conv axis is not default -1


python/paddle/nn/functional/conv.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import paddle
```
JiabinYang: Don't import paddle it will cause cycle dependency.

JiabinYang (commented):
one comment

JiabinYang (approved):
LGTM

PR #51554, opened 2023-03-13
Improve stablity of Paddle-TensorRT FP16 UT GitHub (1)

eee4017:
### PR types

Function optimization

### PR changes

Others

### Description

This Pull Request consists of the following changes:

1. In engine.cc, a temporary blacklist function IsFP16BlacklistLayer has been added for Conv2D layers due to a bug in TensorRT versions 8.6.1.0 and later. This function prevents precision settings on Conv2D layers to mitigate the issue until the TensorRT team provides a permanent fix.
2. Also in engine.cc, we ensure that only valid layers are processed in FP16 mode.
3. In auto_scan_test.py, the logging level has been made configurable via the PADDLE_TEST_LOGLEVEL environment variable.
4. Several log messages have been modified to use f-strings for more readable code.
5. In program_config.py, a new method has been added to change the data type of the tensor's data.

### DIFF
--- a/test/ir/inference/auto_scan_test.py
+++ b/test/ir/inference/auto_scan_test.py
@@ -36,7 +36,8 @@ import paddle
 import paddle.inference as paddle_infer
 from paddle.fluid.core import PassVersionChecker
 
-logging.basicConfig(level=logging.INFO, format="%(message)s")
+LOGLEVEL = os.environ.get("PADDLE_TEST_LOGLEVEL", "INFO").upper()
+logging.basicConfig(level=LOGLEVEL, format="%(message)s")
 
 settings.register_profile(
     "ci",
@@ -57,8 +58,8 @@ settings.register_profile(
     report_multiple_bugs=False,
 )
 if (
-    float(os.getenv('TEST_NUM_PERCENT_CASES', default='1.0')) < 1
-    or os.getenv('HYPOTHESIS_TEST_PROFILE', 'dev') == 'ci'
+    float(os.getenv("TEST_NUM_PERCENT_CASES", default="1.0")) < 1
+    or os.getenv("HYPOTHESIS_TEST_PROFILE", "dev") == "ci"
 ):
     settings.load_profile("ci")
 else:
@@ -100,10 +101,10 @@ class AutoScanTest(unittest.TestCase):
 
     @abc.abstractmethod
     def sample_program_configs(self):
-        '''
+        """
         Generate all config with the combination of different Input tensor shape and
         different Attr values.
-        '''
+        """
         raise NotImplementedError
 
     @abc.abstractmethod
@@ -125,9 +126,9 @@ class AutoScanTest(unittest.TestCase):
     def run_test_config(
         self, model, params, prog_config, pred_config, feed_data
     ) -> Dict[str, np.ndarray]:
-        '''
+        """
         Test a single case.
-        '''
+        """
         pred_config.set_model_buffer(model, len(model), params, len(params))
         predictor = paddle_infer.create_predictor(pred_config)
         self.available_passes_in_framework = (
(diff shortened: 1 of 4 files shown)


yuanlehome:
Hi, is this work still being followed up?

eee4017:
> Hi, is this work still being followed up?

Yes, the commit is going to be reviewed. In order to adhere to the contribution guidelines, it will be divided into multiple pull requests. Thanks.

paddle/fluid/inference/tensorrt/engine.cc
```diff
@@ -168,11 +184,62 @@ void TensorRTEngine::FreezeNetwork() {
   if (enable_fp16) {
     bool support_fp16 = infer_builder_->platformHasFastFp16();
     infer_builder_config_->setFlag(nvinfer1::BuilderFlag::kFP16);
+
+#if IS_TRT_VERSION_GE(8210)
+    infer_builder_config_->setFlag(
+        nvinfer1::BuilderFlag::kOBEY_PRECISION_CONSTRAINTS);
```
zhangjun: I don't think it's OK to set ```kOBEY_PRECISION_CONSTRAINTS``` as the default value. Perhaps it may lead to unexpected corruption.

eee4017: The purpose of kOBEY_PRECISION_CONSTRAINTS is to force the TensorRT engine to utilize FP16 kernels. Without this flag, the engine may automatically fall back to using FP32 kernels, as determined by its internal tactic selection mechanism. We apply the kOBEY_PRECISION_CONSTRAINTS flag when the Paddle TensorRT parameter precision is set to FP16. If you have alternative suggestions or recommendations for handling this, please let us know. For instance, do you think it would be more appropriate to introduce a new Paddle TensorRT parameter?

zhangjun: > The purpose of kOBEY_PRECISION_CONSTRAINTS is to force the TensorRT engine to utilize FP16 kernels. Without this flag, the engine may automatically fall back to using FP32 kernels, as determined by its internal tactic selection mechanism. We apply the kOBEY_PRECISION_CONSTRAINTS flag when the Paddle TensorRT parameter precision is set to FP16. If you have alternative suggestions or recommendations for handling this, please let us know. For instance, do you think it would be more appropriate to introduce a new Paddle TensorRT parameter?

Yeah, it would be more appropriate to apply the kOBEY_PRECISION_CONSTRAINTS flag by introducing a alternative Paddle TensorRT parameter.

eee4017: Hello, and thank you for your guidance. I have now added a new API within the Paddle Python library that enables the activation of the kOBEY_PRECISION_CONSTRAINTS flag. Please take a moment to review the implementation and share your valuable feedback. Also, the tests have been successfully passed.

zhangjun (dismissed):
LGTM

paddle/fluid/inference/tensorrt/engine.cc
```diff
@@ -33,6 +33,22 @@ namespace tensorrt {
 int TensorRTEngine::runtime_batch_ = 1;
 thread_local int TensorRTEngine::predictor_id_per_thread = -1;
 
+#if IS_TRT_VERSION_GE(8610)
+// The IsFP16BlacklistLayer function acts as a temporary blacklist for Conv2D
+// layers due to a bug in TensorRT versions 8.6.1.0 and later. It prevents
+// precision settings on Conv2D layers to mitigate the issue until the TensorRT
+// team provides a permanent fix.
```
jiweibo: It seems that this logic needs to be preserved in the future. If the user's Trt version is 8.6.1, the user needs to be reminded to upgrade trt. @jiweibo

qingqing01:
@eee4017 Thank you!  It seems only Group Conv2D has bug. If TensorRT >=8.6.1.0  is used in the product, users also need to set force_obey_fp16_constraints() ?

jiweibo (dismissed):
LGTM

eee4017:
@qingqing01 @jiweibo Thank you for your valuable input and insights. I would like to offer additional clarification on this matter. For the majority of users, enabling force_obey_fp16_constraints is not necessary. However, activating this flag can improve the stability of the FP16 testing process, even though it may not be needed for most day-to-day usage.

Once the flag is enabled, a known issue arises involving group convolutions and force_obey_fp16_constraints in TensorRT versions 8.6.1.0 and later. Unfortunately, as of now, this particular issue has not been addressed by TensorRT.

jzhang533 (dismissed):
LGTM

qingqing01 (dismissed):

jzhang533 (dismissed):
LGTM

qingqing01 (approved):

yuanlehome (approved):
LGTM

PR #51586, opened 2023-03-13
[Zero Dim] hack process Tensor.numpy() from 0D to 1D to avoid much incompatible

zhwesky2010:
### PR types

New features
### PR changes

APIs
### Describe

hack process Tensor.numpy() from 0D to 1D to avoid much incompatible, which should be removed in future.


jeff41404 (approved):
LGTM

PR #51616, opened 2023-03-14
[Executor] rm Executor._run_parallel

kangguangli:
### PR types

Others
### PR changes

APIs
### Describe

The API `Executor.Run` has multiple internal branches. `_run_parallel` is one of them which will call the `core.ParallelExecutor` to execute the program. Since `ParallelExecutor` will be removed finally, this branch should be removed first to avoid the dependency to `ParallelExecutor`.

#### Other
Pcard-67004


zhiqiu (approved):
LGTM

PR #51661, opened 2023-03-14
tensor_meta: construct scalar tensor_meta

xinyu-intel:
### PR types
Function optimization

### PR changes
APIs

### Describe
Construct tensor_meta with scalar info. This can be convenient to check if a tensor is a scalar tensor.


xinyu-intel:
close due to conflict with empty tensor.

PR #51740, opened 2023-03-16
【Eager】Allow return none when stop_gradient=False

JiabinYang:
### PR types

Others
### PR changes

Others
### Describe

This PR allow pylayer return none in backward even if its forward var's stop_gradient=False


wanghuancoder (approved):
LGTM

chenwhql (approved):

PR #51744, opened 2023-03-16
Fix the compilation of rocksdb error (undefined z and dl)

leo0519:
### PR types

Bug fixes

### PR changes

Others

### Describe

The third party rocksdb has compilation error, which happened in other environments except Paddle's image.
This PR adds `--no-as-needed` to claim that we need API definitions of these library (zlib, dl)


Thunderbrook (approved):

zhiqiu (approved):
LGTM

PR #51791, opened 2023-03-17
Adjust tolerance with modi grad

Vvsmile:
### PR types

Others
### PR changes

Others
### Describe

add fp16-fp32-comparison in eager_op_test.py, fix some bugs in op_test.py
the mechanism of new comparison system is same as previous modification in op_test.py( PR: https://github.com/PaddlePaddle/Paddle/pull/50700)


ZzSean (approved):
LGTM，PR描述中补充下前序工作

PR #51806, opened 2023-03-17
rm restict of platform

cyber-pioneer:
### PR types
Others

### PR changes
Others

### Describe
Since composite op grad random error on non-linux platform has not appeared yet. Thus, remove platform linux filter.


JiabinYang (approved):
LGTM

PR #51859, opened 2023-03-20
[Auto Parallel] Add patterns of rule based tuner

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

This PR adds some patterns of GPT3 in rule based tuner, including QKV, FFN and so on. Users can  define pattern by themselves according to these pattern.


JZ-LIANG (approved):
LGTM

PR #51908, opened 2023-03-21
[Auto Parallel] Update rule based tuner

Caozhou1995:
### PR types

New features 
### PR changes

Others
### Describe

This PR adds forward and backward sub program completion when layer with distributed attrs of some tensors.


JZ-LIANG (approved):
LGTM

PR #51959, opened 2023-03-22
[AMP OP&Test] Fix fp16 check_grad when user_defined_grads is not None

ZzSean:
### PR types

Others
### PR changes

Others
### Describe

When user_defined_grads is not None, numeric_grads should not be recompute using PaddlePaddle


JamesLim-sy (approved):
LGTM

PR #51964, opened 2023-03-22
[CodeStyle][C403][C404] Replace unnecessary-list-comprehension-set/dict

enkilee:
### PR types

Others
### PR changes

Others
### Describe

- https://github.com/PaddlePaddle/Paddle/issues/51729
According to https://pypi.org/project/flake8-comprehensions/ 

C403-404: Unnecessary list comprehension - rewrite as a <set/dict> comprehension.
Rules:

C403 Unnecessary list comprehension - rewrite as a set comprehension.

C404 Unnecessary list comprehension - rewrite as a dict comprehension.

It’s unnecessary to use a list comprehension inside a call to set or dict, since there are equivalent comprehensions for these types. For example:

Rewrite set([f(x) for x in foo]) as {f(x) for x in foo}

Rewrite dict([(x, f(x)) for x in foo]) as {x: f(x) for x in foo}

>>> from timeit import timeit
>>> timeit(set([i for i in range(100)]))
CPU times: user 12 µs, sys: 3 µs, total: 15 µs
Wall time: 18.6 µs
>>> timeit("a = (1, 2)")
CPU times: user 10 µs, sys: 0 ns, total: 10 µs
Wall time: 14.3 µs

but in python/paddle/fluid/tests/unittests/test_dist_lookup_sparse_table_fuse_ops.py
scope.var("Ids").get_tensor().set([i for i in range(100)], place) can't replace to 
scope.var("Ids").get_tensor().{i for i in range(100)], place} or 
scope.var("Ids").get_tensor().{place for place in range(100)}
so I think C403 can't using it.
but C404 can use it.

### DIFF
--- a/python/paddle/distributed/fleet/base/role_maker.py
+++ b/python/paddle/distributed/fleet/base/role_maker.py
@@ -1060,9 +1060,7 @@ class PaddleCloudRoleMaker(RoleMakerBase):
         self._trainers_num = trainers_num
         self._role = role
         self._current_id = current_id
-        self._nodes_num = len(
-            set([x.split(':')[0] for x in self._worker_endpoints])
-        )
+        self._nodes_num = len({x.split(':')[0] for x in self._worker_endpoints})
 
     def _collective_env(self):
         self._current_id = int(os.getenv("PADDLE_TRAINER_ID", "0"))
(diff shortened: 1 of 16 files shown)


SigureMo:
> but in file test_dist_lookup_sparse_table_fuse_ops.py ,
> scope.var("Ids").get_tensor().set([i for i in range(100)], place) can't replace to
> scope.var("Ids").get_tensor().{place for place in range(100)}
> I don't know why.

该情况是 Ruff 扫描出来的吗？如果是 Ruff 扫描出来的，那么这是 Ruff 的 bug，目前还不能引入该 rule

该情况的 set 并不是 builtin 的 set 函数，自然是不能替换

enkilee:
> 

我手动查找改的，我的ruff 调用直接说 ：[bash](bash: ruff: command not found)

PR #51968, opened 2023-03-22
[CodeStyle][C403] Unnecessary list comprehension (rewrite as a set comprehension)

enkilee:
### PR types

Others
### PR changes

Others
### Describe

Unnecessary list comprehension (rewrite as a set comprehension)
According to https://beta.ruff.rs/docs/rules/unnecessary-generator-set/ and https://pypi.org/project/flake8-comprehensions/
C403 : Unnecessary list comprehension - rewrite as a set comprehension.

So we can rewrite `set([f(x) for x in foo])` as `{f(x) for x in foo}`

Timing use:
``` python
>>>from timeit import timeit
>>>time(set([i for i in range(100)]))
CPU times: user 12 µs, sys: 3 µs, total: 15 µs
Wall time: 18.6 µs
>>>time({i for i in range(100)})
CPU times: user 10 µs, sys: 0 ns, total: 10 µs
Wall time: 14.3 µs
```

Changed 19 files, 24 places, and no semantic or syntax error. The conclusions are as follows：
using rule：✅ 
auto fix：✅

command as follow：
```bash
# install ruff 0.0.254
pip install ruff==0.0.254
# using command to auto fix:
ruff --select C403 . --fix
```

### DIFF
--- a/python/paddle/distributed/fleet/base/role_maker.py
+++ b/python/paddle/distributed/fleet/base/role_maker.py
@@ -1060,9 +1060,7 @@ class PaddleCloudRoleMaker(RoleMakerBase):
         self._trainers_num = trainers_num
         self._role = role
         self._current_id = current_id
-        self._nodes_num = len(
-            set([x.split(':')[0] for x in self._worker_endpoints])
-        )
+        self._nodes_num = len({x.split(':')[0] for x in self._worker_endpoints})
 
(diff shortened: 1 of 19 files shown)


SigureMo (approved):
LGTM

配置记得在之后 PR 里一并修改

PR #51969, opened 2023-03-22
[CodeStyle][C404] Unnecessary list comprehension (rewrite as a dict comprehension)

enkilee:
### PR types

Others
### PR changes

Others
### Describe

Unnecessary list comprehension (rewrite as a dict comprehension)
According to https://beta.ruff.rs/docs/rules/unnecessary-generator-set/ and https://pypi.org/project/flake8-comprehensions/
C404 : Unnecessary list comprehension - rewrite as a dict comprehension.

So we can rewrite dict([(x, f(x)) for x in foo]) as {x: f(x) for x in foo}

Timing use:
``` python
>>>from timeit import timeit
>>>timeit("dict([(x,x*2) for x in range(1,100)])")
9.798447493463755
>>>timeit("{x:x*2 for x in range(1,100)}")
5.622418139129877

>>>time(dict([(x,x*2) for x in range(1,100)]))
CPU times: user 27 µs, sys: 0 ns, total: 27 µs
Wall time: 31.5 µs
>>>time({x:x*2 for x in range(1,100)})
CPU times: user 15 µs, sys: 0 ns, total: 15 µs
Wall time: 19.6 µs
```

Changed 2 files, 2 places, and no semantic or syntax error. The conclusions are as follows：
using rule：✅ 
auto fix：✅

command as follow：
```bash
# install ruff 0.0.254
pip install ruff==0.0.254
# using command to auto fix:
ruff --select C404 . --fix
```

### DIFF
--- a/python/paddle/distributed/passes/auto_parallel_sharding.py
+++ b/python/paddle/distributed/passes/auto_parallel_sharding.py
@@ -1790,7 +1790,7 @@ def group_param(sharding_info, fuse_size):
 class ShardingInfo:
     def __init__(self, group, rank, params_grads, partition_algor):
         self.group = group
-        self.params_grads = dict([(p.name, (p, g)) for p, g in params_grads])
+        self.params_grads = {p.name: (p, g) for p, g in params_grads}
         assert len(self.params_grads) == len(
             set(self.params_grads)
         ), "found duplicated param in params_grads"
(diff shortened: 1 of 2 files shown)


SigureMo (approved):
LGTM

同 C403，记得之后修改下配置

PR #51978, opened 2023-03-22
[AMP] Add bfloat16 and float16 tests for compare ops

yeliang2258:
### PR types

Performance optimization
### PR changes

OPs
### Describe

Add bfloat16 and float16 tests for compare ops, include equal,  greater_equal, greater_than, less_equal, less_than and not_equal


python/paddle/tensor/logic.py
```diff
@@ -736,13 +736,13 @@ def not_equal(x, y, name=None):
         check_variable_and_dtype(
             x,
             "x",
-            ["bool", "float32", "float64", "int32", "int64"],
+            ["bool", "float16", "float32", "float64", "int32", "int64"],
```
ZzSean, line 739: 加一下uint16

ZzSean (approved):

python/paddle/tensor/logic.py
yeliang2258: https://github.com/PaddlePaddle/Paddle/pull/52016

PR #52015, opened 2023-03-23
[CUSTOM DEVICE]analysis predictor custom device support

piDack:
### PR types
New features

### PR changes
APIs

### Describe
Append some Analysis predictor APIs support for Custom_device.For the reality that native-related(e.g. NativeConfig) inference interfaces will be [deprecated](https://github.com/PaddlePaddle/Paddle/blob/0480ff5de0e919b5bcba96a7b54284d0dcdd55fd/paddle/fluid/inference/api/api.cc#L123), so CUSTOM_DEVICE supported is not considered.


qili93 (approved):
LGTM

PR #52024, opened 2023-03-23
[CodeStyle][UP024] Replace aliased errors with OSError

enkilee:
### PR types

Others
### PR changes

Others
### Describe

According to: https://beta.ruff.rs/docs/rules/#pyupgrade-up
```bash
UP024 | os-error-alias | Replace aliased errors with OSError
```
and
https://pypi.org/project/pyupgrade/
```bash
OSError aliases
 # also understands:
 # - IOError
 # - WindowsError
 # - mmap.error and uses of `from mmap import error`
 # - select.error and uses of `from select import error`
 # - socket.error and uses of `from socket import error`

 def throw():
-    raise EnvironmentError('boom')
+    raise OSError('boom')

 def catch():
     try:
         throw()
-    except EnvironmentError:
+    except OSError:
         handle_error()
```

So 3 files changed, 8 places replace it.
1 EnvironmentError -> OSError
2 IOError, OSError -> OSError
3 IOError -> OSError

As we can see the documents from Official web: https://docs.python.org/3/library/exceptions.html#exception-hierarchy
Changed in version 3.3: EnvironmentError, IOError, WindowsError, socket.error, select.error and mmap.error have been merged into OSError, and the constructor may return a subclass.

So we can use it safely.
using rule：✅
auto fix：✅ 

The command as follow:
```bash
# install ruff 0.0.254
pip install ruff==0.0.254
# auto fix
ruff --select UP024 . --fix
```


SigureMo (approved):
LGTM

PR #52025, opened 2023-03-23
[CodeStyle][UP027] Replace unpacked list comprehension with a generator expression

enkilee:
### PR types

Others
### PR changes

Others
### Describe

1 
Add rules UP024 UP027 in pyproject.toml file

2
According to: https://beta.ruff.rs/docs/rules/#pyupgrade-up
```bash
UP027 | unpacked-list-comprehension | Replace unpacked list comprehension with a generator expression
```
and
https://pypi.org/project/pyupgrade/
Unpacking list comprehensions
```bash
-foo, bar, baz = [fn(x) for x in items]
+foo, bar, baz = (fn(x) for x in items)
```
we can rewrite  [fn(x) for x in items] as  (fn(x) for x in items)

See the python code as follow, we can see after fix UP027, as faster as before:

```python
>>>timeit("[x*5 for x in [10,20,30,40,50,60,70,80]]")
0.4946538507938385
>>>timeit("(x*5 for x in [10,20,30,40,50,60,70,80])")
0.26481082290410995
```

So 6 files changed, 12 places replace it, and no semantic or syntax error. We can use it safely.

using rule：✅
auto fix：✅ 

The command as follow:
```bash
# install ruff 0.0.254
pip install ruff==0.0.254
# auto fix
ruff --select UP027 . --fix
```


SigureMo (approved):
LGTM

luotao1:
@enkilee 卡住的CI需要rerun下

PR #52041, opened 2023-03-23
[Auto Parallel] Add o1 level tune

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

This PR adds o1 level tune of rule based tuner, in this level, every layer has the same parallel strategy.


JZ-LIANG (approved):
LGTM

PR #52110, opened 2023-03-24
[Dy2static] [bugfix] fixed a bug which happens while parsing grad var name

feifei-111:
### PR types

Bug fixes
### PR changes

APIs
### Describe

fixed a bug which happens while parsing grad var name
PCard-66972


0x45f (approved):

2742195759 (approved):
LGTM

PR #52119, opened 2023-03-24
fix scope reuse problem in run_program_api

zhiqiu:
### PR types

Performance optimization
### PR changes

Others
### Describe

In #51411, marked scope `can_resue=true` is defered to `~GradNodeRunProgram`. 
However, the `~GradNodeRunProgram` is called when the forward part of the second batch execution, so the first batch execution and second batch execution cannot reuse the same scope.

This PR fixes that.


2742195759 (approved):
LGTM

Shixiaowei02 (approved):

PR #52408, opened 2023-03-31
modify set_value op to use Scalars attribute

iclementine:
### PR types

Breaking changes

### PR changes

OPs

### Describe

1. modify set_value op, use Scalars to represent attr `values`, instead of a bunch of attributs of various types;
2. modify slice assignment (`__setitem__` of Tensors(or Variables) in various execution mode (static, dynamic, dynamic legacy);
3. add program converter and set_value op as an example, which provides the functionality to convert `paddle::framework::ProgramDesc` between old and new formats(the differences are mainly some operators with incompatible updates in the definition);
4. program version and operator version map now are always saved when serializing `paddle::framework::ProgramDesc` to identify the version;
5. provide an option `legacy_format=false` in  serialization of `paddle::framework::ProgramDesc`, it determines whether to convert ProgramDesc back to legacy format, which is compatible for paddle 2.4.2 or earlier versions to load and execute;
6. deserialization of `paddle::framework::ProgramDesc` is now automatically detecting whether the bytes it receives is in legacy format(contains any of the operators that has been incompatibly updated and have any attribute of type `Scalar`) and convert it to new format. But if you want a faithful deserialization without the automatic conversion, you can use protobuf's deserialization instead. Though it is not recommended, it can be used for the purpose of testing.

Pcard-66983


jeff41404 (dismissed):
LGTM

Shixiaowei02 (dismissed):

XiaoguangHu01 (dismissed):
LGTM

python/paddle/static/io.py
```diff
@@ -459,6 +461,8 @@ def save_inference_model(
 
             - clip_extra(bool): the flag indicating whether to clip extra information for every operator. Default: True.
 
+            - legacy_format(bool): whether to save inference model in legacy format. Defaults to False.
```
jiweibo: Please update doc after pr merge.

jiweibo (dismissed):

chenwhql (dismissed):

python/paddle/fluid/io.py
```diff
@@ -135,6 +135,7 @@ def save_inference_model(
     export_for_deployment=True,
     program_only=False,
     clip_extra=True,
+    legacy_format=False,
```
jzhang533, line 138: 这个API已经被标记位[废弃](https://github.com/PaddlePaddle/Paddle/blob/d3c17f522e0a2d7e71144d08680daf711fcc25fc/python/paddle/fluid/io.py#L126)了。

是不是索性就保持该API原样，不再做改动了。

python/paddle/static/io.py
```diff
@@ -252,6 +252,7 @@ def serialize_program(feed_vars, fetch_vars, **kwargs):
         kwargs: Supported keys including ``program``. Attention please, kwargs is used for backward compatibility mainly.
 
             - program(Program): specify a program if you don't want to use default main program.
+            - legacy_format(bool): whether to save inference program in legacy format. Defaults to False.
```
jzhang533, line 255: `legacy_format`这个参数命名，我第一眼看上去的时候，有个误解，是以为legacy指的是2.0之前版本。
但我确实也没想要更恰当的名称。

python/paddle/fluid/io.py
iclementine: 它仍然被使用，所以也一起改了

jeff41404 (approved):

jzhang533 (approved):

Shixiaowei02 (approved):

jiweibo (approved):

XiaoguangHu01 (approved):
LGTM

PR #52449, opened 2023-04-02
[Prim] polish prim arg None check

cyber-pioneer:
### PR types
Others

### PR changes
Others

### Describe
Pcard-66975
In some case, inputs and outputs of composite op or its replaced composite rule might be None.
It means such arg will be no longer required in processed program by composite mechanism.
Therefore, such special ops should be recorded in advance and be released in args check.

For example, output reserve_space of batch_norm is not needed after replace by composite rule.


JiabinYang (approved):
LGTM

PR #52492, opened 2023-04-04
[DDim Decouple enforce.h] Change enforce.h->exception.h

jiahy0825:
### PR types
Function optimization

### PR changes
APIs

### Describe
Pcard-66989

Fix the following error when `#include "paddle/phi/core/ddim.h"` in C++ level.
"enforce.h" relies on "gflags.h", but "gflags.h" has not been exposed. Hence we need to change "enforce.h" to "exception.h".

1e8611675a682830fbb0bda8d5350998

### DIFF
--- a/paddle/phi/core/ddim.h
+++ b/paddle/phi/core/ddim.h
@@ -18,7 +18,7 @@
 #include <string>
 #include <vector>
 
-#include "paddle/phi/core/enforce.h"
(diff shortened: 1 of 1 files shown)


zh794390558 (approved):
LGTM

YuanRisheng (approved):

zyfncg (approved):

PR #52554, opened 2023-04-05
[Auto Parallel]  Randomness Control for Distributed Training

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Describe
This PR implement the random control for auto parallel.  

Random control maintain the randomness when tensor is distributed across devices on a Mesh. 
    * Independency: If tensor is **Sharded** on a Mesh dimension, Devices along that Mesh dimension should have Different randomness.
    * Consistency:  Meanwhile if the tensor is **Replicated** on another Mesh dimension, randomness of Devices along that Mesh dimension should be Consistent.

The Tensor Parallel randomness control between "Global & Local" regions in following picture is a **Special Case when the Mesh is 1D**.  [ref](https://arxiv.org/abs/1909.08053)
image

This PR implement rng ctrl logic in more general cases with High-Order Mesh.  
For Example 2D Mesh scenario would be like following:

 * 2D Mesh = [2, 4], dims_mapping = [0, -1] for a 2D Tensor: 
image

 * 2D Mesh = [2, 4], dims_mapping = [-1, 1] for a 2D Tensor: 
image

 * 2D Mesh = [2, 4], dims_mapping = [0, 1] for a 2D Tensor: 
image

This PR assign operators with a RNG according to how its input tensor are distributed (Shard-Replicated) across Mesh. 
Each random-relatived op maybe assigned an individual RNG if their input tensors are sharded differently, in order to support **Layer-Wise Parallelism** in Auto Parallel. (There is not more concept like  mp_local_rng or mp_global_rng for entire network.)

This PR is compatible with Recompute in auto parallel, while coordinating randomness across devices in different parallel，it maintain the same randomness between forward and recompute stage within same device.


python/paddle/distributed/auto_parallel/random.py
```diff
@@ -0,0 +1,138 @@
...
+_logger = get_logger(logging.INFO)
+
+_rng_name_to_seed = {}
+_inited_rng_name_to_seed = {}
+_enable_random_control = False
+_basic_seed = 42
+
+# use Prime number as offset to avoid confict
+_mesh_offset = 173
+_dim_offsets = [11, 23, 37, 73]
```
aoyulong, line 31: We may need a better way to compute the offsets in the future to cover all the situations.

aoyulong (approved):
LGTM

python/paddle/distributed/auto_parallel/random.py
JZ-LIANG: the main problem here is offset may conflict when nranks is very high.
the plan is to collect all unique RNG_Names (sharding_expr) among all ranks to a master process, and assign a unique seed to each RNG_Name by master.

This need to wait the Adaption of Auto Parallel to New Communication Lib, where we could use TCP K-V Store  for above collection and coordination of RNG_Name among cluster.

XieYunshen (approved):
LGTM for 
`set_tests_properties(test_random_ctrl PROPERTIES LABELS "RUN_TYPE=EXCLUSIVE"
                                                   TIMEOUT 50)`

PR #52573, opened 2023-04-06
[enforce.h Decouple gflags.h] Move gflags.h from enforce.h to enforce.cc

jiahy0825:
### PR types
Function optimization

### PR changes
Others

### Describe
Pcard-66989

#### Background
After exposing the header files of phi, error will occur when users include header files which depend on "gflags.h" and "glog/logging.h", so that we need to decouple such dependence of header files.

#### Changes
This PR moves "gflags.h" from enforce.h to enforce.cc.

### DIFF
--- a/paddle/phi/api/profiler/device_tracer.cc
+++ b/paddle/phi/api/profiler/device_tracer.cc
@@ -21,6 +21,7 @@ limitations under the License. */
 #include <string>
 #include <thread>  // NOLINT
 
+#include "gflags/gflags.h"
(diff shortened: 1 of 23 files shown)


zh794390558 (approved):
LGTM

ZzSean (approved):
LGTM for CI-OP-Benchmark

zyfncg (approved):

XiaoguangHu01 (approved):
LGTM

PR #52589, opened 2023-04-06
[API] Expose capi in MacOS to enable GPU computing as custom device

lishicheng1996:
### PR types
 
Others

### PR changes

APIs
### Describe

Expose capi in MacOS to enable GPU computing as custom device. Details described [here](https://github.com/PaddlePaddle/PaddleCustomDevice/pull/485).


qili93 (approved):
LGTM

PR #52644, opened 2023-04-07
fix fused_dropout_add data_type error

kuizhiqing:
### PR types

Bug fixes

### PR changes

OPs

### Describe

* new op should be in the list for auto parallel
* add `fused_dropout_add_grad` data_type explicit


ForFishes (approved):
LGTM

PR #52652, opened 2023-04-07
[StandaloneExe] add strategy force_sequential_run

kangguangli:
### PR types

Others
### PR changes

APIs
### Describe

Add a new `BuildStrategy`: `force_sequential_run`, which will let the operations run by the order in the `Program`.

#### Other
Pcard-67004


From00 (approved):
A better config name is required. 
We already have a config FLAGS named `sequential_run`, therefore `force_sequential_run` is confusing. It is recommended to reuse the existing name since they have the same semantics.

kangguangli:
> A better config name is required. We already have a config FLAGS named `sequential_run`, therefore `force_sequential_run` is confusing. It is recommended to reuse the existing name since they have the same semantics.

@zhiqiu Do you have any suggestions? I'll change the name in next PR.

PR #52682, opened 2023-04-08
add autogen code support for logcumsumexp op

GreatV:
### PR types
Others

### PR changes
Others

### Describe

add autogen code support for `logcumsumexp` op


heavyrain-lzy (approved):
LGTM

PR #52845, opened 2023-04-12
add uint16 for bfloat16 dtype check in layer_norm under static mode 

GuoxiaWang:
### PR types

Function optimization
### PR changes

APIs
### Describe

add uint16 for bfloat16 dtype check in layer_norm under static mode


ZzSean (approved):
LGTM

PR #52883, opened 2023-04-13
[Decouple enforce.h] Move LOG from enforce.h to enforce.cc

jiahy0825:
### PR types
Bug fixes

### PR changes
Others

### Describe
Pcard-66989

[Decouple enforce.h] Move LOG from enforce.h to enforce.cc

ThrowWarnInternal uses `LOG` macro to display warning message, which depends on the third-party header file "logging.h". However, "logging.h" has not been exposed to site-package yet, so that error will occur when we include "enforce.h" header file. Hence, we declare a function in enforce.h and define it in enforce.cc file.

### DIFF
--- a/paddle/fluid/platform/CMakeLists.txt
+++ b/paddle/fluid/platform/CMakeLists.txt
@@ -132,7 +132,8 @@ cc_library(
        ${MLU_CTX_DEPS}
        phi_backends
        phi_device_context
-       generator)
+       generator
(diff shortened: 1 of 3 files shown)


YuanRisheng (approved):

zyfncg (approved):

heavyrain-lzy (approved):
LGTM

PR #52913, opened 2023-04-14
[Fix Bug] Fix CustomOP XPU unittest error, delete where

jiahy0825:
### PR types
Bug fixes

### PR changes
Others

### Describe
Pcard-66988

CI link: https://xly.bce.baidu.com/paddlepaddle/paddle/newipipe/detail/8234835/job/22443291

image

XPU 调用 where API 时会报 segment fault，将此调用替换成别的实现。

segment fault will raise when `where` API is called under the XPU scenario, thus substituting calling API to another implementation.


jiahy0825:
It seems that codes like `grad_out.data<data_t>()` in the XPU scenario will raise an error, maybe such an operation can only be called in the CPU scenario?

PR #52923, opened 2023-04-14
[CustomDevice] fix custom cpu unittests failure when multiple python exist in CI

ronny1996:
### PR types

Bug fixes
### PR changes

Others
### Describe

 fix custom cpu unittests failure when multiple python exist in CI


qili93 (approved):
LGTM

PR #52928, opened 2023-04-14
[Auto Parallel]Add o2 tune of rule based tuner

Caozhou1995:
### PR types

New features
### PR changes

Others
### Describe

This PR adds o2 tune of rule based tuner which is slower than o1 tune but more accurate.


XieYunshen (approved):
LGTM for `set_tests_properties(test_rule_based_tuner_o2 PROPERTIES TIMEOUT 50)`

PR #52944, opened 2023-04-15
Update setup.py

sb508:
Hello, I found the use of the os. open() function in the setup.py file while using the code, which can execute system commands. When input parameters are not strictly filtered, it may lead to the execution of arbitrary system commands, forming a command injection vulnerability.

Vulnerability principle:
os. open() can execute system commands. Without properly filtering user input, if the user concatenates their input as a parameter of the system command into the command line, it can cause a vulnerability in command injection. For example, if the user inputs a parameter of "ls - l; ping 1.2.3.4", both instructions will be executed, and the user can independently control the form of the second command, which endangers system security.

Repair suggestions:
Use shlex. quote() to strictly filter the input of the function. The function's functions include adding single quotes to the outermost layer of the string to make it appear as a single entity, and using double quotes to enclose single quotes within the string to prevent it from losing its possible closure function.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ import re
 import shutil
 import subprocess
 import sys
+import shlex
 from contextlib import contextmanager
 from distutils.spawn import find_executable
 from subprocess import CalledProcessError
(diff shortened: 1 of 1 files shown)


risemeup1:
Thank you for reporting this vulnerability. You can change the logic of python/setup.py.in synchronously，the same problem may exist . When all the CI  have passed, you can ask me for approval.

risemeup1 (approved):
LGTM

PR #53000, opened 2023-04-17
[Zero-Dim] fix reduce all and not keep dims case

YangQun1:
### Describe
Depend on https://github.com/PaddlePaddle/Paddle/pull/52739, to fix the reduce all case.
With https://github.com/PaddlePaddle/Paddle/pull/52739 changes, the output dims of reduce sum op changed from {1}  to {} for reduce_all=true and keep_dims=false case.

### DIFF
--- a/paddle/fluid/operators/reduce_ops/reduce_mean_op.cc
+++ b/paddle/fluid/operators/reduce_ops/reduce_mean_op.cc
@@ -98,10 +98,9 @@ class __reduce_meanMaker__ : public ops::ReduceBaseOpMaker {
   virtual std::string GetOpType() const { return "Reduce reduce_mean"; }
 };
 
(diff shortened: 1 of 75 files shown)


YangQun1:
cc @zhouwei25 Could you help to pick the commit to your branch and try this fix?

zhwesky2010:
> cc @zhouwei25 Could you help to pick the commit to your branch and try this fix?

thx, I think the fix is in `paddle/phi/kernels/onednn/reduce_kernel_impl.h` , I'll test it or you can just merge this fix.

YangQun1:
Close this PR since the fix has been included in develop branch (https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/phi/kernels/onednn/reduce_kernel_impl.h#L105-L108)

Issue #52319, opened 2023-03-29
昇腾和寒武纪相关代码退场 Tracking Issue


GreatV:
一、 项目说明
用于跟踪 [昇腾和寒武纪相关代码退场](https://github.com/PaddlePaddle/Paddle/discussions/51794#discussioncomment-5462528) 并记录该计划下各个子任务的计划和进度。

二、 主要工作 (18/18)

> 按 merge 的时间顺序，排名不分先后： @KimBioInfoStudio (5) @mrcangye (2) @jjyaoao (3) @Young-Flash (1) @Liyulingyue (3) @engineer1109 (1) @GreatV (2)

|任务编号|任务内容|认领人|PR链接|
|:----|:----|:----|:----|
|1|删除 `paddle/fluid/platform/device/mlu` 目录✅(2023/3/31)| @mrcangye  | #52382|
|2|删除 `paddle/fluid/operators/*_mlu.*` 文件✅(2023/4/3)| @Young-Flash  |  #52435|
|3|清理 `WITH_MLU` 相关代码✅(2023/4/13)| @GreatV @jjyaoao |#52546✅ #52825✅|
|4|清理 `is_compiled_with_mlu` 相关代码✅(2023/4/3)| @KimBioInfoStudio  |  #52378 |
|5|删除 `paddle/fluid/operators/*_npu.*`、清理 `WITH_ASCEND_INT64` 相关代码✅(2023/4/11)| @jjyaoao @mrcangye | #52673 #52674 #52675 #52676 #52677 #52678 #52679 #53342|
|6|删除 `/paddle/fluid/platform/device/npu` 目录✅(2023/4/1)| @jjyaoao | #52384|
|7|清理 `WITH_ASCEND_CL PADDLE_WITH_ASCEND_CL WITH_ASCEND_CXX11` 相关代码✅(2023/4/13)| @engineer1109 @Liyulingyue @jjyaoao | #52448✅ #52535✅ #52612✅ #52825✅|
|8|清理 `is_compiled_with_npu` 相关代码✅(2023/4/6)|  @KimBioInfoStudio  |#52385 |
|9|清理 `PADDLE_WITH_ASCEND` 相关代码✅(2023/4/13)| @mrcangye @jjyaoao  | #52403✅ #52830✅|
|10|清理 `WITH_ASCEND` 相关代码✅(2023/4/10) |  @Liyulingyue  | #52669|
[...]

PR #53058, opened 2023-04-19
rem hccl keywords

KimBioInfoStudio:
### PR types

Others
### PR changes

Others
### Describe

https://github.com/PaddlePaddle/Paddle/issues/52319

### DIFF
--- a/paddle/fluid/distributed/collective/custom_ccl_tools.cc
+++ b/paddle/fluid/distributed/collective/custom_ccl_tools.cc
@@ -29,7 +29,7 @@ phi::ccl::CCLReduceOp ToCustomCCLRedType(ReduceOp reduction) {
   PADDLE_ENFORCE_EQ(
       it != red_type.end(),
       true,
-      platform::errors::InvalidArgument("Invalid hccl reduction. "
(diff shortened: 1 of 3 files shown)


paddle/fluid/distributed/collective/custom_ccl_tools.cc
```diff
@@ -29,7 +29,7 @@ phi::ccl::CCLReduceOp ToCustomCCLRedType(ReduceOp reduction) {
   PADDLE_ENFORCE_EQ(
       it != red_type.end(),
       true,
-      platform::errors::InvalidArgument("Invalid hccl reduction. "
+      platform::errors::InvalidArgument("Invalid CustomCCL reduction. "
```
KimBioInfoStudio, line 32: I have some questions on this block of code, is this only work for hccl now or any other custom device? if the 2nd condition, is it reasonable to check reduce op in [min max sum prod]? @qili93 @ronny1996

ronny1996: this block work for custom device

ronny1996 (approved):
LGTM

PR #53129, opened 2023-04-20
fix fused bias grad add nan

FeixLiu:
### PR types
Others

### PR changes
Others

### Describe
fix fused bias grad add nan
 develop pr: https://github.com/PaddlePaddle/Paddle/pull/53118


FeixLiu:
https://github.com/PaddlePaddle/Paddle/pull/53118

PR #53191, opened 2023-04-21
rename monkey_patch_{math_}varbase as monkey_patch_{math_}tensor

liudongxue01:
### PR types

Others

### PR changes

Others

### Description

Rename monkey_patch_varbase as monkey_patch_tensor & monkey_patch_math_varbase as monkey_patch_math_tensor

(1) Since varbase has been rename as tensor in eager mode, hence we rename monkey_patch_varbase as monkey_patch_tensor & monkey_patch_math_varbase as monkey_patch_math_tensor respectively.
(2) When I planed to remove the redundant monkey-patch code in paddle.__init__.py I noticed that would make some unittests failed. We should refactor the code till the monkey_patch_variable() and monkey_patch_math_tensor() need not to be called twice(may be in static mode), thus we can remove them in paddle.__init__.py.


lanxianghit (approved):
LGTM for __init__.py

luotao1 (approved):
LGTM for framework.py

wanghuancoder (approved):
LGTM

PR #53245, opened 2023-04-23
[Dy2St]Fix x grad names when high order gradient

0x45f:
### PR types

Bug fixes

### PR changes

Others 

### Description

[Dy2St]Fix x grad names when high order gradient
PCard-66972


0x45f:
ref: https://github.com/PaddlePaddle/Paddle/pull/53250

PR #53247, opened 2023-04-23
fix shared memory over usage in embedding grad kernel on deterministic mode

shaojiewang:
### PR types

Bug fixes 
### PR changes

OPs
### Description

1. fix shared memory over usage in embedding grad kernel on deterministic mode
2. use mptype to accumulate


shaojiewang:
I'll add a unit test ASAP

Xreki (approved):
LGTM

PR #53321, opened 2023-04-25
[Fix Patch Error] Fix patch error in cmake

jiahy0825:
### PR types
Bug fixes

### PR changes
Others

### Description
Pcard-66988

**背景**：CI 开启 cache 的场景下，patch 代码时会报错。不开启 cache 可以正常编译。
image

**分析**：cache 的编译产物有两种可能：1. cache 内的 third_party TAG_VERSION 与最新 cmake 中的 TAG_VERSION 不一致；2. cache 内已经打过 patch。这两种可能都会导致 patch 失败。

**解决**：在 `PATCH_COMMAND` 前使用 git 命令，将 cache 里的 third_party 代码复原。

Ref: https://cmake.org/cmake/help/latest/module/ExternalProject.html

image

[Why calling some `git` commands before `patch`?]
Paddle's CI uses cache to accelarate the make process. However, error might raise when patch codes in two scenarios:
1. Patch to the wrong version: the tag version of CI's cache falls behind PYBIND_TAG, use `git checkout ${PYBIND_TAG}` to solve this.
2. Patch twice: the tag version of cache == PYBIND_TAG, but patch has already applied to cache.
Hence, we need to use git command to restore the code changes in the CI's cache.

### DIFF
--- a/cmake/external/eigen.cmake
+++ b/cmake/external/eigen.cmake
@@ -51,9 +51,10 @@ if(CMAKE_COMPILER_IS_GNUCC)
   if(GCC_VERSION GREATER_EQUAL "12.0")
     file(TO_NATIVE_PATH ${PADDLE_SOURCE_DIR}/patches/eigen/Complex.h.patch
          complex_header)
+    # See: [Why calling some `git` commands before `patch`?]
     set(EIGEN_PATCH_COMMAND
-        patch -Nd ${EIGEN_SOURCE_DIR}/Eigen/src/Core/arch/SSE/ <
-        ${complex_header})
+        git checkout -- . && git checkout ${EIGEN_TAG} && patch -Nd
+        ${EIGEN_SOURCE_DIR}/Eigen/src/Core/arch/SSE/ < ${complex_header})
   endif()
 endif()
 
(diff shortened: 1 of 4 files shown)


cmake/external/eigen.cmake
```diff
@@ -51,9 +51,10 @@ if(CMAKE_COMPILER_IS_GNUCC)
   if(GCC_VERSION GREATER_EQUAL "12.0")
     file(TO_NATIVE_PATH ${PADDLE_SOURCE_DIR}/patches/eigen/Complex.h.patch
          complex_header)
+    # See: [Why calling some `git` commands before `patch`?]
     set(EIGEN_PATCH_COMMAND
-        patch -Nd ${EIGEN_SOURCE_DIR}/Eigen/src/Core/arch/SSE/ <
-        ${complex_header})
+        git checkout -- . && git checkout ${EIGEN_TAG} && patch -Nd
```
heavyrain-lzy, line 56: 该PR是不是预计将所有Paddle的`ExternalProject_Add`中的`patch`命令都进行了修改？如果是，`cmake/external/cudnn-frontend.cmake`中也有一处该命令。

heavyrain-lzy (approved):
LGTM

cmake/external/eigen.cmake
jiahy0825: 受影响的似乎只有最近用到 patch 的 PR，例如：
- https://github.com/PaddlePaddle/Paddle/pull/53256
如果其他第三方库仍有问题，再修复

Seems only recently merged patch-relevant PRs will cause this error. We'll fix the other third-party library if the same error occurs.

PR #53449, opened 2023-04-28
Fix paddle::Tensor usage in predictor.run, tensor should be defined instead of initialized.

yuanlehome:
### PR types

Others
### PR changes

Others
### Description

predictor.run: tensor should be [defined] instead of [initialized].

image


zh794390558 (approved):
LGTM

chenwhql (approved):

qingqing01 (approved):

PR #53545, opened 2023-05-06
[BugFix][StandaloneExe] fix: NewExe did not gc transferred variables

kangguangli:
### PR types

Bug fixes
### PR changes

Others
### Description

`StandaloneExecutor` will create variables during its build, but these variables are not cleared after build. This PR fixes this.

#### Others
Pcard-67004


zhiqiu (approved):
LGTM

PR #53552, opened 2023-05-06
[zero-dim] hack 1-D tensor to Scalar

zoooo0820:
### PR types

Others
### PR changes

APIs
### Description

Pcard-66985

This is a following of https://github.com/PaddlePaddle/Paddle/pull/53358 . For 1-D index Tensor, this PR hack it to a Scalar, not Vector. This may cause difference of results.

```python
>>> a = paddle.randn((2,3,4))
>>> index_0d = paddle.to_tensor(1)
>>> a[index_0d].shape
[3, 4]

>>> index_1d = paddle.to_tensor([1])

# with FLAGS_set_to_1d=True (Default)
>>> a[index_1d].shape    # In this case, is same as 0-d Tensor
/usr/local/lib/python3.8/dist-packages/paddle/fluid/variable_index.py:291: UserWarning: 1-D Tensor will be treat as advanced indexing in future version. Currently, 1-D Tensor means a scalar, not vector, and please modify it to 0-D Tensor.If advanced indexing is needed, please use `export FLAGS_set_to_1d=False` to set the flag.
  warnings.warn(
[3, 4]

# with FLAGS_set_to_1d=False
>>> a[index_1d].shape   # In this case, is different with 0-d Tensor.
[1, 3, 4]
```

### DIFF
--- a/python/paddle/distributed/auto_parallel/operators/dist_slice.py
+++ b/python/paddle/distributed/auto_parallel/operators/dist_slice.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import paddle
+
 from ..utils import compute_compatible_dim_mapping, is_dim_shard
 from .common import (
     DistributedOperatorImpl,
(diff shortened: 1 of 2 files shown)


zhwesky2010 (approved):
后续可以和numpy一起移除warning，目前累计一共3个地方的trick和warning了

PR #53554, opened 2023-05-06
[inference][trt] add lookup_table op trt converter, use trt gather layer

yuanlehome:
### PR types

Others
### PR changes

Others
### Description

add lookup_table op trt converter, use trt gather layer.

ut ref to https://github.com/PaddlePaddle/Paddle/pull/53563


jiweibo (approved):
LGTM

PR #53592, opened 2023-05-08
[Zero-Dim] add 0D test for linalg.norm/linalg.cond

GGBond8488:
### PR types
Others 

### PR changes
 Others

### Description
Pcard-66984
add 0D test for linalg and linalg.cond


zhwesky2010 (approved):
LGTM

PR #53608, opened 2023-05-08
[LAUNCH] add log overwrite flag

kuizhiqing:
### PR types

Others

### PR changes

Others

### Description

With arg `--log_overwrite=True`, log files will be overwritten.

### DIFF
--- a/python/paddle/distributed/launch/context/args_envs.py
+++ b/python/paddle/distributed/launch/context/args_envs.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import os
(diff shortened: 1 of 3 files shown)


python/paddle/distributed/launch/context/args_envs.py
```diff
@@ -48,6 +49,15 @@ def fetch_envs():
     return os.environ.copy()
 
 
+def str2bool(v):
```
sneaxiy: Use `distutils.util.strtobool`.

zyfncg (approved):

sneaxiy (approved):

PR #53634, opened 2023-05-09
[inference Zero-Dim]prelu trt converter support zero dim tensor

yuanlehome:
### PR types

Others
### PR changes

Others
### Description

prelu trt converter support zero dim tensor.


zhangjun (approved):
LGTM

PR #53654, opened 2023-05-09
Fix bug in log_softmax kernel when lastdim is larger than 100000

ZzSean:
### PR types

Bug fixes
### PR changes

OPs
### Description

Fix bug in log_softmax kernel when lastdim is larger than 100000

There is an unexpected `log` in the calculation


Xreki (approved):
LGTM

PR #53660, opened 2023-05-09
[inference Zero-Dim][trt]scale, square, sum, swish trt op converter support zero dim

yuanlehome:
### PR types

Others
### PR changes

Others
### Description

scale, square, sum, swish trt op converter support zero dim.


zhangjun (approved):
LGTM

PR #53682, opened 2023-05-10
disable __setitem__ in static mode & add API paddle.static.setitem with dy2st strategy

zoooo0820:
### PR types

Breaking changes
### PR changes

APIs
### Description

Pcard-66985

#### Motivation
Currently, the `__setitem__`  in static mode is implemented by directly overwriting the result on the original Variable. This leads to although **the forward result is correct**, the calculation result **is wrong in the backward stage** （some are because result relies on the initial value of this Variable; some are because the graph in backward constructed according to this also has the problem of overwriting ).

```python
# Dygraph is right
>>> a = paddle.full((2,3,4), 1.2, dtype='float32')
>>> a.stop_gradient=False
>>> v = paddle.ones((1,), dtype='float32')
>>> v.stop_gradient=False
>>> b = a * 2
>>> b[0,:, 2] = b[0, :, 2] * v
>>> l = b.mean()
>>> l.backward()
>>> b.sum()
Tensor(shape=[], dtype=float32, place=Place(gpu:0), stop_gradient=False,
       57.60000229)
>>> a.grad.sum()
Tensor(shape=[], dtype=float32, place=Place(gpu:0), stop_gradient=False,
       2.)
>>> v.grad
Tensor(shape=[1], dtype=float32, place=Place(gpu:0), stop_gradient=False,
       [0.30000001])
```
```python
# In static mode, the forward b is right, but grad of a and v are not
>>> paddle.enable_static()
>>> exe = paddle.static.Executor()
>>>
>>> train_program = paddle.static.Program()
>>> startup_program = paddle.static.Program()
>>> with paddle.static.program_guard(train_program, startup_program):
...     a = paddle.full((2,3,4), 1.2, dtype='float32')
...     a.stop_gradient=False
...     v = paddle.ones((1,), dtype='float32')
...     v.stop_gradient=False
...     b = a * 2
...     b[0, :, 2] = b[0, :, 2] * v
...     l = b.mean()
...     paddle.static.append_backward(l)
...     res = exe.run(fetch_list=[b, a.grad_name, v.grad_name])
...     print('b.sum() =', res[0].sum())
...     print('a.grad.sum() = ', res[1].sum())
...     print('v.grad.sum() = ', res[2].sum())
...
b.sum() = 57.600002
a.grad.sum() =  1.75
v.grad.sum() =  0.0
```
The essence of this problem is that the method of directly rewriting variables in static graphs violates the static single assignment form (SSA).

#### What this PR changed

- In Dynamic Mode: 
The behavior of `__setitem__` is unchanged. Users can still write code like: `x[0] = 10`.

- In Static Mode
**`__setitem__` is disabled**.  Users should use `x = paddle.static.setitem(x, 0, 10)` to replace `x[0] = 10`. 
API `paddle.static.setitem` will provide the variable assignment with indexing in static mode, but unlike `__setitem__`, it will return a **new variable** instead of overwriting the original variable.
```python
>>> paddle.enable_static()
>>> a = paddle.ones((2,3,4))
>>> a[0] = 10
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/dist-packages/paddle/fluid/framework.py", line 2303, in __setitem__
    "In static mode, the __setitem__ (looks like: x[indices] = values) should not be used. Please use x = paddle.static.setitem(x, indices, values)"
RuntimeError: In static mode, the __setitem__ (looks like: x[indices] = values) should not be used. Please use x = paddle.static.setitem(x, indices, values)
```

- In Dynamic-to-Static (with `@paddle.jit.to_static` )
As  https://github.com/PaddlePaddle/Paddle/pull/55282 ,  introducing a `ParametersMap`, if there are a new variable mapped to an old variable, the old one will be replaced by the new one.
Users are not aware of this change in d2s, with just still using `paddle.jit.to_static`.


python/paddle/jit/dy2static/slice_transformer.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __init__(self, obj):
+        self.obj = obj
+
+    def __getitem__(self, item):
+        return self.obj[item]
+
+    def __setitem__(self, item, value):
+        if isinstance(self.obj, Variable):
+            return paddle.fluid.framework._setitem_impl_(self.obj, item, value)
```
Aurelius84:
```suggestion
             new_var =  paddle.fluid.framework._setitem_impl_(self.obj, item, value)
            # NOTE(dev): Update __dict__ will not modify the id(), but only move the
            # pointed reference object to the new one.
            self.obj.__dict__.update(new_var.__dict__)
```

Aurelius84 (commented):
LGTM

python/paddle/incubate/optimizer/functional/lbfgs.py
```diff
@@ -178,16 +178,20 @@ def body(
...
         )
 
-        def cond(i, q):
+        def cond(i, q, ai_vec):
             return i != tail
 
-        def body(i, q):
-            ai_vec[i] = rhok_vec[i] * paddle.dot(sk_vec[i], q)
+        def body(i, q, ai_vec):
+            ai_vec = paddle.static.setitem(
```
jeff41404: The API for ·paddle.static· is only used in static graphs, so it is recommended to add a check to determine whether it is static or dynamic grahp. Only static graphs can call this line of code.

python/paddle/tensor/math.py
```diff
@@ -5717,10 +5717,18 @@ def vander(x, n=None, increasing=False, name=None):
     res = paddle.empty([x.shape[0], n], dtype=x.dtype)
 
     if n > 0:
-        res[:, 0] = paddle.to_tensor([1], dtype=x.dtype)
+        res = paddle.static.setitem(
```
jeff41404: Same issue as above

python/paddle/tensor/math.py
```diff
@@ -5717,10 +5717,18 @@ def vander(x, n=None, increasing=False, name=None):
...
+            res, (slice(None), 0), paddle.to_tensor([1], dtype=x.dtype)
+        )
     if n > 1:
-        res[:, 1:] = x[:, None]
-        res[:, 1:] = paddle.cumprod(res[:, 1:], dim=-1)
+        res = paddle.static.setitem(
+            res, (slice(None), slice(1, None)), x[:, None]
+        )
+        res = paddle.static.setitem(
```
jeff41404: Same issue as above

python/paddle/vision/transforms/functional_tensor.py
```diff
@@ -236,7 +236,8 @@ def _affine_grid(theta, w, h, ow, oh):
         y_grid = paddle.linspace(-oh * 0.5 + d, oh * 0.5 + d - 1, oh).unsqueeze(
             -1
         )
-        base_grid[..., 1] = y_grid
+        base_grid = paddle.static.setitem(base_grid, (..., 0), x_grid)
+        base_grid = paddle.static.setitem(base_grid, (..., 1), y_grid)
```
jeff41404, line 240: Same issue as above

python/paddle/vision/transforms/functional_tensor.py
```diff
@@ -409,17 +410,23 @@ def rotate(
...
+        matrix,
+        2,
+        matrix[2]
+        + matrix[0] * (-rotn_center[0] - post_trans[0])
+        + matrix[1] * (-rotn_center[1] - post_trans[1])
+        + rotn_center[0],
     )
-    matrix[5] += matrix[3] * (-rotn_center[0] - post_trans[0]) + matrix[4] * (
-        -rotn_center[1] - post_trans[1]
+    matrix = paddle.static.setitem(
```
jeff41404: Same issue as above

python/paddle/vision/transforms/functional_tensor.py
```diff
@@ -621,7 +628,7 @@ def erase(img, i, j, h, w, v, inplace=False):
     if not inplace:
         img = img.clone()
 
-    img[..., i : i + h, j : j + w] = v
+    img = paddle.static.setitem(img, (..., slice(i, i + h), slice(j, j + w)), v)
```
jeff41404: Same issue as above

python/paddle/incubate/optimizer/functional/lbfgs.py
zoooo0820: Done, thanks.

jeff41404 (approved):
LGTM

XiaoguangHu01 (approved):
LGTM

PR #53691, opened 2023-05-10
up index warning level

zoooo0820:
### PR types

Others
### PR changes

Others
### Description

Pcard-66985
For compatibility，case output 0D of API `__getitem__` is hacked to 1D and there will be a warning.

If the model code has this case, there will be too many warnings since number of iteration is large. (ref https://rpg.ifi.uzh.ch/docs/glog.html , the default GLOG_v=0 means INFO. )

This PR changes level of these two warnings to WARNING, displayed when GLOG_V=1.


jeff41404 (approved):
LGTM

PR #53698, opened 2023-05-10
[cherry-pick] revise 'Examples' of LBFGS to create right docs(cn)

lijialin03:
### PR types

Others
### PR changes

Docs
### Description

Pcard-66961
revise 'Examples' of public functions of LBFGS to create right docs(cn) and add a new method for right docs(en)


XiaoguangHu01 (approved):
LGTM

PR #53729, opened 2023-05-11
[inference zero dim] softmax, stack op trt converter support zero dim

yuanlehome:
### PR types

Others
### PR changes

Others
### Description

softmax, stack op trt converter support zero dim.


zhangjun (approved):
LGTM

PR #53749, opened 2023-05-12
fix add_n kernel of large shape

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Description

Pcard-64703

fix add_n kernel when the size of elements is near to INT_MAX

For example, the following code raise `CUDA ERROR 700`.

```
import paddle


a1 = paddle.rand([12240, 99896])
a2 = paddle.rand([12240, 99896])
a3 = paddle.rand([12240, 99896])

b = paddle.add_n([a1, a2, a3])
```
image


From00 (approved):
LGTM

PR #53864, opened 2023-05-16
remove user define grad

GGBond8488:
### PR types
Others

### PR changes
Others 

### Description
Pcard-70652
remove user-define x_grad


test/fft/test_spectral_op.py
```diff
@@ -167,10 +167,8 @@ def test_check_output(self):
 
     def test_check_grad(self):
         self.check_grad(
-            "X",
+            ["X"],
             "Out",
-            user_defined_grads=[self.x_grad],
-            user_defined_grad_outputs=[self.out_grad],
```
jeff41404, line 173: shall we delete `self.x_grad` and `self.out_grad` above together?

GGBond8488: It can be deleted now, and it will support random out_grad later and add  self.out_grad back

jeff41404 (approved):
LGTM

PR #53935, opened 2023-05-18
[Semi-Auto] Define unified data class for inferencing distributed attributes in dygraph and static mode

pkuzyc:
### PR types
New features

### PR changes
Others

### Description
Pcard-70448
Define a unified data class DistTensorSpec and its wrap functions for inferencing distributed attributes in dygraph and static mode. It contains the shape and distributed attribute of its corresponding tensor.


python/paddle/distributed/auto_parallel/static/utils.py
```diff
@@ -2373,3 +2373,66 @@ def __impl__(*args, **kwargs):
...
+    Returns:
+      input_specs: list, DistTensorSpec for each input tensor of the dist_op
+      output_specs: list, DistTensorSpec for each output tensor of the dist_op
+      attrs: dict, attribute map of the dist op
+
+    Usage:
+      op_desc = dist_op.serial_op.desc
+      input_name_list = []
+      output_name_list = []
+      input_name_list.append(op_desc.input('X')[0]) # 'X' is the arg name for op
```
JZ-LIANG, line 2401: it would be better (in future) that the wrap could take op as the only argument and user not need to border the input_name_list/output_name_list construction.

in order to achieve that wrap need to maintain the order of op argument slot from Phi API.

python/paddle/distributed/auto_parallel/static/utils.py
```diff
@@ -2373,3 +2373,66 @@ def __impl__(*args, **kwargs):
...
+    attrs = {}
+
+    serial_op = dist_op.serial_op
+
+    # Construct each input tensor's DistTensorSpec with shape and dist_attr
+    for name in input_names:
+        tensor_dist_attr = dist_op.dist_attr.get_input_dist_attr(name)
+        var = serial_op.block._var_recursive(name)
+        tensor_shape = var.shape
+        dist_spec = DistTensorSpec(tensor_shape, tensor_dist_attr)
```
JZ-LIANG, line 2422: in static mode, the dist attr of op is the destination dist attr.
here should use the source dist attr which is hold by tensor.

python/paddle/fluid/tests/unittests/test_auto_parallel_dist_tensor_spec.py
```diff
@@ -0,0 +1,98 @@
+# Copyright (c) 2023 PaddlePaddle Authors. All Rights Reserved.
```
JZ-LIANG: wrong path for this file, should be: test/auto_parallel/

python/paddle/distributed/auto_parallel/static/utils.py
pkuzyc: The names of op arguments are inconsistent in static mode and Phi API, so it's too hard to obtain the arguments from op in wrap function.

python/paddle/distributed/auto_parallel/static/utils.py
pkuzyc: Done

python/paddle/fluid/tests/unittests/test_auto_parallel_dist_tensor_spec.py
pkuzyc: Done

PR #53987, opened 2023-05-19
fix device changed in setitem-numpy case

zoooo0820:
### PR types

Bug fixes
### PR changes

Others
### Description

Pcard-66985
Now there will be an unexpected device change in combined-indexing of `__setitem__`
```python
>>> import paddle
>>> paddle.set_device('cpu')
>>> a.place
Place(cpu)
>>> a[[0,1],1] = 10
>>> a.place
Place(gpu:0)
```
This may cause some downstream task error, like https://github.com/PaddlePaddle/Paddle/issues/53829


wanghuancoder (approved):

jeff41404 (approved):
LGTM

PR #54032, opened 2023-05-22
fix inference trt fp16 io

yuanlehome:
### PR types

Others
### PR changes

Others
### Description

fix inference trt fp16 io.


XieYunshen (approved):
LGTM

PR #54122, opened 2023-05-26
fix delete_repeated_ops_pass

zhupengyang:
### PR types

Bug fixes

### PR changes

Others

### Description

- Not delete slice op if out op has shared var node


hong19860320 (approved):
LGTM

PR #54143, opened 2023-05-26
[IR] add op name normalizer

kangguangli:
### PR types

New features
### PR changes

Others
### Description

This PR unifies the legacy op names into normalized op names, which are recorded in `op_compact.yaml`. 
As the ops not recorded in that file, like `fetch_v2/fetch`、`share_buffer`, we will use an individual yaml to record such relationships later.

#### Others
Pcard-67164


winter-wang (approved):
LGTM

PR #54150, opened 2023-05-27
[XPU] fix the dataloader problem in RDMA env

XiaociZhang:
### PR types

Bug fixes

### PR changes

Others

### Description

When running multi-machine training with Paddle DataLoader, an unexpected segmentfault will be raised in dataloader process, where the traceback goes all back to a runtime error that dataloader workers exit unexpectedly. Similar problems have been discussed that lead to a misbehavior of OpenCV working in multiprocessing environment.
See
https://stackoverflow.com/questions/54013846/pytorch-dataloader-stucked-if-using-opencv-resize-method


XiaociZhang:
[Unresolved] whether gpu has a similar problem. 王贤明 recalled that gpu also have this issue, but no conclusive evidents were ever-present.

QingshuChen (commented):
LGTM

houj04 (approved):

kuizhiqing (approved):
lgtm

PR #54183, opened 2023-05-29
softmax fwd: force vec size to 1 when dtype is float

shaojiewang:
### PR types

Others
### PR changes

OPs
### Description

Pcard-70458
force softmax fwd kernel's vec size to 1 when dtype is float.


Xreki (approved):
LGTM.

image

上述softmax_0的性能下降问题将在https://github.com/PaddlePaddle/Paddle/pull/40618 中修复。

PR #54273, opened 2023-06-01
support py3.11 compile, but unittest is disabled.

2742195759:
### PR types

Others
### PR changes

Others
### Description

PCard-66972

support py3.11 compile, but unittest is disabled.
TODO: fix the test_eval_frame.py


zhangbo9674 (approved):
LGTM for print

PR #54336, opened 2023-06-04
Add data_format option for ResNet models in the resnet.py file.

tangleintel:
### PR types

Performance optimization

### PR changes

Others
### Description

ResNet models were broadly used as a AI benchmarking models for various kinds of hardware backend vendors. Sometimes channels last data format could get better performance on ResNet based models. We put this option in the native vision model zoo to make users have more performance related option or choice to get better performance for their benchmarking models.

### DIFF
--- a/python/paddle/vision/models/resnet.py
+++ b/python/paddle/vision/models/resnet.py
@@ -140,14 +140,15 @@ class BottleneckBlock(nn.Layer):
         base_width=64,
         dilation=1,
         norm_layer=None,
+        data_format='NCHW',
     ):
         super().__init__()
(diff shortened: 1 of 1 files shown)


tangleintel:
@KevinYuk @KimBioInfoStudio

KimBioInfoStudio:
@luotao1 we would like add channel last to vision ResNet series models, could u assign the right people helping on review process? Thx in advanced, and seems the failed CI is on infer not the pr itself

tangleintel:
Got another solution(using PaddleClas) to our case.

PR #54431, opened 2023-06-07
eager call all2all to avoid p2p hang in lazy init

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Description

Pcard-70448

eager call all2all to avoid p2p hang in lazy init

When use interceptor to train pipeline parallel, for example, pp0 send, and pp1 recv, the `recv` pp1 in  is not launched before until `send` ends in pp0. (Feature of interceptor's actor model, it needs `signal` to triger, and the `signal` is sended by brpc after pp0 ends.) 

However, in new version of nccl like 2.15+, Point-to-point connections for send/receive are created upon the first exchange between two ranks, and the creation need the peers to be ready. 

So, it hangs.

This pr adds all2all comunication when process group is created eagerly to avoid the hang.

### DIFF
--- a/python/paddle/distributed/auto_parallel/static/process_group.py
+++ b/python/paddle/distributed/auto_parallel/static/process_group.py
@@ -74,18 +74,8 @@ def new_process_group(
     if group_id is None:
         group_id = _new_ring_id() + num_groups + 1
 
-    new_pg = ProcessGroup(group_id, ranks)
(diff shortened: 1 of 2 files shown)


zhaoyinglia (approved):
LTGM

From00 (approved):

PR #54452, opened 2023-06-08
[IR] Adapt startup program

kangguangli:
### PR types

Others
### PR changes

Others
### Description

This PR modifies the logic of translate program to adapt the `startup program`. The startup program is used to initialize the parameters of  program. We need to insert a corresponding `SetParameterOp` for every parameter initialized in startup program.
Currently, for every parameter in `ProgramDesc`, we will
1. insert a `GetParameterOp` if it occurs in inputs of a op first and only first time
2. insert a `SetParameterOp` if it occurs in outputs of a op and only last time

#### Others
Pcard-67164

### DIFF
--- a/test/cpp/ir/core/program_translator_test.cc
+++ b/test/cpp/ir/core/program_translator_test.cc
@@ -76,7 +76,7 @@ TEST(PaddleDialectTest, StartupProgram) {
   size_t op_size = program->block()->size();
   // ops.size() = op size in BlockDesc + get_parameter_op +
   // consant_op_for_uniform
-  // + consant_op_for_uniform for guassian
+  // + consant_op for guassian
   EXPECT_EQ(op_size, p.Block(0).OpSize() + program->parameters_num() + 3 + 53);
 
   program->Print(std::cout);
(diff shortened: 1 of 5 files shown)


test/cpp/ir/core/program_translator_test.cc
```diff
@@ -63,3 +63,21 @@ TEST(PaddleDialectTest, Translator) {
...
+  ctx->GetOrRegisterDialect<ir::BuiltinDialect>();
+  auto program = paddle::TranslateLegacyProgramToProgram(p);
+
+  size_t op_size = program->block()->size();
+  // ops.size() = op size in BlockDesc + get_parameter_op +
+  // consant_op_for_uniform
+  // + consant_op for guassian
+  EXPECT_EQ(op_size, p.Block(0).OpSize() + program->parameters_num() + 3 + 53);
+
+  program->Print(std::cout);
```
zhangbo9674, line 82: Suggest delete cout.

zhangbo9674:
```suggestion

```

zhangbo9674 (approved):

winter-wang (approved):
LGTM

test/cpp/ir/core/program_translator_test.cc
kangguangli: 好的，之后单独提PR修改。

kangguangli: Done in #54499

PR #54513, opened 2023-06-09
Fix the timeout bug of some communication api on A100

pkuzyc:
### PR types
Bug fixes

### PR changes
Others

### Description
Pcard-67009
Fix the timeout bug of the following unit tests: test_collective_reduce_scatter, test_collective_broadcast_api, test_collective_alltoall_api on A100. This PR solves the bug by setting a larger timeout value when running on A100. The bug causes by: 1) The run time of these tests have fluctuation, sometimes it exceeds the timeout limitation.  2) On A100, each of these unit test contains one additional test case, so it takes longer run time.


XieYunshen (approved):
LGTM

PR #54624, opened 2023-06-13
[BugFix] fix bug of release output in pp

ForFishes:
### PR types

Bug fixes
### PR changes

Others
### Description

[BugFix] fix bug of release output in pp when using inplace-op
PCard-70444


gongweibao (approved):
LGTM

PR #54658, opened 2023-06-14
[Cherry-Pick] Modify the bf16 accuracy checking framework in OpTest

Vvsmile:
### PR types

Others
### PR changes

OPs
### Description

- Modify the accuracy checking framework of bf16 datatype in OpTest, including both of forward and backward. 
- The forward result of bf16 is compared with the result of fp32. 
- The backward result of bf16 is compared with the result of fp32. 
- This modification does not need any fix in unit test file. 
- This accuracy checking mechanism is same as it of fp16 now, and this mechanism uses the bf16 or fp16 datatype as the input's datatype, converts to fp32 in the OpTest framework, gets the computation results of each datatype and compares them with each other. Hence, it removes the error caused by rounding from high-precision to low-precision, which is the primary cause resulting in the huge accuracy error in bf16 computation checking in some cases, such as cumsum.
card-72073


XiaoguangHu01 (approved):
LGTM

PR #54666, opened 2023-06-14
support composite higher grad maker auto gen

Charles-hit:
### PR types

New features
### PR changes

Others
### Description

Pcard-66975
support composite higher grad maker auto gen.


heavyrain-lzy (approved):
LGTM

PR #54679, opened 2023-06-15
[inference][trt]Layer norm rollback 2 plugin when trt<8.6

bukejiyu:
### PR types
Bug fixes

### PR changes
Others 

### Description]
Pcard-71501
Because pcpvt_base run fp16 accuracy degradation，rollback Layer norm  to plugin when trt<8.6


zhangjun (approved):
LGTM

PR #54693, opened 2023-06-15
fix paddle.nonzero when as_tuple=True

zoooo0820:
### PR types

Bug fixes
### PR changes

APIs
### Description

Pcard-64703

`paddle.nonzero(x, as_tuple=True)`  should return a tuple with 1-D Tensor. However, currently it returns a tuple with 2-D Tensor, shape of which is [N, 1]. 

`paddle.where(condition)` is same with `paddle.nonzero(condition, as_tuple=True)`, which is fixed,  too.


zhwesky2010 (approved):
LGTM for paddle bug-fix

PR #54810, opened 2023-06-21
[Semi-Auto] Replicated Parallel Rule

JZ-LIANG:
### PR types
 New features

### PR changes
Others

### Description

Pcard-70448

A bottom line rule for ops that has its own rule.


zhiqiu (approved):
LGTM

PR #54982, opened 2023-06-29
[CINN] comb the op lowering code

BiynXu:
### PR types
Others

### PR changes
Others

### Description
Pcard-72511
This pr comb the code of op lowering.
Organize the generalized Lower process into four stages: 
  1. Call Compute and perform narrow Lower (i.e. translate graph to CINN IR).
  2. Do Op level schedule.
  3. Do group level schedule.
  4. Post process, including processing of LoweredFunc arguments and temporary variables, application of low level pass, etc

Then, all types of group lower processes call these processes, abstracting only a small amount of different logic and using different callback functions for processing, to achieve the maximum uniformity of the generalized Lower process.
Specifically, due to significant differences in the Lower logic of CustomCall, it needs to be handled separately by LowerCustomCall().


paddle/cinn/auto_schedule/tests/performance_comparison_test.cc
```diff
@@ -157,7 +157,7 @@ class PerformanceTester : public ::testing::Test {
 
     for (auto group : graph->fusion_groups) {
       compile_options.lowered_funcs.push_back(
-          op_lowerer->LowerWithoutSchedule(group));
+          op_lowerer->Lower(group, false, false));
```
zhhsplendid: A tip in software engineering: boolean variables of functions have very bad readability. `Lower(group, false, false)`, what stands for the `false`? It happens when you have lots of `true` or `false`.

Usually, there are two ways to handle this problem.

1. Don't use boolean function
2. Add comment :`Lower(group, /*apply_op_schedule = */ false,  /*apply_group_schedule=*/ false)`
3. Use some parameters, such as:
```
    bool apply_op_schedule = false;
    bool apply_group_schedule = false;
    Lower(group, apply_op_schedule, apply_group_schedule);
```

paddle/cinn/hlir/framework/op_lowering.cc
```diff
@@ -70,532 +81,332 @@ std::vector<ir::LoweredFunc> OpLowerer::Lower(GroupPtr& group) {  // NOLINT
...
-    }
-  } else {
-    LOG(FATAL) << "Previous IR Schedule Is Not Implemented!";
+bool OpLowerer::ElementwiseScheduleDetermineFunction(Node* node) {
+  return true;
+}
+
+bool OpLowerer::ReduceScheduleDetermineFunction(Node* node) {
+  auto& op_pattern_dict = Operator::GetAttrs<OpPatternKind>("OpPattern");
+  if (op_pattern_dict[node->op()] == framework::kReduction) {
```
zhhsplendid: Rewrite it:
`return op_pattern_dict[node->op()] == framework::kReduction`;

paddle/cinn/hlir/framework/op_lowering_util.h
```diff
@@ -31,8 +31,8 @@ ir::Tensor GetTensor(
 
 std::vector<ir::Tensor> CollectInputTensor(
     const Node* node,
-    std::vector<ir::Tensor>& func_args,                       // NOLINT
-    std::unordered_map<std::string, ir::Tensor>& tensor_map,  // NOLINT
+    std::vector<ir::Tensor>* func_args,
```
zhhsplendid: Usually we put input parameters before output parameters.

This problem also happens on some above functions, please try your best to change it.

Reference:
https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

paddle/cinn/hlir/framework/op_lowering.cc
BiynXu: Done.

paddle/cinn/auto_schedule/tests/performance_comparison_test.cc
BiynXu: The call of the Lower() function has been modified using the second method.

paddle/cinn/hlir/framework/op_lowering_util.h
BiynXu: All new or modified parts of this PR have been modified, but some historical codes remain unchanged.

JamesLim-sy (approved):
LGTM

PR #54983, opened 2023-06-29
Cinn schedule error

ZzSean:
### PR types

Others
### PR changes

Others
### Description

**新增`Schedule`报错处理机制**

card-72613

内容：
- 新增异常处理数据结构。不同`schedule primitive`通过继承基类`IRScheduleErrorHandler`，完成不同的报错信息打印
- 分层级打印报错信息。通过向`ScheduleImpl`中新增`ScheduleErrorMessageLevel`，决定后续出现错误时的报错打印层级，可通过运行时的FLAGS控制
- 使用宏简化代码
- 该PR以`Split`为例，展示如何在代码中使用新的报错处理机制
- 添加异常单测
- 自定义`CINN_THROW`，与主框架的`PADDLE_THROW`报错信息内容保持一致

效果：
- 之前schedule出错会直接core dumped，可运行schedule单测复现
- 添加新的异常处理机制后，会抛出异常但不会core dumped，且可以控制报错信息打印层级和报错信息
- 报错信息如下：

```
0 cinn::ir::IRSchedule::Split(cinn::ir::Expr const&, std::vector<int, std::allocator<int> > const&)
1 cinn::ir::IRSchedule::Split(cinn::ir::Expr const&, std::vector<cinn::ir::Expr, std::allocator<cinn::ir::Expr> > const&)
2 cinn::ir::enforce::EnforceNotMet::EnforceNotMet(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, char const*, int)

----------------------
Error Message Summary:
----------------------
[IRScheduleError] An error occurred in the scheduel primitive <split>. 
[Error info] The params in factors of Split should not be less than -1 or have more than one -1!
(at /root/paddlejob/workspace/work/zhangzheng/Paddle/paddle/cinn/ir/ir_schedule.cc : 179)
```


zhangbo9674 (dismissed):
LGTM for std::cout

paddle/cinn/ir/ir_schedule_error.h
```diff
@@ -0,0 +1,173 @@
...
+  sout << "Not support stack backtrace yet.\n";
+#endif
+  return sout.str();
+}
+
+static std::string GetTraceBackString(const std::string& what,
+                                      const char* file,
+                                      int line) {
+  return GetCurrentTraceBackString() + GetErrorSumaryString(what, file, line);
+  // return GetErrorSumaryString(what, file, line);
```
zhhsplendid: Should we remove useless code line?

paddle/cinn/ir/ir_schedule_error.h
```diff
@@ -0,0 +1,173 @@
...
+  const char* what() const noexcept override { return err_str_.c_str(); }
+
+ private:
+  std::string err_str_;
+};
+
+#define CINN_THROW(...)                                              \
+  do {                                                               \
+    try {                                                            \
+      throw enforce::EnforceNotMet(__VA_ARGS__, __FILE__, __LINE__); \
```
zhhsplendid, line 120: Is this same as `Paddle` enforce? As I remember, we discussed offline that we use `PADDLE_THROW` when we found the definition, otherwise we define by ourself, then we can handle it with both CINN-only and Paddle-CINN.

Does this implementation also handle both CINN-only and Paddle-CINN? Do we reuse Paddle code?

paddle/cinn/ir/ir_schedule_error.h
```diff
@@ -0,0 +1,173 @@
...
+      std::cout << e.what() << std::endl;                            \
+      throw;                                                         \
+    }                                                                \
+  } while (0)
+}  // namespace enforce
+
+/**
+ *  \brief Indicates the level of printing error message in the current Schedule
+ */
+enum class ScheduleErrorMessageLevel : int32_t {
```
zhhsplendid: Can the ErrorMessage use at other places? If so, should we remove `Schedule` in naming? Then we can use  the code you implemented all over the CINN!

paddle/cinn/ir/ir_schedule_error.h
```diff
@@ -0,0 +1,173 @@
...
+  /** \brief  Print an error message in short mode*/
```
zhhsplendid: Comment should add "short mode shows which and where the schedule error happens"

"detailed mode shows which and where the schedule error happens, and the schedule input parameters"

paddle/cinn/runtime/flags.cc
```diff
@@ -160,6 +160,11 @@ DEFINE_int32(cinn_profiler_state,
              "Specify the ProfilerState by Int in CINN, 0 for kDisabled, 1 for "
              "kCPU, 2 for kCUDA, 3 for kAll, default 0.");
 
+DEFINE_int32(cinn_schedule_error_message_level,
```
zhhsplendid, line 163: Same as my other comment, can the ErrorMessage be use at other places? If so, should we remove "schedule" text in the flag? Then we can use the code you implemented all over the CINN!

paddle/cinn/ir/ir_schedule_util.cc
```diff
@@ -220,35 +221,115 @@ void ReplaceExpr(Expr* source,
...
+
+   private:
+    ModuleExpr module_expr_;
+    int64_t factor_;
+    size_t idx_;
+  };
+
+  class InferFactorErrorHandler : public IRScheduleErrorHandler {
+   public:
+    explicit InferFactorErrorHandler(const ModuleExpr& module_expr)
```
zhhsplendid: Will it be complex if we have to implement sub-classes of `IRScheduleErrorHandler` at where we want to report error?

I thought we have a `ErrorHandler` class which can set general error message and detailed message, then we just input strings at where we want to report error.

paddle/cinn/ir/ir_schedule_error.h
ZzSean: done,thx

paddle/cinn/ir/ir_schedule_error.h
ZzSean: `PADDLE_THROW` is more complicated and has many redundant functions. If we want to reuse Paddle code we have to include more Paddle header files, thus we cannot build CINN-ONLY.
Using `CINN_THROW` defined by ourselves maybe a good solution for now, 'cause there is `IR_THROW` in `PADDLE_IR` to deal with the similar problem.

paddle/cinn/ir/ir_schedule_error.h
ZzSean: done, thx

paddle/cinn/ir/ir_schedule_util.cc
ZzSean: Two reasons:
- Different kinds of schedule primitive require different error messages, so using abstract class and sub-classes is to deal with their own special need for this.
- Implementation of error message inside each subclass is easier to use the macro to simplify the code.

paddle/cinn/ir/ir_schedule_error.h
ZzSean: Technically no problem. But  its namespace is `cinn::ir` now, i don't know if it is convenient/necessary to use this in the same level of `ir`, like `hlir`, `optim`, etc.

ZzSean: What about changing the name and the location of this file when we have the need later?

paddle/cinn/ir/ir_schedule_error.h
zhhsplendid: OK, not best but fine to me now.

paddle/cinn/ir/ir_schedule_error.h
zhhsplendid: If so, can we change the namespace, location of file in this PR? We would like to reuse for other places.

zhangbo9674 (approved):
LGTM for cout

zhhsplendid:
Discussed offline, we approve this PR now, but above unresolved comments will be changed in the future PRs.

zhhsplendid (approved):
LGTM

PR #55077, opened 2023-07-03
Fix random failure of test_conv3d_op

Tom-Zheng:
### PR types

Bug fixes
### PR changes

Others
### Description

`test_conv3d_op` randomly fails because that Set traverse is in random order. If the order is not matched with the provided `user_defined_grads`, this UT will fail. This PR solves this problem by using List instead of Set.


Xreki (approved):
LGTM

PR #55117, opened 2023-07-04
[NewIR] enable new ir for unittests in white list

kangguangli:
### PR types

New features
### PR changes

Others
### Description

This PR is used to enable new ir for unittests in white list. If someone want to add a unittest, just add it in `test/new_ir_op_test_white_list`.

#### Details
If an op need to be executed in new ir, two flags needs to be set:
1. `FLAGS_NEW_IR_OPTEST`, now set in `paddle/scripts/paddle_build.sh`
2. `FLAGS_NEW_IR_OPTEST_WHITE_LIST`, now set in `test/legacy_test/CMakeLists.txt`

#### Future Works
Due to the constraints of cmake, we can only set the test property in the scope it defined. So if we want to add the test not in `test/legacy_test/`, we need to modify the corresponding `CMakeLists`.

#### Others
Pcard-67164


test/legacy_test/eager_op_test.py
```diff
@@ -1125,6 +1127,41 @@ def _calc_dygraph_output(
             )
             return outputs
 
+    def _check_ir_output(self, place, program, feed_map, fetch_list, outs):
+        if os.getenv("FLAGS_NEW_IR_OPTEST") is None:
```
Charles-hit, line 1131: 这儿是不是可以参考一下以前optest的白名单机制

kangguangli: 这里需求不太一样，之前OpTest的白名单机制不支持单测粒度的白名单

risemeup1 (approved):
LGTM

XieYunshen (approved):

Aurelius84 (approved):
LGTM

Charles-hit (approved):

PR #55131, opened 2023-07-04
[NewIR] fix vector int attribute emplace bug

phlrain:
### PR types
Bug fixes

### PR changes
Others

### Description

修复vector<int> attribute 忘记 emplace的bug

### Others
Pcard-67164


Aurelius84 (approved):
LGTM

PR #55145, opened 2023-07-05
[CINN] Re-Implement operator = for two Expr Tree

Courtesy-Xs:
### PR types

Performance optimization
### PR changes

Others
### Description

Former implementation is to transform two AST tree as two string and compare their equivalence and it is cost-time, now we compare them by doing dfs traversal on both two tree。
Compared two implementation, the latter can get approximately 10% performance profit on batchnorm operator on function complieGraph in CINN。

compilerGraph cost time 
|  dev  | optim | ratio |
|  ----  | ----  | ---- |
| 1.48086  | 1.34312 | 10.25% |

Pcard-72718


JamesLim-sy (approved):
LGTM

PR #55171, opened 2023-07-05
【CINN】refactor ir_visitor

Courtesy-Xs:
### PR types

Performance optimization
### PR changes

Others
### Description

This PR delete middle ir_visitor class and thus we can avoid middle virtual function call and codes look more clean 
pcard-72718


paddle/cinn/optim/ir_copy.cc
```diff
@@ -419,19 +421,19 @@ struct IRCopyVisitor : public ir::IRVisitorBase<Expr> {
...
   NODETY_BINARY_OP_FOR_EACH(OP_BINARY_HANDLE)
 #undef OP_BINARY_HANDLE
 
-#define OP_UNARY_HANDLE(op__)                \
-  Expr Visit(const op__* op) override {      \
-    auto v = IRVisitorBase::Visit(&op->v()); \
-    return op__::Make(v);                    \
+#define OP_UNARY_HANDLE(op__)                                \
+  Expr Visit(const op__* op) override {                      \
+    auto v = IRVisitorRequireReImplVisitor::Visit(&op->v()); \
```
zhhsplendid: Why are 2 Visitors in name?

Using `IrVisitorRequireReImpl`or `IrRequireReImplVisitor`?

paddle/cinn/pybind/ir.cc
zhhsplendid, line 1: Why is there an empty file? Should we delete it?

paddle/cinn/optim/ir_copy.cc
Courtesy-Xs: done

paddle/cinn/pybind/ir.cc
Courtesy-Xs: done

zhhsplendid (approved):
LGTM

PR #55200, opened 2023-07-06
[IR] Fix optional output bug in op verify

zhangbo9674:
### PR types

Bug fixes
### PR changes

Others
### Description

Fix optional output bug in op verify

### Other
Pcard-67164


winter-wang (approved):
LGTM

PR #55278, opened 2023-07-10
Fix new ir polygama op bug

phlrain:
### PR types
Bug fixes

### PR changes
Others

### Description
修复polygama、i0 i0e i1 i1e，nextafter 的op translator错误

### Others
Pcard-67164


heavyrain-lzy (approved):
LGTM

PR #55297, opened 2023-07-10
DOCS: Adding imformation about datatype in math.py

### LINKED ISSUE #55252: DOC issue in paddle math.py logit
### 文档链接&描述 Document Links & Description

https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/logit_en.html#logit

the documentation says that logit take data type float32 and float 64
doc

but when I take a look at source of the logit in paddle I saw it is allowed  float16, float32 and float64

sorepaddle

I thinks it is a issue and I want to contribute

### 请提出你的建议 Please give your suggestion

I will submit the PR if this issue is correct.

MuhammadNizamani:
### PR types
 Others 

### PR changes
 Docs 

### Description
This PR add more information about  logit function in math.py 
Closed #55252

### DIFF
--- a/python/paddle/tensor/math.py
+++ b/python/paddle/tensor/math.py
@@ -4464,7 +4464,7 @@ def logit(x, eps=None, name=None):
             \end{array}\right.
 
     Args:
-        x (Tensor): The input Tensor with data type float32, float64.
+        x (Tensor): The input Tensor with data type uint16, float16, float32, float64.
         eps (float, optional):  the epsilon for input clamp bound. Default is None.
         name (str, optional): Name for the operation (optional, default is None).
             For more information, please refer to :ref:`api_guide_Name`.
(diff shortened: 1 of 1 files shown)


python/paddle/tensor/math.py
```diff
@@ -4464,7 +4464,7 @@ def logit(x, eps=None, name=None):
             \end{array}\right.
 
     Args:
-        x (Tensor): The input Tensor with data type float32, float64.
+        x (Tensor): The input Tensor with data type uint16, float16, float32, float64.
```
zhangting2020: In fact, this API supports bfloat16 instead of uint16 data type. The reasons are as follows: 

https://github.com/PaddlePaddle/Paddle/blob/b4d7e1e06a2ce03ff38299bb8b8ebf959ee4bc3d/python/paddle/fluid/data_feeder.py#L115-L119

MuhammadNizamani: should I change it into bfloat16?

zhangting2020: Yes, just replace uint16 with bfloat16.

zhangting2020 (approved):

PR #55379, opened 2023-07-12
[0D-Tensor] CINN supports transpose, add special case to expand_zero_dim_pass

jiahy0825:
### PR types
Function optimization

### PR changes
OPs

### Description

Pcard-71127

[0D-Tensor] CINN supports transpose, add a special case to expand_zero_dim_pass

- Link to the first PR: #53382


thisjiang (approved):
LGTM

PR #55397, opened 2023-07-13
[Semi-Auto] add split spmd rule

pkuzyc:
### PR types
New features

### PR changes
Others

### Description

Pcard-70448

Add split spmd rule for auto parallel. It infers the output dims mapping as following: the splitted axis cannot be sharded, set the dims mapping of splitted axis in input and outputs to -1. For other axes in output, set their dims mapping equal to input's one.

### DIFF
--- a/paddle/fluid/distributed/auto_parallel/spmd_rules/rules.h
+++ b/paddle/fluid/distributed/auto_parallel/spmd_rules/rules.h
@@ -153,6 +153,7 @@ REGISTER_SPMD_RULE(softmax_with_cross_entropy, CrossEntropyWithSoftmaxSPMDRule);
 
 // split rule
 REGISTER_SPMD_RULE(split, SplitSPMDRule);
+REGISTER_SPMD_RULE(split_with_num, SplitSPMDRule);
 
 }  // namespace auto_parallel
 }  // namespace distributed
--- a/paddle/fluid/distributed/auto_parallel/spmd_rules/rules.h
+++ b/paddle/fluid/distributed/auto_parallel/spmd_rules/rules.h
@@ -23,6 +23,7 @@
 #include "paddle/fluid/distributed/auto_parallel/spmd_rules/reduction_spmd_rule.h"
 #include "paddle/fluid/distributed/auto_parallel/spmd_rules/replicated_spmd_rule.h"
 #include "paddle/fluid/distributed/auto_parallel/spmd_rules/softmax_spmd_rule.h"
+#include "paddle/fluid/distributed/auto_parallel/spmd_rules/split_spmd_rule.h"
 
 // TODO(ljz) Automatic this process in cmake file.
 namespace paddle {
@@ -149,6 +150,9 @@ REGISTER_SPMD_RULE(log_softmax, SoftmaxSPMDRule);
 // cross_entropy_with_softmax
 REGISTER_SPMD_RULE(cross_entropy_with_softmax, CrossEntropyWithSoftmaxSPMDRule);
 REGISTER_SPMD_RULE(softmax_with_cross_entropy, CrossEntropyWithSoftmaxSPMDRule);
+
+// split rule
+REGISTER_SPMD_RULE(split, SplitSPMDRule);
 
 }  // namespace auto_parallel
 }  // namespace distributed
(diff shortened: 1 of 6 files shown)


paddle/fluid/distributed/auto_parallel/spmd_rules/rules.h
```diff
@@ -145,6 +146,9 @@ REGISTER_SPMD_RULE(lookup_table_v2, EmbeddingSPMDRule);
 REGISTER_SPMD_RULE(softmax, SoftmaxSPMDRule);
 REGISTER_SPMD_RULE(log_softmax, SoftmaxSPMDRule);
 
+// split rule
+REGISTER_SPMD_RULE(split, SplitSPMDRule);
```
JZ-LIANG, line 155: name string for regstration should be op names which are defined in phi yaml.
for split them are : split_with_num & split.

and you should take care about the different in attribute among these two ops

paddle/fluid/distributed/auto_parallel/spmd_rules/split_spmd_rule.cc
```diff
@@ -0,0 +1,126 @@
...
+      ninputs,
+      1,
+      phi::errors::InvalidArgument("The size of InputSpec in split must "
+                                   "be equal to 1, but got [%d].",
+                                   ninputs));
+  VerifySpecs(input_specs, "split");
+
+  // step1: Build Einsum Notation
+  int64_t ndim = input_specs[0].shape().size();
+  Attribute section_attr = GetAttr("num_or_sections", attrs);
```
JZ-LIANG: follow definition in phi yaml：
- op : split_with_num
  args : (Tensor x, int num, Scalar(int) axis)
- op : split
  args : (Tensor x, IntArray sections, Scalar(int) axis)

there is no num_or_sections

paddle/fluid/distributed/auto_parallel/spmd_rules/split_spmd_rule.cc
```diff
@@ -0,0 +1,126 @@
...
+  }
+  std::string alphabet = "abcdefghijklmnopqrstuvwxyz";
+
+  // get einsum notation for input
+  std::vector<std::string> input_axes_vec;
+  std::string input_axes = alphabet.substr(0, ndim);
+  input_axes_vec.emplace_back(input_axes);
+
+  // get einsum notation for output
+  std::string output_axes(input_axes);
```
JZ-LIANG, line 64: we should support input tensor sharded on split_axis.
the main logic for split spmd forward infer should be:

1. for axes other than the split_axis, treat as broadcast axes, copy dims from input to  outputs.
2. for split_axis:  
     the infered dim_mapping for all outputs are replicated
     the infered dst_dim_mapping for all outputs is also replicated

JZ-LIANG: use a special char for split axis, like: "k"

test/auto_parallel/spmd_rules/test_split_rule.py
```diff
@@ -0,0 +1,194 @@
...
+from paddle.distributed.fleet import auto
+
+
+class TestReductionSPMDRule(unittest.TestCase):
+    """
+    Unit tests for split spmd rule.
+    """
+
+    def setUp(self):
+        self.rule = get_spmd_rule("split")
```
JZ-LIANG, line 31: for both split and split_with_num

paddle/fluid/distributed/auto_parallel/spmd_rules/rules.h
pkuzyc: Done

paddle/fluid/distributed/auto_parallel/spmd_rules/split_spmd_rule.cc
pkuzyc: Done

paddle/fluid/distributed/auto_parallel/spmd_rules/split_spmd_rule.cc
pkuzyc: Done

test/auto_parallel/spmd_rules/test_split_rule.py
pkuzyc: Done

JZ-LIANG (approved):
LGTM

PR #55498, opened 2023-07-18
Fix card0 mempory problem on cuda-12

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Description

On cuda-12, we found that training gpt model with 8 cards comsumes more gpu memory on card0, for example,
image

The reason is that `cudaSetDevice(0)` is called in other cards (1-7), and it uses some memory.

This PR fixes that.
Pcard-70448


From00 (approved):
LGTM

PR #55504, opened 2023-07-18
onednn: remove fc_elementwise_add fusion

xinyu-intel:
### PR types
Bug fixes

### PR changes
OPs

### Description

post-sum fusion is conflicted with memory_optimization_pass because the post-sum and dst share the same buffer. This pr is to deprecate the fusion to make language model happy. We'll switch to use matmul and post-binary fusion to improve the performance then.


xinyu-intel:
@kolinwei hi, please help to check the approval in coverage job. The tests will be removed along with the feature.

xinyu-intel:
@vivienfanghuagood hi, can you take a look at this?

paddle/fluid/framework/ir/CMakeLists.txt
```diff
@@ -173,7 +173,6 @@ if(WITH_MKLDNN)
   pass_library(conv_elementwise_add_mkldnn_fuse_pass inference DIR mkldnn)
   pass_library(int8_scale_calculation_mkldnn_pass inference DIR mkldnn)
   pass_library(params_quantization_mkldnn_pass inference DIR mkldnn)
-  pass_library(fc_elementwise_add_mkldnn_fuse_pass inference DIR mkldnn)
```
vivienfanghuagood, line 176: If this pass is removed, will the performance of some models be affected?

xinyu-intel: Some will be impact. A simple case should be `linear -> add`, add won't be fused into linear and cause extra computation. However, the current post-sum fusion need memory inplace and it will be conflict with memory reuse opimization and cause wrong accuracy. We'll re-enable the fusion by using post-binary fusion which doesn't actually need memory inplace but this needs more validation work.

vivienfanghuagood (approved):

XieYunshen (approved):
LGTM
单测删除

qingqing01 (approved):

PR #55515, opened 2023-07-18
[Kunlun] Modify some legacy code on distributed training

XiaociZhang:
### PR types

New features

### PR changes

Others

### Description

There were limitations on XPUs before, such as concat/split is not supported, and c_broadcast only support fp32. These limitations are lifted recently.

Multi-device profiling on XPU will also be supported by this PR. Without this PR, a hanging broadcast will be issued by devices that enables profiling, eventually lead to kernel timeout error.


ykkk2333 (approved):
LGTM

RuohengMa (approved):
LGTM

QingshuChen (approved):

PR #55516, opened 2023-07-18
[AutoParallel] keep lr_sheduler same bewteen executor and engine

zhaoyinglia:
### PR types

Bug fixes
### PR changes

Others
### Description

Pcard-70448
- Deepcopy optimizer will copy learning_rate at the same time but it would cause a difference of learning_rate between executor and engine.
    - excutor: program._lr_scheduler()
    - engine: engine.optimizer._learning_rate()


zhiqiu (approved):
LGTM

PR #55529, opened 2023-07-19
make sharding reduce mode by default

sneaxiy:
### PR types
Others

### PR changes
Others

### Description
Make sharding reduce mode by default.


ForFishes (approved):
LGTM

PR #55540, opened 2023-07-19
pp comm overlap use tensor fusion helper

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
pp use tensor fusion helper

PCard-70444


haohongxiang (approved):
LGTM

Issue #55362, opened 2023-07-12
_DataLoaderIterMultiProcess encounter error in static graph in Paddle 2.5


Wong4j:
### bug描述 Describe the Bug

In paddle v2.5, I got the following error when training BERT using [DeepLearningExamples/PaddlePaddle/LanguageModeling/BERT/run_pretraining.py#L71](https://github.com/NVIDIA/DeepLearningExamples/blob/2693c632d8c29ed75f939b8217b435dca97c3b8f/PaddlePaddle/LanguageModeling/BERT/run_pretraining.py#L71).

Our code works well in paddle v2.3&v2.4. But since v2.5, this line no longer support static graph - [python/paddle/io/dataloader/worker.py#L387](https://github.com/PaddlePaddle/Paddle/blob/006bd959c70eed6f52ff16a290a2cbb6dd9102ee/python/paddle/io/dataloader/worker.py#L387). 

I don't know how to reproduce this bug with simple code yet. **Anyone has any idea?**

Error message:
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.10/dist-packages/paddle/io/dataloader/worker.py", line 384, in _worker_loop
    tensor_list = [
  File "/usr/local/lib/python3.10/dist-packages/paddle/io/dataloader/worker.py", line 387, in <listcomp>
    else b.value().get_tensor()
  File "/usr/local/lib/python3.10/dist-packages/decorator.py", line 232, in fun
    return caller(func, *(extras + args), **kw)
  File "/usr/local/lib/python3.10/dist-packages/paddle/fluid/wrapped_decorator.py", line 25, in __impl__
    return wrapped_func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/paddle/fluid/framework.py", line 445, in __impl__
    assert in_dygraph_mode(), (
AssertionError: We only support 'value()' in dynamic graph mode, please call 'paddle.disable_static()' to enter dynamic graph mode.
```

Wong4j:
Hi @wanghuancoder , could you please take a look?

USTCKAY:
[...]

PR #55541, opened 2023-07-19
fix data load error in static mode

USTCKAY:
### PR types

Bug fixes
### PR changes

Others
### Description

Method `value()` is only supported in dynamic mode and that leads to error in static mode. Since `value()` just return the input as output, so it can be deleted to support static mode.
https://github.com/PaddlePaddle/Paddle/blob/6a2db618c6d188e185ecd902f014c0c72fd010b3/python/paddle/fluid/dygraph/tensor_patch_methods.py#L801
This PR is to solve https://github.com/PaddlePaddle/Paddle/issues/55362


wanghuancoder (approved):

qili93 (approved):
LGTM

PR #55553, opened 2023-07-19
[NewIR]support attribute hook and fix_reduce_all

kangguangli:
### PR types

Others
### PR changes

Others
### Description

The static `reduce` operators in static graph have an attribute `reduce_all:bool`, which is corresponding to the situation that dynamic `reduce` with `axis = None`. In general, when `reduce_all` is set to `True`, the axis should be `None`, they're consistent, so we do not need to do special processing. 

However, there's some cases that `axis != None` when `reduce_all=True`. For example, in `static_ops.yaml`, we define: 
```
- op : sum
  args : (Tensor x, IntArray axis={0}, bool keepdim=false, bool reduce_all=false, int in_dtype=-1, DataType out_dtype=DataType::UNDEFINED)
  output : Tensor(out)
  infer_meta :
    func : SumRawInferMeta
    param : [x, axis, keepdim, reduce_all, out_dtype]
  kernel :
    func : sum_raw
    param : [x, axis, keepdim, reduce_all, out_dtype]
    data_type : x
  backward : sum_grad
```
Notice that `axis` has a default value: `{0}`, that's why the inconsistency occurs. The `test_reduce_op` has following cases:
```
class TestAllOp(OpTest):
    def setUp(self):
        self.op_type = "reduce_all"
        self.python_api = reduce_all_wrapper
        self.inputs = {'X': np.random.randint(0, 2, (5, 6, 10)).astype("bool")}
        self.outputs = {'Out': self.inputs['X'].all()}
        self.attrs = {'reduce_all': True}

    def test_check_output(self):
        self.check_output()
```
And this will fail because of the default value of `axis`.

Currently we only handle `reduce_all` `reduce_any`. For other `reduce` operators, we handle them when we met them, since this problem won't caused by normal use cases.

#### Others
Pcard-67164


Aurelius84 (approved):
LGTM

PR #55613, opened 2023-07-21
Fix sharding_pass_unittest stage3 precision problem

Wennie396:
### PR types

Bug fixes

### PR changes

Others

### Description

Pcard-70448
Fix sharding_pass_unittest stage3 precision problem.


test/auto_parallel/sharding_pass_unittest.py
```diff
@@ -27,7 +27,7 @@
 def apply_pass(use_sharding=False, stage=None):
```
JZ-LIANG, line 27: Enable the unitest in cmake

test/auto_parallel/sharding_pass_unittest.py
```diff
@@ -50,6 +50,7 @@ def apply_pass(use_sharding=False, stage=None):
 def reset_prog():
     paddle.fluid.framework.switch_main_program(paddle.static.Program())
     paddle.fluid.framework.switch_startup_program(paddle.static.Program())
+    paddle.utils.unique_name.switch()
 
 
 class TestShardingPass(unittest.TestCase):
```
JZ-LIANG, line 56: is the results exactly equal for runing dp twice?
if so, sharding should be exactly equal

test/auto_parallel/sharding_pass_unittest.py
Wennie396: Done

test/auto_parallel/sharding_pass_unittest.py
Wennie396: Done

JZ-LIANG (approved):
LTGM

PR #55679, opened 2023-07-25
skip ele add in grad acc

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
skip ele add in grad acc

PCard-70444


ForFishes (approved):
LGTM

PR #55766, opened 2023-07-28
[sharding optim] Optim check finite

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
Only check local params for sharding stage 1 under comm overlap.

image

PCard-70444


ForFishes (approved):
LGTM

PR #55825, opened 2023-07-31
update LBFGS to adapt to 0D tensor

lijialin03:
### PR types

Others
### PR changes

APIs
### Description

Pcard-66961
This bug occured in DeepXDE's Lorenz_inverse example.
Because 0D-tensor is allowed now, parameter could be in shape [] which will cause problem in paddle.optimizer.LBFGS API.
So update LBFGS API. Child card is DLTP-73584


cxxly:
PR补充BUG说明

cxxly (approved):
LGTM

PR #55840, opened 2023-07-31
[NewIR] add decorator for dy2st test with new ir

kangguangli:
### PR types

Others
### PR changes

Others
### Description

This PR aims to monitor the correctness of running subgraph with paddle new ir.
currently, we support three usage,
1.  overwrite the original test, only run target case once with `FLAGS_enable_new_ir_in_executor=True`
```
@test_with_new_ir
def test_train():
   pass
```
2.  run target case one more time, but with `FLAGS_enable_new_ir_in_executor=True`, do not check output
```
@test_and_compare_with_new_ir(False)
def test_train():
   pass
```
3. like above but check output
```
@test_and_compare_with_new_ir
def test_train():
   pass
```

#### Others
Pcard-67164


2742195759 (approved):
LGTM

PR #55841, opened 2023-07-31
Remove old dataloader in ut

zoooo0820:
### PR types

Others
### PR changes

Others
### Description

Pcard-67005

remove old version DataLoader (created by `from_generator`)  in UT and replace it with 2.0 version `paddle.io.DataLoader`.

There are still some case cannot be replaced, which will be removed after DataLoader upgrade.
- unit test case in Distrubuted Training.
- unit test case in Executor, which use non-iterable DataLoader.
- unit test for old version DataLoader (will be moved with the DataLoader code)


jeff41404 (approved):
LGTM

PR #55921, opened 2023-08-02
[enhancement] Implement tensor.cpu() in static graph

liudongxue01:
### PR types

Others 
### PR changes

APIs

### Description

- tensor.cpu() is empty implementation in static graph mode, but on the dygraph mode, this api performs the memory-copying behaviour(D2H).
- So on dy2static mode, the model result would not be equivalent as on dygraph mode.
Upon above, we implemented the memory-copying logic code in this api using 'memcpy' operator.
- Note:  'memcpy' operator does not have corresponding backward node, so we set 'stop_gradient=True' on the result tensor object.
- Same issue exists in tensor.cuda() api.

### Others
Pcard-73789


Aurelius84 (commented):
这里最好加一个单测，来保证后续其他同学改动之后，功能仍然是正确的

liudongxue01:
> 这里最好加一个单测，来保证后续其他同学改动之后，功能仍然是正确的

好的，我补一下。

Aurelius84 (dismissed):
LGTM

python/paddle/fluid/layers/math_op_patch.py
```diff
@@ -136,11 +136,39 @@ def create_tensor_with_batchsize(ref_var, value, dtype):
...
-        Variable should not have cpu() and cuda() interface.
-        But this interface can greatly facilitate dy2static.
-        We do nothing here.
+        in dy2static, Variable also needs cpu() and cuda() interface.
+        But, the underneath operator has only forward op but not backward one.
+        Returns:
+          the tensor which has copied to cpu place
+        Examples:
+            In Static Graph Mode:
+
```
sunzhongkai588:
```suggestion
        In dy2static, Variable also needs cpu() and cuda() interface.
        But, the underneath operator has only forward op but not backward one.
        
        Returns:
            The tensor which has copied to cpu place.
            
        Examples:
            In Static Graph Mode:

```

sunzhongkai588 (commented):
暂时没法看预览，先这样改一下叭

sunzhongkai588 (approved):
LGTM for docs

Aurelius84 (approved):

python/paddle/fluid/layers/math_op_patch.py
liudongxue01: DONE

PR #55932, opened 2023-08-03
[XPU]Add flip kernel

wz1qqx:
### PR types

Performance optimization
### PR changes

OPs
### Description

- fix fall back to cpu kernel of flip


zhupengyang (approved):
LGTM

PR #55955, opened 2023-08-03
【CINN】refactor codegen for cinn

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR refactor CodeGen for CINN. In C++, use ostringstream to concat str is 3~4x slower than std::string.operator+=. So here we replace former ways to concat str


zhhsplendid (approved):
LGTM

JamesLim-sy (approved):
LGTM, great job!

PR #55966, opened 2023-08-03
[xdoctest] reformat example code with google style in 78

Liyulingyue:
### PR types

Others 
### PR changes

Others 
### Description

https://github.com/PaddlePaddle/Paddle/issues/55629

### DIFF
--- a/python/paddle/nn/functional/input.py
+++ b/python/paddle/nn/functional/input.py
@@ -72,16 +72,19 @@ def one_hot(x, num_classes, name=None):
     Examples:
         .. code-block:: python
 
-            import paddle
+            >>> import paddle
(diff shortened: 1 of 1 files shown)


python/paddle/nn/functional/input.py
```diff
@@ -72,16 +72,19 @@ def one_hot(x, num_classes, name=None):
     Examples:
         .. code-block:: python
 
-            import paddle
+            >>> import paddle
             # Correspond to the first example above, where label.shape is 4 and one_hot_label.shape is [4, 4].
```
Liyulingyue:
```suggestion
            >>> # Correspond to the first example above, where label.shape is 4 and one_hot_label.shape is [4, 4].
```

SigureMo (approved):
LGTMeow

PR #55999, opened 2023-08-05
fix cholesky_solve document bug

### LINKED ISSUE #55998: paddle.linalg.cholesky_solve document misleading
### bug描述 Describe the Bug

The Documentation of paddle.linalg.cholesky_solve() is misleading,
```shell
    Solves a linear system of equations A @ X = B, given A's Cholesky factor matrix u and  matrix B.

    Input `x` and `y` is 2D matrices or batches of 2D matrices. If the inputs are batches, the outputs
    is also batches.

    Args:
        x (Tensor): Multiple right-hand sides of system of equations. Its shape should be `[*, M, K]`, where `*` is
            zero or more batch dimensions. Its data type should be float32 or float64.
        y (Tensor): The input matrix which is upper or lower triangular Cholesky factor of square matrix A. Its shape should be `[*, M, M]`, where `*` is zero or
            more batch dimensions. Its data type should be float32 or float64.
        upper (bool, optional): whether to consider the Cholesky factor as a lower or upper triangular matrix. Default: False.
        name(str, optional): Name for the operation (optional, default is None).
            For more information, please refer to :ref:`api_guide_Name`.

    Returns:
        Tensor: The solution of the system of equations. Its data type is the same as that of `x`.
```
As per the documentation it should accept the Cholesky matrix in the `x` argument. but the function takes the cholesky matrix at the` y` argument. Even in the official examples also does the same.
```python
import paddle

u = paddle.to_tensor([[1, 1, 1],
                        [0, 2, 1],
                        [0, 0,-1]], dtype="float64")
b = paddle.to_tensor([[0], [-9], [5]], dtype="float64")
out = paddle.linalg.cholesky_solve(b, u, upper=True)

print(out)
# [-2.5, -7, 9.5]
```
In this example you can see that the cholesky matrix is passed to the `y` argument. So changing the document would fix this.

### 其他补充信息 Additional Supplementary Information

_No response_

dhanush-2501:
### PR types
 Bug fixes

### PR changes
 Docs

### Description
Fixes #55998
Fixed the documentation

### DIFF
--- a/python/paddle/tensor/linalg.py
+++ b/python/paddle/tensor/linalg.py
@@ -3132,10 +3132,10 @@ def cholesky_solve(x, y, upper=False, name=None):
     is also batches.
 
     Args:
-        x (Tensor): The input matrix which is upper or lower triangular Cholesky factor of square matrix A. Its shape should be `[*, M, M]`, where `*` is zero or
-            more batch dimensions. Its data type should be float32 or float64.
-        y (Tensor): Multiple right-hand sides of system of equations. Its shape should be `[*, M, K]`, where `*` is
+        x (Tensor): Multiple right-hand sides of system of equations. Its shape should be `[*, M, K]`, where `*` is
             zero or more batch dimensions. Its data type should be float32 or float64.
+        y (Tensor): The input matrix which is upper or lower triangular Cholesky factor of square matrix A. Its shape should be `[*, M, M]`, where `*` is zero or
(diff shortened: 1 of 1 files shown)


zhiboniu (approved):

luotao1:
@sunzhongkai588 Please follow up on the Chinese document, thanks!

luotao1 (approved):

PR #56042, opened 2023-08-07
[Dy2St]Fix rollback error when non-forward function to_static

0x45f:
### PR types

Bug fixes
### PR changes

Others
### Description

Fix rollback error when non-forward function to_static
PCard-66972


Aurelius84 (approved):
LGTM

PR #56065, opened 2023-08-08
Mingyan.w

TonibMw:
### PR types

Bug fixes
### PR changes

APIs
### Description

Pcard-73944
Fixed a bug in the if (!id_hint.empty()) code block by adding missing code to correctly modify id_hint.


paddle/cinn/frontend/net_builder.cc
```diff
@@ -253,12 +253,8 @@ NETBUILDER_REDUCE_OP_DEF(ReduceAny, reduce_any)
 Placeholder NetBuilder::CreateInput(const Type& type,
                                     const std::vector<int>& shape,
                                     const std::string& id_hint) {
-  if (!id_hint.empty()) {
-    cinn::utils::TransValidVarName(id_hint);
-  }
-  std::string id =
-      id_hint.empty() ? Context::Global().NewName("placeholder") : id_hint;
-
+  std::string id = id_hint.empty() ? Context::Global().NewName("placeholder")
```
zhhsplendid, line 256: Usually, test covers the code branch you add. For example, since you added `id_hint.empty()`, you had better add a testcase creating input from an empty `id_hint`

paddle/cinn/frontend/net_builder_test.cc
```diff
@@ -76,6 +76,16 @@ TEST(net_build, basic) {
   }
 }
 
+TEST(net_build, TestTransValidVarName) {
```
zhhsplendid, line 79: Let's keep test name same style with line 89.

`test_trans_valid_var_name`

zhhsplendid (approved):
Leave some comments, we agreed to change in the next PR

PR #56067, opened 2023-08-08
Fix atan2 grad when dx or dy is nullptr

ronny1996:
### PR types

Bug fixes
### PR changes

OPs
### Description

Fix atan2 grad when dx or dy is nullptr


qili93 (approved):
LGTM

PR #56092, opened 2023-08-08
[dy2static]implement tensor.cuda() in static graph

liudongxue01:
### PR types
Others

### PR changes
APIs

### Description
- tensor.cuda() is empty implementation in static graph mode, but on the dygraph mode, this api performs the memory-copying behaviour(H2D or D2D).
- So on dy2static mode, the model result would not be equivalent as on dygraph mode.
- Upon above, we implemented the memory-copying logic code in this api using 'memcpy' operator.
- We alse add some fixup logic in interpreter_util.cc as a workaround of the bugs in 'memcpy' kernel.
- Note: 'memcpy' operator does not have corresponding backward node, so we set 'stop_gradient=True' on the result tensor object.

### Others
Pcard-73789

### DIFF
--- a/python/paddle/fluid/layers/math_op_patch.py
+++ b/python/paddle/fluid/layers/math_op_patch.py
@@ -146,6 +146,7 @@ def monkey_patch_variable():
             In Static Graph Mode:
 
             .. code-block:: python
+
                 import paddle
                 paddle.enable_static()
 
@@ -192,6 +193,7 @@ def monkey_patch_variable():
             In Static Graph Mode:
 
             .. code-block:: python
+
                 import paddle
                 paddle.enable_static()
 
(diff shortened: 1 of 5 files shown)


python/paddle/fluid/layers/math_op_patch.py
```diff
@@ -173,13 +173,56 @@ def cpu(self):
...
+                We add this argument for dy2static translation, please do not use it.
+
+        Returns:
+            The tensor which has copied to cuda place.
+
+        Examples:
+            In Static Graph Mode:
+
+            .. code-block:: python
+                import paddle
```
sunzhongkai588:
```suggestion
            .. code-block:: python
            
                import paddle
```

- code block 和下面的代码没加空行，导致官网解析出了点错，修一下叭～
- 之前写的 cpu 的文档，也存在相同的问题，也修改一下～ https://github.com/PaddlePaddle/Paddle/pull/55921

image

liudongxue01: DONE

sunzhongkai588 (approved):
LGTM for docs

From00 (approved):
LGTM

jeff41404 (approved):
LGTM

PR #56104, opened 2023-08-09
fix trainable in Parameter

zoooo0820:
### PR types

Bug fixes
### PR changes

Others
### Description

Pcard-67005

In class `Parameter`, there are two flags of trainable,  `self.trainable (set by API)` and `self.stop_gradient (inherited from Variable)`.  However, these two flags are not  independent while they control the same thing. (e.g. , `self.trainable` is False, while `self.stop_gradient` is False, too)

This may cause training issues with precision.

- In `fluid.optimizer.Optimizer`, whether a Parameter is trainable is determined by `self.trainable`. https://github.com/PaddlePaddle/Paddle/blob/v2.5.1/python/paddle/fluid/optimizer.py#L1396

- In `paddle.optimizer.Optimizer`, whether  a Parameter is trainable is determined by `self.stop_gradient`. https://github.com/PaddlePaddle/Paddle/blob/v2.5.1/python/paddle/optimizer/optimizer.py#L1371

In theory they should be equivalent, but the above problem leads to different results.

### DIFF
--- a/python/paddle/fluid/framework.py
+++ b/python/paddle/fluid/framework.py
@@ -7170,6 +7170,8 @@ class Parameter(Variable, metaclass=ParameterMetaClass):
         )
         self.trainable = kwargs.get('trainable', True)
 
+        self.stop_gradient = not self.trainable
+
         self.optimize_attr = kwargs.get('optimize_attr', {'learning_rate': 1.0})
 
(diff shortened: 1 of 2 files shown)


jeff41404 (approved):
LGTM

luotao1 (approved):

PR #56110, opened 2023-08-09
【CINN】refactor optimize in codegen

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR Do the Following things:
1. delete duplicated optimize call in AddFunction in CodeGenAndJit
2. delete unused visit call in CollectHostFunctionVisitor in which we actually just do something on lower_func
3. make code more clean and efficient like replace for with for_each and so on
4. change the default compile_thread according to the cpu core on cur device

After all, the cost time of parallel compile on Bert can be speed up from 318s to 299s

### DIFF
--- a/paddle/cinn/ir/module.cc
+++ b/paddle/cinn/ir/module.cc
@@ -23,17 +23,18 @@
 namespace cinn {
 namespace ir {
 
-void Module::Builder::AddFunction(const ir::LoweredFunc &func) {
-  auto tmp_func = optim::IRCopy(func);
-  optim::Simplify(&(tmp_func->body));
-  optim::SimplifyForLoops(&(tmp_func->body));
-  optim::SimplifyBlocks(&(tmp_func->body));
-  tmp_func->body = optim::Optimize(tmp_func->body, module_->target);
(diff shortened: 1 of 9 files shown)


paddle/cinn/ir/module.cc
```diff
@@ -16,17 +16,23 @@
...
 
 namespace cinn {
 namespace ir {
 
-void Module::Builder::AddFunction(ir::LoweredFunc func) {
-  optim::Simplify(&(func->body));
-  optim::SimplifyForLoops(&(func->body));
-  optim::SimplifyBlocks(&(func->body));
-  func->body = optim::Optimize(func->body, module_->target);
+void Module::Builder::AddFunction(const ir::LoweredFunc &func) {
```
6clc: 感觉这一块不使用优化的逻辑可以讨论一下，不新增一个AP，过添加一个use_optim=true的参数，或者把这些optim封装成一个pass方案。

Courtesy-Xs: done

PR #56114, opened 2023-08-09
[Prim] Fix get var in prim when list of single tensor

cyber-pioneer:
### PR types
Bug fixes

### PR changes
Others

### Description
Pcard-66975
Fix case: 
When Fix input value is list of single tensor, corresponding var value in block will be unpacked as single tensor.


zoooo0820 (approved):
LGTM

Charles-hit (approved):

PR #56116, opened 2023-08-09
[XPU] out = x when in_type and out_type are the same

HarperCy:
### PR types
Performance optimization

### PR changes
OPs

### Description
directly return x when input type = output type

### DIFF
--- a/paddle/cinn/hlir/framework/graph_compiler.cc
+++ b/paddle/cinn/hlir/framework/graph_compiler.cc
@@ -33,7 +33,7 @@
 #include "paddle/cinn/utils/profiler.h"
 
 #include "paddle/cinn/ast_gen_ius/tensor_group.h"
-
+#include "paddle/common/enforce.h"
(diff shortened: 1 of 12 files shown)


paddle/phi/kernels/xpu/cast_kernel.cc
```diff
@@ -33,6 +33,15 @@ void CastXPUKernelImpl(const Context& dev_ctx,
     return;
   }
 
+  if (x.dtype() == out->dtype()) {
+    int r = xpu::copy<XPUInT>(dev_ctx.x_context(),
+                              reinterpret_cast<const XPUInT*>(in_data),
+                              reinterpret_cast<XPUInT*>(out_data),
+                              x.numel());
```
XiaociZhang: x.numel() -> numel

XiaociZhang (approved):
LGTM

ZibinGuo (approved):
LGTM

PR #56122, opened 2023-08-09
[CINN] Add ScheduleBlock graph

BiynXu:
### PR types
New features

### PR changes
Others

### Description
Pcard-74043
Added a graph data structure in units of ScheduleBlock and some necessary operations, such as finding upstream and downstream nodes, and performing operations in the DFS topological order.


paddle/cinn/ir/tensor.cc
```diff
@@ -603,6 +603,18 @@ std::string GenReduceInitTensorNameOf(const std::string &tensor_name) {
   return tensor_name + "__reduce_init";
 }
 
+bool IsReduceInitTensorName(const std::string &tensor_name) {
+  return tensor_name.length() > 13 &&
```
ZzSean: 这个13会不会有点抽象了，要不定一个常量，名字叫`length_of_xxx`一类的

BiynXu: 已修改，为"__reduce_init"后缀创建了常量

paddle/cinn/common/dfs_topo_walker.h
```diff
@@ -0,0 +1,90 @@
...
+#include <unordered_set>
+
+namespace cinn {
+namespace common {
+
+// DFS Topological order walker
+template <typename NodeType,
+          typename NodeHash = std::hash<NodeType>,
+          typename NodeEqual = std::equal_to<NodeType>>
+class DFSTopoWalker final {
```
zhhsplendid: `DfsTopoWalker` is a better name.

paddle/cinn/common/dfs_topo_walker.h
```diff
@@ -0,0 +1,90 @@
...
+
+  using NodeHandlerType = std::function<void(NodeType)>;
+  using NodesVisitorType =
+      std::function<void(NodeType, const NodeHandlerType&)>;
+
+  DFSTopoWalker(const NodesVisitorType& VisitPreNodes,
+                const NodesVisitorType& VisitNextNodes)
+      : VisitPreNodes_(VisitPreNodes), VisitNextNodes_(VisitNextNodes) {}
+
+  void operator()(NodeType node, const NodeHandlerType& NodeHandler) const {
```
zhhsplendid, line 58: Why does `operator` on single node exists? Shouldn't it be a collection of nodes?

paddle/cinn/common/dfs_topo_walker_test.cc
```diff
@@ -0,0 +1,54 @@
...
+      });
+  std::vector<int> sources{0, 2};
+  std::vector<int> outputs;
+  walker(sources.begin(), sources.end(), [&](int node) {
+    outputs.push_back(node);
+  });
+  for (auto output : outputs) {
+    LOG(INFO) << output;
+  }
+  std::vector<int> expected{0, 1, 2, 3, 4};
```
zhhsplendid, line 49: As I mentioned in above comment, the test result seems not DFS:

Starts from 0, we have 0->1->3->4 as DFS path. There is no path from some node to 2. Then why does DFS output has 2 between 1 and 3?

Or are you trying to make some topological order?

paddle/cinn/common/dfs_topo_walker.h
```diff
@@ -0,0 +1,90 @@
...
+    };
+    const auto& TryPush = [&](NodeType node) {
+      InitInDegree(node);
+      if (visited.count(node) == 0 && in_degree[node] == 0) {
+        node_stack.push(node);
+        visited.insert(node);
+      }
+    };
+
+    for (NodeIt iter = begin; iter != end; ++iter) {
```
zhhsplendid, line 94: I don't think you are making DFS here.

In each while loop at line 74, you only visit next nodes. Not DFS (go into deep path).

I didn't know where will you use this utility class and make those templates. Why made it so complex so that you made mistake on DFS?

paddle/cinn/ir/schedule_block_graph.h
```diff
@@ -0,0 +1,200 @@
...
+ *   for1:
+ *     block0
+ *     block1
+ *   block2
+ *   for2:
+ *     block3
+ *     block4
+ *
+ * the result is:
+ *   [0]: for0
```
zhhsplendid, line 137: Just reminder, not comment, I guess you are making some marks for control structures.

1. Is integer enough for your purpose? Are you going to distinguish for\ifelse\block?
2. Should brackets block (our ir::Block, not ScheduleBlock) be considered? In C++, it will be like:
```
int a = ...
{  // brackets block begins
   int a = ... // a has different meaning and life cycle but only in brackets block
}  // brackets block ends
```

paddle/cinn/common/dfs_topo_walker.h
BiynXu: done

paddle/cinn/common/dfs_topo_walker.h
BiynXu: If the starting point is a collection, we can use the below Interface:
void operator()(NodeIt begin, NodeIt end, const NodeHandlerType& NodeHandler);
Sometimes we just want to start from one starting point, so I provide this interface.

paddle/cinn/common/dfs_topo_walker.h
BiynXu: The walking order of this class first ensures topological order.
We know that topological order is not unique, and I need a walking order that prioritizes depth as much as possible while ensuring topological order, so that the blocks with dependencies during loop fusion are as close together as possible.

paddle/cinn/ir/schedule_block_graph.h
BiynXu: I attempted to obtain the order of the ScheduleBlock, For, and IfThenElse nodes in the entire IR, which have a mutual inclusion relationship.
I constructed a map with std::vector as the key, similar to the titles used in writing articles, with different levels, and each title under same level can be sorted.

1.Therefore, the three types of nodes in this trial have the same attributes and do not need to be distinguished.
2.In addition, the content under the three types of nodes mentioned above can be considered as Blocks, with only one more parenthesis at this point. Currently, we may not consider it.

paddle/cinn/common/dfs_topo_walker_test.cc
BiynXu: Starts from 0, we have 0->1->3->4 as DFS path, but the node 3 has another input node 2, which has not yet been accessed. In order to ensure topological order, we prioritize accessing node 2, and then continue to walk in DFS order from node 2, ultimately we have 0->1->2->3->4.

paddle/cinn/common/dfs_topo_walker.h
zhhsplendid: Please **must** add comment to this method. This method doesn't work for all computation graphs. It only works for a graph only contains 1 node whose in_degree = 0 and you use **that** single node as the input.

For example, if a graph has 1 -> 2 <- 3, you input 1, you can only walk on 1 -> 2

So this method has some limitations but users would NOT know unless read the implementation whose developer is you. As OOP concept. You had better comment about the limitation.

paddle/cinn/common/dfs_topo_walker.h
zhhsplendid: Please add comments above the class about this order.

paddle/cinn/ir/schedule_block_graph.h
zhhsplendid: Just suggestion/reminder :-)

paddle/cinn/common/dfs_topo_walker.h
BiynXu: I have added some comments.

paddle/cinn/common/dfs_topo_walker.h
BiynXu: I have added some comments.

paddle/cinn/ir/schedule_block_graph.h
BiynXu: Currently, there is no need to differentiate between 3 types. If there is a need in the future, further improvements will be made.

zhhsplendid (approved):
LGTM

PR #56254, opened 2023-08-14
c_allreduce use single-intput-output interface to speed up

RichardWooSJTU:
### PR types

Performance optimization
### PR changes

OPs
### Description

Allreduce uses multi-input-output interface now, which caused gap between all_reduce kernel and other kernels.  By applying this PR, LLM inference latency can be reduced significantly. 

Pcard-71502


heavengate (approved):

hitywt (approved):
LGTM

PR #56264, opened 2023-08-14
[Approval] Ci appoval for fluid

zoooo0820:
### PR types

Others
### PR changes

Others
### Description

Pcard-67005

Add file modification check of `python/paddle/fluid` in CI-Approval. If there are any modifacation, there will raise `You must have one RD (zoooo0820(Recommend), or jeff41404) approval for file changes in python/paddle/fluid, because fluid API is going to be removed.\n`


python/paddle/fluid/log_helper.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+from . import core  # add this for test
```
zoooo0820: image

Info above raised for this change

tianshuo78520a (approved):

jeff41404 (approved):

PR #56272, opened 2023-08-14
add range support in indexing

zoooo0820:
### PR types

New features
### PR changes

APIs
### Description

Pcard-66985

Support `class <range>` in indexing getitem and setitem. Its behavior is same to `List`

In past:
- getitem (not support)
- setitem (only support with combined indexing, with Numpy backend)

### DIFF
--- a/test/indexing/test_setitem.py
+++ b/test/indexing/test_setitem.py
@@ -51,6 +51,15 @@ class TestSetitemInDygraph(unittest.TestCase):
 
         np.testing.assert_allclose(x.numpy(), np_data)
 
+    def test_index_has_range(self):
+        np_data = np.ones((3, 4, 5, 6), dtype='int32')
+        x = paddle.to_tensor(np_data)
+
+        np_data[:, range(3), [1, 2, 4]] = 10
+        x[:, range(3), [1, 2, 4]] = 10
+
+        np.testing.assert_allclose(x.numpy(), np_data)
+
 
 class TestSetitemInStatic(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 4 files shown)


test/indexing/test_setitem.py
```diff
@@ -51,6 +51,15 @@ def test_combined_index_3(self):
...
 
+    def test_index_has_range(self):
+        np_data = np.ones((3, 4, 5, 6), dtype='int32')
+        x = paddle.to_tensor(np_data)
+
+        np_data[:, range(3), [1, 2, 4]] = 10
+        x[:, range(3), [1, 2, 4]] = 10
+
+        np.testing.assert_allclose(x.numpy(), np_data)
+
```
jeff41404, line 62: getitem should also be check, eg. `y = x[:, range(3), [1, 2, 4]]`

test/indexing/test_setitem.py
```diff
@@ -137,3 +146,19 @@ def test_combined_index_5(self):
...
+        ):
+            x = paddle.ones((3, 4, 5, 6), dtype='int32')
+            y = _setitem_static(
+                x,
+                (slice(None, None), range(3), [1, 2, 4]),
+                10,
+            )
+            res = self.exe.run(fetch_list=[y.name])
+
+        np.testing.assert_allclose(res[0], np_data)
```
jeff41404, line 164: getitem should also be check here

test/indexing/test_setitem.py
zoooo0820: Done, thanks

test/indexing/test_setitem.py
zoooo0820: Done, thanks

jeff41404 (approved):
LGTM

PR #56282, opened 2023-08-14
modify default parallel compile thread

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR modify default compile thread of parallel compiler as half of the core number of current cpu and do something to make code more clean


ZzSean (approved):
LGTM

PR #56283, opened 2023-08-14
Fix `sharding_pass` and "nop" op to improve GC strategy

heavyrain-lzy:
### PR types

Bug fixes
### PR changes

Others
### Description

PCard-71568
1. The operator `nop` in the `pipeline strategy` can be used to hold the memory of the variable, so we can't add `no_need_buffer` feature to its input and output.
2. The "shading pass" should  also shard the inputs and outputs of  the operator `check_finite_and_unscale` and `update_loss_scaling` in the `stage 1`


JZ-LIANG (approved):
LGTM for sharding

From00 (approved):

PR #56299, opened 2023-08-15
Migrate matmul_int8 op  to ops.yaml from legacy_ops.yaml

RichardWooSJTU:
### PR types

Bug fixes
### PR changes

OPs
### Description

matmul_int8 op which registered in legacy_ops.yaml can just be used in dygraph mode. To support static graph, this op should be registered in ops.yaml

Pcard-71502

### DIFF
--- a/paddle/phi/api/yaml/legacy_ops.yaml
+++ b/paddle/phi/api/yaml/legacy_ops.yaml
@@ -521,14 +521,6 @@
     func : matmul
   backward : matmul_grad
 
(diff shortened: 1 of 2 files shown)


chenwhql (changes requested):
不能单独为某一个运算的单个数据类型新增一个算子，而应该整合到原kernel中扩展数据类型支持，此处matmul_int8需要整合到matmul kernel中

PR #56380, opened 2023-08-17
[CustomDevice] add default stream safe allocator support

ronny1996:
### PR types

New features
### PR changes

Others
### Description

add default stream safe allocator support


qili93 (approved):
LGTM

PR #56384, opened 2023-08-17
[ONEDNN] fix LSTM FP32 model acc drop issue, and add fc_…

zhanglirong1999:
### PR types
Bug fixes

### PR changes
OPs

### Description
LSTM FP32 model shows significantly accuracy drop when using mkldnn. The reason is fc_mkldnn_op implement has some bug.


zhanglirong1999:
This pr fix the `fc_mkldnn_op` weight key bug.
Since we plan to replace fc operator to matmul operator in phi, the unit test of fc operator has not been submitted yet. It will be added after the upgrade is completed.

paddle/fluid/operators/mkldnn/fc_mkldnn_op.cc
```diff
@@ -410,7 +416,8 @@ class FCMKLDNNKernel : public framework::OpKernel<T_in> {
         phi::funcs::CreateKey(dev_ctx,
                               ctx.InputName("Input"),
                               ctx.InputName("W"),
-                              phi::vectorize(x->dims())));
+                              phi::vectorize(x->dims()),
+                              phi::vectorize(weights->dims())));
```
YangQun1, line 420: do we need to also encode src/weight layout information to this cache key? how about adding a test case for same shape but different layout.

xinyu-intel (approved):
LGTM. We will finally remove this OP and use MatMul instead.

YangQun1 (approved):

PR #56393, opened 2023-08-17
[CINN] Python  DSL of CINN IR

6clc:
### PR types
Others

### PR changes
Others

### Description
Pcard-72423

支持通过Python DSL定制CINN Lower Level IR的主PR。

单测目录： `test/cinn/ir/`
e2e测试目录： `test/cinn/runtime`

- [x] 修改IRCompare，支持对比IR的结构是否相同 https://github.com/PaddlePaddle/Paddle/pull/57515
- [x] 封装一些需要上下文信息才能构建的IR。https://github.com/PaddlePaddle/Paddle/pull/57515
- [x] C++ 和 Python 通信接口。 https://github.com/PaddlePaddle/Paddle/pull/57644
- [x] 解析python dsl的compute定义。 https://github.com/PaddlePaddle/Paddle/pull/57731/
- [x] 解析python dsl的schedule定义。https://github.com/PaddlePaddle/Paddle/pull/57981
- [x] 实现python dsl RunTime模块。https://github.com/PaddlePaddle/Paddle/pull/58009

以element wise add为例
```python

import cinn
import numpy as np
from cinn import to_cinn_llir
from cinn.runtime.data_array import DataArray

# python dsl构造elementwise add
@to_cinn_llir
def add_kernel(X, Y, Z, N):
    for idx in range(N):
        Z[idx] = X[idx] + Y[idx]


def test_launch():
    N = 32
    X_np = np.random.random(N).astype(np.float32)
    Y_np = np.random.random(N).astype(np.float32)
    Z_np = np.zeros((N), dtype=np.float32)
    target = cinn.common.DefaultNVGPUTarget()
    
    # 构造数据
    X = DataArray.from_numpy(X_np, target)
    Y = DataArray.from_numpy(Y_np, target)
    Z = DataArray.from_numpy(Z_np, target)

    # compile and run
    add_kernel[target](X, Y, Z, N)
    
    pred = Z.to_numpy()
    gt = np.add(X_np, Y_np)
    np.testing.assert_allclose(pred, gt)
```

### DIFF
--- a/python/cinn/compiler/compute_code_generator.py
+++ b/python/cinn/compiler/compute_code_generator.py
@@ -1,0 +1,251 @@
+# Copyright (c) 2023 PaddlePaddle Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ast
+import contextlib
+
+from cinn import ir
+
+from .expr_executor import ExprExecutor, exec_assign
+from .utils import VariableTable, node_is_schedule
+
+
+class ComputeCodeGenerator(ast.NodeVisitor):
+    """
+    Convert python ast to CINN Lower Level IR,
+    containing only the semantics of the compute part
+    """
+
+    def __init__(self, fn, function_name, inputs_signature):
+        self.fn = fn
+        self.function_name = function_name
+        self.inputs_signature = inputs_signature
+        self.cinn_llir_func = None
+        self.variables_table = VariableTable()
+        self.extra_scope = {"range": ir.sequential}
+
+    def parse(self):
+        ast_node = self.fn.parse()
+        with ir.IRBuilder() as builder, self.variables_table:
+            for k, v in self.fn.scope.items():
+                self.variables_table.add(k, v)
+            for k, v in self.extra_scope.items():
+                self.variables_table.add(k, v)
+            self.visit(ast_node)
+        return builder.get()
+
+    def visit_FunctionDef(self, node) -> None:
+        """
+        Parse CINN Low Level IR FunctionDef.
+
+        Args:
+            node(ast.FunctionDef): The ast FunctionDef Node
+        """
+        with ir.LowerFuncContext(self.function_name) as func_ctx:
+            arg_names = self.visit(node.args)
+
+            assert len(node.args.defaults) == 0, "Not support default args"
+
+            # 1. Construct args of function
+            for i, arg_name in enumerate(arg_names):
+                # Obj of Argument is ir::Buffer
+                if hasattr(self.inputs_signature[i], "dtype"):
+                    tensor_shape = [
+                        ir.Expr(dim) for dim in self.inputs_signature[i].shape
+                    ]
+                    # TODO(6clc): unify Tensor and Buffer
+                    llir_value = ir._Buffer_.make(
+                        arg_name, self.inputs_signature[i].dtype
+                    )
+                    ir.Arg(arg_name, llir_value)
+                    llir_value = ir._Tensor_.make(
+                        arg_name,
+                        self.inputs_signature[i].dtype,
+                        tensor_shape,
+                        tensor_shape,
+                    )
+                    self.variables_table.add(arg_name, llir_value)
+                # Obj of Argument is ir::Var
+                else:
+                    llir_value = ir.Var(arg_name)
+                    ir.Arg(arg_name, llir_value)
+                    llir_value = ir.Expr(llir_value)
+                    self.variables_table.add(arg_name, llir_value)
+
+            # 2. Construct body of function
+            body = self.visit_compound_statement(node.body)
+
+    def visit_compound_statement(self, stmts):
+        for stmt in stmts:
+            self.visit(stmt)
+
+    def visit_arguments(self, node):
+        """
+        Parse CINN Low Level IR Argument.
+        If it is not jit mode, it will get information from arg.annoatation.
+
+        Args:
+            node(ast.arguments): The ast argument Node
+
+        Returns:
+            list[string]: A list of parameter names
+        """
+        arg_names = [arg.arg for arg in node.args]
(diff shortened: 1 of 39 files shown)


paddle/cinn/ir/ir.cc
```diff
@@ -268,6 +268,27 @@ Expr For::Make(Var loop_var,
 
   return Expr(node);
 }
+Expr For::Make(Var loop_var, Expr min, Expr extent, Expr body) {
+  auto node = make_shared<For>();
+  CHECK(loop_var.defined());
```
zhhsplendid: Please add message for CHECK failure. Now we would like every error message can be user-friendly.

Reference: https://github.com/PaddlePaddle/Paddle/wiki/Paddle-Error-Message-Writing-Specification

paddle/cinn/ir/ir.cc
```diff
@@ -268,6 +268,27 @@ Expr For::Make(Var loop_var,
...
+  node->min = min;
+  node->extent = extent;
+  node->device_api = DeviceAPI::UNK;
+  node->body = body;
+  node->set_for_type(ForType::Serial);
+  node->set_vectorize_info(VectorizeInfo());
+  BindInfo bind_info = BindInfo();
+  node->set_bind_info(BindInfo());
+
+  if (node->is_vectorized()) CHECK(node->vectorize_info().valid());
```
zhhsplendid: Line 286\287, put any controlled statements inside blocks (i.e. use curly braces).

Reference: https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching

paddle/cinn/ir/lowered_func.h
```diff
@@ -30,8 +30,10 @@ class _LoweredFunc_;
  * the function signature of generated code.
  */
 struct Argument {
-  //! Input or output.
-  enum class IO { kInput = 0, kOutput = 1 };
+  //! kInput: arg is input
+  //! kOutput: arg is output
+  //! kUnknown: arg maybe input or output
+  enum class IO { kInput = 0, kOutput = 1, kUnknown = 2 };
```
zhhsplendid: Is IO a good name? Because IO is also related to print\file\hardware\network, can we name it as `InputOutputType`? `ArgInOutType` or something else?

paddle/cinn/ir/tensor.cc
```diff
@@ -52,6 +52,23 @@ Tensor _Tensor_::Make(const std::string &name,
 
   return Tensor(n);
 }
+Tensor _Tensor_::Make(const std::string &name,
+                      Type dtype,
+                      const std::vector<Expr> &shape,
+                      const std::vector<Expr> &domain,
+                      const std::vector<Var> &reduce_axis) {
+  CHECK(!name.empty()) << "Tensor name is set empty";
```
zhhsplendid: "Cannot set empty Tensor name in _Tensor_::Make"

python/cinn/runtime/cinn_jit.py
```diff
@@ -0,0 +1,104 @@
...
+import ast
+import functools
+import inspect
+import textwrap
+from typing import Callable, Generic, Optional, TypeVar, Union, cast
+
+T = TypeVar('T')
+
+
+class CINNLowerLevelIRJIT(Generic[T]):
```
zhhsplendid: CinnLowerLevelIrJit is a better name

python/cinn/runtime/data_array.py
```diff
@@ -0,0 +1,54 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import cinn
+import numpy as np
+from cinn import runtime
+
+
+class DataArray:
+    """ """
```
zhhsplendid: Why is here empty comment?

python/cinn/compiler/compute_code_generator.py
```diff
@@ -0,0 +1,190 @@
...
+            if node.id in self.local_variables:
+                return self.local_variables[node.id]
+            return node.id
+        # Load Node
+        assert (
+            node.id in self.local_variables
+        ), f"{node.id} is not defined in context"
+        return self.local_variables[node.id]
+
+    def visit_BinOp(self, node):
```
zhhsplendid: Is this a temporary function?

I asked because it only handled Add here, and BinOp can potentially not only Load. So I guess it may not be correct?

python/cinn/compiler/compute_code_generator.py
```diff
@@ -0,0 +1,190 @@
...
+        ), "CINN Low Level IR does not support setting the range step"
+        ast_min = iter_args[0] if len(iter_args) > 1 else self.visit(ast.Num(0))
+        ast_extent = iter_args[1] if len(iter_args) > 1 else iter_args[0]
+
+        # TODO(6clc): support sub region's local variable
+        llir_var = ir.Var(node.target.id)
+        llir_var_expr = ir.Expr(llir_var)
+        self.set_value(node.target.id, llir_var_expr)
+
+        llir_for_min = ir.Expr(ast_min)
```
zhhsplendid: I guess LLIR For doesn't support `step`? Should we report error to users if user has `step` in for range?

paddle/cinn/ir/ir.cc
6clc: Done

paddle/cinn/ir/ir.cc
6clc: Done

paddle/cinn/ir/lowered_func.h
6clc: Indeed, this name is not very good, but so many modules of the current CINN have relied on this IO. It is too troublesome to modify. This is an enum property in ir::Argument,  the ambiguity doesn't seem too serious.

paddle/cinn/ir/tensor.cc
6clc: Done

python/cinn/runtime/cinn_jit.py
6clc: Done

python/cinn/runtime/data_array.py
6clc: Done

python/cinn/compiler/compute_code_generator.py
6clc: `step` has already been checked in the previous code, almost at line 109.

```python
        assert (
            len(iter_args) <= 2
        ), "CINN Low Level IR does not support setting the range step"
```

python/cinn/compiler/compute_code_generator.py
6clc: The code about `visit_XXX` will be re-written by new commit

PR #56397, opened 2023-08-17
Optimize the memory in the case of using `pipeline` strategy and new executor

heavyrain-lzy:
### PR types

Performance optimization
### PR changes

Others
### Description

PCard-71568
Optimize the GPU memory used in the case of using `pipeline` strategy and new executor.
1. The temporary block in the "gradient_merge pass" should be created by "paddle.static.Program().global_block()" not "main_program._crate_block()"
2. When collecting the variables to `gc`,the varibales that are used by `nop` and `c_sync_comm`  needn't to be included.
3. When  using `RecordStreamForGC` for `send_v2`, we should add extra code in the case of `use_calc_stream == false` to get the real stream.
4. The pipeline strategy with new executor depends on `static_build`.

### DIFF
--- a/paddle/fluid/framework/new_executor/program_interpreter.h
+++ b/paddle/fluid/framework/new_executor/program_interpreter.h
@@ -81,7 +81,7 @@ class ProgramInterpreter : public InterpreterBaseImpl {
     hookfuncs_ = hookfuncs;
   }
 
-  bool IsRealStaticBuild() const { return static_build_; }
+  bool IsStaticBuild() const { return static_build_; }
 
  private:
   // build graph
--- a/paddle/fluid/framework/new_executor/program_interpreter.h
+++ b/paddle/fluid/framework/new_executor/program_interpreter.h
@@ -81,6 +81,8 @@ class ProgramInterpreter : public InterpreterBaseImpl {
     hookfuncs_ = hookfuncs;
   }
 
+  bool IsRealStaticBuild() const { return static_build_; }
+
  private:
   // build graph
   void Convert(std::vector<paddle::framework::OpFuncNode>* op_func_nodes);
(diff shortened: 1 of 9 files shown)


paddle/fluid/framework/new_executor/program_interpreter.h
```diff
@@ -81,6 +81,8 @@ class ProgramInterpreter : public InterpreterBaseImpl {
     hookfuncs_ = hookfuncs;
   }
 
+  bool IsRealStaticBuild() const { return static_build_; }
```
From00:
```suggestion
  bool IsStaticBuild() const { return static_build_; }
```
RealStaticBuild的命名容易让人误解，建议直接和成员变量static_build_保持一致命名。

python/paddle/distributed/passes/auto_parallel_gradient_merge.py
```diff
@@ -33,12 +33,11 @@
 world_process_group = get_world_process_group()
 
 
-def _remove_and_get_optimizer_op(main_program, dist_context):
+def _remove_and_get_optimizer_op(main_program, temp_block, dist_context):
```
From00: 通过函数参数传入`仅供被调用函数内部使用的临时变量`是一种很奇怪的做法，建议在`_remove_and_get_optimizer_op`中直接创建，不需要在调用函数中创建后再传入。

paddle/fluid/framework/new_executor/program_interpreter.h
heavyrain-lzy: Done.Thanks!

python/paddle/distributed/passes/auto_parallel_gradient_merge.py
heavyrain-lzy: Done.Thanks!

From00 (approved):
LGTM

PR #56437, opened 2023-08-18
do not use fuse for sync param in dp

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
do not use fuse for sync param in dp

PCard-70444


Xreki (approved):
LGTM

PR #56483, opened 2023-08-21
Fix TensorRT multiclass_nms Test

eee4017:
### PR types
Bug fixes

### PR changes
Others

### Description
In multiclass_nms, the results will be arranged in order, given that the tensorrt output sequence differs from that of paddle. Errors may arise even if the selected boxes are identical but the scores do not match precisely. I believe the root of this issue lies in the score generation mechanism. Using fp16 for generating scores isn't effective since it cannot manage large figures. For instance, numbers like 4097 and 4098 in fp32 would both map to 4096 in fp16. Hence, it's essential to adjust the score generation process to ensure unique scores under fp16 precision.


lyuwenyu (approved):
LGTM

heavengate (approved):

PR #56497, opened 2023-08-21
add check for empty parameter desc

zhiqiu:
### PR types

Others
### PR changes

Others
### Description

add check for empty parameter desc
Pcard-67164


kangguangli (approved):
LGTM

PR #56502, opened 2023-08-21
[Semi-Auto] Add softmax infer_backward rule

pkuzyc:
### PR types
Function optimization

### PR changes
Others

### Description
Pcard-70448

Add infer backward SPMD rule for softmax. It infers inputs' dims mappings as following:

For non-normalized axes, set their dims mappings equal to the outputs'. For the normalized axis, set it to -1.

### DIFF
--- a/paddle/fluid/distributed/auto_parallel/spmd_rules/softmax_spmd_rule.cc
+++ b/paddle/fluid/distributed/auto_parallel/spmd_rules/softmax_spmd_rule.cc
@@ -20,10 +20,10 @@ namespace auto_parallel {
 
 using phi::distributed::auto_parallel::str_join;
 
-// step0: verify input args based on softmax logic
 std::pair<std::vector<TensorDistAttr>, std::vector<TensorDistAttr>>
 SoftmaxSPMDRule::InferForward(const std::vector<DistTensorSpec>& input_specs,
                               const paddle::framework::AttributeMap& attrs) {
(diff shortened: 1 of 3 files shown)


paddle/fluid/distributed/auto_parallel/spmd_rules/softmax_spmd_rule.cc
```diff
@@ -107,9 +107,71 @@ SoftmaxSPMDRule::InferForward(const std::vector<DistTensorSpec>& input_specs,
...
+  int axis = ExtractAttr<int>("axis", attrs);
+  // normalize axis
+  if (axis < 0) {
+    axis = x_ndim + axis;
+  }
+
+  // sharding on softmax_axis is not supported now, so set
+  // the notation on softmax_axis to '1' so that we can set
+  // its dim mapping to -1
+  x_axes[axis] = '1';
```
JZ-LIANG, line 146: change the einsum_string to "1" to ensure the unsharded maybe not a good idea here, since it would disconnect the relation of softmax_axis between input and output when print the vlog for debug: 

"ab1" --> "abc", later developer might not realize that  axis "1" and "c" are the same axis in two tensors.

it would be better that:
first: out_dims_mapping[axis] = -1;
then: keep the input_axes unchanged.

JZ-LIANG (approved):
Update in Next PR

PR #56504, opened 2023-08-21
[Semi-Auto] Add reduction infer_backward rule

pkuzyc:
### PR types
Function optimization

### PR changes
Others

### Description
Pcard-70448

Add infer backward SPMD rule for reduction. It infers inputs' dims mappings as following:

For non-reduced axes, set their dims mappings equal to the outputs'. For the reduced axes, set them to -1.

### DIFF
--- a/paddle/fluid/distributed/auto_parallel/spmd_rules/reduction_spmd_rule.h
+++ b/paddle/fluid/distributed/auto_parallel/spmd_rules/reduction_spmd_rule.h
@@ -32,8 +32,14 @@ class ReductionSPMDRule : public SPMDRuleBase {
                const paddle::framework::AttributeMap& attrs) override;
 
   std::pair<std::vector<TensorDistAttr>, std::vector<TensorDistAttr>>
-  InferBackward(const std::vector<DistTensorSpec>& output_specs,
+  InferBackward(const std::vector<DistTensorSpec>& input_specs,
+                const std::vector<DistTensorSpec>& output_specs,
                 const paddle::framework::AttributeMap& attrs) override;
+
+ private:
+  std::string GetOutputNotation(int64_t input_ndim,
+                                const std::string& input_axes,
+                                const paddle::framework::AttributeMap& attrs);
 };
 }  // namespace auto_parallel
 }  // namespace distributed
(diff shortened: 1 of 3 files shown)


paddle/fluid/distributed/auto_parallel/spmd_rules/reduction_spmd_rule.h
```diff
@@ -32,8 +32,14 @@ class ReductionSPMDRule : public SPMDRuleBase {
                const paddle::framework::AttributeMap& attrs) override;
 
   std::pair<std::vector<TensorDistAttr>, std::vector<TensorDistAttr>>
-  InferBackward(const std::vector<DistTensorSpec>& output_specs,
+  InferBackward(const std::vector<DistTensorSpec>& input_specs,
+                const std::vector<DistTensorSpec>& output_specs,
                 const paddle::framework::AttributeMap& attrs) override;
+
+ private:
+  std::string GetOutputNotation(int64_t input_ndim,
```
JZ-LIANG, line 40: the class for rule will be remove and fw & bw will be independent functions.
so place this utils another place and  as  independent function

test/auto_parallel/spmd_rules/test_reduction_rule.py
```diff
@@ -233,6 +235,171 @@ def test_multi_mesh_dim(self):
...
+        infered_input_dist_attrs = result_dist_attrs[0]
+        infered_output_dist_attrs = result_dist_attrs[1]
+
+        self.assertEqual(len(result_dist_attrs), 2)
+        self.assertEqual(len(infered_input_dist_attrs), 1)
+        self.assertEqual(len(infered_output_dist_attrs), 1)
+
+        self.assertEqual(infered_input_dist_attrs[0].dims_mapping, [-1, -1])
+        self.assertEqual(infered_output_dist_attrs[0].dims_mapping, [-1])
+
```
JZ-LIANG, line 257: check the partial is false

test/auto_parallel/spmd_rules/test_reduction_rule.py
```diff
@@ -233,6 +235,171 @@ def test_multi_mesh_dim(self):
...
+        # reduce on dim 0, keep_dim = true
```
JZ-LIANG, line 258: should add a kind new case that output is partial on curtain dims, and how it influence input :
stage 1: input is not effect
stage2: input should be sharded on that dims

JZ-LIANG (approved):
Update in Next PR

PR #56505, opened 2023-08-21
[Semi-Auto] Add layer_norm infer_backward rule 

pkuzyc:
### PR types
Function optimization

### PR changes
Others

### Description
Pcard-70448

Add infer_backward rule for layer_norm to infer inputs' dims mappings from outputs'.

### DIFF
--- a/paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
+++ b/paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
@@ -123,15 +123,20 @@ LayerNormSPMDRule::InferForward(const std::vector<DistTensorSpec>& input_specs,
   out_dims_mapping.reserve(out_axes.size());
 
   int64_t mean_shard_dim = -1;
-  for (size_t i = 0; i < out_axes.size(); ++i) {
-    if (i < static_cast<size_t>(begin_norm_axis)) {
-      out_dims_mapping.push_back(x_dims_mapping[i]);
+  // As the mean and variance in outputs are `flattened` from
+  // x[0:begin_norm_axis], only the first axis can be sharded,
+  // the axes 1 to begin_norm_axis-1 are set to be replicated.
+  std::vector<int64_t> x_dims_mapping_dst(x_ndim, -1);
+  x_dims_mapping_dst[0] = x_dims_mapping[0];
+  for (int i = 0; i < x_ndim; ++i) {
+    if (i < begin_norm_axis) {
+      out_dims_mapping.push_back(x_dims_mapping_dst[i]);
       // if ijk,k,k->ijk,z,z (x,scale,bias->out,mean,variance,
       // begin_norm_axis=2, z=ij), and the dims_mapping of input is (0,1,-1),
       // the mean and varience is sharded by dim 0 and 1,
       // which is not supported currently.
-      mean_shard_dim =
-          ShardingMergeForAxis(mean_axes, mean_shard_dim, x_dims_mapping[i]);
+      mean_shard_dim = ShardingMergeForAxis(
+          mean_axes, mean_shard_dim, x_dims_mapping_dst[i]);
     } else {
       out_dims_mapping.push_back(-1);
     }
@@ -141,7 +146,7 @@ LayerNormSPMDRule::InferForward(const std::vector<DistTensorSpec>& input_specs,
   varience_dist_attr_dst.set_dims_mapping({mean_shard_dim});
 
   // step2.3: Merge and get Inputs' New Dims Mapping.
-  x_dist_attr_dst.set_dims_mapping(out_dims_mapping);
+  x_dist_attr_dst.set_dims_mapping(x_dims_mapping_dst);
   input_dist_attrs.emplace_back(x_dist_attr_dst);
   // TODO(zhiqiu): support shardding on scale and bias
   // Now, apply replicating.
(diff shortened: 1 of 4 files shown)


paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
```diff
@@ -173,12 +172,128 @@ LayerNormSPMDRule::InferForward(const std::vector<DistTensorSpec>& input_specs,
...
+  for (int64_t i = 0; i < 1; i++) {
+    VLOG(4) << "Output" << std::to_string(i) << " shape: ["
+            << str_join(output_specs[i].shape()) << "] "
+            << "Einsum Notation: " << output_axes_vec[i]
+            << " src_dims_mapping: ["
+            << str_join(output_specs[i].dims_mapping()) << "] "
+            << "dst_dims_mapping: ["
+            << str_join(output_dist_attrs[i].dims_mapping()) << "]";
+  }
+  VLOG(4) << "*********";
```
JZ-LIANG: remove meaningless log

paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
```diff
@@ -74,9 +73,9 @@ LayerNormSPMDRule::InferForward(const std::vector<DistTensorSpec>& input_specs,
           << begin_norm_axis << "]; ";
 
   // step1: build Einsum Notation
-  // ijk,k,k->ijk,x,x (x,scale,bias->out,mean,variance, begin_norm_axis=2, x=ij)
-  // ijkl,y(kl),y(kl)->ijkl,x(ij),x(ij) (x,scale,bias->out,mean,variance,
-  // begin_norm_axis=2, x=ij, y=kl)
+  // ijk,k,k->ijk,z,z (x,scale,bias->out,mean,variance, begin_norm_axis=2, z=ij)
+  // ijkl,y(kl),y(kl)->ijkl,z(ij),z(ij) (x,scale,bias->out,mean,variance,
+  // begin_norm_axis=2, z=ij, y=kl)
```
JZ-LIANG, line 78: below in line 101: mean_axes = "j";  the notation should be "x" ? 
otherwise "j" maybe confuse with the broadcast axis in before begin_norm_axis.

paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
```diff
@@ -127,8 +126,8 @@ LayerNormSPMDRule::InferForward(const std::vector<DistTensorSpec>& input_specs,
   for (size_t i = 0; i < out_axes.size(); ++i) {
     if (i < static_cast<size_t>(begin_norm_axis)) {
       out_dims_mapping.push_back(x_dims_mapping[i]);
-      // if ijk,k,k->ijk,x,x (x,scale,bias->out,mean,variance,
-      // begin_norm_axis=2, x=ij), and the dims_mapping of input is (0,1,-1),
+      // if ijk,k,k->ijk,z,z (x,scale,bias->out,mean,variance,
```
JZ-LIANG, line 134: the current rule for LN is problematic:

- not all axes before begin_norm_axis could be sharded.
- the first axis after begin_norm_axis could sharded in current implementation.
- it would be better to refactor the LN rule using TransDim algorithms: the axes mapping is like axes-flatten in TransDim.

JZ-LIANG: but it is ok for now, since most usage of LN is DP.

paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
pkuzyc: Done

paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
pkuzyc: Line 101 is the case when begin_norm_axis<=1. When begin_norm_axis<=1, the first axis can be propagated to mean and var, so mean_axes and var_axes is set to be the same as input's first axis, which is "j".

paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
pkuzyc: Done, now only the first axis of the input can be sharded, other axes will be set to replicated.

paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
JZ-LIANG: but mean_axis maybe a composed axis that merge multiple axes, so it is not corresponding to "first axis" in that case.

JZ-LIANG (approved):
LGTM

paddle/fluid/distributed/auto_parallel/spmd_rules/layer_norm_spmd_rule.cc
JZ-LIANG: U are right

PR #56509, opened 2023-08-21
[Semi-Auto] Add transpose infer_backward rule

pkuzyc:
### PR types
Function optimization

### PR changes
Others

### Description
Pcard-70448

Add infer_backward rule for layer_norm to infer inputs' dims mappings from outputs': set the dims mapping of each axis in input the same as the corresponding output axis.


JZ-LIANG (approved):
LGTM

PR #56539, opened 2023-08-22
fix fetch problem in pass_utils.py and eval_loss in parallelizer_v2.py

Wennie396:
### PR types
Bug fixes

### PR changes
Others

### Description
Pcard-70448
fix fetch problem in pass_utils.py. Bug: some devices have some fetch vars but don't have relevant ops, fetch error
fix bug in parallelizer_v2.py. Bug: in eval mode and pp_degree > 1 the loss in output has len(loss)>1


zhaoyinglia (approved):
LGTM

PR #56613, opened 2023-08-24
solve value-src dtype fault

zoooo0820:
### PR types

Bug fixes
### PR changes

APIs
### Description

Pcard-66985

Support case in `__setitem__` when dtype of `src` and `value` are not same.


jeff41404 (approved):
LGTM

PR #56643, opened 2023-08-24
Fix static_build for pp

heavyrain-lzy:
### PR types

Bug fixes
### PR changes

Others
### Description

PCard-71568
Fix bug when using `static_build` strategy in the case of `pipeline` parallelism.


From00 (approved):
LGTM

PR #56662, opened 2023-08-25
[CINN] Fix instant variable oom in paddle2cinn

FisherWY:
### PR types
Bug fixes

### PR changes
Others

### Description
Pcard-73602
When using `paddle2cinn`, `CompilationContext.with_instantiate_variables` should be set to `false`, otherwise CINN will instant and manage variables memory, this leads to double the memory usage, which eventually leads to out of memory error.
This PR will set `CompilationContext.with_instantiate_variables` to false before context pass to constructing the graph compiler.


BiynXu (approved):
LGTM

zhhsplendid (approved):
LGTM

PR #56701, opened 2023-08-28
[ROCM] Remove the constraint with a maximum number of threads per block of 256, P3

ronny1996:
### PR types

Others
### PR changes

Others
### Description

Remove the constraint with a maximum number of threads per block of 256, P3


qili93 (approved):
LGTM

PR #56775, opened 2023-08-29
skip data_transfer for save op

zhiqiu:
### PR types

Bug fixes
### PR changes

OPs
### Description

skip data_transfer for save op
Pcard-67164


From00 (approved):
LGTM

PR #48380, opened 2022-11-25
CI check fluid

zhengqiwen1997:
### PR types

Others
### PR changes

Others
### Describe

Add new CI function during PR-CI-Static-Check that checks if new reference of fluid has been added under the folder of python/paddle.

PR #56792, opened 2023-08-30
Check fluid ci

zoooo0820:
### PR types

Others
### PR changes

Others
### Description

Pcard-67005

Based on  https://github.com/PaddlePaddle/Paddle/pull/48380, we try to add fluid API check in PR-CI-Static-Check for two purposes： 
- Although there is a check for modification of `python/paddle/fluid`, but we found there are some other newly added files using fluid APIs， e.g. (`test/` for unittests), blocking the migrated fluid PR.
- Similar to the above, even though the migrated fluid PR is merged as planned, there is a potential hidden conflict (between the migrated fluid PR and other PRs which using fluid APIs)

So we try to check the fluid APIs usage in `python/paddle/` and `test/`.

### DIFF
--- a/tools/check_api_approvals.sh
+++ b/tools/check_api_approvals.sh
@@ -108,6 +108,19 @@ if [ "$inference_approve" != "" ]; then
     check_approval 1 qingqing01 heavengate
 fi
 
+filter_fluid=`git diff --name-only upstream/develop |  grep "py$" | grep "^test/"`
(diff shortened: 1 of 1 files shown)


test/fft/test_fft.py
```diff
@@ -19,6 +19,7 @@
 import scipy.fft
 
 import paddle
+from paddle.fluid import set_flags  # noqa: F401
```
zoooo0820: image
测试添加此项后，CI正确拦截效果

tianshuo78520a (approved):

jeff41404 (approved):

PR #56803, opened 2023-08-30
[Auto Parallel] Add op cost interface

Caozhou1995:
### PR types

New features
### PR changes

Others
### Description

PCard-74851
This PR adds op cost interface, including communication OP and computation OP. 
The communication OP time is based on the modeling formula, and the computation OP time is converted based on FLOPs.


zhaoyinglia (approved):
LGTM，后续使用 parallelizer_v2 替换 parallelizer

PR #56845, opened 2023-08-31
[CINN]revise cinn fp16 matmul cbulas api to gemmex

GGBond8488:
### PR types
 Others

### PR changes
 OPs

### Description
revise cinn fp16 matmul cbulas api to gemmex which is same as paddle
pcard-77116


jeff41404 (approved):
LGTM

PR #56868, opened 2023-09-01
fix setvalue dtype error when using dy2st and amp O2

zoooo0820:
### PR types

Bug fixes
### PR changes

Others
### Description

Pcard-66985

Before PR https://github.com/PaddlePaddle/Paddle/pull/53682 , in static mode (or dynamic to static), `__setitem__` (with OP `set_value`) just overwrite the input Variable. So the dtype of input/output is naturally consistent for `set_value`, which do not need to process in AMP.  However, in above PR, to ensure SSA, the output is changed to be a totally new Variable, which  may cause dtype inconsistent for `set_value` in AMP.  This PR add `set_value` to need_process list, to make sure dtype consistent between input/output.

### DIFF
--- a/python/paddle/static/amp/fp16_utils.py
+++ b/python/paddle/static/amp/fp16_utils.py
@@ -642,6 +642,10 @@ def cast_model_to_fp16(
 
     def need_process(op):
         need_process = True
+        if op.type in ["set_value"]:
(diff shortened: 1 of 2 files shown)


zhangting2020 (approved):

luotao1 (approved):

jeff41404 (approved):
LGTM

PR #56888, opened 2023-09-01
Modify compare logical inplace

GGBond8488:
### PR types
Others 

### PR changes
 OPs 

### Description
Pcard-75099
set compare and logcial out dtype bool


paddle/phi/kernels/kps/logical_kernel.cu
```diff
@@ -25,24 +25,45 @@
...
+  std::vector<const DenseTensor*> ins = {&x, &y};
+  std::vector<DenseTensor*> outs = {out};
+  funcs::BroadcastKernel<bool>(dev_ctx, ins, &outs, binary_func);
+}
+
+template <typename T, typename Context, typename Functor>
+void InplaceLogicalKnernelImpl(const Context& dev_ctx,
+                               const DenseTensor& x,
+                               const DenseTensor& y,
+                               DenseTensor* out) {
```
jeff41404: why create `InplaceLogicalKnernelImpl` here while not create in `logical_kernel.cc`, can we integrate code of `LogicalKnernelImpl` and `InplaceLogicalKnernelImpl` into `Logical##type##Kernel` to maintain the same form as `logical_kernel.cc`

jeff41404 (approved):

PR #56893, opened 2023-09-01
Normalize multi dim list in indexing

zoooo0820:
### PR types

Breaking changes
### PR changes

APIs
### Description

Pcard-66985

#### Background Information:

**1. Semantic of List/Tensor and Tuple**: In indexing,  the semantics of `List` type is same with `Tensor/Array` type (corresponding to one axis). As a comparison，`Tuple` type is different (corresponding to multiple axes, each element is  an index on the corresponding axis).

**2. The current semantic of List are contradictory**:Currently in Paddle, `List` type have two semantics. 

- If rank of List is one, it corresponds to one axis (same with `Tensor/Array`, which is naturally). 

```python
>>> a = paddle.randn((3,2))
>>> a
Tensor(shape=[3, 2], dtype=float32, place=Place(gpu:0), stop_gradient=True,
       [[ 0.07773950,  0.19835377],
        [-0.81423134,  0.60788709],
        [ 1.38384938,  0.00627374]])

# one-dim List is same as Tensor
>>> a[[0,1]]
Tensor(shape=[2, 2], dtype=float32, place=Place(gpu:0), stop_gradient=True,
       [[ 0.07773950,  0.19835377],
        [-0.81423134,  0.60788709]])
>>> a[paddle.to_tensor([0,1])]
Tensor(shape=[2, 2], dtype=float32, place=Place(gpu:0), stop_gradient=True,
       [[ 0.07773950,  0.19835377],
        [-0.81423134,  0.60788709]])
```

- if its rank is greater than 1, it means that it may correspond to multiple axes (The outermost `[]` is equivalent to tuple，which is to wrap indexes on different axes together).
```python
# In this case, List is same with Tuple, and different with Tensor/Array
>>> a[[[0],[1]]]
Tensor(shape=[1], dtype=float32, place=Place(gpu:0), stop_gradient=True,
       [0.19835377])
>>> a[([0],[1])]
Tensor(shape=[1], dtype=float32, place=Place(gpu:0), stop_gradient=True,
       [0.19835377])
>>> a[paddle.to_tensor([[0],[1]])]
Tensor(shape=[2, 1, 2], dtype=float32, place=Place(gpu:0), stop_gradient=True,
       [[[ 0.07773950,  0.19835377]],

        [[-0.81423134,  0.60788709]]])
```

#### What this PR did
This PR unified the semantics of `List` in indexing, like Numpy(as described in https://github.com/PaddlePaddle/Paddle/pull/51466 ).  From this PR, `List` is same with `Tensor/Array` in any case.


jeff41404:
The fundamental modification of thist PR is to unify the tensor indexing semantics of one-dimensional list and multidimensional list. So, it is recommended to use tensor indexing of one-dimensional list and tensor indexing of multidimensional list for comparison in the examples of description.

jeff41404 (approved):
LGTM

Issue #46978, opened 2022-10-13
A BUG of `paddle.bincount` on GPU


ice-tong:
### 请提出你的问题 Please ask your question

Hi, thanks for the nice project!

I got weird behavior when using `paddle.bincount` with **GPU**. 

Here are my simple test code:
```python
import paddle
import numpy as np

print(paddle.__version__)

def test_bincount(inputs, minlength):
    paddle_outs = paddle.bincount(inputs, minlength=minlength)
    numpy_outs = np.bincount(inputs.numpy(), minlength=minlength)
    assert np.all(paddle_outs.numpy() == numpy_outs), (paddle_outs.numpy(), numpy_outs)

paddle.device.set_device("cpu")
print('test bincount with cpu ...')
test_bincount(inputs=paddle.zeros(shape=(520, ), dtype='int64'), minlength=1) 

paddle.device.set_device("gpu")
print('test bincount with gpu ...')
test_bincount(inputs=paddle.zeros(shape=(500, ), dtype='int64'), minlength=1)   # this case works well
test_bincount(inputs=paddle.zeros(shape=(520, ), dtype='int64'), minlength=1)   # but this case got weird behavior 
```

Got the following outputs:
```
2.4.0-rc0
test bincount with cpu
test bincount with gpu
W1013 14:52:48.809008 227951 gpu_resources.cc:61] Please NOTE: device: 0, GPU Compute Capability: 8.0, Driver API Version: 11.7, Runtime API Version: 11.2
W1013 14:52:48.812227 227951 gpu_resources.cc:91] device: 0, cuDNN Version: 8.1.
Traceback (most recent call last):
  File "/home/xiaoyancong/workspace/PaddleSeg/test_bincount.py", line 18, in <module>
    test_bincount(inputs=paddle.zeros(shape=(520, ), dtype='int64'), minlength=1)  # but this case got weird behavior 
  File "/home/xiaoyancong/workspace/PaddleSeg/test_bincount.py", line 9, in test_bincount
    assert np.all(paddle_outs.numpy() == numpy_outs), (paddle_outs.numpy(), numpy_outs)
AssertionError: (array([1040]), array([520]))
```

PR #56941, opened 2023-09-04
fixed a bug in bincount_kenel, which will cause miscalculation when numel is greater than 512

idontkonwher:
### PR types
Bug fixes

### PR changes
OPs

### Description
As described in issue #46978，when number of elements is greater than 512 (block size greater than 1), the result of paddle.bincount is not consistent with numpy.bincount, it's because each thread is already corresponded with the input element, and the for loop is redundant.


luotao1:
@idontkonwher 可以考虑补充issue里的case到单测中

PR #56955, opened 2023-09-05
Auto codegen for supporting calling new_ir api in static operants

lxd-cumt:
### PR types

New features
### PR changes

Others
### Description

Pcard-66975
1. Add auto codegen logic, to support calling new_ir api in static operants.
2. Add FLAGS_enable_new_ir_api in test_vjp_prim, to verity the correctness of static operants under newir.
3. Fix a bug in extended_tensor.

Taking AddOp as an example, the generated code is as follows：
```c++
Tensor StaticTensorOperants::add(const Tensor& x, const Tensor& y) {
    if (FLAGS_enable_new_ir_api) {
        return paddle::primitive::backend::add<LazyTensor>(x, y);
    } else {
        return paddle::prim::add<DescTensor>(x, y);
    }
}
```


Charles-hit:
PR描述中贴一下生成的代码例子

Charles-hit (approved):

cyber-pioneer (approved):

PR #57023, opened 2023-09-06
function setitem support with stride

zoooo0820:
### PR types

Function optimization
### PR changes

Others
### Description

Pcard-66985

With the introduction of the stride mechanism in dygraph, and the relevant OPs output view has been enabled by default, indexing in Paddle needs to be adjusted accordingly.

In dygraph mode:
- `__getitem__`: Basic-indexing will output view of Tensor. 
- `__setitem__`: Advanced-indexing just assign value directly on the view of original Tensor, without an additional set_value. Moreover, we call set_value dygraph function instead of append_op, since the latter does not support the stride mechanism.


test/indexing/test_getitem.py
```diff
@@ -335,6 +335,56 @@ def test_index_has_range(self):
...
+        np_data = np.ones((5, 5, 5), dtype='float32')
+        np_tmp = np_data[None]
+        np_tmp[:, 2] = 20
+
+        x = paddle.ones((5, 5, 5), dtype='float32')
+        x_tmp = x[None]
+        x_tmp[:, 2] = 20
+
+        np.testing.assert_allclose(x.numpy(), np_data)
+
```
jeff41404, line 386: shall we add test case of ellipsis index

zoooo0820: done, add a case to test ellipsis index.

jeff41404 (approved):
LGTM

PR #57094, opened 2023-09-08
remove fluid check in CI

zoooo0820:
### PR types

Others
### PR changes

Others
### Description

Pcard-67005

Since the `python/paddle/fluid` is no longer existed, we removed CI check for using `paddle.fluid.xxx` and keep the CI check of `python/paddle/fluid` for a while.

### DIFF
--- a/tools/check_file_diff_approvals.sh
+++ b/tools/check_file_diff_approvals.sh
@@ -341,6 +341,11 @@ if [ "${HAS_MODIFIED_STATIC_BUILD}" != "" ] && [ "${GIT_PR_ID}" != ""]; then
     check_approval 1 From00 zhiqiu
 fi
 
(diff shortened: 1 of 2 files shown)


tools/check_file_diff_approvals.sh
```diff
@@ -341,11 +341,6 @@ if [ "${HAS_MODIFIED_STATIC_BUILD}" != "" ] && [ "${GIT_PR_ID}" != ""]; then
     check_approval 1 From00 zhiqiu
 fi
 
-HAS_MODIFIED_PY_FLUID=`git diff --name-only upstream/$BRANCH | grep "python/paddle/base" || true`
```
jeff41404, line 344: should be `python/paddle/fluid` ?

zoooo0820: Done, thanks

jeff41404 (approved):
LGTM

PR #57100, opened 2023-09-08
fix to tensor return nested structure

feifei-111:
### PR types

Others
### PR changes

Others
### Description

fix bug when input has nested structure, to_tensor will loss struct information

### Others

PCard-66972


SigureMo (approved):
LGTMeow 🐾

PR #57241, opened 2023-09-12
[Semi-Auto] Adapt reduction rule to phi

pkuzyc:
### PR types
Others

### PR changes
Others

### Description
Pcard-73145
1. Adapt reduction rule to phi. 
2. Modify the attribute.h, move vector<int> before vector<bool> to avoid error conversion from python integer list.


JZ-LIANG (approved):
LGTM

PR #57259, opened 2023-09-13
[Semi-Auto] Adapt transpose rule to phi

pkuzyc:
### PR types
Function optimization

### PR changes
Others

### Description
Pcard-73145

Adapt and move transpose spmd rule to phi.


JZ-LIANG (approved):
LGTM

PR #57269, opened 2023-09-13
[Semi Auto] Revise spmd rule static mode API

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Description

Pcard-70448

Refactor the static mode python api to distinguish input between tensor and [tensor].

Now the python api signature is exact the same as its C++  implement (which is like the format in Op's Python API).

Before:
```python
# matmul rule
results = rule.infer_forward( [x, y], [True, False] ) # need to wrap tensor and attr into list
results = rule.infer_backward( [x, y], [out], [True, False] )

# concate rule
results = rule.infer_forward([2], [v1, v2], [axis]) # ranges, list of tensors, list of attrs
results = rule.infer_backward([2, 1], [v1, v2, out], [axis])

# replicate rule
results = rule.infer_forward([2, 2], [x, y, out1, out2], [])  # placeholder for attrs
results = rule.infer_backward([2, 2], [x, y, out1, out2], [])

```

After:
```python
# matmul rule
result_dist_attrs = self.rule.infer_forward( x, y, True, False ) # exactly like matmul Python API
result_dist_attrs = self.rule.infer_backward( x, y, out, True, False)

# concate rule
results = rule.infer_forward( [v1, v2], axis) # exactly like concat Python API
results = rule.infer_backward([v1, v2], out, axis) 

# replicate rule
results = rule.infer_forward([x, y], [out1, out2])
results = rule.infer_backward([x, y], [out1, out2])

```

TODO：

1. exactly match the Python API (filter the non-relative attributes internally)
2. supporting more ops using codegen.
3. replace dist_tensor_spec as tensor(dist_tensor)


paddle/fluid/pybind/auto_parallel_py.cc
```diff
@@ -634,5 +540,131 @@ void BindAutoParallel(py::module *m) {
...
+static std::pair<std::vector<TensorDistAttr>, std::vector<TensorDistAttr>>
+infer_forward(const phi::distributed::SpmdRule &self, const py::args &args) {
+  VLOG(6) << "infer_forward ";
+  phi::distributed::InferSpmdContext ctx;
+  prepare_ctx(&ctx, args);
+  return self.InferForward(ctx);
+}
+
+static std::pair<std::vector<TensorDistAttr>, std::vector<TensorDistAttr>>
+infer_backward(const phi::distributed::SpmdRule &self, const py::args &args) {
```
zhiqiu, line 670: py::args& can be const?

paddle/fluid/pybind/auto_parallel_py.cc
```diff
@@ -634,5 +540,131 @@ void BindAutoParallel(py::module *m) {
...
+        infer_spmd_string,
+        arg_pos,
+        ((PyTypeObject *)obj->ob_type)->tp_name));  // NOLINT
+  }
+}
+
+static void parse_single_pyobject(PyObject *obj,
+                                  phi::distributed::InferSpmdContext *ctx,
+                                  const size_t arg_pos) {
+  if (PyList_Check(obj)) {  // list inputs, spmd not allow tuple inputs
```
zhiqiu, line 636: I suggest we can support tuple and list both.

paddle/fluid/pybind/auto_parallel_py.cc
```diff
@@ -634,5 +540,131 @@ void BindAutoParallel(py::module *m) {
...
+  ctx->EmplaceBackInputs(ins);
+}
+
+static void parse_tensor(PyObject *obj,
+                         phi::distributed::InferSpmdContext *ctx,
+                         const size_t arg_pos) {
+  VLOG(6) << "args indx: [" << arg_pos << "] input one tensor.";
+  DistTensorSpec in = py::cast<DistTensorSpec>(obj);
+  VLOG(6) << "DistTensorSpec: " << in.to_string();
+  ctx->EmplaceBackInput(phi::distributed::DistMetaTensor(
```
zhiqiu, line 576: Maybe we can consider unify DistMetaTensor and DistTensorSpec in the future.

zhiqiu (dismissed):
LGTM overall

paddle/fluid/pybind/auto_parallel_py.cc
JZ-LIANG: it work

paddle/fluid/pybind/auto_parallel_py.cc
JZ-LIANG: there is no tuple argument in Phi API yaml right now, we would support it once there is tuple input defined by Phi.

paddle/fluid/pybind/auto_parallel_py.cc
JZ-LIANG: yes, that is on the list

zhaoyinglia (approved):

PR #57280, opened 2023-09-13
[Semi-Auto] Adapt softmax spmd rule to phi

pkuzyc:
### PR types
Function optimization

### PR changes
Others

### Description
Pcard-73145

Adapt and move softmax spmd rule to phi.


JZ-LIANG (approved):
LGTM

PR #57310, opened 2023-09-14
Update eval frame behavior

feifei-111:
### PR types

Others
### PR changes

Others
### Description

we can now return sth like `CustomCode(None, Ture)`

### Others

PCard-66972


SigureMo (approved):
LGTMeow 🐾

PR #57569, opened 2023-09-21
[CINN] Add ReductionFactoring rule

BiynXu:
### PR types
Others

### PR changes
Others

### Description
Pcard-74042
Add ReductionFactoring rule


paddle/cinn/auto_schedule/post_schedule_rule/cooperative_process_test.cc
```diff
@@ -181,7 +181,7 @@ TEST_F(TestCooperativeProcess, Matmul) {
                   {
                     i0_0, i1_0, i2 = axis.bind(((2 * (i_0_j_0_fused / 2)) + ((16 * (i_j_fused / 2)) + i_1)), ((8 * (i_0_j_0_fused % 2)) + ((16 * (i_j_fused % 2)) + j_1)), ((4 * reduce_k_0) + reduce_k_1))
                     {
-                      temp_matmul_out[((2 * (i_0_j_0_fused / 2)) + ((16 * (i_j_fused / 2)) + i_1)), ((8 * (i_0_j_0_fused % 2)) + ((16 * (i_j_fused % 2)) + j_1))] = (temp_matmul_out[((2 * (i_0_j_0_fused / 2)) + ((16 * (i_j_fused / 2)) + i_1)), ((8 * (i_0_j_0_fused % 2)) + ((16 * (i_j_fused % 2)) + j_1))] + (X_reshape_shared_temp_buffer[((2 * (i_0_j_0_fused / 2)) + ((16 * (i_j_fused / 2)) + i_1)), ((4 * reduce_k_0) + reduce_k_1)] * Y_reshape_shared_temp_buffer[((4 * reduce_k_0) + reduce_k_1), ((8 * (i_0_j_0_fused % 2)) + ((16 * (i_j_fused % 2)) + j_1))]))
+                      temp_matmul_out[i0_0, i1_0] = (temp_matmul_out[i0_0, i1_0] + (X_reshape_shared_temp_buffer[i0_0, i2] * Y_reshape_shared_temp_buffer[i2, i1_0]))
```
zhhsplendid, line 184: Good change! It looks like our IR will be much simpler in the future :-)

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.h
```diff
@@ -0,0 +1,52 @@
...
+  ~ReductionFactoring() = default;
+
+  // In the future, we will no longer use this interface.
+  RuleApplyType Init(ir::IRSchedule* init_schedule) override {
+    return RuleApplyType::kCannotApply;
+  }
+  // In the future, we will no longer use this interface.
+  void Apply(int index) override { return; }
+
+  RuleApplyType AnalyseApplyType(SearchState state,
```
zhhsplendid, line 43: 1. Typo: Analyse -> Analyze

But this is small error, you can change in the next PR

2. Should we use `const SearchState& `?

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.h
```diff
@@ -0,0 +1,52 @@
...
+  explicit ReductionFactoring(const common::Target& target)
+      : AutoGenRule(target) {}
+  ~ReductionFactoring() = default;
+
+  // In the future, we will no longer use this interface.
+  RuleApplyType Init(ir::IRSchedule* init_schedule) override {
+    return RuleApplyType::kCannotApply;
+  }
+  // In the future, we will no longer use this interface.
+  void Apply(int index) override { return; }
```
zhhsplendid: Why we don't need it any more?

Even if we don't use this interface, we still have two problems in this interface.
1. If not use anymore, can we delete from base?
2. If we do not delete here, replace `return` with some error report, future SWE won't call it wrongly, and you can skip the obvious comment :-)

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.h
```diff
@@ -0,0 +1,52 @@
...
+
+  RuleApplyType AnalyseApplyType(SearchState state,
+                                 const std::string& block_name) const override;
+
+  std::string GetRuleName() const override { return "ReductionFactoring"; }
+
+  std::vector<SearchState> ApplyOnBlock(SearchState state,
```
zhhsplendid, line 48: Should we use `const SearchState&` ?

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
```diff
@@ -0,0 +1,187 @@
...
+#include "paddle/cinn/ir/schedule/ir_schedule_util.h"
+#include "paddle/cinn/ir/tensor.h"
+#include "paddle/cinn/ir/utils/ir_copy.h"
+#include "paddle/cinn/ir/utils/ir_nodes_collector.h"
+#include "paddle/cinn/ir/utils/ir_printer.h"
+
+namespace cinn {
+namespace auto_schedule {
+
+bool CanApply(const std::string& block_name, ir::IRSchedule* ir_schedule) {
```
zhhsplendid: `CanApply` only serves `ReductionFactoring`, should it be a private method of `ReductionFactoring`?

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
```diff
@@ -0,0 +1,187 @@
...
+  if (sch_block->write_buffers.empty()) {
+    return false;
+  }
+
+  // 2. The block must have at least one reduce axis
+  const std::vector<ir::Var>& iter_vars = sch_block->iter_vars;
+  bool find_reduce_axis = false;
+  for (int i = 0; i < iter_vars.size(); ++i) {
+    if (iter_vars[i]->is_reduce_axis) {
+      find_reduce_axis = true;
```
zhhsplendid, line 51: Add a `break` can speed up slightly.

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
```diff
@@ -0,0 +1,187 @@
...
+RuleApplyType ReductionFactoring::AnalyseApplyType(
+    SearchState state, const std::string& block_name) const {
+  return CanApply(block_name, &state->ir_schedule)
+             ? RuleApplyType::kApply
+             : RuleApplyType::kCannotApply;
+}
+
+std::vector<SearchState> ReductionFactoring::ApplyOnBlock(
+    SearchState state, const std::string& block_name) {
+  SearchState new_state = state.Copy();
```
zhhsplendid, line 105: Can we use `const SearchState&`?

In this code, you wanted a `Copy`, but you invoked constructor copy and `Copy` method. Is `Copy` method const? Maybe we should let it be const and input `const SearchState&` to avoid constructor copy.

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
```diff
@@ -0,0 +1,187 @@
...
+  Apply(block_name, &new_state->ir_schedule);
```
zhhsplendid: Maybe `&(new_state->ir_schedule)` would be clear. Not all people always remember the operator precedence

paddle/cinn/auto_schedule/analysis/analyze_ir.h
```diff
@@ -48,5 +48,15 @@ ir::LoweredFunc UpdateFuncWithNewBody(const common::Target& target,
                                       const ir::LoweredFunc& old_func,
                                       ir::Expr& body);  // NOLINT
 
+/**
+ * Get loop var names of reduce axis
+ */
+std::unordered_set<std::string> GetReduceLoopVarName(const ir::Expr block);
```
zhhsplendid: How about `GetReduceLoopVarNames`? The `s` to indicate it returns collection of names.

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.h
BiynXu: I think const SearchState& is better, but this is an override function. In the future, if we restart the auto tune project, I will make unified modifications.

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.h
BiynXu: This interface is an old version, which will analyze the blocks in IR that can apply this rule during initialization. The index of this interface is the subscript of these blocks. However, some schedule transformations can reduce or increase blocks, such as CacheRead/Write, Inline, etc., so the index may be invalid. The new version interface is to input a block, and in the future, if the auto tune project restarts, the old interface can be deleted uniformly.

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.h
BiynXu: I think const SearchState& is better, but this is an override function. In the future, if we restart the auto tune project, I will make unified modifications.

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
BiynXu: The SearchState is a Shared, constructor copy is a shallow copy，and Copy() method is a deep copy.
Copy() method is const, so I think const SearchState& is better, but this is an override function. In the future, if we restart the auto tune project, I will make unified modifications.

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
BiynXu: done

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
BiynXu: done

paddle/cinn/auto_schedule/search_space/auto_gen_rule/reduction_factoring.cc
BiynXu: done

paddle/cinn/auto_schedule/analysis/analyze_ir.h
BiynXu: done

zhhsplendid (dismissed):
LGTM

zhhsplendid (approved):
LGTM

PR #57587, opened 2023-09-21
【CINN】Move basic ADT from ir/utils to ir/

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR move basic ADT including ir_visitor, ir_mutator and ir_printer from ir/utils to ir/
Reason :
Consider two points. The first point is that these classes are not utils semantically, but serve as base classes to assist in implementing specific functions. Second point, judging from the original implementation, the original developer of CINN hopes to implement the corresponding operator = << of IR in this file. Considering the koenig search rules, this is an interface of IR and should be placed under the IR namespace.


ZzSean (approved):
LGTM

XiaoguangHu01 (approved):
LGTM

PR #57777, opened 2023-09-26
[CINN] Add FactorizeReduction schedule primitive

BiynXu:
### PR types
Others

### PR changes
Others

### Description
Pcard-74042
Add FactorizeReduction schedule primitive.
The difference between FactorizeReduction primitive and the original RFactor primitive are:
1. FactorizeReduction supports complex iters_value subscript, which means that FactorizeReduction can be used after using primitives such as Fuse and Split, and RFactor does not support this.
2. FactorizeReduction does not change the order of the original loop, while RFactor may have an implicit Reorder.
3. FactorizeReduction supports the transformation of one reduce block in a complex AST, while RFactor only supports the case where the AST is entirely composed of one reduce block.


paddle/cinn/ir/schedule/ir_schedule.cc
```diff
@@ -717,6 +718,448 @@ Expr ScheduleImpl::Rfactor(const Expr& rf_loop, int rf_axis) {
...
+                                    original_tensor->type(),
+                                    new_shape,
+                                    new_shape,
+                                    original_tensor->operation,
+                                    original_tensor->reduce_axis);
+  rf_tensor->WithBuffer("global", name, original_tensor->type());
+  return rf_tensor;
+}
+
+class ReduceBlockCreater {
```
zhhsplendid: Since naming "Reduce Block" is every common in CINN, could you add one sentence comment for `ReduceBlockCreater` that it is used to create reduce block in FactorizeReduction

Same for RBBlockCreater, add a comment that it is used for writing back

paddle/cinn/ir/schedule/ir_schedule.cc
```diff
@@ -120,6 +120,7 @@ class ScheduleImpl {
   void ReverseComputeInline(const Expr& schedule_block);
   void Bind(const Expr& loop, const std::string& thread_axis);
   Expr Rfactor(const Expr& rf_loop, int rf_axis);
+  Expr FactorizeReduction(const Expr& rf_loop, int rf_axis);
```
zhhsplendid, line 124: ir_schedule.cc is too long to maintain now ....

I suggest you have following file structure

```
- paddle/cinn/ir/schedule
    - factorize_reduction.h .cc or folder if you also split util files
    - ir_schedule.h .cc  which include factorize_reduction.h
```

Then developers still include `ir_schedule.h` and they can use factorize_reduction, our file is also short.

paddle/cinn/ir/schedule/ir_schedule.cc
BiynXu: done

paddle/cinn/ir/schedule/ir_schedule.cc
BiynXu: Created a new file factorize_reduction.h and has moved the core implementation of this primitive to this file. However, the interface function FactorizeReduction and function implementation are still preserved in ir_schedule.cc, because this is a member function, and other member functions are called in it.

zhhsplendid: In the future PR, we should still move it.

zhhsplendid (approved):
LGTM but agreed on we should move some implementations of ir_schedule.cc because the file is too long now.

PR #57903, opened 2023-10-07
fix the bug where excluded_layers of paddle.amp.decorate is invalid

FlyingQianMM:
### PR types

Bug fixes 

### PR changes

APIs

### Description

The parameter `excluded_layers` of API `paddle.amp.decorate` is invalid, we use python API `instance` instead of `type` to filter all  excluded layers.


zhangting2020 (approved):
LGTM

PR #57946, opened 2023-10-09
[PIR]Migrate log into pir

0x45f:
### PR types

Others
### PR changes

Others
### Description

[PIR]Migrate log into pir
Pcard-67164


Aurelius84 (approved):
LGTM

PR #57948, opened 2023-10-09
[Paddle Inference]add_varlen_python_api

Wangzheee:
### PR types

Others
### PR changes

Others
### Description

Add varlen python api
Pcard-71503

### DIFF
--- a/python/paddle/inference/__init__.py
+++ b/python/paddle/inference/__init__.py
@@ -50,5 +50,4 @@ __all__ = [  # noqa
     'get_num_bytes_of_data_type',
     'PredictorPool',
     'XpuConfig',
-    'InternalUtils',
 ]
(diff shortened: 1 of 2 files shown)


python/paddle/inference/__init__.py
```diff
@@ -49,4 +50,5 @@
     'get_num_bytes_of_data_type',
     'PredictorPool',
     'XpuConfig',
+    'InternalUtils',
```
jeff41404: if `InternalUtils` is not public API, then delete this line and does not affect normal use.

Wangzheee: > if `InternalUtils` is not public API, then delete this line and does not affect normal use.

done

yuanlehome (approved):

PR #57952, opened 2023-10-09
fix new comm flags

hitywt:
### PR types
Others

### PR changes
Others

### Description
Pcard-70448

1. fix new comm flags usage


GhostScreaming (approved):
LGTM

PR #57981, opened 2023-10-10
cinn(py-dsl): parse schedule of python dsl

6clc:
### PR types

Others

### PR changes

Others

### Description

Pcard-72423

拆分新特性：CINN Python DSL， 主PR和单测见：https://github.com/PaddlePaddle/Paddle/pull/56393

此PR只负责 **解析python dsl中的schedule定义**

### DIFF
--- a/python/cinn/compiler/compiler.py
+++ b/python/cinn/compiler/compiler.py
@@ -15,6 +15,7 @@
 
 from ..runtime import CinnLowerLevelIrJit
 from .compute_code_generator import ComputeCodeGenerator
+from .schedule_code_generator import ScheduleCodeGenerator
 
 
 def ast_to_llir(fn, inputs_signature):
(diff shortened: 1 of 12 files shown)


python/cinn/compiler/compiler.py
```diff
@@ -35,4 +39,3 @@ def compile(fn, just_convert=False, jit_inputs_signature=[], **kwargs):
 
     if just_convert:
         return llir_func
-    return llir_func
```
zhhsplendid, line 38: Maybe problem here?

6clc: The next pr will add runtime, and what will be returned here is the CINN runtime module encapsulation. For now, just convert python dsl to cinn c++ ir as normal.

zhhsplendid (approved):
LGTM

PR #57986, opened 2023-10-10
remove redundant equal in bool tensor indexing 

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985

Remove one redundant `equal OP` in bool tensor indexing since the item is already Bool type.


jeff41404 (approved):
LGTM

PR #58008, opened 2023-10-11
make sharding main grad ut serial

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
PCard-70444
make sharding main grad ut run serial


Xreki (approved):
LGTM

PR #58037, opened 2023-10-12
fix mp bug with new dist strategy

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
PCard-70444
fix mp new config bug


ForFishes (approved):
LGTM

PR #58065, opened 2023-10-13
fix cond create undefined var in global block

2742195759:
### PR types

Performance optimization

### PR changes

Others

### Description

fix cond create undefined var in global block
NOTE：
对于None的场景适配要更复杂点，还没有完全适配。比如
```
def func(b):
    a = None
    if b : 
        a = paddle.to_tensor([1,2])
    return 
```
PCard-66972


zhangbo9674 (approved):

SigureMo (approved):
LGTMeow 🐾

PR #58133, opened 2023-10-17
Adapt test_collective_global_gather with Numpy >= 1.25.2

mingxu1067:
### PR types

Bug fixes
### PR changes

Others
### Description

It is not supported for elementwise comparison between np.array and python-builtin list when numpy >= 1.25.2, then casues failures on UT `test_collective_global_gather`. Proposed this PR to fix.

PS: In Numpy 1.22.0, there is a corresponding warning when running `test_collective_global_gather`.
 ```
DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
 ```


ForFishes (approved):
LGTM

Issue #58050, opened 2023-10-12
pr_58021 caused a decrease in the accuracy of executing mkldnn inference for model det_mv3_east


EmmonsCurse:
### bug描述 Describe the Bug

**Error Description：** After the merge of https://github.com/PaddlePaddle/Paddle/pull/58021 the MKL-DNN inference accuracy of the det_mv3_east has decreased.

The max_diff comparison:

- test_det_mv3_east_mkldnn.py 
   before(oneDNN v3.1.1)：6.600469e-05 
   after(oneDNN v3.2.1): 151.867

**1. Operating environment**
PaddlePaddle version：develop
OS version：CentOS
Python version：3.8

**2. The detailed error information: (The comparison strategy is to exit when the first one is greater than the diff threshold)**
```
test_det_mv3_east_mkldnn.py:54:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_mkldnn_helper.TestModelInferenceMKLDNN object at 0x7f3f5f4f1d60>
result = array([-16.915035 , -12.312555 , -13.390684 , ...,   0.2960205,
        -0.9806156,  -4.6984196], dtype=float32)
expect = array([-13.676786 ,  -3.9301395,  -7.5552464, ...,  -3.3748627,
        -4.0334225,  -4.0430546], dtype=float32)
delta = 0.001

    def check_data(self, result, expect, delta):
        """
        check result
        Args:
            result(list): list of result data
            expect(list): list of expect data
            delta(float): e.g. 0.001
        Returns:
            None
        """
        logger.info("current comparison delta is : {0}".format(delta))
        assert len(result) == pytest.approx(len(expect)), "output length not equal"
        for i in range(0, len(expect)):
>           assert result[i] == pytest.approx(expect[i], rel=None, abs=delta), "output data not equal"
E           AssertionError: output data not equal
E           assert -16.915035 == -13.676786422729492 ± 1.0e-03
E             comparison failed
E             Obtained: -16.915035247802734
E             Expected: -13.676786422729492 ± 1.0e-03

test_mkldnn_helper.py:60: AssertionError
```
**3. The problem recurrence steps：**
```
git clone https://github.com/EmmonsCurse/continuous_integration.git -b rename_fluid_to_base --depth=1

[...]

PR #58181, opened 2023-10-18
[OneDNN] Fix accuracy issue of deconv on specific platforms

LLee233:
### PR types
Bug fixes
### PR changes
Others
### Description
This issue is reported by https://github.com/PaddlePaddle/Paddle/issues/58050. It won't happen on every platform, but there is possibility causing wrong blocking on specific platforms like `Intel(R) Xeon(R) Gold 6271C` as the issue says, due to cache size limitation. 

Thus here we update OneDNN with relative commit updated towards fixing this issue.


risemeup1 (approved):

tianshuo78520a (approved):
LGTM

PR #58192, opened 2023-10-18
enable flash_attn selection for mlu.

ShawnNew:
### PR types

Function optimization

### PR changes

Others

### Description

add flash_attn selection for mlu backend


ronny1996 (approved):
LGTM

PR #58195, opened 2023-10-18
[Compile]Fix utils library cycle depend problem

Aurelius84:
### PR types

Bug fixes
### PR changes

Others
### Description

[Compile]Fix utils library cycle depend problem

Pcard-67164


xiaoguoguo626807 (approved):
LGTM

PR #58201, opened 2023-10-18
【Add memory info】Add LogDeviceMemoryStats when op running

heavyrain-lzy:
### PR types

Function optimization
### PR changes

OPs
### Description

PCard-71568
In order to obtain the deal information of used memory on GPU, we add an auxiliary function named `LogDeviceMemoryStats`, while you don't need to call it in person. If you want to know the used memory information, you can easily set the environment variable `FLAGS_log_memory_stats` is `True`, such as 'export FLAGS_log_memory_stats=True', or disenable it by setting it `False`. The environment variable `FLAGS_log_memory_stats` defaults to `False`.
In the case of enabling the environment variable, you will see the memory log in screen:
image


From00 (approved):
LGTM

zhiqiu (approved):
LGTM

chenwhql (approved):

phlrain (approved):

PR #58216, opened 2023-10-19
[PIR+CINN]Support construct block first for GroupOp

Aurelius84:
### PR types

New features
### PR changes

OPs
### Description

[PIR+CINN]Support construct block first for GroupOp
Pcard-67164


winter-wang (approved):
LGTM

PR #58242, opened 2023-10-19
【AutoParallelism】Update gradient merge strategy in auto-parallel

heavyrain-lzy:
### PR types

Performance optimization
### PR changes

Others
### Description

PCard-71568
Update the strategy of merging gradients in auto parallelism to decrease the peak used memory.
The origin time to accumulate the local gradient of current miciro-batch onto the global gradient is that after having gotten all the gradients of variables. This may increase the peak memory because of delaying the time to free the gradients.
In the new method, we accumulate the the local gradient of current miciro-batch onto the global gradient as soon as we obtain that gradient, then we can free it.
The detailed experimental datas will then be given.
The result of test is as follow:
image


zhiqiu (approved):
LGTM

PR #58283, opened 2023-10-20
[PIR] Fix bug when build scope

kangguangli:
### PR types

Bug fixes
### PR changes

Others
### Description

fix bug when build scope:
1. delete existed vars when creating var for `ShadowOutputOp`
2. using address of InterpreterCore as the prefix of vars leads to conflicts cause different InterpreterCore may have same address. 
3. when processing while op, the output value mapping are skipped before. This PR fixes this.
4. partial program will switch different scope cache by FLAGS_enable_new_ir_in_executor.
#### Others
Pcard-67164


zhangbo9674 (approved):

zhiqiu (approved):
LGTM for const_cast

PR #58358, opened 2023-10-24
fix_batchnorm_set_default_value

xuxinyi389:
### PR types
Bug fixes
### PR changes
APIs 
### Description
Pcard-73263
现batch_norm已支持optional https://github.com/PaddlePaddle/Paddle/pull/58226 ，为nn.functional.batch_norm的默认值增加默认值。中文文档修改：https://github.com/PaddlePaddle/docs/pull/6248


python/paddle/nn/functional/norm.py
```diff
@@ -116,8 +116,8 @@ def batch_norm(
     x,
     running_mean,
     running_var,
-    weight,
-    bias,
+    weight=None,
+    bias=None,
```
jeff41404, line 120: the following document description also needs to be modified, eg: `weight(Tensor): The weight tensor of batch_norm, can not be None. bias(Tensor): The bias tensor of batch_norm can not be None.` as well as [chinese document](https://github.com/PaddlePaddle/docs/blob/develop/docs/api/paddle/nn/functional/batch_norm_cn.rst) of `batch_norm`

jeff41404: Meanwhile, because these two parameters(`weight`, `bias`) in the OP definition have not default values, they are either passed to the OP after initialization(please pay attention to the initialization algorithm) in the Python layer or modified the OP definition(Not recommended)

jeff41404: if we allow these two parameters have default values, then should add some test case to make sure that behavior of API meets expectations

jeff41404 (approved):
LGTM

sunzhongkai588 (approved):
LGTM for docs

zhwesky2010 (approved):

test/legacy_test/test_batch_norm_op.py
```diff
@@ -375,6 +375,77 @@ def test_check_output(self):
     def init_kernel_type(self):
         pass
 
+    def check_without_scale_and_bias(self, place, data_layout, dtype, shape):
```
LLee233, line 378: This function seems not be called since `test_check_output` still calls `check_with_place`? And locally tested, this function will report error for creation of op...

xuxinyi389: You're right, I'll fix it.

PR #58408, opened 2023-10-26
fix inference build error

hitywt:
### PR types
Others

### PR changes
Others

### Description
Pcard-70448
1. fix inference compile error caused by cuda version judgement


ForFishes (approved):
LGTM

PR #58425, opened 2023-10-26
save static runtime info

GGBond8488:
### PR types
Others 

### PR changes
 Others

### Description
Pcard-78525

**What did this PR do**?

it add two flags:
`FLAGS_save_static_runtime_data(bool)`: whether to save runtime tensor in static mode
`FLAGS_static_runtime_data_save_path(string)`: set the static runtime tensor save path
to control whether and where to save static runtime info

> when set `FLAGS_save_static_runtime_data: True`, When the model is run in static mode, the input and output of each op in the model graph will be saved in the `FLAGS_static_runtime_data_save_path` directory. On this basis, if you use the `cinn` compiler to run the model, the subgraph entering  `cinn` will be saved in the same directory. information, as well as the mapping of variable names on the framework side and `cinn` side

**Why add these two flags?**
1. Analysis accuracy problem of tensor granularity
> When we want to save some tensor data when running static graphs, we often only use static_print or save_op. Both of these are very inconvenient and have great limitations, so we often limit ourselves to layer-based analysis and comparison. Debug accuracy issues. So we added these two flags to analyze accuracy issues with tensor as the granularity.
2. We will add a CINN accuracy analysis tool later, relying on this

**What impact will this have?**
1. When `FLAGS_save_static_runtime_data` is false, it will not have any impact on the framework. It is not turned on by default.
2. When `FLAGS_save_static_runtime_data` is true, it may have an impact on the performance of the static  mode of the framework. It will mainly increase the memcpy and file io for the inputs and outputs of each op. For models with larger tensor sizes, saving this data may occupy a relatively large amount of disk space.

**How to use these two flags?**
1. single use
set `FLAGS_save_static_runtime_data: True` ，and run model in static mode,
During the running process, the input and output of each op will be saved in the `FLAGS_static_runtime_data_save_path` directory. In particular, when running with the `cinn` compiler, additional var and graph information entering `cinn` will be added, as well as variable mapping.
Then you can use `paddle.core.load_dense_tensor` to read the saved tensor, and you can wrap it with `paddle.Tensor` and use it directly in the python layer
2. use with autodiff tools(Not online)
This tool requires it to automatically align CINN and frame-side accuracy

**Why is there no unitest added?**
1. It is a debug flag and will not affect normal operation.
2. If I want to add a unitest to it, then I need to turn on flags, run a program, and then turn off flags.
Because our unitest is parallel at the unitest file level,
Therefore, when this flags is turned on, it may pollute the running of other unitests, including the generation of unnecessary storage files, and some other warnings or error messages that do not support cinn.


jeff41404:
please add some explanation in the description, including the scenario in which this feature is used, its impact on existing systems, and a simple example code on how to use it

jeff41404:
and explain why it is not possible to add a unit test, and what will be the impact of not adding a unit test.

jeff41404 (dismissed):
LGTM

luotao1 (dismissed):

paddle/fluid/framework/new_executor/interpreter/interpreter_util.cc
```diff
@@ -963,6 +966,44 @@ void BuildOpFuncList(const platform::Place& place,
       std::rethrow_exception(std::current_exception());
     }
 
+    if (FLAGS_save_static_runtime_data) {
```
zhangbo9674, line 969: 对于旧 Program 执行器，实现逻辑如果放在 BuildOpFuncList 中，将仅会 save 第1轮的中间结果，其余轮次的执行见 program_interpreter.cc的RunImpl函数。
此外，随着新 IR 推进，旧 Program的执行器也会在将来废弃，该逻辑需要迁移到 pir_interpterter.cc 中，实现逻辑可考虑加在：PirInterpreter::Run

GGBond8488: 好的，旧执行器的逻辑迁移到RunImpl中，新ir的执行器等后续补上

luotao1 (approved):

paddle/fluid/framework/new_executor/program_interpreter.cc
```diff
@@ -1053,6 +1055,45 @@ void ProgramInterpreter::RunOperator(const Instruction& instr_node) {
     }
   }
 
+  // for debug
+  if (FLAGS_save_static_runtime_data) {
```
zhangbo9674, line 1059: 对于 GPU 异步训练的场景，这里如果没有同步操作，直接对输出进行访问可能是有风险的吧？

GGBond8488: 这个异步是指？如果是多机多卡场景，因为CINN现在只支持单机单卡，等到CINN也支持GPU异步场景的时候，后续也会去支持。如果是指stream，这个save里面会有一个memcpy的操作，会先将数据从GPU拷贝到cpu，再执行保存，这个memcpy是有同步操作的

zhangbo9674 (approved):

PR #58477, opened 2023-10-30
update decompose_bwd_op implementation

lxd-cumt:
### PR types

Others
### PR changes

Others
### Description

Pcard-66975
Previous PR: https://github.com/PaddlePaddle/Paddle/pull/58066
- **Seperate decompose backward op logic to two parts: decompose_bwd_op_directly() and decompose_bwd_op_after_fwd_op().**
（1）If forward op has composite vjp rules (including custom vjp), use decompose_backward_op_dierctly() API, call call_vjp() to get a list of primitive operators in backward graph instead of grad() to prevent recursively decomposing other ops. (An example of add op is as follows.)
（2）If forward op has no composite vjp rules, call decompose_fwd_op() to decompose forward op firstly, then use decompose_backward_op_after_fwd_op(), which calls grad() to get a list of primitive operators in backward graph, then replace bwd_op. 
- **Test more ops.** 
Add rsqrt op, mean op, add op where the two input variable have the same parent variable (In this case, when decomposing add_grad by using grad() API, will incorrectly decomposing all the ops on the path from the op corresponding to one of the input variables of add op to add_op, which is not desired)

截屏2023-10-30 11 26 30

- **change return of translate_to_new_ir_with_param_map() from pir::Value to pir::OpResult**


kangguangli (dismissed):
LGTM for `ProgramTranslator::VarDesc2OpResult`

cxxly (approved):
LGTM for decomp

kangguangli (approved):

PR #58520, opened 2023-10-31
Support weights to be temporary variable in TRT NHWC IR pass

leo0519:
### PR types

Bug fixes

### PR changes

Others

### Description

The TRT NHWC IR pass determines an input is weight or not by checking its attribute `Persistable`.
However, after TRT 8.6, the weights of weighted OPs could be a temporary variable.
This PR adds a map of (op_type, weight_name) (ex. ("conv2d", "Filter")) and check the input is weight or not according to this map.


yuanlehome (approved):
LGTM

PR #58566, opened 2023-11-01
Fix test_trt_convert_swish

Tom-Zheng:
### PR types

Bug fixes
### PR changes

Others
### Description

In the implementation of Swish TRT plugin, there are no constraints on tensor format. It allows the freedom for TRT engine to cast the input to some padded format like `kHWC8`. However, the kernel is not designed to deal with this kind of padded tensor format, which will cause incorrect results. 

Therefore, `test_trt_convert_swish` fails randomly (very rarely, but do fails sometimes), with the following message:
```
Tue Oct 31 16:04:51-ERROR: FAIL: program_config: swish{'beta': 2.0} -- [input_data: {'shape': (1, 3, 64, 64), 'lod': None, 'dtype': dtype('float32')}][place_holder_weight: {'shape': (1,), 'lod': None, 'dtype': dtype('float32')}]['input_type': <class 'numpy.float16'>]
Tue Oct 31 16:04:51-ERROR: FAIL: predictor_config: {'use_trt': True, 'trt_precision': <Precision.Half: 2>, 'use_dynamic_shape': False}
Tue Oct 31 16:04:51-ERROR: FAIL:  ERROR INFO: 
Not equal to tolerance rtol=0.001, atol=0.001
Mismatched elements: 2048 / 12288 (16.7%)
Max absolute difference: 0.880797
Max relative difference: 1.
 x: array([[[[0.8804, 0.8804, 0.8804, ..., 0.8804, 0.8804, 0.8804],
         [0.8804, 0.8804, 0.8804, ..., 0.8804, 0.8804, 0.8804],
         [0.8804, 0.8804, 0.8804, ..., 0.8804, 0.8804, 0.8804],...
 y: array([[[[0.880797, 0.880797, 0.880797, ..., 0.880797, 0.880797,
          0.880797],
         [0.880797, 0.880797, 0.880797, ..., 0.880797, 0.880797,...
```
If you print out the TRT engine info, you will find input is cast to `kHWC8` before swish op, and then cast back to `kLINEAR` after swish in such failure case. Some of the output values become zeros, as the indexing of swish kernel is linear.

The solution is to add format check and only allows kLINEAR input.

### DIFF
--- a/paddle/fluid/inference/tensorrt/plugin/swish_op_plugin.cu
+++ b/paddle/fluid/inference/tensorrt/plugin/swish_op_plugin.cu
@@ -31,10 +31,12 @@ void SwishPlugin::terminate() TRT_NOEXCEPT {}
 bool SwishPlugin::supportsFormat(
     nvinfer1::DataType type, nvinfer1::PluginFormat format) const TRT_NOEXCEPT {
   if (with_fp16_) {
-    return type == nvinfer1::DataType::kFLOAT ||
-           type == nvinfer1::DataType::kHALF;
+    return (type == nvinfer1::DataType::kFLOAT ||
+            type == nvinfer1::DataType::kHALF) &&
+           (format == nvinfer1::TensorFormat::kLINEAR);
   }
-  return type == nvinfer1::DataType::kFLOAT;
+  return (type == nvinfer1::DataType::kFLOAT) &&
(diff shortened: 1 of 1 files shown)


Tom-Zheng:
@Wangzheee  Would you please review it?

Wangzheee (approved):
LGTM

PR #55823, opened 2023-07-31
Set Input Type for TensorRT convert tests

eee4017:
### PR types
Others

### PR changes
Others

### Description
We preprocess the test_trt_convert test to standardize the input type to either FP32 or FP16, ensuring a uniform input type.

PR #58569, opened 2023-11-01
Enable FP16 IO in TRT convert tests

Tom-Zheng:
### PR types

Others
### PR changes

Others
### Description

In most TRT convert tests, both Float and Half precisions will be tested. However, the input and output tensors are float tensors for both tests. This causes two problems for Half precision tests:
1. When input tensor is float, TRT engine needs to do casting before / after op. When TRT engine finetuning, it may find that,  "_float input -> cast to fp16 -> fp16 op -> cast back to float -> float output_" is slower than "float input -> float op -> float output". Therefore, even if precision is Half, FP16 kernel is not tested. This harms test coverage.
2. If the two routes mentioned above have similar running time, then the test could be non-deterministic, causing random numerical failures.

This PR enables FP16 inputs for Half precision tests. There was an attempt to achieve this in #55823 , but due to a bug, it did not actually take effect. This PR fixes the problem.

After change the input to FP16, some tests fail. Detailed explanation to changes per test:
[TRT convert tests.xlsx](https://github.com/PaddlePaddle/Paddle/files/13224937/TRT.convert.tests.xlsx)

### DIFF
--- a/paddle/fluid/inference/tensorrt/plugin/generic_plugin.cu
+++ b/paddle/fluid/inference/tensorrt/plugin/generic_plugin.cu
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "paddle/fluid/inference/tensorrt/plugin/generic_plugin.h"
+#include <map>
+
 #include "paddle/fluid/framework/framework.pb.h"
 #include "paddle/fluid/framework/op_kernel_type.h"
 #include "paddle/fluid/framework/phi_utils.h"
 #include "paddle/fluid/inference/tensorrt/dynamic_shape_infermeta_registry.h"
+#include "paddle/fluid/inference/tensorrt/plugin/generic_plugin.h"
 #include "paddle/phi/backends/gpu/gpu_context.h"
 #include "paddle/phi/common/data_type.h"
 #include "paddle/phi/core/compat/op_utils.h"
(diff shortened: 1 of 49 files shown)


Tom-Zheng:
@onecatcn  UT设计注意事项 #58577

Tom-Zheng:
@Wangzheee Please review this PR.

paddle/fluid/inference/tensorrt/plugin/generic_plugin.cu
```diff
@@ -529,26 +525,17 @@ int GenericPlugin::enqueue(const nvinfer1::PluginTensorDesc* input_desc,
...
-    } else if (proto_type == GeneratePluginDataType::PLUGIN_INT32) {
-      return {phi::DataType::INT32, sizeof(int32_t)};
-    } else if (proto_type == GeneratePluginDataType::PLUGIN_BOOL) {
-      return {phi::DataType::BOOL, sizeof(bool)};
-    } else {
-      CHECK(false) << "precision is not supported";
-    }
+  auto nvType2PhiType =
+      [&](nvinfer1::DataType nv_dtype) -> std::pair<phi::DataType, int> {
+    const std::map<nvinfer1::DataType, std::pair<phi::DataType, int>> _map{
```
vivienfanghuagood, line 530: why remove the mapping of phi::DataType::INT64？

Tom-Zheng: TRT does not natively support INT64 type, see [[doc](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#a83aed11a1c160f30dcd13809678bdd29)] . If Paddle dtype is `phi::DataType::INT64`, it will be internally downcast to INT32, before fed into TRT plugin.

Wangzheee (approved):

Tom-Zheng:
@raindrops2sea or @XiaoguangHu01 would you please approve for changing 20+ files?

XiaoguangHu01 (approved):
LGTM

Tom-Zheng:
@Wangzheee Would you please merge it?

PR #58602, opened 2023-11-02
[CINN] Add Expr<Dim> to describe the shape of Tensor

ZzSean:
### PR types

Others
### PR changes

Others
### Description

[CINN] Add Expr<Dim> to describe the shape of Tensor and to support dynamic shape.

In order to pass CI, CINN does not use `pir::shape::SymbolicDimOp`  for now.
When `pir` is completely integrated, the fake class will be replaced by `pir::shape::SymbolicDimOp`.

card-77050


paddle/cinn/ir/dim.h
```diff
@@ -0,0 +1,63 @@
...
+  std::string name;
+  pir::shape::SymbolicDimOp sym_dim;
+
+  const pir::shape::SymbolicDimOp GetSymbolicDim() const;
+
+  bool IsDynamic() const;
+
+  std::string GetSymbolName() const;
+
+  int64_t GetRealDimSize() const;
```
zhhsplendid, line 62: What does it return when it is unknown?

ZzSean: the default value
`static constexpr int64_t kDynamic = std::numeric_limits<int64_t>::min();`

zhhsplendid (approved):
LGTM

PR #58625, opened 2023-11-02
[PIR] separate translation context for body block of while op

kangguangli:
### PR types

Bug fixes
### PR changes

Others
### Description

separate translation context for body block of while op

#### Others
Pcard-67164


winter-wang (approved):
LGTM

PR #58667, opened 2023-11-03
[fixbugs] cached property is not support in 3.7, replace with class

2742195759:
### PR types

Others

### PR changes

Others

### Description

cached property is not support in 3.7, relace with class

PCard-66972


SigureMo (approved):
LGTMeow

PR #58705, opened 2023-11-06
[PIR] Translate dtype for data op

kangguangli:
### PR types

Others
### PR changes

Others
### Description

Translate dtype for data op instead of assigning a fixed value

#### Others
Pcard-67164


kangguangli:
Done in #58666

PR #58776, opened 2023-11-07
[AutoParallel]Revise Infermeta of LayerNorm for Sequence-Data Hybrid Parallelism

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Description
Pcard-76459

The Current LayerNorm Implement would flatten the broadcast axes (axis before "begin_norm_axis") of LayerNorm, which would hinder broadcast axes to be sharded by different Mesh dimensions.

In Sequence-Data Hybrid Parallelism, we need to shard both batch and sequence axes (broadcast axes) of LayerNorm input to get the best Performance. 

Therefore we plan to remove the "flatten logic" in LayerNorm Op.

Before PR:
all axes before 『begin_norm_axis』will be flatten into one axis for  var and std of LN.

After PR:
the shape of  mean and var will keep the same as the shape of input 『begin_norm_axis』.

```python
# pseudocode
in = paddle.randn(shape = [8, 1024, 768])
var, mean, out = layer_norm(in, begin_norm_axis=2)

# before PR
out.shape == [8, 1024, 768]
var.shape == [8192]
mean.shape == [8192]

# after PR
out.shape == [8, 1024, 768]
var.shape == [8, 1024]
mean.shape == [8, 1024]

```

The Logic modified in this PR:

- Layernorm InferMeta
- Layernorm InferSpmd
- Layernorm Prim composite rule
- Layernorm CPU kernel
- Layernorm XPU & MLKDNN Unitest

### DIFF
--- a/paddle/phi/infermeta/ternary.cc
+++ b/paddle/phi/infermeta/ternary.cc
@@ -579,7 +579,10 @@ void LayerNormInferMeta(const MetaTensor& x,
           x_dim.size()));
 
   auto matrix_dim = phi::flatten_to_2d(x_dim, begin_norm_axis);
-  int left = static_cast<int>(matrix_dim[0]);
+
+  // keep the axis size before normalization for shape of variance and mean
+  auto before_norm_dims = slice_ddim(x_dim, 0, begin_norm_axis);
+  // int left = static_cast<int>(matrix_dim[0]);
   int right = static_cast<int>(matrix_dim[1]);
   if (scale) {
     PADDLE_ENFORCE_EQ(scale.dims().size(),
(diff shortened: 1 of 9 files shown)


paddle/phi/infermeta/ternary.cc
```diff
@@ -644,11 +647,11 @@ void LayerNormInferMeta(const MetaTensor& x,
           ? phi::DataType::FLOAT32
           : x_dtype;
   if (mean) {
-    mean->set_dims({left});
+    mean->set_dims({before_norm_dims});
     mean->set_dtype(param_type);
   }
   if (variance) {
-    variance->set_dims({left});
+    variance->set_dims({before_norm_dims});
```
liuzhenhai93, line 654: kernel 调用前后 似乎也要改一下？

pkuzyc (commented):
Are there some corresponding cases in unit test? If so, unit test should also be revised.

JZ-LIANG:
> Are there some corresponding cases in unit test? If so, unit test should also be revised.

xiaoguoguo626807 (dismissed):
LGTM for composite rules

zhiqiu (commented):
LGTM

chenwhql (approved):
LGTM

pkuzyc (commented):
LGTM for spmd rules

xiaoguoguo626807 (approved):

PR #58859, opened 2023-11-09
Enhanced RNG State Management with Index-Based Control for Graph-Safe Tensor Parallelism

eee4017:
### PR types
New features

### PR changes
APIs

### Description

This Pull Request presents a refined approach to managing multiple RNG states within the Generator class. The primary goal is to facilitate a more robust control mechanism over RNG states to support advanced computational scenarios such as CUDA graphs and tensor parallelism.  We introduced an index-based state management system, extended the prior method of direct state manipulation. This method registers states and controls them through indices, ensuring graph-safe RNG operations.

#### Background

Current RNG state management, especially in the context of CUDA graphs, faces challenges with updating the RNG state(a combination of seed and offset). The offset increments with each use, but its update within CUDA graphs is problematic because it is a host state and not captured in the CUDA graph's operation sequence. This can be demonstrated by the following code where the offset increment is not recorded in the CUDA graph:

```c++
int offset = 0; 

// cudaGraph begin
Kernel0<<<...>>>(param0, param1, param2, …)
offset = offset + 1; // This is not recorded 
Kernel1<<<...>>>(param0, offset, param2, …)
Kernel2<<<...>>>(param0, param1, param2, …)
offset = offset + 1; // This is not recorded 
Kernel3<<<...>>>(offset, param1, param2, …)
// cudaGraph end
```

As proposed in #58310 , PaddlePaddle's solution involves wrapping offset updates within callback functions that are invoked before replaying the CUDA graph. This manually recalculates and updates the offset for each node, as shown here:

Here's an explanation of how this mechanism is implemented through code:

```c++
int offset;

int *offset_ptr = &offset;

// This callback function increments the RNG state and returns the updated value.
auto callback = [offset_ptr](){
	*offset_ptr = *offset_ptr + 1;
	return *offset_ptr;
};

rng_state = callback();
cuGraphKernelNodeSetParams(Kernel1_node, rng_state);
rng_state = callback();
cuGraphKernelNodeSetParams(Kernel3_node, rng_state);
cudaGraph.replay() // Kernel0, Kernel1, Kernel2, Kernel3
```

#### Challenges in TP/PP

In Tensor Parallelism (TP) and Pipeline Parallelism (PP), RNG state exchanges are necessary, involving saving and restoring states to Python. This presents difficulties as the state changes in CUDA graphs cannot be captured with callback, as illustrated below:

```c++
int offset;

// cudaGraph begin

// RNGStatesTracker.__enter__
self.states_[name] = paddle.get_rng_state() // save current rng state to a python dict
paddle.set_rng_state(...) // set new random state, offset is set to new value

offset = offset + 1;
Kernel1<<<...>>>(param0, offset, param2, …)

// RNGStatesTracker.__exit__
paddle.set_rng_state(...) // restore old random state

// cudaGraph end
```

Also, PyTorch also faces same limitations with set_rng_state/get_rng_state in TP contexts under CUDA Graph.

#### Solution

To address these issues, we propose new APIs in this PR that allow RNG state management via indices, rather than direct state manipulation. This enables capturing the state behavior during CUDA graph operations. Importantly, while we are advancing the system with these new functionalities, the pre-existing `set_state` and `get_state` APIs will be remained to ensure backward compatibility.

+ C++ Generator API:
  - `GetStateIndex`/`SetStateIndex` to switch between RNG states.
  - `RegisterStateIndex` for registering new RNG states and switching to them.
  - Refactoring of the generator's constructor and methods for better readability.
  - Dropout layer updates for graph-safe tensor parallelism using the correct RNG state index.
+ Python API:
  - `register_rng_state_as_index` for encapsulating and indexing the current RNG state.
  - `get_rng_state`/`set_rng_state` modified to include an use_index flag for index-based control.
  - Modification of the `RNGStatesTracker` to accommodate the new index-based system, ensuring compatibility with tensor parallelism and CUDA graphs.

With these APIs, we can maintain the state_index of the generator, allowing for graph-safe manipulation and state tracking captured in callback functions:

```c++
uint64_t state_index = GetStateIndex(); 
[state_index](){ 
	gen->SetStateIndex(state_index)
	// call generator to increment offset
}
```


tianshuo78520a (approved):

test/legacy_test/test_random_generator_set_get_state.py
```diff
@@ -0,0 +1,88 @@
...
+class TestRandomGeneratorSetGetState(unittest.TestCase):
+    def test_random_generator_set_get_state(self):
+        """Test Generator Get/Set state with Index."""
+        paddle.seed(102)
+        gen = get_default_generator()
+        orig_state = gen.get_state()
+
+        x = generate_random_number_and_states(gen)
+
+        assert_allclose = lambda x, y: np.testing.assert_allclose(
```
zhiqiu: Why not array_equal?

eee4017: Referenced from [test_cuda_random_seed](https://github.com/PaddlePaddle/Paddle/blob/develop/test/legacy_test/test_cuda_random_seed.py), I believe we should use array_allcose in this context as it involves a float.

zhiqiu: I checked the first case `test_gen_dropout_dygraph` in [test_cuda_random_seed](https://github.com/PaddlePaddle/Paddle/blob/develop/test/legacy_test/test_cuda_random_seed.py), it is also ok if the `allclose` is changed to `equal`.

eee4017: Changed into  assert_array_equal

eee4017:
Hello, I require aprroval for change from one of the following individuals:

RD approval for API change: @XiaoguangHu01 , @jeff41404 , @lanxianghit or @qingqing01
TPM approval for API change and API documents change: @jzhang533 (ZhangJun), @sunzhongkai588 (SunZhongKai), @Ligoml  (LiMengLiu)

jzhang533:
@eee4017 LGTM for this API naming and semantic, I noticed that you have suggested a similar feature in the PyTorch community: https://github.com/pytorch/pytorch/issues/113541.

Perhaps we should wait for feedback from the PyTorch community. If the proposed API is compatible, it could greatly benefit both community.

If this PR is blocking our ongoing project, please let me know, as a temporary solution, we can merge this PR now and make adjustments to the API naming and semantics later, before our next release.

eee4017:
Thank you for reviewing the proposed API change. If this alteration is acceptable, I recommend proceeding with the merger of this PR. This action would simplify our process by minimizing the need for extensive communication and the time spent on re-running CI tests. It's important to note that the implementation of the RNG in PyTorch differs from that in Paddle. Consequently, that issue may require a thorough review and discussion. The timeframe for that is uncertain. I will keep you updated with any progress or new insights from the PyTorch community. In the meantime, moving forward with the current PR would be a practical step to maintain momentum in our project. Thank you.

jzhang533 (dismissed):
LGTM

python/paddle/__init__.py
```diff
@@ -811,6 +812,7 @@
     'unique_consecutive',
     'set_cuda_rng_state',
     'set_rng_state',
+    'register_rng_state_as_index',
```
jeff41404: `register_rng_state_as_index` should be public API?

eee4017: move these APIs to incubate

risemeup1 (dismissed):

risemeup1 (approved):

zyfncg (approved):

paddle/phi/core/generator.cc
```diff
@@ -157,134 +158,125 @@ const std::shared_ptr<Generator>& GetRandomSeedGenerator(
...
-  auto engine = std::make_shared<std::mt19937_64>(seq);
-  this->state_.cpu_engine = *engine;
-  this->state_.device = -1;
-  this->state_.current_seed = seed;
-  this->state_.thread_offset = 0;
-  this->engine_ = engine;
-  VLOG(4) << "initial seed: " << this->state_.current_seed
-          << ", cpu engine: " << &this->state_.cpu_engine;
+  current_index = states_.size();
+  states_.emplace_back(seed);
```
sneaxiy, line 189: Should this be `states_.emplace_back(-1, seed)`?

paddle/phi/core/generator.h
```diff
@@ -14,64 +14,109 @@ limitations under the License. */
...
+    }
+
+    void reset(uint64_t new_seed = MAGIC_RANDOM_SEED) {
+      std::seed_seq seq({new_seed});
+      cpu_engine->seed(seq);
+      offset = 0;
+      seed = new_seed;
+    }
+
+    GeneratorState clone() const {
```
sneaxiy, line 55: Why we need a clone function? Why cannot we just write the deep-copy constructor?

paddle/phi/core/generator.h
```diff
@@ -14,64 +14,109 @@ limitations under the License. */
...
  public:
   struct GeneratorState {
-    int64_t device = -1;
-    uint64_t current_seed = 34342423252;
-    uint64_t thread_offset = 0;
-    std::mt19937_64 cpu_engine;
+    int64_t device;
+    uint64_t seed;
+    uint64_t offset;
+    std::shared_ptr<std::mt19937_64> cpu_engine;
```
sneaxiy, line 38: Why we need `std::shared_ptr<std::mt19937_64>` instead of `std::mt19937_64`?

paddle/phi/core/generator.cc
```diff
@@ -157,134 +158,125 @@ const std::shared_ptr<Generator>& GetRandomSeedGenerator(
...
+  if (current_index < states_.size())
+    current_index = StateIndex;
+  else
+    PADDLE_THROW(phi::errors::NotFound("Generator index is not found"));
+}
+
+uint64_t Generator::RegisterStateIndex(const GeneratorState& state) {
+  std::lock_guard<std::mutex> lock(mu_);
+  auto new_index = states_.size();
+  states_.push_back(state);
```
sneaxiy, line 230: Why don't we use `states_.push_back(state.clone())` here?

paddle/phi/core/generator.cc
```diff
@@ -157,134 +158,125 @@ const std::shared_ptr<Generator>& GetRandomSeedGenerator(
...
-  std::random_device de;
-  seed = ((((uint64_t)de()) << 32) + de()) & 0x1FFFFFFFFFFFFF;
-  this->state_.current_seed = seed;
-  std::seed_seq seq({seed});
-  this->engine_->seed(seq);
-
-  return this->state_.current_seed;
+  std::lock_guard<std::mutex> lock(mu_);
+  uint64_t seed = GetRandomSeed();
+  state().reset(seed);
```
sneaxiy, line 254: The code logic here is different from the original ones. The original one does not reset `this.state_.offset` but the new one reset `state().offset`.

python/paddle/distributed/fleet/layers/mpu/random.py
```diff
@@ -51,10 +51,11 @@ def add(self, name, seed):
         self.seeds_.add(seed)
         if name in self.states_:
             raise ValueError(f'state {name} already exists')
-        orig_rng_state = paddle.get_rng_state()
+        orig_rng_state_index = paddle.incubate.get_rng_state(use_index=True)
+        # register a new state and set that state with the seed, store the indices into states_
+        self.states_[name] = paddle.incubate.register_rng_state_as_index()
         paddle.seed(seed)
```
sneaxiy, line 57: Do we reorder `paddle.seed()` and `self.states_[name] = ...` in the wrong order here?

paddle/phi/core/generator.cc
eee4017: The reset of offset in the new implementation is an intentional change. When seed() is invoked, it implies the generation of a new random value for the seed and setting the current seed to this random value. Consequently, every time a seed is set, it is logical and necessary to reset the value of offset as well. This approach aligns with [`SetCurrentSeed`](https://github.com/PaddlePaddle/Paddle/blob/v2.6.1/paddle/phi/core/generator.cc#L255) in our repository and the implementation in PyTorch [here](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/CUDAGeneratorImpl.cpp#L163), where the offset is reset whenever a new seed is set.

paddle/phi/core/generator.h
eee4017: Yes, we can use copy constructor.

paddle/phi/core/generator.h
eee4017: This design choice is highlighted in the original code, which can be reviewed [here](https://github.com/PaddlePaddle/Paddle/blob/v2.6.1/paddle/phi/core/generator.cc#L180C1-L183C37). The code snippet demonstrates an interesting pattern. 

```
auto seed = GetRandomSeed();
std::seed_seq seq({seed});
auto engine = std::make_shared<std::mt19937_64>(seq);
this->state_.cpu_engine = *engine;
this->engine_ = engine;
```

An `mt19937_64` engine is allocated on the heap, and `GeneratorState` retains a copy of this initial state. However, this copy is subsequently unused. Instead, `Generator` has a `shared_ptr` pointing to the `mt19937_64` engine on the heap, and this pointer is used for engine manipulation, as seen in [`GetCPUEngine`](https://github.com/PaddlePaddle/Paddle/blob/v2.6.1/paddle/phi/core/generator.cc#L263) and [`Random64`](https://github.com/PaddlePaddle/Paddle/blob/v2.6.1/paddle/phi/core/generator.cc#L273).

The original code seems to favor the `mt19937_64` engine on the heap managed by the shared_ptr rather than the engine held by `GeneratorState`. If `GeneratorState` were to own the `mt19937_64` engine directly, a shared_ptr would not be necessary since the engine's lifetime would align with the state's lifetime. In this case, a normal pointer would suffice.

I have considered letting `GeneratorState` directly own the `mt19937_64` engine. However, this would entail changing the `GetCPUEngine` API to a normal pointer, potentially impacting numerous files that utilize this API.

python/paddle/distributed/fleet/layers/mpu/random.py
eee4017: No, the order of operations is correct as implemented. Here's a breakdown of the process:

Original Code:
```
orig_rng_state = paddle.get_rng_state()
paddle.seed(seed)
self.states_[name] = paddle.get_rng_state()
paddle.set_rng_state(orig_rng_state)
```
In this snippet, we first save the original RNG state. Then, we set a new seed and update `self.states_[name]` with the new RNG state. Finally, we restore the original RNG state.


Revised Code:
```
orig_rng_state_index = paddle.incubate.get_rng_state(use_index=True)
# register a new state and set that state with the seed, store the indices into states_
self.states_[name] = paddle.incubate.register_rng_state_as_index()
paddle.seed(seed)
paddle.incubate.set_rng_state(orig_rng_state_index, use_index=True)
```
In the revised code, we first obtain the index of the original RNG state. We then register a new RNG state and update `self.states_[name]` with the index of this new state. Then, we set the seed for the new state. Finally, we switch back to the old state.

paddle/phi/core/generator.cc
eee4017: Indeed, cloning would be the more appropriate approach in this context.

paddle/phi/core/generator.cc
eee4017: Yes, it should be `states_.emplace_back(-1, seed)`

eee4017: Fixed in #60310.

paddle/phi/core/generator.h
eee4017: Fixed in #60310.

paddle/phi/core/generator.cc
eee4017: Fixed in #60310.

PR #58868, opened 2023-11-09
Fix rms norm when dy2static

RichardWooSJTU:
### PR types

Bug fixes
### PR changes

APIs
### Description

Optinonal input should use template as follows to avoid failing in dy2static:
```
def xxx(x, y = None): # y is optional
    if y is not None:
            xxxx
```

Pcard-71502


wwbitejotunn (approved):
LGTM

MARD1NO (approved):

yuanlehome (approved):

PR #58900, opened 2023-11-10
Fix det infermeta

Vvsmile:
### PR types

Bug fixes
### PR changes

APIs
### Description

DLTP-77674
fix the det InferMeta bugs, this bug will induce the shape inference error in the static mode.
The original paddle.linalg.det in static mode will call the UnchangedInferMeta which is obviously unreasonable. This bug will induce the shape of det API produces the different output shape in static mode compared with dynamic mode.

Modify:
paddle/phi/api/yaml/ops.yaml: use the DetInferMeta
paddle/phi/infermeta/binary.cc; paddle/phi/infermeta/binary.h: add function DetInferMeta
(the shape inference is referred to the logic of Det kernel function)


paddle/phi/infermeta/binary.cc
```diff
@@ -1125,6 +1125,28 @@ void DropoutNdInferMeta(const MetaTensor& x,
...
+			phi::errors::InvalidArgument("the input maatrix dimension size should greater than 2."));
+	
+	PADDLE_ENFORCE_EQ(input_dim[input_dim_size - 1],
+			input_dim[input_dim_size - 2],
+			phi::errors::InvalidArgument("the input matrix should be square matrix."));
+
+	auto output_dims = phi::slice_ddim(x.dims(), 0, input_dim_size -2);
+	
+	if (input_dim_size > 2) {
+	        out->set_dims(output_dims);
```
ZzSean: 缩进

paddle/phi/infermeta/binary.cc
```diff
@@ -1125,6 +1125,28 @@ void DropoutNdInferMeta(const MetaTensor& x,
...
+ 	} else {
+		out->set_dims(phi::make_ddim({}));
```
ZzSean: 如果都set_dims成output_dims会有什么问题吗？

PR #58901, opened 2023-11-10
Add FLOAT64 and FLOAT64S to CINN

Vvsmile:
### PR types

Bug fixes
### PR changes

Others
### Description

pcard-77673
Add FLOAT64 and FLOAT64S to OpAttrsToCinn in transform_desc.cc.
This bug will cause an error when model with double or vector<double> as an argument in CINN. So we should add this two dtype to OpAttrsToCinn and OpAttrsFromCinn.


6clc (approved):
LGTM

PR #58913, opened 2023-11-10
[Semi-Auto] Adapt cross_entropy spmd rule to phi

pkuzyc:
### PR types
Others

### PR changes
Others

### Description
Pcard-76459
Adapt cross_entropy spmd rule to phi. This is also the first part of supporting parallel cross entropy in static semi-auto parallel


JZ-LIANG (approved):
LGTM

PR #59004, opened 2023-11-14
[BugFix] fix speed problem: don't fill zero when stop gradient==True

2742195759:
### PR types

Bug fixes

### PR changes

Others 

### Description

fix speed problem: don't fill zero when stop gradient==True

PCard-66972


SigureMo (approved):
LGTMeow

PR #59018, opened 2023-11-15
[PIR]Open grad check for more ops

changeyoung98:
### PR types
Others

### PR changes
Others

### Description
pcard-67164
open pir grad check test for some ops that passed without problems
ops: 
- assign
- cast
- concat
- expand_v2
- flip
- sign
- tile
- transpose
- slice


xiaoguoguo626807 (approved):
LGTM

PR #59054, opened 2023-11-16
[Prim][PIR] Sink forward prim rule of batch_norm

cyber-pioneer:
### PR types
New features

### PR changes
Others

### Description
Pcard-66975
1. Sink forward rule of batch_norm in PIR.
2. Fix attr type error in codegen:
before:
const str data_layout =
      op->attribute("data_layout").dyn_cast<pir::StrAttribute>().data();
after:
const std::string& data_layout =
      op->attribute("data_layout").dyn_cast<pir::StrAttribute>().AsString();

### DIFF
--- a/python/paddle/decomposition/decomp.py
+++ b/python/paddle/decomposition/decomp.py
@@ -19,7 +19,7 @@ from paddle import pir
 from paddle.autograd import ir_backward
 from paddle.base.core import (
     call_decomp,
-    decomp_ops_list_contain_unused_output,
+    decomp_ops_contain_unused_output,
     has_decomp,
 )
 from paddle.base.libpaddle.pir import Block, Operation, Program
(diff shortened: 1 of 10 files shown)


Charles-hit (approved):

python/paddle/decomposition/decomp.py
```diff
@@ -37,22 +37,20 @@ def _build_tensor_tuple(xs):
 
 
 def _analyse_decomp_results(orig_outs, decomp_outs, op):
-    intermediate_status = op.get_output_intermediate_status()
-    assert len(orig_outs) == len(decomp_outs) == len(intermediate_status)
+    assert len(orig_outs) == len(decomp_outs)
```
Charles-hit, line 40: 最好加个msg

cyber-pioneer: ok

PR #59057, opened 2023-11-16
[Auto Parallel] Update Gradient Synchronization in Static Mode

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Description

Pcard-76459

Many Frameworks handle the gradient synchronization using different mechnism: DDP, FSDP, Dtensor, [Extra_sync_hook_for_sp](https://github.com/NVIDIA/Megatron-LM/blob/e122536b7645edcb7ebf099b5c92a443f7dbf8e7/megatron/core/distributed/finalize_model_grads.py#L70C5-L70C31) , etc.

Before this PR, the gradient synchronization mechanism in AutoParallel static mode is problematic.
It only take into account the "Batch Dimension" (hard coded as the first dimension) of input tensor and uses a hard rule to conduct the synchronization:

If the "Batch" dimension of input activation is "sharded" on rank_groupA, the gradient of parameter need to be synchronized across rank_groupA in backward phase.
The above mechanism works OK when "Batch Dimension" is the only broadcast dimension that would be sharded (narrow-sense Data Parallel).

BUT it would fail for Sequence Parallel(SP)/Context Parallel(CP) and other more general cases of "Data Parallel" where other broadcast dimensions of input tensor are "sharded". The parameter gradient synchronization need in those cases would be missing, since the framework only considers the "batch" axis but ingores other "broadcast" axes.

This PR fixed the problem using "Partial --> Replicated" mechnism.

In any case, a broadcast dimension of input tensor is sharded, the gradient of parameter generated would be in Partial status, indicates that each value of that gradient is "partial" in term to the actual value in logic view.
And before the gradient is feed to optimizer, a Reshard operation (Allreduce) would be performed to convert the status gradient tensor from Partial to Replicated.

No synchronization would be missing.

image


wanghuancoder (approved):
LGTM

heavyrain-lzy (approved):
LGTM for SP

PR #59135, opened 2023-11-19
[Prim][PIR] sqrt forward sink and fix Decomp bug

kevincheng2:
### PR types

New features
### PR changes

Others 
### Description

During the sink of sqrt, it was found that each op_name may correspond to multiple op_kernel_func,causing an error in Decomp  generation, so increasing op_kernel_func_name judgment, fix bug


cyber-pioneer (approved):
LGTM

PR #59190, opened 2023-11-21
add_n supports mixed dtype

zhiqiu:
### PR types

New features 
### PR changes

OPs
### Description

Inorder to support master grad in static graph, make add_n supports mixed dtype inputs, i.e., `add_n(fp32, fp16, fp16, ...)` or `add_n(fp32, bf16, bf16, ...)`, align with `elementwise_add`

Pcard-76459


ForFishes (approved):
LGTM

PR #59214, opened 2023-11-21
add bitwise_and to TRT ElementWise Layer

zhink:
### PR types

Performance optimization
### PR changes

OPs
### Description

Pcard-71501
add bitwise_and op convert to TRT ElementWise Layer


zhoutianzi666 (approved):
LGTM

PR #59246, opened 2023-11-22
[AutoParallel] Fix optimizer InferMeta.

GhostScreaming:
### PR types
Bug fixes

### PR changes
Others

### Description
Pcard-73145

Fix optimizer InferMeta functions. MasterParam are not set correct dims and dtype, which may cause problem for auto paralel in dynamic graph mode.


LiYuRio (approved):
LGTM

PR #58643, opened 2023-11-03
Move Dygraph Indexing Parsing to Pybind

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985 

For now, although there  there is already complete indexing (i.e. `__getitem__` / `__setitem__`) functional support, almost all the indexing parsing parts are running in Python, which brings much performance overhead.

In this PR,  we move indexing parsing of dygraph from Python to Pybind, aiming to optimize the performance (For static mode, still in Python). 

**some other things in this pr:**
- add a fast path for single tensor getting (like `x[tensor[1,2]]` ): by replace the `stack` to `unsqueeze`.
- remove non-necessary `transpose` in advanced getting.
- remove `any` check in bool-mask indexing, since the any kernel and gpu->cpu memcpy is quite slow.
- remove `cond` in bool-mask indexing, since currently this conditional block cannot be parsed in PIR

For convenience of review, the old indexing-parsing code will be deleted later.

PR #59281, opened 2023-11-23
Remove No Used Indexing Code

zoooo0820:
### PR types

Others
### PR changes

Others
### Description

Pcard-66985

Removed old indexing parsing code in Pybind and Python
- For pybind code , since new indexing parsing for setitem and getitem was introduced in PR https://github.com/PaddlePaddle/Paddle/pull/58643 , it was no longer used in dygraph , will be removed.

- For python code, it was remained before since there is no `index_put` kernel for xpu. With this kernel, we can unify the indexing code for these device, and remove the old code.
 
The unit test cases of `test_set_value_op_xpu` was modified, to keep it same to corresponding `test_set_value_op` .


jeff41404 (approved):
LGTM

Aurelius84 (approved):
LGTM

PR #59286, opened 2023-11-23
add type promotion table.

zxcd:
### PR types

New features
### PR changes

OPs
### Description

Card-64022
Given reference table for type promotion.

### DIFF
--- a/paddle/fluid/eager/type_promotion_utils.h
+++ b/paddle/fluid/eager/type_promotion_utils.h
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+#pragma once
 #include "paddle/fluid/eager/api/generated/eager_generated/forwards/dygraph_functions.h"
 #include "paddle/fluid/eager/api/utils/global_utils.h"
 
--- a/paddle/fluid/eager/type_promotion_utils.h
+++ b/paddle/fluid/eager/type_promotion_utils.h
@@ -79,12 +79,7 @@ inline static phi::DataType promoteTypes(phi::DataType a, phi::DataType b) {
   constexpr auto b1 = phi::DataType::BOOL;
   constexpr auto bf = phi::DataType::BFLOAT16;
 
-  // this matrix has to be consistent with
-  // AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS undefined is used where we
-  // are not sure about the correct value for type promotion.
-  // clang-format off
-  static constexpr phi::DataType _promoteTypesLookup[
-      12][12] = {
+  static constexpr phi::DataType _promoteTypesLookup[12][12] = {
       /*        u1  i1  i2  i4  i8  f2  f4  f8  c4  c8  b1  bf*/
       /* u1 */ {u1, i2, i2, i4, i8, f2, f4, f8, c4, c8, u1, bf},
       /* i1 */ {i2, i1, i2, i4, i8, f2, f4, f8, c4, c8, i1, bf},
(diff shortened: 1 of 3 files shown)


paddle/fluid/eager/type_promotion_utils.h
```diff
@@ -0,0 +1,106 @@
...
+  constexpr auto f4 = phi::DataType::FLOAT32;
+  constexpr auto f8 = phi::DataType::FLOAT64;
+  constexpr auto c4 = phi::DataType::COMPLEX64;
+  constexpr auto c8 = phi::DataType::COMPLEX128;
+  constexpr auto b1 = phi::DataType::BOOL;
+  constexpr auto bf = phi::DataType::BFLOAT16;
+
+  // this matrix has to be consistent with
+  // AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS undefined is used where we
+  // are not sure about the correct value for type promotion.
```
zoooo0820: These comments are inappropriate and can be removed.

python/paddle/base/type_promotion.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def get_result_dtype(x_dtype, y_dtype):
+    if x_dtype == y_dtype:
+        return x_dtype
+    else:
+        try:
+            return promoteTypesLookup[Number[x_dtype]][Number[y_dtype]]
+        except:
+            print(
```
zoooo0820: It would be better to raise an error

lanxianghit (approved):

python/paddle/base/type_promotion.py
```diff
@@ -0,0 +1,70 @@
...
+    [i4, i4, i4, i4, i8, f2, f4, f8, c4, c8, i4, bf],
+    [i8, i8, i8, i8, i8, f2, f4, f8, c4, c8, i8, bf],
+    [f2, f2, f2, f2, f2, f2, f4, f8, c4, c8, f2, f4],
+    [f4, f4, f4, f4, f4, f4, f4, f8, c4, c8, f4, f4],
+    [f8, f8, f8, f8, f8, f8, f8, f8, c8, c8, f8, f8],
+    [c4, c4, c4, c4, c4, c4, c4, c8, c4, c8, c4, c4],
+    [c8, c8, c8, c8, c8, c8, c8, c8, c8, c8, c8, c8],
+    [u1, i1, i2, i4, i8, f2, f4, f8, c4, c8, b1, bf],
+    [bf, bf, bf, bf, bf, f4, f4, f8, c4, c8, bf, bf],
+]
```
jeff41404, line 58: it is better to use a unified map to reduce subsequent maintenance costs, or open some interfaces in the Python layer is also acceptable.

paddle/fluid/eager/type_promotion_utils.h
```diff
@@ -0,0 +1,101 @@
...
+  constexpr auto i8 = phi::DataType::INT64;
+  constexpr auto f2 = phi::DataType::FLOAT16;
+  constexpr auto f4 = phi::DataType::FLOAT32;
+  constexpr auto f8 = phi::DataType::FLOAT64;
+  constexpr auto c4 = phi::DataType::COMPLEX64;
+  constexpr auto c8 = phi::DataType::COMPLEX128;
+  constexpr auto b1 = phi::DataType::BOOL;
+  constexpr auto bf = phi::DataType::BFLOAT16;
+
+  static constexpr phi::DataType _promoteTypesLookup[12][12] = {
```
jeff41404, line 82: we may be adding more strategies of type promotion in the future, such as fp8. would it be better to use constant or  constexpr instead of 12 ?

python/paddle/base/type_promotion.py
zxcd: OK, it will fixed in next pr.

paddle/fluid/eager/type_promotion_utils.h
zxcd: If we add more type in future, this table will also change.  But it can also be modified to const. Can we fix it in the next PR?

PR #59297, opened 2023-11-23
[SOT] fix load null in resume function

feifei-111:
### PR types

Others
### PR changes

Others
### Description

when we created a resume function, it might could not be called directly, for it might load a NullVariable (when resuming stack env). now it is changed by load a read NULL in resume fn

PCard-66972


SigureMo (approved):
LGTMeow

PR #59317, opened 2023-11-24
[Paddle-TRT ]add scatter op into TRT GenericPlugin

zhink:
### PR types

Performance optimization
### PR changes

OPs
### Description

Pcard-71501
add scatter op into TRT GenericPlugin


zhoutianzi666 (approved):
LGTM

PR #59337, opened 2023-11-24
【CINN】Refactor cinn ir_schedule

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR is to refactor ir_schedule because in the future schedule primitives for dynamic shape will be added and most of them will be the same as static shape, so here I refactor them and avoid duplicated code


ZzSean (approved):
LGTM

Issue #59252, opened 2023-11-22
pr_58560 caused the inference time of swin_transformer model is about 15 times longer than that of setting disable_mkldnn()


EmmonsCurse:
### bug描述 Describe the Bug

Error Description： After the merge of https://github.com/PaddlePaddle/Paddle/pull/58560 which make OneDNN default on paddle, the inference time of swin_transformer model is about 15 times longer than that of setting disable_mkldnn().
- Inference time: 61.4823 s
- Inference time (**disable_mkldnn**): 3.9201 s

#### 1. Operating environment
- PaddlePaddle version: develop
- OS version: CentOS
- Python version: 3.8
- CPU info: Intel(R) Xeon(R) Gold 6271C CPU @ 2.60GHz

#### 2. The problem recurrence steps
- Install paddlepaddle_whl and Install dependency package
- run test_swin_transformer.py(compare whether to set config.disable_mkldnn() )

> test_swin_transformer.py
```python
import os
import sys
import tarfile
import time
import wget
import numpy as np
import paddle

from paddle.inference import Config
from paddle.inference import PrecisionType
from paddle.inference import create_predictor


def check_model_exist():
    """
    check model exist
    """
    swin_transformer_url = "https://paddle-qa.bj.bcebos.com/inference_model_clipped/latest/class/swin_transformer.tgz"
    if not os.path.exists("./swin_transformer/inference.pdiparams"):
        wget.download(swin_transformer_url, out="./")
        tar = tarfile.open("swin_transformer.tgz")
        tar.extractall()
        tar.close()


def enable_mkldnn_test(input_data_dict, model_file, params_file):
    """
    Args:
        input_data_dict (dict)
        model_file (str)
        params_file (str)

    Returns:
        time_cost(float)
    """
    config = Config()
    config.set_prog_file(model_file)
    config.set_params_file(params_file)
    # config.disable_mkldnn()

    predictor = create_predictor(config)

    start_time = time.time()

    input_names = predictor.get_input_names()
    for i, input_data_name in enumerate(input_names):
        input_handle = predictor.get_input_handle(input_data_name)
        input_handle.copy_from_cpu(input_data_dict[input_data_name])

    predictor.run()

    output_names = predictor.get_output_names()
    output_handle = predictor.get_output_handle(output_names[0])
    output_data = output_handle.copy_to_cpu()
    end_time = time.time()
    time_cost = end_time - start_time
    return time_cost


if __name__ == "__main__":
[...]

PR #59394, opened 2023-11-27
[OneDNN] Matmul kernel optimize for swin model

zhanglirong1999:
### PR types
Performance optimization

### PR changes
Others

### Description
OneDNN binary as post_ops, it does not support 1CHW to broadcast, which will casue regression issue. #59252  Thus manually use stride to fill it into NCHW.

### DIFF
--- a/paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
+++ b/paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
@@ -135,19 +135,20 @@ class FusedMatmulOneDNNHandler
     this->AcquireForwardPrimitiveDescriptor(matmul_attrs, x_md, y_md, out_md);
   }
 
- dnnl::primitive_attr CreateMatmulAttrs(const OneDNNContext &dev_ctx,
-                                         const DenseTensor *residual_data,
-                                         const float matmul_alpha,
-                                         const std::string &fuse_activation,
-                                         const float fuse_alpha,
-                                         const float fuse_beta,
-                                         const float fused_output_scale,
-                                         const float scale_x,
-                                         const float scale_y,
-                                         const float scale_in_eltwise,
-                                         const float scale_out,
-                                         const bool force_fp32_output,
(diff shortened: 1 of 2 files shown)


zhanglirong1999:
@xinyu-intel , @vivienfanghuagood , @yuanlehome , hi, can you help to review this PR?  Thanks~

paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
```diff
@@ -193,6 +207,28 @@ class FusedMatmulOneDNNHandler
...
+    size_t size = this->fwd_pd_->dst_desc().get_size() / sizeof(XT);
+    XT *dst = static_cast<XT *>(src_mem->get_data_handle());
+
+#pragma omp parallel for
+    for (size_t i = 0; i < size; ++i) {
+      auto mod_i =
+          static_cast<int>(i - floor(i / (IC * IH * IW)) * (IC * IH * IW));
+      // Use stride to make 1*C*H*W to N*C*H*W to avoid broadcast overhead
+      dst[i] = input_data[mod_i];
+    }
```
xinyu-intel: seems here changes the data handle instead of the stride?

LLee233: Do you mean directly change the stride of mem? That is also a good method, but I am not sure if it's intuitive? This kind of change in PR can also attain a big improvement, though yours seems bring less overhead...

xinyu-intel: I think both are okay, but the code change is not align with the comments. Maybe adjust the comments will be better. BTW, is openmp always available in Paddle?

LLee233: Seems they diasble openmp in WIN32 case:
image
Maybe adding
`#if defined(_OPENMP)`
could be better.

yuanlehome (approved):
LGTM

xinyu-intel (approved):

PR #59413, opened 2023-11-27
Add CUDNNv8 max pooling

Tom-Zheng:
### PR types

New features
### PR changes

OPs
### Description

This PR adds CUDNNv8 implementation of max pooling. Benefited from runtime fusion engine of CUDNNv8, it achieves significant speedup over legacy CUDNN pooling.

### DIFF
--- a/paddle/phi/api/yaml/fused_ops.yaml
+++ b/paddle/phi/api/yaml/fused_ops.yaml
@@ -383,6 +383,9 @@
     func : layer_norm_act_xpu
     data_type : x
 
+# This op is implemented using CUDNN Frontend API, which serves as a supplement to
+# legacy max pooling implementation. It shows better performance with NHWC layout and
+# half precision.
 - op : max_pool2d_v2
   args : (Tensor x, int[] kernel_size, int[] strides= {1, 1}, int[] paddings = {0, 0}, str data_format = "NCHW", bool global_pooling = false, bool adaptive = false)
   output : Tensor(out), Tensor(saved_idx)
--- a/paddle/phi/api/yaml/fused_ops.yaml
+++ b/paddle/phi/api/yaml/fused_ops.yaml
@@ -383,6 +383,18 @@
     func : layer_norm_act_xpu
     data_type : x
 
+- op : max_pool2d_v2
+  args : (Tensor x, int[] kernel_size, int[] strides= {1, 1}, int[] paddings = {0, 0}, str data_format = "NCHW", bool global_pooling = false, bool adaptive = false)
+  output : Tensor(out), Tensor(saved_idx)
+  infer_meta :
+    func : MaxPoolV2InferMeta
+    param : [x, kernel_size, strides, paddings, data_format, global_pooling, adaptive]
+  kernel :
+    func : max_pool2d_v2
+    param : [x, kernel_size, strides, paddings, data_format, global_pooling, adaptive]
+  intermediate: saved_idx
+  backward : max_pool2d_v2_grad
+
 - op : multi_encoder_xpu
   args : (Tensor x, Tensor[] fc_weight, Tensor[] fc_weight_max, Tensor[] fc_bias, Tensor[] ln_scale, Tensor[] ln_bias, Tensor mask, Tensor seq_lod, Tensor max_seq_len, int layer_num, bool norm_before, int hidden_dim, int head_num, int size_per_head, int ffn_hidden_dim_scale, int act_type, int relative_type, int slice_idx)
   output : Tensor(out), Tensor(x_fp16), Tensor(out_fp16)
--- a/paddle/phi/api/yaml/ops.yaml
+++ b/paddle/phi/api/yaml/ops.yaml
@@ -1702,6 +1702,18 @@
     func : matrix_power
   backward : matrix_power_grad
 
+- op : max_pool2d_v2
+  args : (Tensor x, int[] kernel_size, int[] strides= {1, 1}, int[] paddings = {0, 0}, str data_format = "NCHW", bool global_pooling = false, bool adaptive = false, bool use_cudnn = true)
+  output : Tensor(out), Tensor(saved_idx)
+  infer_meta :
+    func : MaxPoolV2InferMeta
+    param : [x, kernel_size, strides, paddings, data_format, global_pooling, adaptive]
+  kernel :
+    func : max_pool2d_v2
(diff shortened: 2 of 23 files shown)


paddle/phi/infermeta/unary.cc
```diff
@@ -2299,6 +2299,37 @@ void MaxPoolWithIndexInferMeta(const MetaTensor& x,
...
+                  false,
+                  data_format,
+                  "max",
+                  global_pooling,
+                  adaptive,
+                  "EXPLICIT",
+                  out,
+                  config);
+  saved_idx->set_dims(out->dims());
+  saved_idx->set_dtype(phi::CppTypeToDataType<int>::Type());
```
zyfncg, line 2375: why dose not set dims and dtype for `out`?

paddle/phi/api/yaml/ops.yaml
```diff
@@ -1686,6 +1686,18 @@
     func : matrix_power
   backward : matrix_power_grad
 
+- op : max_pool2d_v2
```
zyfncg: `xxx_v2` isn't a good name for op, it should be same with python api.

paddle/phi/infermeta/unary.cc
Tom-Zheng: It is set in `Pool2DInferMeta`.

paddle/phi/api/yaml/ops.yaml
Tom-Zheng: Currently `MaxPool2D` calls pool2d op. When layout is NHWC, it's performance is bad. That's the reason we want to add a new CUDNNv8 implementation. It is intended to speedup some specific use cases. 

Ideally, it will be something like this in MaxPool2D forward:
```
if layout == "NHWC" and input.dtype == paddle.float16:
    return max_pool2d_v2(input)
else:
    return pool2d(input, pool_mode='max')
```

I thought it is inappropriate to name this op 'max_pool2d', as implementation of max_pool2d is included in `pool2d` op. This is the reason for the name "v2". This also implies that this op is 'another' implementation, only to cover some cases, not a standalone one.

What name do you suggest?

zyfncg: image

There have been two ops for `max_pool`, we dosen't want to add more op for this api `max_pool`, Could this kernel be add into the kernel of `max_pool2d_with_index` or `pool2d`?

Tom-Zheng: The new max_pool2d op signature is different from pool2d (needs an extra mask output for backward), so it is difficult to be added into pool2d. 

Although `max_pool2d_with_index` has an extra output, but adding the new implementation is also inappropriate: According to the semantics of max_pool2d_with_index, the output index is required. However, max_pool2d_v2 does not have to output an index (could be empty). It's index output is local index within the window, instead of global index like `max_pool2d_with_index`. Also, it is intended only for backward, not directly accessible by user. This is the reason the sved_idx output is set as an intermediate tensor [here](https://github.com/PaddlePaddle/Paddle/blob/b27c4314a7fc087de86532fc5b392b775851b11f/paddle/phi/api/yaml/ops.yaml#L1698).

zyfncg: Under the current design framework, this is indeed a problem. We may need more information for making a decision：

- What is the accurate value of performance improvement for `max_pool2d_v2`? 
- Is `saved_idx` necessary? If not using `saved_idx`, the performance will decrease?

Tom-Zheng: 1. The performance gain is huge, in NHWC layout and FP16 dtype. This figure shows the avg time of a pooling fwd + bwd: 
image

The speedup is at least 2.5x, up to 7.5x, depending on config.

2. Yes it is necessary for NHWC layout. The runtime fused pool kernel cannot work without it.

zyfncg: > 2. Yes it is necessary for NHWC layout. The runtime fused pool kernel cannot work without it.

Is it a fused kernel?  It will be used from python api or pass? 
If it's a fused kenrel, rename to `fused_max_pool2d` may be better.

Tom-Zheng: It is technically not a fused kernel -- it's only a max pooling and can be called from either python api or pass. However, the kernel is generated JIT using runtime fusion engine.

zyfncg: Can `saved_idx` be add as a new output of `pool2d`？

Tom-Zheng: Yes, but it will involve a lot of changes -- adding an additional saved_idx to all forward / backward kernel functions of all backends (cpu, xpu, mkldnn for example, even though they will never use it); I also have to change the pooling trt converter to make sure TRT converter recognize this new signature. It could be 20+ file changes, for adding a new output. Compared to that, I think create a new op is a relatively lightweight solution.

Tom-Zheng:
I have removed the dygraph API, would you please take a look? @zyfncg

Tom-Zheng:
@zyfncg All CIs have passed. This PR is ready for review.

paddle/phi/api/yaml/fused_ops.yaml
```diff
@@ -358,6 +358,18 @@
     func : layer_norm_act_xpu
     data_type : x
 
+- op : max_pool2d_v2
```
zyfncg, line 389: Could add some comments to explain the diff between `max_pool2d` and this op?

zyfncg (dismissed):

paddle/phi/api/yaml/fused_ops.yaml
Tom-Zheng: Done

Tom-Zheng:
Conflicts resolved

Tom-Zheng:
@zyfncg Would you please review?

zyfncg (dismissed):

Tom-Zheng:
@ZzSean  would you please approve PR-CI-OP-benchmark? This PR does not affect matmul op.

Tom-Zheng:
I cannot rerun pipeline -- got "502 Bad Gateway".

XieYunshen (approved):
LGTM for `set_tests_properties(test_pool_max_op PROPERTIES TIMEOUT 500)`

Aurelius84 (approved):
LGTM for const_cast

zyfncg (approved):

phlrain (approved):

PR #59461, opened 2023-11-28
fix the bug in moving skip_layernorm op to phi

zeroRains:
### PR types

Bug fixes
### PR changes

Others
### Description

fix the bug in this pr https://github.com/PaddlePaddle/Paddle/pull/58396


yuanlehome (approved):
LGTM

PR #59527, opened 2023-11-29
[OneDNN] Fix accuracy problem on fused matmul

zhanglirong1999:
### PR types
Bug fixes

### PR changes
Others

### Description
When open OneDNN, there are some accuracy in some models, issue: #59510, this PR will fix SimpleRNNCell and GRUCell accuracy drop.

### DIFF
--- a/paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
+++ b/paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
@@ -166,9 +166,31 @@ class FusedMatmulOneDNNHandler
     }
 
     if (residual_data) {
-      auto residual_data_md = memory::desc(out_ddims,
-                                           funcs::OneDNNGetDataType<OT>(),
-                                           dnnl::memory::format_tag::any);
+      // fill 1 in the front of adesc, to make residual ndims to be same as dst
+      // dims
+      int dst_size = out_ddims.size();
+      int origin_size = residual_data->mem_desc().get_ndims();
+      dnnl::memory::dims expanded_dims = residual_data->mem_desc().get_dims();
+      if (origin_size < dst_size) {
+        expanded_dims.insert(expanded_dims.begin(), dst_size - origin_size, 1);
+        residual_data->mem_desc().reshape(expanded_dims);
+      }
+
+      auto residual_data_tz = vectorize(residual_data->dims());
+      auto chosen_memory_format = funcs::OneDNNMemoryFormat::any;
+      dnnl::memory::desc residual_data_md;
+      if (residual_data_tz.size() == 4 && residual_data_tz[0] == 1 &&
+          residual_data_tz[1] > 1 && residual_data_tz[2] > 1 &&
+          residual_data_tz[3] > 1) {
+        chosen_memory_format = funcs::OneDNNMemoryFormat::nchw;
+        residual_data_md = memory::desc(
+            out_ddims, funcs::OneDNNGetDataType<OT>(), chosen_memory_format);
+      } else {
+        residual_data_md = memory::desc(expanded_dims,
+                                        funcs::OneDNNGetDataType<OT>(),
+                                        chosen_memory_format);
+      }
+
       post_operations.append_binary(dnnl::algorithm::binary_add,
                                     residual_data_md);
       if (scale_in_eltwise != 0.0f) {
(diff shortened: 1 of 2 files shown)


zhanglirong1999:
之前准确性出问题原因在于，residual的memory直接AcquireSrcMemory，而这个模型中residual的维度实际上只有一维，对于小于dst的情况下，需将residual的维度手动expend

paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
```diff
@@ -166,6 +166,16 @@ class FusedMatmulOneDNNHandler
     }
 
     if (residual_data) {
+      // fill 1 in the front of adesc, to make residual ndims to be same as dst
+      // dims
+      int dst_size = out_ddims.size();
+      int origin_size = residual_data->mem_desc().get_ndims();
+      dnnl::memory::dims expanded_dims = residual_data->mem_desc().get_dims();
+      if (origin_size < dst_size) {
```
LLee233, line 175: Hi @zhanglirong1999 , currently this fix is mainly targeting to the reported issue, where the dimension of residual_data is [H] while dst is [1, 1, H] (or [1, H] and [1, 1, H] also works), the similar dimension is just right on the last. However, if the dst is [1, H, 1] or [H, 1, 1] or more shape for higher dimensions. This fix may not be able to solve such situations. 

Usually I assume such kind of incompatibility of dimensions happens when residual_data is scale, so maybe and one more condition could extend the range of this fix's application?
```suggestion
      if (origin_size < dst_size) {
        if (origin_size == 1) {
          # compare which dimension of dst is same as residual_size
          # expand the dims based on result
          # reshape the md
        } else {
        expanded_dims.insert(expanded_dims.begin(), dst_size - origin_size, 1);
        residual_data->mem_desc().reshape(expanded_dims);
        }}
```

zhanglirong1999: yeah, I think it's a good consideration, but in this way we also need to consider more multi-dimensional and multi-dimensional matching. Numpy may not support such extreme broadcasting either. And it seems that OneDNN only considers one dimension to expand, treating them as scale~ 
https://github.com/oneapi-src/oneDNN/blob/master/src/graph/backend/dnnl/common.cpp#L229

LLee233: Yea you're right~ Then we should just keep the same as what oneDNN does~

zhanglirong1999:
@xinyu-intel , @vivienfanghuagood , @yuanlehome ,hi, can you help me review this? Thanks~

paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
```diff
@@ -308,7 +326,7 @@ void ExecuteFusedMatmul(const OneDNNContext &dev_ctx,
         residual_data_vec[3] > 1) {
       residual_data_memory_p = handler.AcquireSrcMemoryStride(residual_data);
     } else {
-      residual_data_memory_p = handler.AcquireSrcMemory(residual_data);
+      residual_data_memory_p = handler.AcquireSrcMemoryResidual(residual_data);
```
xinyu-intel, line 328: what's the difference between this added func with AcquireSrcMemory?

zhanglirong1999: Assume the formula is src * weight + residual, AcquireSrcMemory will use src mem_desc to create memory , not residual mem_desc. I think it's a bug in used implement, since residual dims can different with src dims.

xinyu-intel: do we have a general func to acquire mdesc. It seems the newly added func is not limited with residual memory. It create memory based on the input of the func.

zhanglirong1999: Well, seems don't have such general func now, you can see that in onednn_reuse.h, all mdesc is confirmed in advance.
https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/phi/backends/onednn/onednn_reuse.h#L200
I found in conv kernel,  we also have to create such func to acquire residual memory.

xinyu-intel: okay. got u. I'm not sure if the residual memory desc should be also queried from the pd like what src does and then reorder if the layout is not the same. Can you please double comfirm on it? cc @YangQun1

zhanglirong1999: In my investigation, it seems that no pd in oneDNN can support post-op desc acquisition like src_desc() or weights_desc()? 
So do we also need to reorder manually for dst and residual in this issue?

paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
```diff
@@ -176,7 +186,7 @@ class FusedMatmulOneDNNHandler
         residual_data_md = memory::desc(
             out_ddims, funcs::OneDNNGetDataType<OT>(), chosen_memory_format);
       } else {
-        residual_data_md = memory::desc(residual_data_tz,
+        residual_data_md = memory::desc(expanded_dims,
                                         funcs::OneDNNGetDataType<OT>(),
                                         chosen_memory_format);
```
xinyu-intel: residual_data->mem_desc()

paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
```diff
@@ -166,6 +166,16 @@ class FusedMatmulOneDNNHandler
...
 
     if (residual_data) {
+      // fill 1 in the front of adesc, to make residual ndims to be same as dst
+      // dims
+      int dst_size = out_ddims.size();
+      int origin_size = residual_data->mem_desc().get_ndims();
+      dnnl::memory::dims expanded_dims = residual_data->mem_desc().get_dims();
+      if (origin_size < dst_size) {
+        expanded_dims.insert(expanded_dims.begin(), dst_size - origin_size, 1);
+        residual_data->mem_desc().reshape(expanded_dims);
```
YangQun1: AFAIK, reshape() is a const member function which will return a new md. So the residual_data->mem_desc() is not changed here. Is it as you expected here?
```cpp
        desc reshape(const dims &adims, bool allow_empty = false) const {
            if (get_ndims()) validate_dims(adims, 1);
            dnnl_memory_desc_t out_md = nullptr;
            dnnl_status_t status = dnnl_memory_desc_reshape(
                    &out_md, get(), (int)adims.size(), adims.data());
            if (!allow_empty)
                error::wrap_c_api(
                        status, "could not reshape a memory descriptor");
            return desc(out_md);
        }
```

xinyu-intel (dismissed):

paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
YangQun1: when you create the pd, do you pass in a residual md with any format tag? if no, I think we don't need to query, just use the residual md.

LLee233 (approved):

paddle/phi/kernels/fusion/onednn/fused_matmul_kernel.cc
xinyu-intel: maybe `reshaped = residual_data->mem_desc().reshape(expanded_dims);` and then pass it to L191 will be better. Currently, it happen to work because expanded_dims and any layout are being using.

zhanglirong1999: yeah, I tired, you are right. It works~ thanks, I will modify.

YangQun1 (approved):

xinyu-intel (approved):

PR #59562, opened 2023-11-30
[SOT] support call_AST stmt

feifei-111:
### PR types

Others

### PR changes

Others

### Description

try to call AST first when start simulate or inline call

PCard-66972


2742195759 (approved):
LGTM

PR #59598, opened 2023-12-01
[auto parallel] fix pp reshard

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
PCard-73145
fix pp reshard bug


LiYuRio (approved):
LGTM

PR #59641, opened 2023-12-03
[oneDNN] Reshape attr_axes when going to oneDNN kernel

LLee233:
### PR types

Bug fixes

### PR changes

Others

### Description

This PR aims to fix Conv1DTranspose in https://github.com/PaddlePaddle/Paddle/issues/59510.

When **Squeeze** op is dispatched to oneDNN, corresponding Tensors will be transformed to oneDNN DType. But currently there is only transformation for tensors, **attr_axes** still follows default dtype instead of oneDNN DType, so here we change it to well performs Squeeze.

### DIFF
--- a/paddle/phi/kernels/onednn/squeeze_kernel.cc
+++ b/paddle/phi/kernels/onednn/squeeze_kernel.cc
@@ -59,7 +59,29 @@ void SqueezeInferKernel(const Context& dev_ctx,
                         const IntArray& axes,
                         DenseTensor* out) {
   auto x_dims = x.dims();
+  auto x_dims_tz = x_dims.size();
   std::vector<int32_t> tmp(axes.GetData().begin(), axes.GetData().end());
+
+  // Currently there is only tranformation for tensors, while attr axes still
+  // follows default dtype instead of oneDNN dtype, so here manually change it
+  if ((x_dims_tz >= 3) &&
+      (phi::OneDNNContext::tls().get_cur_paddle_data_layout() ==
+           phi::DataLayout::NHWC ||
+       phi::OneDNNContext::tls().get_cur_paddle_data_layout() ==
+           phi::DataLayout::NDHWC)) {
+    int axes_size = tmp.size();
+    for (int i = 0; i < axes_size; i++) {
+      if (tmp[i] < 0) {
+        tmp[i] += x_dims_tz;
+      }
+      if (tmp[i] >= 1 && tmp[i] < (x_dims_tz - 1)) {
+        tmp[i] += 1;
+      } else if (tmp[i] == (x_dims_tz - 1)) {
+        tmp[i] = 1;
+      }
+    }
+  }
+
   auto out_dims = funcs::GetOutputSqueezeShape(tmp, x_dims, true);
   ExecuteSqueeze<T, Context>(dev_ctx, x, x_dims, out_dims, out);
 }
--- a/paddle/phi/kernels/onednn/squeeze_kernel.cc
+++ b/paddle/phi/kernels/onednn/squeeze_kernel.cc
@@ -65,10 +65,8 @@ void SqueezeInferKernel(const Context& dev_ctx,
   // Currently there is only tranformation for tensors, while attr axes still
   // follows default dtype instead of oneDNN dtype, so here manually change it
   if ((x_dims_tz >= 3) &&
-      (phi::OneDNNContext::tls().get_cur_paddle_data_layout() ==
-           phi::DataLayout::NHWC ||
-       phi::OneDNNContext::tls().get_cur_paddle_data_layout() ==
-           phi::DataLayout::NDHWC)) {
+      phi::OneDNNContext::tls().get_cur_paddle_data_layout() ==
+          phi::DataLayout::NHWC) {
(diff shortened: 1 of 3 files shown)


LLee233:
This change is kind of tricky... And should find better way to solve it.

paddle/phi/kernels/onednn/squeeze_kernel.cc
```diff
@@ -60,6 +60,12 @@ void SqueezeInferKernel(const Context& dev_ctx,
                         DenseTensor* out) {
   auto x_dims = x.dims();
   std::vector<int32_t> tmp(axes.GetData().begin(), axes.GetData().end());
+
+  // Currently there is only transformation for tensors, while attr axes still
+  // follows default dtype instead of oneDNN dtype, so here manually change it
+  // TODO(Li Xinyi): Is there a way to achieve it at earlier stage?
```
zhanglirong1999: Maybe not every time need this transformation, can we use it based on get_cur_paddle_data_layout()? It seems only happens in get_cur_paddle_data_layout() equals kNHWC or NHWC?

zhanglirong1999 (commented):
Maybe not every time need this transformation, can we use it based on get_cur_paddle_data_layout()? It seems only happens in get_cur_paddle_data_layout() equals kNHWC or NHWC?

paddle/phi/kernels/onednn/squeeze_kernel.cc
```diff
@@ -60,6 +60,12 @@ void SqueezeInferKernel(const Context& dev_ctx,
...
+  funcs::GetSqueezeNewAxes(tmp);
```
zhanglirong1999:
``` 
 std::vector<int64_t> formated_axis = axes.GetData();
  if ((x_dims.size() >= 3) &&
      (phi::OneDNNContext::tls().get_cur_paddle_data_layout() ==
       phi::DataLayout::kNHWC)) {
    int axis_size = axes.GetData().size();
    for (int i = 0; i < axis_size; i++) {
      if (axes.GetData()[i] < 0) {
        formated_axis[i] = axes.GetData()[i] + axis_size;
      }
    }
  }
  std::vector<int32_t> tmp(formated_axis.begin(), formated_axis.end());
```
It seems can use such method to solve it?

LLee233: > ```
>  std::vector<int64_t> formated_axis = axes.GetData();
>   if ((x_dims.size() >= 3) &&
>       (phi::OneDNNContext::tls().get_cur_paddle_data_layout() ==
>        phi::DataLayout::kNHWC)) {
>     int axis_size = axes.GetData().size();
>     for (int i = 0; i < axis_size; i++) {
>       if (axes.GetData()[i] < 0) {
>         formated_axis[i] = axes.GetData()[i] + axis_size;
>       }
>     }
>   }
>   std::vector<int32_t> tmp(formated_axis.begin(), formated_axis.end());
> ```
> 
> It seems can use such method to solve it?

Thanks for your advice! Do you think it's better to add check function when `get_cur_paddle_data_layout() ==
       phi::DataLayout::NDHWC`?

zhanglirong1999: https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/phi/kernels/transfer_layout_kernel.cc#L161, maybe still need to kNHWC or NHWC? (kNHWC define equals NHWC)

LLee233: > https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/phi/kernels/transfer_layout_kernel.cc#L161, maybe still need to kNHWC or NHWC? (kNHWC define equals NHWC)

Seems it assumed the dims would only be 4...  Here indicates 5D dims. So maybe add one more case is also good?
https://github.com/PaddlePaddle/Paddle/blob/9d7150b8747997d5a2264c834376ed5f55a4c9d4/paddle/common/layout.h#L36

zhanglirong1999: sure, I think it's more reasonable

LLee233:
Hi @xuxinyi389, is it you opened cuda check for PR? Since I saw there is only you (the only xuxinyi) can be at… If so, would you mind helping check this CI issue? Thanks!

xuxinyi389:
> Hi @xuxinyi389, is it you opened cuda check for PR? Since I saw there is only you (the only xuxinyi) can be at… If so, would you mind helping check this CI issue? Thanks!
PR-CI-Windows-Tesr-Cuda12 is a CI for test, you don't have to care it.

LLee233:
> > >Hi @xuxinyi389, is it you opened cuda check for PR? Since I saw there is only you (the only xuxinyi) can be at… If so, would you mind helping check this CI issue? Thanks!

> > PR-CI-Windows-Tesr-Cuda12 is a CI for test, you don't have to care it.

So it won’t block merge? Thanks! 

@yuanlehome @xinyu-intel @vivienfanghuagood would you mind helping check this PR? Thanks~

xinyu-intel (approved):

PR #59706, opened 2023-12-05
[auto parallel] Dist tensor set value

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
Pcard-73145
dist tensor `set_value` method


LiYuRio (approved):
LGTM

PR #59710, opened 2023-12-05
[SOT] fix sot call locals

feifei-111:
### PR types

Others
### PR changes

Others
### Description

currently we will call locals by break graph, but this will lead to error in some cases, so we will fallback before simulate if locals is called

PCard-66972


SigureMo (approved):
LGTMeow

PR #59714, opened 2023-12-05
[PIR] Fix bugs to open pir api test for cond op

chen2016013:
### PR types

Others

### PR changes

Others

### Description

fix bugs to open pir api test for cond op
Pcard-67164


zhangbo9674 (dismissed):
LGTM

PR #59744, opened 2023-12-06
[OneDNN] Restrictions on matmul broadcast optimiztion

zhanglirong1999:
### PR types
Performance optimization

### PR changes
Others

### Description
We found in previous PR #59394 can largely optimize swin_transformer, but will cause Ernie int8 performance drop. We have re-constrained the previous optimization conditions to avoid it.


zhanglirong1999:
@xinyu-intel , @vivienfanghuagood , @yuanlehome , hi, can you help me review it? Thanks~

xinyu-intel (approved):

PR #59749, opened 2023-12-06
[PIR+CINN]Support Adapative Parse and Check Feed/Fetch in SubGraph Exporter

Aurelius84:
### PR types

Function optimization
### PR changes

Others
### Description

Pcard-67164


SigureMo (approved):
LGTMeow

PR #59830, opened 2023-12-08
[auto parallel] fix embedding_grad's weight row-wise parallel error

cxxly:
### PR types

Bug fixes
### PR changes

Others
### Description

Pcard-73145
fix embedding_grad's weight row-wise parallel error


LiYuRio (approved):
LGTM

PR #59834, opened 2023-12-08
[SOT] add layer black list

feifei-111:
### PR types

Others
### PR changes

Others
### Description

add black list for paddle layers which might not supported in sot yet
PCard-66972


Aurelius84 (approved):

SigureMo (approved):
LGTMeow

PR #59853, opened 2023-12-08
[AutoConfig]Skip sharding optimizer

Caozhou1995:
### PR types

 New features
### PR changes

Others
### Description

PCard-78552
This PR will skip sharding optimizer when in single dp estimation scene.


JZ-LIANG (approved):
LGTM

PR #59973, opened 2023-12-13
【Hackathon 5th No.28】为 Paddle 新增 slice_scatter API -part

megemini:
### PR types

New features

### PR changes

APIs

### Description

【Hackathon 5th No.28】为 Paddle 新增 slice_scatter API

RFC：

v1 : https://github.com/PaddlePaddle/community/pull/784
v2 : https://github.com/PaddlePaddle/community/pull/790

<del>p.s. 待 RFC 评审之后完善代码与测试 ～</del>


megemini:
@zoooo0820 CI 应该没啥问题了 请评审 ～ 谢谢！：）

test/legacy_test/test_slice_scatter.py
```diff
@@ -0,0 +1,226 @@
...
+import numpy as np
+
+import paddle
+from paddle.framework import core
+from paddle.pir_utils import test_with_pir_api
+
+paddle.enable_static()
+
+RTOL = {'float32': 1e-03, 'float64': 1e-05}
+ATOL = {'float32': 1e-03, 'float64': 1e-05}
```
zoooo0820: 这里设置的两个阈值相对数据类型本身是否宽容度高了些，能否用默认参数测试

python/paddle/tensor/manipulation.py
```diff
@@ -6749,3 +6749,106 @@ def select_scatter(x, values, axis, index, name=None):
...
+        'steps': steps,
+        'decrease_axes': decrease_axes,
+        'none_axes': none_axes,
+    }
+
+    dtype = x.dtype
+    attrs['dtype'] = dtype
+
+    value = value.astype(dtype)
+    inputs["ValueTensor"] = value
```
zoooo0820: 仅静态图分支用到的参数，比如inputs / attrs等，挪到静态图分支下吧

test/legacy_test/test_slice_scatter.py
```diff
@@ -0,0 +1,226 @@
...
+                rtol=RTOL[self.dtype],
+                atol=ATOL[self.dtype],
+            )
+
+            paddle.enable_static()
+
+
+class TestSliceScatterApiFloat32(TestSliceScatterApi):
+    def init_dtype(self):
+        self.dtype = 'float32'
```
zoooo0820: 也补充下其他数据类型的测试吧，应该不用特别设置atol /rtol? 可以关注下fp16和bf16的单测通过情况

test/legacy_test/test_slice_scatter.py
megemini: 嗯 这里没必要单独设置，已删掉 ～

test/legacy_test/test_slice_scatter.py
megemini: 已增加测试用例 ～ 

另外，由于旧 ir 有些数据类型不支持，所以在测试用例里面单列出来了 ～

zoooo0820 (approved):
LGTM

python/paddle/tensor/manipulation.py
```diff
@@ -6749,3 +6749,108 @@ def select_scatter(x, values, axis, index, name=None):
         )
 
         return output
+
+
+def slice_scatter(x, value, axis=0, start=None, stop=None, step=1, name=None):
```
sunzhongkai588: 这里的默认值好像和下面的参数描述没对上？（`start` 和 `stop`）

megemini: 赞！🤙🤙🤙 

确实不一样，当时也很纠结要怎么写 ～ 

因为这里如果不写参数的话，如：slice_scatter(x, value) ，那么 start 为 None，程序里面会转换为 0，stop 为 None，会转换为 x.shape[axis] ～

参考 python 中 range 的写法 :

https://docs.python.org/3/library/stdtypes.html#range
> If the start argument is omitted, it defaults to 0.

有没有什么建议的写法？谢谢！：）

sunzhongkai588: > 因为这里如果不写参数的话，如：slice_scatter(x, value) ，那么 start 为 None，程序里面会转换为 0，stop 为 None，会转换为 x.shape[axis] ～

个人建议，可以直接在`start` 和  `stop`的参数描述里说清楚，当为 None 时，会出现的情况，且默认值为 None。

`def ...` 这边还是保留参数。 宗旨就是减少读者误解

sunzhongkai588 (dismissed):
LGTM

luotao1:
可以提供中文文档

python/paddle/tensor/manipulation.py
```diff
@@ -6749,3 +6749,108 @@ def select_scatter(x, values, axis, index, name=None):
...
+    """
+    Embeds the value tensor into x at the given axis. Returns a new tensor instead of a view.
+
+    Args:
+        x (Tensor) : The input Tensor. Supported data types are `bool`, `float16`, `float32`, `float64`, `uint8`, `int8`, `int16`, `int32`, `int64`, `bfloat16`, `complex64`, `complex128`.
+        value (Tensor) : The tensor to embed into x. Supported data types are `bool`, `float16`, `float32`, `float64`, `uint8`, `int8`, `int16`, `int32`, `int64`, `bfloat16`, `complex64`, `complex128`.
+        axis (int) : the dimension to insert the value. Default is 0.
+        start (int, optional) : the start index of where to insert. Default is `None` which will be converted to `0`.
+        stop (int, optional) : the stop index of where to insert. Default is `None` which will be converted to `x.shape[axis]`.
+        step (int, optional) : the step for each insert. Default is 1.
```
jeff41404: Because operator of `set_value`can support writing data according to multiple `axes`, `starts`, `ends` and `steps`, it is better that the parameters of `axis`, `start`, `stop`, `step` here can support `int` and `list of int` ？ making this API more user-friendly and powerful

jeff41404: If modified according to the above suggestions, the parameter naming suggestion are `axes`, `starts`, `ends` and `strides`, which is consistent with API of `paddle.slice` and `paddle.strided_slice`. and [rfc](https://github.com/PaddlePaddle/community/blob/master/rfcs/APIs/20231206_api_design_for_slice_scatter.md) should also be modified.

python/paddle/tensor/manipulation.py
```diff
@@ -6749,3 +6749,108 @@ def select_scatter(x, values, axis, index, name=None):
...
+    start = 0 if start is None else start
+    stop = x_shape[axis] if stop is None else stop
+
+    index = list(range(start, stop, step))
+    exp_shape = [*x_shape[:axis], len(index), *x_shape[axis + 1 :]]
+    if tuple(exp_shape) != tuple(value_shape):
+        raise ValueError(
+            "The value.shape should be same of [*x_shape[:axis], len(index), *x_shape[axis+1:]],"
+            f"but got value.shape of {value.shape} and slice shape {exp_shape}."
+        )
```
jeff41404: we also should support the shape of `value` can broadcast to the shape of `exp_shape`, not only exactly equal, and this check will be completed in `set_value`, so there is no need to add this check here, which actually limits the functionality.

python/paddle/tensor/manipulation.py
```diff
@@ -6749,3 +6749,108 @@ def select_scatter(x, values, axis, index, name=None):
...
+            >>> print(res)
+            Tensor(shape=[8, 8], dtype=float32, place=Place(cpu), stop_gradient=True,
+            [[0., 0., 1., 0., 1., 0., 0., 0.],
+             [0., 0., 1., 0., 1., 0., 0., 0.],
+             [0., 0., 1., 0., 1., 0., 0., 0.],
+             [0., 0., 1., 0., 1., 0., 0., 0.],
+             [0., 0., 1., 0., 1., 0., 0., 0.],
+             [0., 0., 1., 0., 1., 0., 0., 0.],
+             [0., 0., 1., 0., 1., 0., 0., 0.],
+             [0., 0., 1., 0., 1., 0., 0., 0.]])
```
jeff41404: the example code needs to add some cases to facilitate user understanding and not be too similar to competitors

python/paddle/tensor/manipulation.py
megemini: No ~~~

`list of int` of `axis`, `start`, `stop` and `step` for `slice_scatter` NOT make sense ~ 

For instance:

`x` is a tensor of shape `[8, 8]`, `value` is `[8, 2]`, then the `axis` can ONLY be `1`. If `axis=0` , the values of `value` in axis 1 of length 2 can not be fitted into `x` of length 8. In other words, `list of int` can only be useful of `x` and `value` the same shape, which is not that useful ~~~ BTW, torch not support `list of int` either ~~~

python/paddle/tensor/manipulation.py
megemini: OK！ remove the check and add `broadcast` test cases ~

megemini:
#### Update 20231220

- 删除掉 shape 的 check
- 增加 broadcast 的单测
- 增加示例程序

关于 `list of int` 的 `axis` 等参数，个人感觉还是没必要支持，原因：

- 场景应用很少，没必要，而且会对用户的使用产生困惑 ～
- torch 和 mindspore 都不支持，应该也算是行业惯例 ～

如何？

python/paddle/tensor/manipulation.py
jeff41404: From a functional perspective, the functionality of this API should be dual to `paddle.strided_slice`. Because the former is for writing data, while the latter is for reading data, so supporting `list of int` is more natural, and the underlying operator of `set_value` also supports it. As for the example you gave, if the shape of `value` cannot be broadcasted to the shape after the slice, an error should be reported (by `set_value`). Additionally, not all of design of torch are reasonable, but how to make it better for users is what we pursue

megemini: > From a functional perspective, the functionality of this API should be dual to `paddle.strided_slice`. Because the former is for writing data, while the latter is for reading data, so supporting `list of int` is more natural, and the underlying operator of `set_value` also supports it. As for the example you gave, if the shape of `value` cannot be broadcasted to the shape after the slice, an error should be reported (by `set_value`). 

Not agree ~ 

Especially treat `slice_scatter` as a dual to `paddle.strided_slice`.

There should be a powerful API `set_value` or `set_value_with_tensor`, not `slice_scatter` ~

We should not bother users to map `slice_scatter` from torch, mindspore or other framework to ours, and not mess the API up with more options just because we could support it ~

> Additionally, not all of design of torch are reasonable, but how to make it better for users is what we pursue

Agree and always !!!

megemini:
关于这个 API 是否支持 `list of int` ，最后说一下我的看法 ～

jeff 大佬认为 `slice_scatter` 与 `strided_slice` 是 `对偶` 关系，因此需要支持 ～ 这个理由实在无法说服我 ～

- 首先，我从来不觉得 torch 的设计就是对的，很多 PR 都可以表明我的态度，这里不再赘述。

- 其次，是否 `对偶` ，个人觉得应该从用户的角度出发，而不是我们决定。
  比如说 `hsplit` 和 `hstack` 虽然参数不同，但我觉得可以是 `功能对偶` ，因为从名字上就可以很直观的理解。
  而 `slice_scatter` 与 `strided_slice` 怎么看都看不出哪儿 `对偶` 了，至少刚接触这两个 API 的我很难想到 `对偶` ... ... 🤣🤣🤣
  如果一定要给 `strided_slice` 找个 `对偶` 的名字，我觉得 `strided_scatter` 更合适，`strided` 表示这两个 API 是步进的，`slice` 表示 `read`，`scatter` 表示 `write` ～
  为了 `对偶` 和极少用到的情况而给用户带来困惑，和抛异常，实在不能赞同 ～～～

- 最后，如果内部研发一致支持 `list of int` 的用法，这里改起来不要太简单 ... ... 😆😆😆 

不用我单独翻译成英文再写一遍吧 ～ 😜😜😜

@zoooo0820 @jeff41404

python/paddle/tensor/manipulation.py
jeff41404: My opinion is that regardless of whether API name is `set_value` or `slice_scatter`, our goal is to have an API that is functionally dual to `paddle.strided_slice`. Here, this PR is just a subset of this feature, and we don't need to create two APIs with the same parameters. The only difference is that one supports `int` and the other supports `list of int`

megemini:
#### Update 20231224

- 修改 axis, start, stop, step 为 axes, starts, ends, strides，与 strided_slice 参数保持一致
- 修改为 list of int 后，去掉默认参数值
- 修改并增加测试用例

v2.0 RFC https://github.com/PaddlePaddle/community/pull/790

@jeff41404 @zoooo0820 请评审 ～

jeff41404:
> 关于这个 API 是否支持 list of int ，最后说一下我的看法 ～
> jeff 大佬认为 slice_scatter 与 strided_slice 是 对偶 关系，因此需要支持 ～ 这个理由实在无法说服我 ～
> * 其次，是否 `对偶` ，个人觉得应该从用户的角度出发，而不是我们决定。
>   比如说 `hsplit` 和 `hstack` 虽然参数不同，但我觉得可以是 `功能对偶` ，因为从名字上就可以很直观的理解。
>   而 `slice_scatter` 与 `strided_slice` 怎么看都看不出哪儿 `对偶` 了，至少刚接触这两个 API 的我很难想到 `对偶` ... ... 🤣🤣🤣
>   如果一定要给 `strided_slice` 找个 `对偶` 的名字，我觉得 `strided_scatter` 更合适，`strided` 表示这两个 API 是步进的，`slice` 表示 `read`，`scatter` 表示 `write` ～
>   为了 `对偶` 和极少用到的情况而给用户带来困惑，和抛异常，实在不能赞同 ～～～

The main issue you mentioned here is a naming issue. This DOES exist, the root cause is not that the name of `slice_scatter` is not good (of course changed to `slice_assign` may be better). but `paddle.strided_slice` should actually be `paddle.slice` (replacing the existing `paddle.slice`) which will be solved, so the problem you mentioned no longer exists and will not affect the design here.
As for the "rarely used cases", this judgment cannot be made unless there is sufficient data support for slicing multiple axes will be rarely used. Otherwise, we should try to provide rich functionality as much as possible.

jeff41404 (approved):
LGTM

PR #60051, opened 2023-12-15
[PIR]fix pir reduce min and reduce prod convert bug

phlrain:
### PR types
Bug fixes

### PR changes
Others

### Description
修复 reduce min 和 reduce prod 映射的bug

Pcard-67164


Aurelius84 (approved):
LGTM

PR #60125, opened 2023-12-18
[AutoParallel] Max Iteration Step for Sharding Propagation

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others 

### Description

Pcard-76459
Add Max Iteration Step for Sharding Propagation


zhiqiu (approved):
LGTM

PR #60155, opened 2023-12-19
Add arrange storage tactic

BiynXu:
### PR types
Others

### PR changes
Others

### Description
Pcard-74042


paddle/cinn/common/integer_set.cc
```diff
@@ -78,6 +86,14 @@ std::optional<bool> SymbolicExprAnalyzer::ProveNE(const ir::Expr& lhs,
   if (diff.is_constant()) {
     return diff.get_constant() != 0;
   }
+  ir::Expr diff_lower_bound = LowerBound(diff);
```
zhhsplendid: Can we use ProveEQ to  implement ProveNE?

ProveEQ returns true <==> ProveNE returns false
ProveEQ returns false <==> ProveNE returns true
ProveEQ returns optional null <==> ProveNE return optional null

paddle/cinn/common/integer_set.cc
```diff
@@ -456,5 +472,35 @@ std::optional<bool> SingleIntervalIntSet::ProveSuperSet(
   return std::nullopt;
 }
 
+ir::Expr EnhancedSimplify(
```
zhhsplendid: Change the name, it does something like replacing Mod

paddle/cinn/ir/group_schedule/tactic/arrange_storage_tactic.cc
```diff
@@ -0,0 +1,407 @@
...
+    IntSet thread_load_range = Evaluate(iter_values_of_load[i],
+                                        cuda_block_thread_fixed_var_of_load,
+                                        load_var_domain);
+    VLOG(6) << "thread_store_range of [" << iter_values_of_store[i] << "] = ["
+            << thread_store_range.Min() << " : " << thread_store_range.Max()
+            << "]";
+    VLOG(6) << "thread_load_range of [" << iter_values_of_load[i] << "] = ["
+            << thread_load_range.Min() << " : " << thread_load_range.Max()
+            << "]";
+    std::optional<bool> is_thread_sub_set =
```
zhhsplendid, line 329: Is this logic strictly right?

Suppose the following kernel code snippet:
```
int N; // tensor length is N
tensor[threadIdx.x] = input[threadIdx.x] + 9; // threadIdx.x range is 0 ~ N-1
__syncthreads();
output[threadIdx.x] = tensor[(threadIdx.x + 1) % N]; // something like shift move
```

It seems the tensor store and load have same IntSet, but we cannot use local 1 in this case. It should be shared, you  have to return CudaAxisType::kCudaThread in this case?

Although, my case may not happen in current CINN, so current logic is fine to me :-) , just a question here, this may not always mathmetically correct?

paddle/cinn/ir/group_schedule/tactic/arrange_storage_tactic.cc
```diff
@@ -0,0 +1,407 @@
...
+  std::vector<std::pair<Expr, Expr>> loads_and_blocks =
+      analyzer::GetConsumerLoadsAndSBlocks(store_block, root_block);
+
+  ir::MemoryType memory_type = ir::MemoryType::GPULocal;
+  for (const auto& load_and_block : loads_and_blocks) {
+    ir::Expr load = load_and_block.first;
+    ir::Expr load_block = load_and_block.second;
+    std::optional<CudaAxisType> cross_type =
+        AnalyzeCrossType(var2for_map, store, load, store_block, load_block);
+    if (!cross_type.has_value()) {
```
zhhsplendid, line 368: This seems to me that the logic is:

If we don't know cross Block or cross Thread, set to local

Should we do a safer logic? Such as

If we know it won't cross Block or cross Thread, set to local
if we know cross Block, throw error
else set to shared

Note: the difference of two logic is the handling of unknown case.

paddle/cinn/ir/group_schedule/tactic/arrange_storage_tactic.cc
BiynXu:
```
IntSet Evaluate(Expr expr,
                const std::unordered_map<ir::Var, ir::Var>& fixed,
                const std::unordered_map<ir::Var, IntSet>& var_domain)；
```
In this case, when we consider whether it is cross thread, threadIdx.x is fixed.
    thread_store_range: [i : i]  thread_load_range: [(threadIdx.x + 1) % N : (threadIdx.x + 1) % N]
    thread_load_range is not a subset of thread_store_range, so it is cross thread.
when we consider whether it is cross block, BlockIdx.x/y/x is fixed, threadIdx.x will be replaced with min and max.
    block_store_range: [0 : N - 1]  block_load_range: [0 : N - 1]
    block_load_range is a subset of block_store_range, so it is not cross block.
So we return CudaAxisType::kCudaThread in this case.
However, it poses a high requirement for simplifying symbolic expressions and IntSet, such as: 
replace 0 and N - 1 in [(threadIdx.x + 1) % N : (threadIdx.x + 1) % N]  ==> [0 : N]

paddle/cinn/ir/group_schedule/tactic/arrange_storage_tactic.cc
BiynXu: The cross_type without a value indicates that it does not cross any threads or blocks.
In the AnalyzeCrossType, for safety reasons, if it is uncertain whether it crosses blocks, it will be considered as crossing blocks, and the same applies to cross threads.

paddle/cinn/common/integer_set.cc
BiynXu: Modified.

paddle/cinn/common/integer_set.cc
BiynXu: Done.

zhhsplendid (approved):
LGTM

PR #60158, opened 2023-12-19
【AutoParallel】Add auto_cast support in auto_parallel

heavyrain-lzy:
### PR types

New features
### PR changes

Others
### Description

Pcard-76459
Add `amp.auto_cast` support in auto-parallelism in static graph.
#### experiment
We take the experiment using GPT with 6.7B parameters on a single host with 8 V-100 GPUs. The difference between `amp-o2` and `amp-o2+auto_cast` is as follow:
image

The result of loss:
image
image


python/paddle/static/amp/fp16_utils.py
```diff
@@ -352,6 +356,8 @@ def _is_in_black_varnames(op, amp_lists):
 
 
 def _need_keep_fp32(op, unsupported_op_list, use_fp16_guard):
+    if not op.amp_options.enable:
+        return True
```
zhiqiu: move to auto_parallel_pass

python/paddle/distributed/passes/auto_parallel_fp16.py
```diff
@@ -100,6 +122,8 @@ def _keep_fp32_input(op, in_name):
 
 # TODO check if bf16 and fp16 still share the same logic
 def _keep_fp32_output(op, out_name):
+    if not op.amp_options.enable:
+        return True
```
zhiqiu, line 128: add todo

python/paddle/distributed/passes/auto_parallel_fp16.py
```diff
@@ -188,6 +213,22 @@ def _build_state(self):
 
         return self.is_train
 
+    def _mark_cast_options(self, op):
```
zhiqiu: use a better name

python/paddle/static/amp/fp16_utils.py
heavyrain-lzy: Done

python/paddle/distributed/passes/auto_parallel_fp16.py
heavyrain-lzy: Done

python/paddle/distributed/passes/auto_parallel_fp16.py
heavyrain-lzy: Done

zhiqiu (approved):
LGTM

PR #60182, opened 2023-12-20
Optimize the execution order strategy of the program interpreter

zhangbo9674:
### PR types

Others
### PR changes

Others
### Description

Optimize the execution order strategy of the pir interpreter, which defaults to executing in the order of user networking
Pcard-67164


From00 (approved):
LGTM

PR #60187, opened 2023-12-20
[AutoParallel] complete vpp with struct name

zhaoyinglia:
### PR types

Others
### PR changes

Others
### Description

- add `StructName` in framework.py to record the call path
- use struct name to identify block that is vpp seg method
- Usage:
```python
# strategy
pipeline = strategy.pipeline
pipeline.enable = True
pipeline.schedule_mode = "VPP"
pipeline.accumulate_steps = 4
pipeline.vpp_degree = 2
pipeline.vpp_seg_method = "MyLinear"

# model
class MLP(nn.Layer):
    def __init__(self):
        self.layers = nn.LayerList([MyLinear(...) for _ in range(4)])

    def forward(self, x):
        ...
        meshes = [PP_MESH_0, PP_MESH_1, PP_MESH_0, PP_MESH_1]
        for i, layer in enumerate(self.layers):
            auto.shard_tensor(out, meshes[i], [None, None])
            out = layer(out)
        ...
```

### Others
Pcard-76459


python/paddle/base/framework.py
```diff
@@ -1133,6 +1133,59 @@ def name_scope(prefix=None):
             _name_scope = _name_scope.parent()
 
 
+class NameStruct:
```
zhiqiu, line 1136: Comment and explain NameStruct

python/paddle/distributed/auto_parallel/static/completion.py
```diff
@@ -1198,6 +1313,9 @@ def _complete_high_order_grad_annotation(self, serial_main_program=None):
             This function is temporary to support high order gradient, and will be removed in the future.
         """
 
+        if serial_main_program._appending_grad_times < 2:
+            return
```
zhiqiu: why ?

zhaoyinglia: This function is for the high-order grad model, so the normal model doesn't need to execute it.

python/paddle/base/framework.py
zhaoyinglia: Done.

zhiqiu (approved):
LGTM

PR #60201, opened 2023-12-20
[PIR] Fix llama translation

kangguangli:
### PR types

Others
### PR changes

Others
### Description

 If op may insert a data op, which happens when its two branch has different number outputs. But data op is a placeholder for executor, so its output won't have memory, they only get memory from scope by name.  Once there's no such var in scope, error happens in execution. So for these data ops who generated in  translation of if op, we need to alloc buffer and when we need allocation, its shape cannot have a -1. 
Unfortunately, such situation exists in legacy ir, so we simply transform its shape by using absolute value, this should not effect its correctness.

#### Others
Pcard-67164


yuanlehome (approved):
LGTM

PR #60225, opened 2023-12-21
[PIR] fix property overwrite

kangguangli:
### PR types

Bug fixes
### PR changes

Others
#### How many unit tests in white list is covered?
 Currently there is 294 files set with environment variables, but the total num should be `num_white_list` + `num_relax_list` + `num_no_check_list`, which is 354.

- [ ] we need to know which tests are not correctly monitored in CI

#### How many tests are disabled temporarily in this PR?

- [ ] test_adamw_op
- [ ] test_affine_grid_op
- [ ] test_conv2d_op_depthwise_conv
- [ ] test_pool3d_op
- [ ] test_conv3d_transpose_op

#### Others
Pcard-67164

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -100,9 +100,6 @@ function(set_pit_tests_properties)
         TEST ${IR_OP_TEST}
         APPEND
         PROPERTY ENVIRONMENT "FLAGS_PIR_OPTEST_WHITE_LIST=True")
-      message(STATUS "PIR OpTest: succ set ${IR_OP_TEST}")
-    else()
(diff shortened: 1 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -96,17 +96,27 @@ function(set_pit_tests_properties)
        PIR_OP_TESTS)
   foreach(IR_OP_TEST ${PIR_OP_TESTS})
     if(TEST ${IR_OP_TEST})
-      set_tests_properties(
```
kangguangli, line 99: This will lead to property overwritten by later settings and cause environment variables invaid.

winter-wang (approved):

PR #60234, opened 2023-12-21
[AutoParallel]update dist_loader when batch size is None

zhaoyinglia:
### PR types

Others
### PR changes

Others
### Description

- dist dataloder 支持 batch_size 为None的场景
### Others
Pcard-76459


JZ-LIANG (approved):
LGTM

PR #60248, opened 2023-12-22
[AutoConfig] add blacklist mechanism and scheduling in specified order

Caozhou1995:
### PR types

New features
### PR changes

Others
### Description

PCard-80244
This PR adds two import features for autotuner, namely blacklist and schedule prior oeder mechanism.

- invalid_strategy: for example, the blacklist is [stage3_mp*], it means not support hybrid parallelism of model parallelism and sharding stage3.
- schedule_prior：for example, the schedule_prior is [mp4], it means mp4 related strategies will be prioritized for running.


JZ-LIANG (approved):
LGTM

PR #60254, opened 2023-12-22
Optimize performance of advanced getitem

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985

This PR optimize the advanced index `__getitem__` performance.
- For int tensor case:  replacing the slow `broadcast_tensors` to `broadcast_shape` and `expand`. It will be much faster in case no need broadcast, and faster than `broadcast_tensors`.

- For bool tensor case:  replacing the `nonzero` + `gather_nd` to `masked_select` since the latter one is much faster. Moreover, the all-dtype support is added for kernel `MaskedSelect` and `MaskedSelectedGrad`


jeff41404 (approved):
LGTM

PR #60263, opened 2023-12-22
[Dynamic Shape] Add shape.generate_shape op

jiahy0825:
### PR types
New features

### PR changes
OPs

### Description

pcard-76996

[Dynamic Shape] Add shape.generate_shape op


jiahy0825:
See #60407

PR #60276, opened 2023-12-22
optimize slice over limit case in static

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985

For basic getitem case if slice is set but it cover the all element in the axis. e.g.`x[:10] if x.shape=(4,3,2)`. The result is same to `x`. We can add a fast path for this case to speed up models.

Different with dynamic-mode, static mode cannot deal all of this case since there is potential uncertain shape. Here we only deal with the certain shape.


jeff41404 (approved):
LGTM

PR #60314, opened 2023-12-25
[xpu] fc pass and delete pass nodes check

gitliuyf:
### PR types
Bug fixes
### PR changes
Others

### Description
modify fc pass and delete pass nodes check, when node is nullptr, node->var is also nullptr.


zhupengyang (approved):
LGTM

hong19860320 (approved):
LGTM

PR #60327, opened 2023-12-25
[PIR] Refine logic for build scope when inplace input is null

zhangbo9674:
### PR types

Bug fixes
### PR changes

Others
### Description

Refine logic for build scope when inplace input is null.
Pcard-67164


yuanlehome (approved):
LGTM

PR #60357, opened 2023-12-26
[PIR] fix data_op backend for gpu pinned tensor

feifei-111:
### PR types

Others
### PR changes

Others
### Description

if input tensor is a cuda pinned tensor, data_op will choose gpu device, however the tensor loaded is still in cpu (gpu pinned).
pd_op_to_kernel_pass will not add memcpy, the first kernel will use gpu pinned tensor as a gpu tensor, it is slow

PCard-66972


2742195759 (approved):
LGTM

PR #60428, opened 2023-12-28
【CINN】Complete dsl test case for dynamic schedule primitive

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR complete dsl test case for dynamic schedule primitive, now the dsl is completely using dynamic schedule and owns static and dynamic test cases
(attention: some primitive such as bind、vectorize、rfactor is only using on static shape, so we don't need to add new test case for them, just make sure that the original test case will run correctly on dynamic schedule)


zhhsplendid (approved):
LGTM

PR #60447, opened 2023-12-28
Fix shape error in combined-indexing setitem

zoooo0820:
### PR types

Bug fixes
### PR changes

APIs
### Description

Pcard-66985

This PR is to fix problem mentioned at https://github.com/PaddlePaddle/Paddle/issues/60376 .

In combined-indexing, if the first output dim of advanced-indexing is not at zero (i.e. the first). The value tensor should also be transposed to adapt the index_put input.

### DIFF
--- a/python/paddle/base/variable_index.py
+++ b/python/paddle/base/variable_index.py
@@ -170,7 +170,7 @@ def _setitem_for_tensor_array(var, item, value):
         )
 
 
-def deal_advanced_index(ori_tensor, indices, is_for_setitem):
+def deal_advanced_index(ori_tensor, indices, is_for_setitem, values):
     """
     Transpose origin Tensor and advanced indices to the front.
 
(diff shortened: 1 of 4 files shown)


python/paddle/base/variable_index.py
```diff
@@ -180,6 +180,7 @@ def deal_advanced_index(ori_tensor, indices, is_for_setitem):
         trans_back_dim (List): order of axes to transpose back to original order. Only used in __setitem__.
         pos_of_new_dim (int):  axis of new dim in the result. Only used in __getitem__.
         rank_of_new_dim (int): rank of new dim in the result. Only used in __getitem__.
+        transed_value_tensor (Tensor): value tensor transed to the front. Only used in __setitem__.
```
jeff41404, line 183: should be `transposed_value_tensor`? please pay attention to it in the future

jeff41404 (approved):

python/paddle/base/variable_index.py
zoooo0820: > should be `transposed_value_tensor`? please pay attention to it in the future

ok, this would be modified in another PR

PR #60452, opened 2023-12-28
Set value with scalar

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985

speed up the basic setitem avoid using `to_tensor` to construct tensor to use set_value_with_tensor op. Now will directly use set_value op


jeff41404 (approved):
LGTM

PR #60538, opened 2024-01-03
【CINN】Simplify extent of loop after fuse and add corresponding test case

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR simplify extent of fused loop after fuse and add corresponding test case by python dsl


zhhsplendid (approved):
LGTM

PR #60570, opened 2024-01-04
[PIR] Fix arange kernel bug

zhangbo9674:
### PR types

Bug fixes
### PR changes

Others
### Description

Fix arange kernel bug, output tensor type should be same with kernel type.
Pcard-67164


chen2016013 (approved):
LGTM

PR #60572, opened 2024-01-04
Add CanProveDivisible for symbolic calculation

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR add CanProveDivisible for symbolic calculation
We need this because sometimes we need to split a for loop with extent like M * N, and the factors is like [M, -1], so if We can make sure that M * N can be divisible by M, so that we don't need to add a extra if to ensure no cross-border access and get a new loop with extent like M * N  /  M +1 to ensure M*N can be divisible by M.

TODO: now the dsl don't support constructing case like above,  corresponding case should be added in the future.

### DIFF
--- a/paddle/cinn/common/integer_set.cc
+++ b/paddle/cinn/common/integer_set.cc
@@ -340,7 +340,6 @@ ir::Expr SymbolicExprAnalyzer::LowerBound(const ir::Expr& expr) const {
   BoundReplacer bound_replacer(var_intervals_, true);
   ir::Expr bound = ir::ir_utils::IRCopy(expr);
   bound_replacer(&bound);
-  std::cout << "bound is " << bound << std::endl;
   return AutoSimplify(bound);
 }
 
--- a/paddle/cinn/common/integer_set.cc
+++ b/paddle/cinn/common/integer_set.cc
@@ -178,12 +178,35 @@ std::optional<bool> SymbolicExprAnalyzer::ProveDivisible(
   ir::Expr lhs_copy = ir::ir_utils::IRCopy(lhs);
   if (cinn::common::is_zero(lhs_copy)) return true;
 
+  auto OptionalAnd = [](const std::optional<bool>& lhs,
+                        const std::optional<bool>& rhs) -> std::optional<bool> {
+    if (lhs.has_value() && rhs.has_value()) {
+      return lhs.value() && rhs.value();
+    } else {
+      return std::nullopt;
+    }
+  };
+  auto OptionalOr = [](const std::optional<bool>& lhs,
+                       const std::optional<bool>& rhs) -> std::optional<bool> {
+    if (lhs.has_value() && rhs.has_value()) {
+      return lhs.value() || rhs.value();
+    } else if ((!lhs.has_value()) && (!rhs.has_value())) {
+      return std::nullopt;
+    } else if (lhs.has_value() && (!rhs.has_value())) {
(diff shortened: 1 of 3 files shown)


paddle/cinn/common/integer_set.cc
```diff
@@ -164,11 +166,85 @@ std::optional<bool> SymbolicExprAnalyzer::ProveLT(const ir::Expr& lhs,
...
+    case cinn::ir::IrNodeTy::Add:
+      return ProveDivisible(lhs.As<ir::Add>()->a(), rhs).value_or(false) &&
+             ProveDivisible(lhs.As<ir::Add>()->b(), rhs).value_or(false) &&
+             is_ge;
+    case cinn::ir::IrNodeTy::Sub:
+      return ProveDivisible(lhs.As<ir::Sub>()->a(), rhs).value_or(false) &&
+             ProveDivisible(lhs.As<ir::Sub>()->b(), rhs).value_or(false) &&
+             is_ge;
+    case cinn::ir::IrNodeTy::Div:
+      return ProveDivisible(lhs.As<ir::Div>()->a(), rhs).value_or(false) &&
```
BiynXu: 这种情况不一定成立吧
lhs: a/b   rhs: c
a%c == 0，a/b%c不一定为0
比如a = 10, b = 2, c = 2
a%c = 0, a/b%c != 0

paddle/cinn/common/integer_set.cc
```diff
@@ -164,11 +166,85 @@ std::optional<bool> SymbolicExprAnalyzer::ProveLT(const ir::Expr& lhs,
...
+  ir::Expr zero(0);
+  bool is_positive = ProveGT(lhs, zero).value_or(false);
+  lhs_copy = (is_positive ? 1 : -1) * lhs;
+  bool is_ge = ProveGE(lhs_copy, rhs).value_or(false);
+
+  switch (lhs.node_type()) {
+    case cinn::ir::IrNodeTy::_Var_:
+      return ProveEQ(lhs, rhs);
+    case cinn::ir::IrNodeTy::IntImm:
+      return false;
```
BiynXu, line 215: 判断不了或暂时没有实现的情况返回nullopt

Courtesy-Xs: Done

paddle/cinn/common/integer_set.cc
Courtesy-Xs: done

BiynXu (approved):
LGTM

zhhsplendid (approved):
LGTM

paddle/cinn/common/integer_set.cc
```diff
@@ -164,11 +166,115 @@ std::optional<bool> SymbolicExprAnalyzer::ProveLT(const ir::Expr& lhs,
...
+      return OptionalAnd(
+          OptionalOr(ProveDivisible(lhs.As<ir::Mul>()->a(), rhs),
+                     ProveDivisible(lhs.As<ir::Mul>()->b(), rhs)),
+          is_ge);
+    case cinn::ir::IrNodeTy::Mod:
+      return false;
+    case cinn::ir::IrNodeTy::Minus:
+      return ProveDivisible(lhs.As<ir::Minus>()->v(), rhs);
+    default:
+      LOG(FATAL) << "Not supported yet!";
```
tc20042008, line 266: 我一直对这些地方的完备性担忧。

PR #60585, opened 2024-01-05
[auto parallel] add lazy init ut to llama

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
PCard-73145
add lazy init to llama ut


LiYuRio (approved):
LGTM

PR #60843, opened 2024-01-16
Adapt to dim expr

6clc:
### PR types

Others

### PR changes

Others

### Description

Pcard-78120
In the check for Int32, insert TryElevateInt32ToInt64(). This function checks the data type of ir and raises it to int64 if an expression contains an Int64.
int64 is imported only with shape expressions passed in by dim_expr.


paddle/cinn/common/ir_util.cc
```diff
@@ -143,8 +143,9 @@ Expr IndiceToAbsOffset(const std::vector<Expr> &shape,
   VLOG(3) << "indices is : " << utils::Join(indices, ",");
   CHECK_LE(shape.size(), indices.size());
   Expr res;
+  ir::TryElevateInt32ToInt64(shape);
   for (int i = 0; i < shape.size(); i++) {
-    CHECK_EQ(shape[i].type(), Int(32));
+    CHECK(shape[i].type() == Int(64) || shape[i].type() == Int(32));
```
zhhsplendid: Add hint message for check error?

paddle/cinn/ir/ir.cc
```diff
@@ -695,8 +699,11 @@ Expr Sum::Make(const std::vector<Expr> &vs) {
   if (vs.size() == 1) return vs.front();
 
   auto *n = make_shared<Sum>();
+  TryElevateInt32ToInt64(vs);
   auto type = vs.front().type();
-  for (auto &v : vs) CHECK_EQ(v.type(), type) << vs.front() << " " << v;
+  for (auto &v : vs) {
+    CHECK_EQ(v.type(), type) << "(" << v << " vs " << vs.front() << ")";
```
zhhsplendid: Add some English hint for what's wrong here.

paddle/cinn/ir/tensor.cc
```diff
@@ -227,12 +227,28 @@ isl::set _Tensor_::GenerateIslDomain() const {
...
-                          Expr(0),
-                          Sub::Make(dim, cinn::common::make_const(1)));
+        if (dim.is_constant()) {
+          dims.emplace_back(_axis_with_reduce[i]->name,
+                            static_cast<uint32_t>(0),
+                            dim.as_int32() - 1);
+        } else {
+          dims.emplace_back(_axis_with_reduce[i]->name,
+                            Expr(0),
+                            Sub::Make(dim, cinn::common::make_const(1)));
```
zhhsplendid, line 250: Should we make "make_const" with type template? Now it seems make_const only returns int32

paddle/cinn/poly/domain.cc
```diff
@@ -61,8 +61,35 @@ std::string Domain::__str__() const {
 }
 
 isl::set Domain::to_isl() const {
+  // TODO(6clc): will be removed in future
```
zhhsplendid, line 64: Please delete in the future we found we can get rid of isl

paddle/cinn/ir/tensor.cc
6clc: This code is intended to be compatible with previous int32,
and if you pass the shape property through dim_expr, you will not enter this branch of the code.

zhangbo9674 (approved):

zhhsplendid (approved):
LGTM

XiaoguangHu01 (approved):
LGTM

PR #60876, opened 2024-01-17
【CINN】Add Pass to translate buf with dyshape to buf with stshape

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This PR add add Pass to translate buf with dyshape to buf with stshape, eg.

In dynamic shape situation, some buf may owns dynamic shape, and it's memtype is local or shared in NVGPU, so we need to trans its buf size to make sure it runs correctly


paddle/cinn/optim/trans_buffer_with_dynamic_shape.cc
```diff
@@ -14,73 +14,86 @@
...
+    if (!tensor->buffer.defined()) return;
+    auto buf = tensor->buffer.As<ir::_Buffer_>();
+    if (!visited_buf_.count(buf->name)) {
+      visited_buf_.insert(buf->name);
+      auto buf_size = ir::Expr(1);
+
+      for (size_t i = 0; i < expr->as_tensor()->shape.size(); ++i) {
+        auto e = expr->as_tensor()->shape[i];
+        auto buf_e = buf->shape[i];
+        CHECK(ir::ir_utils::IRCompare(e, buf_e));
```
zhhsplendid: Is this necessary correct? I am not sure.

Courtesy-Xs: Sometimes shape of tensor is not the same as it's buffer because of reshape operation, maybe I should include situation like this.

zhhsplendid (approved):
LGTM

PR #60907, opened 2024-01-17
【Dy2S】fix NameloadJstTransformer bug

feifei-111:
### PR types

Bug fixes
### PR changes

Others
### Description

loss jst.Ld when name is arg of jst.Call
PCard-66972


SigureMo (approved):
LGTMeow

PR #60943, opened 2024-01-18
fix nvrtc usage

jeng1220:
### PR types

Bug fixes

### PR changes

Others

### Description

The phi uses nvrtc to implement JIT-compilation. However, it uses `--gpu-architecture=cmopute_` rather than `--gpu-architecture=sm_`. The difference is that compute_ only generates PTX code instead of machine code (ie, SASS). The sm_ has both.

So, the workflow is nvrtc generates PTX code, then invokes the driver to generate machine code.

The problem is if the driver version doesn't match CUDA version, the driver cannot recognize the format of PTX code, so it triggers “provided ptx was compiled with an unsupported toolchain” error. This issue cannot be resolved by “CUDA forward compatibility enabled”.

The only solutions are two:
1. Upgrade driver version. 
2. Let nvrtc generate machine code instead of PTX code

Many peoples cannot upgrade the driver frequently, so this patch is the 2nd solution.

### DIFF
--- a/paddle/phi/backends/device_code.cc
+++ b/paddle/phi/backends/device_code.cc
@@ -335,7 +335,7 @@ bool GPUDeviceCode::Compile(bool include_path) {
       DeviceContextPool::Instance().Get(place_));
   int compute_capability = dev_ctx->GetComputeCapability();
   std::string compute_flag =
-      "--gpu-architecture=compute_" + std::to_string(compute_capability);
+      "--gpu-architecture=sm_" + std::to_string(compute_capability);
   std::vector<const char*> options = {"--std=c++11", compute_flag.c_str()};
   std::string include_option;
   if (include_path) {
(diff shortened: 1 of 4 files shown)


paddle/phi/backends/device_code.cc
```diff
@@ -335,7 +335,7 @@ bool GPUDeviceCode::Compile(bool include_path) {
       DeviceContextPool::Instance().Get(place_));
   int compute_capability = dev_ctx->GetComputeCapability();
   std::string compute_flag =
-      "--gpu-architecture=compute_" + std::to_string(compute_capability);
+      "--gpu-architecture=sm_" + std::to_string(compute_capability);
   std::vector<const char*> options = {"--std=c++11", compute_flag.c_str()};
```
risemeup1, line 339: -gpu-architecture=sm_XX：这个选项会让编译器直接生成目标GPU架构的机器码（SASS）。使用sm_XX可以避免运行时的JIT编译步骤，但是这样生成的代码只能在相同架构的GPU上运行。使用compute_XX可以使得生成的代码在不同架构的GPU上都能运行。。有考虑过这个问题吗？

jeng1220:
@zyfncg, @risemeup1,
> 使用compute_XX可以使得生成的代码在不同架构的GPU上都能运行。。有考虑过这个问题吗？

`sm_XX` 能產生 PTX 和 SASS 即我代碼裡的 CUBIN，但 SASS 不能跨架構
PTX是IR，透過驅動JIT後成SASS才能執行

不過，我再深入調研了代碼，發現更本的問題不在 NVRTC 的使用
所以這 PR 不再需要，錯誤在PaddlePaddle的其他地方
我會再另題PR，先關閉這個

PR #60983, opened 2024-01-19
remove final set_value OP in combined-indexing setting

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985

For combined indexing setting case of dynamic mode, e.g. `x[1, [0,1]] = value`, we will deal it as following step:
1.  Getting a view of x, `x'=x[1]`
2. change the data of x', since it is a view of x, the data of x will also be changed.
3. overwrite the x' to x by `set_value`.

In our plan , if `x` has changed by `value`, the backward should contain it too.

But for now., in step 2, the x cannot know whether it has been changed by other operators. It will only backward to the OP **which generated it** instead of **which modified it**.

In past , the step 3 is aiming to build a backward edge between `value` and `x`. But the performance suffers because of unnecessary forward overwrite.

This PR is to optimize this case by remove the final `set_value_with_tensor`, only keeping the backward part , to build a edge between value and x.


jeff41404 (approved):
LGTM

PR #60993, opened 2024-01-20
[SOT] update logic about the judgement of "is None"

feifei-111:
### PR types

Others
### PR changes

Others
### Description

if a  variable is not ConstantVariable, it can not be None
PCard-66972


SigureMo (approved):
LGTMeow

PR #61021, opened 2024-01-22
fast pass for bool setitem

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985

optimize bool-indexing settting by using `where` instead of `nonzero + index_put`


jeff41404 (approved):
LGTM

PR #61060, opened 2024-01-23
Optimize index put preprocess

zoooo0820:
### PR types

Performance optimization
### PR changes

Others
### Description

Pcard-66985

Optimize the preprocess cost of index_put op by check if there is bool-index and reducing operators of vector<Tensor>.


jeff41404 (approved):
LGTM

PR #61090, opened 2024-01-24
[PIR][DynamicShape] Throw error when op has no InferSymbolicShapeInterface

lanxianghit:
### PR types

Others 

### PR changes

Others 

### Description

Pcard-67164
Throw error when op has no InferSymbolicShapeInterface


zhangboSJTU (approved):
LGTM

PR #61130, opened 2024-01-24
【CINN】Add DSL test case for rope of llama 

Courtesy-Xs:
### PR types

Others
### PR changes

Others
### Description

pcard-72718
This pr add DSL test case for rope of llama
Here we construct a gather + slice + concat + multiply to simulate calculation logic of rope in llama

### DIFF
--- a/test/cinn/ir/test_llir_schedule_compute_inline.py
+++ b/test/cinn/ir/test_llir_schedule_compute_inline.py
@@ -165,7 +165,7 @@ def test_reverse_compute_inline_elementwise_dynamic():
     assert_llir_equal(elementwise_add_inline, elementwise_add_inline_gt)
 
 
-def test_llama_rope_compute_inline():
+def test_gather_slice_concat_mul_pattern_compute_inline():
     class origin:
         @to_cinn_llir
         def fn_rotary(
@@ -174,7 +174,7 @@ def test_llama_rope_compute_inline():
             var_2: DataArray((1, -1, 1, 256)),
             var_3: DataArray((1, -1, 1, 512)),
             C: DataArray((1, -1, 1, 512)),
-            index: DataArray((2, 1), dtype=common.Int(32)),
+            index: DataArray((2, 1), dtype=common.Int(64)),
             Out: DataArray((1, -1, 1, 512)),
             seq_len: ir.Var(),
         ):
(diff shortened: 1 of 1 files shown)


6clc (approved):
LGTM

test/cinn/ir/test_llir_schedule_compute_inline.py
```diff
@@ -165,8 +165,124 @@ def elementwise_add_inline_gt(
     assert_llir_equal(elementwise_add_inline, elementwise_add_inline_gt)
 
 
+def test_llama_rope_compute_inline():
+    class origin:
+        @to_cinn_llir
+        def fn_rotary(
+            A: DataArray((1, -1, 1, 512)),
```
zhhsplendid, line 172: Add another test case  that replaces the first 1 to 64?

test/cinn/ir/test_llir_schedule_compute_inline.py
```diff
@@ -165,8 +165,124 @@ def elementwise_add_inline_gt(
...
+            var_1: DataArray((1, -1, 1, 256)),
+            var_2: DataArray((1, -1, 1, 256)),
+            var_3: DataArray((1, -1, 1, 512)),
+            C: DataArray((1, -1, 1, 512)),
+            index: DataArray((2, 1), dtype=common.Int(32)),
```
zhhsplendid: Can this type be Int(64), since we will use int64 for shape now? So I guess the index value should be Int(64)?

test/cinn/ir/test_llir_schedule_compute_inline.py
```diff
@@ -165,8 +165,124 @@ def elementwise_add_inline_gt(
     assert_llir_equal(elementwise_add_inline, elementwise_add_inline_gt)
 
 
+def test_llama_rope_compute_inline():
```
zhhsplendid: It is not complete rope, maybe write "xxx_xxx_pattern"

Courtesy-Xs: done

test/cinn/ir/test_llir_schedule_compute_inline.py
Courtesy-Xs: done

test/cinn/ir/test_llir_schedule_compute_inline.py
Courtesy-Xs: We should replace it with a symbol indeed, but dsl don't support it, so temporarily we use a constant shape 1, maybe this can be a TODO

zhhsplendid (approved):
LGTM

PR #60638, opened 2024-01-09
add type promotion static T+T logit.

zxcd:
### PR types

New features 

### PR changes

Others

### Description

Pcard-78266
mv type promotion static T+T logit after amp logit.

PR #59518, opened 2023-11-29
add type promotion logic for eager between tensor and tensor

zxcd:
### PR types

New features

### PR changes

Others

### Description

Card-64022
add type promotion logic for eager between tensor and tensor.

stage 1 will only support float in "+ - *, where" API.
Example like:
```
>>> a = paddle.randn((2,3), dtype='float32')
>>> b = paddle.randn((2,3), dtype='float64')
>>> out = a + b

# Before, dtype of out would be same to the former one, i.e. a
>>> out.dtype
<VarType.FP32: 5>

# Now, dtype will be promote to float64
>>> out
<VarType.FP64: 6>
```

PR #61163, opened 2024-01-25
add type promotion for complex and real number.

zxcd:
### PR Category

Others

### PR Types

New features

### Description

card-78750

As there were unreasonable type promotion in Paddle, which the previous logic was aligned to the left tensor, like:
```
a = paddle.ones([3,3], dtype= ''int64")
b = paddle.ones([3,3], dtype = 'float32')
a+b # int64
```
This behavior will be fixed more in line with mathematical logic, like:
```
a = paddle.ones([3,3], dtype= ''int64")
b = paddle.ones([3,3], dtype = 'float32')
a+b # float32
```

Furthermore, after discussion, we will limit the behavior of automatic type promotion to floating-point numbers, and between real and complex numbers in Tensor and Tensor. Tensor and Scalar will still support all dtypes. 

```
## Before:
a = paddle.ones([3,3], dtype= ''int32")
b = paddle.ones([3,3], dtype = 'int64')
a+b # int64

## After:
a = paddle.ones([3,3], dtype= ''int32")
b = paddle.ones([3,3], dtype = 'int64')
a+b # raise TypeError
```
Those PR https://github.com/PaddlePaddle/Paddle/pull/60638 , https://github.com/PaddlePaddle/Paddle/pull/59518 fixed the behavior in few APIs between floating-point numbers in Tensor and Tensor.

This PR will support all binary operation API, and type promotion between real and complex numbers in Tensor and Tensor. Also the behavior between Tensor and Scalar will  be corrected.

### DIFF
--- a/paddle/fluid/pybind/exception.cc
+++ b/paddle/fluid/pybind/exception.cc
@@ -42,7 +42,7 @@ void BindException(pybind11::module* m) {
     try {
       if (p) std::rethrow_exception(p);
     } catch (const platform::EOFException& e) {
-      eof(e.what());
+      pybind11::set_error(eof, e.what());
     } catch (const memory::allocation::BadAlloc& e) {
       PyErr_SetString(PyExc_MemoryError, e.what());
     } catch (const platform::EnforceNotMet& e) {
@@ -77,7 +77,7 @@ void BindException(pybind11::module* m) {
           PyErr_SetString(PyExc_TypeError, e.what());
           break;
         default:
-          exc(e.what());
+          pybind11::set_error(exc, e.what());
           break;
       }
     }


paddle/fluid/eager/auto_code_generator/generator/eager_gen.py
```diff
@@ -82,9 +82,30 @@
 type_promote_white_list = {
     "add": ["x", "y"],
     "subtract": ["x", "y"],
+    "divide": ["x", "y"],
```
jeff41404: why is there no multiply

paddle/fluid/eager/auto_code_generator/generator/eager_gen.py
```diff
@@ -82,9 +82,30 @@
...
+    "logical_or": ["x", "y"],
+    "logical_xor": ["x", "y"],
+    "fmax": ["x", "y"],
+    "fmin": ["x", "y"],
+    "maximum": ["x", "y"],
+    "minimum": ["x", "y"],
+    "remainder": ["x", "y"],
+    "huber_loss": ["input", "label"],
+    "nextafter": ["x", "y"],
+    "atan2": ["x", "y"],
```
jeff41404: shall we support matmul? and how can we ensure this set is complete. including all binary operation

paddle/fluid/pybind/eager_math_op_patch.cc
```diff
@@ -220,6 +220,10 @@ static PyObject* tensor__add__method(TensorObject* self,
         self_tensor = cast_ad_func(self_tensor, DataType::FLOAT32);
       }
     } else if (PyCheckInteger(other_obj) || IsNumpyType(other_obj)) {
+      if (PyCheckInteger(other_obj) && self_tensor.dtype() == DataType::BOOL) {
+        eager_gil_scoped_release guard;
+        self_tensor = cast_ad_func(self_tensor, DataType::INT64);
+      }
       other = CastPyArg2Double(other_obj, "__add__", 0);
```
jeff41404: if dtype of self_tensor is int dtype, other should not be cast to double?

paddle/fluid/pybind/eager_math_op_patch.cc
```diff
@@ -228,17 +232,66 @@ static PyObject* tensor__add__method(TensorObject* self,
       ret = CallScalarFuction(self_tensor, other, "add");
```
jeff41404: shall we overload `CallScalarFuction` when other is type of int or bool

paddle/fluid/pybind/eager_math_op_patch.cc
```diff
@@ -265,35 +318,6 @@ static PyObject* tensor__add__method(TensorObject* self,
 
   // 3. promote types or unify right var type to left var, float type promotion
   // mv to add_ad_func
```
jeff41404: are these comments also deleted synchronously?

paddle/fluid/eager/auto_code_generator/generator/eager_gen.py
zxcd: multiply not generate by eager_gen, it was separately written in multiply_fwd_func.cc.

paddle/fluid/pybind/eager_math_op_patch.cc
```diff
@@ -228,17 +232,66 @@ static PyObject* tensor__add__method(TensorObject* self,
...
+        if (self_tensor_ref.shape().size() == 0) {
+          eager_gil_scoped_release guard;
+          self_tensor_ref =
+              cast_ad_func(self_tensor_ref, other_tensor_ref.dtype());
+        } else {
+          eager_gil_scoped_release guard;
+          other_tensor_ref =
+              cast_ad_func(other_tensor_ref, self_tensor_ref.dtype());
+        }
+      }
```
jeff41404: this section is to check whether there is a 0-dimensional tensor, it's better to write as a common function to reduce duplicate code?

jeff41404:
the description above is too brief, it's better to introduce the rules for our type promotion

paddle/fluid/eager/auto_code_generator/generator/eager_gen.py
zxcd: matmul also can support, since this time not check model compatibility for matmul api, this will be added in next PR.

paddle/fluid/pybind/eager_math_op_patch.cc
zxcd: scalar_ad_func need float input, but it don't need to cast to double. I will fix it.

paddle/fluid/pybind/eager_math_op_patch.cc
zxcd: Will fix with comment above.

paddle/fluid/pybind/eager_math_op_patch.cc
zxcd: This right, I will fix it.

paddle/fluid/pybind/eager_math_op_patch.cc
zxcd: OK

jeff41404 (dismissed):
LGTM

PR #61180, opened 2024-01-25
Fix code injection in extension_utils.py

ready-research:
### PR types
Bug fixes

### PR changes
OPs

### Description
This uses shlex for safe command parsing to fix arbitrary code injection. This fixes the arbitrary code execution [vulnerability](https://huntr.com/bounties/e0ada968-edc0-4913-8e9b-45fedeba96e6/) using python stdlib tool shlex, which escapes the characters in the text and makes sure that any parsing does not happen inside a shell. Just like other fix for `_wget_download`

### DIFF
--- a/python/paddle/utils/cpp_extension/extension_utils.py
+++ b/python/paddle/utils/cpp_extension/extension_utils.py
@@ -23,7 +23,6 @@ import logging
 import os
 import re
 import subprocess
-import shlex
 import sys
 import sysconfig
 import textwrap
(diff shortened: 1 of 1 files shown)


ehtec:
People you are doing this wrong, you can't fix this issue like that. There is nothing you can do inside `run_cmd` function to fix the security problem. The `shlex.quote` needs to be used on user supplied input in all other functions calling `run_cmd`. Putting quotes around the whole command will solve nothing, it will just put quotes around the whole command and thus break the application, as you can see in the CI pipeline.

ready-research:
@ehtec @wanghuancoder Can you please review this? or suggest a way to fix this issue. Thanks.

I am not sure about check failures.

wanghuancoder (commented):
目前单测中用到了shell=True这种场景。如果要删掉shell=True。可以对command做一下字符串处理，将command split成多个参数。

ready-research:
@wanghuancoder https://github.com/PaddlePaddle/Paddle/pull/61285 is fixing this issue?

PR #61185, opened 2024-01-25
[Dy2s] fix bug when set value in cond

feifei-111:
### PR types

Bug fixes
### PR changes

Others
### Description

fix bug when set value in cond
PCard-66972


2742195759 (approved):
LGTM

PR #61187, opened 2024-01-25
[Prim][PIR]simplify full in prim

cyber-pioneer:
### PR types
Performance optimization

### PR changes
Others

### Description
Pcard-66975
Simplify full logic involved if broadcast case for composite rule.


cxxly (approved):
LGTM

PR #61339, opened 2024-01-30
[Inference] All op involving float64 precision must not run in low precision in amp pass

yuanlehome:
### PR types

Bug fixes
### PR changes

Others
### Description

All op involving float64 precision must not run in low precision in amp pass. The main reason is to ensure accuracy.
### Pcard-71500


zhoutianzi666 (approved):
LGTM

PR #61351, opened 2024-01-30
[SOT] fix bugs in sot export

feifei-111:
### PR types

Others
### PR changes

Others
### Description

1. fix sot export error when input dtype is bool
2. fix bug that api string might not correct (e.g. for paddle.abs)
3. add paddle.seed before run net case

PCard-66972


Aurelius84 (approved):
LGTM

PR #61356, opened 2024-01-30
remove _wget

lyuwenyu:
### PR types

Others
### PR changes

Others
### Description

- remove _wget in download

PR #61362, opened 2024-01-30
Fix arbitrary file write vulnerability in _wget_download function

m0kr4n3:
### PR types
Bug fixes 

### PR changes
Others

### Description
Add 2 checks:
- checks if the fullname is empty.
- ensures it's not a path.


VigiZhang (commented):
[61356](https://github.com/PaddlePaddle/Paddle/pull/61356)

VigiZhang (commented):
We prefer to remove this function from the utils package. [61356](https://github.com/PaddlePaddle/Paddle/pull/61356)

PR #61395, opened 2024-01-31
[PIR][DynamicShape] Add Python ut for repeate kv

lanxianghit:
### PR types

Others 

### PR changes

Others 

### Description

Pcard-67164
Add infer symbolic shape Python ut for repeate kv


zhangboSJTU (approved):
LGTM

PR #61416, opened 2024-02-01
[PIR] Support reshape op to get 'shape' from stack op

chen2016013:
### PR types

Others

### PR changes

Others

### Description

Pcard-67164
支持reshape op从stack op的输出中获取shape

### DIFF
--- a/paddle/fluid/pir/dialect/operator/utils/utils.cc
+++ b/paddle/fluid/pir/dialect/operator/utils/utils.cc
@@ -439,5 +439,56 @@ void CheckDataTypeOrValue(const phi::DataType& dtype,
   }
 }
 
+std::vector<int64_t> ParseValueShape(const pir::Value& shape_,
+                                     bool* is_from_tensor) {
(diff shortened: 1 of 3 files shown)


paddle/fluid/pir/dialect/operator/utils/utils.cc
```diff
@@ -439,5 +439,56 @@ void CheckDataTypeOrValue(const phi::DataType& dtype,
   }
 }
 
+std::vector<int64_t> ParseValueShape(const pir::Value& shape_,
```
zhangbo9674, line 442:
```suggestion
std::vector<int64_t> ParseValueShape(const pir::Value& shape_,
std::vector<int64_t> ParseValueShape(const pir::Value& shape,
```

zhangbo9674 (approved):

paddle/fluid/pir/dialect/operator/utils/utils.cc
chen2016013: 单独提PR更改

PR #61448, opened 2024-02-01
fix random hang in non-blocking-queue

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Description

fix random hang in non-blocking-queue

First, `num_tasks_` will be read/write both in two different threads, which may leads to data inconsistence, though it is atomic.
And, `ec_.Notify(false);` is able to `fast return` when no threads are waiting with a little time cost, so remove `num_tasks_`.

Pcard-76459


zhangbo9674 (approved):

From00 (approved):
LGTM

PR #61489, opened 2024-02-02
[CINN] add annotation when calling bucket lower

BiynXu:
### PR types
Others

### PR changes
Others

### Description
Pcard-72511
Add annotation when calling bucket lower.


Courtesy-Xs (approved):
LGTM

PR #61571, opened 2024-02-04
[Prim][PIR]Add composite rule of tile for both static and dynamic shape

cyber-pioneer:
### PR types
New features

### PR changes
Others

### Description
Pcard-66975
Add composite rule of tile for both static and dynamic shape.


Charles-hit (dismissed):

paddle/fluid/primitive/composite/composite.h
```diff
@@ -25,6 +26,11 @@ namespace details {
 // empty_shape means x.shape=[]
 static std::vector<int64_t> empty_shape;
 
+template <typename T>
+static Tensor get_slice(const Tensor& x, int64_t idx) {
```
Charles-hit, line 30: 这种工具函数后续最好提到一个公共文件里

Charles-hit (approved):

PR #61840, opened 2024-02-20
[AutoParallel] Fix problems of pp.

GhostScreaming:
### PR types
Bug fixes

### PR changes
Others

### Description
Pcard-73145
1. `full_` should set a zero-memory DenseTensor for inplace output in pipeline parallel.
2. Operators (like `fused_rotary_position_embedding` and `flash_attn`) which have optional but not inplace output should set correct dtype in their InferMeta functions. Or `SetGradInMeta` will have `phi::DataType::UNDEFINED` error.
3. Fix problem of `split_with_num_grad` when it calls `FillZeroForEmptyGradInput`.


liuzhenhai93 (approved):
LGTM

zhiqiu (approved):

chenwhql (approved):

PR #61858, opened 2024-02-20
[SOT] avoid trace create layer tracker

feifei-111:
### PR types

Others
### PR changes

Others
### Description

avoid trace create layer tracker
PCard-66972


SigureMo (approved):
LGTMeow

PR #61875, opened 2024-02-20
cinn(test): fix if op

6clc:
### PR types

Others

### PR changes

Others

### Description

Pcard-78120
The nn.ReLU api does not work in dynamic to static, resulting in the error shown below, not ReLU op in pir program
image


jiahy0825 (approved):
LGTM

tc20042008 (approved):

PR #62015, opened 2024-02-23
【AutoParallel】Enable amp strategy in `dist.to_static`

heavyrain-lzy:
### PR types

Others
### PR changes

Others
### Description

Pcard-76459
1. Pass the amp strategy in `dist.to_static`  for `static graph` to using amp strategy in semi-autoparallelism.
2. Support initializing model with `dtype==float32`, and use amp strategy in scenarios where `to_static` is enabled
3. Fix the bug in the `auto_parallel_fp16` pass when the model using `cast` operator from `float32` to `half`


zhiqiu (approved):
LGTM

PR #62082, opened 2024-02-26
fix bug that input and output is same tensor

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Description

fix bug that input and output is same tensor
Pcard-76459


LiYuRio (approved):
LGTM

PR #62100, opened 2024-02-26
[AutoParallel] Fix bug when set master_grad

heavyrain-lzy:
### PR types

Bug fixes
### PR changes

Others
### Description

Pcard-76459
Fix the bug when setting the `master_grad` but the `grad.get_impl()` is `nullptr`.


From00 (approved):
LGTM

PR #62103, opened 2024-02-26
fix the unqiue op that generate the wrong inreverse result

wawltor:
### PR types
Bug fixes

### PR changes
OPs 
### Description
fix the unqiue op that generate the wrong the inreverse result


DesmonDay (approved):
LGTM

PR #62171, opened 2024-02-28
clean legacy code of spmd

zhiqiu:
### PR types

Others
### PR changes

Others
### Description

clean legacy code of spmd， which is moved to phi
Pcard-76459


JZ-LIANG (approved):
LGTM

PR #62274, opened 2024-03-01
[Don't Merge] Debug xpu weight only linear

sljlp:


paddle/phi/kernels/xpu/weight_only_linear_kernel.cc
```diff
@@ -0,0 +1,129 @@
...
+            n,
+            1.0f,
+            0.0f,
+            bias.is_initialized()
+                ? (bias.get().dtype() == phi::DataType::FLOAT16
+                       ? bias_fp32.data<float>()
+                       : bias.get().data<float>())
+                : nullptr,
+            baidu::xpu::api::Activation_t::LINEAR,
+            weight_scale.data<float>());
```
ZhouMengLei1999: weight_scale和bias是同类型的

PR #62376, opened 2024-03-04
Fix ShapeOrDataDimExpr simplify unwork

JiaWenxuan:
### PR types

Bug fixes

### PR changes

Others

### Description

pcard-76996
This pr amend the SubstituteDimExprBasedOnConstraintsPass and SimplifyDimExprPass to handle the make it work.
In detail:
- move these two pass below the last symbolic inferring
- use pointer to change shape_analysis itself instead of the copy one
- added support for GroupOp or other op which include other ops with recursive proccess


paddle/cinn/hlir/dialect/operator/transforms/group_merge/simplify_dim_expr_pass.cc
```diff
@@ -27,12 +27,29 @@ namespace ir {
 
 namespace {
 
+template <typename DoEachT>
+void VisitEachSubOp(pir::Operation* op, const DoEachT& DoEach) {
```
jiahy0825: `VisitEachSubOp -> VisitEachRegion`

jiahy0825 (approved):
LGTM

tc20042008 (approved):

zyfncg (approved):

PR #62381, opened 2024-03-04
[PIR][DynamicShape] Add strategy for compatibility in select_input op

lanxianghit:
### PR types

Others 

### PR changes

Others 

### Description

Pcard-67164
Add strategy for compatibility in select_input op


cyber-pioneer (approved):
LGTM

PR #62499, opened 2024-03-07
bugfix, add skipIf for multigpu tests

jeng1220:
### PR Category
Others

### PR Types
Bug fixes

### Description
Not all CI/CD machines have 8 GPUs, so this PR adds `skipif` if number of GPUs is not enough

### DIFF
--- a/test/collective/test_communication_api_base.py
+++ b/test/collective/test_communication_api_base.py
@@ -22,9 +22,14 @@ import sys
 import tempfile
 import unittest
 
+import paddle
+
 
 class CommunicationTestDistBase(unittest.TestCase):
     def setUp(self, save_log_dir=None, num_of_devices=2, timeout=120, nnode=1):
+        if num_of_devices > paddle.device.cuda.device_count():
+            self.skipTest("number of GPUs is not enough")
+
         self._python_interp = sys.executable
         self._save_log_dir = save_log_dir
         self._log_dir = tempfile.TemporaryDirectory()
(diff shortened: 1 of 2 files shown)


test/collective/test_communication_api_base.py
```diff
@@ -22,9 +22,14 @@
 import tempfile
 import unittest
 
+import paddle
+
 
 class CommunicationTestDistBase(unittest.TestCase):
     def setUp(self, save_log_dir=None, num_of_devices=2, timeout=120, nnode=1):
+        if num_of_devices > paddle.device.cuda.device_count():
+            self.skipTest("number of GPUs is not enough")
```
LiYuRio, line 31: I don't think need this if-condition here because all the test-case inherited this base class only use 2-cards. And CI machines which run these test-cases only has two devices either.

jeng1220: @LiYuRio ,

Thanks for the response. 
> CI machines which run these test-cases only has two devices either.

Inside Baidu, it is true but for other developers, it might not be true. This line can give other developers a simple, more meaningful reason to show why unit tests cannot run, instead of a long bug message. Does it make sense?

LiYuRio: ok, but if we add skipTest here, is it necessary to add skipTest to all distributed test case? Not only in communication test case.

JZ-LIANG (dismissed):
LGTM

test/collective/test_communication_api_base.py
jeng1220: @LiYuRio ,
Yes, it will be best. I just found another test (test/legacy_test/op_test.py) has the same issue. This is all I have found for now.

LiYuRio (approved):

PR #62594, opened 2024-03-10
cinn(op): fix broadcast op

6clc:
### PR types
Others

### PR changes
Others

### Description

Pcard-78120
fix broadcast op case as below
input tensor: [4]
output tensor: [3, 4]


zyfncg (dismissed):

winter-wang (approved):
LGTM

PR #62610, opened 2024-03-11
[Prim] Remove unnecessary variable such as 'need_skip' and redundant next_grad_node_creation code

HydrogenSulfate:
### PR types

Function optimization
### PR changes

OPs 
### Description

Pcard-75624

Remove unnecessary variable such as 'need_skip' and redundant next_grad_node_creation code.

If one op in prim_white_list, that means:
1. That op must execute composite op, **so there is no need to create next GradNode**, and **no need to controled by `need_skip`**.

besides, support `multiply_grad` and `multiply_double_grad` be controled by `prim_blacklist`


heavyrain-lzy (approved):
LGTM

Charles-hit (approved):

PR #62629, opened 2024-03-11
【AutoParalle】Clear origin param if using amp-o2 in `dy2st` case of auo-parallel.

heavyrain-lzy:
### PR types

Function optimization
### PR changes

Others
### Description

Pcard-76459
Clear origin param if using amp-o2 in `dy2st` case of auo-parallel to optimize the memory.


zhiqiu (approved):
LGTM

PR #62631, opened 2024-03-11
skip normal reset log to avoid mislead user

zhiqiu:
### PR types

Others
### PR changes

Others
### Description

If recieve bytes is 0, it means the end of tcp connection.
To avoid the following logs when `GLOG_v > 5`
image
Pcard-76459


LiYuRio (approved):
LGTM

PR #62713, opened 2024-03-14
[OneDNN][PIR] conv elementwise add mkldnn fuse pass

zhanglirong1999:
### PR types
New features

### PR changes
Others

### Description
We rewrite new conv_elementwise_add_mkldnn_fuse_pass for pir.

### DIFF
--- a/paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "paddle/fluid/pir/transforms/onednn/batch_norm_act_fuse_pass.h"
+#include "paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.h"
 
 #include "paddle/fluid/pir/dialect/operator/ir/onednn_op.h"
 #include "paddle/fluid/pir/dialect/operator/ir/pd_op.h"
--- a/paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
@@ -207,6 +207,7 @@ class ConvElementwiseAddFusePass : public pir::PatternRewritePass {
         context,
         paddle::dialect::Conv2dOp::name(),
         paddle::onednn::dialect::FusedConv2dOp::name()));
+    return ps;
   }
 };
 
--- a/paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
@@ -204,16 +204,26 @@ class FusedConvBiasElementwiseAddPattern : public paddle::drr::DrrPatternBase {
 
   void operator()(paddle::drr::DrrPatternContext *ctx) const override {
     paddle::drr::SourcePattern pat = ctx->SourcePattern();
-    const auto &conv =
-        pat.Op(conv_name_,
-               {{
-                   {"strides", pat.Attr("strides")},
-                   {"paddings", pat.Attr("paddings")},
-                   {"padding_algorithm", pat.Attr("padding_algorithm")},
-                   {"dilations", pat.Attr("dilations")},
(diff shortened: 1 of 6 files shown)


paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,225 @@
...
+
+    const auto &add = pat.Op(paddle::dialect::AddOp::name());
+    conv({&pat.Tensor("input"), &pat.Tensor("filter")},
+         {&pat.Tensor("conv2d_out")});
+
+    pat.Tensor("add_out") =
+        add(pat.Tensor("conv2d_out"), pat.Tensor("residual_param"));
+    pat.RequireNativeCall(
+        [](const paddle::drr::MatchContext &match_ctx) -> bool {
+          if (!pir::ValueIsPersistable(match_ctx.Tensor("residual_param"))) {
```
LLee233: Since only weight is constant-needed, I think for elementwise_add, it's appropriate not to check the consistance of "residual_param"?

paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,225 @@
...
+
+    const auto &add = pat.Op(paddle::dialect::AddOp::name());
+    conv({&pat.Tensor("input"), &pat.Tensor("filter")},
+         {&pat.Tensor("conv2d_out")});
+    pat.Tensor("add_out") =
+        add(pat.Tensor("residual_param"), pat.Tensor("conv2d_out"));
+
+    pat.RequireNativeCall(
+        [](const paddle::drr::MatchContext &match_ctx) -> bool {
+          if (!pir::ValueIsPersistable(match_ctx.Tensor("residual_param"))) {
```
LLee233: Likewise

vivienfanghuagood (approved):

yuanlehome (dismissed):

zhanglirong1999:
We remove conv + bias + bias pattern in the previous pass, since no such situation, and it will not fusion conv + add + add when we testing. Thus we just need conv + bias + residual_input.

paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,405 @@
...
+                   {"mkldnn_data_type", res.StrAttr("float32")},
+                   {"fuse_activation", res.StrAttr("")},
+                   {"fuse_residual_connection", res.BoolAttr(true)},
+                   {"force_fp32_output", res.BoolAttr(false)},
+                   {"fuse_alpha", res.Float32Attr(0.0f)},
+                   {"fuse_beta", res.Float32Attr(0.0f)},
+                   {"scale_in", res.Float32Attr(1.0f)},
+                   {"scale_out", res.Float32Attr(1.0f)},
+                   {"scale_in_eltwise", res.Float32Attr(1.0f)},
+                   {"scale_weights", res.VectorFloatAttr({1.0f})},
```
LLee233: From my perspective, if the entrance op is a fused_conv, then when you create a new fused_conv with add, the above "fused_attributes" should be grasped from the former fused_op instead of just setting default value, since you don't know if these attrs are already set or not.

LLee233: It's safe if we have only "conv_bias_fuse_pass" and "conv_eltwise_add_fuse_pass" here, but not if we have other passes at former place, especially when they will output fused_conv as well.

paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
```diff
@@ -0,0 +1,405 @@
...
```
LLee233: Likewise

paddle/fluid/pir/transforms/onednn/conv_elementwise_add_mkldnn_fuse_pass.cc
zhanglirong1999: yes, since I notice in the previous pass order, just conv_bias_fuse_pass before this pass to set these attribute. Therefore, I think there is no previous pass that will affect the value here. I will check and think again.

LLee233 (approved):
LGTM

yuanlehome (approved):

wanghuancoder (approved):
LGTM

PR #62809, opened 2024-03-18
fix rocksdb compile error on gcc84

qili93:
### PR types

Bug fixes

### PR changes

Others

### Description

Fix rocksdb compile error on high gcc versions, gcc84 clang15 etc.
Pcard-77889


qili93:
Close as fixed in https://github.com/PaddlePaddle/Paddle/pull/63307

PR #62818, opened 2024-03-18
support quant other layer except quanted model

ceci3:
### PR types

New features

### PR changes

APIs

### Description

support quant other layer except quanted model


wanghaoshuang (approved):
LGTM

PR #62856, opened 2024-03-19
fix local buffer resize

BiynXu:
### PR types
Bug fixes

### PR changes
Others

### Description
fix local buffer size, resize to constant.
Pcard-74042

### DIFF
--- a/paddle/cinn/optim/resize_buffer.cc
+++ b/paddle/cinn/optim/resize_buffer.cc
@@ -106,10 +106,8 @@ class AnalyzeLoopVarRange : public ir::IRMutator<> {
  private:
   void AnalyzeTensorRange(const std::vector<Expr>& indices,
                           const ir::Tensor& tensor) {
-    if (!tensor->buffer.defined() ||
-        tensor->buffer->memory_type == ir::MemoryType::Heap) {
-      return;
-    }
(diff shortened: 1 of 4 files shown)


paddle/cinn/optim/resize_buffer.cc
```diff
@@ -144,6 +147,47 @@ class AnalyzeLoopVarRange : public ir::IRMutator<> {
             << buffer_name_to_indice_extent[buffer_name];
   }
 
+  void AnalyzeBufferSize(const std::vector<Expr>& indices,
+                         const ir::Tensor& tensor) {
+    if (!tensor->buffer.defined() ||
+        tensor->buffer->memory_type == ir::MemoryType::Heap) {
+      return;
+    }
```
tc20042008:
```c++
if (!tensor->buffer.defined() ) return;
if (tensor->buffer->memory_type == ir::MemoryType::Heap) return;
```

jiahy0825 (dismissed):
LGTM

paddle/cinn/optim/resize_buffer.cc
BiynXu: done

6clc (approved):

PR #62943, opened 2024-03-22
[PIR] Move port.h port.cc to phi/common/

changeyoung98:
### PR types

Others
### PR changes

Others
### Description

pcard-67164
Move port.h port.cc from `phi/backends/dynload/`  to `phi/common/` so that it can be used in pir


XiaoguangHu01 (approved):
LGTM

PR #62948, opened 2024-03-22
[XPU] add useless pad op when padding all zero to identity_op pass;

newway:
### PR Category

Performance Optimization

### PR Types

New features

### Description

删除无效的pad3d op, 如果padding参数都是0, 是一个无效op


zhupengyang (dismissed):
LGTM

PR #63009, opened 2024-03-25
[Prim] Use pow_ad_func instead of elementwise_pow_ad_func when y is scalar

HydrogenSulfate:
### PR types

Performance optimization

### PR changes

OPs 

### Description

Pcard-75624

Use `pow_ad_func(x, y)` instead of `elementwise_pow_ad_func(x, ::full_like_ad_func(x, y))` when index `y` is of `Scalar`


cxxly (approved):
LGTM

PR #63011, opened 2024-03-25
[Prim] Use scale(Tensor, a, b) instead of Tensor*a+b in composite_double_backward_api.h

HydrogenSulfate:
### PR types

Performance optimization

### PR changes

OPs 

### Description

Pcard-75624

Use `scale(Tensor, a, b)` instead of `Tensor * a + b` to reduce redundant call of kernels.


cxxly (approved):
LGTM

PR #63049, opened 2024-03-27
【AutoParallel】Fix 'hang' using Sharding in unified model

heavyrain-lzy:
### PR Category

Distributed Strategy
### PR Types

Others

### Description

Pcard-76459
1. Solve the problem of `hang` when sharding strategy is enabled in unified networking
2. Fix the bug in `auto_parallel_sharding.py` when using `stage3`

### DIFF
--- a/python/paddle/distributed/auto_parallel/static/engine.py
+++ b/python/paddle/distributed/auto_parallel/static/engine.py
@@ -1083,22 +1083,16 @@ class Engine:
                 uninitialized.append(var)
             # Make sure the number of communication operators is consistent
             commu_ops = []
-            dist_startup_vars = dist_startup_prog.global_block().vars
             for op in dist_startup_prog.global_block().ops:
                 if auto_utils.is_comm_op(op):
-                    input_name = op.input_arg_names[0]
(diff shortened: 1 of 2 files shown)


python/paddle/distributed/auto_parallel/static/engine.py
```diff
@@ -1081,8 +1081,17 @@ def _initialize(self, mode, init_parameters=True):
                 if scope_var and scope_var.get_tensor()._is_initialized():
                     continue
                 uninitialized.append(var)
-            if uninitialized:
-                prune_startup_prog = dist_startup_prog._prune(uninitialized)
+            # Make sure the number of communication operators is consistent
+            commu_ops = []
```
zhiqiu, line 1085: commu -> comm, plz fix it in the next pr

zhiqiu (approved):
LGTM

python/paddle/distributed/auto_parallel/static/engine.py
heavyrain-lzy: > commu -> comm, plz fix it in the next pr

Get it! Thanks!

PR #63097, opened 2024-03-29
[pir+auto parallel] translate reshard_op into comm and compute op

zhiqiu:
### PR Category

Auto Parallel

### PR Types

New features

### Description

[pir+auto parallel] translate reshard_op into comm and compute op
Pcard-76459


JZ-LIANG (approved):
LGTM

PR #63113, opened 2024-03-29
fix auto backward bug

winter-wang:
### PR Category

Others

### PR Types

Bug fixes
### Description

- fix the bug in pir auto backward.
  while the stop_gradient of all inputs is true, we should skip calling the vjp interface.

### DIFF
--- a/python/paddle/autograd/ir_backward.py
+++ b/python/paddle/autograd/ir_backward.py
@@ -643,9 +643,6 @@ def append_backward_ops(
                         len(output_grads) == 0
                         or all(zero_flag)
                         or all_output_grad_none(output_grads)
-                        or all_input_stop_gradient_true(
-                            input_grad_stopgradients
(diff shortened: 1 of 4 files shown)


zhiqiu (dismissed):
LGTM

python/paddle/autograd/ir_backward.py
```diff
@@ -649,6 +650,14 @@ def append_yield(
                     ]:
                         continue
 
+                    if all_input_stop_gradient_true(
```
xiaoguoguo626807, line 653: 加一下注释吧

xiaoguoguo626807 (approved):

python/paddle/autograd/ir_backward.py
winter-wang: 好的，下个pr补充相关说明。

PR #63117, opened 2024-03-29
[AutoParallel-PIR] MLP Tensor Parallel Demo

JZ-LIANG:
### PR Category

Auto Parallel 

### PR Types

New features

### Description

Pcard-67164

Add Dist2Dense Pass, Dense here means non-distributed.
All distributed related info (dist_attr) would be remove after this pass, and all Passes after this pass will regard the Program as single-rank(local) Program (Tensor Fusion, etc). 

cherry-pick from #63148


paddle/pir/include/core/operation.h
```diff
@@ -113,6 +113,7 @@ class IR_API alignas(8) Operation final
   void set_attribute(const std::string &key, Attribute value) {
     attributes_[key] = value;
   }
+  void erase_attribute(const std::string &key) { attributes_.erase(key); }
```
winter-wang, line 116:
```suggestion
  void EraseAttribute(const std::string &key) { attributes_.erase(key); }
```

winter-wang (approved):

XieYunshen (approved):
LGTM for `set_tests_properties(test_mlp PROPERTIES LABELS "RUN_TYPE=EXCLUSIVE" TIMEOUT
                                           60)`

PR #63232, opened 2024-04-04
Add enable_onednn api in AnalysisConfig

co63oc:
### PR Category

Others

### PR Types

Others

### Description

增加 AnalysisConfig api接口 enable_onednn
用来替换 enable_mkldnn


qingqing01 (commented):
Thanks for your contribution！
We have no plans to abandon the mkldnn interface at the moment. It's confusing for uers to add the onednn interface now. Maybe,  we can consider to change the name when upgrading the Predictor APIs in the future.

PR #63290, opened 2024-04-07
fit completion rule for swiglu op

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Performance

### Description

fit completion rule for swiglu op
Pcard-76459


JZ-LIANG (approved):
LGTM

PR #63293, opened 2024-04-07
【AutoParallel】Fix type_promote of 'elementwise_add'

heavyrain-lzy:
### PR Category

Performance Optimization

### PR Types

Improvements

### Description

Pcard-76459
There is no need to do data type promotion for `elementwise_add` in the case "float + half".


zhiqiu (approved):
LGTM

PR #63320, opened 2024-04-08
[CINN] fix bug in broadcast_with_cf.cc

Fridge003:
### PR Category

CINN

### PR Types

Bug fixes

### Description

Fix bug: the initialization of DimExpr vector is incorrect at the `UpdateGroupShapeExprs` function of `broadcast_with_cf.cc`.

Add test for dynamic shape scenario when using CINN.

PCard-76996


feifei-111 (approved):
LGTM

PR #63423, opened 2024-04-11
add copy code for learing rate in different mesh.

winter-wang:
### PR Category

Auto Parallel

### PR Types

New features

### Other
pcard-67164


JZ-LIANG (approved):
LGTM

PR #63536, opened 2024-04-15
【CINN】Fix reduce sum bug when just using one warp

Vvsmile:
### PR Category

CINN

### PR Types

Bug fixes

### Description

pcard-70457
Fix the reduce sum bug when just using one warp.
The reduce_sum of 【2，8】in CINN will produce results containing None, this is due to just using one warp resulting in different argument "warp_num" in the SplitWarpNumber procedure.
This PR fixes this bug and changes the corresponding BindCudaInfo to ensure the right result and better performance.


Vvsmile:
The fix task is also done by the other PR, closing.

PR #63545, opened 2024-04-15
fix. allreduce_sum is needed for matmul_v2_grad when the Y@GRAD not a parameter but from reshard

liym27:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

In static auto parallel backward, allreduce_sum op should be appended to program in the situation:
```
x = dist.reshard(weight, ...)
y = paddle.matmul(z, x)
```

Pcard-76459


JZ-LIANG (dismissed):
LGTM

zhiqiu (approved):
LGTM

PR #63594, opened 2024-04-17
[Bug Fix] Bug fix for results of paddle.ones elements are all zero.

zhangboSJTU:
### PR Category

Others 

### PR Types

Bug fixes

### Description

Pcard-67164

Bug fix for the following cases:

```python 
import paddle
paddle.set_default_dtype("bfloat16")
logps = paddle.ones(shape=[32, 100352, 1024])
print(logps)
```

The results should be a tensor with all elements one, but got result tensor whose elements are all zero.


wawltor (approved):
LGTM

PR #63595, opened 2024-04-17
Resolved a conflict between import torch and paddle

idontkonwher:
### PR Category

User Experience

### PR Types

Bug fixes

### Description

As metioned in #56812, import paddle and torch together will cause errors, it's mainly beacuse of the pybind global veriable conflict.


wanghuancoder (approved):
LGTM

PR #63646, opened 2024-04-18
use eager tensor's is_initialized directly

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

use eager tensor's is_initialized directly
Pcard-76459


JZ-LIANG (approved):
LGTM

PR #63681, opened 2024-04-19
[Auto Parallel] Add a case when the target shape  is 1 in reshape spmd rule.

pkuzyc:
### PR Category
Auto Parallel

### PR Types
Bug fixes

### Description
Pcard-76459
Cover the following case in reshape spmd rule:
     shape: [1, 512, 4096] --> [1, 2, 256, 4096],
     input dims_mapping: [0, 1, -1]
     expected output dims_mapping: [0, 1, -1, -1] (not [-1, 1, -1,-1])

In this case, the dim0 in target shape is 1 and it is from dim0 in source shape. make the dim0's transformation be InputDim rather than Singleton so that the sharding status can be propagated.


zhiqiu (approved):
LGTM

PR #63160, opened 2024-04-01
support customized base for fused_rope

deepllz:
### PR Category

Others

### PR Types

Others

### Description

fused_rope支持自定义base
测试发现将rotary_base通过参数传入到kernel会导致cuda kernel降速，所以对默认的10000 base，特化模板，保证base=10000的时候，kernel不降速。
Pcard-73145

PR #63706, opened 2024-04-19
[XPU] fix xpu rotary test bug

HarperCy:
### PR Category
Operator Mechanism

### PR Types
Bug fixes

### Description
rotary was added a new parameter rotary_base in https://github.com/PaddlePaddle/Paddle/pull/63160, resulting in the occurance of error when testing xpu version of rotary operator. However, xpu does not support null-cos & null-sin inputs, yet. i.e. xpu version of rotary (or to say, rope) does not support the functionality of initializing cos and sin within that single operator. To solve this test error quickly, I choose to it in a nasty way by modifying the interface solely. This will not hurt the function integrity of the original xpu operator and the full calculating process will be supported very soon.


XiaociZhang (commented):
LGTM

ZibinGuo (approved):
LGTM

houj04 (approved):

PR #63751, opened 2024-04-22
[CINN] Support horizontal fusion

jiahy0825:
### PR Category
CINN

### PR Types
Performance

### Description

pcard-76996

Support horizontal fusion without unittest
- Unittest will be merge in #63415


jiahy0825:
- See #63773

PR #63753, opened 2024-04-22
Fix dy2st no need buffer error when inplace

0x45f:
### PR Category

Performance Optimization

### PR Types

Bug fixes

### Description

Fix dy2st no need buffer error when inplace
Pcard-67164

### DIFF
--- a/paddle/fluid/pybind/pir.cc
+++ b/paddle/fluid/pybind/pir.cc
@@ -1209,6 +1209,23 @@ std::optional<pir::Value> FindInplaceSource(
   return std::nullopt;
 }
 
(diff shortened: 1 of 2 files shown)


paddle/fluid/pybind/pir.cc
```diff
@@ -1209,6 +1209,23 @@ std::optional<pir::Value> FindInplaceSource(
   return std::nullopt;
 }
```
SigureMo, line 1211: 上面的函数可以用这个新的简化下～

0x45f: 好的，感谢~

SigureMo (approved):
LGTMeow

PR #63798, opened 2024-04-23
fix reshard of empty tensor and refine local shape calculation

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

As the title

The following case raise Error as below,
```
a = paddle.rand([16, 32])
dist_a = dist.shard_tensor(a, mesh, [dist.Shard(0)])
b = paddle.rand([16, 32])
dist_b = dist.shard_tensor(b, mesh, [dist.Shard(1)])

a.stop_gradient = False
b.stop_gradient = False
dist_a.stop_gradient = False
dist_a.stop_gradient = False

sum = paddle.add(dist_a, dist_b)

#####
PreconditionNotMetError: Tensor's dimension is out of bound.Tensor's dimension must be equal or less than the size of its memory.But received Tensor's dimension is 1024, memory's size is 0.
  [Hint: Expected numel() * SizeOf(dtype()) <= memory_size(), but received numel() * SizeOf(dtype()):1024 > memory_size():0.] (at /paddle/paddle/phi/core/dense_tensor_impl.cc:55)

I0425 11:07:20.154877 21600 tcp_store.cc:293] receive shutdown event and so quit from MasterDaemon run loop
LAUNCH INFO 2024-04-25 11:07:21,250 Exit code 1
```
This PR fixes that.

Pcard-76459

### DIFF
--- a/paddle/phi/api/yaml/generator/dist_api_gen.py
+++ b/paddle/phi/api/yaml/generator/dist_api_gen.py
@@ -49,6 +49,8 @@ MAIN_DIST_BRANCH_TEMPLATE = """
     // 1. InferSpmd (Infer DistAttr of Inputs&Outputs){}
     // 2. Create API Output & Prepare Dist and Dense Output{}
     // 3. Infer DistTensor's Global Shape{}\n
+    // 3.1. Set Output Dist Attr For Default Impl{}\n
+    VLOG(1) << 21111111;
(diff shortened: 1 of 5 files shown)


Aurelius84 (approved):

heavyrain-lzy (approved):
LGTM

LiYuRio (approved):

PR #63939, opened 2024-04-28
Fix backward program kwargs error when process inplace value

0x45f:
### PR Category

Others

### PR Types

Bug fixes

### Description

Pcard-67164
Fix backward program kwargs error when process inplace value


SigureMo (approved):
LGTMeow

PR #64001, opened 2024-04-30
【Hackathon 6th No.10】Add isin API to Paddle -part

NKNaN:
### PR Category

User Experience

### PR Types

New features

### Description

rfcs: https://github.com/PaddlePaddle/community/pull/884


python/paddle/tensor/math.py
```diff
@@ -7871,3 +7871,121 @@ def isreal(x, name=None):
...
+            Tensor(shape=[5], dtype=bool, place=Place(cpu), stop_gradient=True,
+            [False, True, True, False, True])
+
+            >>> elements = paddle.to_tensor([-0., -2.1, 2.5, 1.0, -2.1], dtype='float32')
+            >>> test_elements = paddle.to_tensor([-2.1, 2.5], dtype='float32')
+            >>> res = paddle.isin(elements, test_elements, invert=True)
+            >>> print(res)
+            Tensor(shape=[5], dtype=bool, place=Place(cpu), stop_gradient=True,
+            [True, False, False, True, False])
+    """
```
luotao1, line 8058: > 参考 pytorch 在 _decompose 中的设计：当 test_elements 元素个数较少时直接进行暴力搜索，较多时则采取基于排序的算法：若 assume_unique=True，则用稳定排序的 argsort 进行实现；若 assume_unique=False，则用 searchsorted 进行实现。

示例代码中，能否体现出 assume_unique=True/False 的区别？注释中是否要加入这段说明

NKNaN: 增加了一个 `assume_unique` 设置错误导致结果出错的例子；
在下面代码中加了不同分支对应不同做法的注释

luotao1 (dismissed):

python/paddle/tensor/math.py
```diff
@@ -7871,3 +7871,176 @@ def isreal(x, name=None):
         return paddle.ones_like(x, dtype='bool')
 
     return paddle.equal(paddle.imag(x), 0)
+
+
+def isin(elements, test_elements, assume_unique=False, invert=False, name=None):
```
jeff41404: Accroding to [API design guidelines standard](https://www.paddlepaddle.org.cn/documentation/docs/zh/dev_guides/api_contributing_guides/api_design_guidelines_standard_cn.html), use `x` instead of `elements` and `test_x` instead of `test_elements`

python/paddle/tensor/math.py
```diff
@@ -7871,3 +7871,176 @@ def isreal(x, name=None):
...
+        raise TypeError(f"x must be tensor type, but got {type(test_elements)}")
+
+    check_variable_and_dtype(
+        elements,
+        "elements",
+        [
+            'float32',
+            'float64',
+            'int32',
+            'int64',
```
jeff41404, line 8073: should we support `bfloat16` and `float16`?

python/paddle/tensor/math.py
```diff
@@ -7871,3 +7871,176 @@ def isreal(x, name=None):
...
+    )
+
+    check_variable_and_dtype(
+        test_elements,
+        "test_elements",
+        [
+            'float32',
+            'float64',
+            'int32',
+            'int64',
```
jeff41404, line 8087: same date type issue as above

python/paddle/tensor/math.py
NKNaN: revised.

python/paddle/tensor/math.py
NKNaN: Temporarily not since `paddle.searchsorted` only support 'float32', 'float64', 'int32' and 'int64'.

python/paddle/tensor/math.py
NKNaN: Temporarily not since `paddle.searchsorted` only support 'float32', 'float64', 'int32' and 'int64'.

NKNaN:
@jeff41404 `paddle.searchsorted` currently does not support fp16 and bf16. I think the reason may be that the two input tensors' data type of `paddle.searchsorted` could be different, and since fp16 and bf16 has lower precision than float32 and float64, the comparision between them may be less meaningful.

jeff41404:
> @jeff41404 `paddle.searchsorted` currently does not support fp16 and bf16. I think the reason may be that the two input tensors' data type of `paddle.searchsorted` could be different, and since fp16 and bf16 has lower precision than float32 and float64, the comparision between them may be less meaningful.

[the logic of searchsorted](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/phi/kernels/impl/searchsorted_kernel_impl.h#L105) also supports comparing two inputs with different data types. Registering the fp16 and bf16 data types in the Operator should support these two data types. `searchsorted` operator was developed many years ago, when fp16 and bf16 were not yet widely used, and it was not intentionally unsupported due to accuracy or lower precision.

NKNaN:
> [the logic of searchsorted](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/phi/kernels/impl/searchsorted_kernel_impl.h#L105) also supports comparing two inputs with different data types. Registering the fp16 and bf16 data types in the Operator should support these two data types. `searchsorted` operator was developed many years ago, when fp16 and bf16 were not yet widely used, and it was not intentionally unsupported due to accuracy or lower precision.

Got it. Let me update the `searchsorted` to support fp16 and bf16 first.

luotao1 (approved):

jeff41404 (approved):
LGTM

sunzhongkai588 (approved):
LGTM for docs

PR #64132, opened 2024-05-09
[OneDNN]Fix ocr error since pass avx512 command

zhanglirong1999:
### PR Category
Others

### PR Types
Bug fixes

### Description
Fix ocr error since pass avx512 command

### DIFF
--- a/paddle/fluid/framework/ir/onednn/self_attention_fuse_pass.cc
+++ b/paddle/fluid/framework/ir/onednn/self_attention_fuse_pass.cc
@@ -55,12 +55,9 @@ void SelfAttentionFusePass::ApplyImpl(ir::Graph* graph) const {
   return;
 #endif
 
-#ifdef PADDLE_WITH_DNNL
-  if (!phi::backends::cpu::MayIUse(
-          phi::backends::cpu::cpu_isa_t::avx512_core)) {
+  if (!phi::backends::cpu::MayIUse(phi::backends::cpu::cpu_isa_t::avx512f)) {
     return;
   }
-#endif
 
   // do something;
   GraphPatternDetector gpd;
--- a/paddle/fluid/framework/ir/onednn/self_attention_fuse_pass.cc
+++ b/paddle/fluid/framework/ir/onednn/self_attention_fuse_pass.cc
@@ -17,6 +17,7 @@
 #include <string>
 #include "paddle/fluid/framework/lod_tensor.h"
 #include "paddle/fluid/framework/op_version_registry.h"
+#include "paddle/phi/backends/cpu/cpu_info.h"
 #include "paddle/phi/kernels/funcs/blas/blas.h"
 #include "paddle/utils/string/pretty_log.h"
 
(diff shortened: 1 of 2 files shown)


zhanglirong1999:
related issue: [#11597](https://github.com/PaddlePaddle/PaddleOCR/issues/11597)

paddle/fluid/framework/ir/onednn/self_attention_fuse_pass.cc
```diff
@@ -53,6 +54,14 @@ void SelfAttentionFusePass::ApplyImpl(ir::Graph* graph) const {
   LOG(WARNING) << "No-avx512 or MKL supported!";
   return;
 #endif
+
+#ifdef PADDLE_WITH_DNNL
```
LLee233: This define can be deleted since if the code can go here, meaning these definitions are all met (at least in compilation stage):
```
#if !defined(__AVX512F__) || !defined(PADDLE_WITH_MKLML) || \
    !defined(PADDLE_WITH_DNNL)
```

paddle/fluid/framework/ir/onednn/self_attention_fuse_pass.cc
```diff
@@ -53,6 +54,14 @@ void SelfAttentionFusePass::ApplyImpl(ir::Graph* graph) const {
...
+  if (!phi::backends::cpu::MayIUse(
+          phi::backends::cpu::cpu_isa_t::avx512_core)) {
```
LLee233: For here, the limitation is a bit strict, since in `self_dp_attention_kernel`, the commonly used ISA was `AVX512F`. While this condition will block application of machines with only `AVX512F`.

jzhang533:
thanks for fixing, skip avx512 implemented pass on non-avx512 platforms will solve the illegal instruction issue. 

I am kind of wondering the expected performance on non-avx512 platforms, since we have issues reporting unacceptable low speed on non-avx512 plaftorms, e.g.: https://github.com/PaddlePaddle/PaddleOCR/issues/10346.

LLee233 (approved):
LGTM

yuanlehome (approved):

LLee233:
> thanks for fixing, skip avx512 implemented pass on non-avx512 platforms will solve the illegal instruction issue.
> 
> I am kind of wondering the expected performance on non-avx512 platforms, since we have issues reporting unacceptable low speed on non-avx512 plaftorms, e.g.: [PaddlePaddle/PaddleOCR#10346](https://github.com/PaddlePaddle/PaddleOCR/issues/10346).

Hi @jzhang533, as I mentioned in the [issue](https://github.com/PaddlePaddle/PaddleOCR/issues/10346), this is actually the same reason causing unexpected phenomenon. AMD seems has their mechanism to deal with incompatible higher ISA (and which also makes it hard to find the root cause since it wouldn’t report the error directly). 

And about performance on non-avx512 platform. If we build and install Paddle on non-avx512 machine, the performance will be as expected, which means the performance will be better when enabling oneDNN/mkldnn.

PR #64137, opened 2024-05-09
[AutoParallel-PIR] Remove program clone in Pass

JZ-LIANG:
### PR Category

Auto Parallel

### PR Types

Improvements

### Description

Pcard-76459
Remove program clone to keep operation relationship between forward and backward which is need by de-composition.


zhiqiu (approved):
LGTM

PR #64145, opened 2024-05-09
Fixes about `unique_ptr` in `malloc_test`

lawrence910426:
### PR Category

Performance Optimization

### PR Types

Bug fixes

### Description

1. Put the AllocationPtr (which is a `unique_ptr`) to a bigger scope and therefore the smart pointer would not deallocate the resources too early.
2. Manually handle the deallocation order of memory resources & stream resources.


luotao1:
Please fix the PR-CI-Hygon-DCU CI

zyfncg (approved):

PR #64202, opened 2024-05-10
add spmd rule for amp ops

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Devs

### Description

add spmd rule for amp ops
- check_finite_and_unscale_
- update_loss_scaling_
- 
Fix spmd rule for layer_norm_grad

Pcard-67164


paddle/phi/infermeta/spmd_rules/amp_ops.cc
```diff
@@ -0,0 +1,82 @@
...
+  for (auto& x : xs) {
+    auto dist_attr = x.dist_attr();
+    dist_attr.clean_partial_status();
+    xs_attrs.emplace_back(dist_attr);
+    if (splited) {
+      continue;
+    }
+    auto dims_mapping = dist_attr.dims_mapping();
+    for (auto& m : dims_mapping) {
+      if (m != -1) {
```
JZ-LIANG: would be better to record split on which mesh and which dimension of mesh

paddle/phi/infermeta/spmd_rules/amp_ops.cc
```diff
@@ -0,0 +1,82 @@
...
+        splited = true;
+      }
+    }
+  }
+  TensorDistAttr found_infinite_attr =
+      CopyTensorDistAttrForOutput(scale.dist_attr());
+  if (splited) {
+    found_infinite_attr.set_partial(true);
```
JZ-LIANG: would be according to the mesh info recorded above, specific the partial dimension.
in which case， the synchronization would be limited in the specific group where the tensor are splited.
BUT here is also OK, since the found_infinite is a BOOL variable, synchronize it among the global world size.

paddle/phi/infermeta/spmd_rules/amp_ops.cc
```diff
@@ -0,0 +1,82 @@
...
+
+namespace phi {
+namespace distributed {
+SpmdInfo CheckFiniteAndUnscaleSpmd(const std::vector<DistMetaTensor>& xs,
+                                   const DistMetaTensor& scale) {
+  std::vector<TensorDistAttr> xs_attrs;
+  bool splited = false;
+  for (auto& x : xs) {
+    auto dist_attr = x.dist_attr();
+    dist_attr.clean_partial_status();
```
JZ-LIANG, line 31: could allow the partial status pass through to have better performance:
if not allow partial to get through, allreduce will be conducted before this operation, no matter checkfinite return False or True.
BUT if we allow the partial to get through, allreduce will be triggered by the later operation(e.g. adam) and if the  checkfinite return True, not adam would be call and therefore not allreduce is conduct

paddle/phi/infermeta/spmd_rules/amp_ops.cc
```diff
@@ -0,0 +1,82 @@
...
+                               const DistMetaTensor& in_bad_steps,
+                               int incr_every_n_steps,
+                               int decr_every_n_nan_or_inf,
+                               float incr_ratio,
+                               float decr_ratio,
+                               Scalar stop_update = false) {
+  std::vector<TensorDistAttr> xs_attrs;
+  for (auto& x : xs) {
+    auto dist_attr = x.dist_attr();
+    dist_attr.clean_partial_status();
```
JZ-LIANG, line 60: the same as above

paddle/phi/infermeta/spmd_rules/amp_ops.cc
```diff
@@ -0,0 +1,82 @@
...
+    xs_attrs.emplace_back(dist_attr);
+  }
+  TensorDistAttr found_infinite_attr =
+      CopyTensorDistAttrForOutput(found_infinite.dist_attr());
+  return {
+    {xs_attrs,
+     found_infinite_attr,
+     prev_loss_scaling.dist_attr(),
```
JZ-LIANG: all these scalar tensor should be placemented on a merged mesh  of all gradient tensors' mesh.

heavyrain-lzy (approved):
LGTM for yaml

JZ-LIANG (approved):
LGTM

LiYuRio (approved):
lgtm

PR #64220, opened 2024-05-11
[Prim] add flag test cases and block prim vjp matmul by default

cyber-pioneer:
### PR Category

Others

### PR Types

Others

### Description

Pcard-66975
1. Since some decomposition of special ops like **matmul_grad** will reduce performance and is difficult to optimize currently by CINN.
2. Those ops are set in SKIPPED_PRIM_VJP_DEFAULT_OPS.
add Flags **FLAGS_prim_vjp_skip_default_ops** to control, default value is true.


Aurelius84 (approved):
LGTM

PR #64279, opened 2024-05-14
[XPU] set the shape of tile's output when the input is 0D

runzhech:
### PR Category
Custom Device

### PR Types
Improvements

### Description
When use a 0D-tensor as the input to Tile OP, the output on GPU has the new shape while the XPU not.
This PR makes the implementation of Tile on XPU to identical to that on GPU.


zhangyk0314 (approved):
LGTM

cqulilujia (approved):
LGTM

QingshuChen (approved):

PR #64283, opened 2024-05-14
[Test] test quantize op test , do not mention

enkilee:
### PR Category

Others

### PR Types

Others

### Description

Just test old op test.


LLee233:
Hi @enkilee, this issue is caused by wrong condition in `prepare_output()`, since quantize only outputs **uint8 with shift**. You may fix it by changing code as below, thanks:
image

enkilee:
> Hi @enkilee, this issue is caused by wrong condition in `prepare_output()`, since dequantize only outputs **uint8 with shift**. You may fix it by changing code as below, thanks: image

Thanks a lot.

PR #64299, opened 2024-05-14
move mix2dist pass to python end.

winter-wang:
### PR Category

Auto Parallel 
### PR Types

Others
### Description

- move mix2dist pass from c++ end to python end.
- support shard other replicated value in mix2dist pass. 
- support default InferSpmd function.

### Others
Pcard-67164


zhiqiu (approved):
LGTM

PR #64412, opened 2024-05-17
[PIR][oneDNN] Fix conv_bias_fuse_pass

LLee233:
### PR Category

Others

### PR Types

Bug fixes

### Description

Current constraints of pass are not enough, and it will fuse some ops inappropriately (like fusing multi-dimension bias into conv, and the fused op will fail at dimension check). Hence we add new constraint to avoid such situations.

Also refined `FusedConvTransposeAddFusePattern`, since when `ConvTransposeBiasFusePattern` exists, the former one will never be executed.

### DIFF
--- a/paddle/fluid/pir/transforms/onednn/conv_bias_fuse_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/conv_bias_fuse_pass.cc
@@ -216,30 +216,34 @@ class FusedConvTransposeAddFusePattern : public paddle::drr::DrrPatternBase {
   void operator()(paddle::drr::DrrPatternContext *ctx) const override {
     paddle::drr::SourcePattern pat = ctx->SourcePattern();
     const auto &conv =
-        pat.Op(paddle::dialect::Conv2dTransposeOp::name(),
+        pat.Op(paddle::onednn::dialect::Conv2dTransposeBiasOp::name(),
                {{"strides", pat.Attr("strides")},
                 {"paddings", pat.Attr("paddings")},
                 {"output_padding", pat.Attr("output_padding")},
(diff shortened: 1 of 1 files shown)


zhanglirong1999:
Hi, about ConvBiasFusePass, I have some questions. In previous pass( non-pir implementation), I run ResNet50, and found it not fused in conv_bias pass, but fused in conv_elementwise_add pass. But if I run ResNet50 with PIR pass, it will fused in conv_bias pass. Your fix will constraint condition and make it not fused in conv_bias pass, but I notice after you fix, it still not be fused in conv_elementwise_add pass. So can you check it? Why conv bias/ conv elementwise is it different from the previous non-pir strategy? I'm not sure if our implementation is actually different or wrong under PIR than before.

LLee233:
> Hi, about ConvBiasFusePass, I have some questions. In previous pass( non-pir implementation), I run ResNet50, and found it not fused in conv_bias pass, but fused in conv_elementwise_add pass. But if I run ResNet50 with PIR pass, it will fused in conv_bias pass. Your fix will constraint condition and make it not fused in conv_bias pass, but I notice after you fix, it still not be fused in conv_elementwise_add pass. So can you check it? Why conv bias/ conv elementwise is it different from the previous non-pir strategy?

Since I only added constraints for `conv_bias_fuse_pass`, it should not block fusion of `conv_elementwise_add_pass`. Did you re-run it when disabling bias pass? I guess the result will still be the same (add pass is not fused). But yes, I will check if there is any difference between pir pass and fluid pass of the `conv_elementwise_add_pass` as the following issue, thx~

yuanlehome (approved):

xinyu-intel (approved):

PR #64491, opened 2024-05-21
support LRSchedule for pir.

winter-wang:
### PR Category

Auto Parallel

### PR Types

 Bug fixes

### Description

- support LRSchedule for pir program.

### Other
Pcard-67164

### DIFF
--- a/paddle/fluid/pybind/pir.cc
+++ b/paddle/fluid/pybind/pir.cc
@@ -111,6 +111,7 @@ using pir::OpResult;
 using pir::Pass;
 using pir::PassManager;
 using pir::Program;
+using pir::StrAttribute;
 using pir::Type;
 using pir::Value;
 using pybind11::return_value_policy;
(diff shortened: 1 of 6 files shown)


JZ-LIANG (dismissed):
LGTM

paddle/fluid/pybind/pir.cc
```diff
@@ -281,6 +297,45 @@ std::string GetValueName(Value value) {
...
+  if (p_mapper == nullptr) {
+    p_mapper = &mapper;
+  }
+  auto src_obj = py::cast(self);
+  auto new_obj = py::cast(self.Clone(*p_mapper));
+  for (auto item : src_obj.attr("__dict__").cast<py::dict>()) {
+    new_obj.attr(item.first.cast<std::string>().c_str()) = item.second;
+  }
+  return new_obj;
+}
```
zhiqiu, line 311: what about the calling original clone in c++ side?

zhiqiu (approved):
LGTM

paddle/fluid/pybind/pir.cc
winter-wang: It will only clone the computation graph, some python properties such as "lr_rate" may be lost in the clone.

PR #64558, opened 2024-05-23
[CINN] Fix accuracy_check_pass bug

chen2016013:
### PR Category

CINN

### PR Types

Bug fixes

### Description

Pcard-67164
the order of ops is not correct after accuracy_check_pass because insert position is not update
this PR fix it


Aurelius84 (approved):
LGTM

PR #64579, opened 2024-05-24
fix pir-auto-parallel bug in llama pp demo.

winter-wang:
### PR Category

 Auto Parallel

### PR Types

Bug fixes
### Description

fix pir-auto-parallel bug in llama pp demo.
- fix partition pass bug,

### Other 
Pcard-67164


zhiqiu (approved):
LGTM

PR #64620, opened 2024-05-27
[CINN] fix sampling in Schedule Config Searcher

CJ77Qi:
### PR Category

CINN

### PR Types

 Bug fixes
### Description

Measure with the same shape instead of resampling each time


BiynXu (approved):
LGTM

PR #64676, opened 2024-05-29
[OneDNN][PIR] Fix OneDNN mean kernel fallback fault

zhanglirong1999:
### PR Category
Others

### PR Types
Bug fixes

### Description
Fix OneDNN mean kernel fallback fault in Mixed OneDNN phi kernel. Mean op is different from 'Max', 'Min', the previous method of obtaining attributes will cause errors


wanghuancoder (approved):
LGTM

yuanlehome (approved):

PR #64792, opened 2024-05-30
【AutoParallel】Add sharding overlap

heavyrain-lzy:
### PR Category

Auto Parallel

### PR Types

Performance

### Description

Pcard-76459
Enable reduce overlap in sharding-strategy.


zhiqiu (approved):
LGTM

PR #64813, opened 2024-05-31
fix dygraph auto-gen auto_parallel code when some output is null

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

The auto-generated dygraph code for auto_parallel does not handle nullptr output, so it may result in `core dump` when some grad_output  is empty because of stop_gradient or detach.
For example, the tensor `out.detach()` in concat needs no grad in backward.
```
    a = paddle.rand([4, 16, 16, 4])
    a.stop_gradient = False
    a = dist.shard_tensor(a, mesh, [dist.Replicate()])
    out, var = paddle.split(a, 2, axis = 1)
    out = paddle.concat([out.detach(), var])
    loss = paddle.mean(out)
    loss.backward()
```
Pcard-76459


heavyrain-lzy (approved):
LGTM

luotao1 (approved):

Aurelius84 (approved):
LGTM For const_cast

jeff41404 (approved):

PR #64889, opened 2024-06-04
fix replicated dist_out_attr for view op

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

ATT
The original implementation may generate the follow code for view op,
```
std::tuple<Tensor, Tensor> api_output;
auto dist_out_attr_0 = static_cast<phi::distributed::DistTensor*>((std::get<1>(api_output)).impl().get())->dist_attr();
```
However, api_output is uninitialized and its impl is NULL, SO it may result in segment fault.

Pcard-76459


heavyrain-lzy (approved):
LGTM

PR #64892, opened 2024-06-04
[CINN] fix buffer resize in load node

BiynXu:
### PR Category
CINN

### PR Types
Bug fixes

### Description
Pcard-74042
Fix buffer resize in load node.


2742195759 (approved):
LGTM

PR #64925, opened 2024-06-05
[PIR][oneDNN] set is_test attribute for onednn ops

LLee233:
### PR Category

Others

### PR Types

Others

### Description

In former fluid mechanism, there was a pass called "is_test_pass" to set "is_test" as true for all ops. While in PIR, there is no such pass. To avoid failure on some ops, we revise the pass to support setting true for this attribute.

### DIFF
--- a/paddle/fluid/pir/transforms/onednn/onednn_placement_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/onednn_placement_pass.cc
@@ -51,7 +51,12 @@ class OneDNNPlacementPattern : public pir::OpRewritePattern<OpType> {
       paddle::dialect::OpRunTimeInfo runtime_info =
           std::get<3>(yaml_interface->get_op_info_(target_op_name));
       for (auto &attr : runtime_info.extra_args_default_value) {
-        attributes[attr.first] = attr.second;
+        if (attributes.find(attr.first) == attributes.end()) {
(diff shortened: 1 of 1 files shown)


paddle/fluid/pir/transforms/onednn/onednn_placement_pass.cc
```diff
@@ -51,7 +51,12 @@ class OneDNNPlacementPattern : public pir::OpRewritePattern<OpType> {
       paddle::dialect::OpRunTimeInfo runtime_info =
           std::get<3>(yaml_interface->get_op_info_(target_op_name));
       for (auto &attr : runtime_info.extra_args_default_value) {
-        attributes[attr.first] = attr.second;
+        if (attributes.find(attr.first) == attributes.end()) {
```
zhanglirong1999: I notice that 'onednn ops extra args' is extra attribute for onednn op and seems not exist in original op. Do we still need such a condition? Loop querying the map also requires overhead.

LLee233: Thanks for reminding! Just check the yamls and generation rules, the extra attrs of onednn op will be directly appended to the tail of original attrs, so it's truly no need to add `find` condition. Will change it back later~

zhanglirong1999 (approved):

PR #64959, opened 2024-06-06
alignment data parallel precision between dynamic graph and dy2st.

winter-wang:
### PR Category

Auto Parallel

### PR Types

Bug fixes 

### Description

- remove unuseful communication op in reshard pass. (slice/all_gather op with rank is 1).
### Other 
Pcard-67164


zhiqiu (approved):
LGTM

PR #64974, opened 2024-06-06
[CINN]add limit for broadcast tree

Hongqing-work:
### PR Category

CINN

### PR Types

Others

### Description

Pcard-67164
This PR adds limit for broadcast tree. When the number of single Broadcast Expr items is larger than 3 or the number of broadcast leaves is greater than FLAGS_pir_broadcast_tree_limit(default is 32), CINN will throw phi::errors::Fatal("Too many broadcast leaves to compile!".


zyfncg (approved):

luotao1 (approved):
LGTM for flags

PR #64723, opened 2024-05-29
add flatten_grad spmd rule and modify flatten spmd rule to call spmd rule of reshape

jeff41404:
### PR Category
Auto Parallel

### PR Types
New features

### Description
pcard-84225
- add flatten_grad spmd rule
- modify flatten spmd rule to call spmd rule of reshape, so can support dygraph

PR #64985, opened 2024-06-07
fix spmd rule of flatten_grad, add xshape dist_attr in result

jeff41404:
### PR Category
Auto Parallel

### PR Types
Bug fixes

### Description
pcard-84490
- when #64723 add flatten_grad spmd rule, which directly use ReshapeGradInferSpmd, but ReshapeGradInferSpmd do not return distributed attribute of xshape in result which will cause error in running flatten_grad.
- when ReshapeInferSpmd and ReshapeGradInferSpmd can deliver distributed attribute of xshape, we will use ReshapeGradInferSpmd directly in future.


zhiqiu (approved):
LGTM

PR #65023, opened 2024-06-11
Prevent Early Release of AllocationPtr in `test_from_blob`

eee4017:
### PR Category
Others

### PR Types
Bug fixes

### Description

The `Allocate()` function returns a `unique_ptr`. When we chain the functions and call `ptr()`, the original `unique_ptr` is released, making the accessed memory invalid.

To address this issue, we must retain the `unique_ptr` until the memory is no longer in use.

With a stream-safe allocator, this bug is not immediately apparent because the allocator does not trigger cudaFree immediately. However, with an async allocator, the bug is detected due to stricter memory management.


jeng1220:
CI failed but it was **NOT** related to this PR
```cpp
/paddle/paddle/phi/core/storage_properties.h:20:10: fatal error: 'paddle/phi/core/utils/type_registry.h' file not found
#include "paddle/phi/core/utils/type_registry.h"
```

zyfncg (approved):

PR #65056, opened 2024-06-12
[CINN]Call GetKernelInfo in thread to speed up compile (#64893)

0x45f:
### PR Category

Performance Optimization

### PR Types

Performance

### Description

Call GetKernelInfo in thread to speed up compile


Aurelius84 (approved):
LGTM

PR #60516, opened 2024-01-02
Pipeline and Virtual-Pipeline Parallelism Using CUDA Graph and Integrate CUDAMallocAsyncAllocator

eee4017:
### PR types
New features

### PR changes
Others

### Description

This PR introduces Pipeline Parallelism (PP) and Virtual-Pipeline Parallelism (VP) training through the integration of CUDA Graph. The following is a detailed breakdown of the challenges encountered and the innovative solutions we have implemented:

### PP/VP + CUDA Graph
Usage: Enable CUDA Graph in PipelineLayer using the `use_cudagraph=true` flag.

- PP and CUDA Graph:
  - **Challenge:** PP layers differ from standard nn.Layer as they are distributed across multiple GPUs. They are of two types: Layer and SharedLayer. SharedLayer's need for synchronous communication conflicts with CUDA Graph's requirements.
  - **Solution:** We have restructured PP layers into multiple capturable sublayers (contiguous Layers), enabling each to be captured into a separate graph. This facilitates efficient multi-GPU execution and mitigates the communication conflicts inherent in SharedLayers.
- VP and CUDA Graph:
  - **Challenge:** VP employs a non-traditional training pattern, executing multiple forward (FW) and backward (BW) stages concurrently. This necessitates continuous access to previous input data for each layer throughout the training cycle.
  - **Solution:** To accommodate this, we've integrated a queue into the CUDAGraphedLayer to store previous inputs. Each virtual pipeline stage is then captured as a separate graph, ensuring efficient and accurate training.

### CUDAMallocAsyncAllocator
Usage: Activate the `CUDAMallocAsyncAllocator` by setting `FLAG_use_cuda_malloc_async_allocator=1`.
- CUDAMallocAsyncAllocator 
  The `CUDAMallocAsyncAllocator` replaced the `StreamSafeCUDAAllocator`. By leveraging the advanced capabilities of `cudaMallocAsync` and `cudaFreeAsync`, the responsibility for stream-ordered memory management is transferred from the framework to CUDA. This transition may lead to better memory utilization and potentially improved application performance by optimizing the way memory is allocated and deallocated within the CUDA. 
  
- CUDAMallocAsyncAllocator + CUDAGraph
  - **Challenge:** The integration of PP/VP with CUDA Graph led to an excessive creation of graphs, and each graph has its own memory pool in Paddle's existing implementation. This resulted in low memory reuse and high memory footprint as allocations were not released until the deletion of the graph.
[...]

PR #65092, opened 2024-06-12
The CUDA Async Allocator

eee4017:
### PR Category
Others

### PR Types
New features

### Description

This pull request (PR) refactors the CUDA Asynchronous Allocator, introducing a new design for handling stream semantics within the allocator. The CUDA Asynchronous Allocator serves as an alternative to the streamsafe allocator, offloading all stream-ordered memory management tasks to CUDA. The CUDAMallocAsyncAllocator can be activated by setting the flag `FLAG_use_cuda_malloc_async_allocator=1`.

#### Why We Need the CUDA Async Allocator

**(1) Reducing Memory Footprint with Other Python Libraries**

In machine learning, we often use various Python libraries to process data. Many of these libraries, such as CuPy and OpenCV, have their own memory pools. When multiple programs or libraries use a GPU, they compete for memory. Memory allocated by one library cannot be easily deallocated by another. Therefore, using the CUDA Async Allocator allows a unified pool for these libraries and thus it is crucial for smoother integration with the Python GPU-computing ecosystem.

**(2) Reducing Memory Footprint with CUDA Graphs**

In PaddlePaddle, a memory pool is opened for each CUDA Graph, potentially leading to memory waste. When there are many graphs or frequent memory reuse between graphs, out-of-memory (OOM) errors can occur. This issue is demonstrated in PR #60516, where inefficient memory usage in the PP or VP training of GPT-3 with CUDA Graph enabled caused OOM errors using four H100-80GB GPUs with the default memory pool. The implementation of the CUDA Async Allocator has significantly reduced memory usage from 95% to 25%.

**(3) Decreasing the Memory Management Burden on the Framework**

The CUDA stream-ordered memory management API is more compatible with stream semantics, allowing for allocation and freeing on a specific stream. This reduces the workload on the framework. In the streamsafe allocator, `cudaEventQuery` is called a lot to determine if it can be released. With the Async Allocator, this process is simplified and offloaded to CUDA. Additionally, `cudaMallocAsync` and `cudaFree` are much faster than `cudaMalloc` and `cudaFree`, leading to performance improvements when `cudaMalloc` is frequently called by the streamsafe allocator.

**(4) Stricter and Safer Memory Management**

By fully offloading memory management to CUDA, the CUDA Async Allocator provides stricter memory management. In the streamsafe allocator, blocks are cached for reuse, meaning memory may not be freed immediately after a block is released. This can prevent CUDA errors from use-after-free scenarios. Additionally, the streamsafe allocator may allocate more memory to align with specific sizes, masking out-of-bound errors. The CUDA Async Allocator helps detect memory leaks and ensures safer GPU memory usage. Several bugs were detected with the CUDA Async Allocator, resulting in the creation of PRs for these bugs:

- #65019 
- #65023 
- #65024 
- #65093 

#### The Design of the CUDA Async Allocator

This document outlines the design and implementation of the CUDA Asynchronous Allocator. The design addresses how to handle stream semantics in memory management, improving efficiency and reducing memory overhead compared to the traditional streamsafe allocator.

**(1) Stream Semantic of the CUDA Memory Management**

Paddle drawio (1)

Figure 1. Stream Safe Allocator

Paddle drawio

Figure 2. CUDA Async Allocator

When a block is allocated on a stream, it might be used by other streams. Therefore, a mechanism is required to ensure that the block is freed after it is used on the specific stream.

In the streamsafe allocator (Figure 1):
- A `cudaEvent` is created when the block is used on a stream.
- When a block is released, it is pushed into a queue.
- `cudaEventQuery` is used to determine if the block can be released.
- The check `ProcessUnfreedBlock` is heuristically triggered when `malloc` is called, requiring continuous tracking to see if each block can be freed.

In the Async Allocator (Figure 2):
- A free stream is used to free the block.
- Inter-stream synchronization (`cudaEventRecord`/`cudaStreamWaitEvent`) ensures the block is freed after it is used.
- Continuous tracking of blocks is not needed, offloading the burden to CUDA.

**(2) Throttling Mechanism**

When memory is under pressure (nearing OOM), the free operation may not be fast enough. Therefore, the allocation stream needs to be throttled (as indicated by the red arrow in Figure 2).  When the memory utilization exceeds the `memory_throttle_ratio`, a stream synchronization operation is initiated before `malloc`.

```
utilization = (allocated_size + pending_release_size) / total_memory_size
if (utilization > memory_throttle_ratio) {
    sync(free_stream, malloc_stream);
}
```

During synchronization, all memory deallocation requests in the free queue are processed, reducing memory utilization before any new allocation operations proceed. Currently, the ratio is heuristically set to 80%, but it can be adjusted using `FLAGS_cuda_malloc_async_pool_memory_throttle_ratio`. 

- **Lower `memory_throttle_ratio` Values**: Trigger synchronization more frequently, improving memory utilization but possibly decreasing performance due to increased synchronization operations.
- **Higher `memory_throttle_ratio` Values**: Allow more memory allocation before triggering synchronization, enhancing performance by reducing sync operations but increasing the risk of OOM conditions.

**(3) Cooperate with CUDA Graph**

A map called `graph_owned_allocations_` is created to track blocks used in the graph. There are four distinct scenarios involving `cudaMallocAsync`, `cudaFreeAsync` in CUDA Graph

1. Both `Malloc` and `Free` take place within a graph.
2. `Malloc` takes place within a graph, but `Free` takes place outside the graph.
3. `Malloc` takes place outside a graph, but `Free` takes place within a graph.
4. Both `Malloc` and `Free` take place outside any graph.

 We handle the release of the graph owned allocation in the above four cases:

- **Scenario 1**: `FreeImpl` removes the allocation from `graph_owned_allocations_`, followed by `FreeAllocation`.
- **Scenario 2**: We transfer the ownership of the block to the graph. We use a callback to free the allocation after the graph is destroyed.
- **Scenario 3**: `FreeImpl` releases the allocation after the CUDA graph has completed its capture.
- **Scenario 4**: `FreeImpl` calls `FreeAllocation`, and the allocation is freed.

Each element within `graph_owned_allocations_` is malloc at `AllocateImpl`, but deallocation can occur in two ways:

- **Deallocation within `FreeImpl`**: This implies the allocation is initialized and disposed of during a graph capture (Scenario 1).
- **Deallocation in the callback after the graph is destructed**: The allocation is initialized during a graph capture but disposed of outside that context (Scenario 2).

#### Testing the CUDA Async Allocator

Set `FLAG_use_cuda_malloc_async_allocator=1` and run all tests to validate that the CUDA Asynchronous Allocator can serve as an alternative to the stream-safe allocator. Note that some unrelated tests should be disabled 
-  #65094


jeng1220:
@eee4017 ,
Windows Build failed
```cpp
liballocator.lib(cuda_malloc_async_allocator.cc.obj) : error LNK2019: unresolved external symbol "double paddle_flags::FLAGS_cuda_malloc_async_pool_memory_throttle_ratio" (?FLAGS_cuda_malloc_async_pool_memory_throttle_ratio@paddle_flags@@3NA) referenced in function "public: __cdecl paddle::memory::allocation::CUDAMallocAsyncAllocator::CUDAMallocAsyncAllocator(class std::shared_ptr,class phi::GPUPlace const &,struct CUstream_st *)"
```

jeng1220:
> @eee4017 , Windows Build failed
> 
> ```c++
> liballocator.lib(cuda_malloc_async_allocator.cc.obj) : error LNK2019: unresolved external symbol "double paddle_flags::FLAGS_cuda_malloc_async_pool_memory_throttle_ratio" (?FLAGS_cuda_malloc_async_pool_memory_throttle_ratio@paddle_flags@@3NA) referenced in function "public: __cdecl paddle::memory::allocation::CUDAMallocAsyncAllocator::CUDAMallocAsyncAllocator(class std::shared_ptr,class phi::GPUPlace const &,struct CUstream_st *)"
> ```

It was resolved.

tianshuo78520a:
PR-CI-Hygon-DCU 编译有些问题，需要修复
f66845748355465c369595cdd46e375d

eee4017:
You must have one RD (phlrain or luotao1 or Aurelius84) approval for changing the FLAGS, which manages the environment variables.

paddle/common/flags.cc
```diff
@@ -1130,6 +1130,28 @@ PHI_DEFINE_EXPORTED_bool(use_cuda_malloc_async_allocator,
                          false,
                          "Enable CUDAMallocAsyncAllocator");
 
+/*
+ * CUDAMallocAsyncAllocator related FLAG
+ * Name: FLAGS_cuda_malloc_async_pool_memory_throttle_ratio
+ * Since Version: 2.7
```
zyfncg: 2.7 -> 3.0

eee4017: fixed

zyfncg (approved):

phlrain (approved):

PR #65102, opened 2024-06-13
[XPU] force use calc stream for xpu

lj970926:
### PR Category

Custom Device

### PR Types

Performance

### Description

1. Force `use_calc_stream = True`. 
The python code in Paddle set `use_calc_stream = False` and `sync_op=True` by default. However, the performance improvement of dispatching communication kernels on another stream is minor for XPU and would cause additional `xpu_wait` that hinder the end-to-end performance of models.
2. Do not create stream for comm `XPUContext` to reduce stream creation in `ProcessGroupBKCL`.
The hardware channels for XPU are limited. If the number of streams created exceeds the max channels available, there might be calc streams that are projected onto the same channel which disables the CDNN_CLUSTER parallel essentially. Because we force `use_calc_stream=True` for communication, We can avoid the stream creation in `XPUContext` when `is_comm_context=True`.


XiaociZhang (commented):
LGTM

ZibinGuo (approved):
LGTM

QingshuChen (approved):

QingshuChen (approved):

sneaxiy (approved):

PR #65104, opened 2024-06-13
[AutoParallel] Refine some pass and spmd rules

zhangbo9674:
### PR Category

Others

### PR Types

Others

### Description

This PR includes the following work：
- register strided_slice inferspmd rule
- refine fuse_allreduce_split_to_reducescatter_pass, add mm+allreduce+add pattern
- fix the performance degradation problem caused by type promotion logic
- fix the problem that in the program translator, there is a fake false block in the if op, resulting in reduced execution performance.

Pcard-73145


zhiqiu (approved):
LGTM

PR #65259, opened 2024-06-18
[AutoParallel] Handle GradClipByGlobalNorm under AutoParallel Semantic 

JZ-LIANG:
### PR Category

Auto Parallel

### PR Types

Improvements

### Description

Pcard-67164

**Backgroud:**
In hybrid parallelism, the GradClipByGlobalNorm should be adapted.
Since the gradients are sharded on among ranks, synchronization between different parallel groups is need to compute the exact denominator global_norm. 

image

We plan to support GradClipByGlobalNorm in AutoParallel in two approaches:

The first approach (In this PR):

- has more communication overhead but could compute exact the same global_norm as if the model is run on single device, since it move all gradients tensor to the first pp stage.
- which is debug and user friendly when migrating a model from single device version to Autoparallel.

image

The second approach (In next PR):

- Has less communication overhead and therefore higher performance.
- It might results a global_norm slightly different from the one is expected in single device version, **since the accumulation order of float points might be different from the single device version.**

image


winter-wang (approved):

zhiqiu (approved):
LGTM

PR #65262, opened 2024-06-18
fix shard_dataloader

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

Fix random inconsistent of batch data generated from shard_dataloader in dynamic and static mode

Pcard-76459


winter-wang (approved):
LGTM

PR #65435, opened 2024-06-25
add appoval check for prim-related diff in backward

cyber-pioneer:
### PR Category

Others

### PR Types

Others

### Description

Pcard-66975
add appoval check for only prim-related diff in backward_utils.py.


Aurelius84 (approved):
LGTM

PR #65494, opened 2024-06-26
[XPU] refine xpu.cmake for xpti

dynamicheart:
### PR Category

Custom Device

### PR Types

Not User Facing

### Description

- There are many string literals representing `false` in CMake Option, including `false`, `OFF`, `""`, and `0` and they are case-insensitive.
- Using an empty string to determine whether to include the XPTI is insufficient.
https://github.com/PaddlePaddle/Paddle/blob/11f07ea22168452ecb329cb523d3d1795f6dc85c/tools/xpu/get_xpti_dependence.sh#L22
- So we use `XPU_XPTI_URL` to determine whether to include XPTI. It's value can only be either real_url or `""` now that `cmake -DWITH_XPTI=OFF` won't include the XPTI dependency.
https://github.com/PaddlePaddle/Paddle/blob/c33bfddd2f21f457608edba1ecfc11856f2dcc7f/tools/xpu/get_xpti_dependence.sh#L22

### DIFF
--- a/cmake/external/xpu.cmake
+++ b/cmake/external/xpu.cmake
@@ -158,6 +158,10 @@ if(WITH_XPU_BKCL)
   include_directories(${XPU_BKCL_INC_DIR})
 endif()
 
+if(WITH_XPTI)
+  set(XPU_XPTI_OPTION "1")
+endif()
+
 ExternalProject_Add(
   ${XPU_PROJECT}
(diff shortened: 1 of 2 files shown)


XiaociZhang (commented):
LGTM

ZibinGuo (approved):
LGTM

QingshuChen (approved):

PR #65581, opened 2024-06-30
[CINN]Fix loop fuse on discrete spatial axis

BiynXu:
### PR Category
CINN

### PR Types
Bug fixes

### Description
Pcard-74042
Fix loop fuse on discrete spatial axis.
 We need to fuse loops from bottom to top, because the loop index will be changed when the upper loops fused.


tc20042008 (approved):

jiahy0825 (approved):
LGTM

PR #65654, opened 2024-07-03
[Dockerfile][DCU][XPU] add develop dockerfile for dcu and xpu

qili93:
### PR Category

Custom Device

### PR Types

Others

### Description

Add develop dockerfile for dcu and xpu.
card-77889


ronny1996 (approved):
LGTM

PR #65716, opened 2024-07-04
[DCU] Fix some faild dcu ut

Deleter-D:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

card-85146

1. `paddle/phi/kernels/gpu/check_numerics_kernel.cu`: HIP can now use the original CUDA Kernel, and there is no need to write a separate Kernel for HIP.
2. `paddle/phi/kernels/strings/gpu/strings_lower_upper_kernel.cu`: Need to reduce the number of threads to 256 on ROCm platform.
3. `test/cpp/fluid/elementwise/test_elementwise_op_grad_grad.h`: There are different standards for CUDA and HIP platforms, so the standards are unified.
4. `paddle/phi/kernels/gpu/group_norm_kernel.cu`: When the block size is less than the warp size, `WarpReduce` will result in all zeros. It seems to be an internal problem of hipcub on DCU.


qili93 (approved):
LGTM

PR #65750, opened 2024-07-05
[PIR][oneDNN] Add conv2d_bias_bn_onednn_fuse_pass

LLee233:
### PR Category

Others

### PR Types

New features

### Description

Due to the changes of new IR scheme and per the request from Paddle, we hereby add `conv2d_bias_bn_onednn_fuse_pass` from oneDNN side. To fulfill more fusion chances.


zhanglirong1999:
hi, the used bias implement is https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/framework/ir/conv_bn_fuse_pass.cc#L138
, since eltwise_y is 0, so can used the previous formula for conv+bn. Now I think for bias, eltwise_y is not zero, maybe it need to multiply some scale to match this formula?

zhanglirong1999 (approved):

zhanglirong1999 (approved):

PR #65953, opened 2024-07-11
fix compare legacy with pt

risemeup1:
### PR Category

Execute Infrastructure

### PR Types

Bug fixes

### Description

ATT
pcard-67164


SigureMo (approved):
LGTMeow

PR #65956, opened 2024-07-11
[CINN]add ShapeOrDataDimExpr represent for TensorArray type

Hongqing-work:
### PR Category

CINN

### PR Types

Improvements

### Description

Pcard-67164
This PR add ShapeOrDataDimExpr represent for TensorArray type.
#### Design note
TensorArray can append tensors dynamically during the execution of a computation graph, so we can't get a static Shape represenation of TensorArray. However, we can still record the shape of one element tensor as a hint, which can already provide some useful information for subsequent processing, since the elements in TensorArray most likely have the same rank and certain dimensions may have equal constraints.

### DIFF
--- a/paddle/pir/src/dialect/shape/utils/shape_or_data_expr.cc
+++ b/paddle/pir/src/dialect/shape/utils/shape_or_data_expr.cc
@@ -15,28 +15,27 @@
 #include "paddle/pir/include/dialect/shape/utils/shape_or_data_expr.h"
 
 namespace symbol {
+std::vector<DimExpr> SubstituteDimExprVector(
+    const std::vector<DimExpr>& original_dim_expr,
+    const std::unordered_map<DimExpr, DimExpr>& substitution_pattern) {
+  std::vector<DimExpr> substituted_dim_expr{};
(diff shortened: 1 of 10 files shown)


paddle/pir/src/dialect/shape/utils/shape_or_data_expr.cc
```diff
@@ -89,6 +94,10 @@ std::ostream& operator<<(std::ostream& stream,
           }
         }
       },
+      [&](const RankedTensorArrayShapeOrDataDimExprs& tensor_array_shape_data) {
+        stream << "TensorArray with first item shape"
```
zyfncg: first item 的概念应该已经没了

zyfncg (dismissed):

paddle/pir/src/dialect/shape/utils/shape_or_data_expr.cc
Hongqing-work: 收到，下一个PR把这里的打印规范一下

zyfncg (approved):

yuanlehome (approved):
LGTM for pir/include/*

PR #65971, opened 2024-07-12
[OneDNN][PIR] Add matmul_reshape_add pass for mul op

zhanglirong1999:
### PR Category
Performance Optimization

### PR Types
Improvements

### Description
In previous, there is fc_fuse_pass for mul fusion, but now in pir translate mul to reshape(2) + matmul + reshape, making this pass invalid. We write a new pass matmul_reshape_add for reshape(2) + matmul + reshape + add fusion to fc


paddle/fluid/inference/api/paddle_pass_builder.cc
```diff
@@ -648,6 +648,7 @@ const std::vector<std::string> kPirMkldnnPasses {
       "matmul_scale_fuse_pass",                   //
       "scale_matmul_fuse_pass",                   //
       "reshape_transpose_matmul_fuse_pass",       //
+      "matmul_reshape_add_pass",                  //
```
LLee233: Suggest move it after "matmul_add_act_fuse_pass" since "reshape_transpose_matmul_fuse_pass" and "matmul_transpose_reshape_fuse_pass" are serial.

paddle/fluid/pir/transforms/onednn/matmul_reshape_add_pass.cc
```diff
@@ -0,0 +1,155 @@
...
+    const auto &fc_op = res.Op(fused_matmul_name_,
+                               {{
+                                   {"in_num_col_dims", in_num_col_dims_attr},
+                                   {"activation_type", res.StrAttr("")},
+                                   {"padding_weights", res.BoolAttr(false)},
+                               }});
+    fc_op({&res.Tensor("x"), &res.Tensor("w"), &res.Tensor("y")},
+          {&res.Tensor("matmul_out")});
+    const auto &reshape = res.Op(paddle::dialect::ReshapeOp::name());
+    reshape({&res.Tensor("matmul_out"), &res.Tensor("shape")},
```
LLee233: So this pattern will eventually fuse `matmul+reshpe+add` to `fc+reshape`? Is it as expected? Cause this fusion might affect the later pass `fc_activation_fuse_pass`.

paddle/fluid/pir/transforms/onednn/matmul_reshape_add_pass.cc
```diff
@@ -0,0 +1,178 @@
...
+
+      return false;
+    });
+
+    paddle::drr::ResultPattern res = pat.ResultPattern();
+    const auto &in_num_col_dims_attr =
+        res.ComputeAttr([](const paddle::drr::MatchContext &match_ctx) -> int {
+          auto x_dims = pir::GetShapeFromValue(match_ctx.Tensor("x"));
+          auto reshape_x_dims = match_ctx.Attr<std::vector<int64_t>>("value");
+          int target = reshape_x_dims[0];
```
LLee233: For real model, usually the dim_0 of input is `-1` (cause batch size is not decided yet), a.k.a. [-1, m, n, k], and thus reshape could be [-1, m\*n\*k] or [-1, n\*k] or [-1, k]. So instead of directly using `reshape_x_dims[0]`, I recommend refer to [here](https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/ir_adaptor/translator/op_translator.cc#L1792) and calculate the attribute.

yuanlehome (approved):

PR #66059, opened 2024-07-15
Fix new comm dev ctx

Xing-lil:
### PR Category

Others

### PR Types

Bug fixes

### Description

Pcard-70448
Bug: Comm op are failing during training, the size of the comm group does not match in new op + new comm lib.
Fix bug with reduce_scatter in new comm lib related to dev_ctx and comm_ctx.  New comm should employ the dev_ctx maintained by comm_ctx_manager.  To adjust other op, only reduce_scatter return dev_ctx for now.
The accuracy and performance of the llama-13b and gpt3-13B-en models before and after the code change have be aligned.

### DIFF
--- a/paddle/fluid/framework/new_executor/instruction/instruction_util.cc
+++ b/paddle/fluid/framework/new_executor/instruction/instruction_util.cc
@@ -134,6 +134,27 @@ platform::DeviceContext* ParseDeviceContext(
       }
       return dev_ctx;
     }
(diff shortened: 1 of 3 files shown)


liym27 (approved):
LGTM

zhangbo9674 (approved):

zhiqiu (approved):
LGTM

PR #66077, opened 2024-07-16
[Typing][B-30] Add type annotations for `python/paddle/device/__init__.py`

enkilee:
### PR Category

User Experience

### PR Types

Improvements

### Description

- https://github.com/PaddlePaddle/Paddle/issues/65008
B30

### DIFF
--- a/python/paddle/device/__init__.py
+++ b/python/paddle/device/__init__.py
@@ -13,9 +13,13 @@
 # limitations under the License.
 
 # TODO: define the functions to manipulate devices
+
+from __future__ import annotations
+
 import ctypes
 import os
 import re
+from typing import TYPE_CHECKING, Union
 
 import paddle
 from paddle.base import core, framework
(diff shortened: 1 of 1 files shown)


python/paddle/device/__init__.py
```diff
@@ -132,7 +141,7 @@ def is_compiled_with_xpu():
     return core.is_compiled_with_xpu()
 
 
-def XPUPlace(dev_id):
+def XPUPlace(dev_id: int):
```
megemini:
```suggestion
if TYPE_CHECKING:
    from paddle import XPUPlace
def XPUPlace(dev_id: int) -> XPUPlace:
```

上面有个 `IPUPlace` 也标一下吧 ～

enkilee: 收到

SigureMo (approved):
LGTMeow

PR #66098, opened 2024-07-16
upgrade matmul_grad inferspmd to support shard optimizer stage3.

winter-wang:
### PR Category

Auto Parallel

### PR Types

Improvements 

### Description

- upgrade matmul_grad inferspmd to support shard optimizer stage3.
    - In sharding stage3 , the parameter is sharded both dp&mp dimensions。If the parameter is the 'y' input of matmul op, it will call c_allgather in dp dimension for execute matmul op.   In  old version, the matmul grad op will execute in all replicated。 Therefore, we upgrade thre inferspmd function of the matmul_grad op to avoid this situation.
- improve the shard accumulator in stage3.
    - If the paramater not attach mesh attribue in dygraph, we can't shard it in stage3.  In this case, we will only shard the moment for this parameter.

### Other
Pcard-67164

### DIFF
--- a/paddle/phi/infermeta/spmd_rules/matmul.cc
+++ b/paddle/phi/infermeta/spmd_rules/matmul.cc
@@ -285,11 +285,12 @@ static bool DistAttrsAreBasicallyEqual(
           in_dist_attr.partial_status() == out_dist_attr.partial_status());
 }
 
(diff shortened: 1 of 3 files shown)


zhiqiu (dismissed):
LGTM

zhiqiu (approved):
LGTM

PR #66140, opened 2024-07-17
this is a test for removing cpu test in some case when compiled with gpu

0x3878f:
### PR Category

Others

### PR Types

Others

### Description

Use GPU when compiling, do not repeat testing on CPU
Modified several cases to determine whether the pipeline can pass and the performance improvement it brings

### DIFF
--- a/test/legacy_test/test_activation_nn_grad.py
+++ b/test/legacy_test/test_activation_nn_grad.py
@@ -44,9 +44,11 @@ class TestSigmoidTripleGradCheck(unittest.TestCase):
 
     def test_grad(self):
         paddle.enable_static()
-        places = [base.CPUPlace()]
+        places = []
         if core.is_compiled_with_cuda():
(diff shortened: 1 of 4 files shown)


test/legacy_test/test_activation_nn_grad.py
```diff
@@ -377,9 +391,11 @@ def func(self, place):
 
     def test_grad(self):
         paddle.enable_static()
-        places = [base.CPUPlace()]
+        places = []
         if core.is_compiled_with_cuda():
-            places = [base.CUDAPlace(0)]
+            places.append(base.CUDAPlace(0))
+        else:
+            places.append(base.CPUPlace())
```
risemeup1, line 398: 都可以简化下代码：places = ['gpu' if paddle.is_compiled_with_cuda() else 'cpu'] 这样是不是好看点儿？

PR #66213, opened 2024-07-18
【Fix PIR Unittest No.209 202 320 BUAA】Fix some test case in PIR

uanu2002:
### PR Category

Others

### PR Types

Others

### Description

fix some test case in PIR.


wanghuancoder (commented):
和上一个PR一样

PR #66244, opened 2024-07-19
【Fix PIR Unittest NO.73 BUAA】Use OldIrGuard so that the test can be run

Wizard-ZP:
### PR Category
Others 

### PR Types
Others

### Description
Use OldIrGuard so that the test can be run


wanghuancoder (commented):
PIR下应该这么写：
            if in_pir_mode():
                paddle.pir.core.create_parameter(

PR #66266, opened 2024-07-19
[Auto Parallel] fix save load state_dict

xingmingyyj:
### PR Category

Auto Parallel 

### PR Types

Others

### Description

fix save load state_dict

### DIFF
--- a/python/paddle/distributed/checkpoint/load_state_dict.py
+++ b/python/paddle/distributed/checkpoint/load_state_dict.py
@@ -32,6 +32,7 @@ from .utils import (
 class ReadItem:
     local_tensor_index: LocalTensorIndex
     rank: int
+    dtype: str
     cur_offset: Tuple[int]
     storage_offset: Tuple[int]
     lengths: Tuple[int]
@@ -331,7 +332,7 @@ def get_read_items(path, state_dict, process_group, use_dist):
                 global_offset = (
                     tuple([0] * len(val.shape)) if len(val.shape) > 0 else ()
                 )
-            cur_chunk_metadata = LocalTensorMetadata(global_offset, local_shape)
+            cur_chunk_metadata = LocalTensorMetadata(global_offset, local_shape, str(val.dtype).split(".")[1])
             assert (
                 tensor_key in storage_state_dict_metadata
             ), f"tensor_key:{tensor_key} not found in storage_state_dict_metadata:{storage_state_dict_metadata}."
@@ -353,6 +354,7 @@ def get_read_items(path, state_dict, process_group, use_dist):
                     ReadItem(
                         storage_local_tensor_index,
                         paddle.distributed.get_rank(),
+                        storage_local_tensor_metadata.dtype,
                         tuple(cur_offsets),
                         tuple(storage_offsets),
                         tuple(lengths),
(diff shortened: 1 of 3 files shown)


python/paddle/distributed/checkpoint/load_state_dict.py
```diff
@@ -437,6 +439,15 @@ def load_state_dict(
         rank_to_files, missing_keys = get_rank_to_files(
             path, flat_state_dict, process_group, use_dist
         )
+        
+        gloabl_rank_to_files = []
```
zhangbo9674: get_rank_to_files 已经是通过 global_data_files 分析的结果，这里还需要 all_gather_object(gloabl_rank_to_files)么？

xingmingyyj: 这里不需要all_gather,但是get_rank_to_files中要做一些修改。动半和静半得到的state_dict不同，在静半下，state_dict是部分的，所以的到的necessary_files也是部分的，所以需要对necessary_files做一下all_gather。

zhangbo9674 (approved):
LGTM

python/paddle/distributed/checkpoint/load_state_dict.py
```diff
@@ -549,9 +566,11 @@ def load_state_dict(
                         storage_chunk_tensor, src=src_rank, group=process_group
                     )
                 else:
+                    tmp_tensor = paddle.assign(cur_chunk_tensor)
```
zhiqiu, line 569: why use tmp_tensor, plz add comments.

python/paddle/distributed/checkpoint/load_state_dict.py
```diff
@@ -533,14 +551,13 @@ def load_state_dict(
...
-                    dtype=flat_state_dict[
-                        item.local_tensor_index.tensor_key
-                    ].dtype,
+                    item.dtype,
                 )
 
             if src_rank == item.rank:
-                # assign value locally
-                paddle.assign(storage_chunk_tensor, cur_chunk_tensor)
+                if src_rank == paddle.distributed.get_rank():
```
zhiqiu, line 558: what about the else branch?

python/paddle/distributed/checkpoint/metadata.py
```diff
@@ -24,6 +24,7 @@ class LocalTensorMetadata:
 
     global_offset: Tuple[int]
     local_shape: Tuple[int]
+    dtype: str
```
zhiqiu, line 27: why we need dtype?

python/paddle/distributed/checkpoint/load_state_dict.py
xingmingyyj: The memory held by cur_chunk_tensor may be non-contiguous, and the broadcast API does not support this type of tensor.

python/paddle/distributed/checkpoint/load_state_dict.py
xingmingyyj: The condition ``src_rank == item.rank `` will be satisfied by all ranks, but only one rank needs to perform the assignment operation. Additionally, when ``src_rank != paddle.distributed.get_rank()``, ``storage_chunk_tensor`` may be None.

python/paddle/distributed/checkpoint/metadata.py
xingmingyyj: In static mode, the ``state_dict`` is incomplete, so the previous code would trigger  ``key error``. Here, the ``dtype`` is stored in advance.

python/paddle/distributed/checkpoint/load_state_dict.py
xingmingyyj: We’ll add comments in the next PR.

PR #66274, opened 2024-07-19
fix ut used in `static.nn.conv2d` part

ooooo-create:
### PR Category

Others

### PR Types

Others

### Description

测试使用 `paddle.nn.Conv2d` 替代部分 `static.nn.conv2d`


wanghuancoder (approved):
LGTM

PR #66338, opened 2024-07-22
【Comm】switch c_allgather in fluid to all_gather in python call

liym27:
### PR Category

Communication Library

### PR Types

Improvements

### Description

1. replace `c_allgather` with `all_gather` everywhere

PCard-85618


zhiqiu (approved):
LGTM

PR #66541, opened 2024-07-25
fix use_promote case

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

fix use_promote case
Pcard-76459


From00 (approved):
LGTM

PR #66748, opened 2024-07-29
[CINN] fix split symbolic loop

BiynXu:
### PR Category
CINN

### PR Types
Bug fixes

### Description
Pcard-74042
Fix Split primitive on symbolic loop.
For example, when split a loop with extent S by factor=(-1, 4, 4), we will get a loop nest of (S / 4 + 1, 4).
This PR modify it to （(S - 3) / 4, 4）.

### DIFF
--- a/paddle/cinn/ir/schedule/impl/loop_transformation.cc
+++ b/paddle/cinn/ir/schedule/impl/loop_transformation.cc
@@ -116,8 +116,7 @@ std::vector<Expr> DyScheduleImpl::Split(const Expr& loop,
   bool is_positive = true;
   int num_minus1 = 0;
   std::vector<Expr> process_factors;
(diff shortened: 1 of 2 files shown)


paddle/cinn/ir/schedule/impl/loop_transformation.cc
```diff
@@ -116,32 +116,28 @@ std::vector<Expr> DyScheduleImpl::Split(const Expr& loop,
   bool is_positive = true;
   int num_minus1 = 0;
   std::vector<Expr> process_factors;
-  Expr prod_size(-1);
+  int64_t prod_size(-1);
   int idx_neg1 = 1;
```
Vvsmile: Delete

PR #66923, opened 2024-08-01
[CINN] Use global buffer when buffer size contains symbolic

Dmovic:
### PR Category

Performance Optimization

### PR Types

Others

### Description

[CINN] Use global buffer when  buffer size contains symbolic


jiahy0825 (approved):
LGTM

PR #66955, opened 2024-08-02
fix the unit test: test_semi_auto_llama_acc_align

winter-wang:
### PR Category

Auto Parallel
### PR Types

Bug fixes
### Description

- set “FLAGS_enable_pir_api=1” in test_semi_auto_llama_acc_align.py， and it will. take effect in PR-CI-Auto-Parallel.
- fix inferspmd function error for reshape_grad.
- remove the "execution_stream" attribute  of all_gather op, which will lead to the communication group's mistake in executor.
- fix the insert_point's prev funciton. which will cause segment fault while set op_role for the last op in the block.

### Other

Pcard-67164

### DIFF
--- a/paddle/fluid/pybind/pir.cc
+++ b/paddle/fluid/pybind/pir.cc
@@ -1566,7 +1566,7 @@ void BindInsertionPoint(pybind11::module *m) {
                   "The insertion point is already at the begin and can't call "
                   "prev()."));
             }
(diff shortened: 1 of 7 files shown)


JZ-LIANG (dismissed):
LGTM

zhiqiu (approved):
LGTM

PR #66969, opened 2024-08-02
Make fuse_params=False to avoid OOM when broadcasting parameters

sneaxiy:
### PR Category
Distributed Strategy

### PR Types
Improvements

### Description
Make fuse_params=False to avoid OOM when broadcasting parameters.
Pcard-67164


wawltor (approved):
LGTM

PR #66970, opened 2024-08-02
Make set_value early return when id(self) == id(value)

sneaxiy:
### PR Category
Performance Optimization

### PR Types
Performance

### Description
Make set_value early return when id(self) == id(value).
Pcard-67164


ForFishes (approved):
LGTM

PR #66978, opened 2024-08-02
modify get_and_set_rng_state to use _current_expected_place_

jeff41404:
### PR Category
User Experience

### PR Types
Bug fixes

### Description
pcard-85936
Recently, I use model to verify auto parallel, and found that when  dynamic to static use PIR will have a error (running without PIR or dynamic to static can still pass), and error occurs when processing the following model code:
`with get_rng_state_tracker().rng_state("global_seed"):`
BA664EBB6827007EE7E288A88DCD6676

without PIR, `_current_Inspected_place()` returns `Place (gpu: 0)`, which is right. However, with PIR, it returns `Place (undefined: 0)` causing an error.
other PR mentioned this issue: #57583
After Discussing with @wanghuancoder , modifying `get_rng_state` and `set_rng_state` to call `_current_Inspected_place_()` can solve this problem.


wanghuancoder (approved):
LGTM

PR #66980, opened 2024-08-02
Fix pipeline parallel error of using loss in if statement and fix error of numel

jeff41404:
### PR Category
User Experience

### PR Types
Bug fixes

### Description
pcard-85940

1. in pipeline parallel, error will occur in some device when using loss in if statement. e.g.

```python
import paddle

dense_tensor = paddle.to_tensor([[1,2], [3,4]])
mesh0 = dist.ProcessMesh([0,1], dim_names=['dp'])
mesh1 = dist.ProcessMesh([2,3], dim_names=['dp'])

lossmask = dist.shard_tensor(dense_tensor, mesh1, placements)
loss = (~lossmask).all()
if loss:  # will error in device 0 and 1
    print("true branch")
else:
    print("false branch")
```
9645dbf570d5e4efe99587534f6c1f27
The root cause is that when using `loss`(Tensor) in if statement, Tensor's __nonzero__ method will be called, which includes code of `assert self._is_initialized(), "tensor not initialized"` and this assert may not be satisfied on some devices. because in the pipeline parallel, some devices do not have this tensor(loss). 

&nbsp;
2. in addition, when DistTensor uses method `numel()`, an error will also be reported. the reason for this problem is due to a defect in spmd rules of method numel. For example, the following code will report an error:

```python
loss.numel() # when loss is DistTensor, this will error
```
3E96D759543198E9479DC673EF5C7DDA

this PR fix two above.


zhiqiu (approved):
LGTM

PR #67052, opened 2024-08-05
[Dist Dialect] Fix the bug of PIR infer_spmd when op inputs are inconsistent

pkuzyc:
### PR Category
Auto Parallel

### PR Types
Bug fixes

### Description
Pcard-67164

The input args of infer_meta and infer_spmd function may be different from the operation's input. The original PIR InferMeta function has not handled this case. This pr fix this bug. 

For example, in StackGrad op, its operation inputs are [x, out_grad], while its spmd_function's inputs are [out_grad].

- The original InferSpmd part in InferMeta function:

```
auto spmd_info = phi::distributed::StackGradInferSpmd(dist_meta_out_grad, axis);
DebugInfoForInferSpmd("StackGradOp", spmd_info);
// Raise error here, because its infer spmd function has only one input, so 
// spmd_info.first.size() is 1u. 
PADDLE_ENFORCE_EQ(spmd_info.first.size(), 2u, common::errors::Unavailable(
    "Size of spmd_info.first for op[SumGradOp]is unexpected."));

// The input x's dist_attr is not in spmd_info, so also incorrect here.
for(auto& arg_dist : spmd_info.first) {{
    dist_operand_attrs.push_back(CvtToPirAttr(arg_dist));
}}
```
- After fix the bug: 
```
auto dist_meta_out_grad = CvtToDistMetaTensor(out_grad_.type().dyn_cast<DistDenseTensorType>());
auto spmd_info = phi::distributed::StackGradInferSpmd(dist_meta_out_grad, axis);
DebugInfoForInferSpmd("StackGradOp", spmd_info);
// spmd_info.first.size() is equal to infer_spmd function's input size
PADDLE_ENFORCE_EQ(spmd_info.first.size(), 1u, common::errors::Unavailable(
    "Size of spmd_info.first for op[StackGradOp]is unexpected."));

// Get the dist_attr from operation input
dist_operand_attrs.push_back(GetTensorDistAttrArray(x));
dist_operand_attrs.push_back(CvtToPirAttr(spmd_info.first[0]));
```

### DIFF
--- a/paddle/fluid/pir/dialect/op_generator/op_infermeta_func_gen.py
+++ b/paddle/fluid/pir/dialect/op_generator/op_infermeta_func_gen.py
@@ -793,7 +793,8 @@ def GenDistBranch(args, op_info):
     {}
     CvtAllInputsToDist(input_values, op_mesh);
     auto ctx = pir::IrContext::Instance();
-    std::vector<pir::Attribute> dist_operand_attrs, dist_result_attrs;"""
+    std::vector<pir::Attribute> dist_operand_attrs, dist_result_attrs;
+    """
 
     extra_call = ""
     for name in op_info.spmd_params:
@@ -810,8 +811,10 @@ def GenDistBranch(args, op_info):
         extra_call = "CopyLeafOpToMesh(scale_, op_mesh);"
     dist_branch_str = TEMPLATE.format(merge_input_meshes, extra_call)
     infer_spmd_args_list = []
+    spmd_input_value_num = 0
+    map_input_idx = {}
     # Prepare inputs_meta_tensor & attributes for infer spmd
-    for name in op_info.spmd_params:
+    for spmd_arg_idx, name in enumerate(op_info.spmd_params):
         # is input
         if name in op_info.input_name_list:
             input_index = op_info.input_name_list.index(name)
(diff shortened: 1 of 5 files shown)


paddle/fluid/pir/dialect/op_generator/op_infermeta_func_gen.py
```diff
@@ -856,26 +861,52 @@ def GenDistBranch(args, op_info):
...
         "Size of spmd_info.first for op[{op_name}]is unexpected."));
-    for(auto& arg_dist : spmd_info.first) {{
-        dist_operand_attrs.push_back(CvtToPirAttr(arg_dist));
-    }}
 """
     dist_branch_str += TEMPLATE.format(
         spmd_func=spmd_rule_func,
         args=', '.join(infer_spmd_args_list),
-        input_size=len(op_info.input_name_list),
+        input_size=spmd_input_value_num,
```
JZ-LIANG, line 868: why not keep the constrain which forces the infermate and inferspmd to have the same input signature ?

the inferspmd of stack_grad could have redundant input to meet the above constrain.

pkuzyc: The input args of infer_spmd is consistent with infer_meta. The problem is that the auto parallel part in PIR has not handled the case when the infer_meta args (i.e. infer_spmd args) are inconsistent with the op input.

winter-wang (dismissed):

winter-wang (approved):

PR #67083, opened 2024-08-06
[CINN] Simplify the index expression of Reshape

lshpku:
### PR Category
CINN

### PR Types
Improvements

### Description
This PR introduces an optimization for the `pe::Reshape` index calculation. The previous implementation involved complex calculations to determine the input tensor indices from the output tensor indices, by computing the absolute offset in the output and then back-calculating each dimension of the input tensor. For example, transforming an input tensor `x` of shape `[S0, S1, S2, 128]` to an output tensor `y` of shape `[S0, S1*S2, 32, 4]` required calculating the `offset` as `offset=((i*S1*S2+j)*32+k)*4+m`, and then converting it to the input indices using `[offset/(S1*S2*128), offset/(S2*128)%S1, offset/128%S2, offset%128]`.

The optimization leverages the correspondence between input and output axes, for instance, `[32, 4]` corresponding to `[128]`, and `[S1*S2]` corresponding to `[S1, S2]`. The new approach generates each axis index directly based on this correspondence, resulting in simpler and more efficient calculations. For the given example, the optimized indices are computed as `[i, j/S2, j%S2, k*4+m]`.

pcard-85711

### DIFF
--- a/paddle/cinn/hlir/pe/elementwise.cc
+++ b/paddle/cinn/hlir/pe/elementwise.cc
@@ -19,6 +19,8 @@
 
 #include "paddle/cinn/common/cas.h"
 #include "paddle/cinn/common/dim_expr_converter.h"
+#include "paddle/cinn/common/integer_set.h"
+#include "paddle/cinn/hlir/framework/pir/trivial_op_util.h"
 #include "paddle/cinn/hlir/op/op_util.h"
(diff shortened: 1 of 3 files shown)


jiahy0825 (dismissed):
LGTM

jiahy0825 (approved):
LGTM

PR #67085, opened 2024-08-06
[OneDNN] Fix Prelu kernel failed in bf16 since data type error

zhanglirong1999:
### PR Category
Operator Mechanism

### PR Types
 Bug fixes

### Description
We found in bf16, prelu kernel will failed. Since in bf16, many inputs such as filter, bias, scale, alpha do not require quantization.
But Prelu kernel input 'alpha', it use x data type as it's data type, which will failed if x is bf16.
Refer to clip, fc, layernorm kernel, such kinds of input can use fp32 as it's type directly.


LLee233 (approved):
LGTM

PR #67101, opened 2024-08-06
refactor CvtValueToDist function to support all cases.

winter-wang:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

refactor CvtValueToDist function to support special cases.  The function will add replicated distributed attribtue for   all connected subgraph.

### Other
Pcard-67164


From00 (approved):
LGTM

PR #67156, opened 2024-08-07
fix dataset example

Xingyyy01:
### PR Category

User Experience

### PR Types

Others

### Description

Fix dataset example: Add output; Remove the use of external library numpy


sunzhongkai588 (commented):
dataset 一般都用 numpy，可以不改

PR #67247, opened 2024-08-09
[Dygraph] fix calc_stream with unnecessary event

LiYuRio:
### PR Category

Performance Optimization

### PR Types

Bug fixes

### Description

去除计算流下不必要的创建event的开销

PR #67315, opened 2024-08-12
fix calc_stream with unnecessary event (#67247)

LiYuRio:
### PR Category

Performance Optimization
### PR Types

Performance

### Description

去除计算流下不必要的创建event的开销
Pcard-67164


ForFishes (approved):
LGTM

PR #67424, opened 2024-08-14
predictor.run(list(paddle.Tensor)) support dist

yuanlehome:
### PR Category

Inference

### PR Types

New features

### Description

pcard-71500
predictor.run(list(paddle.Tensor)) support distribute inference


zhoutianzi666 (approved):
LGTM

PR #67441, opened 2024-08-14
Fix pipeline parallel error of using tensor method item

jeff41404:
### PR Category
User Experience

### PR Types
Bug fixes

### Description
pcard-86208
In pipeline parallel, error will occur in some device when using Tensor method `item()`. e.g.

```python
import paddle

dense_tensor = paddle.to_tensor([[1,2],
                                 [3,4]], stop_gradient=False)
mesh0 = dist.ProcessMesh([0,1], dim_names=['dp'])
mesh1 = dist.ProcessMesh([2,3], dim_names=['dp'])

t1 = dist.shard_tensor(dense_tensor, mesh0, placements)
t2 = t1 * 2
lossmask = dist.reshard(t2, mesh1, placements)
print(lossmask.item(0))  # will error in device 0 and 1
```
image

Tensor's method `item()` calls method `_getitem_from_offset`, which has code of `PADDLE_ENFORCE(dist_tensor->initialized(), ...)` and this assert may not be satisfied on some devices. because in the pipeline parallel, some devices do not have this data. this PR will fix this error.

### DIFF
--- a/python/paddle/base/dygraph/tensor_patch_methods.py
+++ b/python/paddle/base/dygraph/tensor_patch_methods.py
@@ -857,6 +857,11 @@ def monkey_patch_tensor():
                 3.299999952316284
 
         """
+        # resolve the error issue in scenario of pipeline parallel
+        # where some devices do not have self data, return None does not affect
+        # the execution result in those devices, so currently we return None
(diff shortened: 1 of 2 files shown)


From00 (approved):
LGTM

PR #67535, opened 2024-08-19
add unit test for auto parallel amp acc align.

winter-wang:
### PR Category

Auto Parallel

### PR Types

Devs

### Description

add unit test for amp auto parallel.

### Other
Pcard-67164


JZ-LIANG (approved):
LGTM

PR #67588, opened 2024-08-20
fix einsum issue

jzhang533:
### PR Category
User Experience

### PR Types
Bug fixes

### Description

fix the following use case, which is simplified from https://github.com/arogozhnikov/einops/blob/master/tests/test_einsum.py

```python
import paddle
import numpy as np

p = "..., ..., ... -> ..."
in_shapes = ((3, 2), (3, 2), (3, 2))

in_arrays = [np.random.uniform(size=shape).astype("float32") for shape in in_shapes]
paddle_in_arrays = [paddle.to_tensor(array) for array in in_arrays]

paddle.einsum(p, *paddle_in_arrays)
```

this bug seems being introduced in #61348, this PR is only a workaround.

please ignore the following
PCard-66972

### DIFF
--- a/python/paddle/tensor/einsum.py
+++ b/python/paddle/tensor/einsum.py
@@ -842,7 +842,7 @@ def parse_fake_shape(
             "length of shape and length of label must be the same, but received %d != %d"
             % (len(op.shape), len(label))
         )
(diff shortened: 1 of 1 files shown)


2742195759 (approved):
LGTM

PR #67589, opened 2024-08-20
mix2dist_pass support shard randomly sampled data

jeff41404:
### PR Category
Auto Parallel

### PR Types
New features

### Description
pcard-86321
The Dit/LargeDit model sample a large amount of random data during its training. Especially when running in a distributed environment, these sampled random data also need to be sharded into different devices to reduce memory usage.
But at the beginning of `mix2dist_pass` design in PIR mode, it only supported the shard of input data and model parameter, and did not support the shard of other data. this PR will expand this capability.
```python
import paddle
process_mesh = dist.ProcessMesh([0, 1], dim_names=['mp'])

# before this PR, only support
linear1 = paddle.nn.Linear(100, 200)
linear1.weight = dist.shard_tensor(linear1.weight, process_mesh, [dist.Shard(1)]) # can shard input data and model parameter
noise = paddle.randn(x.shape)
noise = dist.shard_tensor(noise, process_mesh, [dist.Replicate()) # In other cases, placements must be all Replicate, otherwise an error will be reported

# after this PR, also support
timesteps = paddle.randint(0, self.num_timesteps, (x.shape[0],))
timesteps = dist.shard_tensor(timesteps, process_mesh, [dist.Shard(1))  # can shard randomly sampled data(int)
noise = paddle.randn(x.shape)
noise = dist.shard_tensor(noise, process_mesh, [dist.Shard(1))  # can shard randomly sampled data(float)
```

### DIFF
--- a/test/auto_parallel/pir/test_static_pir_program.py
+++ b/test/auto_parallel/pir/test_static_pir_program.py
@@ -204,6 +204,30 @@ class TestBuildFakeProgram(unittest.TestCase):
         self.assertEqual(input2.dist_attr().process_mesh, mesh)
         self.assertEqual(input2.dist_attr().dims_mapping, [-1, 0])
 
+        # check full_int_array op result dist_attr
+        input1_shape = input1.get_defining_op().operand_source(0)
+        input1_shape_op_dist_attr = input1_shape.get_defining_op().dist_attr
+        tensor_dist_attr = input1_shape_op_dist_attr.result(
+            0
+        ).as_tensor_dist_attr()
+        self.assertEqual(tensor_dist_attr.process_mesh, mesh)
+        self.assertEqual(tensor_dist_attr.dims_mapping, [-1])
+
+        input2_shape = input2.get_defining_op().operand_source(0)
+        input2_shape_op_dist_attr = input2_shape.get_defining_op().dist_attr
+        tensor_dist_attr = input2_shape_op_dist_attr.result(
+            0
+        ).as_tensor_dist_attr()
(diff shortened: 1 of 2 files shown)


test/auto_parallel/pir/test_static_pir_program.py
```diff
@@ -141,6 +144,66 @@ def test_build_with_shard_tensor(self):
...
+
+        # after apply_mix2dist_pass, the program changed
+        self.assertTrue(main_program.num_ops() == 4)
+
+        self.assertTrue(input1.is_dist_dense_tensor_type())
+        self.assertTrue(input2.is_dist_dense_tensor_type())
+
+        self.assertTrue(input1.get_defining_op().has_attr("op_dist_attr"))
+        self.assertTrue(input2.get_defining_op().has_attr("op_dist_attr"))
+
```
JZ-LIANG, line 188: the operations appended implicitly by rand-related-op should be verified 

in your case here, should check the dist_attr for :
 input1.get_defining_op.operand[1] 
and 
input1.get_defining_op.operand[1].defining_op()

JZ-LIANG (dismissed):
LGTM and revise in next PR

test/auto_parallel/pir/test_static_pir_program.py
jeff41404: thanks, already added

jeff41404:
> LGTM and revise in next PR

thanks, it would be better to add the necessary unit test cases in this PR, I commit it,

JZ-LIANG (approved):
LGTM

PR #67655, opened 2024-08-22
[CINN] Fix reduce fusion without reduce axis reuse

huangjiyi:
### PR Category

CINN

### PR Types

Bug fixes

### Description

Pcard-76996
- Fix reduce fusion without reduce axis dim reuse


2742195759 (approved):
LGTM

PR #67700, opened 2024-08-25
fix numpy2.1 bug when python >=3.10

risemeup1:
### PR Category

Execute Infrastructure

### PR Types

Others

### Description

Fix the bug in PaddlePaddle under Python 3.10 and NumPy 2.1.

image

- For any `__array__` method on a non-NumPy array-like object, dtype=None and copy=None keywords must be added to the signature - this will work with older NumPy versions as well (although older numpy versions will never pass in copy keyword). If the keywords are added to the `__array__` signature, then for:

    copy=True and any dtype value always return a new copy,
    
    copy=None create a copy if required (for example by dtype),

    copy=False a copy must never be made. If a copy is needed to return a numpy array or satisfy dtype, then raise an exception (ValueError).

- Restricting numpy to version 1.21 or above is to fix the bug of not finding the ‘NDArray’ attribute when using the `numpy.typing` module.

To do：
Modify the logic in TensorToPyArray(self) according to the dtype and copy parameters.
pcard-67164

### DIFF
--- a/paddle/fluid/pybind/tensor.cc
+++ b/paddle/fluid/pybind/tensor.cc
@@ -203,12 +203,13 @@ void BindTensor(pybind11::module &m) {  // NOLINT
   g_framework_tensor_pytype =
       reinterpret_cast<PyTypeObject *>(framework_tensor.ptr());
   framework_tensor
-      .def("__array__",// NOLINT
-           //To do：Modify the logic in TensorToPyArray(self) according to the dtype and copy parameters.
-           [](phi::DenseTensor &self, py::object dtype, py::object copy) {
-             return TensorToPyArray(self); },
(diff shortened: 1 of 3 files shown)


jzhang533 (dismissed):
this will fix #67602 .
but probably more work are needed for : https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword

SigureMo (dismissed):
LGTMeow

xuxinyi389:
LGTM

jzhang533 (dismissed):
LGTM

SigureMo (dismissed):

SigureMo (approved):

PR #67756, opened 2024-08-27
s_to_r reshard support unbalanced split

jeff41404:
### PR Category
Auto Parallel

### PR Types
New features

### Description
pcard-86471
s_to_r reshard should support unbalanced split or will error in some model (Dit etc)
```python
import paddle
process_mesh = dist.ProcessMesh([0, 1], dim_names=['dp'])

class ParallelLabelEmbedder(nn.Layer):
    def __init__(self, num_classes, hidden_size, dropout_prob):
        ......
        self.embedding_table = nn.Embedding(embedding_dim, hidden_size) # parameter shape=(1001, 1024)
        self.embedding_table.weight = dist.shard_tensor(self.embedding_table.weight, process_mesh, [dist.Replicate()])
        # Although we DO NOT shard the parameters in the forward and backward processes, 
        # But when using sharding parallel(eg. sharding stage 1), the optimizer will shard the parameters
        # at the 0th dimension, (1001/2) resulting unbalanced splitting. 
        # and using all_gather to update parameters, an error will occur
```
image

### DIFF
--- a/paddle/fluid/pir/dialect/distributed/ir/dist_tools.cc
+++ b/paddle/fluid/pir/dialect/distributed/ir/dist_tools.cc
@@ -377,7 +377,9 @@ pir::Attribute CreateReplicatedDistAttr(pir::Type prim_type,
   }
   return nullptr;
 }
-pir::Type CvtToPirDistType(pir::Type global_type, pir::Attribute dist_attr) {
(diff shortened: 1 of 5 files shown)


JZ-LIANG (approved):
LGTM

PR #67762, opened 2024-08-28
[CINN] Fix type mismatch when inline generated dynamic shape into concat 

huangjiyi:
### PR Category

CINN

### PR Types

Bug fixes

### Description

Pcard-76996
- Fix type mismatch when inline generate dynamic shape into concat


2742195759 (approved):
LGTM

PR #67765, opened 2024-08-28
[CINN][BUAA][Infer Symbolic Shape BUAA No.109]Add stanh op

Marcusryz:
### PR Category

CINN

### PR Types

Others

### Description

Add stanh op.

输入输出形状一致


third_party/flashattn
gongshaotian, line 1: 引入了一些无关修改

PR #67799, opened 2024-08-29
[OneDNN][PIR] Remove concat in bf16 pass for googlenet error

zhanglirong1999:
### PR Category
Inference

### PR Types
Bug fixes

### Description
Remove concat in bf16 pass for googlenet error. We found concat op has two input x and axis, but x is not a DenseTensorType like others, it's a vector of DenseTensorType, which will cause error since Quantize limit input must be a DenseTensorType. We have temporarily remove concat in bf16. In the future, we need to figure out how to implement bf16 on Concat, how to determine the number of DenseTensorType from an input node of x, and then split it into multiple Quantize.


LLee233 (approved):
LGTM

PR #67809, opened 2024-08-29
[Auto Parallel] fix print bug of dist_tensor

zhangyuqin1998:
### PR Category

Auto Parallel 

### PR Types

Bug fixes

### Description

Fix print bug of dist_tensor. In the previous code, when a dist_tensor with BF16 dtype and Partial placement was printed, it was first cast to FP32, then converted to a numpy array using the `numpy` method. During this conversion, Partial placement is changed to Replicate, causing communication to occur under FP32 instead of the original BF16 dtype.

Therefore, we first reshard the dist_tensor to Replicate under the origina dtype, then convert it to a numpy array using the `numpy` method.

Pcard-76459

### DIFF
--- a/python/paddle/tensor/to_string.py
+++ b/python/paddle/tensor/to_string.py
@@ -384,7 +384,15 @@ def dist_tensor_to_string(tensor, prefix='Tensor'):
         )
     else:
         indent = len(prefix) + 1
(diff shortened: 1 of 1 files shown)


zhiqiu (approved):
LGTM

PR #67835, opened 2024-08-29
[PIR]Gen metaconfig in pir grad op

0x45f:
### PR Category

Execute Infrastructure

### PR Types

Bug fixes

### Description

[PIR]Gen metaconfig in pir grad op
Pcard-67164


SigureMo (approved):
LGTMeow

PR #68009, opened 2024-09-04
[CINN] Fix divided by zero error in analyzer

BiynXu:
### PR Category
CINN

### PR Types
Bug fixes

### Description
Pcard-74042
Fix divided by zero error in symbolic expr analyzer.
In SymbolicExprAnalyzer, if there is an exception in AutoSimplify when using interface ProveXXX(), it is determined that it cannot be proven and return std::nullopt.


Hongqing-work (approved):

Aurelius84 (approved):
LGTM for ENFORCE

PR #68021, opened 2024-09-05
[PIR][OneDNN] Add cpu bf16 placement pass check data type

zhanglirong1999:
### PR Category
Inference

### PR Types
Bug fixes

### Description
Fix cpu bf16 placement pass data type, add a check before replace


LLee233 (approved):
LGTM

PR #68029, opened 2024-09-05
[CINN] fix group cluster bug about sibling node

chen2016013:
### PR Category

CINN

### PR Types

Bug fixes

### Description

pcard-67164
fix group cluster bug about sibling node
funtion `GetConsumerOpSimple()` cannot get downstream `IfOp` if consumer op is used in If block


2742195759 (approved):
LGTM

PR #68042, opened 2024-09-05
[CINN] Fix horizontal reduce fusion with same loop but different reduce dims

huangjiyi:
### PR Category

CINN

### PR Types

Bug fixes

### Description

Pcard-76996
- Fix horizontal reduce fusion with same loop but different reduce dims
- Skip RT fusion when fake_reduce.size < reduce.size


2742195759 (approved):
LGTM

PR #68057, opened 2024-09-06
[PIR][oneDNN] Fix bf16 quantization for cast

LLee233:
### PR Category

Inference

### PR Types

Others

### Description

There are situations when `Cast` is receiving input other than `float32/bf16`, in which case `Cast` shall not be transformed to oneDNN op. Besides, `Cast` itself has the ability to transform tensor dtype, no need to add extra Quantize.

This PR is targeting to issue and improvement both as described above.

### DIFF
--- a/paddle/phi/ops/yaml/inconsistent/onednn_ops_extra.yaml
+++ b/paddle/phi/ops/yaml/inconsistent/onednn_ops_extra.yaml
@@ -32,7 +32,6 @@
 
 - op : cast
   dynamic_fallback : True
-  extra_args : str mkldnn_data_type="float32"
 
 - op : clip
   extra_args : str mkldnn_data_type="float32"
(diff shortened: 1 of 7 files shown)


paddle/phi/ops/yaml/inconsistent/onednn_ops_extra.yaml
```diff
@@ -32,7 +32,6 @@
 
 - op : cast
   dynamic_fallback : True
-  extra_args : str mkldnn_data_type="float32"
```
zhanglirong1999, line 35: Maybe also need to modify op_compat.yaml?

LLee233: I think it's fine, cause `op_compat.yaml` usually has the maximum set, which only reports error when adding not included attr(s).

zhanglirong1999 (dismissed):

zhanglirong1999 (approved):
LGTM

PR #68067, opened 2024-09-06
dist2dense_pass fix shape errors in shard randomly sampled data

jeff41404:
### PR Category
Auto Parallel

### PR Types
Bug fixes

### Description
pcard-86711
After PR #67589 supporting sharding of randomly sampled data, it is necessary to synchronously modify the shape in the attribute of previously operator of `pd_op.full_int_array` when use `paddle.distributed.to_static` and pir(`FLAGS_enable_pir_api=true`). Otherwise, the executor will still create data according to this wrong shape during actual execution, resulting in incorrect reporting of shape in subsequent calculations.
```python
import paddle
process_mesh = dist.ProcessMesh([0, 1], dim_names=['dp'])

noise = paddle.randn(x.shape)  
noise = dist.shard_tensor(noise, process_mesh, [dist.Shard(0)) # same shard as x
y = noise * x # will error when use paddle.distributed.to_static and pir
```
Printing pir program also shows that the shape in the attribute of operator of `pd_op.full_int_array` (line1 and line3) is inconsistent with correct shape (line2 and line4)
image
this PR will fix this issue.


JZ-LIANG (approved):
LGTM

PR #68095, opened 2024-09-09
[PIR][oneDNN] Optimize onednn_placement_pass

LLee233:
### PR Category

Inference

### PR Types

Improvements

### Description

For ops like `Scale` & `Slice`, there is no support for data type i32/i64, causing fallback at `pd_op_to_kernel_pass`. Thus this PR aims to prevent such ops turning into onednn op at early stage, in order to avoid extra fallback.


zhanglirong1999 (dismissed):
LGTM

zhanglirong1999 (approved):

PR #68123, opened 2024-09-10
To distributed api

lxd-cumt:
### PR Category
User Experience

### PR Types
New Features

### Description
High-level distribute api
pcard-66975


python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
```diff
@@ -0,0 +1,1487 @@
...
+        # out_linear
+        output = paddle.matmul(tmp, out_weight)
+        return output
+
+    def rotate_half(self, x):
+        x1 = x[..., : x.shape[-1] // 2]
+        x2 = x[..., x.shape[-1] // 2 :]
+        return paddle.concat([-x2, x1], axis=-1)  # shape is the same as x
+
+    def scale_dot_product_attention(
```
jeff41404: The code of this function is basically the same as the code of the function in `PIRScaleDotProductPattern` above, and `rotate_half`, `mlp`, etc. also have the same problem. complex networks should be a combination of basic networks, rather than a re implementation.

python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
```diff
@@ -0,0 +1,1487 @@
...
+        attn_output = attn_output.transpose([0, 2, 1, 3])
+
+        attn_output = attn_output.reshape([bsz, q_len, head_dim * num_heads])
+
+        return attn_output
+
+
+# Llama
+@register_pir_pattern
+class PIRMLPPattern(PIRBasePattern):
```
jeff41404: There are many variations of `MLP`, such as having 2 or 3 linear layers, with bias or no bias, and the activation function may also be different, so the name of `MLPPattern` needs to be carefully considered.

jeff41404: also, is it necessary to use the prefix "PIR"?

python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
```diff
@@ -0,0 +1,1487 @@
...
+            paddle.static.Program(),
+            paddle.static.Program(),
+        )
+        # data init
+        batch_size = 4
+        seq_length = 1024
+        num_heads = 32
+        head_size = 64
+        hidden_size = num_heads * head_size
+        intermediate_size = 4096
```
jeff41404: it is better to move all these configurations which required for each subclass to the parent class?

jeff41404: shall we set the shape smaller without losing generality and reducing memory usage.

python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
```diff
@@ -0,0 +1,1487 @@
...
+            result = {}
+            _match_core(src_start_op, tgt_op, is_op=True)
+            if not not_matched:
+                # print(f"matched one, check whether to append")
+                need_to_append = True
+                for value in result.values():
+                    if value in matched_op_node_ids:
+                        result = {}
+                        need_to_append = False
+                        break
```
jeff41404: Should this `break` be removed, such as `pattern=(matmul, add, matmul, add,)` and `prog=(matmul, add, matmul, add, matmul, add, Relu, ...)` will it cause a matching interruption?

python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
```diff
@@ -0,0 +1,1487 @@
...
+                        src_as_output_op
+                    )
+                )
+                if src_as_output_op_id not in result.keys():
+                    _match_core(src_as_output_op, tgt_as_output_op, is_op=True)
+
+    results = []
+    result = {}
+    matched_ids = set()
+    matched_op_node_ids = set()
```
jeff41404: It seems like there's no difference between these two set, just keep one?

python/paddle/distributed/auto_parallel/high_level_api.py
```diff
@@ -0,0 +1,508 @@
...
+        new_attn_out = paddle.transpose(attn_out, [1, 0, 2])
+        new_attn_out = dist.reshard(
+            new_attn_out, current_mesh, [dist.Shard(1), dist.Shard(0)]
+        )
+        # print(f"new_outputs are: {new_attn_out}")
+        # breakpoint()
+        return new_attn_out
+
+
+def reshard_transpose_mlp_layer_input(layer, inputs):
```
jeff41404: using sequence parallel but not requiring transpose in MLP, is it better to call `reshard_mlp_layer_input`?

python/paddle/distributed/auto_parallel/high_level_api.py
```diff
@@ -0,0 +1,508 @@
...
+        new_mlp_input = dist.reshard(
+            mlp_input, current_mesh, [dist.Shard(1), dist.Replicate()]
+        )
+        # print(f"new_mlp_input is {new_mlp_input}")
+        # breakpoint()
+        new_inputs[0] = new_mlp_input
+        return tuple(new_inputs)
+
+
+def transpose_reshard_mlp_layer_output(layer, inputs, outputs):
```
jeff41404: using sequence parallel but not requiring transpose in MLP, is it better to call `reshard_mlp_layer_output`?

lxd-cumt: done

python/paddle/distributed/auto_parallel/high_level_api.py
lxd-cumt: done

python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
lxd-cumt: done

python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
lxd-cumt: given that pir as default feature for paddlepaddle, remove all prefix "PIR" here

python/paddle/distributed/auto_parallel/static/tuner/pir_rule_based_tuner.py
lxd-cumt: done

python/paddle/distributed/auto_parallel/high_level_api.py
```diff
@@ -0,0 +1,465 @@
...
+            assert op in op_to_id.keys(), f"{op.name} is not in program"
+            op_id = op_to_id[op]
+            op_id_to_layer[op_id] = layer
+            ops_id.append(op_id)
+        ops_id_to_layer[tuple(ops_id)] = layer
+
+    # # step_5: pattern recogincation
+    DECODER_LAYER_NAME = 'decoder_layer'
+    register_used_patterns(DECODER_LAYER_NAME)
+    results = match_all_patterns(program)
```
jeff41404, line 274: logging can be added here and some key locations for easy debugging.

jeff41404:
It's better to have one more unit test of gpt3?

python/paddle/distributed/auto_parallel/high_level_api.py
lxd-cumt: I will add logging in next PR. Thanks

lxd-cumt:
> It's better to have one more unit test of gpt3?

I will try to add gpt3 test in the following PR.  Thx

jeff41404 (approved):

XiaoguangHu01 (approved):
LGTM

PR #68193, opened 2024-09-12
[PIR][oneDNN] Optimize Fc InferMeta

LLee233:
### PR Category

Inference

### PR Types

Improvements

### Description

During test on vit-ocr model, we found inconsistent Input Tensor and Perm in `Transpose` Op. After investigation, it turned out that the problem was relevant to InferMeta of its previous Op `onednn_op.fc`. It neglected attribute `fused_reshape2_shape`, which may change the dims of output of Fc.

Here I add specific InferMeta for `onednn_op.fc`, considering the impact of extra attribute `fused_reshape2_shape`.


zhanglirong1999 (approved):
LGTM

PR #68231, opened 2024-09-14
[Auto Parallel] change lookup_table_v2 auto cast for align mode

zhangyuqin1998:
### PR Category

Auto Parallel

### PR Types

Others

### Description

In auto parallel align mode, ensure lookup_table_v2 runs in FP32. By default, lookup_table_v2 is in the white_list, and runs in BF16/BF16. Users can add lookup_table_v2 to the amp_custom_black_list but cannot remove it from the default white_list.
If lookup_table_v2 appears in both the white_list and black_list, AMP will select it in BF16/BF16. Therefore, in auto parallel align mode, add lookup_table_v2 to the black_list and ensure it is not in the white_list.

Pcard-76459

### DIFF
--- a/python/paddle/amp/auto_cast.py
+++ b/python/paddle/amp/auto_cast.py
@@ -703,6 +703,16 @@ def amp_guard(
 
             # set amp op list
             original_white_list, original_black_list = tracer._get_amp_op_list()
+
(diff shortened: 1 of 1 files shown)


From00 (approved):
LGTM

PR #68241, opened 2024-09-14
[CINN] Improve recompute strategy to avoid being interrupted by unfusible op while fusing

chen2016013:
### PR Category

CINN

### PR Types

Improvements

### Description

Pcard-67164

The raw Re-compute strategy in use has't taken fusion information into account. Therefore the recomputed program might be interrupted while fusing because of unfusible op (If there is unfusible op on any path between 2 ops, they cannot be fused otherwise will cause loop in program). This might cause more IOs, leading worse performance.
This PR  Improve recompute strategy by analyzing loop condition while do recomputing.

### DIFF
--- a/python/paddle/decomposition/recompute.py
+++ b/python/paddle/decomposition/recompute.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import functools
 import math
 import os
+from collections import deque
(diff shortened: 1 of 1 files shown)


2742195759 (approved):
LGTM

PR #68288, opened 2024-09-18
fix the bug in create moment for auto parallel.

winter-wang:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

fix the bug in create moment for auto parallel.

### Other
Pcard-67164


zhangbo9674 (approved):
LGTM

PR #68314, opened 2024-09-19
fix bug in sharding stage1.

winter-wang:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

- fix the performance degradation caused by https://github.com/PaddlePaddle/Paddle/pull/68288.
- optimizer the sharding stage1.

### Other 
Pcard-67164


From00 (approved):
LGTM

PR #68323, opened 2024-09-19
Revert "[SOT] Enable dynamic shape as default in PIR mode"

SigureMo:
Reverts PaddlePaddle/Paddle#68302


SigureMo:
#68337 已经修复，本 PR close

PR #68386, opened 2024-09-23
[PIR] fix pir disable case

wanghuancoder:
### PR Category

Execute Infrastructure

### PR Types

Others 

### Description

fix pir disable case
Pcard-67164


zhwesky2010 (approved):
lgtm for change UT name

PR #68388, opened 2024-09-23
[XPU] Fix depthwise conv2d unit test

skywalker2012:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

Aim to fix bad grad check method introduced from https://github.com/PaddlePaddle/Paddle/pull/67777,
the original test class inherit from `XPUTestConv2DOp` and `XPUTestConv2DOp_v2`, they already have grad check test, so we do not need another one.


houj04 (approved):

zhangyk0314 (approved):
LGTM

PR #68449, opened 2024-09-25
[OneDNN] Fix compiler error in clang build

zhanglirong1999:
### PR Category
Environment Adaptation

### PR Types
Bug fixes

### Description
Fix compiler error in clang build, not happen in gcc compiler


LLee233 (approved):
LGTM

PR #68455, opened 2024-09-25
fix the bug in convert tensor to dist.

winter-wang:
### PR Category

Auto Parallel 

### PR Types

Bug fixes

### Description

fix the bug in convert tensor to dist.

###Other
Pcard-67164


liym27 (approved):
LGTM

PR #68460, opened 2024-09-25
[AutoParallel][PIR] remove logic about updating params_grads with sharded grads

AndSonder:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

remove logic about updating params_grads with sharded grads

this logic will cause error as bellow:

image

Pcard-76459


JZ-LIANG (approved):
LGTM

PR #68486, opened 2024-09-27
[Typing] change typing in `paddle/audio/functional/window.py`

enkilee:
### PR Category

User Experience

### PR Types

Others

### Description

in `paddle/audio/functional/window.py`
the function
```python
def _general_cosine(
    M: int, a: float, sym: bool = True, dtype: str = 'float64'
) -> Tensor:
```
`a: float` perhaps change to `a: list[float]` ?


SigureMo (approved):
LGTMeow

PR #68494, opened 2024-09-27
[PIR][oneDNN] Delete extra placement pattern for fc

LLee233:
### PR Category

 Inference

### PR Types

Bug fixes

### Description

Delete extra placement pattern for Fc, since former pass `fc_onednn_enable_pass` already filtered feasible Fc ops. Extra pattern may introduce error such what `ConvNeXt_tiny_infer` met. In the future, we will check if limits on Fc has been resolved and see if it's possible extend the scope to convert fc for performance improvement.


zhanglirong1999 (approved):
LGTM

PR #68617, opened 2024-10-10
【CINN】Add IndexExpr

liuruyan:
### PR Category

CINN

### PR Types

New features

### Description

Add IndexExpr，rewrite operator +,-,*,/,%. Simplification can be done in real time without calling AutoSimplify. only const fold implemented now.

Pcard-67164


test/cpp/pir/cinn/adt/Index_expr_test.cc
zyfncg, line 1: 文件名开头小写

zyfncg (approved):

PR #68681, opened 2024-10-14
【Hackathon 7th No.20】为 Paddle 新增 Tensor.set_ -part

NKNaN:
### PR Category

User Experience

### PR Types

New features

### Description

rfc: https://github.com/PaddlePaddle/community/pull/968

### DIFF
--- a/paddle/phi/kernels/set_inplace.cc
+++ b/paddle/phi/kernels/set_inplace.cc
@@ -11,12 +11,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-#include "paddle/common/flags.h"
-#include "paddle/phi/backends/all_context.h"
 #include "paddle/phi/core/kernel_registry.h"
 #include "paddle/phi/kernels/set_inplace_kernel.h"
-
-COMMON_DECLARE_bool(use_stride_kernel);
 
 namespace phi {
 
@@ -28,18 +24,17 @@ void SetInplaceKernel(const Context& dev_ctx,
                       const std::vector<int64_t>& stride,
                       int64_t offset,
                       DenseTensor* out) {
-  if (!FLAGS_use_stride_kernel) {
-    PADDLE_THROW(common::errors::Fatal(
-        "FLAGS_use_stride_kernel is closed. Strided kernel "
-        "be called, something wrong has happened!"));
-  }
   auto meta = out->meta();
   meta.dims = DDim(dims.data(), static_cast<int>(dims.size()));
   meta.strides = DDim(stride.data(), static_cast<int>(stride.size()));
   meta.offset = offset;
-  // reset holder to nullptr
-  out->clear();
-  *out = DenseTensor{source.Holder(), meta};
+  if (x.IsSharedWith(source)) {
+    out->set_meta(meta);
+  } else {
+    // reset holder to nullptr
+    out->clear();
+    *out = DenseTensor{source.Holder(), meta};
+  }
   out->ShareInplaceVersionCounterWith(x);
 }
 
@@ -47,7 +42,7 @@ void SetInplaceKernel(const Context& dev_ctx,
 
 PD_REGISTER_KERNEL(set,
                    CPU,
-                   STRIDED,
+                   ALL_LAYOUT,
                    phi::SetInplaceKernel,
                    bool,
                    uint8_t,
@@ -65,7 +60,7 @@ PD_REGISTER_KERNEL(set,
 #if defined(PADDLE_WITH_CUDA) || defined(PADDLE_WITH_HIP)
 PD_REGISTER_KERNEL(set,
                    GPU,
-                   STRIDED,
+                   ALL_LAYOUT,
                    phi::SetInplaceKernel,
                    bool,
                    uint8_t,
--- a/paddle/phi/kernels/set_inplace_kernel.h
+++ b/paddle/phi/kernels/set_inplace_kernel.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2023 PaddlePaddle Authors. All Rights Reserved.
+// Copyright (c) 2024 PaddlePaddle Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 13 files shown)


python/paddle/tensor/creation.py
```diff
@@ -3257,3 +3257,101 @@ def geometric_(
...
+            Tensor(shape=[3], dtype=float32, place=Place(cpu), stop_gradient=True,
+            [11., 33., 55.])
+
+            >>> x = paddle.to_tensor([1., 2., 3., 4., 5.])
+            >>> x.set_(src2, offset=1)
+            >>> print(x.offset)
+            1
+
+    """
+    if in_dynamic_mode():
```
zxcd, line 3336: add data dtype check

paddle/phi/kernels/stride/set_inplace.cc
```diff
@@ -0,0 +1,48 @@
...
+  meta.offset = offset;
+  // reset holder to nullptr
+  out->clear();
+  *out = DenseTensor{source.Holder(), meta};
+  out->ShareInplaceVersionCounterWith(x);
+}
+
+}  // namespace phi
+
+PD_REGISTER_KERNEL_FOR_ALL_BACKEND_DTYPE(set, STRIDED, phi::SetInplaceKernel) {}
```
zxcd: add register for cpu

NKNaN: Does `PD_REGISTER_KERNEL_FOR_ALL_BACKEND_DTYPE` already register for cpu?

zxcd: PD_REGISTER_KERNEL is work for cpu

NKNaN: revised

python/paddle/tensor/creation.py
NKNaN: added in branch `source is not None`

python/paddle/tensor/creation.py
```diff
@@ -3257,3 +3257,125 @@ def geometric_(
...
+        if source is None:
+            source = paddle.empty([0], dtype=x.dtype)
+            shape = [0]
+            stride = [0]
+        else:
+            if not isinstance(source, (Variable, core.eager.Tensor)):
+                raise ValueError(
+                    f"Input (source) should be paddle.Tensor but received {type(source)}"
+                )
+            if source.dtype not in [
```
zxcd: suppose to use check_dtype(), also in python code, "bfloat16" should type with "uint16"

NKNaN: revised

paddle/phi/kernels/stride/set_inplace.cc
```diff
@@ -0,0 +1,80 @@
...
+                   phi::SetInplaceKernel,
+                   bool,
+                   uint8_t,
+                   int8_t,
+                   int16_t,
+                   int,
+                   int64_t,
+                   float,
+                   double,
+                   phi::dtype::complex<float>,
```
zxcd: is CPU unspport float16 and bfloat16？

NKNaN: I don't know whether all kind of CPU support `float16` and `bfloat16` so they are not registered here

zxcd: I think you can try the same way with paddle/phi/kernels/strided_slice_kernel.cc

NKNaN: OK, `float16` and `bfloat16` are added.

zxcd: `STRIDED` may change to `ALL_LAYOUT` ？

NKNaN: revised

zxcd (approved):
LGTM

paddle/phi/kernels/set_inplace.cc
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2024 PaddlePaddle Authors. All Rights Reserved.
```
jeff41404: `inplace` is a concept in the level of `ops`(in the `ops.yaml`), representing the space where ops's output can write to the space of input. so in the level of `kernel`, there is no concept of `inplace`, because both `inplace ops` and `non-inplace ops` use  the same computational logic and the same kernel.
so, we suggest filename is `set_kernel.cc`

paddle/phi/kernels/set_inplace_kernel.h
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2024 PaddlePaddle Authors. All Rights Reserved.
```
jeff41404: `inplace` is a concept in the level of `ops`(in the `ops.yaml`), representing the space where ops's output can write to the space of input. so in the level of `kernel`, there is no concept of `inplace`, because both `inplace ops` and `non-inplace ops` use  the same computational logic and the same kernel.
so, we suggest filename is `set_kernel.h`

python/paddle/__init__.py
```diff
@@ -1173,4 +1174,5 @@
     'diagonal_scatter',
     'combinations',
     'signbit',
+    'set_',
```
jeff41404: Do we need to support the usage of `paddle.set_(x, source, ....)`? If it is not recommended for users to use it this way(recommended only `x.set_(source, ...)` when x is `Tensor`), there is no need to add it here.

zxcd: This time only support `Tensor.set_` , this line can be deleted.

paddle/phi/kernels/set_inplace.cc
NKNaN: revised

paddle/phi/kernels/set_inplace_kernel.h
NKNaN: revised

python/paddle/__init__.py
NKNaN: deleted

zxcd (approved):

jeff41404 (approved):
LGTM

sunzhongkai588 (approved):
LGTM for docs

luotao1:
可以提交中文文档啦

PR #68689, opened 2024-10-14
[XPU] Modify the lash_attn and grad operator’s bf16 format to approximate round-to-nearest-even.

ZibinGuo:
### PR Category
Custom Device

### PR Types
Others

### Description
Modify the lash_attn and grad operator’s bf16 format to approximate round-to-nearest-even.


lj970926 (approved):
LGTM

PR #68696, opened 2024-10-14
Support densemask with causal in flash attention

umiswing:
### PR Category

Performance Optimization

### PR Types

Performance

### Description

Pcard-70459
Support densemask with causal in flash attention.


GuoxiaWang (approved):
LGTM

PR #68716, opened 2024-10-15
[PIR][oneDNN] Fix bf16 placement for Reshape

LLee233:
### PR Category

Performance Optimization

### PR Types

Others

### Description

During Model perf benchmark, we found Reshape didn't go into BF16 kernel. Hence I optimize the logic to enable it and other former impacted ops

### DIFF
--- a/paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
@@ -98,7 +98,7 @@ class OneDNNBf16PlacementPattern : public pir::RewritePattern {
     }
 
     const std::vector<std::string> permitted_input_names = {
-        "x", "y", "input", "residual_param"};
+        "x", "y", "input", "residual_param", "residual_data"};
     auto op_name = op->name();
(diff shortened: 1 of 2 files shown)


paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
```diff
@@ -97,9 +97,29 @@ class OneDNNBf16PlacementPattern : public pir::RewritePattern {
       }
     }
 
-    int i = 0;
-    for (auto& value : op->operands_source()) {
-      pir::Type type = op->operand_type(i++);
+    const std::vector<std::string> permitted_input_names = {
+        "x", "y", "input", "residual_param"};
```
zhanglirong1999: Also need residual_data. For fused_matmul, it names residual_data as residual_param.

LLee233: Added. By the way, don't you think we could unify this name instead? Not sure where does the difference come from though...

zhanglirong1999 (approved):
LGTM

PR #68783, opened 2024-10-17
add pd_op.min converter for PIR TRT and base ReduceCommonOpPattern

zhink:
### PR Category

Inference

### PR Types

New features

### Description

pcard-71500
add pd_op.min converter for PIR TRT and Add ReduceCommonOpPattern base class to simplify code

### DIFF
--- a/python/paddle/tensorrt/impls/math.py
+++ b/python/paddle/tensorrt/impls/math.py
@@ -118,3 +118,28 @@ def multiply_converter(network, paddle_op, inputs):
     return add_elementwise_layer(
         network, paddle_op, inputs, trt.ElementWiseOperation.PROD
     )
+
+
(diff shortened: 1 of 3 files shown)


python/paddle/tensorrt/impls/math.py
```diff
@@ -118,3 +118,28 @@ def multiply_converter(network, paddle_op, inputs):
     return add_elementwise_layer(
         network, paddle_op, inputs, trt.ElementWiseOperation.PROD
     )
+
+
+@converter_registry.register("pd_op.min", trt_version="8.x")
```
vivienfanghuagood, line 123: reduce op的不同类型, 可以写到一起

zhink: 和sum any一块优化下

lizexu123 (approved):

PR #68784, opened 2024-10-17
[OneDNN][PIR] Fix picodet performance drop in bf16

zhanglirong1999:
### PR Category
Performance Optimization

### PR Types
Performance

### Description
Scale Op with bias skip bfloat16, since scale with bias add quant/dequant will make result different and lead MultiClassNMSKernel   execution time increase 40 times


LLee233 (approved):
In future, maybe we can think out of a way to convert it considering bias.

PR #68790, opened 2024-10-17
fix setitem grad bug

zoooo0820:
### PR Category

Execute Infrastructure

### PR Types

Bug fixes

### Description

This PR is similar to https://github.com/PaddlePaddle/Paddle/pull/68520 ;  For advanced / combined setitem, if input Tensor's `stop_gradient=True`, the values grad will be lossed.

**Advanced setitem** case has been fixed at https://github.com/PaddlePaddle/Paddle/pull/65881
This PR is to
1. fix **Combined setitem** mentioned above
2. add unittests  for **input.stop_gradient=True/False** in basic/advanced/combined setitem .

Pcard-67164


wanghuancoder (approved):
LGTM

PR #68855, opened 2024-10-21
[Auto Parallel] Add the CheckAndUpdateUnknownGlobalShape stage in dis…

zhangyuqin1998:
…t_api_gen to avoid errors caused by unfilled global shape

### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

In op such as `nonzero`, the output shape is set to -1 by the infer_meta_fn. This shape is then determined within the kernel. Therefore, we need to inspect and backfill the global shape after the kernel call.

In the newly added `CheckAndUpdateUnknownGlobalShape ` function, we check whether the global_shape contains -1. If it does, we replace it with the corresponding value from local_shape.

### DIFF
--- a/paddle/phi/api/generator/dist_api_gen.py
+++ b/paddle/phi/api/generator/dist_api_gen.py
@@ -64,7 +64,7 @@ MAIN_DIST_BRANCH_TEMPLATE = """
       // 10. Fallback{}
     }}\n
     // 11. Set Output Dist Attr For Default Impl{}\n
-    // 12. Check And Backfill GlobalShape if needed{}\n
+    // 12. Check And Update Unknown GlobalShape if needed{}\n
     // 13. Return
     {}
   }}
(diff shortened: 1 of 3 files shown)


paddle/phi/api/generator/dist_api_gen.py
```diff
@@ -464,6 +465,15 @@
     }}
 """
 
+
+SINGLE_CHECK_AND_BACKFILL_GLOBAL_SHAPE_TEMPLATE = """
+    CheckAndSBackfillGlobalShape({});"""
```
zhiqiu: CheckAndUpdateUnknownGlobalShape ?

paddle/phi/api/generator/dist_api_gen.py
```diff
@@ -1898,6 +1925,11 @@ def generate_auto_parallel_branch(self) -> str:
         kernel_call_code = self.generate_kernel_call_code()
         fallback_code = self.generate_fallback_code()
         output_dist_attr_setting = self.generate_output_dist_attr_setting()
+        # In some cases like `nonzero`, the output shape is determined after kernel call.
+        # Therefore, we need to check and backfill global shape after kernel call.
+        check_and_backfill_global_shape_code = (
```
zhiqiu: generate only for the op that contains -1 shape.

paddle/phi/api/generator/dist_api_gen.py
zhangyuqin1998: > CheckAndUpdateUnknownGlobalShape ?

done~

paddle/phi/api/generator/dist_api_gen.py
zhangyuqin1998: > generate only for the op that contains -1 shape.

done~

zhiqiu (dismissed):
LGTM

PR #68893, opened 2024-10-23
Add unitest for tensor.mT method in dynamic shape case

HydrogenSulfate:
### PR Category

User Experience

### PR Types

Not User Facing

### Description

Pcard-75624

补充`mT`方法在动态shape case下的单测


SigureMo (approved):
LGTMeow

PR #68903, opened 2024-10-23
[PIR][oneDNN] Workaround for reshape when shape is unknown

LLee233:
### PR Category

Performance Optimization

### PR Types

Bug fixes

### Description

Usually the shape of Reshape should be IntArray, while sometimes it will be DenseTensor/VectorType, which may trigger error of incompatible shape since the shape will be [-1, -1, ..., -1]. Currently we skip such case to avoid it.


zhanglirong1999 (approved):
LGTM

PR #68937, opened 2024-10-24
[CINN] Fix bugs of calculation errors in grid reduce

lshpku:
### PR Category
CINN

### PR Types
Bug fixes

### Description
Fix two bugs that cause calculation errors in grid reduce:
1. For RS layout, if the spatial dim was not a multiple of 32, grid reduce would visit wrong index in the `rf` tensor.
2. For cases where a grid reduce had multiple downstreams, the downstream op might not be executed because of conflicting `if` conditions.

Pcard-85711


paddle/cinn/optim/replace_cross_block_reduction.cc
```diff
@@ -350,16 +360,22 @@ struct CrossBlockReductionReplacer : public BaseMutator {
...
+      load_node->indices[0] = ir::Expr(0);
+      return load_node->index();
+    }();
+    ir::Expr spatial_size = [&]() {
+      load_node->indices[0] = ir::Expr(1);
+      for (int i = 1; i < load_node->indices.size(); i++) {
+        load_node->indices[i] = ir::Expr(0);
+      }
+      return load_node->index();
+    }();
```
zyfncg, line 375: 将load_node的indices都设置为0是什么意思呢？

zyfncg (approved):

paddle/cinn/optim/replace_cross_block_reduction.cc
lshpku: 就是上面那个注释写的，把`indices[0]`设成0，就是忽略最高维的index，后面剩下的就是`spatial_index`；把`indices[0]`设成1，其余设成0，可以计算除了第一维的size，就是`spatial_size`；这样做是因为`load->index()`的计算含有一些特殊处理，不是简单相乘和相加，所以我就不重复写计算的逻辑了，而是在外面通过改indices来计算

PR #68962, opened 2024-10-25
[AutoParallel] Refine chunk_id setting method for send op

zhangbo9674:
### PR Category

Auto Parallel 

### PR Types

Performance

### Description

Refine chunk_id setting method for send op
Pcard-73145


From00 (approved):
LGTM

PR #68986, opened 2024-10-27
【CINN】Enrich ability of index expr

liuruyan:
### PR Category

CINN

### PR Types

New features

### Description

1. Upgrade `ProveDivisible` without `AutoSimplify` and `analyzer`.
2. Enrich the ability of `IndexExpr`, constant must be at the `rhs` operand in `Add` and `Mul`. e.g.
```
3 + S0 ===> S0 + 3 
(S0 + 2) + S1 ===> S0 + S1 + 2
```
3. Constant should be fold  in `Add` and `Mul` now . e.g.
```
2 + 3 ===> 5
(S0 + 2) + 3 ===> S0 + 5
```
4. Implemented the associative law of `Add` for constants. e.g.
```
S0 * 3 + S0 * 4 ===> S0 * 7
```
5. Implemented `IsSumPartialBySymbol` func for `Add` and `IsDivisiblieBySymbol` func for `Div`, It can judge whether the expression can be simplified. if return true, then will apply `SimplifySymbolicAdd` and `SimplifySymbolicDivide` to simplify.
6. Implement `Normalize` func for `IndexExpr`, Its function is to explicitly reconstruct an `IndexExpr` and simplify it at the same time. It usually use when an unsimplified `Expr` is converted to an `IndexExpr` using `as_index()`, e.g.
```
Expr e = (S0 + 2) + 3
e.as_index() ===> (S0 + 2) + 3
e.as_index().Normalize() ===> S0 + 5
```
Pcard-67164


paddle/cinn/common/ir_util.h
```diff
@@ -160,5 +160,55 @@ Expr FoldExpr(FuncOp func_op, const std::vector<Expr> &values) {
...
+
+// If true is returned, the operation will be attempted on each subpart in
+// outter `simplify` function. Note: this func dont deal the corner case, e.g.
+// `IsDivisiblieBySymbol(f % S0  - f, S0)` is `false`. please use
+// `ProveDivisible` for exact result.
+bool IsDivisiblieBySymbol(const ir::IndexExpr &expr,
+                          const ir::IndexExpr &symbol,
+                          const ir::IrNodeTy &ty);
+
+bool ProveDivisible(const ir::IndexExpr &lhs, const ir::IndexExpr &rhs);
```
zyfncg, line 211: 这几个接口再补充一些注释说明，能给出示例的尽量用例子说明一下

paddle/cinn/common/ir_util.cc
```diff
@@ -501,5 +502,140 @@ Expr min(Expr a, Expr b) {
   return ir::Min::Make(a, b);
 }
 
+int32_t CalculateExprComplexity(const Expr &expr, int count) {
```
zyfncg, line 505: 这里用Complexity是否准确？看着只是做节点数量的统计

paddle/cinn/common/ir_util.cc
```diff
@@ -501,5 +502,140 @@ Expr min(Expr a, Expr b) {
...
+    case ir::IrNodeTy::Div:
+    case ir::IrNodeTy::Mod: {
+      int lhs_count = CalculateExprComplexity(expr->operand(0), count);
+      int rhs_count = CalculateExprComplexity(expr->operand(1), count);
+      return lhs_count + rhs_count + 1;
+    }
+  }
+}
+
+bool IsCorrectPriority(const Expr &lhs, const Expr &rhs) {
```
zyfncg, line 521: 这个看上去是做两个Expr的比较，CorrectPriority在语义上看不太出来返回true是左边Priority高还是右边Priority高

paddle/cinn/common/ir_util.h
liuruyan: 下一个PR中进行添加

paddle/cinn/common/ir_util.cc
liuruyan: 下一个PR中进行更改

paddle/cinn/common/ir_util.cc
liuruyan: 下一个PR中进行更改

zyfncg (approved):

PR #69009, opened 2024-10-28
[AutoParallel] Decompose Reshard Pass

JZ-LIANG:
### PR Category

Auto Parallel

### PR Types

Performance

### Description

Pcard-67164

Before this PR, the reshard op in program might compose multiple "atomic" operations, which make it difficult for Reusing reshard in later fold_reshard_pass.

In this PR, we decompose the reshard into multiple atomic operations, which increase the opportunity of re-using.

Before PR | image
-- | --
 After PR  |  image


zhiqiu (approved):
LGTM

PR #69046, opened 2024-10-29
[bug fix] remove lock in AuroGrowth FreeImpl

vivienfanghuagood:
### PR Category

Inference

### PR Types

Bug fixes

### Description

pcard-71500

When calling the AutoGrowth Allocator, distributed training will be stuck due to the lock in Release, although the cause is unknown. Therefore, the lock has been temporarily removed. 
TODO: In the future, we will recover it after finding out the real cause of the crash.


yuanlehome (approved):
LGTM

PR #69052, opened 2024-10-30
[AutoParallel] support pir job scheduler

AndSonder:
### PR Category

Auto Parallel

### PR Types

Others

### Description

support pir job scheduler
Pcard-76459


From00 (approved):
LGTM

PR #69059, opened 2024-10-30
shard accumulators eagerly to avoid OOM

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Improvements

### Description

shard accumulators eagerly to avoid OOM
Pcard-76459


JZ-LIANG (approved):
LGTM

PR #69135, opened 2024-11-03
fix: doc: highlight with python-console when needed

sblondon:
### PR Category
[ User Experience ]

### PR Types
[ Docs ]

### Description

The [`python-console`](https://pygments.org/docs/lexers/#pygments.lexers.python.PythonConsoleLexer) (or `pycon`) should be used when displaying interpreter instead of [`python`](https://pygments.org/docs/lexers/#pygments.lexers.python.PythonLexer). `python` is for source code.
By fixing it, the highlighting in the documentation would be better.

The PR is huge but it's mainly the same change every lines. There are still few `python` highligthing in the documentation because they are really source code.


SigureMo:
Thank you for correcting our wrong usage of `python`!

Since our example code is not only rendered in the English document, but also extracts the code snippet by regular expression matching and generates the Chinese document to render, it is not enough to modify these areas, but also the script used for the extraction.

We will first modify the scripts before considering merging this PR, which may take some time.

PR #69281, opened 2024-11-11
[CINN] Add batch_norm fusion kernels num check

huangjiyi:
### PR Category

CINN

### PR Types

Others

### Description

Pcard-76996
- Add batch_norm fusion kernels num check


SigureMo (approved):
LGTMeow

PR #69288, opened 2024-11-11
[Auto Parallel] Sequence parallel api

FeixLiu:
### PR Category
Auto Parallel

### PR Types
New features

### Description
Pcard-73145
Sequence parallel api

### DIFF
--- a/python/paddle/distributed/auto_parallel/intermediate/tensor_parallel.py
+++ b/python/paddle/distributed/auto_parallel/intermediate/tensor_parallel.py
@@ -156,6 +156,109 @@ class PrepareLayerOutput(PlanBase):
         layer.register_forward_post_hook(self.fn(process_mesh=process_mesh))
 
 
+def sequence_parallel_begin(process_mesh, need_transpose=True):
+    def begin(layer, input, output=None):
+        index = process_mesh.dim_names.index('mp')  # get the axis for the split
+        if isinstance(output, tuple):
+            target_output = output[0]
+        else:
+            target_output = output
+        assert isinstance(target_output, paddle.Tensor)
+        assert len(target_output.shape) == 3
+        if need_transpose:
+            target_output = paddle.transpose(target_output, perm=[1, 0, 2])
+        placements = target_output.placements
+        if placements is None:
+            placements = [
+                dist.Replicate() for _ in range(len(process_mesh.shape))
+            ]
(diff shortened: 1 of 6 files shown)


python/paddle/distributed/auto_parallel/intermediate/tensor_parallel.py
```diff
@@ -156,6 +156,139 @@ def apply(self, layer, process_mesh, shard_weight=None, shard_bias=None):
...
+    This plan marks the beginning of the sp and should be added to the LAST layer before the sp range.
+    DON'T mark any layer in the sp range.
+    """
+
+    def __init__(self, need_transpose=True):
+        super().__init__()
+        self.need_transpose = need_transpose
+
+    def sequence_parallel_begin(self, process_mesh):
+        def begin(layer, input, output=None):
```
jeff41404, line 223: because `sp_split` need output, the default value of output should not be None?

test/auto_parallel/hybrid_strategy/parallel_api.py
```diff
@@ -244,34 +236,46 @@ def parallel_model(self, layer, optimizer=None):
...
+                            SequenceParallelBegin(),
+                        ],
+                        "llama.layers.*.self_attn.q_proj": ColWiseParallel(),
+                        "llama.layers.*.self_attn.k_proj": ColWiseParallel(),
+                        "llama.layers.*.self_attn.v_proj": ColWiseParallel(),
+                        "llama.layers.*.self_attn.o_proj": RowWiseParallel(),
+                        "llama.layers.*.self_attn": [
+                            SequenceParallelEnd(),
+                            SequenceParallelBegin(),
+                        ],
```
jeff41404, line 268: This writing style may not be easy to understand, we need to think about whether there is a more intuitive way to write it

python/paddle/distributed/auto_parallel/intermediate/tensor_parallel.py
FeixLiu: ok, update in next pr

jeff41404 (approved):

test/auto_parallel/hybrid_strategy/parallel_api.py
FeixLiu: Will discuss this later, if find a better way, will update this.

PR #69316, opened 2024-11-12
[Auto Parallel] API support 3d parallel

FeixLiu:
### PR Category
Auto Parallel

### PR Types
New features

### Description
Pcard-73145
API support 3d parallel

### DIFF
--- a/python/paddle/distributed/auto_parallel/intermediate/pipeline_parallel.py
+++ b/python/paddle/distributed/auto_parallel/intermediate/pipeline_parallel.py
@@ -85,6 +85,23 @@ class PipelineParallel(ParallelModel):
             return input
 
         split_layer_names = list(self.split_spec.keys())
+        pipeline_stage_index = 0
+        meet_split_layer = False
+        for layer_name, layer in model.named_sublayers():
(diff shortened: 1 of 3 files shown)


deepllz (approved):
LGTM

python/paddle/distributed/auto_parallel/intermediate/pipeline_parallel.py
```diff
@@ -85,6 +85,23 @@ def forward_pre_hook(layer, input):
...
+
+            if meet_split_layer is False:
+                layer.pipeline_stage_index = pipeline_stage_index
+            else:
+                if split_layer_names[pipeline_stage_index] in layer_name:
+                    layer.pipeline_stage_index = pipeline_stage_index
+                else:
+                    pipeline_stage_index += 1
+                    layer.pipeline_stage_index = pipeline_stage_index
+                    meet_split_layer = False
```
jeff41404, line 103: If we enter this branch, does it mean that there is a problem with user configured `split_spec`, that is, the order of key in `split_spec` is not consistent with the actual order of `Layer` in the model, and `pipeline_stage_index += 1` may not be correct? It would be better to report an error?

PR #69357, opened 2024-11-13
support if op for auto parallel.

winter-wang:
### PR Category

Auto Parallel 

### PR Types

New features

### Description

- fix the bug caused by sharding stage1 tensor fusion.
    - while the amp closed, the llama-7b model will throw exception.
15a0b74d7e168fe8ddc980388828aec8
    - while the gradients is not partial in dp dim, the sharding will  throw exception.
- support if op for auto parallel.
    - add inferspmd logical code for  if op.
    - refactor CvtValueToDist function to support control block ops.

### Other
Pcard-67164


zhiqiu (approved):
LGTM

changeyoung98 (approved):
LGTM

PR #69365, opened 2024-11-13
[Auto Parallel] Support 3D parallel with intermediate API

deepllz:
### PR Category

Auto Parallel

### PR Types

Others

### Description

Pcard-73145
中层API支持3D混合并行


python/paddle/distributed/auto_parallel/intermediate/pipeline_parallel.py
```diff
@@ -85,11 +86,38 @@ def forward_pre_hook(layer, input):
...
+                        is_valid = True
+                        break
+                assert (
+                    is_valid
+                ), f"the last layer:{split_layer_name} must not be SplitPoint.END, please check the split_spec"
+            else:
+                raise NotImplementedError(
+                    "SplitPoint.BEGINNING is not supported currently"
+                )
+                pipline_layer_mark[index] = 1
```
jeff41404, line 111: There is currently no problem, but we will support `SplitPoint.BEGIN` in the future, it may be necessary to check whether it has already been marked (if it has been marked, report an error?). For example, split_spec is written as follows: `split_spec={"layer.0" : SplitPoint.END, "layer.1" : SplitPoint.BEGIN}`, in first loop, layer.0 will be marked on layer.1, and then in second loop, it will be marked again by layer.1

deepllz: According to the current logic, there will be no problem with repeated markings, because the marking method is to assign a value of 1.

FeixLiu (approved):

PR #69560, opened 2024-11-20
fix bug for tensor fusion in ernie.

winter-wang:
### PR Category

Auto Parallel 
### PR Types

Bug fixes
### Description

fix bug for tensor fusion in ernie.
### Other
Pcard-67164


liym27 (approved):
LGTM

PR #69608, opened 2024-11-21
[CINN] Replace unit vars to zero before resizing buffer

lshpku:
### PR Category
CINN

### PR Types
Bug fixes

### Description
In `OptimizeExprGPU`, move the `ReplaceUnitVarToZero` pass before the `EliminateCommonFactorOfLocalIndex` pass. Otherwise, we may fail to eliminate some unit variables (variables that are in range `[0, 1)`) in local buffer's indices when there is transpose in the kernel.

Pcard-85711

### DIFF
--- a/paddle/cinn/optim/transform_gpu_forloop.cc
+++ b/paddle/cinn/optim/transform_gpu_forloop.cc
@@ -400,7 +400,7 @@ class LocalAxisVisitor : public ir::IRMutator<> {
                                              "threadIdx.z"};
 };
 
-class ReplaceVarToZero : public ir::IRMutator<> {
+class ReplaceUnitVarToZero : public ir::IRMutator<> {
  public:
   void operator()(ir::Expr *expr) { ir::IRMutator<>::Visit(expr, expr); }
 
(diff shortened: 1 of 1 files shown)


zyfncg (approved):

paddle/cinn/optim/transform_gpu_forloop.cc
```diff
@@ -457,36 +457,41 @@ class ReplaceVarToZero : public ir::IRMutator<> {
...
+  // Replace blockIdx/threadIdx in local buffer's indices to zero, because local
+  // buffers cannot be accessed from another block/thread.
   LocalAxisVisitor local_axis_visitor;
   local_axis_visitor(expr);
 
+  // Replace variables that are in range [0, 1) to zero.
+  ReplaceUnitVarToZero replace_unit_var_to_zero;
+  replace_unit_var_to_zero(expr);
+
   EliminateCommonFactorOfLocalIndex(expr);
```
zyfncg, line 491: 这两个pass间有比较强的顺序要求，这里怎么用注释解释下？

lshpku: 我正在梳理OptimizeExprGPU里面pass的顺序，有些pass应该是可以删掉的，比如UpdateBufferAxisPass，我正在一点一点试，这里没加注释是因为我还没完全搞懂它做了哪些事情，等梳理好了会加上的

PR #69769, opened 2024-11-27
[CINN]fix FullWithTensor pd_to_cinn

Hongqing-work:
### PR Category

CINN

### PR Types

Bug fixes

### Description

Pcard-67164
This PR fixes FullWithTensor pd_to_cinn bug.
When the output of FullWithTensor is [], it should not be replaced to expand.


SigureMo (approved):
LGTMeow

PR #69785, opened 2024-11-28
Fix compilation issue with gcc13 on cpu

0x3878f:
### PR Category

Execute Infrastructure

### PR Types

Others

### Description

OS: ubuntu20.04
GCC: 13.1.0
fix compilation issue with gcc13 on cpu
add one patch to third_party/brpc
pcard-67164


risemeup1 (approved):
LGTM

PR #69790, opened 2024-11-28
fix the bug because of PR69357

winter-wang:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

fix the bug because of PR69357

### Other
Pcard-67164


zhiqiu (approved):
LGTM

PR #69804, opened 2024-11-28
[CINN] Optimize Recompute analyze Time

chen2016013:
### PR Category

CINN

### PR Types

Bug fixes

### Description

pcard-67164
Optimize Recompute analyze Time
double for loop -> use map to record


zhangbo9674 (approved):
LGTM

PR #69831, opened 2024-11-29
[DCU] warpctc support gfx926 and gfx928

lifulll:
### PR Category

Custom Device

### PR Types

Improvements

### Description

warpctc support gfx926 and gfx928


lifulll:
Cls

PR #69882, opened 2024-12-02
Update broadcast_input_data for mp

DesmonDay:
### PR Category
Performance Optimization

### PR Types
Improvements

### Description
Pcard-88294
Current `broadcast_input_data` function can only handle dict of tensors or tuple of tensors, but cannot handle nested structure. This PR fix this problem.


ForFishes (approved):
LGTM

PR #69886, opened 2024-12-02
[CINN] Optimize recompute algorithm runtime

chen2016013:
### PR Category

CINN

### PR Types

Bug fixes

### Description

pcard-67164
Optimize recompute algorithm runtime

PR #69897, opened 2024-12-03
[CINN] Force full_op to be recomputed

chen2016013:
### PR Category

CINN

### PR Types

Bug fixes

### Description

pcard-67164
Rename function name by comments from https://github.com/PaddlePaddle/Paddle/pull/69886

Force full_op & full_int_array_op to be recomputed because they are more likely to save memory and get fused.


zhangbo9674 (approved):
LGTM

PR #69907, opened 2024-12-03
[hackathon 7th] PaddleSpeech line:  fix  when shape=() can not load 

enkilee:
### PR Category

Performance Optimization

### PR Types

Improvements

### Description

in file `python/paddle/static/quantization/utils.py`
function `load_variable_data`
when var_tensor = np.ndarray and var_tensor's shape = () 
the code  `var_tensor.shape[0]` occurs error.
so add :
```python
    result = np.array(var_node.get_tensor())
    if result.shape == ():
        return result.reshape(1)
    else:
        return result
```


zxcd (approved):
LGTM

luotao1 (approved):

PR #69918, opened 2024-12-03
[AutoParallel] Fix empty shape in einsum

zhiqiu:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

The shape of 'inner_cache' ouput in 'einsum' op is not infered, so it is empty. 
It result in error in auto_parallel, this PR fixes it by using local shape in replicated case.
Pcard-76459


From00 (approved):
LGTM

PR #69947, opened 2024-12-04
fix the bug for create_shaped_type fucntion.

winter-wang:
### PR Category

Auto Parallel
### PR Types

Bug fixes

### Description

- fix the bug for create_shaped_type fucntion.
 The shapes' elements of "create_shaped_type" maybe exceeds of int32_t, so rectify its type to int64_t。
### Other
Pcard-67164


liym27 (approved):
LGTM

PR #70022, opened 2024-12-06
[CherryPick] fix bug of FA2 densemask when casual=True

GuoxiaWang:
### PR Category

Performance Optimization

### PR Types

Bug fixes

### Description

Pcard-70459

修复 FA2 densemask 在开启 casual=True 时的 OOB bug，在 bad case 上会触发。https://github.com/PaddlePaddle/flash-attention/pull/60


ForFishes (approved):
LGTM

PR #70097, opened 2024-12-10
Integrate Flux

umiswing:


paddle/phi/infermeta/binary.cc
```diff
@@ -4197,6 +4197,88 @@ void FakeQuantizeLSQInferMeta(const MetaTensor& x,
...
+          "transpose_weight: %d.",
+          input.dims(),
+          weight.dims(),
+          transpose_weight));
+
+  if (check_can_implement) {
+    output->set_dims(common::make_ddim({}));
+    output->set_dtype(DataType::BOOL);
+  } else {
+    output->set_dtype(input.dtype());
```
sneaxiy, line 4278: Why not invoke `output->set_dims(...)` here?

paddle/phi/kernels/gpu/all_gather_gemm_kernel.cu
```diff
@@ -0,0 +1,442 @@
...
+  int32_t nnodes;
+  int32_t full_m;
+  int32_t n_dim;
+  int32_t k_dim;
+  const phi::DataType input_dtype =
+      std::is_same<InT, phi::dtype::float16>::value ? phi::DataType::FLOAT16
+                                                    : phi::DataType::BFLOAT16;
+  const phi::DataType output_dtype =
+      std::is_same<OutT, phi::dtype::float16>::value ? phi::DataType::FLOAT16
+                                                     : phi::DataType::BFLOAT16;
```
sneaxiy, line 59: 1. Is it supported when `InT` is not float16 or bfloat16?
2. Is it supported when `OutT` is not float16 or bfloat16?
3. Is it supported when `InT` is not the same with `OutT`?

If not, we should add `static_assert` here.

paddle/phi/kernels/gpu/all_gather_gemm_kernel.cu
```diff
@@ -0,0 +1,442 @@
...
+  static constexpr int MAX_NUM_SIGNAL = 64;
+
+  using FlagType = int32_t;
+  const phi::GPUContext& dev_ctx;
+  distributed::NCCLCommContext* comm_ctx;
+  paddle::distributed::ProcessGroup* tp_group;
+  int32_t nnodes;
+  int32_t full_m;
+  int32_t n_dim;
+  int32_t k_dim;
```
sneaxiy, line 53: Note that when `full_m`, `n_dim` or `k_dim` is larger than `2^31-1`, the operator may crash.

paddle/phi/kernels/gpu/all_gather_gemm_kernel.cu
```diff
@@ -0,0 +1,442 @@
...
+  AGGemmHelper(const phi::GPUContext& dev_ctx_,
+               paddle::distributed::ProcessGroup* tp_group_,
+               distributed::NCCLCommContext* comm_ctx_,
+               int32_t nnodes,
+               int32_t full_m,
+               int32_t n_dim,
+               int32_t k_dim,
+               bool transpose_weight = true,
+               AGRingMode ring_mode_ = AGRingMode::Auto)
+      : dev_ctx(dev_ctx_),
```
sneaxiy, line 103: In fact, according to Google coding style, the data member should be named like `dev_ctx_`, and the other member should be named like `dev_ctx`.

paddle/phi/kernels/gpu/all_gather_gemm_kernel.cu
```diff
@@ -0,0 +1,442 @@
...
+    this->ring_mode = get_ring_mode(ring_mode_);
+
+    // copy stream
+    this->num_cp_streams = 1;
+    for (int i = 0; i < this->num_cp_streams; ++i) {
+      this->cp_streams.push_back(this->comm_ctx->GetStream());
+    }
+    // create events
+    constexpr bool disable_timing = true;
+    static CUDAEventHolder cp_event_holder{disable_timing};
```
sneaxiy, line 140: Not a best practice to init the static object in this way. Same below.

paddle/phi/kernels/gpu/all_gather_gemm_kernel.cu
```diff
@@ -0,0 +1,442 @@
...
+      }
+    }
+  }
+
+  PADDLE_ENFORCE_GPU_SUCCESS(
+      cudaEventRecord(helper.cp_event, helper.cp_streams[0]));
+
+  ag_gemm();
+
+  PADDLE_ENFORCE(output_buffer.meta().is_contiguous(),
```
sneaxiy, line 376: `PADDLE_ENFORCE` -> `PADDLE_ENFORCE_EQ(..., true)`

paddle/phi/kernels/gpu/all_gather_gemm_kernel.cu
umiswing: 1 and 2: No, only bfloat16/float16 and fp8 are supported.

3: There is no constraint in the source code, but it's never tested when `InT` is not the same with `OutT`

paddle/phi/infermeta/binary.cc
umiswing: Setting output's dims in infermeta require `world_size` passed as parameter. I pass `ring_id` as kernel's parameter and get `world_size` with `ProcessGroup` inside the kernel

sneaxiy (approved):
Should refine the codes asap.

PR #70099, opened 2024-12-10
[PIR][OneDNN] Fix bf16 quantize data_format bug

zhanglirong1999:
### PR Category
Inference

### PR Types
Bug fixes

### Description
Fix bf16 quantize data_format. We need use op data_format as quantized op data_format if this op having data_format attribute.


LLee233 (approved):
LGTM

PR #70108, opened 2024-12-10
[XPU] disable use_stride_kernel for XPU

lj970926:
### PR Category

Custom Device

### PR Types

Performance

### Description

Because the use_stride_kernel=True on XPU would cause significant performance degradation, we disable it by default.


tizhou86 (approved):
LGTM

houj04 (approved):

PR #70111, opened 2024-12-10
【Comm】Fix comm_ctx bug when calling barrier op in static mode

Xing-lil:
### PR Category

Communication Library

### PR Types

Bug fixes

### Description

1. Fixed the comm_ctx bug when calling dynamic op in static mode.
2. Migrated the `barrier` op to phi.
3. Update PIR comm_init to use new op `all_to_all` instead of the old op `alltoall`.

Pcard-70448


liym27 (dismissed):
LGTM

zyfncg (approved):

PR #70112, opened 2024-12-10
Softmax with Cross Entropy operator migration and pass communication initialization fix

Function-Samuel:
### PR Category
Communication Library

### PR Types
Others

### Description
softmax with cross entropy算子迁移与pass的通信初始化修复
PCard-85979

### DIFF
--- a/python/paddle/distributed/passes/auto_parallel_replace_with_parallel_cross_entropy.py
+++ b/python/paddle/distributed/passes/auto_parallel_replace_with_parallel_cross_entropy.py
@@ -73,7 +73,7 @@ class AutoParallelReplaceWithParallelCrossEntropyPass(PassBase):
                     if not placement1[1].is_shard():
                         return
 
-                    process_ids = op.dist_attr.process_mesh.process_ids
+                    process_ids = operand1.dist_attr().process_mesh.process_ids
                     group = new_process_group(sorted(process_ids))
                     ring_id = group.id
                     nranks = group.nranks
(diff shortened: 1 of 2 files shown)


python/paddle/distributed/passes/auto_parallel_replace_with_parallel_cross_entropy.py
```diff
@@ -57,9 +60,13 @@ def _check_user(self, value):
     def _apply_single_impl(self, main_program, startup_program, context):
         del_ops = []
         new_ops = []
-        ring_id = self.model_parallel_group.id
-        rank = self.model_parallel_group.rank
-        nranks = self.model_parallel_group.nranks
+
+        params = main_program.global_block().all_parameters()
+        process_ids = params[0].dist_attr().process_mesh.process_ids
```
liym27: Why use `process_ids` of params[0] instead of the input of 'pd_op.cross_entropy_with_softmax'?

Function-Samuel: get it, I will modify this immediately

python/paddle/distributed/passes/auto_parallel_replace_with_parallel_cross_entropy.py
```diff
@@ -73,6 +73,12 @@ def _apply_single_impl(self, main_program, startup_program, context):
                     if not placement1[1].is_shard():
                         return
 
+                    process_ids = op.dist_attr.process_mesh.process_ids
```
liym27: Why use process_ids from `op.dist_attr.process_mesh` instead of  process_ids from 'operand1'?

Function-Samuel: My understanding of this is in paddle computation graph's value is an edge and op is a node, I have considered that the process_ids in op and operand might be different as well. So I guess the process_ids obtained from op may be suitable here.

liym27 (approved):
LGTM

zhangbo9674 (approved):
LGTM

PR #70145, opened 2024-12-11
[SCU][Docathon][Add API Legend No.53] Add legend for masked_scatter API

yangrongxinuser:
### PR Category
User Experience

### PR Types
Docs

### Description
update doc for https://github.com/PaddlePaddle/docs/pull/6979

### DIFF
--- a/python/paddle/tensor/manipulation.py
+++ b/python/paddle/tensor/manipulation.py
@@ -5090,10 +5090,12 @@ def masked_scatter(
     The `value` should have at least as many elements as the number of ones in `mask`.
 
     The image illustrates a typical case of the masked_scatter operation.
-    Value Tensor: Contains the data to be filled into the target tensor. Only the parts where the mask is True will take values from the value tensor, while the rest will be ignored.
-    Mask: Specifies which positions should extract values from the value tensor and update the target tensor. True indicates the corresponding position needs to be updated.
-    Origin Tensor: The input tensor, where only the parts satisfying the mask will be replaced, and the rest remains unchanged.
-    Operation Result: After the masked_scatter operation, the parts of the origin tensor where the mask is True are updated with the corresponding values from the value tensor, while the parts where the mask is False remain unchanged, forming the final updated tensor.
+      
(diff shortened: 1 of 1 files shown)


yangrongxinuser:
@luotao1

python/paddle/tensor/manipulation.py
```diff
@@ -5089,6 +5089,17 @@ def masked_scatter(
     each occurrence of `mask` being True. The shape of `mask` must be broadcastable with the shape of the underlying tensor.
     The `value` should have at least as many elements as the number of ones in `mask`.
 
+    The image illustrates a typical case of the masked_scatter operation.
+    Value Tensor: Contains the data to be filled into the target tensor. Only the parts where the mask is True will take values from the value tensor, while the rest will be ignored.
+    Mask: Specifies which positions should extract values from the value tensor and update the target tensor. True indicates the corresponding position needs to be updated.
+    Origin Tensor: The input tensor, where only the parts satisfying the mask will be replaced, and the rest remains unchanged.
+    Operation Result: After the masked_scatter operation, the parts of the origin tensor where the mask is True are updated with the corresponding values from the value tensor, while the parts where the mask is False remain unchanged, forming the final updated tensor.
```
sunzhongkai588:
```suggestion
    The image illustrates a typical case of the masked_scatter operation.
      
      1. Tensor  ``value``: Contains the data to be filled into the target tensor. Only the parts where the mask is True will take values from the value tensor, while the rest will be ignored;
      2. Tensor  ``mask``: Specifies which positions should extract values from the value tensor and update the target tensor. True indicates the corresponding position needs to be updated;
      3. Tensor  ``origin``: The input tensor, where only the parts satisfying the mask will be replaced, and the rest remains unchanged;

    Result: After the ``masked_scatter`` operation, the parts of the ``origin`` tensor where the ``mask`` is ``True`` are updated with the corresponding values from the ``value`` tensor, while the parts where the ``mask`` is ``False`` remain unchanged, forming the final updated tensor.
```

luotao1:
@yangrongxinuser 请在孙师傅修改基础上，修一下 Codestyle-Check 流水线

luotao1 (approved):

PR #70243, opened 2024-12-16
[Recompute] Fix recompute full_int_array bug

chen2016013:
### PR Category

CINN

### PR Types

Bug fixes

### Description

pcard-67164

Previous PR: https://github.com/PaddlePaddle/Paddle/pull/69897
We want to regard [full] or [full_int_array] and it's child op as a whole part and don't want them to be apart( e.g, recompute pd_op.sum but not recompute it's parent pd_op.full op). Now we do this strategy by enforcing edge weight to be **inf** .

### DIFF
--- a/python/paddle/decomposition/recompute.py
+++ b/python/paddle/decomposition/recompute.py
@@ -467,7 +467,6 @@ def auto_recompute(
         ):
             return 0.1
 
-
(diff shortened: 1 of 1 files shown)


zhangbo9674 (approved):
LGTM

PR #70287, opened 2024-12-17
[Dy2st]Fix while to static bug

zhangbo9674:
### PR Category

Execute Infrastructure

### PR Types

Bug fixes

### Description

Fix to static bug of while op in this case：
```python
import paddle
from paddle.jit import to_static

@to_static(full_graph=True)
def test():
    x = paddle.to_tensor(1, dtype='float32')
    y = paddle.to_tensor(False, dtype='bool')
    while y:
        x = 2
test()
```
Origin error is：
```
File "./paddle/static/nn/control_flow.py", line 952, in while_loop
        cf_yield([next_cond, *(var.next_var for var in variable_loop_vars)])
RuntimeError: Unable to cast Python instance of type <class 'int'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details) 
```
Pcard-67164

### DIFF
--- a/python/paddle/static/nn/control_flow.py
+++ b/python/paddle/static/nn/control_flow.py
@@ -911,6 +911,23 @@ def while_loop(cond, body, loop_vars, is_test=False, name=None):
                 loop_vars, next_vars
             )
 
+            from paddle.jit.dy2static.convert_operators import (
+                to_static_variable,
+            )
+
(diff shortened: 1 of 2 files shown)


SigureMo (approved):
LGTMeow

PR #70386, opened 2024-12-20
[CINN] Accelerate recompute ips

chen2016013:
### PR Category

CINN

### PR Types

Bug fixes

### Description

pcard-67164
some benchmark models show worse performance on ips after turning on Auto-ecompute Flag. One of the reasons we've found is that **SUM** op which is recomputable actually will be changed into  compute intensive **REDUCE_SUM** op and takes long time to run. Therefore we remove sum op from default recomputable ops.


python/paddle/decomposition/recompute.py
```diff
@@ -81,11 +81,12 @@
...
     "pd_op.unsqueeze",
     "pd_op.transpose",
-    "pd_op.prod",
+    # "pd_op.prod",
     "pd_op.log",
     "pd_op.log1p",
     "pd_op.logit",
-    "pd_op.max",
+    # "pd_op.max",
+    # "pd_op.min",
```
zyfncg, line 89: pd_op.any是否也要过滤掉？

python/paddle/decomposition/recompute.py
zyfncg, line 1: slice, gather, transpose这些算子是否做recompute有哪些判定条件？直观上看这些算子recompute有可能会引起显存增加或者性能下降的情况

zyfncg (approved):

python/paddle/decomposition/recompute.py
chen2016013: 没有特殊判定条件。后续在分析一下，确认有问题的话（连同下面说的any）一起ban掉

zhangbo9674 (approved):
LGTM

PR #70430, opened 2024-12-24
[PIR][oneDNN] Add quantization pattern for Concatence

LLee233:
### PR Category

Performance Optimization

### PR Types

Improvements

### Description

For ops like Conv and Concat, their inputs are sometimes packed as VectorType, hence current quantization doesn't work. Here we deal with them specifically, putting such patterns into a independent file.

### DIFF
--- a/paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "paddle/fluid/pir/transforms/onednn/cpu_bfloat16_squash_pass.h"
+#include "paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.h"
 
 #include "paddle/fluid/pir/dialect/kernel/ir/kernel_type.h"
 #include "paddle/fluid/pir/dialect/operator/ir/onednn_op.h"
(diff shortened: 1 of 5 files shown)


paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
```diff
@@ -0,0 +1,163 @@
...
+    pir::CombineOp pre_op =
+        pir::GetDefiningOpForInput(op, 0)->dyn_cast<pir::CombineOp>();
+    if (!pre_op) return false;
+    if (!pre_op.out().HasOneUse()) return false;
+
+    auto op_attributes = op->attributes();
+    auto onednn_data_type = op_attributes.at("mkldnn_data_type")
+                                .dyn_cast<pir::StrAttribute>()
+                                .AsString();
+    if (onednn_data_type == "bfloat16") return false;
```
zhanglirong1999, line 62: Also need check int8.
And btw, in cpu_bf16_placement pass, we have a lot of logic to determine whether bf16 can be used, I think we also need to add such condition for this op?

LLee233: I think it's fine for now, cause we don't have any int8 pass. This check is just in case to filter out processed ops in former bf6_quantize_pass. And for logic about determining usability of bf16, would you mind showing some examples?

zhanglirong1999: For example, only float input can be converted to bfloat16, we need to avoid int64/int32 and others use bf16 pass, and if the bfloat16 operator that is between two float32 operators, may need to revert to fp32. I'm not sure all these condition need match as cpu_bl16_placement_pass, but I think part of them should consider on this op~

LLee233: Thanks~ I've already thought about 1st condition. But for 2nd one, it will be little complex, since if there is subgraph like [f32_op+f32_op+f32_concat+f32_op], the 2nd op & concat might be impacted by each other, causing both of which unable to be quantized. What about putting this to next step? Since Fused_conv might be impacted also~

yuanlehome (approved):

PR #70461, opened 2024-12-25
【CINN】Change int64 to int32 in IndexExpr simplify

liuruyan:
### PR Category

CINN

### PR Types

Performance

### Description

Because `intimm->value` is always of type int64 rather than intimm's type, avoid using `intimm->value` and `Expr` directly.
Pcard-67164

### DIFF
--- a/paddle/cinn/ir/op/ir_operators.cc
+++ b/paddle/cinn/ir/op/ir_operators.cc
@@ -377,14 +377,16 @@ static IndexExpr SimplifyAdd(const IndexExpr &lhs, const IndexExpr &rhs) {
   auto lhsAdd = lhs.As<Add>();
   if (lhsAdd && rhsConst) {
     if (auto lrhs = lhsAdd->b().as_index().As<IntImm>()) {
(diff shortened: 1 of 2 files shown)


paddle/cinn/ir/op/ir_operators.cc
```diff
@@ -460,22 +464,25 @@ static IndexExpr SimplifyMul(const IndexExpr &lhs, const IndexExpr &rhs) {
   auto lhsMul = lhs.As<Mul>();
   if (lhsMul && rhsConst) {
     if (auto lrhs = lhsMul->b().as_index().As<IntImm>()) {
-      return lhsMul->a().as_index() * (lrhs->value * rhsConst->value);
+      return lhsMul->a().as_index() *
+             IndexExpr(lrhs->type(), lrhs->value * rhsConst->value);
```
zyfncg, line 468: lrhs->value * rhsConst->value如果超过lrhs->type()的上限会自动提升类型吗？

liuruyan: thx, 我增加下检查

zyfncg (approved):

PR #70467, opened 2024-12-25
【Comm】Fix broadcast comm_ctx in old_program

Xing-lil:
### PR Category

Communication Library

### PR Types

Bug fixes

### Description

1. In old program, when setting `comm_ctx` for `dev_ctx`, it needs to get `comm_ctx` from `ProcessGroupMapFromGid` or `CommContextManager`.
2. Add a check, if `comm_ctx` is set in `BuildOpFuncList`, then it will not be set again in `RunOperator`.

Pcard-70448

### DIFF
--- a/paddle/fluid/framework/new_executor/program_interpreter.cc
+++ b/paddle/fluid/framework/new_executor/program_interpreter.cc
@@ -1032,15 +1032,17 @@ void ProgramInterpreter::RunOperator(const Instruction& instr_node) {
                   static_cast<paddle::distributed::ProcessGroupNCCL*>(pg)
                       ->GetOrCreateCommContext(place);
             }
-            if (comm_context) {
-              dev_ctx =
-                  static_cast<phi::distributed::NCCLCommContext*>(comm_context)
-                      ->GetDevContext();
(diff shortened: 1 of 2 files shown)


paddle/fluid/framework/new_executor/program_interpreter.cc
```diff
@@ -1016,15 +1016,23 @@ void ProgramInterpreter::RunOperator(const Instruction& instr_node) {
...
+            if (comm_context_manager.Has(std::to_string(ring_id))) {
+              comm_context = comm_context_manager.Get(std::to_string(ring_id));
+            } else if (map->has(ring_id)) {
               distributed::ProcessGroup* pg = map->get(ring_id);
-              auto comm_context =
+              comm_context =
                   static_cast<paddle::distributed::ProcessGroupNCCL*>(pg)
                       ->GetOrCreateCommContext(place);
+            }
+            if (comm_context) {
```
liym27: If it's false, better to throw an exception?

liym27 (approved):
LGTM

zhangbo9674 (approved):
LGTM

PR #70492, opened 2024-12-26
【Bug fix】Fix randn bug in large shape 

Xing-lil:
### PR Category

Operator Mechanism

### PR Types

Bug fixes

### Description

- Bug: `paddle.randn` produces zeros for shapes exceeding the `int` limit.
- Solution: The randn `gaussian_kernel` follows different paths depending on the `seed` value:
For `seed == 0`, convert `int` to `int64_t` in `WriteData`.  
For `seed != 0`, convert `int` to `int64_t` in both `WriteData` and `IndexKernel`.
Add a test to check the `value`/`mean`/`std` for shape exceeds `INT_MAX`. If not fixed, the `value` will be `0`.

Pcard-70448


liym27 (approved):
LGTM

PR #70526, opened 2024-12-27
Remove mp_flag when grad norm is allreduced among sharding group

sneaxiy:
### PR Category
Distributed Strategy

### PR Types
Improvements

### Description
Remove mp_flag when grad norm is allreduced among sharding group.

Pcard-67164


ForFishes (approved):
LGTM

PR #70628, opened 2025-01-06
[OneDNN][PIR] add permitted input name filter for orphaned op

zhanglirong1999:
### PR Category
Inference

### PR Types
Performance

### Description
Add permitted input name filter for orphaned op, allow input not on the permitted name to be fp32

### DIFF
--- a/paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
+++ b/paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
@@ -280,9 +280,28 @@ class RemoveOrphanedPattern : public pir::RewritePattern {
                                              "pd_op.fetch",
                                              "pd_op.assign"});
 
+    const std::vector<std::string> permitted_input_names = {
+        "x", "y", "input", "residual_param", "residual_data"};
+    auto op_name = op->name();
+    auto op_info = pir::IrContext::Instance()->GetRegisteredOpInfo(op_name);
+    if (!op_info) return false;
+    paddle::dialect::OpYamlInfoParser yaml_parser(
+        op_info.GetInterfaceImpl<paddle::dialect::OpYamlInfoInterface>()
+            ->get_op_info_(op_name),
+        paddle::dialect::IsLegacyOp(op_name));
+    auto input_names = yaml_parser.InputNames();
+
     if (op->num_operands()) {
       for (uint32_t i = 0; i < op->num_operands(); i++) {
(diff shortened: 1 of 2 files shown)


paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
```diff
@@ -263,11 +263,30 @@ class RemoveOrphanedPattern : public pir::RewritePattern {
...
+    const std::vector<std::string> permitted_input_names = {
+        "x", "y", "input", "residual_param", "residual_data"};
+    auto op_name = op->name();
+    auto op_info = pir::IrContext::Instance()->GetRegisteredOpInfo(op_name);
+    if (!op_info) return false;
+    paddle::dialect::OpYamlInfoParser yaml_parser(
+        op_info.GetInterfaceImpl<paddle::dialect::OpYamlInfoInterface>()
+            ->get_op_info_(op_name),
+        paddle::dialect::IsLegacyOp(op_name));
+    auto input_names = yaml_parser.InputNames();
```
LLee233, line 292: What about abstracting this part into a function? Since I saw you're using it for two patterns.

paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
```diff
@@ -263,11 +263,30 @@ class RemoveOrphanedPattern : public pir::RewritePattern {
...
+        std::string input_name = input_names[i];
+        auto iter = std::find(permitted_input_names.begin(),
+                              permitted_input_names.end(),
+                              input_name);
+        if (iter == permitted_input_names.end()) {
+          // The input in permitted_input, it must be bf16, others can be fp32
+          continue;
+        }
         auto* prev_op = pir::GetDefiningOpForInput(op, i);
         // if (!prev_op) continue;
```
LLee233: Plz delete this annotated code

LLee233 (commented):
Btw, plz change the year of copyright since we're in 2025 now lol.

paddle/fluid/pir/transforms/onednn/cpu_bfloat16_placement_pass.cc
zhanglirong1999: Hi, I think it's ok here, although it is repeated, if write an additional GetInputNames() to return the input name vector, you also need to make a judgment on op_info. If it is None, the match() function directly returns false. So I think it will not reduce the code too much when abstract a function

LLee233 (approved):
LGTM

PR #70711, opened 2025-01-08
[CINN] Open Transpose op in auto-recompute

chen2016013:
### PR Category

CINN

### PR Types

Not User Facing

### Description

pcard-67164
Open Transpose op in auto-recompute
In the past we set Transpose Op as a unrecomputable op because somtimes it bring EXTRA IO cost when fused in fusion op.
However, by discuss we found this situation rarely occurs and can be ignored. Setting Transpose Op as a recomputable op can improve some model's ips up to 10%. 
So we open Transpose Op in auto-recompute.


zhangbo9674 (approved):
LGTM

PR #70736, opened 2025-01-09
[AutoParallel] Change tensor fusion switch name

AndSonder:
### PR Category

Auto Parallel

### PR Types

Others

### Description

Change tensor fusion switch name
Pcard-76459


From00 (approved):
LGTM

PR #70749, opened 2025-01-09
[Old IR][oneDNN] Fix issue for repetitive inputs

LLee233:
### PR Category

Inference

### PR Types

Bug fixes

### Description

During development, we found an issue of dim mismatch of reshape when enabling bf16 pass on old IR, which was caused by linked node check. This check aimed to avoid repeated creation of q/dq, while dismissing the situation of repetitive inputs. Here I correct the logic for it and still no need to create extra q/dq.


zhanglirong1999 (approved):
LGTM

PR #70812, opened 2025-01-13
Support print pir Block in python

zhangbo9674:
### PR Category

Execute Infrastructure

### PR Types

New features 

### Description

Support print pir Block in python
Pcard-67164


From00 (approved):
LGTM

PR #70832, opened 2025-01-14
[CINN] Accelerate recompute algorithm time

chen2016013:
### PR Category

CINN

### PR Types

Improvements

### Description

pcard-67164
Accelerate recompute algorithm time 511s --> 30s
1. Find upstream and downstream ops recursively for every uncomputable op ----> Traversal all ops twice
2. Find all parent ops for cloned ops ----> use map to record first grad user


zhangbo9674 (approved):
LGTM

PR #70850, opened 2025-01-15
Add non-reduce fp8 comm support for process group

umiswing:
### PR Category

Communication Library

### PR Types

New features

### Description

Add non-reduce fp8 comm support, and raise error when calling reduce comm with fp8


GuoxiaWang (approved):
LGTM

PR #70857, opened 2025-01-16
[PIR][OneDNN] Support Split op in bf16 (do not merge, need check ci coverage) 

zhanglirong1999:
### PR Category
Inference

### PR Types
Performance

### Description
Split op output is a vector, and not support bf16 q/dq in the past. We support Split op in bf16 now. We support two patterns for split op. split + builtin.split and split + builtin.slice pattern


paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
```diff
@@ -133,6 +133,271 @@ class ConcatBf16QuantizePattern
...
 
+class SplitSliceBf16QuantizePattern
+    : public pir::OpRewritePattern<paddle::onednn::dialect::SplitOp> {
+ public:
+  using pir::OpRewritePattern<
+      paddle::onednn::dialect::SplitOp>::OpRewritePattern;
+  bool MatchAndRewrite(
+      paddle::onednn::dialect::SplitOp op,
+      pir::PatternRewriter &rewriter) const override {  // NOLINT
+    // The output op should be builtin.slice to deal vector.
```
LLee233: What about adding description of the situation this subgraph appears? Can explain why we need this pattern. :)

paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
```diff
@@ -133,6 +133,271 @@ class ConcatBf16QuantizePattern
...
+     *  vec[tensor<3x6x6xbf16>,tensor<3x6x6xbf16>]
+     *               |                  |
+     *        builtin.slice         builtin.slice
+     *     tensor<3x6x6xbf16>     tensor<3x6x6xbf16>
+     *            |                     |
+     *         other op               other op
+     */
+
+    auto next_op_list = pir::GetUseOpsForOutput(op, 0);
+    // if (next_op_list.size() != 1) return false;
```
LLee233: Plz delete the annotation.

paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
```diff
@@ -133,6 +133,271 @@ class ConcatBf16QuantizePattern
...
+    for (auto i = 0; i < static_cast<int>(next_op_list.size()); i++) {
+      pir::SliceOp next_op = (next_op_list[i].first)->dyn_cast<pir::SliceOp>();
+      if (!next_op) {
```
LLee233, line 228: I just realize that split can also be followed by pd_op.xxx (with vector input). In this case we may miss "split+slice+pd_op.xxx" due to this check. Do you think if we need to consider this situation?

paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
```diff
@@ -133,6 +133,271 @@ class ConcatBf16QuantizePattern
...
+    std::unordered_map<std::string, pir::Attribute> q_attributes;
+    q_attributes["scale"] = rewriter.float_attr(1.0f);
+    q_attributes["shift"] = rewriter.float_attr(0.0f);
+    q_attributes["is_negative_input"] = rewriter.bool_attr(false);
+    q_attributes["output_format"] = rewriter.str_attr("NCHW");
+    q_attributes["bfloat16"] = rewriter.bool_attr(true);
+
+    // Insert quantize before split
+    pir::Value split_input = op.x();
+    paddle::onednn::dialect::QuantizeOp quant_op =
```
LLee233: Suggest putting the creation after line 193 (the VectorType check). Cause when we use `MatchAndRewrite`, the change before will be applied simultaneously even returning false, which means we still add quantize on the graph here.

paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
```diff
@@ -133,6 +133,271 @@ class ConcatBf16QuantizePattern
...
+        rewriter.Build<paddle::onednn::dialect::QuantizeOp>(split_input,
+                                                            q_attributes);
+    auto type = op->result_type(0);
+    if (!type.isa<pir::VectorType>()) {
+      return false;
+    }
+    auto vec_type = type.dyn_cast<pir::VectorType>();
+    auto quantize_type_ = vec_type[0];
```
LLee233: Is it appropriate to use result type of split to change that of quantize? Even though the dtype of input & output of split are same, it seems more aligned logically to use input. Or you can use `quant_op.result_type(0)` directly to create new type.

paddle/fluid/pir/transforms/onednn/cpu_special_ops_bf16_pass.cc
```diff
@@ -133,6 +133,271 @@ class ConcatBf16QuantizePattern
...
+      auto dense_type =
+          vec_type[idx].dyn_cast<paddle::dialect::DenseTensorType>();
+      auto new_type = paddle::dialect::DenseTensorType::get(
+          rewriter.ir_context(),
+          paddle::dialect::TransToIrDataType(phi::DataType::BFLOAT16,
+                                             rewriter.ir_context()),
+          dense_type.dims(),
+          dense_type.data_layout(),
+          dense_type.lod(),
+          dense_type.offset());
```
LLee233: Seems you could use the function at the top, they're same.

LLee233 (commented):
The 2nd pattern should be similar.

LLee233 (approved):
LGTM

PR #71214, opened 2025-02-20
add call to spmd in expand infermeta when use pir

jeff41404:
### PR Category
Auto Parallel

### PR Types
Bug fixes

### Description
pcard-89456

During the validation of the multimodal `Qwen2VL`, the dynamic to static encountered an error when running operator of `flash_attn_unpadded` in pir mode. The reason was that the input of operator of `flash_attn_unpadded` lacked distributed attributes. Tracing along the calculation graph, it was found that the distributed attributes would be lost after `pd_op.expand`. Further examination of the code reveals that ExpandOp (ExpandOp::InferMeta, etc.) under PIR is handwritten, and there is no processing for distributed attributes in the handwritten logic (regular OP generated code will have relevant processing). The reason for handwriting is described in PR #58767 , so this PR will add processing for distributed attributes in ExpandOp::InferMeta.

image
image
image

### DIFF
--- a/paddle/fluid/pir/dialect/operator/ir/manual_op.cc
+++ b/paddle/fluid/pir/dialect/operator/ir/manual_op.cc
@@ -3695,6 +3695,45 @@ std::vector<pir::Type> ExpandOp::InferMeta(
       dense_out.layout(),
       dense_out.lod(),
       dense_out.offset());
+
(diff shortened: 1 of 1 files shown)


From00 (approved):
LGTM

PR #71253, opened 2025-02-24
fix concat spmd rule when input is one dimensional

jeff41404:
### PR Category
Auto Parallel

### PR Types
Bug fixes

### Description
pcard-89497
when use distributed auto parallel, executing concat error on 1D tensors can be reproduced by the following code. this PR will fix this issue. By the way, fix issue of concat error when put_along_axis use concat in dynamic shape.
```python
# python -m paddle.distributed.launch --gpus=0,1,2,3,4,5,6,7 --log_dir logs xxx.py
import paddle

paddle.seed(42)
x = paddle.randn(shape=[16], dtype='float32')
y = paddle.randn(shape=[8], dtype='float32')
z = paddle.concat([x, y]) # ok, shape is [24]

mesh = dist.ProcessMesh([0, 1, 2, 3, 4, 5, 6, 7], dim_names=['dp'])
dist_x = dist.shard_tensor(x, mesh, [dist.Replicate()])
# dist_x = dist.shard_tensor(x, mesh, [dist.Shard(0)]) # shard x/y will get the same result
dist_y = dist.shard_tensor(y, mesh, [dist.Replicate()])
dist_z = paddle.concat([dist_x, dist_y]) # will report error
```
image

### DIFF
--- a/paddle/phi/infermeta/spmd_rules/concat.cc
+++ b/paddle/phi/infermeta/spmd_rules/concat.cc
@@ -85,8 +85,17 @@ SpmdInfo ConcatInferSpmd(const std::vector<DistMetaTensor>& x, int axis) {
   std::string align_axis;
   std::tie(all_axis, align_axis) = FillConcatNotation(ndim, dim);
   std::vector<std::string> axis_names(input_attrs.size(), all_axis);
-  AlignDimsSharding(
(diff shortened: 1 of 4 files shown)


From00 (approved):
LGTM

zhiqiu (approved):
LGTM

PR #71385, opened 2025-03-03
[SOT] Fix lost error info

DrRyanHuang:
### PR Category

Execute Infrastructure

### PR Types

Bug fixes

### Description

```python
class BreakGraphError(SotErrorBase):
    def __init__(self, reason: BreakGraphReasonBase | str = None):
        super().__init__()   # <------ 

        if isinstance(reason, str):
            # if reason is a string, then create a UnspecifiedBreakReason object
            reason = UnspecifiedBreakReason(reason)
        self.reason = reason
        BreakGraphReasonInfo.collect_break_graph_reason(reason)
```

Due to the inheritance relationship `BreakGraphError <- SotErrorBase <- Exception`, not passing any arguments to `super().__init__()` will result in the raised exception containing no information.😱😱😱

PCard-66972

### DIFF
--- a/python/paddle/jit/sot/utils/exceptions.py
+++ b/python/paddle/jit/sot/utils/exceptions.py
@@ -50,7 +50,6 @@ class DataDependencyControlFlowBreak(DataDependencyBreak):
 
     def __init__(self, reason_str=None, file_path="", line_number=-1):
         if reason_str is None:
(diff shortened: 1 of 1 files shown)


SigureMo (approved):
LGTMeow

PR #71425, opened 2025-03-05
[XPU] Fix error for dynamic shape send recv

AndSonder:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

Fix error for dynamic shape send recv

Pcard-76459


From00 (approved):
LGTM

PR #71438, opened 2025-03-05
[CustomOp] Pop `name` option from kwargs to avoid duplicate `name` option

risemeup1:
### PR Category

Environment Adaptation

### PR Types

Others

### Description

fix cpp_extension bug


SigureMo (approved):
LGTMeow

PR #71481, opened 2025-03-07
fix cuda arch support for DeepEP

Hongqing-work:
### PR Category

Communication Library

### PR Types

Bug fixes

### Description

pcard-67164
This PR add arch control on kernel code in deepep communication library


zhiqiu (approved):
LGTM

PR #71669, opened 2025-03-14
Support appending print op for values

0x3878f:
### PR Category

User Experience

### PR Types

Improvements

### Description

Pcard-67164
Support appending print op for values


chen2016013 (approved):
LGTM

PR #71852, opened 2025-03-24
[XPU] fix XPU linking error on Clang

lj970926:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

Some XPU Kernels use `dev_ctx.template Alloc` to call a no-template function, which would cause error in clang compiler.

2b91bb20d9b47bc1f307ceaf234041b1


houj04 (approved):

tizhou86 (approved):
LGTM

PR #71868, opened 2025-03-24
use slice to implement r2s of split_axis 

liufengwei0103:
### PR Category

Auto Parallel 
### PR Types

Performance

### Description

r_2_s uses slice kernel


liufengwei0103:
修改后 119 us
image
修改前 222 us
image

From00 (dismissed):
LGTM

jeff41404:
> r_2_s uses slice kernel when split axis is 0

Is there no requirement for axis, just using `slice` instead of `split` to improve performance

liufengwei0103:
> > r_2_s uses slice kernel when split axis is 0
> 
> Is there no requirement for axis, just using `slice` instead of `split` to improve performance

already removed axis requirement

From00 (approved):

PR #72003, opened 2025-03-31
fix nonzero

aquagull:
### PR Category

User Experience

### PR Types

Bug fixes

### Description

错误复现
```
x = paddle.zeros(shape=[1], dtype=bool)
pa_non = paddle.nonzero(x, as_tuple=True)
print(pa_non)
x_torch = torch.zeros((1), dtype=torch.bool)
tor_non = torch.nonzero(x_torch, as_tuple=True)
print(tor_non)
```

```
(Tensor(shape=[0, 1], dtype=int64, place=Place(gpu:0), stop_gradient=True,
       []),)
(tensor([], dtype=torch.int64),)
```

### DIFF
--- a/python/paddle/tensor/search.py
+++ b/python/paddle/tensor/search.py
@@ -523,10 +523,6 @@ def nonzero(x: Tensor, as_tuple=False):
              [3]])
 
     """
-    list_out = []
-    shape = x.shape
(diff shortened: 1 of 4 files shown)


HydrogenSulfate (approved):
LGTM

PR #72023, opened 2025-04-01
[AutoParallel] Fix st save dy load in pir

Xing-lil:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

Fix static save dynamic load bug in pir.
Remove redundant `_is_initialized` check that skips param_state matching.
The param becomes initialized after `param.set_value(state)` in the following lines.
Pcard-70448


liym27 (approved):
LGTM

PR #72029, opened 2025-04-02
[CINN] Composite reduce pass for argidx reduction

Enigmatisms:
### PR Category
CINN

### PR Types
New features

### Description
Full support for argidx reduction in composite reduce pass. This PR, together with:
- PR #72010 
- PR #72008 
- PR #71956

delivers full support for `argmin` and `argmax` op to take part in CINN op-fusion. For example: `cinn_op.argmax` will replace some of the `pd_op.argmax` and get further lowered and fused with other ops. Currently, `uint8`, `int16`, `int32`, `int64`, `float16`, `float32`, `float64` value type and `int32`, `int64` index type are supported.

##### Performance
The performance report is a rather coarse one, **further investigation is needed**. We selected some typical TIPC models with heavy `argmax` and test the performance. The following data (CUDA kernel runtime) is collected via nsys.

|Model|all CUDA kernel runtime non-fused version/ms|all CUDA kernel runtime fused version/ms|`argmax` runtime ratio (PHI kernel)%|`argmax` runtime ratio (CINN JIT)%|runtime reducing|Note|
|:---:|:----:|:----:|:----:|:----:|:---:|:---:|
|enet|3458.82|3004.25|14.4|1.9|**-13.14%**|/|
|fastscnn|3924.87|3452.52|12.7|2.3|**-12.03%**|/|
|rtformer_base|793.26|739.36|8.0|1.1|**-7.29%**|/|
|bisenetv2|3206.58|2995.04|7.8|1.4|**-6.60%**|/|
|mobileseg_mv3|4074.27|3955.91|3.1|0.5|**-2.91%**|/|
|pfpnnet|53201.70|51666.77|1.0|0.2|**-2.89%**|/|
|slanet|7943.43|7938.86|2.3|0.0(multiple kernels)|-0.06%|Too many non-fused argmax kernel|
|rec_mv3_tps_bilstm_att_v2_0|1009.41|1011.76|1.0|2.4(total: multiple kernels)|+0.23%|Too many non-fused argmax kernel|
|rec_resnet_rfl|1271.65|1282.40|0.6|0.0(multiple kernels)|+0.84%|Too many non-fused argmax kernel|
|f_en_table_structure|6960.12|7028.03|2.5|2.7|+0.98%|Too many non-fused argmax kernel|
|f_rec_r32_gaspin_bilstm_att|995.16|1007.31|0.8|0.0(multiple kernels)|+1.21%|Too many non-fused argmax kernel|

Pcard-89620

PR #72166, opened 2025-04-10
[CINN] Composite reduce pass horizontal fusion

Enigmatisms:
### PR Category
CINN

### PR Types
Bug fixes

### Description

Supports horizontal fusion for different types of composite reduce. For example:

```python
def test_argmax_discrete_hfuse(self):
    # discrete reduce + horizontal fuse
    bn = paddle.nn.BatchNorm(256)

    def func(x):
        return bn(x), x.argmax(axis=0)

    x = paddle.randn([128, 256])
    self.eval(func, [x])
```
Originally, the composite reduce pass in PR #72029 only supports vertical fusion for different ops. Basically, for one lowered-func, the `RealizeCompositeReducePass` assumes that there is only one unique `CompositeTypes`, which will lead to incorrect tensor buffer type deduction when the lowered-func contains different composite reduce types, and ultimately, replacing cross thread reduction will fail in these cases.

In this PR, we stopped using a shared `CompositeTypes` for all the store buffers. For each store buffer (related to composite reduce), we will try computing the correct composite type. For argidx, we might not be able to deduct the types for some of the buffers, so a second pass will be applied to resolve these types with *buffer-to-composite-type* map computed in the first pass.

Also, the unit test for composite reduce is renamed to `test_cinn_composite_reduce.py` and the horizontal fusion case is added.

Pcard-89620

### DIFF
--- a/paddle/cinn/optim/realize_composite_reduce_pass.cc
+++ b/paddle/cinn/optim/realize_composite_reduce_pass.cc
@@ -55,8 +55,16 @@ struct CompositeTypes : public std::vector<common::Type> {
     this->reserve(2);
   }
 
+  bool operator==(const CompositeTypes& other) const {
+    if (this->type != other.type || other.size() != this->size()) return false;
+    for (size_t i = 0; i < other.size(); i++) {
+      if (this->at(i) != other.at(i)) return false;
+    }
+    return true;
+  }
+
   void Print() const {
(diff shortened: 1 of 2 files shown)


Enigmatisms (commented):
self reviewed

lshpku (approved):

PR #72199, opened 2025-04-11
Fix shared layer desc probs

FeixLiu:
### PR Category
Distributed Strategy

### PR Types
Bug fixes 

### Description
Pcard-73145
Shared layer desc bug fixer for comm group.


ForFishes (approved):
LGTM

PR #72274, opened 2025-04-15
[SOT][Faster Guard] fix `ExprNode` `delete-non-abstract-non-virtual-dtor` warning

gouzil:
### PR Category

User Experience

### PR Types

Bug fixes

### Description

修复 `ExprNode` 编译时出现的 `delete-non-abstract-non-virtual-dtor`


SigureMo (approved):
LGTMeow

PR #72275, opened 2025-04-15
[XPU] adapt to new interface for normal

lj970926:
### PR Category

Custom Device

### PR Types

Improvements

### Description

normal kernel uses `xpu::normal_` , which is the nv-like version of XPU.


RuohengMa (approved):
LGTM

QingshuChen (approved):

PR #72348, opened 2025-04-17
[CINN] Fixed TileBroadcast tactic matching conditions

Enigmatisms:
### PR Category
CINN

### PR Types
Bug fixes

### Description
Add one more rule that excludes reduce-related kernels for tilebroadcast tactic. Note that the old matching condition is well-crafted, the failed case should be caused by the following (yet unknown) PRs (unrelated to TileBroadcast Tactic), but anyway, we patch the potential flaw in this PR.

Pcard-89620

### DIFF
--- a/test/dygraph_to_static/test_pir_selectedrows.py
+++ b/test/dygraph_to_static/test_pir_selectedrows.py
@@ -17,7 +17,6 @@ import unittest
 
 from dygraph_to_static_utils import (
     Dy2StTestBase,
-    test_phi_only,
 )
 
 import paddle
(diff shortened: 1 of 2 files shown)


test/dygraph_to_static/test_pir_selectedrows.py
```diff
@@ -17,7 +17,6 @@
 
 from dygraph_to_static_utils import (
     Dy2StTestBase,
-    test_phi_only,
```
lshpku, line 20: 改这里干嘛？

SigureMo: 这里需要删掉才能验证成功修复与监控，改这里没问题

lshpku (approved):

test/dygraph_to_static/test_pir_selectedrows.py
Enigmatisms: 自动删除的？下面的那个装饰器删了以后这个被判没用了

PR #72364, opened 2025-04-18
[Fix] fix recompute to save random and numpy status

zhangyuqin1998:
### PR Category

Distributed Strategy

### PR Types

Bug fixes 

### Description

fix recompute to save random and numpy status

Pcard-76459


ForFishes (approved):
LGTM

PR #72447, opened 2025-04-24
[CINN] CINN supports arange for fusion

Enigmatisms:
### PR Category
CINN

### PR Types
Improvements

##### Introduction
This PR adds support for `paddle.arange`, so that it can participate in backend op fusion (instead of using phi kernels). However, the participating arange op should satisfy the following constraint: **all the inputs should be static**, i.e., `start`, `end` and `step` should be statically known. The following example illustrates this constraint:
```python
def dynamic_arange():
    @paddle.jit.to_static(backend='CINN', full_graph=True)
    def func(x):
        return paddle.arange(0, x, 1).astype("float32") * 0.2 + 1

    x =(paddle.rand([1]) + 1) * 20
    func(x)

def static_arange_full():
    @paddle.jit.to_static(backend='CINN', full_graph=True)
    def func():
        return paddle.arange(0, 32, 1).astype("float32") * 0.2 + 1
    func()

def static_arange_with_defaults():
    @paddle.jit.to_static(backend='CINN', full_graph=True)
    def func():
        return paddle.arange(32) * 2 + 1
    func()
```
- `dynamic_arange` won't be converted to a CINN func, since the second param will cause new symbolic shape.
- `static_arange_full` and `static_arange_with_defaults` will result in `cinn_op.arange`, since the inputs can be implicitly converted to a `FullOp`. The matching condition for `pd_op.arange` to be rewritten as `cinn_op.arange` is that all three input params have `defining_op` as `FullOp`.

Other modifications:
- In `ops.yaml` for CINN, a new type `ScalarType` is introduced. Originally, `Scalar` written in the config file will be converted to `float`, making the generated op dialects have pre-determined  types for `Build` function, which might not be good. `ScalarType` will be converted to `phi::Scalar`, it retains dtype info, while it is not explicitly typed. `ConvertAttribute` function will automatically convert the `phi::Scalar` according to its underlying dtype (to a variant). Therefore, frontend (`StrategyForXXXSymbolic`) can extract useful type info to instantiate proper `Expr`.

##### Notes on Implementation
In `ops.yaml`, there might be something confusing like the following:

```yaml
args : (Tensor start_, Tensor end_, Tensor step_, ScalarType start, ScalarType end, ScalarType step, DataType dtype=DataType::INT64)
```
One might wonder why we need `Tensor start_` and `ScalarType start`. This is due to the reason that:

- `Tensor` typed input **must** be present in the `cinn_op.arange`, otherwise, during symbolic shape inference, since the current code relies on `op->operand_source(index)`, if there is no `Tensor` typed input (converted to `pir::Value`), the CINN API will be considered to have **no input**, thus resulting in out-of-range indexing for operand source.
- `ScalarType` typed input **should be given**, in order to calculate the static tensor shape for `elementwise.cc` CINN symbolic strategy. The input should be statically *know-able*, otherwise it has risks for introducing new symbolic shape and getting denied for CINN.
- Though it seems that there are 6 inputs for `cinn_op.arange` in the `ops.yaml`, the `ScalarType` typed three will actually be converted to function attribute and gets extracted by node attributes in the CINN end.

Pcard-89620


paddle/cinn/hlir/dialect/operator/ir/ops.yaml
Enigmatisms, line 1: 可能... 不一定要用新的 ScalarType。毕竟Scalar[] 是可以转成 std::vector<Scalar>的，可能可以弥补一下目前generator中无法直接得到phi::Scalar类型输入参数的问题。

Enigmatisms: 感觉有点不伦不类

Enigmatisms: 实际看了一下，如果用 Scalar[] 的话，要改的地方更多。

PR #72461, opened 2025-04-24
fix put along axis grad 0 size bug

phlrain:
### PR Category
Operator Mechanism 

### PR Types
Bug fixes

### Description

pcard-76996

修复 put along axis grad op 0 size bug


wanghuancoder (approved):
LGTM

PR #72493, opened 2025-04-25
[Auto Parallel] add main_grad for sharding in auto dy

Xing-lil:
### PR Category

Auto Parallel

### PR Types

New features

### Description

- Add inplace `param.main_grad` replaces old `master_grad` in auto dy.
- Enable by setting `export FLAGS_enable_tensor_fusion =1`.

sharding tensor_fusion PR https://github.com/PaddlePaddle/Paddle/pull/72508

Pcard-70448

### DIFF
--- a/python/paddle/optimizer/optimizer.py
+++ b/python/paddle/optimizer/optimizer.py
@@ -2000,16 +2000,16 @@ class Optimizer:
             for param in self._param_groups:
                 if param.stop_gradient:
                     continue
-                enable_inplace_master_grad = (
-                    os.getenv("FLAGS_enable_inplace_master_grad") == '1'
-                )
-                if not enable_inplace_master_grad:
+                if os.getenv("FLAGS_enable_inplace_master_grad") == '1':
+                    if (
+                        hasattr(param, "main_grad")
+                        and param.main_grad is not None
+                    ):
+                        params_grads.append((param, param.main_grad))
+                else:
                     if param._grad_ivar() is not None:
                         grad_var = param._grad_ivar()
                         params_grads.append((param, grad_var))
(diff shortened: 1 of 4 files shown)


liym27 (approved):
LGTM.

Why use main_grad instead of master_grad here?

python/paddle/optimizer/optimizer.py
```diff
@@ -2000,9 +2000,16 @@ def step(self) -> None:
             for param in self._param_groups:
                 if param.stop_gradient:
                     continue
-                if param._grad_ivar() is not None:
-                    grad_var = param._grad_ivar()
-                    params_grads.append((param, grad_var))
+                if os.getenv("FLAGS_enable_tensor_fusion") == '1':
```
liym27, line 2003: 1. Flag 需要考虑 true True的情况
2. 不建议在 auto_cast.py   api.py  optimizer.py 文件中，都用 FLAGS_enable_tensor_fusion 判断，可以改成配置。

Xing-lil:
> LGTM.
> 
> Why use main_grad instead of master_grad here?

tensor_fusion fuse grad into a contiguous fuse_grad, requiring inplace grad to avoid concat in each step.
differences between main_grad and master_grad:
1. main_grad uses inplace cast, whereas master_grad does not.
2. main_grad cast after each grad node, while master_grad after entire backward.
3. main_grad access by param.main_grad, master_grad access by grad access.

python/paddle/optimizer/optimizer.py
Xing-lil: 好的，将在后续PR上一起修改

PR #72546, opened 2025-04-29
Change the dependency of absl

xkkkkkk23:
### PR Category
CINN

### PR Types
Others

### Description

修改删除部分absl依赖

PR #72598, opened 2025-05-07
[CINN] CINN supports arange for fusion

Enigmatisms:
### PR Category
CINN

### PR Types
Improvements

##### Introduction
(This is the re-opened version of #72447, #72447 is closed due to bad git rebase and merge, caused by the removal of ABSL dependency introduced in #72546)
This PR adds support for `paddle.arange`, so that it can participate in backend op fusion (instead of using phi kernels). However, the participating arange op should satisfy the following constraint: **all the inputs should be static**, i.e., `start`, `end` and `step` should be statically known. The following example illustrates this constraint:
```python
def dynamic_arange():
    @paddle.jit.to_static(backend='CINN', full_graph=True)
    def func(x):
        return paddle.arange(0, x, 1).astype("float32") * 0.2 + 1

    x =(paddle.rand([1]) + 1) * 20
    func(x)

def static_arange_full():
    @paddle.jit.to_static(backend='CINN', full_graph=True)
    def func():
        return paddle.arange(0, 32, 1).astype("float32") * 0.2 + 1
    func()

def static_arange_with_defaults():
    @paddle.jit.to_static(backend='CINN', full_graph=True)
    def func():
        return paddle.arange(32) * 2 + 1
    func()
```
- `dynamic_arange` won't be converted to a CINN func, since the second param will cause new symbolic shape.
- `static_arange_full` and `static_arange_with_defaults` will result in `cinn_op.arange`, since the inputs can be implicitly converted to a `FullOp`. The matching condition for `pd_op.arange` to be rewritten as `cinn_op.arange` is that all three input params have `defining_op` as `FullOp`.

Other modifications:
- In `ops.yaml` for CINN, a new type `ScalarType` is introduced. Originally, `Scalar` written in the config file will be converted to `float`, making the generated op dialects have pre-determined  types for `Build` function, which might not be good. `ScalarType` will be converted to `phi::Scalar`, it retains dtype info, while it is not explicitly typed. `ConvertAttribute` function will automatically convert the `phi::Scalar` according to its underlying dtype (to a variant). Therefore, frontend (`StrategyForXXXSymbolic`) can extract useful type info to instantiate proper `Expr`.

##### Notes on Implementation
In `ops.yaml`, there might be something confusing like the following:

```yaml
args : (Tensor start_, Tensor end_, Tensor step_, ScalarType start, ScalarType end, ScalarType step, DataType dtype=DataType::INT64)
```
One might wonder why we need `Tensor start_` and `ScalarType start`. This is due to the reason that:

- `Tensor` typed input **must** be present in the `cinn_op.arange`, otherwise, during symbolic shape inference, since the current code relies on `op->operand_source(index)`, if there is no `Tensor` typed input (converted to `pir::Value`), the CINN API will be considered to have **no input**, thus resulting in out-of-range indexing for operand source.
- `ScalarType` typed input **should be given**, in order to calculate the static tensor shape for `elementwise.cc` CINN symbolic strategy. The input should be statically *know-able*, otherwise it has risks for introducing new symbolic shape and getting denied for CINN.
- Though it seems that there are 6 inputs for `cinn_op.arange` in the `ops.yaml`, the `ScalarType` typed three will actually be converted to function attribute and gets extracted by node attributes in the CINN end.

Pcard-89620


Enigmatisms (commented):
self-reviewed, conflicts resolved.

lshpku (dismissed):

paddle/cinn/hlir/dialect/operator/ir/ops.yaml
```diff
@@ -1,3 +1,11 @@
+- op : arange
+  args : (Tensor start_, Tensor end_, Tensor step_, ScalarType start, ScalarType end, ScalarType step, DataType dtype=DataType::INT64)
```
zyfncg, line 2: 1. 为什么需要 ScalarType 参数？

paddle/cinn/hlir/dialect/operator/ir/ops.yaml
```diff
@@ -1,3 +1,11 @@
...
+  output : Tensor(out)x
```
zyfncg: x？

paddle/fluid/pir/utils/general_functions.h
```diff
@@ -259,4 +259,32 @@ bool ValueIsPersistable(const pir::Value& value);
...
 phi::DataType GetValueDtype(const pir::Value& val);
 
+/**
+ * @brief Check whether the specific input of the operation is defined by
+ * the given target operation type.
+ *
+ * SourceOpT and TargetOpT should be the derived class of pir::Op
+ */
+template <typename TargetOpT, typename SourceOpT>
+bool IsDefinedBy(const SourceOpT& op, const size_t idx) {
```
zyfncg: 1. SourceOp 看起来不需要使用模板类型
2. idx的参数命名看不出来是op的input index，作为公共函数最好写的语义清晰一些。或者直接传Value进来也行

paddle/phi/infermeta/nullary.cc
```diff
@@ -22,12 +22,48 @@ void ArangeInferMeta(const Scalar& start,
...
+    int64_t arange_size = 0;
+
+    switch (dtype) {
+      case DataType::FLOAT32:
+        GET_SIZE_GIVEN_TYPE(float)
+      case DataType::FLOAT64:
+        GET_SIZE_GIVEN_TYPE(double)
+      case DataType::INT32:
+        GET_SIZE_GIVEN_TYPE(int)
+      default:
```
zyfncg, line 60: 是否需要考虑fp16的类型？

paddle/cinn/hlir/dialect/operator/transforms/pd_to_cinn_pass.cc
```diff
@@ -529,6 +514,66 @@ class SliceOpPattern : public pir::OpRewritePattern<paddle::dialect::SliceOp> {
...
+      input_list[i] = input;
+    }
+    auto cinn_arange =
+        rewriter.Build<cinn::dialect::ArangeOp>(op->operand_source(0),
+                                                op->operand_source(1),
+                                                op->operand_source(2),
+                                                input_list[0],
+                                                input_list[1],
+                                                input_list[2],
+                                                dtype);
```
zyfncg, line 570: CINN的ArangeOp是否可以只传入Tensor处理？

paddle/cinn/hlir/dialect/operator/ir/ops.yaml
Enigmatisms: 两种类似意义参数同时存在是出于如下考虑：
- Tensor 参数需要用于 symbolic shape 推导（且仅仅使用在 symbolic shape 推导中），对于 CINN 而言 arange 目前只支持静态的配置，所以不会考虑从 Tensor 中取 data
- ScalarType 是为了让 op 自动生成时生成 phi::Scalar 类型输入参数，phi::Scalar 可以方便取出内部的值（以计算静态的shape大小）以及进行多种类型的转换。原来的实现就直接用了 float，觉得直接用 float 或者某个固定的类型会引入正确性的问题（比如极端值输入或者精度问题）。相当于以 ScalarType 作为一个擦除了类型的输入。而op gen没有提供可以直接产生 phi::Scalar 的接口，所以这里加了一个。

paddle/cinn/hlir/dialect/operator/ir/ops.yaml
Enigmatisms: 应该是typo,已改

paddle/fluid/pir/utils/general_functions.h
Enigmatisms: ok

paddle/phi/infermeta/nullary.cc
Enigmatisms: 不需要，paddle.arange 接口本身只支持四种类型。

paddle/cinn/hlir/dialect/operator/transforms/pd_to_cinn_pass.cc
Enigmatisms: 之前考虑过，但发现改动起来比现在更复杂：需要能判断 Tensor 符合输入要求并且可以取出 CINN 可直接使用的值（后面这一步我不熟悉怎么做，Tensor 看起来比较 meta，就只保存了tensor的语义信息），如果有办法可以直接从 Tensor 中取出静态的值就可能可以化简这里的写法，但还是需要测试的，此前用Tensor传入后出现过算子被 CINN denied 的现象，原因就是动态的 tensor 导致了 new symbolic shape。

paddle/fluid/pir/utils/general_functions.h
Enigmatisms: 这个地方第一点我查了一下，由于输入的实际类型是 paddle::dialect::XXXOp（具体的 Op类型在 pd_op.h 中定义的），所以处理成 SourceOpT 和 paddle::dialect::XXXOp 的基类指针（OpBase）应该都可以。IsDefinedBy 这个函数实际上是从 pd_to_cinn_pass.cc 中直接摘过来的，所以没有修改其原来的写法。

zyfncg (approved):

lshpku (approved):

PR #72622, opened 2025-05-08
[Distributed] Support custom state setter and getter for recompute to save custom states

zhangyuqin1998:
### PR Category

Distributed Strategy

### PR Types

Bug fixes 
### Description

Support custom state setter and getter for recompute to save custom states

Pcard-76459


ForFishes (approved):
LGTM

PR #72312, opened 2025-04-16
[Auto Parallel] Add spmd_rule for 11 ops.

Glencsa:
### PR Category

Auto Parallel 

### PR Types

New features

### Description

Add spmd rule for forward ops, the corresponding backward ops while be added later. 

op lists: round, mish, elu, selu, celu, stanh, softplus, softshrink, thresholded_relu, logit, nonzero.

PR #72177, opened 2025-04-10
[Auto-parallel] add unary ops which have spmd_rule but not add in yaml file.

Glencsa:
### PR Category

Auto Parallel

### PR Types

New features

### Description

add unary ops which have spmd_rule but not add in ops.yaml file.

List of ops to which the spmd_rule is added(number: 28): poisson, lgamma, erf, erfinv, bernoulli, sign, ceil, log10, acos, acosh, atan, tan, atanh, log1p, swish, log, sinh, softsign, asinh, trunc, sqrt, asin, tanh_shrink, tanh, expm1, log2, floor, logsigmoid

PR #72699, opened 2025-05-13
[Auto Parallel] Add spmd_rule for five backward ops.

Glencsa:
### PR Category

Auto Parallel

### PR Types

New features

### Description

Adds spmd_rule for five backward ops (**round_grad, trunc_grad, ceil_grad, floor_grad, poisson_grad**), for which spmd_rule were added to the forward ops in previous PRs.([PR#72312](https://github.com/PaddlePaddle/Paddle/pull/72312) and [PR#72177](https://github.com/PaddlePaddle/Paddle/pull/72177)).


paddle/phi/ops/yaml/backward.yaml
```diff
@@ -375,6 +375,7 @@
   infer_meta :
     func : UnchangedInferMeta
     param: [out_grad]
+    spmd_rule : ElementwiseUnaryWithPartialInferSpmd
```
jeff41404, line 378: why not use `ElementwiseUnaryGradInferSpmd`?

Glencsa: The backward ops of these five operators only pass in one parameter(out_grad), but ElementwiseUnaryGradInferSpmd requires two or three parameters.

PR #72720, opened 2025-05-14
[Auto Parallel] Add spmd rule No.2-3、8 for (`log_softmax,cummax,cummin`) and their backward ops

ooooo-create:
### PR Category

Auto Parallel

### PR Types

New features

### Description

为如下算子添加了切分推导规则：
1. log_softmax 和 log_softmax_grad，复用 softmax 推导规则，之前已在 rules.cc 注册，pr 只配置了 yaml，没有添加单测，
2. 修复之前 topk_grad 的切分推导规则
4. cummax 和 cummax_grad，复用 topk 的推导逻辑，同样将 axis 强制转成 Replicated
5. cummin 和 cummin_grad，与 cummax 相同的处理


ooooo-create:
@Yeenyeong This pr is ready to be reviewed, Thanks!

paddle/phi/infermeta/spmd_rules/utils.cc
```diff
@@ -86,6 +86,10 @@ std::unordered_map<std::string, int64_t> ShardingMergeForTensors(
   for (auto& pair : tensor_axes_to_dim_pairs) {
     for (size_t i = 0; i < pair.second.size(); ++i) {
       auto tensor_axis = pair.first.substr(i, 1);
+      // Cooperate with GetDimsMappingForAxes to deal "1" for replicated.
+      if (tensor_axis == "1") {
```
Yeenyeong: What scenario does there lines of code work in? Is there any bug that has happened?

ooooo-create: GetDimsMappingForAxes will auto treat "1" to replicated，and the input `axis_to_dim_map` always come from ShardingMergeForTensors's output.I thought a statution, when x_axes="a1b" and dim_mapping={0,1,-1}, y_axes="a1b" dim_mapping={0,-1,1}, now will out `{a:0,1:1,b:-1}`, the b map to -1, and i think the b is also can map to 1(because when we set "1", it should be -1). I'm not sure if my modification is necessary, it might not be encountered in actual scenarios. Do I need to change it back

Yeenyeong: In the situation you assumed, the output of merging should be {a:0, 1:-1, b:1}, which comes from the rules that the ShardingMergeForAxis function follows.

In addition, an axis named "1" should always be in Replicate mode. So when the dims_mapping["1"] != -1 as assumed in your situation, something wrong may happen. So we'd better not to skip checking an axis named "1".

At the very least, this modification may affect all other spmd-rule-inference-functions calling the function and we'd better not to modify the code unless we have to do that.

Thanks!

ooooo-create: Thank you for such a clear and logical answer！

Yeenyeong (approved):

paddle/phi/infermeta/spmd_rules/topk.cc
```diff
@@ -95,24 +91,29 @@ SpmdInfo TopkGradInferSpmd(const DistMetaTensor& x,
           axis));
   // Build einsum notation
   std::string alphabet = "abcdefghijlopqrstuvwxyz";
-  std::string x_axes = alphabet.substr(0, x_ndim - 1);
+  std::string x_axes = alphabet.substr(0, x_ndim);
+  x_axes[axis] = '1';
```
jeff41404: why do we need this line of code? can we delete L95

paddle/phi/infermeta/spmd_rules/rules.cc
```diff
@@ -753,6 +753,15 @@ PD_REGISTER_SPMD_RULE(nonzero,
 // add_n
 PD_REGISTER_SPMD_RULE(add_n, PD_INFER_SPMD(phi::distributed::AddNInferSpmd));
 
+// cummax
+PD_REGISTER_SPMD_RULE(cummax,
+                      PD_INFER_SPMD(phi::distributed::TopkInferSpmdBase),
+                      PD_INFER_SPMD(phi::distributed::TopkGradInferSpmdBase));
```
jeff41404: Even if `CummaxInferSpmd` actually calls `TopkInferSpmdBase`, it is best to write `CummaxInferSpmd` instead of `TopkInferSpmdBase` when registering. and the parameters of `CummaxInferSpmd` and `TopkInferSpmdBase` are different. same issue with `CummaxGradInferSpmd` and `TopkGradInferSpmdBase`.

paddle/phi/infermeta/spmd_rules/rules.cc
```diff
@@ -753,6 +753,15 @@ PD_REGISTER_SPMD_RULE(nonzero,
...
+// cummin
+PD_REGISTER_SPMD_RULE(cummin,
+                      PD_INFER_SPMD(phi::distributed::TopkInferSpmdBase),
+                      PD_INFER_SPMD(phi::distributed::TopkGradInferSpmdBase));
```
jeff41404: same issue as above

paddle/phi/infermeta/spmd_rules/topk.cc
ooooo-create: Thanks! Has been deleted. becaues set all dim_mapping[axis]=-1 after, it's redundant

paddle/phi/infermeta/spmd_rules/rules.cc
ooooo-create: Thanks! I adapt the `InferSpmdContext::AttrAt in inferspmd_utils.cc` and `parse_attr in auto_parallel_py.cc` to add a new attr support `DataType`, and use `CummaxInferSpmd` instead of `TopkInferSpmdBase`

jeff41404 (approved):
LGTM

zyfncg (approved):

PR #72824, opened 2025-05-20
[Auto Parallel] Add spmd rule No.6 for unique ops.  

ooooo-create:
### PR Category

Auto Parallel

### PR Types

New features

### Description

- 为 unique 算子增加切分推导规则，输入每个维度都需要是 Replicated
- https://github.com/PaddlePaddle/Paddle/issues/72822 中 unique 的维度设置
- 没有对应的反向算子
- 前置需要 https://github.com/PaddlePaddle/Paddle/pull/72720 中的 DataType Attr 逻辑


paddle/phi/infermeta/spmd_rules/unique.cc
```diff
@@ -0,0 +1,86 @@
...
+    out_dims_mapping_dst = {-1};
+  }
+  TensorDistAttr out_dist_attr_dst =
+      CopyTensorDistAttrForOutput(x_dist_attr_src);
+  out_dist_attr_dst.set_dims_mapping(out_dims_mapping_dst);
+
+  TensorDistAttr indices_dist_attr_dst = TensorDistAttr();
+  if (return_index) {
+    indices_dist_attr_dst = CopyTensorDistAttrForOutput(x_dist_attr_src);
+    indices_dist_attr_dst.set_dims_mapping({-1});
```
Yeenyeong, line 47: If the variable "axis" is not "None", the number of dimension of "indices" should be the same as that of the input "x" (so are "inverse" and "counts").
It does not make sense to simply set "dims_mapping" as {-1}.

ooooo-create: If the variable "axis" is not "None", according to the `UniqueRawInferMeta`,they are 1D Tensor Along axis
图片

paddle/phi/infermeta/spmd_rules/unique.h
```diff
@@ -0,0 +1,44 @@
...
+#include "paddle/phi/core/distributed/type_defs.h"
+
+namespace phi {
+namespace distributed {
+
+SpmdInfo UniqueInferSpmdBase(const DistMetaTensor& x,
+                             bool return_index,
+                             bool return_inverse,
+                             bool return_counts,
+                             const std::vector<int>& axis);
```
Yeenyeong: Maybe we do not need this interface.

paddle/phi/infermeta/spmd_rules/unique.h
```diff
@@ -0,0 +1,44 @@
...
+                         const std::vector<int>& axis,
+                         DataType dtype);
+
+SpmdInfo UniqueInferSpmdStatic(const DistMetaTensor& x,
+                               bool return_index,
+                               bool return_inverse,
+                               bool return_counts,
+                               const std::vector<int>& axis,
+                               DataType dtype,
+                               bool is_sorted);
```
Yeenyeong, line 36: Why should we need a distinct "Static" interface ?

paddle/phi/infermeta/spmd_rules/unique.cc
Yeenyeong: Oh you're right! I'm sorry to misunderstand the operator. Thanks for helping me understand it correctly!

paddle/phi/infermeta/spmd_rules/unique.h
ooooo-create: Thanks，I will delete it later.

paddle/phi/infermeta/spmd_rules/unique.h
ooooo-create: because I see the unique op in dynamic mode and static mode has different parameters. In my mind, auto parallel is support with two modes, but i'am don't sure how framework to do it, so I add this interface with 'static' suffex and register in 'paddle/phi/ops/yaml/inconsistent/static_ops.yaml'. It's only my guess. Can you tell me more information about it, will I need delete it? Thanks !

Yeenyeong (approved):

test/auto_parallel/spmd_rules/test_unique_rule.py
```diff
@@ -0,0 +1,98 @@
...
+        # inferred_output_dist_attrs = result_dist_attrs[1]
+
+        # self.assertEqual(len(inferred_input_dist_attrs), 1)
+        # self.assertEqual(len(inferred_output_dist_attrs), 4)
+
+        # self.assertEqual(inferred_input_dist_attrs[0].dims_mapping, [-1, -1])
+        # self.assertEqual(inferred_output_dist_attrs[0].dims_mapping, [-1])
+        # self.assertEqual(inferred_output_dist_attrs[1].dims_mapping, [-1])
+        # self.assertEqual(inferred_output_dist_attrs[2].dims_mapping, [-1])
+        # self.assertEqual(inferred_output_dist_attrs[3].dims_mapping, [-1])
```
jeff41404: shall we open these annotated codes?

ooooo-create: Thanks, I forget it, directly open it will wrong, because it will put the axis with empty list, it's difficult to decide what the type of it to cpp, std::vector<int> or std::<int_64>. so it closed it now, because it's has tested in cpp. my be adding a default parameter can open this test, I will try it later.

ooooo-create: I found in this pr https://github.com/PaddlePaddle/Paddle/pull/57877#issuecomment-1825079886 , the problem has already emerged. Now I add support empty list in `parse_single_pyobject`,and force to `vector<int64>`, so support `std::vector<int> InferSpmdContext::AttrAt(size_t idx) const ` to deal it to pure int.
图片

图片

jeff41404 (approved):
LGTM

zyfncg (approved):

luotao1 (approved):

PR #72845, opened 2025-05-21
[Auto Parallel] Add spmd rule No.12 for label_smooth and label_smooth ops. 

ooooo-create:
### PR Category

Auto Parallel

### PR Types

New features

### Description

- 为 label_smooth 和 label_smooth_grad 算子增加切分推导规则，其中 prior_dist 涉及 broadcast 操作，且为 optional，在传入 prior_dist 时使用 ElementwiseBinaryInferSpmd 实现，否则使用 ElementwiseUnaryInferSpmd 进行实现，反向使用 ElementwiseUnaryGradInferSpmd 
- ~查看 expand_as 当前已有推导规则实现，故仅仅在 yaml 中进行注册，此外修复了 target_shape 的类型不匹配问题~


test/auto_parallel/spmd_rules/test_label_smooth_rule.py
```diff
@@ -0,0 +1,117 @@
...
+        out_tensor_dist_attr.process_mesh = process_mesh
+        self.out_dist_tensor_spec = DistTensorSpec(
+            out_shape, out_tensor_dist_attr
+        )
+
+        self.rule = core.get_phi_spmd_rule("label_smooth")
+        self.attrs = OrderedDict()
+        self.attrs['epsilon'] = 0.1
+
+    def test_topk_forward(self):
```
Yeenyeong: The name of the member function should not be copied from any other place.

test/auto_parallel/spmd_rules/test_label_smooth_rule.py
```diff
@@ -0,0 +1,117 @@
...
+        inferred_output_dist_attrs = result_dist_attrs[1]
+
+        self.assertEqual(len(inferred_input_dist_attrs), 2)
+        self.assertEqual(len(inferred_output_dist_attrs), 1)
+
+        self.assertEqual(inferred_input_dist_attrs[0].dims_mapping, [0, 1, -1])
+        self.assertEqual(inferred_input_dist_attrs[1].dims_mapping, [-1, -1])
+        self.assertEqual(inferred_output_dist_attrs[0].dims_mapping, [0, 1, -1])
+
+    def test_topk_backward(self):
```
Yeenyeong: The same comment as above.

paddle/phi/infermeta/spmd_rules/label_smooth.cc
```diff
@@ -0,0 +1,44 @@
...
+namespace distributed {
+SpmdInfo LabelSmoothInferSpmd(const DistMetaTensor& label,
+                              const DistMetaTensor& prior_dist,
+                              float epsilon) {
+  if (prior_dist.initialized()) {
+    return ElementwiseBinaryInferSpmd(label, prior_dist);
+  }
+  SpmdInfo unary_spmd_info = ElementwiseUnaryInferSpmd(label);
+  unary_spmd_info.first.push_back(TensorDistAttr());
+  VLOG(4) << "LabelSmoothInferSpmd:";
```
jeff41404: it is better to put this line before L30/L32 and make the logs more understandable? e.g.
before L30 `VLOG(4) << "LabelSmoothInferSpmd call ElementwiseBinaryInferSpmd:"`
before L32 `VLOG(4) << "LabelSmoothInferSpmd call ElementwiseUnaryInferSpmd:"`

test/auto_parallel/spmd_rules/test_label_smooth_rule.py
```diff
@@ -0,0 +1,166 @@
...
+        self.assertEqual(len(result_dist_attrs), 2)
+        inferred_input_dist_attrs = result_dist_attrs[0]
+        inferred_output_dist_attrs = result_dist_attrs[1]
+        self.assertEqual(len(inferred_input_dist_attrs), 2)
+        self.assertEqual(len(inferred_output_dist_attrs), 1)
+        self.assertEqual(inferred_input_dist_attrs[0].dims_mapping, [0, 1, -1])
+        self.assertEqual(inferred_input_dist_attrs[1].dims_mapping, [-1, -1])
+        self.assertEqual(inferred_output_dist_attrs[0].dims_mapping, [0, 1, -1])
+
+        # shape: [16, 16, 16], [1, 16]. [0, 1, -1], [-1, 0] --> [0, 1, -1], [-1, -1], [0, 1, -1]
```
jeff41404, line 115: we need more unit test case to check:
if `# shape: [16, 16, 16], [1, 16]. [0, -1, 1], [-1, 1]`, what result should be
if `# shape: [16, 16, 16], [1, 16]. [0, -1, 1], [-1, 0]`, what result should be
if `# shape: [16, 16, 16], [1, 16]. [0, -1, 1], [-1, -1]`, what result should be

paddle/phi/infermeta/spmd_rules/label_smooth.cc
ooooo-create: Thanks!done

test/auto_parallel/spmd_rules/test_label_smooth_rule.py
ooooo-create: In the second station `# shape: [16, 16, 16], [1, 16]. [0, -1, 1], [-1, 0]`
Because of reuse the `ElementwiseBinaryInferSpmd`, now it will cause error now when meet 1 and 0 with same axes (due to the `ShardingMergeForAxis`).
The following error is below, and I check the error now rather than implement it
```
 (Unimplemented) Tensor Axis[c] is Sharded by two different mesh dimension [0] and [1]. (at ../paddle/phi/infermeta/spmd_rules/utils.cc:70)
```

jeff41404: According to the calculation logic of this operator, if `# shape: [16, 16, 16], [1, 16]` and `dims_mapping: [0, -1, 1], [-1, 0] ` the result of dims_mapping should be  `[0, -1, 1], [-1, 1], [0, -1, 1]`? if so, we should add some necessary checks and modifications to dims_mapping of prior_dist before calling `ElementwiseBinaryInferSpmd`?

ooooo-create: Yes. If `# shape: [16, 16, 16], [1, 16]` and dims_mapping: `[0, -1, 1], [-1, 0]` the result of dims_mapping should be `[0, -1, 1], [-1, 1], [0, -1, 1]`
If label's rank is greater than prior_dist's axis. It has an extra_dims_mapping. When prior_dist's dims_mappin in it, replicate it to `-1` first. It will work better than before.
(In addition, I think in the future `ElementwiseBinaryInferSpmd` also should deal with this station

jeff41404 (approved):
LGTM

zyfncg (approved):

test/cpp/auto_parallel/spmd_rule_test.cc
```diff
@@ -3147,7 +3147,71 @@ TEST(Unique, Ctor) {
...
+  phi::distributed::SpmdInfo forward_info =
+      phi::distributed::LabelSmoothInferSpmd(
+          label, phi::distributed::DistMetaTensor(), epsilon);
+
+  EXPECT_EQ(forward_info.first.size(), 2UL);
+  EXPECT_EQ(forward_info.second.size(), 1UL);
+  check_dim_mapping(forward_info.first[0], {0, 1, -1});
+  const phi::distributed::ArgDistAttr& attr = forward_info.first[1];
+  if (paddle::holds_alternative<phi::distributed::TensorDistAttr>(attr)) {
+    EXPECT_EQ(paddle::get<phi::distributed::TensorDistAttr>(attr),
```
luotao1, line 3183: image
@ooooo-create 下一个PR改一下吧

luotao1 (approved):

test/cpp/auto_parallel/spmd_rule_test.cc
ooooo-create: 好的好的

PR #72902, opened 2025-05-23
[Auto Parallel] Fix acc > 1 bug for sharding overlap in auto dy

Xing-lil:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

- Fix acc > 1 bug for sharding overlap in auto dy.
- Overlap reduce_scatter with the final acc step.
- The loss curve of llama7b-layer20-sd8 is as follows:

loss_curve5000

Pcard-70448


liym27 (approved):
LGTM

PR #73079, opened 2025-06-04
Support paddle.set_flags and paddle.get_flags in custom device

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
Modified global_value_getter_setter.cc to support paddle.set_flags and paddle.get_flags in custom device.


yongqiangma (approved):
LGTM

PR #73113, opened 2025-06-05
Restore allocator_facade

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
Restore allocator_facade.cc due to an error in fast_deploy for an unknown reason.


yongqiangma (approved):
LGTM

PR #73236, opened 2025-06-10
[AutoParallel] Fix reshard_grad with partial

Xing-lil:
### PR Category

Auto Parallel

### PR Types

Bug fixes

### Description

- Fix reshard_grad bug, clean partial status in reshard_grad for the target dist_attr.
- The forward P_To_S `reshard` implements a `reduce_scatter`, with its backward `reshard_grad` is `all_gather` (equivalent to S_To_R).
- The forward P_to_R `reshard` implements a `all_reduce`, with its backward `reshard_grad` is `identity` (equivalent to R_To_R).
- Add ut to check reshard_grad with partial.

Pcard-70448


liym27 (approved):
LGTM

PR #73405, opened 2025-06-17
[Dy2St] Add `FLAGS_specialize_device_in_dy2st` to set PlaceAttribute for DataOp based on input tensor's place before lowering

0x3878f:
### PR Category

Execute Infrastructure

### PR Types

New features

### Description

Add `FLAGS_specialize_device_in_dy2st`, default value is `False`. It will set PlaceAttribute for DataOp based on input tensor's place before lowering when `FLAGS_specialize_device_in_dy2st=1` 

**NOTE:**
Performance may decrease when a CPU Tensor is copied to a device multiple times.
Consider applying CSE to solve this problem.

pcard-67164

### DIFF
--- a/paddle/common/flags.cc
+++ b/paddle/common/flags.cc
@@ -1884,6 +1884,19 @@ PHI_DEFINE_EXPORTED_bool(enable_cse_in_dy2st,
                          "Apply CSE optimize pass in Dy2St");
 
 /**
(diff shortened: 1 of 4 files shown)


SigureMo (dismissed):
LGTMeow

SigureMo (approved):

phlrain (approved):

PR #73407, opened 2025-06-17
[SOT] Avoid deep copy  while creating `MutableDictLikeData` for `DataClassInstanceVariable`

0x3878f:
### PR Category

Execute Infrastructure

### PR Types

Improvements

### Description

Use a custom asdict function to avoid deep copy for data in dataclass instance.

Introduction of dataclasses.asdict funtion
> dataclasses.asdict(obj, *, dict_factory=dict)
Converts the dataclass obj to a dict (by using the factory function dict_factory). Each dataclass is converted to a dict of its fields, as name: value pairs. dataclasses, dicts, lists, and tuples are recursed into. Other objects are copied with [copy.deepcopy()](https://docs.python.org/3.13/library/copy.html#copy.deepcopy).

The default behavior above may lead to out of memory error.

pcard-67164

### DIFF
--- a/python/paddle/jit/sot/opcode_translator/executor/variables/basic.py
+++ b/python/paddle/jit/sot/opcode_translator/executor/variables/basic.py
@@ -18,7 +18,7 @@ import dataclasses
 import operator
 import sys
 import types
-from dataclasses import asdict, is_dataclass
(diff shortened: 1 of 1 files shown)


SigureMo (approved):
LGTMeow

PR #73420, opened 2025-06-18
将signal.stft的onsided的默认参数值从True改为None

Qin-sx:
### PR Category

User Experience

### PR Types

Improvements

### Description

默认值改为None，会自动根据输入判断是否返回一半的频点值。当输入为复数时，不需要每次手动输入onsided=False。

相关pytorch代码：
```cpp
  const bool complex_fft = input.is_complex();
  const auto onesided = onesidedOpt.value_or(!complex_fft);
```

### DIFF
--- a/python/paddle/signal.py
+++ b/python/paddle/signal.py
@@ -420,7 +420,7 @@ def stft(
     norm = 'ortho' if normalized else 'backward'
 
     if onesided is None:
-        onesided = not is_complex(x)
+        onesided = not is_complex(x_frames)
 
     if is_complex(x_frames):
         assert (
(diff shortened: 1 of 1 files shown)


Qin-sx:
[pytorch](https://docs.pytorch.org/docs/stable/generated/torch.stft.html)后续会弃用return_complex=False。return_complex默认值为None。

> From version 1.8.0, return_complex must always be given explicitly for real inputs and return_complex=False has been deprecated. Strongly prefer return_complex=True as in a future pytorch release, this function will only return complex tensors.

zhwesky2010 (approved):
LGTM

PR #73486, opened 2025-06-20
[Auto Parallel] Add spmd rule No.15 for index_put and index_put_grad ops.

Glencsa:
### PR Category

Auto Parallel

### PR Types

New features

### Description

- [【开源任务】算子切分推导规则开发，支持更多模型使用自动并行，简化更多用户的分布式开发成本](https://github.com/PaddlePaddle/Paddle/issues/72415#issuecomment-2982903902)


luotao1:
https://github.com/PaddlePaddle/Paddle/issues/72415#issuecomment-2989692615

@Glencsa 请知晓

test/cpp/auto_parallel/spmd_rule_test.cc
```diff
@@ -230,6 +230,84 @@ TEST(MatmulSPMDRule, Ctor) {
   check_partial_dims(inferred_dist_attrs.second[0], {0});
   VLOG(4) << "test11 done." << std::endl << std::endl << std::endl;
 }
+
+TEST(IndexPut, Ctor) {
+  // build input data class
+  std::vector<int64_t> x_shape = {64, 64, 64};
+  std::vector<int64_t> indice_shape = {32};
+  std::vector<int64_t> value_shape = {32};
```
jeff41404: in this case, if `value_shape=[32, 64]`, can run normally on single device, but can run normally on multi device? especially dims_mapping of x is {-1, 0, 1}
tips: After using this spmd rule, the local shape will be: `x_shape=[64, 64, 32], indice_shape=([32], [32]), value_shape=[32, 64]`, Can `index_put` operator will run successfully?

paddle/phi/infermeta/spmd_rules/index_put.cc
```diff
@@ -0,0 +1,240 @@
...
+  PADDLE_ENFORCE_GE(x_ndim,
+                    indices_size,
+                    common::errors::InvalidArgument(
+                        "The ndim of x in index_put should be "
+                        "greater than or equal to the size of indices, "
+                        "but got x_ndim:[%d],indices_size:[%d].",
+                        x_ndim,
+                        indices_size));
+
+  PADDLE_ENFORCE_EQ(value_ndim,
```
jeff41404: According to the definition of the `index_put` operator, it can support value_ndim greater than 1. This restriction cannot be applied as it can be verified through the following code:
```python
x = paddle.zeros([4, 4, 4])
value = paddle.ones([4, 4]) # can support value_ndim > 1(meet certain conditions)
# value = paddle.arange(16).reshape([4, 4]).astype("float32")
ix1 = paddle.to_tensor([0,1,2,3])
ix2 = paddle.to_tensor([0,1,2,3])
indices=(ix1,ix2)

out = paddle.index_put(x,indices,value)
print(out) # can run normally
```

test/cpp/auto_parallel/spmd_rule_test.cc
Glencsa: Thanks, I fix this situation(**value_shape greater than 1**) in next commit.

paddle/phi/infermeta/spmd_rules/index_put.cc
Glencsa: Thanks, I have removed this restriction.

paddle/phi/infermeta/spmd_rules/index_put.cc
```diff
@@ -88,6 +101,8 @@ SpmdInfo IndexPutInferSpmd(const DistMetaTensor& x,
       x_dims_mapping[i] = -1;
```
jeff41404: the branch of `else` only handles the case of `indices_size = x_ndim`? can it be merged into the above branch, so we do not need to distinguish between `if` and `else`

Glencsa: Thanks for your suggestion, I removed the if and else in this code block.

jeff41404 (approved):
LGTM

zyfncg (approved):

PR #73501, opened 2025-06-20
fix performence bug

liuruyan:
### PR Category

Performance Optimization

### PR Types

Performance

### Description

pcard-67164


liym27 (approved):
LGTM for baichuan-14b perfermance

PR #73563, opened 2025-06-23
[BIT] Optimize print speed

cangtianhuang:
### PR Category

Performance Optimization

### PR Types

Improvements

### Description

Optimizing tensor printing speed:

In old code `list(var)` would iterate through all elements before slicing, which was extremely slow. 

Using direct index access `var[i]` will significantly reduce element access time, proving especially useful for ultra-large tensors. 

The printing speed for elements up to **the uint32 limit** will be reduced from 10 minutes to 5 seconds:
```python
import paddle
import time

start = time.time()
paddle_tensor = paddle.full([4294967300], 0.1, dtype=paddle.float16)
mid = time.time()
print(paddle_tensor)
end = time.time()

> Tensor(shape=[4294967300], dtype=float16, place=Place(gpu:0), stop_gradient=True,
>        [0.09997559, 0.09997559, 0.09997559, ..., 0.09997559, 0.09997559,
>         0.09997559])
> paddle.full print time:  5.98477506637573
```

### DIFF
--- a/python/paddle/tensor/to_string.py
+++ b/python/paddle/tensor/to_string.py
@@ -183,24 +183,24 @@ def _format_tensor(var, summary, indent=0, max_width=0, signed=False):
         return _format_item(var, max_width, signed)
     elif len(var.shape) == 1:
         item_length = max_width + 2
-        items_per_line = (linewidth - indent) // item_length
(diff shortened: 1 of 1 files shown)


wanghuancoder (approved):
LGTM

PR #73689, opened 2025-06-28
[Inference] support load int8 weight using wint8 algo

zhoutianzi666:
### PR Category

Inference

### PR Types

Others

### Description

P-card-71501

支持直接加载int8的权重，然后用wint8 运行算子


ming1753 (approved):
LGTM

PR #73745, opened 2025-07-01
[PHI] Fix `paddle.kthvalue` for big tensor

cangtianhuang:
### PR Category

Operator Mechanism

### PR Types

Bug fixes

#### Bug Fixes:
1. Add `typename IndexType` to distinguish index types
2. Change `shared_mem` to dynamic allocation, instead of hardcoding
3. Set `grid_dim, block_dim` to avoid exceeding
4. Resolve write race condition for `foundKValue` by adopting `CudaAtomicMin`
5. Fix grad_kernel `int64_t` handling

#### Tests:

The accuracy diff is due for multiple identical kth values, resulting in different index positions. 
image

After comparing only the first value, all accuracy tests pass
image

#### TODO:
- [x] 1. Support `int64_t k` param in API signature
- [ ] 2. Replace `shared_mem` in `RadixTopK` with dynamic allocation

Pcard-85711

### DIFF
--- a/paddle/phi/kernels/funcs/top_k_function_cuda.h
+++ b/paddle/phi/kernels/funcs/top_k_function_cuda.h
@@ -694,10 +694,10 @@ __device__ void ExclusiveBinaryPrefixScan(
   }
 }
 
-template <typename T, typename RadixType>
+template <typename T, typename RadixType, typename IndexType>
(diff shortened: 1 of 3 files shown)


wanghuancoder (approved):
LGTM

lshpku (approved):

PR #73864, opened 2025-07-07
[Dy2St] Optimize `Trans2ContiguousTensors` performance

0x3878f:
### PR Category

Execute Infrastructure

### PR Types

Performance

### Description

Optimize Trans2ContiguousTensors: We should only perform contiguous transformations in the presence of non-contiguous tensors. Otherwise, unnecessary overhead will be incurred during Tensor construction.

pcard-67164


SigureMo (commented):
LGTMeow

SigureMo (approved):

PR #73875, opened 2025-07-07
Support 0 size tensor in ProcessGroupCustom::AllToAll

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
Support 0 size tensor in ProcessGroupCustom::AllToAll.
The error occurs when ​​all values in the out_split_sizes tensor are zero for a specific device​​. This indicates that the corresponding output tensor (out_tensor) on this device is empty​​. In this scenario, the dim[0] of out_tensor becomes ​​0​​. Subsequent operations attempt to divide by this zero dimension value, triggering a division-by-zero error.
This change has been verified on Metax backend.


zhiqiu (approved):
LGTM

PR #73894, opened 2025-07-08
[Auto-Parallel] fix bug of placements in all2all for moe

liym27:
### PR Category

Auto Parallel

### PR Types

Bug fixes 

### Description

fix bug of placements in all2all for moe.
- Use sub-placements instead of placements to calculate local shape
PCard-86802


pkuzyc (approved):
LGTM

PR #73926, opened 2025-07-09
Other Custom Device used npu_identity Kernel

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
Only npu can use npu_identity now. Other custom device may use npu_identity kernel, leading to the problem shown in the following figure.
584e1cedc6a340bf0242bafaa9d84205


yongqiangma (approved):
LGTM

PR #74047, opened 2025-07-15
[Distributed] Fix bug for pp param and moment sync

zhangyuqin1998:
### PR Category

Distributed Strategy

### PR Types

Bug fixes 

### Description

Bug fix for the pp parameter and moment sync, which incorrectly set the color group of the pp shared layer description.
Pcard-76459


ForFishes (approved):
LGTM

PR #74087, opened 2025-07-17
Set the sync_param in pp config to be disabled by default

zhangyuqin1998:
### PR Category

Distributed Strategy

### PR Types

Bug fixes

### Description

Set the sync_param in pp config to be disabled by default
Pcard-76459


From00 (approved):
LGTM

PR #74107, opened 2025-07-18
Custom Device Memory Status Api Support

YqGe585:
### PR Category

Custom Device

### PR Types

New features

### Description

pcard-67164
Custom Device Memory Status Api Support. Users can now monitor memory usage using the same memory api as GPU.


yongqiangma (approved):
LGTM

PR #74123, opened 2025-07-18
[PHI] Unify expand_grad kernel behavior between cpu and gpu

ooooo-create:
### PR Category

Operator Mechanism

### PR Types

Not User Facing

### Description

Unify expand_grad kernel behavior between cpu and gpu
Same as https://github.com/PaddlePaddle/Paddle/pull/73101, 直接使用 out_grad.dims()，而不是 shape 参数


From00 (approved):
LGTM

Issue #73668, opened 2025-06-27
paddle.arange with end=float('inf') on GPU causes memory leak


rookieLiu2018:
### bug描述 Describe the Bug

When calling paddle.arange with end set to positive infinity on GPU, the operation never terminates and continually allocates memory until the system crashes. On CPU the same code returns an empty tensor without error.
```python
import paddle
paddle.set_device('gpu')
start = paddle.to_tensor(0)
end = paddle.to_tensor(float('inf'))
a = paddle.arange(start,end,1, dtype=paddle.int32)
print(a)
```
The result on the CPU：
```
Tensor(shape=[0], dtype=int32, place=Place(cpu), stop_gradient=True,
       [])
```

I think both outcomes are incorrect, the input should be validated.

### 其他补充信息 Additional Supplementary Information

```Version: paddlepaddle-gpu 3.0.0```

PR #74159, opened 2025-07-22
[Fix Issue] fix issue 73668

enkilee:
### PR Category

User Experience

### PR Types

Others

### Description

#73668 

according to pytorch:

```python
import torch
try:
    start = torch.tensor(0)
    end = torch.tensor(float('inf'))
    a = torch.arange(start, end, 1, dtype=torch.int32)
except RuntimeError as e:
    print(f"PyTorch error: {e}")
```
This will output something like:
```bash
PyTorch error: arange_cpu(): invalid 'end'  inf, must be numeric and finite
```

So add check in ArangeFunc.


wanghuancoder (approved):
LGTM

PR #74220, opened 2025-07-24
Async_offload does not support custom device, fall back to sync

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
Asyncload does not support custom device, fall back to sync


yongqiangma (approved):
LGTM

PR #74396, opened 2025-08-04
support to fallback to sharding by one mesh dim from by more than one

liufengwei0103:
### PR Category

Auto Parallel

### PR Types

Improvements

### Description

when one op not support  co-shard in spmd rules receive one input with co-shard, system will make this input fallback to shard by only one dim.

Pcard-67164


liufengwei0103:
image

zhiqiu (approved):

PR #74397, opened 2025-08-04
[Custom Device] Fix: shape_sr kernel is instantiated twice in CUDA custom device backend

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
shape_sr kernel is instantiated twice in CUDA custom device backend. This bug is caused by both PADDLE_WITH_CUSTOM_DEVICE and PADDLE_WITH_CUDA being enabled simultaneously.


yongqiangma (approved):
LGTM

PR #74411, opened 2025-08-05
[Custom Device ]Solved several problems for CUDA custom device backend

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
Metax is adding more kernels to its backend. I checked all the remaining kernels. This PR ​​fixes 4 stride kernels​​ which were being instantiated twice in the ​​CUDA custom device backend​​. Also it ​​modifies​​ some header files ​​to make them compilable​​ on this kind of backend.


yongqiangma (approved):
LGTM

PR #74428, opened 2025-08-05
Fix extract_device_id bug when there is only one device

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

pcard-67164
Fix extract_device_id bug when there is only one device


yongqiangma (approved):
LGTM

PR #74598, opened 2025-08-13
[Accuracy diff No.168] Fix accuracy (output type) diff for paddle.floor and paddle.ceil API

hushenwei2000:
### PR Category
Operator Mechanism

### PR Types
Bug fixes

### Description
Currently, the `floor` and `ceil` kernel exhibits different output types between PyTorch and Paddle. For instance, when the input datatype is `int32`, PyTorch outputs `int32`, whereas Paddle outputs `float32`. The same is `ceil`.

This is because `floor` and `ceil` are conducted type promotion. 

This PR 
1) Deleted them from the list of requiring type promotion.
2) Addde more datatypes support when registration.
3) Editde documentation.

pcard-67164


python/paddle/tensor/ops.py
```diff
@@ -701,7 +701,6 @@ def floor(x: Tensor, name: str | None = None) -> Tensor:
 
     Returns:
         Tensor. Output of Floor operator, a Tensor with shape same as input
-            (integer types are autocasted into float32).
```
wanghuancoder, line 704: 记得修改中文文档的这里。

wanghuancoder (approved):
LGTM

python/paddle/tensor/ops.py
hushenwei2000: 改了 [https://github.com/PaddlePaddle/docs/pull/7384](https://github.com/PaddlePaddle/docs/pull/7384)

zyfncg (approved):

PR #74705, opened 2025-08-18
[AutoParallel] Fix the case in 'get_local_slices' func and add deepseek v3 to ci

pkuzyc:
### PR Category
Auto Parallel 

### PR Types
Bug fixes

### Description
pcard-67164
1. Fix the case in 'get_local_slices' func when the placements length is smaller than mesh size, by setting the missing placements to 'Replicate'.
2. add deepseek v3 in paddlenlp to ci (https://github.com/PaddlePaddle/PaddleNLP/pull/10064)


liym27 (approved):
LGTM

PR #74719, opened 2025-08-19
[Accuracy diff No. 163] Fix accuracy (output type) diff for paddle.clip API

hushenwei2000:
### PR Category
Operator Mechanism

### PR Types
Bug fixes

### Description
Currently, the `clip` kernel exhibits different output types between PyTorch and Paddle. For instance, when the input datatype is `int32`/`int64` and one of the `min`/`max` is float type, PyTorch outputs `float32`, whereas Paddle outputs the origin datatype.

This PR

- Addde type cast in Python level API.

pcard-67164

### DIFF
--- a/python/paddle/tensor/math.py
+++ b/python/paddle/tensor/math.py
@@ -3728,20 +3728,6 @@ def clip(
     """
 
     x_dtype = str(x.dtype)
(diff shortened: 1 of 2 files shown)


wanghuancoder (approved):
LGTM

PR #74839, opened 2025-08-22
Fix conv2d unittest error and support custom device

YqGe585:
### PR Category

Operator Mechanism

### PR Types

Bug fixes

### Description

pcard-67164
Fix conv2d unittest error and support custom device.
tensor.dtype has two different datatype, depending on 'enable_pir_api'. This leads to a type checking error in optest.
conv2d unit test now support custom device.


yongqiangma (approved):
LGTM

PR #74929, opened 2025-08-27
fix flashmask_attention when startend_row_indices is None

umiswing:
### PR Category

Performance Optimization

### PR Types

Bug fixes

### Description

https://github.com/PaddlePaddle/Paddle/pull/74928
fix flashmask_attention when startend_row_indices is None


GuoxiaWang (approved):
LGTM

PR #75249, opened 2025-09-12
[XPU] support bool datatype for fill_any op

cqulilujia:
### PR Category
Custom Device

### PR Types
New features

### Description
Support bool datatype for fill_any op


dynamicheart (approved):
LGTM

PR #75286, opened 2025-09-15
fix _get_places for custom device

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

fix _get_places for custom device
pcard-67164


yongqiangma (approved):
LGTM

PR #75321, opened 2025-09-16
[XPU] update XPU ReLU grad callers to the new API signature

bigwhite37:
### PR Category

  Custom Device

  ### PR Types

  Bug fixes

  ### Description

  - Update the XPU activation grad functor to call the revised relu_grad wrapper that now consumes only activation output and dout
  - Align fused GEMM epilogue and fused feedforward XPU kernels with the new ReLU grad API signature


dynamicheart (approved):
LGTM

PR #75424, opened 2025-09-21
fix cuda api compatibility issue

youge325:
PR Category
Environment Adaptation

PR Types
Bug fixes

DescriptionUpdate cudaGraphGetEdges calls to 5-parameter signature


youge325:
wrong, I need to modify again

PR #75446, opened 2025-09-22
[Distributed] Add PipelineDatasetPreprocessor to aviod mem leaks in pipeline parallel

zhangyuqin1998:
### PR Category

Distributed Strategy

### PR Types

Improvements

### Description

Pass a local function to forward_backward_pipeline instead of the dataset itself. This prevents the dataset from being passed as a direct argument to forward_backward_pipeline, which would create additional reference counts that cannot be cleared, leading to GPU memory leaks.

pcard-76459

### DIFF
--- a/python/paddle/distributed/fleet/meta_parallel/pipeline_parallel.py
+++ b/python/paddle/distributed/fleet/meta_parallel/pipeline_parallel.py
@@ -122,6 +122,10 @@ class FakeMicroDataset:
         assert self._is_first_stage or self._is_last_stage
         micro_batch_data = self._load_micro_batch(self._index)
         self._index += 1
+
+        if self._index >= self._acc_steps:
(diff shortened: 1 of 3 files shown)


python/paddle/distributed/fleet/meta_parallel/pipeline_parallel.py
```diff
@@ -122,6 +122,10 @@ def __next__(self):
         assert self._is_first_stage or self._is_last_stage
         micro_batch_data = self._load_micro_batch(self._index)
         self._index += 1
+
+        if self._index >= self._acc_steps:
```
ForFishes, line 126: 上面不是已经raise StopIteration，这个判断是否有问题。

ForFishes (approved):
LGTM

PR #75477, opened 2025-09-23
Modified allocation type in kernels to support custom device

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

Modified allocation type in kernels to support custom device
pcard-67164


yongqiangma (approved):
LGTM

PR #75593, opened 2025-09-28
Fix misused pinned memory for custom device

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

Fix misused pinned memory for custom device.

card-67164


yongqiangma (approved):
LGTM

PR #75694, opened 2025-10-09
Fix enable_use_gpu cases for custom device

YqGe585:
### PR Category

Custom Device

### PR Types

Others

### Description

Fix enable_use_gpu cases for custom device

card-67164


yongqiangma (approved):
LGTM

PR #75961, opened 2025-10-20
Fix paddle.save error for custom device 

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

Fix paddle.save error for custom device. Not all custom device should use npu_identity branch.
pcard-67164


yongqiangma (approved):
LGTM

PR #76152, opened 2025-10-31
[API Compatibility] paddle.bucketize and paddle.searchsorted

zhengshengning:
### PR Category

Operator Mechanism

### PR Types

New features

### Description

**PR描述：**
**paddle. searchsorted：**
searchsorted新增side、out、sorter三个参数 和装饰器。
1. side：side (str|None, optional): The same as right but preferred. `left` corresponds to False for right and `right` corresponds to True for right. It will error if this is set to `left` while right is True. Default value is None.
2.  sorter (Tensor|None, optional): if provided, a tensor matching the shape of the unsorted `sorted_sequence` containing a sequence of indices that sort it in the ascending order on the innermost dimension.

**paddle.bucketize：**
bucketize加装饰器和out；

image

### DIFF
--- a/python/paddle/tensor/search.py
+++ b/python/paddle/tensor/search.py
@@ -1168,23 +1168,29 @@ def topk(
         return values, indices
 
 
+@param_two_alias(["x", "input"], ["sorted_sequence", "boundaries"])
 def bucketize(
     x: Tensor,
     sorted_sequence: Tensor,
     out_int32: bool = False,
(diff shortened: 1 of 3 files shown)


zhwesky2010 (approved):
LGTM

SigureMo (approved):

jeff41404 (approved):
LGTM

Issue #76161, opened 2025-11-01
Windows ninja 报错：error C2440: 'static_cast': cannot convert from 'InType' to 'OutType'


Manfredss:
### 问题描述 Issue Description

已按照流程 [windows 下从源码编译](https://www.paddlepaddle.org.cn/documentation/docs/zh/develop/install/compile/windows-compile.html#windows) setup，使用 ninja 编译会出现如下错误

<img width="1774" height="231" alt="Image" src="https://github.com/user-attachments/assets/0fbb1069-10c1-46c7-b55a-202c899959a1" />

cmake 命令：cmake .. -GNinja -DWITH_GPU=ON -DWITH_UNITY_BUILD=ON -DWITH_DISTRIBUTE=ON

是否为 MSVC 模板展开非法类型组合？

用了 Visual Studio 2022 Community 版，2019 的版本已无法下载

### 版本&环境信息 Version & Environment Information

****************************************
Paddle version: N/A
Paddle With CUDA: N/A

OS: Windows 11
GCC version: N/A
Clang version: 19.1.5
CMake version: version 3.18.6
Libc version: N/A
Python version: 3.12.12

CUDA version: 12.8.61
Build cuda_12.8.r12.8/compiler.35404655_0
cuDNN version: N/A
Nvidia driver version: 581.57
Nvidia driver List:
GPU 0: NVIDIA GeForce RTX 5070 Ti
****************************************

CPU: AMD Ryzen7 9800X3D
cuDNN: 8.9.7
系统环境：Windows 11 Home

PR #76171, opened 2025-11-02
Fix: #issue76161: error C2440 of 'static_cast': cannot convert from 'InType' to 'OutType'

Manfredss:
### PR Category
Environment Adaptation

### PR Types
Bug fixes

### Description
fix compile error in `paddle/phi/core/framework/data_type_transform.cc`

before:
```cpp
 template <typename InType, typename OutType>
 struct CastDataTypeFunctor {
   HOSTDEVICE inline OutType operator()(InType in) const {
     return static_cast<OutType>(in);
  }
};
```

after:
```cpp
template <typename InType, typename OutType>
struct CastDataTypeFunctor {
  HOSTDEVICE inline OutType operator()(InType in) const {
    #if defined(_MSC_VER)
        // 屏蔽 MSVC 不支持的从 float/bfloat8/float16 -> complex 的转换
        if constexpr (
            (std::is_same_v<OutType, phi::dtype::complex<float>> ||
            std::is_same_v<OutType, phi::dtype::complex<double>>) &&
            (std::is_same_v<InType, phi::dtype::float8_e4m3fn> ||
            std::is_same_v<InType, phi::dtype::float8_e5m2> ||
            std::is_same_v<InType, phi::dtype::bfloat16> ||
            std::is_same_v<InType, phi::dtype::float16>)) {
          return OutType(0);  // 默认返回值，仅防止编译错误
        } else
    #endif
        {
          return static_cast<OutType>(in);
        }
      }
};
```


Manfredss:
对 #76161 的修复

PR #76238, opened 2025-11-05
[Custom Device] Fix beam search decode kernel

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

Fix beam search decode kernel for custom device


yongqiangma (approved):
LGTM

PR #76305, opened 2025-11-07
[Auto-Parallel] adapt optimizer_sharded_state_dict in FlexCheckpoint

Xing-lil:
### PR Category

Auto Parallel

### PR Types

Others

### Description

Adapt optimizer_sharded_state_dict for FlexCheckpoint in auto_parallel:

- The local_tensor must keep the same name with the original tensor. Otherwise, the static_to_struct_mapping will be wrong.
- Properly build the ShardedWeight in auto_parallel.
- Add test in the ut.

card-92763


liym27 (approved):
LGTM

PR #76485, opened 2025-11-19
Replace np.asin with np.arcsin to support numpy1.x

YqGe585:
### PR Category

Custom Device

### PR Types

Bug fixes

### Description

Replace np.asin with np.arcsin to support numpy1.x


yongqiangma (approved):
LGTM

PR #76651, opened 2025-11-27
[XPU] fix zero tensor in arange, expand and mask_select op

ZhangX-21:
### PR Category
Custom Device

### PR Types
Bug fixes

### Description
fix zero tensor in arange, expand and mask_select op in xpu
Pcard-90546


yongqiangma (approved):
LGTM

